PR #4978, opened 2014-07-10
adding xdot to rosdep.

stonier:
Finally had to fork and do it manually. Seems github's editor causes whitespace problems.


k-okada:
this suppress `ros-indigo-xdot` -> https://github.com/ros/rosdistro/pull/3927,

- should we check within travis ?
- should we remove ros-indigo-xdot?

PR #6334, opened 2014-11-24
OS X Yosemite (10.10) Rosdep entries for vtk and gazebo2

JeremyCade:
#### Homebrew VTK Update

Recent updates to the homebrew/science/vtk brew have created issues for ROS Indigo install on OSX. 
This can be fixed by removing the **--python** and **--qt** options and replacing them with **--with-qt** on the vtk homebrew dependency.
#### ROS & Gazebo

As noted on the [Which combination of ROS/Gazebo versions to use](http://gazebosim.org/tutorials?tut=ros_wrapper_versions&cat=connect_ros) page, ROS Indigo is only compatible with the Gazebo 2.x branch.

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -98,7 +98,7 @@ freetype:
 gazebo:
   osx:
     homebrew:
-      packages: [gazebo]
+      packages: [gazebo2]
 gccxml:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -336,7 +336,7 @@ libv4l-dev:
 libvtk:
   osx:
     homebrew:
-      options: [--python, --qt]
+      options: [--with-qt]
       packages: [vtk]
 libvtk-java:
   osx:


vrabaud:
@wjwwood , what should be do with that one ?

wjwwood:
I should figure out what should be done :stuck_out_tongue:. I don't know. Merging this will cause problems for people using ROS hydro on OS X, but not merging it will cause problems for Indigo.

dirk-thomas:
I think we have to make a call on this. Either decide to "break" on one distro (where I would prefer to break Hydro and make it work well on Indigo) or do the necessary changes to support both.

wjwwood:
I have a plan which allows users to pick different dependencies based on the Homebrew tap they choose, but I haven't had time to implement it.

Basically we'd make the rosdep rule for `gazebo` resolve to the `gazebo_for_ros` Homebrew formula. When installing hydro you'd "tap" the `ros/hydro-deps` tap which contains a version of the `gazebo_for_ros` formula which depends on `gazebo`, but the `ros/indigo-deps` tap will have a different version of the `gazebo_for_ros` formula which depends on `gazebo5`. This would be applied to all the affected formula. This way rosdep always resolves to the same key, but then we can change what the key ends up installing based on which Homebrew tap the user selects. It still doesn't allow a side by side setup with Hydro and Indigo, but at least people can choose.

This will take quite some work to implement, so I'm not sure when I'll get to it.

/cc @scpeters

scpeters:
Sounds reasonable. There are separate formula for each gazebo version (`gazebo` for 1.9, `gazeboN` for each release after) but they cannot be `brew link`ed at the same time.

dirk-thomas:
Should we aim to address this before `Jade` is being released?

scpeters:
We added [new rosdep keys](https://github.com/ros/rosdistro/pull/8243) for gazebo5 in jade. So I think this is currently just a conflict between hydro / indigo, which have the same rosdep key for different versions of gazebo.

wjwwood:
Yeah, this shouldn't be trouble for Jade. I was mistaken above about gazebo5 and indigo, but this still applies for ogre which is 1.7 in hydro and 1.8/9 in indigo and jade.

dirk-thomas:
Can we close this and if necessary convert it into a ticket?

wjwwood:
Closing in favor of: https://github.com/ros/rosdistro/issues/8912

PR #6419, opened 2014-12-03
brew formula changes for vtk

Juxi:
the homebrew formula for `vtk` does not have the `--python` option anymore.
Because of this the installation fails, by removing the option `vtk` installs

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -340,7 +340,7 @@ libv4l-dev:
 libvtk:
   osx:
     homebrew:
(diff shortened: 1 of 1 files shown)


ahundt:
I've run into this identical issue. Where can I find the files to include this patch myself? I'm trying to build indigo.

wjwwood:
`rosdep update` should be sufficient to get this "fix". The question you posted elsewhere (https://github.com/ros-infrastructure/rosdep/issues/363#issuecomment-70604030) is a similar but different failure. Looks like they have changed the `vtk` formula again or you've already installed it with different options.

PR #6739, opened 2015-01-03
Combined fix for unicode and include fedora 21+ fixes

### LINKED ISSUE #6637: update unidiff test usage to read character encoding
Fedora 19 codename has unicode. Discovered here: https://github.com/ros/rosdistro/pull/6619 

Work around here: https://github.com/matiasb/python-unidiff/issues/12

tfoote:
Replaces #6738  and fixes #6637

### DIFF
--- a/test/test_url_validity.py
+++ b/test/test_url_validity.py
@@ -3,7 +3,7 @@
 from __future__ import print_function
 
 
-from io import BytesIO
+from io import StringIO
 import os
 import subprocess
 import yaml
@@ -51,7 +51,9 @@ def detect_lines(diffstr):
     """Take a diff string and return a dict of
     files with line numbers changed"""
     resultant_lines = {}
-    io = BytesIO(diffstr)
+    # Force utf-8 re: https://github.com/ros/rosdistro/issues/6637
+    encoding = 'utf-8'
+    io = StringIO(unicode(diffstr, encoding))
     udiff = unidiff.PatchSet(io)
     for file in udiff:
         target_lines = []
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1658,7 +1658,13 @@ libpopt-dev:
 libpq-dev:
   arch: [postgresql-libs]
   debian: [libpq-dev]
-  fedora: [derelict-postgresql-devel, postgresql-devel]
+  fedora:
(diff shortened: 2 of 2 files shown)


test/test_url_validity.py
```diff
@@ -3,7 +3,7 @@
 from __future__ import print_function
 
 
-from io import BytesIO
+from io import StringIO
```
dirk-thomas: Since the unit tests run with Python 2 this should not use the `io` package but `cStringIO` instead (http://wiki.ros.org/python_2_and_3_compatible_code#StringIO).

k-okada:
Hi, I got
```
Could not resolve rosdep key 'libpq-dev' for distro '24':
No definition of [libpq-dev] for OS version [24]
        rosdep key : libpq-dev
        OS name    : fedora
        OS version : 24
        Data: arch:
- postgresql-libs
debian:
- libpq-dev
fedora:
  '21':
  - derelict-PQ-devel
  - postgresql-devel
  '22':
  - derelict-PQ-devel
  - postgresql-devel
  beefy:
  - derelict-postgresql-devel
  - postgresql-devel
  heisenbug:
  - derelict-postgresql-devel
  - postgresql-devel
  "schr\xF6dinger\u2019s":
  - derelict-postgresql-devel
  - postgresql-devel
  spherical:
  - derelict-postgresql-devel
  - postgresql-devel
gentoo:
  portage:
    packages:
    - dev-libs/libpqxx
ubuntu:
- libpq-dev

Failed to resolve libpq-dev on fedora:24 with: Error running generator: Failed to resolve rosdep key 'libpq-dev', aborting.
libpq-dev is depended on by these packages: ['euslisp']
<== Failed
You can try to address the issues which appear above and try again if you wish.

```
error message,

Should I update bloom? or change rosdistro file to use a version number (24?) instead of code name?

tfoote:
This just means that there are no rules for fedora '23' and '24' for libpq-dev: https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L2017-L2030

It looks like postgresql-devel is the same key in 23 and 24: https://apps.fedoraproject.org/packages/postgresql-devel 

Same with derelict-PQ-devel: https://apps.fedoraproject.org/packages/derelict-PQ-devel

So the `22` line can just be duplicated for `23` and `24`. 

@cottsay FYI

PR #6741, opened 2015-01-04
cleaning up imports as mentioned 

tfoote:
at https://github.com/ros/rosdistro/pull/6739/files#r22433748

@dirk-thomas

### DIFF
--- a/test/test_url_validity.py
+++ b/test/test_url_validity.py
@@ -3,19 +3,24 @@
 from __future__ import print_function
 
 
-from io import StringIO
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import StringIO
 import os
 import subprocess
 import yaml
 from yaml.composer import Composer
 from yaml.constructor import Constructor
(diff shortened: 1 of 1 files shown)


test/test_url_validity.py
```diff
@@ -1,15 +1,19 @@
 #!/usr/bin/env python
 
+
+
```
dirk-thomas: These additional white spaces seem to not follow PEP8.

test/test_url_validity.py
```diff
@@ -1,15 +1,19 @@
...
 from __future__ import print_function
 
 
-from io import StringIO
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import StringIO
```
dirk-thomas: Since the script is currently incompatible with Python 3 anyway the exception handling is not useful.

tfoote: It might as well be partially there. I fixed the other whitespace.

PR #6750, opened 2015-01-05
Added termcolor and netaddr to python rosdep

mitchellwills:


wjwwood:
Thanks, consider adding entries for `pip` in a follow up pull request.

mitchellwills:
Should the pip entries replace the ones I added here? I couldn't find any great documentation for it.

wjwwood:
@mitchellwills no, they would go along side the ubuntu entry and under the different operating system's which do not have `apt-get` like OS X. Just look at some of the other entries which have both, for example: https://github.com/ros/rosdistro/blob/98f60ee9bcd3e954c85195e6f9a5b8bf1b30860c/rosdep/python.yaml#L759-765

mitchellwills:
Is it better to use apt-get dependencies where they exist or to just use pip for everything like python-cpplint does?
https://github.com/ros/rosdistro/blob/98f60ee9bcd3e954c85195e6f9a5b8bf1b30860c/rosdep/python.yaml#L249-L261

wjwwood:
You should prefer `apt-get` for `ubuntu` and `debian` over `pip` because we cannot generate `.debs` for package which depend on things which do not have an `apt-get` entry. So if you replace the `apt-get` with `pip` on `ubuntu` then we could not create binaries for any package which used the key. On other operating systems, however, having the pip entry will make building from source easier.

PR #6753, opened 2015-01-06
Updated termcolor and netaddr python rosdep to include additional platforms via pip

mitchellwills:
In response to discussion in #6750

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -601,7 +601,40 @@ python-multicast:
   ubuntu:
     pip: [py-multicast]
 python-netaddr:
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for paying it forward, that looks much more complete.

P.S. In case you're curious, we have plans to improve the syntax in cases like these for the next generation tool: http://xylem.readthedocs.org/en/latest/spec.html#version-ranges-and-geq

mitchellwills:
Awesome. I always thought it was weird to have ROS specific dep management for a supposedly ROS agnostic tool

PR #6776, opened 2015-01-07
myo_ros: 0.1.2-0 in 'indigo/distribution.yaml' [bloom]

rgariepy:
Increasing version of package(s) in repository `myo_ros` to `0.1.2-0`:
- upstream repository: https://github.com/clearpathrobotics/myo_ros.git
- release repository: https://github.com/clearpath-gbp/myo_ros-release.git
- bloom version: `0.5.16`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -3533,6 +3533,18 @@ repositories:
       url: https://github.com/ual-arm-ros-pkg/mvsim.git
       version: master
     status: maintained
+  myo_ros:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/clearpath-gbp/myo_ros-release.git
+      version: 0.1.2-0
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/myo_ros.git
+      version: master
+    status: maintained
+    status_description: New release. Adding examples planned, but low-priority.
   nanomsg:
     release:
       tags:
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -3533,6 +3533,18 @@ repositories:
...
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/clearpath-gbp/myo_ros-release.git
+      version: 0.1.2-0
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/myo_ros.git
+      version: master
+    status: maintained
+    status_description: New release. Adding examples planned, but low-priority.
```
dirk-thomas: Please remove the status description since it is not intended for such information.

dirk-thomas:
Please also comment on a PR when making further commits in the future since only comments trigger notification emails.

indigo/distribution.yaml
wjwwood: @rgariepy Thanks for updating the pull request. Just to follow up on @dirk-thomas's request, this is what he is talking about:

From http://www.ros.org/reps/rep-0137.html#release-file
> status_description: an optional field describing in a short sentence the current status of the repository. For example detailing the reason for EOL and the recommended upgrade path.

Though it doesn't restrict using this field for anything, I believe the intention was that this should only really be used when detailing why something is EOL and what users should do instead.

rgariepy:
@dirk-thomas Thanks for checking on this, I'm used to our internal setup.

PR #6847, opened 2015-01-13
pcan_topics: 1.0.11-0 in 'indigo/distribution.yaml' [bloom]

najkirdneh:
Increasing version of package(s) in repository `pcan_topics` to `1.0.11-0`:
- upstream repository: https://github.com/najkirdneh/pcan_topics.git
- release repository: https://github.com/najkirdneh/pcan_topics-release.git
- bloom version: `0.5.16`
- previous version for package: `1.0.9-0`
## pcan_topics
- No changes

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -4487,7 +4487,7 @@ repositories:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/najkirdneh/pcan_topics-release.git
-      version: 1.0.9-0
+      version: 1.0.11-0
     source:
       type: git
       url: https://github.com/najkirdneh/pcan_topics.git


najkirdneh:
Dear Mr. Woodall,

I still have one Jenkins build which fails:

Build failed in Jenkins: devel-indigo-pcan_topics » amd64,trusty,devel #3

The other builds have went back to normal. Looking at the failure report it
seems odd that it cannot install the packages it depends on while other
build can. Should this build also be back tot normal?

Kind regards,

Hendrik

Merged #6847 <https://github.com/ros/rosdistro/pull/6847>.

—
Reply to this email directly or view it on GitHub
<https://github.com/ros/rosdistro/pull/6847#event-218390907>.

wjwwood:
So that is a devel job, which happens on commit to your repository listed in your `source` entry in this file. Those jobs are meant to let you know when your commits break things. There is currently no way to opt-out of those, but we are working on that. If you don't want them you'll have to remove the source entry using a manually created pull request.

Without the actual error from the job, or a link to the job, I'm not sure why the job failed, but it may be due to something unrelated to your package.

The binary jobs are what you really care about and you can track their status here: http://www.ros.org/debbuild/indigo.html?q=pcan_topics and it looks like they are all passing, but none of them have made it into the public debian repository yet. You can read more about the different repositories here: http://wiki.ros.org/ShadowRepository

PR #6850, opened 2015-01-14
ceres_solver: 0.10.0-1 in 'indigo/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `ceres_solver` to `0.10.0-1`:
- upstream repository: https://github.com/stonier/ceres_solver.git
- release repository: https://github.com/yujinrobot-release/ceres_solver-release.git
- bloom version: `0.5.16`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -571,6 +571,13 @@ repositories:
       url: https://github.com/ros/catkin.git
       version: indigo-devel
     status: maintained
+  ceres_solver:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/yujinrobot-release/ceres_solver-release.git
+      version: 0.10.0-1
+    status: maintained
   class_loader:
     doc:
       type: git


mikeferguson:
@stonier  any reason not to use the upstream Debian package that I released of ceres?

stonier:
Aha....don't know whether to thank you or thump you :neckbeard:

Wasn't around when we started so we've been source installing for months. Didn't expect it to magically appear - that's a nice surprise!

It looks good enough for our immediate needs...if we start requiring extra in a release (especially like the sparse suite) I'll connect with you then.

From what I can gather, you are doing this in a ppa, then serving it on the ros apt repo via the [https://github.com/ros-infrastructure/reprepro-updater/blob/master/config/ceres.upstream.yaml](reprepo updater)?

I'm also about to build a release for sophus...would you recommend doing it this way or just go with a bloom 3rd party build?

mikeferguson:
Yes, it's via PPA + reprepro updater -- I released into Indigo in May or so of this year I believe. PPA is here: https://launchpad.net/~mikeferguson/+archive/ubuntu/ceres-solver

It's a toss up right now as to PPA vs bloom 3rd party. At first, the Bloom docs were lacking for 3rd party stuff and i ran into a bunch of problems. I recently released SBPL into Indigo via the Bloom 3rd Party approach and it finally worked right. One issue with the PPA release is that Ubuntu has been tearing down the older buildfarms quite quickly and we on ROS have continued to build for non-LTS releases. This was the primary reason I went with Bloom for SBPL, is that I couldn't release to Saucy anymore.

stonier:
I'll go with bloom for sophus as a 3rd party in the short term then as I'm familiar with that.

I can imagine doing a ppa and hooking them via reprepo-updater might be good to get it out there quickly (no intermediate ppa step needed for ros packages) while going through the slow process of getting it officially into ubuntu upstream.

stonier:
Thanks mate!

tfoote:
I would recommend blooming 3rd party stuff for now. We originally couldn't support it and pushed hard to get things upstream into PPAs. However, in that process there is a lot of pain as we start to lose control of the process and cannot make high impact changes. An example is urdfdom where it got upstream, but they didn't get the latest version into precise and then we're stuck with ifdefing and working around the shortcomings of the older version. 

Blooming 3rdparty is a nice middle ground keeping control, and enabling automation, but sticking to the standard build style. 

Another evolution is to look at building our own PPA style repo for ROS distro independent packages. (Note bloom has support for non ROS packages, but it's less well tested.)

mikeferguson:
@tfoote We might want to send your comment, or a link to this thread, to the maintainers/packagers mailing list -- as it is a departure from our earlier stance a while ago that suggesting upstreaming 3rd party repos.

PR #6860, opened 2015-01-14
add ecto_image_pipeline and image_pipeline to jade

vrabaud:


wjwwood:
@vrabaud can you please use the `bloom-release` command when you release? You can still stop before submitting the pull request if you like and make combined ones like this, but since you don't run that then there is no history of what you've done stored in the README of the release repository:

https://github.com/ros-gbp/ecto_image_pipeline-release

Here is an example of a repository which does have a README because `bloom-release` was used:

https://github.com/ros-gbp/rviz-release/blob/master/README.md

Thanks.

vrabaud:
sounds good.

PR #6868, opened 2015-01-15
Rosdep entry for nvidia-cuda-toolkit on ubuntu and debian

NiziL:
Well, all in the title. Here is a new rosdep entry for the nvidia-cuda-toolkit package.


vrabaud:
debian needs to be before ubuntu (alphabetical order)

PR #6941, opened 2015-01-21
Fixing some old indexes

jbohren:
I guess you guys found out what I'm doing, no sense in putting this off.


tfoote:
Thanks for the updates.

PR #6988, opened 2015-01-26
rosdep: split fltk into fltk-dev and fluid

wjwwood:
This splits the `fltk` rosdep key into a dev and runtime part, `libfltk-dev` and `fluid` respectively. This is to resolve https://github.com/ros-simulation/stage_ros/pull/9. I'll leave `fltk` for backwards compatibility.


dirk-thomas:
Should we add a comment to the `fltk` entry to point users to the new keys?

wjwwood:
Comment added.

PR #6993, opened 2015-01-27
add percol

k-okada:
we need this for for @garaemon 's 'ROS + percol integration

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -799,6 +799,19 @@ python-pep8:
     pip:
       packages: [pep8]
   ubuntu: [pep8]
+python-percol:
+  debian:
+    pip:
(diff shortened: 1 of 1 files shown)


wjwwood:
Alphabetical ordering is wrong, it would at least come after `python-pep8`.

Also, be aware that our build farm cannot create a Debian package when you depend on a key which uses `pip` on Ubuntu. So we cannot build any `.deb` packages for ROS packages which depend on this key. However, it's still useful when using the `rosdep` CLI, so we allow them.

k-okada:
thanks fixed

dirk-thomas:
The new version of `jsk_tools` uses this rosdep key and therefore fails to build: https://github.com/ros/rosdistro/pull/7074

PR #7006, opened 2015-01-29
New Repository For Indexing: foscam_8918_driver

tdenewiler:


hydro/distribution.yaml
```diff
@@ -1875,6 +1875,11 @@ repositories:
       type: git
       url: https://github.com/kth-ros-pkg/force_torque_tools.git
       version: hydro
+  foscam_8918_driver:
+    doc:
+      type: git
+      url: https://github.com/tdenewiler/foscam_8918_driver
```
wjwwood: This should probably have a trailing `.git`.

hydro/distribution.yaml
```diff
@@ -1875,6 +1875,11 @@ repositories:
...
+      version: hydro-devel
```
wjwwood: The CI is failing because this has trailing whitespace, can you please remove it?

tdenewiler: Thanks, I couldn't quite figure that out from the Travis output. Fixed in b9a85fe.

PR #7015, opened 2015-01-29
rxparamedit: 1.0.1-0 in 'hydro/distribution.yaml' [bloom]

dornhege:
Increasing version of package(s) in repository `rxparamedit` to `1.0.1-0`:
- upstream repository: https://github.com/dornhege/rxparamedit.git
- release repository: https://github.com/dornhege/rxparamedit-release.git
- distro file: `hydro/distribution.yaml`
- bloom version: `0.5.16`

### DIFF
--- a/hydro/distribution.yaml
+++ b/hydro/distribution.yaml
@@ -7615,6 +7615,21 @@ repositories:
       url: https://github.com/tork-a/rwt_ros.git
       version: hydro-devel
     status: developed
+  rxparamedit:
+    doc:
+      type: git
+      url: https://github.com/dornhege/rxparamedit.git
+      version: hydro-devel
+    release:
+      tags:
+        release: release/hydro/{package}/{version}
+      url: https://github.com/dornhege/rxparamedit-release.git
+      version: 1.0.1-0
+    source:
+      type: git
+      url: https://github.com/dornhege/rxparamedit.git
+      version: hydro-devel
+    status: maintained
   s3000_laser:
     doc:
       type: git


dirk-thomas:
Does `rx` in the package name make sense if it is actually Qt-based?

dornhege:
I'm following the quite old naming scheme from the pre-groovy time. I don't mind changing to rqt_paramedit, although this package does not provide rqt_gui integration.

dirk-thomas:
May be `qtparamedit` then? Optionally with some underscores for readability?

Would you be interested to convert it into a rqt plugin? Since it is already using Qt it should be fairly straight forward.

dornhege:
There is actually an older one rosgui_paramedit derived from this package. Isn't there a parameter plugin for rqt? I couldn't find something by a quick check and I usually use this standalone tool. If something like this is needed I could also convert that.

dirk-thomas:
Actually yes, there is `rosgui_paramedit`. May be it would make sense to merge your package with that?

Any rqt plugin can be used in standalone mode.

dornhege:
Do you mean this: http://wiki.ros.org/rosgui_paramedit ? Because that is mine using rxparamedit internally. If that is the only one available for rqt I'll port that to hydro.

In that case please reject this pull request as I'll do that first and rename the repositories before releasing.

dirk-thomas:
Yes, this is the only parameter editor in rqt currently (as far as I know).
Thanks for working on it!

Closing this.

PR #7053, opened 2015-02-02
yaml_cpp_0_3: 0.3.0-0 in 'indigo/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `yaml_cpp_0_3` to `0.3.0-0`:
- upstream repository: https://github.com/stonier/yaml_cpp_0_3.git
- release repository: https://github.com/yujinrobot-release/yaml_cpp_0_3-release.git
- bloom version: `0.5.16`
## yaml_cpp_0_3

```
* renamespaced library and headers (yaml->yaml-0.3) to avoid system v0.5 conflict
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8087,6 +8087,17 @@ repositories:
       url: https://github.com/rohbotics/xv_11_laser_driver.git
       version: indigo-devel
     status: maintained
+  yaml_cpp_0_3:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/yujinrobot-release/yaml_cpp_0_3-release.git
+      version: 0.3.0-0
+    source:
+      type: git
+      url: https://github.com/stonier/yaml_cpp_0_3.git
+      version: indigo
+    status: maintained
   yocs_msgs:
     doc:
       type: git


vrabaud:
We usually try not to re-add third party and try to use the upstream ones. I'll let @tfoote take a decision on that one.

stonier:
Of course - I usually try and do the same.

The why of it - system yaml_cpp (v0.5) now uses boost. This was rather controversial as yaml_cpp was  a favourite for very embedded projects. The author ultimately decided to keep the boost dependency, but still maintains a [v0.3 release](https://code.google.com/p/yaml-cpp/) for embedded projects.

We use yaml_cpp in cross-compiled projects (via catkin) that do not use boost (nor roscpp and beyond). There is simply not any way to do this with the system yaml cpp.

Note that this package has headers and libraries namespaced appropriately to avoid conflicts.

vrabaud:
Totally makes sense. Wouldn't creating a libyaml-cpp0.3 key in rosdep work ? Pointing to http://packages.ubuntu.com/trusty/libyaml-cpp0.3

stonier:
No - that one can't be installed alongside (isn't renamed). i.e. it conflicts with the 0.5 version. And it doesn't help with a cross compiled project either - having it as a simple catkin project lets us build a deb for it when testing on the desktop and also provides sources that catkin cross compilations can consume with ease.

tfoote:
It's unfortunate, but I think this might be the best approach. With the full renamspacing it should not collide with symbols.

PR #7067, opened 2015-02-02
baxter_interface: 1.1.0-0 in 'indigo/distribution.yaml' [bloom]

ghost:
Increasing version of package(s) in repository `baxter_interface` to `1.1.0-0`:
- upstream repository: https://github.com/RethinkRobotics/baxter_interface.git
- release repository: https://github.com/RethinkRobotics-release/baxter_interface-release.git
- bloom version: `0.5.16`
## baxter_interface

```
* Updates baxter_interface to ROS Indigo
* Upgrades to the Indigo robot required a reworking of how cameras power **on** / **off** - Three cameras can no longer be powered at the same time, and closing cameras turns power **off** to the specified camera and **on** to the other two
* Updates joint trajectory action server to use new Inverse Dynamics Feed Forward Commands and configurations for smoother & more accurate MoveIt trajectory execution
* Updates joint trajectory action server default control mode from *position* to *position_w_id* which uses 'raw' joint position control
* Updates joint trajectory action server to replace linear, cubic, and quintic spline fitting with *Cubic Bezier Spline* interpolation. These splines can use Position, Velocity, and/or Acceleration to more accurately interpolate supplied trajectories
* Adds head trajectory action server interface for controlling Baxter's head pan joint (thanks to @aginika for contributing this)
* Updates verification of Gripper software versions to check firmware Build Date instead of Build Version (SDK gripper version 1.1 conflicted with Manufacturing versions)
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -300,6 +300,13 @@ repositories:
       url: https://github.com/RethinkRobotics-release/baxter_common-release.git
       version: 1.1.0-1
     status: developed
+  baxter_interface:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/RethinkRobotics-release/baxter_interface-release.git
+      version: 1.1.0-0
+    status: developed
   bfl:
     release:
       tags:


rethink-imcmahon:
I'm not sure if there's a better way to do this... `baxter_core_msgs` is a core dependency among all our packages, which prevents me from being able to set up any other package until `baxter_common/baxter_core_msgs` is merged. `bloom-release` gives an error saying that  the rosdep could not be satisfied or something to that effect. 
Similarly, I will need to wait for `baxter_interface` (which depended on `baxter_core_msgs`) to be merged before `baxter_examples` and `baxter_tools`. Finally, after everything else is merged, `baxter` repo can be bloom-released. Granted, this is only because it's the first time we are releasing these package to Indigo, but I imagine this not ideal for you at OSRF to have to merge every 5-10 min rather than all in one go.

dirk-thomas:
This is the only way for a first time release. Don't worry about the number of PRs.

wjwwood:
Yeah, unfortunately bloom only works off of what is released when generating the `.deb` dependencies, so you'll have to release them one at a time for the first release.

PR #7077, opened 2015-02-03
Removed pr2_ethercat_drivers from distribution.yaml. Will fix and releas...

UltronDestroyer:
Regarding actions to be taken from this issue https://github.com/PR2/pr2_ethercat_drivers/issues/68, pr2_ethercat_drivers will be removed. Once tested in Indigo and eml is released as a third party, this package will re-release. Will this effect packages that depend on these catkin packages being inside the distribution.yaml? Do those also have to be removed?

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -5060,26 +5060,6 @@ repositories:
       url: https://github.com/pr2/pr2_controllers.git
       version: indigo-devel
     status: maintained
-  pr2_ethercat_drivers:
-    doc:
-      type: git
-      url: https://github.com/pr2/pr2_ethercat_drivers.git
-      version: indigo-devel
-    release:
-      packages:
-      - eml
-      - ethercat_hardware
-      - fingertip_pressure
-      - pr2_ethercat_drivers
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/pr2-gbp/pr2_ethercat_drivers-release.git
-      version: 1.8.13-1
-    source:
-      type: git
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Any package depending on it will fail in the mean time. You should keep the time window as sorry as possible. Definitely within one sync window so that the regression does not show up in the public repo.

Is that ok?

TheDash:
Thanks dirk. What's the sync window time?

dirk-thomas:
That depends :wink: Tully usually announces syncs to the public repo every few weeks. The last one for Indigo was on Jan 26. So if you can finish this in one/two weeks that would be good. Then nobody using the public repo will be affected.

dirk-thomas:
Do you expect to get everything working again with one/two weeks?

dirk-thomas:
Btw. `eml` is continuously failing to build on the farm.

TheDash:
Yes, that's right. I was going to do a re-release of eml today. I believe
all that needs to be done is already done but the bloom release is being
done.

On Wed, Feb 4, 2015 at 1:59 PM, Dirk Thomas <notifications@github.com>
wrote:

> Btw. eml is continuously failing to build on the farm.
>
> —
> <https://github.com/ros/rosdistro/pull/7077#issuecomment-72915745>.
>



-- 

Devon Ash
<a href="http://ca.linkedin.com/pub/devon-ash/48/478/981"
style="text-decoration:none;"><span style="font: 80% Arial,sans-serif;
color:#0783B6;">View Devon Ash's LinkedIn profileView Devon Ash's profile</span></a>

PR #7097, opened 2015-02-05
I corrected the bugs on the url.

khansari:


wjwwood:
@khansari this is still not correct, did you run `bloom-release` at any point? Please follow the tutorial on releasing a package for the first time:

http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

PR #7099, opened 2015-02-05
Update distribution.yaml

furushchev:
add jsk_smart_gui only for doc


vrabaud:
the repo does not exists.

furushchev:
sorry i mistook repo name.

PR #7115, opened 2015-02-06
Update distribution.yaml

rbirac:


indigo/distribution.yaml
```diff
@@ -7677,6 +7677,21 @@ repositories:
       url: https://github.com/ros-drivers/um6.git
       version: indigo-devel
     status: maintained
+  um7:
+    doc:
+      type: git
+      url: https://github.com/ros-drivers/um7.git
+      version: indigo-devel
```
dirk-thomas: The repository does not have an `indigo-devel` branch.

indigo/distribution.yaml
```diff
@@ -7677,6 +7677,21 @@ repositories:
...
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
```
dirk-thomas: The repository is empty.

dirk-thomas:
I will close this PR for now. Please comment here if you fixed the mentioned problems and it can be reopened or just use bloom feature to create a new PR for you.

PR #7125, opened 2015-02-06
(Indigo, rtshell pkgs) Add doc, source.

130s:


dirk-thomas:
All repository URLs are the same?

130s:
Yes they are stored in the same repository. I thought it was okay since release job seemed to work fine, but is this not the case for doc and source jobs?

dirk-thomas:
Each documentation job would actually document the same - the content of the repository - which does not make sense.

The same for the source entries. A user clone more than one will have duplicate package names in the workspace.

Regarding the release repositories: it looks like you have released every package into a separate release repository. That is not the recommended process for releasing. Therefore I can't tell you if that will work - now or in the future. Is there a reason why you did this differently than it is recommended?

130s:
Ok, thanks for the clarification. I'll discuss with the co-maintainer who actually made this change (these packages used to be in the same meta pkg until Hydro).

tfoote:
The distribution file should follow the structure of the repo, like in hydro. This is independant of whether there is a metapackage or not. The metapackage and the repo name don't need to match. that's mostly a legacy thing where the repo used to be a ros stack. You should just use the same structure as indigo, but remove the rtshell_core package from the list if it's been removed in indigo. 

PS bloom should do all of this for you.

dirk-thomas:
@130s Please also create a new release repo (for all packages of that source repo) and remove the separate ones from the distribution file.

130s:
Thanks for the instruction. The issue was addressed in https://github.com/ros/rosdistro/pull/7163, https://github.com/ros/rosdistro/pull/7160, and https://github.com/ros/rosdistro/pull/7161 by @k-okada.

PR #7150, opened 2015-02-09
fixing rosdep typos

jbohren:


tfoote:
Thanks for the fixes.

jbohren:
I do think "Ox OS" sounds like a cool name for an operating system... "OxOS"... "0xOS"...

PR #7152, opened 2015-02-09
rosdep: fixing osxbrew -> osx platform key

jbohren:
This is the only one like it, is this a valid platform key?


wjwwood:
Uh, it used to be, when you could do OS X with Homebrew or Macports, they were different operating systems. But since then Macports has done their own thing where the ros packages are bundled in with MacPorts.

PR #7159, opened 2015-02-10
remove rtshell_core

k-okada:


tfoote:
Why are you removing this from hydro?

k-okada:
we want to switch to 
http://github.com/tork-a/rtshell-release
http://github.com/tork-a/rtsprofile-release
http://github.com/tork-a/rtctree-release

dirk-thomas:
Can't you change the existing entry instead?

k-okada:
each upstream (rtshell/rtsprofile/rtctree) have different upstream url so I think we need different release repo for each of them and that need different entry, I think.

PR #7192, opened 2015-02-10
pr2_kinematics: 1.0.7-1 in 'indigo/distribution.yaml' [bloom]

UltronDestroyer:
Increasing version of package(s) in repository `pr2_kinematics` to `1.0.7-1`:
- upstream repository: https://github.com/PR2/pr2_kinematics.git
- release repository: https://github.com/pr2-gbp/pr2_kinematics-release.git
- bloom version: `0.5.18`
## pr2_arm_kinematics

```
* Fix pr2_arm_kinematics for indigo by adding cmake_modules
* Add dependency to cmake_modules to solve Eigen depdency on indigo
* Contributors: Ryohei Ueda
```
## pr2_kinematics
- No changes

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -5247,6 +5247,24 @@ repositories:
       url: https://github.com/PR2/pr2_hack_the_future.git
       version: hydro-devel
     status: maintained
+  pr2_kinematics:
+    doc:
+      type: git
+      url: https://github.com/pr2/pr2_kinematics.git
+      version: hydro-devel
+    release:
+      packages:
+      - pr2_arm_kinematics
+      - pr2_kinematics
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/pr2-gbp/pr2_kinematics-release.git
+      version: 1.0.7-1
+    source:
+      type: git
+      url: https://github.com/pr2/pr2_kinematics.git
+      version: hydro-devel
+    status: maintained
   pr2_make_a_map_app:
     doc:
       type: git


tfoote:
@TheDash you can also rerun bloom with --pull-request-only. It will be much quicker since it won't actually do the rerelease, just reopen the pull request.

TheDash:
I did run it with --pull-request-only. It hasn't been released into the distribution.yaml before, and the release repo has changed from hydro to indigo, so It asks me some things.

tfoote:
Huh, maybe it doesn't work for that case, I would have expected it not to
produce a debian increment in the version number.

wjwwood:
According to the README in the release repository you ran it twice without `--pull-request-only`, but I don't know if this pull request was the result of that last one or not:

https://github.com/pr2-gbp/pr2_kinematics-release/blob/master/README.md

TheDash:
Right. The data gleaned which lead me to the conclusion in my prior comment
was from making another bloom-release of a similar package (the
pr2_make_a_map one). I assumed that info would extrapolate here.

On Tue, Feb 10, 2015 at 3:22 PM, William Woodall <notifications@github.com>
wrote:

> According to the README in the release repository you ran it twice without
> --pull-request-only, but I don't know if this pull request was the result
> of that last one or not:
>
> https://github.com/pr2-gbp/pr2_kinematics-release/blob/master/README.md
>
> —
> <https://github.com/ros/rosdistro/pull/7192#issuecomment-73776163>.
>



-- 

Devon Ash
<a href="http://ca.linkedin.com/pub/devon-ash/48/478/981"
style="text-decoration:none;"><span style="font: 80% Arial,sans-serif;
color:#0783B6;">View Devon Ash's LinkedIn profileView Devon Ash's profile</span></a>

PR #7215, opened 2015-02-11
fix doc/source loction to  release package, they needs to be valid ROS package

k-okada:


dirk-thomas:
These release repos do not contain any ros packages on the master branch.

jade/distribution.yaml
```diff
@@ -462,7 +462,7 @@ repositories:
       version: 1.0.2-1
     source:
       type: git
-      url: https://github.com/euslisp/jskeus.git
+      url: https://github.com/tork-a/jskeus-release.git
       version: master
```
wjwwood: I don't know what you're doing here, but pointing to the master branch won't work.

Why are you patching in the `CMakeLists.txt` instead of putting it in the upstream repository?

k-okada:
sorry, fixed to release branch

dirk-thomas:
Can you please explain why you are not placing the CMake file into the source repository? Clearly a developer can not build the package after checking out the source repo??

k-okada:
the source repo is Make file project so ideally we should put both pakcage.xml and CMakeLists.txt in patch directory of release repository, but we put package.xml and CHANGELOG.rst to use catkin_generate_changelog,caktin_prepare_packgae.
It seems if we use `:{auto}` or something in the version tag of package.xml and put them in release repository, then bloom-release fill version tag automatically. is it correct? if so I can remove package.xml from source repo and put them in release repo

wjwwood:
> It seems if we use :{auto} or something in the version tag of package.xml and put them in release repository, then bloom-release fill version tag automatically. is it correct?

Yeah, that's what the patch to upstream thing is for in bloom.

But why don't you just put the `CMakeLists.txt` in the upstream? Just so other users don't get and confused and try to use CMake instead of Make? If you have control over the repository then what reason do you have for using Make over CMake?

I'm fine to merge this, but I'm just trying to get a handle on why this is necessary.

dirk-thomas:
We highly recommend to keep the `package.xml` file as well as the `CMakeLists.txt` file in the source repo. If you do not want to do that for whatever reason please also remove the `package.xml` file. Otherwise the repository claims to contain a CMake package which it doesn't.

Also be aware that by overlaying both files only in the release repo it will be impossible to build a development version of your packages from source anymore.

If you are aware of all these drawbacks / limitations but want to proceed with this PR as it we can merge it.

k-okada:
The reason I did not add CMakeLIsts.txt in source repo is just I'm afraid that requires a lot of effort, specially I wan to not just add CMakeLists.txt but to change whole build system from Make to CMake, I really wan to do that, but this remains my todo list for years. so  I'll choose removing package.xml for now. Hope you merge this PR

k-okada:
BTW, do I also have to change doc entry that point to source repo, without both package.xml and CmakeLists.txt ?
```
  euslisp:
    doc:
      type: git
      url: https://github.com/euslisp/EusLisp.git
      version: master
```

dirk-thomas:
Yes, and the corresponding doc job has been failing every three days since it was registered for exactly that reason: http://jenkins.ros.org/view/Idoc/job/doc-indigo-euslisp/

PR #7217, opened 2015-02-12
fix doc location for rtm packages

k-okada:
sorry to bother you, but I fixed doc entity for other packages as well

### DIFF
--- a/hydro/distribution.yaml
+++ b/hydro/distribution.yaml
@@ -4977,33 +4977,33 @@ repositories:
     status: maintained
   openrtm_aist:
     doc:
-      type: svn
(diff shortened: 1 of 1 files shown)


hydro/distribution.yaml
```diff
@@ -7375,7 +7375,7 @@ repositories:
     source:
       type: git
       url: https://github.com/gbiggs/rtctree.git
-      version: master
+      version: release/hydro/rtctree
```
tfoote: This branch does not exist

k-okada: fixed

--
◉ Kei Okada

PR #7273, opened 2015-02-17
python-colorama is only available on trusty

k-okada:
see http://packages.ubuntu.com/search?lang=en&suite=trusty&searchon=names&keywords=python-colorama

or did you intened to install via pip?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -215,7 +215,16 @@ python-clearsilver:
     trusty: [python-clearsilver]
 python-colorama:
   fedora: [python-colorama]
-  ubuntu: [python-colorama]
+  ubuntu:
+    lucid: []
+    maverick: []
+    natty: []
+    oneiric: []
+    precise: []
+    quantal: []
+    raring: []
+    saucy: []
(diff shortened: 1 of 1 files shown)


tfoote:
These should be undefined, not zero length lists. Zero length lists implies that the item is expected to be already available on the system.

k-okada:
If we have package that depends on python-colorama, and try to build deb on
your buildfirm with indigo, it will fail on utopic  and vivid, except
trusty. Is it right? Of course we can patch for each ubuntu release to
avoid this problem, but may be this is confusing.Should we use pip for this
package?

--
◉ Kei Okada

dirk-thomas:
No, PIP is not an option since released packages are depending on it (I just assume this until proven otherwise). The package is available in Trusty, Utopic as well as Vivid. So the rosdep key should be registered for these.

Then you can use it for Jade releases (using T, U, V) and partly Indigo (T will work, S will fail).

k-okada:
ok, I have created another PR that remove null list -> https://github.com/ros/rosdistro/pull/7289,

Issue #7346, opened 2015-02-24
devel-indigo-openrtm_aist devel job continuously triggering


tfoote:
Has been triggering continuously http://jenkins.ros.org/job/devel-indigo-openrtm_aist/

It looks to correctly be a branch. Thus not a reoccurance of #3989  Not sure why it's continuously triggering. 

I've disabled the job for now.

tfoote:
Also for hydro: http://jenkins.ros.org/job/devel-hydro-openrtm_aist

tfoote:
cc/ @k-okada @130s 
It looks like this is an issue upstream: https://issues.jenkins-ci.org/browse/JENKINS-27115 until that's resolved this won't be resolvable. To that end since this is causing several hundred builds per day on the farm I'm going to recommend that we remove the devel entries for these packages. 

This will result in them not being available from "source" via rosinstall, but since it's in the release repository anyway, it's not actually the source anyway. And this entry will only update when bloomed so it's not subject to the standard development cycles.

k-okada:
Ok,removed these entries https://github.com/ros/rosdistro/pull/7354

Please let me know if there are something that we should fix in our side

tfoote:
Thanks @k-okada I'm working on finding a work around. It should not require anything from your side. 

When/if we find the work around we should be able to reenable the jobs.

PR #7354, opened 2015-02-25
remove pckages which causes #7346

k-okada:


tfoote:
Thanks for putting this in @k-okada. I've filed it upstream but don't expect a quick response from them.

Issue #1389, opened 2013-07-15
Inconsistent rosdep for key fltk


meyerj:
rosdep key `fltk` resolves to `[fluid, libfltk1.1-dev]` for Ubuntu since 21f94b840117fc48f0dd74727b97095c1a0ede93. Package [fluid](http://packages.ubuntu.com/precise/fluid) recommends [libfltk1.3-dev](http://packages.ubuntu.com/precise/libfltk1.3-dev) in Ubuntu precise, which is incompatible with [libfltk1.1-dev](http://packages.ubuntu.com/precise/libfltk1.1-dev) required by rosdep.

Although this is not really a conflict as libfltk1.3-dev is only a recommendation, it obviously causes errors on Hudson, e.g.
- http://build.willowgarage.com/job/devel_groovy_hector_quadrotor_precise_amd64/15/ or
- http://build.willowgarage.com/view/devel_groovy/job/devel_groovy_hector_gazebo_precise_amd64/21/.

I cannot reproduce this result locally, but I assume that apt is somehow configured to always install recommended packages and even to allow removal of conflicting packages on the build servers.

See also ros-infrastructure/jenkins_scripts#46.

tfoote:
Reviewing this it seems that 1.3 is now the recommended one. We should update the rule to point to 1.3 for newer versions. I don't want to break anything already running so my suggestion would be to make the change for vivid and forward. Leaving the older systems as they are.

dirk-thomas:
Why not also for utopic?

tfoote:
Good point I was thinking about utopic already being out. but we haven't released ROS on top yet so that should be fine.

deffi:
Now calling rosdep install (as described on [indigo/Installation/Source](http://wiki.ros.org/indigo/Installation/Source)) alternatingly installs `libfltk1.3-dev` and `libfltk1.1-dev`, and uninstalls the other.

Adding the following line to a local rosdep YAML file resolves the issue, only installing 1.3:
`fltk: { ubuntu: { utopic: ["fluid", "libfltk1.3-dev"] } }`

In fact, looking at [rosdistro/rosdep/base.yaml](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml) reveals that the fix (9fd2c55c0f6ef5957d18160f43e7af25b45a7015) sets the correct version for `libfltk-dev`, but `fltk` still refers to `libfltk1.1-dev` on all Ubuntu versions.

PR #7377, opened 2015-02-26
homogenize fltk and libfltk-dev packages, re #1389

tfoote:
Fix for comment: https://github.com/ros/rosdistro/issues/1389#issuecomment-76171667

@deffi if you can track down who's using the old key please suggest that they update to use libfltk-dev and fluid as independant rosdep rules and drop the deprecated fltk rule.


tfoote:
@wjwwood can you check this logic?

deffi:
@tfoote it's stage. I sent a [pull request](https://github.com/ros-gbp/stage-release/pull/3) to [ros-gbp/stage-release](https://github.com/ros-gbp/stage-release), I hope that's the right place.

wjwwood:
@tfoote I guess that looks fine to me.

@deffi I'll have to merge your pull request and then do a (re)release of stage, but only after this has been merged.

PR #7388, opened 2015-02-27
Adding tf_laser_transform to documentation index for distro

gus484:
I'd like tf_laser_transform to be indexed and documented on ros.org.


vrabaud:
Please respect the alphabetical order. I am not sure the naming is good though: it is not as generic as the name suggests.

tfoote:
I agree with @vrabaud  I think that tinkerforge_laser_transform would be a better name. It doesn't actually use tf, except for a few datatypes. 

Also there's a unrelated change in base.yaml which should be submitted separately.

PR #7426, opened 2015-03-02
Adding tinkerforge_laser_transform to documentation index for indigo

gus484:
I'd like tinkerforge_laser_transform to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8048,6 +8048,16 @@ repositories:
       url: https://github.com/wcaarls/threemxl.git
       version: master
     status: maintained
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -8048,6 +8048,16 @@ repositories:
       url: https://github.com/wcaarls/threemxl.git
       version: master
     status: maintained
+  tinkerforge_laser_transform:
+    doc:
+      type: git
+      url: https://github.com/gus484/ros.git
+      version: indigo-devel
```
dirk-thomas: The repository does not contain an *indigo-devel* branch.

PR #7506, opened 2015-03-14
Adding cob_people_perception to documentation index for groovy, hydro, indigo

ipa-rmb:


vrabaud:
The Travis test says you cannot add documentation to Groovy as it's EOLed.

PR #7572, opened 2015-03-19
Update libpcl-all and libpcl-all dev keys

paulbovbel:


rosdep/base.yaml
```diff
@@ -1660,7 +1667,14 @@ libpcl-all-dev:
...
   opensuse: [pcl]
-  ubuntu: [libpcl-1.7-all-dev]
+  ubuntu:
+    precise: [libpcl-1.7-all-dev]
+    quantal: [libpcl-1.7-all-dev]
+    raring: [libpcl-1.7-all-dev]
+    saucy: [libpcl-1.7-all-dev]
+    trusty: [libpcl-1.7-all-dev]
+    utopic: [libpcl-1.7-all-dev]
+    vivid: [libpcl-dev]
```
wjwwood: It's strange that this is unversioned where as the non`-dev` is versioned. Is this correct?

paulbovbel: Definitely strange, going based on http://packages.ubuntu.com/search?keywords=libpcl&searchon=names

wjwwood: @jspricke Can you comment on why the `-dev` package in Vivid is not versioned but the non`-dev` is?

tfoote: Utopic should match vivid with libpcl-dev. That release is a straight backport of upstream. I believe the `-dev` package is unversioned since it is not side by side compatible.

tfoote:
imported missing utopic libpcl-dev here: https://github.com/ros-infrastructure/ros_bootstrap_dependencies/pull/2

rosdep/base.yaml
jspricke: PCL got accepted into Debian, so it will be in Vivid (it's not in Utopic, as you can see here: http://packages.ubuntu.com/search?suite=all&arch=any&searchon=names&keywords=libpcl-dev). Also, Debian policy is to version the lib package according to the SONAME and have the development package unversioned, as you can only install one at the time.

jspricke: As a second note, I've stopped pushing stuff into the PCL PPA, as it's not needed anymore starting with Vivid.

jspricke: Ah.. and for Utopic we where already working on the PCL package, so I put the new version into the PPA back then. (Sorry for all the noise).

wjwwood: It's pretty confusing that the portable coroutine library is `libpcl1-dev` whereas PCL is `libpcl-dev`, surely that isn't intentional...

jspricke: Yeah, that's unfortunate, but I don't see a good way to change it.

rosdep/base.yaml
```diff
@@ -1660,7 +1667,14 @@ libpcl-all-dev:
...
   macports: [libpcl]
   opensuse: [pcl]
-  ubuntu: [libpcl-1.7-all-dev]
+  ubuntu:
+    precise: [libpcl-1.7-all-dev]
+    quantal: [libpcl-1.7-all-dev]
+    raring: [libpcl-1.7-all-dev]
+    saucy: [libpcl-1.7-all-dev]
+    trusty: [libpcl-1.7-all-dev]
+    utopic: [libpcl-1.7-all-dev]
```
wjwwood: This needs to change to `libpcl-dev`.

rosdep/base.yaml
```diff
@@ -1651,7 +1651,14 @@ libpcl-all:
...
   macports: [libpcl]
   opensuse: [pcl]
-  ubuntu: [libpcl-1.7-all]
+  ubuntu:
+    precise: [libpcl-1.7-all]
+    quantal: [libpcl-1.7-all]
+    raring: [libpcl-1.7-all]
+    saucy: [libpcl-1.7-all]
+    trusty: [libpcl-1.7-all]
+    utopic: [libpcl-1.7-all]
```
wjwwood: This should also match vivid, `libpcl1.7`.

tfoote: This should be `libpcl1.7` for utopic too.

paulbovbel:
Done, thank you

PR #7687, opened 2015-03-30
remove OpenCV3 on Hydro

vrabaud:
It does not compile on Precise because of a gcc bug and I don't
want to patch it.


vrabaud:
for reference of the compilation problem: http://jenkins.ros.org/job/ros-hydro-opencv3_binarydeb_precise_amd64/12/consoleFull

dirk-thomas:
The "reference" will be removed next night on the farm (with the removal of the entry). If you want the information to persist just copy'n'[aste the important part of the output to this ticket.

PR #7688, opened 2015-03-30
add uos_rotunit, switch branch for uos_tools

mintar:


hydro/distribution.yaml
```diff
@@ -8862,15 +8862,24 @@ repositories:
       type: git
       url: https://github.com/uos/uos_slam.git
       version: hydro
+  uos_rotunit:
```
dirk-thomas: Wrong alphabetical order.

mintar: Oops, sorry. Fixed.

PR #7695, opened 2015-03-31
Adding grid_map to documentation index for distro.

pfankhauser:
I'd like grid_map to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2064,6 +2064,16 @@ repositories:
       url: https://github.com/mikeferguson/grasping_msgs-gbp.git
       version: 0.3.1-0
     status: developed
+  grid_map:
+    doc:
+      type: git
+      url: https://github.com/ethz-asl/grid_map.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/ethz-asl/grid_map.git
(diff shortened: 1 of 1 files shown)


pfankhauser:
@vrabaud "hey, that looks cool btw, how does that compare to costmap_2d ?"

Thanks. Grid map has similar features to costmap_2d, but was developed as backbone for the [elevation mapping](https://github.com/ethz-asl/elevation_mapping) package. It has following features:
* Universal application to 2.5-dimensional mapping, any number of layers supported (elevation, variance, color, surface normal, occupancy, traversability, foothold quality etc.).
* Data storage is implemented as two-dimensional circular buffer. This allows re-positioning of the map (to follow the robot) without copying data.
* Data is stored as [Eigen](http://eigen.tuxfamily.org/) data types, allowing to use the available Eigen algorithms directly to the map data.
* Compatibility with ROS message types such as PointCloud2, OccupancyGrid, and GridCells.

vrabaud:
Thx for the info ! Did you advertise it on the navigation SIG http://wiki.ros.org/sig/Navigation ?

pfankhauser:
Not yet, I will do so, thanks!

PR #7777, opened 2015-04-04
mavlink: 2014.4.4-0 in 'jade/distribution.yaml' [bloom]

vooon:
Increasing version of package(s) in repository `mavlink` to `2014.4.4-0`:
- bloom version: `0.5.19`
- previous version for package: `2015.3.11-0`

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -578,7 +578,7 @@ repositories:
       tags:
         release: release/jade/{package}/{version}
       url: https://github.com/mavlink/mavlink-gbp-release.git
-      version: 2015.3.11-0
+      version: 2014.4.4-0
     status: maintained
   mavros:
     doc:


tfoote:
@vooon 

It has been flagged as a regression by our debbuild status page: http://54.183.65.232/status_page/ros_jade_default.html?q=REGRESSION

This is actually a downgrade, from what was synced to public. 
If you want people to get the downgrade you will need to rerelease with a higher version, usually referencing the downgraded version in the build number part of the version string.

vooon:
Ohhhh! It should be 2015! Quick question there is a method to specify ros disrto once in bloom command?

Now i'm do that:
```
bloom-release --rosdistro jade --track jade mavros
```

vooon:
Rereleased #7863.

Also question: how fast mavros 0.11.1 will be in repos (0.11.0 has one little bug fixed in 0.11.1)?

tfoote:
There's not a way to set the track and rosdistro (they are usually used together, but are technically different)  BTW you can use the short version for the --rosdistro, -r and --track -t which will save you several characters

Syncs usually go about every 2 weeks. If you'd like to request an acceleration of that please email ros-release@lists.ros.org to suggest it. That way we can keep all the maintainers in the loop.

dirk-thomas:
I am using the following bash aliases to type less:

    alias bloom-release-indigo="bloom-release -r indigo -t indigo"

vooon:
Thanks, before i used `^R` (history search), so usually long params ok, just annoying (and sometimes i messed up with distro!=track or year :) ).
Now i updated my zsh aliases.

PR #7878, opened 2015-04-08
added release for maggie_*_msgs repos

raulperula:


hydro/distribution.yaml
```diff
@@ -3883,6 +3883,17 @@ repositories:
       type: git
       url: https://github.com/UC3MSocialRobots/maggie_devices_msgs.git
       version: hydro-devel
+    release:
+      packages:
+      - eyelids_msgs
+      - ir_controller_msgs
+      - maggie_devices_msgs
+      - motor_controller_msgs
+      - rfid_msgs
```
dirk-thomas: These package names sound very generic. I am not sure if it makes sense to introduce them as-is. May be it would be better to prefix them with `maggie_` in order to not indicate that they are supposed to be generic messages?

raulperula: That is the idea. What we pretend is to do generic messages for using in several different robots. Now is only maggie robot, but we are developing two different robots and trying to unify the messages for similar devices (hardware). If you still think that is better to include the prefix "maggie_", tell me and I modify the names. Thanks for your comment.

tfoote: Since all packages are in a flat namespace we try to be careful about generic names. rfid_msgs seems like something that once reviewed by the community might be promoted to be inside common_msgs. However if you use the `rfid_msgs` name in your release either it would have to be your implementation  or we would have to pick a different name. There are already other rfid messages defined such as: http://wiki.ros.org/hrl_rfid 

Also when searching for other existing rfid messages I note you've picked the name [rfid](http://wiki.ros.org/rfid) for another package. 

This will be packaged into ros-indigo-rfid so someone coming along may try to install that to add rfid functionality to their robot. And they'll be surprised to find a driver package for ["the Social Robot Maggie"](https://github.com/UC3MSocialRobots/maggie_devices/tree/hydro-devel/rfid)

In general please review the [Package Naming Conventions](http://wiki.ros.org/ROS/Patterns/Conventions) Anything which is robot or family of robots specific should use a prefix to indicate as such.

raulperula: Now I think we are following the Package Naming Conventions for our packages. Please tell me if everything is ok or not. Thanks.

dirk-thomas:
The PR does currently not pass Travis validation since the release repo seem to not contain the claimed tags.

raulperula:
Sorry, there was a bug in the release commit. Now it should be fine.

tfoote:
Your package list is out of alphabetical order. 

```
--- file:///home/travis/build/ros/rosdistro/hydro/distribution.yaml
+++ loaded-and-saved
@@ -3888,5 +3888,5 @@
       packages:
+      - maggie_devices_msgs
       - maggie_eyelids_msgs
       - maggie_ir_controller_msgs
-      - maggie_devices_msgs
       - maggie_motor_controller_msgs
```

Note we usually recommend letting bloom submit the pull-requests for releases. And only manually creating the source and doc entries if necessary. Bloom has an option to only generate the pull-request if you've already run the release, `-p`.

raulperula:
Ok, sorry about that. I have corrected the order. Next releases I will try to use the bloom pull request submit as you recommend.

PR #7954, opened 2015-04-15
Fixes to compile ros indigo on Gentoo.

bluecube:
Requires overlays ros and Gentoo-ROS-overlay (https://github.com/steup/Gentoo-ROS-Overlay)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -84,7 +84,7 @@ assimp:
   fedora: [assimp]
   gentoo:
     portage:
-      packages: [media-libs/assimp]
+      packages: [media-gfx/assimp]
   macports: [assimp]
   opensuse: [libassimp3]
   rhel: [assimp-devel]
@@ -131,7 +131,7 @@ assimp-dev:
   fedora: [assimp-devel]
   gentoo:
     portage:
-      packages: [media-libs/assimp]
+      packages: [media-gfx/assimp]
   opensuse: [libassimp3]
   rhel: [assimp-devel]
   ubuntu:
@@ -430,7 +430,7 @@ collada-dom:
   fedora: [collada-dom-devel]
   gentoo:
     portage:
-      packages: [media-libs/collada-dom]
+      packages: [dev-libs/collada-dom]
   macports: [collada-dom]
   ubuntu: [collada-dom-dev]
 comedi:
@@ -692,6 +692,7 @@ gazebo:
     raring: [gazebo]
     saucy: [gazebo2]
     trusty: [gazebo2]
+  gentoo: [sci-electronics/gazebo]
 gcc-avr:
   arch: [avr-gcc]
   debian: [gcc-avr]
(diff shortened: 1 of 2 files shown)


rosdep/base.yaml
```diff
@@ -692,6 +692,7 @@ gazebo:
     raring: [gazebo]
     saucy: [gazebo2]
     trusty: [gazebo2]
+  gentoo: [sci-electronics/gazebo]
```
wjwwood: Any idea what version of `gazebo` this is?

rosdep/base.yaml
```diff
@@ -2172,6 +2177,7 @@ liburdfdom-headers-dev:
     trusty: [liburdfdom-headers-dev]
     utopic: [liburdfdom-headers-dev]
     vivid: [liburdfdom-headers-dev]
+  gentoo: [media-libs/urdfdom_headers]
```
wjwwood: Where these (`console_bridge`, `urdfdom`, and `urdfdom_headers`) contributed by you or someone else into portage? Any idea what versions?

wjwwood:
Our CI enforces alphabetical ordering: https://travis-ci.org/ros/rosdistro/builds/58676112#L192

rosdep/base.yaml
bluecube: These are from the Gentoo-ROS-Overlay I linked in the description. Versions 0.2.7, 0.3.0 and 0.3.0.

rosdep/base.yaml
bluecube: Gazebo comes from the main gentoo repository and is version 5.1.0. Is this ok?

bluecube:
I was a little too fast with making this pull request. Other than the ordering there is a one more problem with dependencies between urdfdom and urdfdom_headers... but I'll be back :-)

Btw I have a few questions:
1) What should I do with this pull request?
  - close it and open a new fixed one?
  - add new commits with the fixes?
  - rebase content of the first commit?
2) Can I run the checks that travis does locally?
3) At three places (libvtk-java, python-cairo and python-vtk) I'm using `use:` as in ros-infrastructure/rosdep#176, although It probably doesn't do anything. The problem is that there are no separate packages for language support of libraries and AFAIK the only way to get this is with changed use flags. Is it ok to keep them in the yaml files?

wjwwood:
> 1) What should I do with this pull request?

I would close this, then when you're ready you can reopen it and update the branch it points to with the correct content.

> 2) Can I run the checks that travis does locally?

Yes, just do `nosetest -s` in the root of the repository.

> 3) At three places (libvtk-java, python-cairo and python-vtk) I'm using use: as in ros-infrastructure/rosdep#176, although It probably doesn't do anything. The problem is that there are no separate packages for language support of libraries and AFAIK the only way to get this is with changed use flags. Is it ok to keep them in the yaml files?

Yes, if you plan to eventually update `rosdep` to support them.

PR #8011, opened 2015-04-21
add python-requests-pip to rosdep/python.yaml

komeisugiura:


dirk-thomas:
Why do you need this entry beside the already existing `python-requests` one?

komeisugiura:
We would like to use Requests version 2.0, and this can be installed as python-requests-pip. By python-requests, Requests version 1.0 will be installed instead.

dirk-thomas:
Ubuntu Trusty seems to already contain version 2.2 (http://packages.ubuntu.com/trusty/python-requests). On which platform are you trying to use it?

tfoote:
We strongly recommend finding a way to work with the upstream released package versions. Although it's usually a little behind it's more stable and is compatible with other packages deployed on the system. (For example if you install a new version from pip it may break other packages maintained and released against the debian version. 

Also this will only work for builds from source, and debian packages cannot be built using the proposed rule.

komeisugiura:
Thanks for your comments. I'm going to find another way with the upstream released package versions, and I'd like to close this issue.

PR #8093, opened 2015-04-23
urg_node: 0.1.9-0 in 'jade/distribution.yaml' [bloom]

chadrockey:
Increasing version of package(s) in repository `urg_node` to `0.1.9-0`:
- bloom version: `0.5.19`
## urg_node

```
* Merge pull request #7 <https://github.com/ros-drivers/urg_node/issues/7> from mikeferguson/indigo-devel
  add a script to set the IP address of an URG laser
* Updated diagnostics to support configurable parameters.
* add a script to set the IP address of an URG laser
* Contributors: Chad Rockey, Michael Ferguson
```

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -749,13 +749,6 @@ repositories:
       url: https://github.com/ros-perception/laser_geometry.git
       version: indigo-devel
     status: maintained
-  laser_proc:
-    release:
-      tags:
-        release: release/jade/{package}/{version}
-      url: https://github.com/ros-gbp/laser_proc-release.git
-      version: 0.1.4-0
-    status: maintained
   librms:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


chadrockey:
no idea how this collided with its own dependency lol

wjwwood:
So, the problem started at some point when I had not changed bloom. My leading theory is that GitHub is taking longer than they did at one time to update their cached files under `raw.github.com` (or `raw.githubusercontent.com`), which means there is a race condition between what you make a diff against (based on what's in the GitHub cache) and what is in the actual repository. That's why you don't see it on the terminal before submitting.

chadrockey:
Yeah, I think that's right.  It also looks like they may use multiple caches.  I had to rosdep update a few times to get it to pick up laser_proc in the rosdep step so it's very surprising that the diff removes laser_proc.

dirk-thomas:
GitHub caches files for up to 5 minutes through their CDN (when using a URL like the ones above).

PR #8128, opened 2015-04-24
add libav(fedora) to rosdep/base.yaml

komeisugiura:


dirk-thomas:
Why is the list empty?

komeisugiura:
I'd like to use an audio conversion function, and I've been using ffmpeg until Ubuntu 13.10. From Ubuntu 14.04, ffmpeg is not supported any more, and libav(avconv) is supported instead.

The current situation is as follows:
- Fedora supports ffmpeg but does not support liav(avconv).
- Ubuntu 14.04 supports libav but does not support ffmpeg.
- Ubuntu 12.04-13.10 support both ffmpeg and libav.

Considering the above situation, I'd like to use libav for rospeex, and I think I have to put an empty list for Fedora when making an package using bloom. What do you think?

(An empty list itself has been approved for some packages such as xulrunner-1.9.2 and openmpi)

PR #8214, opened 2015-04-28
added python-avahi and python-docutils to ubuntu utopic, vivid

atiderko:


rosdep/python.yaml
```diff
@@ -336,6 +338,8 @@ python-docutils:
     saucy: [python-docutils]
     trusty: [python-docutils]
     trusty_python3: [python3-docutils]
+    utopic: [python-docutils]
```
vrabaud: please add the Python 3 versions too

PR #8227, opened 2015-04-28
adding flask-restful as pip for ubuntu

asmodehn:


wjwwood:
There doesn't seem to be a `.deb` for this in Ubuntu, so this is correct. However, since this key doesn't resolve to an `apt-get` rule, any packages which depend on it cannot be released as binary packages through our farm. This is because we cannot build a `.deb` which depends on a `pip` installed Python package.

PR #8318, opened 2015-05-04
removing OpenCV3 from Indigo until build is fixed on old libavcodec

vrabaud:


vrabaud:
fyi, http://code.opencv.org/issues/3986

PR #8369, opened 2015-05-07
add pygraphviz config for osx

trthanhquang:
bug when install ros indigo desktop version on OS X 10.10.3
`rosdep install --from-paths src --ignore-src --rosdistro indigo -y` 
keep reporting error:

```
to system dependencies:
qt_dotgraph: No definition of [python-pygraphviz] for OS [osx]
```


vrabaud:
out of alphabetical order: osx needs to be before ubuntu

trthanhquang:
haha did not know that it need alphabetical order

vrabaud:
let's keep it clean :) thx

PR #8394, opened 2015-05-08
tornado-couchdb, validictory, jsonref & ws4py

ablasdel:


rosdep/python.yaml
```diff
@@ -1638,6 +1642,10 @@ python-tornado-pip:
   ubuntu:
     pip:
       packages: [tornado]
+python-tornado-couchdb-pip:
```
tfoote: This should be before  python-tornado-pip:

ablasdel:
@tfoote revised

PR #8543, opened 2015-05-24
liburdfdom-tools and sdl-mixer for rosdep

NLZ:
I couldnt find other packages for liburdfdom-tools, hope its okay.


vrabaud:
fine for now: people will add when needed. Thx.

PR #8580, opened 2015-05-31
ubuntu utopic removes mongodb-dev

mikeferguson:


mikeferguson:
best I can tell, there is no new package that is needed. We'll know for sure once this gets merged and we see if moveit_ros_planning_interface actually builds on utopic/vivid

vrabaud:
it will fail, sorry, cf https://github.com/ros-planning/warehouse_ros/issues/20

mikeferguson:
I suppose I should undo this -- so that it continues to work on at least Trusty until we get this fixed.

PR #8592, opened 2015-06-01
use homebrew/science/pcl instead of just pcl to resolve ambiguity

jmtatsch:
otherwise this issue arises https://github.com/ros-infrastructure/rosdep/issues/339


wjwwood:
Thanks!

PR #8613, opened 2015-06-05
Adding alembic

asmodehn:
It s python code, so I assume this is the right place ?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1,5 +1,3 @@
-alembic:
-  ubuntu: [alembic]
 ipython:
   debian: [ipython]
   fedora: [python-ipython]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
It might be better to name the package `python-alembic` and also use that Ubuntu package as of `utopic`. Only on `trusty` it could fall back to the name without the Python prefix.

tfoote:
What is the difference between these: http://packages.ubuntu.com/trusty/alembic and 
http://packages.ubuntu.com/search?keywords=python-alembic

There's also python3 version which a rule might be added for: http://packages.ubuntu.com/utopic/python3-alembic

asmodehn:
I m using python2 on trusty and I'm not sure what ros policy is regarding versions of python...
So I just setup python-alembic for > trusty and alembic for trusty.

dirk-thomas:
Thanks for updating. I think this is the first package for `wily`...

PR #8640, opened 2015-06-09
Break pointcloud_to_laserscan out of perception_pcl (jade)

paulbovbel:


wjwwood:
@dirk-thomas do you see anything wrong with this approach? Seems like it should work to me.

dirk-thomas:
New version is higher - lgtm.

PR #8677, opened 2015-06-16
Adding reinforcment_learning to documentation index for groovy

toddhester:
I'd like my repo reinforcement_learning to be indexed and documented on ros.org. It was previously a long time ago, but I never updated to the new doc process.


dirk-thomas:
`groovy` is EOL for quite a while now. We do not generate documentation for that distribution anymore. Please consider to update it to work with current distributions and add it there.

PR #8680, opened 2015-06-17
rename fetch_driver_msgs to fetch_msgs

mikeferguson:
I'm about to add additional packages to this repo

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1878,21 +1878,6 @@ repositories:
       url: https://github.com/ros-gbp/fcl-release.git
       version: 0.3.2-0
     status: maintained
-  fetch_driver_msgs:
-    doc:
-      type: git
-      url: https://github.com/fetchrobotics/fetch_driver_msgs.git
-      version: master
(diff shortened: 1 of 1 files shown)


mikeferguson:
I think this is actually correct -- but the test complains because renaming the repo also requires it be moved, and the diff can't handle that.

indigo/distribution.yaml
```diff
@@ -1878,37 +1878,39 @@ repositories:
...
     doc:
       type: git
-      url: https://github.com/fetchrobotics/fetch_gazebo.git
+      url: https://github.com/fetchrobotics/fetch_msgs.git
       version: master
     release:
       packages:
-      - fetch_gazebo
-      - fetch_gazebo_demo
+       - fetch_driver_msgs
```
dirk-thomas: As Travis reports the indentation of this line is wrong.

mikeferguson:
@dirk-thomas thanks for catching that, I had looked at the travis log but missed that, just rebased this into a single commit, should make it through testing momentarily

PR #8682, opened 2015-06-17
Adding dlut_laser and dlut_vision to documentation index for distro

ZhuangYanDLUT:
Adding dlut_laser and dlut_vision to documentation index for distro.
dlut_laser contains some laser tools,  such as dlut_hash_icp.
dlut_vision contains dlut_viso2 which was developed from viso2_ros and libviso2, and other packages.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1479,6 +1479,15 @@ repositories:
       type: git
       url: https://github.com/uos/diffdrive_gazebo_plugin.git
       version: indigo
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Some of the packages in the repo are based on rosbuild and won't be indexed. Are you sure that you want to go ahead to document at least the catkin packages?

ZhuangYanDLUT:
Yes, I will catkinlize the package later.
Thanks.

PR #8684, opened 2015-06-17
added uc3m maps for hydro. added maggie robot msgs in jade for documenting wiki

raulperula:


dirk-thomas:
Please be aware that Hydro is EOL by now. We are still building packages and generating documentation but that might end in the near future.

raulperula:
Thanks @dirk-thomas for the advice. We know that, we have in our plans to go to Indigo (LTS).

PR #8685, opened 2015-06-17
Adding bta_ros to documentation index for jade

angelmerino:
I would like to index and add documentation for package bta_ros in jade.

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -170,6 +170,11 @@ repositories:
       url: https://github.com/ros/bond_core.git
       version: master
     status: maintained
+  bta_ros:
+    doc:
+      type: git
+      url: https://github.com/voxel-dot-at/bta_ros.git
+      version: master
   calibration:
     doc:
       type: git


vrabaud:
hi, I am not crazy about having ROS in the name (it's kind of obvious), maybe bta_drivers or bta_bridge ? @tfoote, any advice ?
That reminds me: I need to finalize my REP ....

vrabaud:
oh wait, that's been around for a while as it's in indigo.

angelmerino:
Hi, actually it is a driver for Bluetechnix "Time of Flight" cameras. It is indexed for hydro and indigo. Just wanted to add jade as it works the same after testing.
Do you suggest a name change?

tfoote:
It would be better to make the name more specific and not have ros in the name. It's redundant to state ros in the name of a ROS package. 

For reference the naming convention is here: http://wiki.ros.org/ROS/Patterns/Conventions

i'll merge this since it's already available under that name but it would be recommended to change the name to something more like bluetechnix_camera for clarity and introspection. From just 'bta_ros` I have no idea what the package is or does.

PR #8788, opened 2015-06-24
Added pip install of python websocket for older ubuntu distros and support for python3

mitchellwills:
Does what I did for trusty_python3 and utopic_python3 correctly use python3 pip or do I need to do something special? The python3-websocket ubuntu package is only released in vivid and wily.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1815,7 +1815,31 @@ python-vtk:
     utopic: [python-vtk]
     vivid: [python-vtk]
 python-websocket:
-  ubuntu: [python-websocket]
+  ubuntu:
+    precise:
+      pip:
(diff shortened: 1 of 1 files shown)


wjwwood:
lgtm, just a heads up, if you release a package which depends on this, then it cannot be built on platforms where pip is used instead of apt. So for instance it would work to release against this for Jade, but not Indigo, since Indigo also needs an apt rule for Saucy.

mitchellwills:
Its just being used as a test dependency, not build or run so that should be OK correct?

wjwwood:
It will be ok for a release, and theoretically it will work with a devel or prerelease since rosdep should install it, even on Saucy. However, I'm not sure if we've ever tested that.

PR #8804, opened 2015-06-25
adding iot_bridge

corb555:


tfoote:
Looking briefly at this project it's definitely interesting, but I would suggest that your previous name openhab_bridge was a much better name. It's much more descriptive of what a user can find in the package. Our naming conventions are here: http://wiki.ros.org/ROS/Patterns/Conventions

corb555:
The reason I renamed it is that there is now a second project, "Eclipse SmartHome" in development that uses parts of OpenHAB.   The iot_bridge should be able to work with both Eclipse SmartHome and OpenHAB.  Let me know if that makes sense or you think I should revert to OpenHAB_bridge.

PR #8836, opened 2015-07-02
Adding followme to documentation index for distro distribution.yaml

kendemu:
I'd like followme to be index and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2005,6 +2005,20 @@ repositories:
       url: https://github.com/ros-drivers/flir_ptu.git
       version: master
     status: developed
+  followme:
+    doc:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The validation of the PR failed because the version `indigo` is not a valid branch in the repository.

But beside that you should not use the same repository for the upstream sources and the release. Please follow the bloom tutorial to setup a separate release repository and after that clean your upstream repository.

tfoote:
The first time release tutorial can be found here: http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

tfoote:
Closing. Please follow the above tutorial to open a new PR using bloom.

PR #8838, opened 2015-07-02
Updated two dependencies for gentoo.

allenh1:
There were two small problems with the package selection. This should satisfy everything.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2587,6 +2587,7 @@ nvidia-cg:
   ubuntu: [nvidia-cg-toolkit]
 nvidia-cuda:
   debian: [nvidia-cuda-toolkit]
+  gentoo: [nvidia-cuda-toolkit]
   ubuntu: [nvidia-cuda-toolkit]
 omniidl:
   debian:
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -2708,7 +2708,7 @@ pkg-config:
   debian: [pkg-config]
   fedora: [pkgconfig]
   freebsd: [pkg-config]
-  gentoo: [dev-util/pkgconfig]
+  gentoo: [dev-util/pkgconf]
```
dirk-thomas: Both seems to exist:
* https://packages.gentoo.org/package/dev-util/pkgconfig
* https://packages.gentoo.org/package/dev-util/pkgconf

Why is the new one favored over the previous one?

allenh1:
Most gentoo applications use dev-util/pkgconf, and both cannot be on the system at the same time.

dirk-thomas:
It sounds like a big change for every current user to switch?

I will keep this for @tfoote to merge.

tfoote:
What are the implications for someone who's already using pkgconfig?

allenh1:
pkgconf is an identical system to pkgconfig. The change should not affect anyone. I should also point out that my system won't even let me install pkgconfig. I did a clean install to make sure this wasn't anything I had done abnormally and pkgconf was pulled in with the install of Gnome. I suppose the safest thing would be to allow either, but I don't know how to do that or if it's even a possibility.

tfoote:
If it's incompatible as it stands with a fresh install with gnome that seems reasonable to switch.

PR #8841, opened 2015-07-03
adding python-concurrent.futures

asmodehn:


vrabaud:
that is the name of the Ubuntu package: http://packages.ubuntu.com/precise-backports/python/python-concurrent.futures. The dot bothers me so I'll let somebody else merge it :)

dirk-thomas:
The entry `python-pyside.qtuitools` also has a dot and that works without a problem.

PR #8844, opened 2015-07-03
Update python.yaml: pynmea2

gajan:


vrabaud:
indeed, does not seem to be in Ubuntu.

PR #8864, opened 2015-07-07
add openssl and ca-certificates

k-okada:


vrabaud:
ca-certificates should be before catkin_pkg

k-okada:
thanks, fixed

PR #8873, opened 2015-07-08
Add BSON dependency

DLu:


DLu:
Thanks WW!

wjwwood:
np, nice branch name btw.

PR #8915, opened 2015-07-14
adding ros_rtt to documentation index for distro

sukha-cn:
I did some work about real-time communication in ros. The package supports real-time communication within node at the moment. And it will support communication between nodes in the future. I'd like this repository to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -7516,6 +7516,21 @@ repositories:
       url: https://github.com/ros/ros_realtime.git
       version: hydro-devel
     status: maintained
+  ros_rtt:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


wjwwood:
What's the difference in your work and this?

http://wiki.ros.org/rtt

Also, out naming guidelines suggest not putting `ros_` as a prefix in a name since it is redundant, and instead you should put a more meaningful name that's tied to you or your organization until it becomes widely used.

Also the CI is failing because your repository doesn't seem to have a `package.xml`:

https://travis-ci.org/ros/rosdistro/builds/70866291#L3980

If you just want to document, then you don't need the source or release entries. If you want to release it, you should follow the bloom tutorials: http://wiki.ros.org/bloom

PR #8916, opened 2015-07-14
Add PhantomJS, Selenium, and their various dependencies

mehwang:


rosdep/base.yaml
```diff
@@ -342,6 +342,8 @@ box2d-dev:
   debian: [libbox2d-dev]
   fedora: [Box2D-devel]
   ubuntu: [libbox2d-dev]
+build-essential:
+  ubuntu: [build-essential]
```
wjwwood: We've never needed `build-essential` as a rosdep rule before, can you explain the scenario where you needed this? I think it basically gets pulled in by our other low level dependencies, though maybe that's an oversight on our part. Also this probably won't translate into anything meaningful on other OS's.

mehwang:
@wjwwood That's a good point, so I've removed it.

rosdep/python.yaml
```diff
@@ -1449,6 +1449,10 @@ python-scipy:
     utopic_python3: [python3-scipy]
     vivid: [python-scipy]
     vivid_python3: [python3-scipy]
+python-selenium:
+  ubuntu:
+    pip:
+      packages: [selenium]
```
wjwwood: Just so you know, we cannot build binary packages for Ubuntu for packages which depend on pip only things. So if you have a package which depends on this key, we will not be able to release a deb package automatically.

wjwwood:
I'll let another maintainer double check this before merging it.

rosdep/python.yaml
mikeferguson: @wjwwood we aren't going to be debianising anything using this -- we are just using rosdep as the way to install for CI

dirk-thomas: Why not rely on the Debian package available on Ubuntu: http://packages.ubuntu.com/trusty/python/python-selenium

wjwwood: Well the obvious thing is the version in Trusty is 2.25.0 and 2.46.0 is the latest release. That'd be my guess as to why they want the pip one. But I agree that if there is a Trusty deb and it works we should have that as the default key. Unfortunately there is no good practice for when you need a newer version and pip can provide it, but there is also an apt-get version others might use.

mikeferguson: I'm pretty sure the ubuntu version is indeed ancient and pretty much non-usable at this point. Would changing the keyname to "python-selenium-latest" or "python-selenium-pip" be a better option? (then if someone wanted to use the apt-get version later, they could?)

dirk-thomas: Yes, that sounds like a good compromise.

mikeferguson: any preference as to name? is there any existing convention?

dirk-thomas: A few packages use a `-pip` suffix which looks like a reasonable choice.

mehwang:
I've changed the name to python-selenium-pip as per @dirk-thomas's suggestion.

PR #8917, opened 2015-07-15
adding python-babel

asmodehn:
warning : this package doesn't exist for saucy, and therefore builds depending on it will break for saucy...
ref : https://launchpad.net/ubuntu/+source/python-babel
Should I put a warning somewhere, or somehow specify it differently ? Or maybe not add it here at all ?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -131,8 +131,13 @@ python-avahi:
     utopic: [python-avahi]
     vivid: [python-avahi]
 python-babel:
-  ubuntu: [python-babel]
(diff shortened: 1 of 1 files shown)


tfoote:
The full solution would be to call out each version as in the entry above yours. But as Saucy is already EOL it's find this way. 

The best place to reference the availability of packages is on packages.ubuntu.com like [here](http://packages.ubuntu.com/trusty/python-babel) 

Note your commit failed CI due to alphabetization. fedora comes before ubuntu.

PR #8934, opened 2015-07-17
Adding python-enum34 as a rosdep key

stonier:
Only added to ubuntu trusty as I'm not aware of naming on other platforms.


tfoote:
It's available on trusty and newer. We can just use the versionless `ubuntu` tag http://packages.ubuntu.com/trusty/python-enum34

vrabaud:
please juste write : ```ubuntu: [python-enum34]```

PR #8963, opened 2015-07-22
Add play_motion doc & source jobs

bmagyar:


vrabaud:
ouh yeah, that looks awesome.

PR #8997, opened 2015-07-28
remove sources tag from jsk_hogehoge packages

k-okada:
We realized that our frequent commits to the repository consumes many job resources to your jenkins.ors.org resources, I'm not sure but if removing sources tags stops executing ros-\$ROS_DISTRO-devel-jsk- jobs and that reduces cunsuming ros.org resources, I'm very happy about that.

### DIFF
--- a/hydro/distribution.yaml
+++ b/hydro/distribution.yaml
@@ -3177,10 +3177,6 @@ repositories:
         release: release/hydro/{package}/{version}
       url: https://github.com/tork-a/jsk_3rdparty-release.git
       version: 2.0.2-0
-    source:
-      type: git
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Removing the source entry will also prevent users to use `rosinstall_generator` with the `--upstream(-development)`. They can still use `rosinstall_generator` using the latest release from the gbp repo though. If that is ok for you I can merge this.

In the future (when the new build farm is deployed) we will be able to disable the devel job without affecting the availability of the entry.

k-okada:
thanks you for explanation, that's ok for me. Please merge this.

2015年7月29日水曜日、Dirk Thomas<notifications@github.com>さんは書きました:

PR #9015, opened 2015-07-30
Rosdep for libgv-python

stonier:
Should be available across ubuntu distros that we care about:
- http://packages.ubuntu.com/search?keywords=libgv-python


tfoote:
missing a colon. please amend the commit to fix.

stonier:
Got it, sorry Tully.

PR #9047, opened 2015-07-31
Added python-catkin-lint

alexhenning:


tfoote:
This is in the ROS repos so should be ok.

PR #9085, opened 2015-08-05
Add walrus and responses python packages to rosdep db

mehwang:
Walrus is a redis toolkit for Python.

Responses is a utility library for mocking out the Python requests library.


tfoote:
Are there versions of these in apt? With the pip definitions these dependencies cannot be used in released packages.

mikeferguson:
We're well aware of the pip release issue, that is why we specifically added -pip to the end (if someone wants the APT version, they can still claim the name). I imagine there are not apt versions (or they are horribly out of date).

PR #9091, opened 2015-08-06
remove segbot_apps and segbot_simulator from Indigo distribution

jack-oquin:


jack-oquin:
Is there a problem with this? 

We merged the packages from those two repositories into https://github.com/utexas-bwi/segbot

dirk-thomas:
Please see the announce on the ros-release mailing list: http://lists.ros.org/pipermail/ros-release/2015-August/004661.html

jack-oquin:
It's not critical, but I would like to get a few things into the release pipeline before leaving on vacation next week.

Is it OK to release again, now?

dirk-thomas:
Yes, that's why we have merge all pending PR just minutes ago.

jack-oquin:
OK. 

I'd seen Tully's message about new Indigo packages and assumed it was open for business again.

PR #9104, opened 2015-08-07
Added siemens experimental repo

shaun-edwards:


dirk-thomas:
Please note that hydro is EOL and the farm will be disabled soon (after the next full sync).

PR #9107, opened 2015-08-07
adding overlooked trusty in jade build target

tfoote:


dirk-thomas:
What is the state of this PR?

tfoote:
waiting for review

dirk-thomas:
Did you check that prereleases of Jade for Trusty are not possible now with ros-infrastructure/ros_buildfarm#93 being merged? And that this change makes them work again?

If yes, +1.

dirk-thomas:

tfoote:
Prereleases work in the current state despite this missing information. This is simply making the info reflect [REP 3](http://www.ros.org/reps/rep-0003.html)

dirk-thomas:
If this is not necessary we should close the PR without merging it.

tfoote:
Sure this is not "necessary". Yes it will change out soon when we change formats. This pull request has been pending since the beginning of August.

This is equivalent of a typo. We have 2 of 3 distros written down, the 3rd is missing. If someone looks here we should at least have the best data possible. We should just fix the typo and move on.

PR #9143, opened 2015-08-11
Add postgresql and postgresql-client rosdeps

paulbovbel:


wjwwood:
Seems that it disagrees with you about the alphabetical ordering.

paulbovbel:
@wjwwood , fixed the ordering, but something unknown is causing an error.

wjwwood:
Hmm, looks like someone may have removed the groovy tags from one of their release repositories? @dirk-thomas what do you think?

dirk-thomas:
Travis failed to fetch the cache and is trying to build it from scratch. This fails then since e.g. the repo containing `beliefstate` was moved. Close / reopen should retrigger travis...

PR #9154, opened 2015-08-13
Add python modules required for rospeex in rosdep/python.yaml

komeisugiura:
The modules are required for the next version of rospeex, which will be updated later.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -118,6 +118,19 @@ python-argparse:
       packages: []
     vivid:
       packages: []
+python-autobahn:
+  debian:
+    pip:
+      packages: [autobahn]
+  fedora:
+    pip:
+      packages: [autobahn]
+  osx:
+    pip:
+      packages: [autobahn]
+  ubuntu:
+    pip:
+      packages: [autobahn]
 python-avahi:
   arch: [avahi]
   debian: [python-avahi]


dirk-thomas:
Please consider using the available Debian package instead. Otherwise you won't be able to release any packages depending on this.

komeisugiura:
Do you mean the package should be obtained by apt-get instead of pip? If so, the solution would be the following. What do you think? 

Before:
```
  debian:
    pip:
      packages: [autobahn]
```
After:
```
   debian: [python-autobahn]
```

dirk-thomas:
Yes, correct. The same for Ubuntu.

komeisugiura:
For Ubuntu 14.04 or later, we can install python-autobahn by apt-get. For Ubuntu 13.10 or earlier, python-autobahn is not included in the default packages. Therefore, I think the following will work. What do you think?

```
+python-autobahn:
+  debian: [python-autobahn]
+  fedora:
+    pip:
+      packages: [autobahn]
+  osx:
+    pip:
+      packages: [autobahn]
+  ubuntu:
+    lucid:
+      pip:
+        packages: [autobahn]
+    maverick:
+      pip:
+        packages: [autobahn]
+    natty:
+      pip:
+        packages: [autobahn]
+    oneiric:
+      pip:
+        packages: [autobahn]
+    precise:
+      pip:
+        packages: [autobahn]
+    quantal:
+      pip:
+        packages: [autobahn]
+    raring:
+      pip:
+        packages: [autobahn]
+    saucy:
+      pip:
+        packages: [autobahn]
+    trusty: [python-autobahn]
+    utopic: [python-autobahn]
+    vivid: [python-autobahn]

```

tfoote:
That looks good. Though I'd suggest not including the past EOL'd distros. We don't have a policy, but usually retain existing records for EOL'd distros. However adding EOL'd distros doesn't make a lot of sense.

komeisugiura:
Thank you for your suggestion. EOL'd distros are included because some users want to install from source code.

I'll close this and send another pull request.

PR #9155, opened 2015-08-13
Added rule for pyGPs

lteacy:
Added rule for pyGPS - python Gaussian Processes library.

See http://www-ai.cs.uni-dortmund.de/weblab/static/api_docs/pyGPs/index.html

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1984,3 +1984,16 @@ python-zmq:
 xdot:
   fedora: [python-xdot]
   ubuntu: [xdot]
+python-pyGPs:
+  debian:
+    pip:
+      packages: [pyGPs]
+  fedora:
+    pip:
+      packages: [pyGPs]
+  osx:
+    pip:
+      packages: [pyGPs]
+  ubuntu:
+    pip:
+      packages: [pyGPs]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please insert in alphabetical order.

I assume that there is no Debian package for it? Just as a note: a package depending on this can't be released.

lteacy:
Done. No Debian package that I can find. 
cheers!

tfoote:
Capitalization needs to be taken into account. The capitals should be in front of all the lower case elements. The test results can be seen here: https://travis-ci.org/ros/rosdistro/builds/75578704

You can test it locally with the script check_rosdep.py in the scripts folder. 

Though with that said, I'd suggest changing the key to python-pygps-pip since lowercase is more common and this is a pip only package and it will not be ambiguious/conflict in the future if it gets packaged into debian packaging.

lteacy:
Ok, I've renamed this as per Tully's suggestion, and checks appear to pass.
The capitalisation of the package name itself is a bit unfortunate, though it appears we have to leave this as is in order to correctly detect installation. I guess the author's intention was to avoid confusion with GPS as opposed to GPs (Gaussian Processes).

Let me know if there is anything else.

tfoote:
Thanks looks good. In the future if you can squash the commits it will leave a cleaner history, but I'll merge it for now anyway.

PR #9205, opened 2015-08-18
Moves saucy python-websocket from pip to apt for use in bloom

rctoris:


tfoote:
Is this the same version?

vrabaud:
The pip version is 0.32: https://pypi.python.org/pypi/websocket-client
The Ubuntu version is 0.11: http://old-releases.ubuntu.com/ubuntu/pool/universe/p/python-websocket-client/
@rctoris , does that work with you ?

rctoris:
Thanks for checking! Yes that does work.

PR #9209, opened 2015-08-18
fix wrong commit : key openssl should be openssl, and add openssh key for backword compatibility

k-okada:


k-okada:
original commit (https://github.com/ros/rosdistro/pull/8864) was wrong

PR #9263, opened 2015-08-24
Add mplayer

mvollrath:


rosdep/base.yaml
```diff
@@ -2612,6 +2612,23 @@ mpg123:
...
+mplayer:
+  arch: [mplayer]
+  debian:
+    jessie: [mplayer]
+    sid: [mplayer]
+    squeeze: [mplayer]
+    wheezy: [mplayer]
+  fedora: [mplayer]
+  ubuntu:
+    precise: [mplayer]
```
tfoote: Since these entries are all the same please put a single entry under ubuntu and debian like you did with fedora instead of calling out each individual version.

vrabaud:
@mvollrath , any update on that ?

dirk-thomas:
Closing due to no response.

mvollrath:
It was done this way because precise is the first version of Ubuntu where an mplayer package was available.  Is an apt-get error on oneiric friendlier than a rosdep error?

tfoote:
Also, all distros older than precise are EOL as such we don't want to make our life more complicated to add support for an unsupported distro.

PR #9264, opened 2015-08-25
Adding evapi_ros to documentation index for indigo

makcakoca:
I'd like evapi_ros to be indexed and documented on ros.org.


tfoote:
Your branch is a few days out of date. It needs to be rebased against a fresh copy of master.

PR #9265, opened 2015-08-25
rebased against a fresh copy of master for evapi_ros

makcakoca:
I'd like evapi_ros to be indexed and documented on ros.org.


tfoote:
This looks like it's replacing #9264 but it's also reverting #9245 I think you need to check your merge conflict.

PR #9305, opened 2015-08-26
move ackermann_msgs ownership to ros-drivers

jack-oquin:


tfoote:
This is failing because of a change to an EOL'd distro. It should be ok to merge as this repo has actually moved and it will be helpful for introspectability to use the new repo location.

jack-oquin:
That is what I figured, but am not sure how to handle the error.

jack-oquin:
@tfoote: what do I need to do to get this in shape so it can be merged?

Right now all the build farm links are broken, because it was moved from the old location and that copy no longer exists.

Should I resubmit without the Groovy update? It does not seem to mind the Electric and Fuerte changes.

tfoote:
There's nothing to do to make it pass (with all the info) We just need to manually review it closely and merge it anyway. 

@dirk-thomas @wjwwood we'll need to review it closely. 

Note that if you used github's move method, it will actually forward most requests correctly until you recreate a repo with the same name.

dirk-thomas:
LGTM. All branches and tags are there.

jack-oquin:
I did use the github move method (Settings>Transfer). It seems to work OK and there *is* forwarding in place, at least for the moment.

PR #9310, opened 2015-08-27
Fix Formulae found in multiple taps

balzer82:
Multiple Locations fix

```
$ brew info tango-icon-theme --json=v1
Error: Formulae found in multiple taps:
 * ros/deps/tango-icon-theme
 * ros/groovy/tango-icon-theme

Please use the fully-qualified name e.g. ros/deps/tango-icon-theme to refer the formula.
```

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -652,7 +652,7 @@ swig-wx:
 tango-icon-theme:
   osx:
     homebrew:
-      packages: [tango-icon-theme]
+      packages: [ros/deps/tango-icon-theme]
 tar:
   osx:
     homebrew:


wjwwood:
The idea of not using the fully qualified name is so that you can get different versions of it when you have different taps tapped. You happen to have two incompatible taps tapped at the same time. If you look at the from source instructions it mentions that you'll need to untap old ROS distro taps:

http://wiki.ros.org/indigo/Installation/OSX/Homebrew/Source#Homebrew
> Note: if you are upgrading from previous ROS distro try:
> `$ brew untap ros/DISTRO`

I believe the right fix for you is to untap `ros/groovy`.

PR #9347, opened 2015-09-01
Adding riskrrt to documentation index for distro

gregvi:
I'd like riskrrt to be indexed and documented on ros.org


vrabaud:
```doc``` does not have the proper indentation

PR #9393, opened 2015-09-07
Adding voronoi_planner to documentation index for distro    

frontw:
I'd like voronoi_planner to be indexed and documented on ros.org.


vrabaud:
branch indigo does not exist

frontw:
Branch indigo added.
Would it retry automatically?

vrabaud:
no: how would it know when to retry ? Just triggered it though so we'll see.

vrabaud:
you can also set your version to master if you know you only have one branch

PR #9399, opened 2015-09-08
Add python-path.py entry

wkentaro:


dirk-thomas:
The package name is a bit weird but it looks like that's what upstream chose...

PR #9404, opened 2015-09-08
Removing evapi_ros release since packages for arm

makcakoca:
I'd like to remove evapi_ros release.


vrabaud:
@makcakoca was the original person who added those.

PR #9412, opened 2015-09-08
Adding dynamicvoronoi to documentation index for distro

frontw:
Please add dynamicvoronoi to documentation index for distro.
I want to publish catkinized version of http://wiki.ros.org/dynamicvoronoi which I use in another package.


dirk-thomas:
The two commits together are a no-op?

PR #9421, opened 2015-09-09
added vicon_bridge for hydro,indigo,jade

markusachtelik:


dirk-thomas:
Changing groovy does not have any effect anymore. Since we are not doing any changes to groovy anymore please remove the groovy change from the PR.

markusachtelik:
ok, groovy changes removed

PR #9426, opened 2015-09-10
adding riskrrt to doc index

gregvi:
I'd like riskrrt to be indexed on ros


vrabaud:
https://github.com/spalanza/riskrrt_ros.git does not have an indigo-devel branch

PR #9428, opened 2015-09-10
Changed installation method for autobahn in rosdep/python.yaml

komeisugiura:
I'm working on updating rospeex, multilingual cloud-based speech recognition and synthesis module, and I'd like to use pip for required module (autobahn) instead of apt-get. The reason is as follows.

Based on the comments in the following discussions, I've been trying to use apt-get for installing autobahn.
https://github.com/ros/rosdistro/pull/9154

The problem is that the autobahn package installed by apt-get is sometimes very old (=autobahn-0.5) in Debian and Ubuntu. There are many differences between autobahn-0.5 and autobahn-0.10, and adapting rospeex to both versions will take considerable amount of time.

Considering the fact that pip can install the latest autobahn (and that some packages already use pip even if apt-get can be used), I'd like to install autobahn-0.10 by pip.

By the way, the autobahn versions installed by Ubuntu's apt-get are as follows:
13.10 and earlier: no apt-get package provided for autobahn.
14.04: python-autobahn-0.5
14.10: python-autobahn-0.5
15.04: python-autobahn-0.5
15.10: python-autobahn-0.10

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -121,7 +121,9 @@ python-argparse:
     vivid:
       packages: []
 python-autobahn:
-  debian: [python-autobahn]
+  debian:
+    pip:
+      packages: [autobahn]
   fedora:
     pip:
       packages: [autobahn]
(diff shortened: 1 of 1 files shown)


tfoote:
You cannot switch to pip dependencies and release debian packages based on them. Which means that you will not be able to release rospeex as debian packages anymore. [here](https://github.com/ros/rosdistro/blob/5740c1d3b0d965d994d49ea003e9d2b25650af31/indigo/distribution.yaml#L8298)

If you really do want a pip rule please create a second definition autobahn-pip to not interfere with the autobahn target that others might be using.

PR #9435, opened 2015-09-11
Add pyquery entry

iory:


tfoote:
It looks like there's a deb for this python-pyquery which we should use instead: http://packages.ubuntu.com/trusty/python/python-pyquery

iory:
OK, updated

PR #9442, opened 2015-09-14
added repositories for cassandra_ros, rosR, rosR_demos, and rosshell

andre-dietrich:
would like to add the following packets to ros indigo ...


vrabaud:
```version``` for your sources should be the branch where you get the data from. So I guess it should be ```master``` in oyur case.

vrabaud:
thx ! dont hesitate to release for Jade too.

PR #9450, opened 2015-09-15
Removed pip-related entries from python-autobahn in rosdep/python.yaml

komeisugiura:
According to the comments in https://github.com/ros/rosdistro/pull/9154 and https://github.com/ros/rosdistro/pull/9428 , pip installation entries for python-autobahn are removed.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -122,37 +122,15 @@ python-argparse:
       packages: []
 python-autobahn:
   debian: [python-autobahn]
-  fedora:
-    pip:
-      packages: [autobahn]
-  osx:
-    pip:
-      packages: [autobahn]
   ubuntu:
-    lucid:
-      pip:
-        packages: [autobahn]
-    maverick:
-      pip:
-        packages: [autobahn]
-    natty:
-      pip:
-        packages: [autobahn]
-    oneiric:
-      pip:
-        packages: [autobahn]
-    precise:
-      pip:
-        packages: [autobahn]
-    quantal:
-      pip:
-        packages: [autobahn]
-    raring:
-      pip:
-        packages: [autobahn]
-    saucy:
-      pip:
-        packages: [autobahn]
+    lucid: []
+    maverick: []
+    natty: []
+    oneiric: []
+    precise: []
+    quantal: []
+    raring: []
+    saucy: []
     trusty: [python-autobahn]
     utopic: [python-autobahn]
     vivid: [python-autobahn]


tfoote:
We use the empty entry to imply that the dependency is implicitly resolved and nothing is needed to be installed. That does not look right in this case. What is your intent in making these rules as such?

komeisugiura:
You are right. I intended to solve the problem of the compatibility between python-autobahn and Ubuntu 13.10.

The points are:
- python-autobahn is unavailable for Ubuntu 13.10 or earlier.
- If we keep the pip entry for Ubuntu 13.10, any package using python-autobahn cannot be released as a .deb package since bloom-release will fail.
- Since it is not only the problem of rospeex, I recommend to delete the pip entry for 13.10.
- Simply deleting the entry will result in unresolved dependency for Ubuntu 13.10. This is why I proposed to use the empty entry (=`saucy: []`) since bloom-release will not fail.

I'm wondering if there could be a good idea to ignore Ubuntu 13.10 in bloom-release. If not, what would be the best idea to solve the above problem?

tfoote:
Since Saucy is EOL we do plan to discontinue builds for it at some point. However we are not setup to do that yet. There is some infrastructure work to do to parameterize it cleanly.

If your package depends on python-autobahn how does it work if you simply ignore the dependency?

In general we want to make it such that the rosdep rules reflect what's available in ubuntu which the current state seems to be most accurate.

komeisugiura:
If the line `<run_depend>python-autobahn</run_depend>` is removed in `package.xml`, rospeex can be released. However, when a user tries to install rospeex, he has to run two commands.
- Ubuntu 14.04 users: `# apt-get install ros-indigo-rospeex python-autobahn` is required.
- Ubuntu 13.10 users: `# apt-get install ros-indigo-rospeex; pip install autobahn` is required.
(I think the above ways are not ideal in terms of package management.)

As far as I understand, the problem is not specific to rospeex, and can be generalized to other packages using python-autobahn. 
I'm wondering this problem can further be generalized to the case where:
- A ROS distro supports two or more distros
- xx package contains pip-install entry for the former distro (e.g. 13.10)
- xx package contains .deb-install entry for the latter distro (e.g. 14.04)
- ->  bloom-release will fail.

Is this correct?

Since rospeex's development team has just finished rewriting websocket-related codes with python-tornado (instead of python-autobahn), we can release rospeex even if this pull-request is not accepted. 

I appreciate if you could decide the policy whether to accept the change or not.

tfoote:
Yes, your generalized statement is correct. When releasing software for a ROS distro you need to be aware of the low water mark for available supporting libraries in all targeted distros as defined in [REP 3](http://www.ros.org/reps/rep-0003.html)

I'll close this PR as it reflects what is available from upstream in the current state. Zeroing the dependency on the older platforms hides the error but not the actual problem of the missing dependency.

wjwwood:
If the dependency is optional, you could include it in the package.xml, but for bloom you could exclude the dependency from the debian control file. That way the deb doesn't depend on it, but when someone is building from source `rosdep` will fetch it.

I don't recommend this, since it's a lot of work and prone to errors, but it is an option technically.

komeisugiura:
Thank you for your comments. I agree with closing this issue.

PR #9458, opened 2015-09-15
Add python-redis-pip to python packages

mehwang:
The apt package for python-redis installs redis-py 2.7.2 (last updated 2012) and doesn't include a bunch of bug fixes and useful features.

The pip package has the more recent redis-py 2.10.3 (updated 8/2014).


tfoote:
This is fine, but note that you will not be able to release debian packages against this version.

PR #9476, opened 2015-09-17
Added three pip packages to python.yaml.

peterpolidoro:
I added three pip packages (mightex_device, zaber_device, and modular_device) to python.yaml.


wjwwood:
@peterpolidoro can you squash this? You have several merge commits and a cleanup commit which are extraneous.

Also, just so you know, `pip` only dependencies cannot be used when releasing a package because we cannot create `.deb`'s which depend on `pip` packages.

PR #9525, opened 2015-09-21
Correct spelling of ubuntu for python-jasmine-pip

mehwang:
This was a silly typo.

@tfoote: @mikeferguson tells me that the build probably shouldn't have passed in the first place with the typo, and to let you know that your tests might not be working completely.


tfoote:
Well you could have been developing for the new os `ubutnu` ;-)

We don't validate the keys to be able to support potential new OSs being added which could be used with an OS override in rosdep or when bringing up support for a new OS.

Issue #9569, opened 2015-09-28
media-gfx/assimp is now media-libs/assimp


Alessandro-Barbieri:
https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L95
is referencing old name

tfoote:
If you could submit a PR with the change it would be greatly appreciated. 

@allenh1 FYI

allenh1:
Woops. I thought I fixed that one. I think media-libs/assimp is already mentioned in the yaml, so the media-gfx/assimp should just be removed.

allenh1:
I see. Nevermind. This pull request should fix the issue. Thanks @Alessandro-Barbieri

tfoote:
Fixed in #9571

PR #9571, opened 2015-09-28
media-gfx/assimp entry to media-libs/assimp

Alessandro-Barbieri:
https://github.com/ros/rosdistro/issues/9569


tfoote:
Thanks for the PR!

Should be good according to @allenh1 in the ticket.

PR #9637, opened 2015-10-13
add time for fedra and debian

k-okada:
https://github.com/ros/rosdistro/pull/9604 missing fedora/debian entry

https://admin.fedoraproject.org/pkgdb/package/time/
https://packages.debian.org/jessie/time


wjwwood:
Alphabetization is off. :smile:

k-okada:
oh, fixed.

PR #9639, opened 2015-10-13
Rename CONTRIBUTING.md to CONTRIBUTING

dirk-thomas:
Otherwise the file is not being referenced when create PRs.


wjwwood:
Should work as is: https://github.com/blog/1184-contributing-guidelines#contributors-wanted-inquire-within

dirk-thomas:
Nvm. I checked it wrong.

PR #9644, opened 2015-10-13
added new rosdep rule for python-rrdtool

airballking:


tfoote:
Your block is inserted out of alphabetical order

airballking:
Thanks for catching this! Corrected it.

PR #9726, opened 2015-10-28
Adding rsv_balance, rsv_balance_desktop and rsv_balance_simulator for doc index

vmatos:


jade/distribution.yaml
```diff
@@ -3650,6 +3650,51 @@ repositories:
...
+  rsv_balance:
+    doc:
+      type: git
+      url: https://github.com/robosavvy/rsv_balance.git
+      version: master
+    release:
+      packages:
+      - rsv_balance_description
+      - rsv_balance_msgs
+      url: https://github.com/robosavvy/rsv_balance.git
```
dirk-thomas: This is not a valid `release` repository (same fore the other release entries). Please read the bloom tutorial on how to release a ROS package: http://wiki.ros.org/bloom

vmatos: Thank you,
It was unclear from the tutorial how could I index my packages for wiki only, without binary release.

As far as I can deduce, should I delete the lines pertaining "release:" from my entries?
Such like: cd9f2b68341b6c565cb96976f75f9fd605b2533d

dirk-thomas: Yes, that would be sufficient for documentation.

But I assume your repositories build on-top of each other. Therefore you might also want to remove the `source` blocks since the continuous integration jobs will fail if the dependencies are not available as Debian packages.

vmatos:
Thank you for the help.

Could we clarify on the [wiki](http://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation) which fields should we use in which cases?

dirk-thomas:
I added a small note that only the doc section needs to be added: http://wiki.ros.org/action/diff/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation?action=diff&rev1=7&rev2=8

PR #9737, opened 2015-10-31
Adding depth_nav_tools to documentation index for indigo

mdrwiega:
I would like depth_nav_tools to be indexed and documented on ros.org.


tfoote:
The indigo-devel branch does not exist on that repo. You either need to create that branch or index master which exists.

mdrwiega:
Thank you for information. I just renamed my master branch to indigo-devel.

PR #9733, opened 2015-10-31
Remove broken jobs

tfoote:
We have a lot of repeatedly failing jobs that we're trying to rebuild ever night. This will stop them from trying to build until they are rereleased.

PR #9739, opened 2015-11-01
re-enable eus_qp

k-okada:
https://github.com/ros/rosdistro/pull/9733 should be fixed in https://github.com/jsk-ros-pkg/jsk_control/pull/511

I'm not hurry so please merge when you're ok.


tfoote:
thanks for the quick response.

PR #9760, opened 2015-11-03
Add rule for python-box2d

wrvb:
[box2d](http://box2d.org/) is a `c++` library for two dimensional physics
simulation.  The [python-box2d](https://github.com/pybox2d/pybox2d) package
provides python bindings for `box2d`, and is available in the Ubuntu
repositories for all currently supported releases (including precise, trusty,
vivid, and wily).


wrvb:
Wow, that was fast!

PR #9772, opened 2015-11-06
Needed dependency (for package trac_ik) for the libnlopt headers and libraries on Ubuntu

pbeeson:


tfoote:
Please keep them as separate dependencies like your first commits, with the same names as the ubuntu versions. That way we can separate the build and runtime depends for more optimized packaging. Note it failed CI due to being out of alphabetical order and there's a trailing space on the `ubuntu:` line

pbeeson:
It might be useful on the contributing docs http://docs.ros.org/independent/api/rosdep/html/contributing_rules.html to mention the check_rosdep.py script.

tfoote:
I rebased out the fixups and made this generic to all ubuntus in: 2d531eb89826401e821e164fc4ad8697a67f5b6d

tfoote:
@pbeeson You can now use the rosdep rules in your package.xml Thanks for the suggestion for improving the docs. I opened a PR to improve it https://github.com/ros-infrastructure/rosdep/issues/429

PR #9803, opened 2015-11-10
Adding stereo_slam to documentation index for distro

plnegre:


indigo/distribution.yaml
```diff
@@ -9958,6 +9958,11 @@ repositories:
       type: git
       url: https://github.com/wunderkammer-laboratory/steered_wheel_base_controller.git
       version: master
+  stereo_slam:
+    doc:
+      type: git
+      url: https://github.com/srv/stereo_slam
```
dirk-thomas: Must end with `.git`.

Same below.

dirk-thomas:
It looks like you recently changed the branch name which lets the Jenkins jobs fail: http://build.ros.org/job/Jdoc__stereo_slam__ubuntu_trusty_amd64/12/console Please fix the version entries.

plnegre:
I don't know what is wrong... The value of version tag on 'ros/distro/indigo/distribution.yaml' is 'indigo' and, on my repo (https://github.com/srv/stereo_slam.git) I have a branch with this name.

tfoote:
The problem is in the Jade entry which references `indigo-devel` branch in `jade/distribution.yaml`

PR #9858, opened 2015-11-24
Adding python-tblib

asmodehn:


rosdep/python.yaml
```diff
@@ -1849,6 +1849,9 @@ python-sympy:
     raring: [python-sympy]
     saucy: [python-sympy]
     trusty: [python-sympy]
+python-tblib:
+  ubuntu:
```
tfoote: This is failing CI due to a trailing space on this line.

rosdep/python.yaml
```diff
@@ -1849,6 +1849,9 @@ python-sympy:
     raring: [python-sympy]
     saucy: [python-sympy]
     trusty: [python-sympy]
+python-tblib:
+  ubuntu: 
+    trusty: [python-tblib]
```
dirk-thomas: Only adding it for Trusty won't allow you to release any package. Please fill in additional Ubuntu versions.

asmodehn:
Saucy doesn't have this package... I think this might break the deb build on packages that depend on this, is there a way to fix it ?

tfoote:
Generally we want all dependencies to be available on target platforms. 

Typically for something like this we'd consider importing a backport from a ppa if the dependency is necessary and not available otherwise. 

We also plan to turn off Saucy builds since it's reached EOL. Once that has happened builds will work again. But until then we want to keep compatibility.

asmodehn:
So this will be mergeable when saucy is dropped ? any ETA ?
I would like to avoid doing a backport for an obsolete distro...

asmodehn:
Actually, how about I do a third-party release of  https://github.com/ionelmc/python-tblib ?

I recently found out that the trusty version was too old and didn't work as expected... I might need to use a more recent version, but I am not sure how the dependency will be resolved if there is both a rosdep key, and a rosdistro key...

Doing the thirdparty release would also probably satisfy saucy builds...

tfoote:
Yes a 3rdparty build would satisfy saucy and hopefully get you the needed version.

asmodehn:
In the end I might eventually do this a different way...
Anyway I don't need it at the moment.
I ll reopen if needed.

PR #9863, opened 2015-11-24
Adding rangeonly_driver to documentation index for indigo

felramfab:
I'd like rangeonly_driver to be indexed and documented on ros.org.


dirk-thomas:
In order to be documented you need to add the `doc` block instead.

The `source` block will result in a CI job which will build your repo after every commit.

PR #9879, opened 2015-11-26
Use fltk --devel formula for el capitan

jmtatsch:
because the flak formula either does not compile or function as expected on OS X
versions newer than Yosemite due to an upstream incompatibility.

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -89,8 +89,22 @@ flex:
       packages: []
 fltk:
   osx:
-    homebrew:
-      packages: [fltk]
+    "el capitan":
+      homebrew:
+        install_flags: [--devel]
+        packages: [fltk]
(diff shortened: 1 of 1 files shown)


tfoote:
looks reasonable to me. @wjwwood can you take a look?

wjwwood:
This should be fixed upstream. In my opinion it would be better to add this as a workaround to the troubleshooting guide along with a description of the problem. In the mean time a bug should be filed with the upstream (Homebrew or FLTK or both). I'm still not sure how or why it's failing, you neither explained the failure nor linked to an issue about it. I don't think this pr makes sense because `--devel` will change over time and has the largest chance to be broken over that time period so baking it into our rosdep rules isn't a good idea in my opinion.

wjwwood:
I'll close this for now and if new evidence suggests we should reconsider it I can reopen the issue then.

PR #9937, opened 2015-12-03
Remove libfovis release

plnegre:


dirk-thomas:
Hydro is already EOL and there is no build farm for it anymore. Therefore we usually don't accept any PR for it.

@tfoote This change should not have passed the tests in the first place. Can you please check and update the tests to catch this in the future.

tfoote:
Yeah, we keep the old rosdistro entries in place as documentation for where the packages came from and to support people building from source. If something's hosting is relocated we can override and update the entry, but other than that our policy is to maintain the rosdistro in it's final form.

PR #9989, opened 2015-12-07
Adding iav_depthimage_to_laserscan to documentation index for distro

iav-student:
I'd like iav_depthimage_to_laserscan to be indexed and documented on ros.org.

### DIFF
--- a/hydro/distribution.yaml
+++ b/hydro/distribution.yaml
@@ -2839,12 +2839,6 @@ repositories:
       url: https://github.com/code-iai/iai_common_msgs.git
       version: master
     status: developed
-  iav_depthimage_to_laserscan:
-    doc:
-      type: git
-      url: https://github.com/iav-student/iav_depthimage_to_laserscan.git
-      version: hydro
-    status: maintained
   iiwa:
     release:
       tags:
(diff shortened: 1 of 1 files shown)


iav-student:
iav_depthimage_to_laserscan

dirk-thomas:
Hydro had been EOL quite a while ago and there is no build farm anymore. Therefore we don't accept PR for it anymore. Please consider adding it to Indigo or Jade instead.

@tfoote I don't know how this is being flagged as passing by Travis. Please double check the tests which seem to be not working as desired.

tfoote:
@dirk-thomas we only check doc and release since those are the things which have generated outputs which will never change. We don't assert that the source entries don't change at the moment since they may still be used by the rosinstall generator even for EOL distros.

dirk-thomas:
@tfoote Please feel free to reopen the PR then and merge it.

iav-student:
Then must I adding it to the PR Indigo or Jade or it will be reopened?

tfoote:
@iav-student I reopened it and merged it.

iav-student:
How I index a API in http://docs.ros.org/api/ now? I thought that it'd be automatic.

tfoote:
`source` entries are for continuous integration and rosinstall generation. You would need a `doc` entry to have documentation added. However as mentioned earlier in this thread Hydro is EOL so there are no more doc jobs running.

PR #10016, opened 2015-12-10
Adding iav_depthimage_to_laserscan to documentation index for indigo

iav-student:
I'd like iav_depthimage_to_laserscan to be indexed and documented on ros.org.


iav-student:
The build for my package in jenkins has failed, what it could be?

PR #10075, opened 2015-12-16
relocate docs for utexas-art repository

jack-oquin:
The sources moved from googlecode to github.

### DIFF
--- a/doc/electric/art_vehicle.rosinstall
+++ b/doc/electric/art_vehicle.rosinstall
@@ -1,3 +1,3 @@
 - svn:
     local-name: art_vehicle
-    uri: https://utexas-art-ros-pkg.googlecode.com/svn/trunk/stacks/art_vehicle
+    uri: https://github.com/austin-robot/utexas-art-ros-pkg.git
--- a/doc/fuerte/art_vehicle.rosinstall
+++ b/doc/fuerte/art_vehicle.rosinstall
@@ -1,3 +1,3 @@
 - svn:
     local-name: art_vehicle
-    uri: https://utexas-art-ros-pkg.googlecode.com/svn/trunk/stacks/art_vehicle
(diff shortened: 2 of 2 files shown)


wjwwood:
This will not cause the documentation to be rebuilt, but at least it's updated in our records.

jack-oquin:
Is there no longer any way to get the docs rebuilt?

dirk-thomas:
No, the build farm only operates on the current distros. The jobs for EOLed distros are being removed shortly after announcing the EOL.

wjwwood:
You can build them manually.

jack-oquin:
The last actual release of these packages was Fuerte. They are no longer maintained, and were never converted to catkin. 

Since there has recently been some interest in them, I was hoping to get the docs working again.

Is it possible to build docs for rosbuild packages on Indigo?

dirk-thomas:
No, the last distribution which supported rosbuild documentation was Hydro: http://www.ros.org/reps/rep-0003.html#indigo-igloo-may-2014

wjwwood:
The fuerte documentation still exists: http://docs.ros.org/fuerte/api/art_common/html/index.html

jack-oquin:
Yes, it does. 

The problem may be less pervasive than I had initially thought. The main difficulty I've noticed (so far) is that wiki references like ``<<MsgLink(art_msgs/BrakeCommand)>>`` are not getting resolved correctly. The message pages actually still exist, but the references don't work any more.

See: http://wiki.ros.org/art_msgs

Is this just a bug in the ``MsgLink`` macro? The macro links to:

 * http://docs.ros.org/api/art_msgs/html/msg/BrakeCommand.html

The actual page is here:

 * http://docs.ros.org/groovy/api/art_msgs/html/msg/BrakeCommand.html

Should I open a roswiki issue for this?

wjwwood:
https://github.com/ros-infrastructure/roswiki/issues/137

PR #10093, opened 2015-12-18
[indigo/csm] Correct branch to be used

130s:
I'm not sure why the DEB 1.0.2-2 of this 3rd-party package seems to be working although the branch name is wrong.


k-okada:
because this section is just for doc and sources, I think

130s:
> because this section is just for doc and sources

You're right.

Then there should be no problem in merging this (as done already).

PR #10097, opened 2015-12-19
md49_base_controller: 0.1.0-0 in 'indigo/distribution.yaml' [bloom]

Scheik:
Increasing version of package(s) in repository `md49_base_controller` to `0.1.0-0`:
- upstream repository: https://github.com/Scheik/md49_base_controller.git
- release repository: https://github.com/Scheik/md49_base_controller.git
- bloom version: `0.5.20`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -4866,6 +4866,25 @@ repositories:
       type: git
       url: https://github.com/mikeferguson/maxwell.git
       version: indigo-devel
+  md49_base_controller:
+    doc:
+      type: git
+      url: https://github.com/Scheik/md49_base_controller.git
+      version: master
+    release:
+      packages:
+      - custom_messages
+      - md49_base_controller
+      - serialport
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/Scheik/md49_base_controller.git
+      version: 0.1.0-0
+    source:
+      type: git
+      url: https://github.com/Scheik/md49_base_controller.git
+      version: master
+    status: developed
   media_export:
     doc:
       type: git


indigo/distribution.yaml
```diff
@@ -4866,6 +4866,25 @@ repositories:
...
       url: https://github.com/mikeferguson/maxwell.git
       version: indigo-devel
+  md49_base_controller:
+    doc:
+      type: git
+      url: https://github.com/Scheik/md49_base_controller.git
+      version: master
+    release:
+      packages:
+      - custom_messages
```
dirk-thomas: I am not sure if this is a good name for a package...

wjwwood:
I agree about the `custom_messages` package name, please keep in mind that your package names are in a flat namespace with everyone else's packages too. So unless your package is truly generic and reusable, please consider prefixing it with the project or organization name.

tfoote:
It's also *highly* recommended to keep the upstream and release repos separated. A release repo produces a lot of redundant data from the source repo which is optimized for the release process, and is causes extra overhead when checking out the upstream source if combined in a single repo.

Scheik:
Hi, thanks for your comments. This was my first pull request to contribute to ROS. Maybe i am a bit to inexperienced. Intention was to have a Package/Stack that works with MD49, which is a dual motor driver board driven by serial port. 

The custom_message package just provides the ROS messages that are published by the md49_base_controller node (like encoderdata and other data from MD49, as current speed, acceleration and so on) Instead of renaming the custom_message package i think i will "merge" the custom messages back into md49_base_controller package himself.

The serialport package works as a library, which is used from md49_base_controller, to establish and handle the serial communictaion with the MD49. Maybe i also should find another solution for that?

I think, best way is to rethink some things again, so i will just close the pull request for now.

wjwwood:
@Scheik It's up to you, but I'll just provide some advice on a few of the points.

For the `custom_messages` package I'd recommend renaming, probably to something like `md49_messages`. That way if someone wants to communicate with your controller node using these messages on a different computer, then they only need to build your messages and not the whole controller on their computer. This is a common patter in the ROS community (to separate the messages from the rest in a different package).

If the `serialport` package is truly generic, then the name might be ok, but there's really no reason you couldn't call it `md49_serialport` just to be safe. You can search for similar packages already released, there are a few ways:

- ROS wiki: http://www.ros.org/browse/search.php?distro=indigo&q=serial
- ROS Index (community project): http://rosindex.github.io/search/?q=serial
- Google: https://www.google.com/?gws_rd=ssl#q=ros+package+serial

Through those you'd probably find:

- serial: http://wiki.ros.org/serial
 - Full disclose (I wrote this).
- http://wiki.ros.org/cereal_port

Both of which might be something you could reuse/improve to your needs rather than make a new package.

Hopefully that helps you figure out what to do next.

Scheik:
Thanks a lot for the advice, i really appreciate that! 

At first i now figured out, how and why it is recommended to separate the repos into an upstream- (the one i develop in) and the release- repo. Learning curve is still high after a year of using ROS now :-) Especially the way contributing to official repos was new for me, although i worked a lot with github with my own repos already. The fear of doing something wrong and causing trouble is high.

What i did yesterday, was to "merge" everything into a single package and send a pull request to rosdistro again, which was merged successfully already.

The released package is working, but still under development, so i think it is no problem to change some things again. In the future I will release the messages again as single package with a more clearer name, the advantage you described is obvious and was the reason i had them in a separate package in the beginning.

With the package for serial communications I'm not 100% sure for now, how to handle it. Although it is usable in a generic way to handle serial communications on Linux, i tend to release it as a package as part of the md49- stack at the moment. But maybe it will be better, to use an already released package, as you described. I think that will be closer to the general philosophy of ROS.

As you could have figured out already, i'm not a professional developer, more a hobbyist. Because of that i'm reliant to every advice from more experienced users and glad about every information i get, e.g. if that isn't the right place to discuss my beginners problems here ;-)

PR #10117, opened 2015-12-21
added Edinburgh Speech Tools Library

niyawe:


tfoote:
Can you make the key for this `libestools-dev` we try to mirror the debian names for things for simplicity. Especially something like `est` is too short and not unique enough to be trivially introspectable by someone not knowing what it is already.

niyawe:
done

tfoote:
Thanks

PR #10135, opened 2015-12-23
Adding patrolling_sim repo to documentation index for hydro and indigo

davidbsp:
I'd like my repo to be indexed and documented on ros.org. Thanks!


dirk-thomas:
Hydro had been EOLed a while ago and we don't operate build farm for it anymore. Please only add the repo to Indigo (and if you like to also to Jade).

davidbsp:
Thanks. I will create a new pull request.

PR #10035, opened 2015-12-11
Adding forte_rc_robot to documentation index for distro

aaraujo11:
Please, i'd like forte_rc_robot to be indexed and documented on ros.org.

PR #10163, opened 2015-12-30
remove forte_rc_robot from index due to continuously failing builds.

tfoote:
This reverts #10035


tfoote:
@aaraujo11 FYI I'm removing the binary builds from the distro since they are failing to build. You appear not to have pushed all the necessary tags that bloom created. It looks like you manually create your indigo track. In general it's recommended to let bloom auto create that for you.

aaraujo11:
@tfoote thanks for the info, once a have the opportunity we will correct this issue. Regards

PR #10185, opened 2016-01-02
Update base.yaml

niyawe:
libestools-dev is a virtual package. Since the package cannot be found in [1] and in [2] it looks like the script parses a file, where the virtual package is missing I changed the name to libestools2.1-dev

[1] http://jenkins.ros.org/job/ros-indigo-robot-face_binarydeb_trusty_amd64/2/console
[2] https://github.com/smd-ros-devel/catkin-rpms/blob/master/scripts/assert_package_dependencies_present.py

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1252,7 +1252,13 @@ libespeak-dev:
   ubuntu: [libespeak-dev]
 libestools-dev:
   fedora: [festival-speechtools-devel]
-  ubuntu: [libestools2.1-dev]
+  ubuntu:
+    saucy: [libestools2.1-dev]
+    trusty: [libestools2.1-dev]
+    utopic: [libestools2.1-dev]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
In some Ubuntu distributions the package including the version number is the virtual package. In some it is the other way around. Therefore I don't think that this patch is the right solution.

niyawe:
Ok, what would be the right solution? Different package names for each Ubuntu version?

tfoote:
It is a virtual package in Precise, Trusty, and Vivid, but a real package after that. http://packages.ubuntu.com/search?keywords=libestools-dev

Unfortunately rosdep and our checkers are not able to track virtual packages correctly. You'll need to give the concrete implementation packages for each version.

niyawe:
done

tfoote:
Thanks for the update.

PR #10189, opened 2016-01-03
removing moveit_simple_actions as it is renamed to romeo_moveit_actions

nlyubova:
hello,
could you please accept the removal of moveit_simple_actions ?


dirk-thomas:
@tfoote can you please remove the sourcedebs from shadow-fixed.

PR #10191, opened 2016-01-04
adding pyros_test

asmodehn:
A package with some nodes mostly useful for testing.


dirk-thomas:
I was just curious what the package is doing / providing. From reading the code or looks like this package is missing a dependency on `pyros_setup`?

asmodehn:
Oh... I ll double check tomorrow, but it should be the other way around :
pyros_setup depending on pyros_test for tests only.

pyros_test will just have a bunch of nodes useful for testing complex ros
systems ( echo node, pipe node, etc. ) Whatever is generic and useful
enough.

pyros_setup will be tricky to get right though...

PR #10230, opened 2016-01-09
pyros_setup: 0.0.9-0 in 'indigo/distribution.yaml' [bloom]

asmodehn:
Increasing version of package(s) in repository `pyros_setup` to `0.0.9-0`:
- upstream repository: https://github.com/asmodehn/pyros-setup.git
- release repository: https://github.com/asmodehn/pyros-setup-release.git
- distro file: `indigo/distribution.yaml`
- bloom version: `0.5.20`
- previous version for package: `0.0.8-0`

PR #10233, opened 2016-01-10
Revert "pyros_setup: 0.0.9-0 in 'indigo/distribution.yaml' [bloom]"

tfoote:
Reverts ros/rosdistro#10230

The sourcedeb is failing to build and it consequently keeps trying to rebuild.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -7475,7 +7475,7 @@ repositories:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/asmodehn/pyros-setup-release.git
-      version: 0.0.9-0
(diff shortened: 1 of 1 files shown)


asmodehn:
Yep, lets revert that. I thought it would have been easier than it is, to integrate dh-virtualenv. Not sure about the buildfarm setup, I ll try ros-buildbot (and maybe a prerelease) on a local server to see if I can catch something...

asmodehn:
FYI just installing dh-virtualenv package on my buildbot-ros server is enough to get the sourcedeb step to succeed. However it fails later on because of : https://github.com/spotify/dh-virtualenv/issues/127

PR #10257, opened 2016-01-11
update index.yaml to REP 143

dirk-thomas:
This needs an announcement on ros-users when being merged.

After this has been merged we can start using the new features in distribution files: http://www.ros.org/reps/rep-0143.html#distribution-file

This PR already enables PR jobs for the ROS core repos which have tests.

We should consider:
- enabling `test_pull_requests` for more ROS base repos
- disabled `test_commits` for repos which "devel" jobs will conceptionally never succeed on the farm (the ones which are currently just broken can still be disabled in the build files)

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -1,31 +1,19 @@
 %YAML 1.1
 # ROS index file
-# see REP 141: http://ros.org/reps/rep-0141.html
+# see REP 143: http://ros.org/reps/rep-0143.html
 ---
 distributions:
   groovy:
-    distribution: groovy/distribution.yaml
+    distribution: [groovy/distribution.yaml]
     distribution_cache: http://ros.org/rosdistro/groovy-cache.yaml.gz
-    doc_builds: [groovy/doc-build.yaml]
-    release_builds: [groovy/release-build.yaml]
-    source_builds: [groovy/source-build.yaml]
   hydro:
-    distribution: hydro/distribution.yaml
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Draft of announcement to ros-users (please feel free to edit this):

--

The new build farm [1] will be rolled out soon and in order to benefit from the new options specified in REP 143 [2] the rosdistro files have been upgraded to the new "format 3". If you have kept your system up-to-date there should be nothing to do.

But if you run into the following error:

    Unable to handle 'index' format version '3', please update rosdistro"

when running tools like "rosdep" or "bloom" you need to update your system. A more detailed description as well as the exact steps to update your system are the same as for the previous transition to REP 141. They can be found here: http://answers.ros.org/question/122266/how-can-i-fix-unable-to-handle-index-format-version-2-please-update-rosdistro/

If you run into any issues please comment on the GitHub PR which converted the rosdistro files [3].

Cheers,
- Dirk

[1] Docker-based build farm: http://wiki.ros.org/buildfarm
[2] REP 143: http://www.ros.org/reps/rep-0143.html
[3] Pull request switching to format 3: https://github.com/ros/rosdistro/pull/10257

wjwwood:
lgtm

tfoote:
+1

dirk-thomas:
I updated the draft of the announcement with more info about PR hold time and the planned work on Monday.

I will merge this and send the email tomorrow morning.

PR #10277, opened 2016-01-13
[rosdep/base] Add libgl1-mesa-glx-lts for Ubuntus.

130s:


tfoote:
I'm a little worried about required dependencies on this as some of these packages have been known to break X server. http://wiki.ros.org/Installation/Ubuntu
http://answers.ros.org/question/203610/ubuntu-14042-unmet-dependencies-similar-for-14043/
http://answers.ros.org/question/203610/ubuntu-14042-unmet-dependencies-similar-for-14043/

130s:
I see, that conflict issue looks eerie. I take this back (and my usecase was, as you already found, https://github.com/ros-drivers/leap_motion/issues/26 that's closed already).

PR #10286, opened 2016-01-14
keep build files

dirk-thomas:
I am not sure if this fixes the builds on the old farm (http://jenkins.ros.org/job/devel-jade-rospack/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/7/console) and/or if it is a problem for the new farm.


dirk-thomas:
A partial revert is not an option.

PR #10289, opened 2016-01-15

tn0432:
I 'd like ardrone2control to be indexed and documented on ros.org.

### DIFF
--- a/hydro/distribution.yaml
+++ b/hydro/distribution.yaml
@@ -258,6 +258,16 @@ repositories:
       url: https://github.com/vanadiumlabs/arbotix_ros-release.git
       version: 0.9.2-0
     status: maintained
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Hydro is EOL for a while. We don't run a build farm for it anymore. Please consider adding your package to a current distro (Indigo, Jade).

tn0432:
Hello!

I have my package in Fuerte and Hydro. It will not be a problem to make it in Indigo or Jade. But how do I publish with different distros? Thanks

dirk-thomas:
Sorry, but I don't understand your question.

The distribution files for Indigo and Jade are in a sibling folder. Beside that it is the same process as you have already done it in this PR.

PR #10317, opened 2016-01-19
move rosdistro distribution_caches to the new farm

tfoote:
Indigo and Jade are being generated in that location. 

I copied Groovy and Hydro into that folder for consistency. 

@dirk-thomas for review


dirk-thomas:
lgtm

PR #10318, opened 2016-01-20

JGMonroy:
I'd like mapir-ros-pkg to be indexed and documented on ros-org


JGMonroy:
Ups. Fixing problem and creating new pull request

PR #10319, opened 2016-01-20

JGMonroy:
I'd like mapir-ros-pkgs to be indexed and documented on ros.org


JGMonroy:
Ok, got it, the version should be master. I should have readed http://www.ros.org/reps/rep-0141.html#distribution-file more carefully.

PR #10320, opened 2016-01-20

JGMonroy:
I'd like mapir-ros-pkgs to be indexed and documented on ros.org


dirk-thomas:
We will wait with the merge untl the new buildfarm rollout is finished in about a day.

PR #10321, opened 2016-01-20
Changing branches

AndyZe:


dirk-thomas:
We will wait with the merge untl the new buildfarm rollout is finished in about a day.

PR #10328, opened 2016-01-21
remove not existing repos from Indigo

dirk-thomas:
The following repositories are been removed since they don't exist anymore:
- demo_lidar @jizhang-cmu
- demo_rgbd @jizhang-cmu


dirk-thomas:
@tfoote Please review and merge

PR #10360, opened 2016-01-27
add procps for OSX

k-okada:
PLEASE DO NOT MERGE YET

I do not have OSX computer and this is not tested. Need someone say OK
see discussion https://github.com/ros-perception/vision_opencv/issues/106#issuecomment-175352204


trainman419:
This is no longer necessary now that https://github.com/ros-perception/vision_opencv/pull/109 is merged.

PR #10362, opened 2016-01-27
Install header files for libturbojpeg on Ubuntu

cberner:


tfoote:
+1 that looks reasonable since the fedora rule already pulls in the header. The other option would be to create a second rule with the -dev name.

cberner:
Either sounds good to me. Although personally, I think it makes more sense to do it this way so that Fedora and Ubuntu are consistent. Let me know if you'd like me to switch it though

PR #10363, opened 2016-01-27
waypoints_navigation package deleted

DaikiMaekawa:
[Review on Reviewable](https://reviewable.io/reviews/ros/rosdistro/10363)


DaikiMaekawa:
waypoints_navigation should be deleted because newly one has been released instead of this package.

PR #10396, opened 2016-02-02
Add libhaloc doc to jade

plnegre:


dirk-thomas:
> Repository 'libhaloc' with url 'https://github.com/srv/libhaloc' should end with `.git` but does not.

plnegre:
Done!

PR #10404, opened 2016-02-03
Adding m_explore to documentation index for jade

hrnr:
Hi,

I'm developing new packages for multi robot exploring. I'd like to document them on ros wiki.

Cheers,
Jiri

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -1982,6 +1982,12 @@ repositories:
       url: https://github.com/WPI-RAIL/m4atx_battery_monitor.git
       version: develop
     status: maintained
+  m_explore:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The package names `explore` and `map_merge` are not very specific. Especially since are already packages in Indigo with the names `explorer` and `map_merger`.

tfoote:
Please see the naming conventions here: http://wiki.ros.org/ROS/Patterns/Conventions

hrnr:
Thanks for feedback.

I have renamed the packages to `explore_lite` (to emphasize it's aiming to be very lightweight, it does not create its own costmap etc.) and `multirobot_map_merge`. Is it enough to prevent the conflict?

PR #10439, opened 2016-02-07
[rosdep/base] Add cwiid and cwiid-dev

mdhorn:
Added C language version of the Wiimote library.
Both standalone library for running and development
flavor for building from source.

Needed for upstreaming new C implementation of ROS Wiimote node.


tfoote:
Thanks for filling in all the arches.

PR #10444, opened 2016-02-08
fix distribution version

dirk-thomas:
Incomplete upgrade in #10309.

The distribution files already use the new attributes like `test_pull_requests`.


tfoote:
+1 we need to update the checker to use rep 143 for this and #10443 to pass CI

dirk-thomas:
All green. Any reason to not merge this?

tfoote:
looks good now that ci is passing

PR #10464, opened 2016-02-10
adding pyros-setup as pip dependency

asmodehn:


rosdep/python.yaml
```diff
@@ -42,6 +42,10 @@ paramiko:
     trusty: [python-paramiko]
     utopic: [python-paramiko]
     vivid: [python-paramiko]
+pyros-setup:
```
dirk-thomas: I think it would be good to add a `-pip` suffix to the name (similar to there pip packages). That makes it clear that it can't be used for packages which should be packaged into Debian.

asmodehn: good point.

dirk-thomas:
+1 from me.

Waiting for one more vote for rosdep rules...

PR #10465, opened 2016-02-10
removing pyros_setup

asmodehn:
Trying to get the python package directly from pip via rosdep, without the need to catkin build it.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -7542,13 +7542,6 @@ repositories:
       url: https://github.com/clearpathrobotics/puma_motor_driver.git
       version: master
     status: maintained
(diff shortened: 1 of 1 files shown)


dirk-thomas:
You are aware that it is not usable in other packages which should be packaged into Debian? We also don't recommend to remove packages from a stable distribution since users might rely in it.

asmodehn:
Since I can control If i rely on the pip or on the deb via rosdep key (-pip suffix), lets keep it that way for now.

I am however planning to eventually move to a pip package as with normal python. I am aware about the problem regarding deb package depending on it, for which I am currently investigating solutions (dh-virtualenv)

PR #10466, opened 2016-02-10
add sound-theme-freedesktop and workrave for sound assets

k-okada:


dirk-thomas:
+1 from me.

Waiting for one more vote for rosdep rules...

PR #10482, opened 2016-02-11
rosdep: Add curl runtime packages

### LINKED ISSUE #10467: rosdep: Inconsistent resolve rules for `curl`
[The current rosdep resolve rules](https://github.com/ros/rosdistro/blob/953a366b2d307024b8c22539f23aba03770d9a53/rosdep/base.yaml#L494) for `curl` is inconsistent among different operating systems:

```yaml
curl:
  arch: [curl]
  debian: [libcurl4-openssl-dev]
  fedora: [libcurl-devel]
  freebsd: [curl]
  gentoo:
    portage:
      packages: [net-misc/curl]
  macports: [curl]
  opensuse: [libcurl-devel]
  rhel: [libcurl-devel]
  ubuntu: [libcurl4-openssl-dev]
```

For _Ubuntu_ and _Debian_ for example, only development libraries are listed as resolve rules. Those [does not include the `curl` binary](http://packages.ubuntu.com/trusty/amd64/libcurl4-openssl-dev/filelist). For _Gentoo_ and _Arch_ on the other hand, the resolved packages include both development files and the binary ([+](https://www.archlinux.org/packages/core/x86_64/curl/), [+](http://www.portagefilelist.de/site/query/listPackageFiles/?category=net-misc&package=curl&version=7.47.1&do#result)).

I can think of two ways to solve this issue. We can either add missing resolve rules for the binary (e.g `curl` for _Ubuntu_ and _Debian_) or break down this key to `curl` and `curl-dev` and list appropriate resolve rules for each platform. The latter may cause regressions for packages that are already dependant on this key.

mani-monaj:
Resolves #10467

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -493,16 +493,16 @@ crypto++:
   ubuntu: [libcrypto++-dev]
 curl:
   arch: [curl]
-  debian: [libcurl4-openssl-dev]
-  fedora: [libcurl-devel]
+  debian: [libcurl4-openssl-dev, curl]
+  fedora: [libcurl-devel, curl]
   freebsd: [curl]
   gentoo:
     portage:
       packages: [net-misc/curl]
   macports: [curl]
-  opensuse: [libcurl-devel]
-  rhel: [libcurl-devel]
-  ubuntu: [libcurl4-openssl-dev]
+  opensuse: [libcurl-devel, curl]
+  rhel: [libcurl-devel, curl]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Multiple list item must be separated with a colon.

tfoote:
Where by colon he means comma. It's a yaml list.

dirk-thomas:
Yeah that one :wink:

mani-monaj:
My bad! I fixed the issue and re-based/squashed the changes.

PR #10496, opened 2016-02-15
add royale library (pmd driver)

mikeferguson:


tfoote:
I can't find this package. http://packages.ubuntu.com/trusty/libroyale

mikeferguson:
It's not currently public -- I'm working towards getting the OKs for it to go public, but in the meantime, I'm building on a private farm and need the rosdep key for blooming the ROS driver on top of it.

tfoote:
Until it's public I'd suggest that you add a custom rosdep source with this entry and others hosted in the same repo to the machines where you're planning to use this dependency. (when installing from source or running bloom)

mikeferguson:
yeah -- except that means also hacking the buildfarm as well. I guess we'll continue to fork more of ROS...

wjwwood:
@mikeferguson out of curiosity which part needs to be hacked to take a custom rosdep rule. This is a pattern that should be made easier if it is currently hard. Is that the buildbot build farm or the osrf build farm?

mikeferguson:
Private buildfarm (buildbot-ros). For release, it does not really matter as the information becomes part of the gbp as long as we use custom rosdep keys on the machine we release on -- but for a test/pr builds of the actual ROS drivers, we need to be able to look up this package when it is part of the package.xml.

Currently, while our buildbots use various (private) rosdistros to configure what they build, the builds themselves rely on the public rosdistro for rosdep resolution, so that we don't have to bake custom rosdep sources into the /etc/ros/rosdep files (this is actually one area I consider to be a pain point, for instance, even if I have a custom ROSDISTRO_INDEX_URL, it does nothing to change my rosdep sources, those have to be updated manually as far as I can tell).

(I should note, there is a bit of a hack in buildbot-ros that maybe we have relied on a bit too much, in that if a rosdep key fails to be resolved, we attempt to install with "ros-distro-KEY" as this allows us to still build our packages that are unlisted in the public rosdistro)

tfoote:
For a private buildfarm that wants to have rosdep rules inject I think that the best place to do this would be to add an entry to the target _config entries in the build file: https://github.com/ros/rosdistro/blob/master/indigo/release-build.yaml#L15 
http://www.ros.org/reps/rep-0141.html

That's where we already allow injecting mirrors which could also be extended to PPAs. And with that could come the appropriate rosdep rules since they basically should be associated with the PPAs.

PR #10515, opened 2016-02-18
Moved wpi-rail to gt-rail

dekent:


vrabaud:
Hydro is eol-ed, please do not change those files.

PR #10553, opened 2016-02-23
Add pull request building for common_msgs and geometry*

tfoote:


dirk-thomas:
I would recommend to first get the Jade devel job of `geometry_experimental` (http://build.ros.org/view/Jdev/job/Jdev__geometry_experimental__ubuntu_trusty_amd64/) green before enabling the PR check for it

PR #10554, opened 2016-02-23
urdfdom_py: 0.3.1-1 in 'indigo/distribution.yaml' [bloom]

jackieokay:
Increasing version of package(s) in repository `urdfdom_py` to `0.3.1-1`:
- upstream repository: https://github.com/ros/urdf_parser_py/
- release repository: https://github.com/ros-gbp/urdfdom_py-release.git
- bloom version: `0.5.20`
- previous version for package: `0.3.0-2`
## urdfdom_py

```
* Add travis
* Add package.xml for ROS release
* Tweak CMakeLists to reflect migration from urdfdom
* Contributors: Jackie Kay
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -11191,11 +11191,19 @@ repositories:
       version: master
     status: maintained
   urdfdom_py:
+    doc:
+      type: git
+      url: https://github.com/ros/urdf_parser_py.git
+      version: 0.3.1
     release:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/ros-gbp/urdfdom_py-release.git
-      version: 0.3.0-2
+      version: 0.3.1-1
+    source:
+      type: git
+      url: https://github.com/ros/urdf_parser_py.git
+      version: indigo-devel
     status: maintained
   urg_c:
     doc:


jacquelinekay:
the first time I ran bloom for this release it was interrupted because I made a typo while copy-pasting a URL into the prompt to link to a source repo... so I'm not sure this is going to work

indigo/distribution.yaml
```diff
@@ -11191,11 +11191,19 @@ repositories:
...
     release:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/ros-gbp/urdfdom_py-release.git
-      version: 0.3.0-2
+      version: 0.3.1-1
+    source:
+      type: git
+      url: https://github.com/ros/urdf_parser_py.git
+      version: 0.3.1
```
dirk-thomas: Both version entries should be branch names.

jacquelinekay: ok, a tag name is not ok?

Do you suggest I change this manually or re-run bloom somehow?

wjwwood:
@jacquelinekay you can always `ctrl-c` bloom and start over. Until you say yes to the push question nothing is changed on GitHub.

indigo/distribution.yaml
dirk-thomas: A tag does not make sense for a devel job. For the doc entry that would work but then you would need to update it with every release.

You can edit the PR in place.

wjwwood: A tag doesn't make sense for the source entry... It might make sense for the doc.

You can just close this pr and re-run bloom with the same arguments as before but with `--pull-request-only` added to them.

wjwwood: Both options work, editing this in place is ok too.

jacquelinekay: a45ad8c thanks for the advice! I'll squash this after Travis finishes

PR #10579, opened 2016-02-25
Added first of my packages, common-sensors, for pre-release test

JenniferBuehler:
I want to perform a pre-release test before my first official release.

There is a number of packages which I want to add, but because this is the first time I am going through this process, I wanted to try with a simple package first and will then add the others after I know the process.

If I understand correctly, for a pre-release test only the "source" information is required.

Thanks & Kind regards
Jennifer

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1410,6 +1410,16 @@ repositories:
       url: https://github.com/ros/common_msgs.git
       version: indigo-devel
     status: maintained
+  common_sensors:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


JenniferBuehler:
This is superseded by my next pull request which is the result of bloom-release. Followed instructions [here](http://wiki.ros.org/bloom/Tutorials/PrereleaseTest?distro=indigo) to wait for the right place to stop, but thought it needs to first generate the PR for distribution.yaml otherwise the website wouldn't find my package... and then the whole process ended up finishing.

I guess next time I'll have to manually do the PR on distribution.yaml without using bloom-release to generate it.

PR #10580, opened 2016-02-25
common_sensors: 0.1.0-0 in 'indigo/distribution.yaml' [bloom]

JenniferBuehler:
Increasing version of package(s) in repository `common_sensors` to `0.1.0-0`:
- upstream repository: https://github.com/JenniferBuehler/common-sensors.git
- release repository: https://github.com/JenniferBuehler/common-sensors-release.git
- bloom version: `0.5.20`
## common_sensors

```
* Added sources to repository
* Contributors: Jennifer Buehler
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1411,6 +1411,21 @@ repositories:
       url: https://github.com/ros/common_msgs.git
       version: indigo-devel
     status: maintained
+  common_sensors:
+    doc:
+      type: git
+      url: https://github.com/JenniferBuehler/common-sensors.git
+      version: master
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/JenniferBuehler/common-sensors-release.git
+      version: 0.1.0-0
+    source:
+      type: git
+      url: https://github.com/JenniferBuehler/common-sensors.git
+      version: master
+    status: developed
   common_tutorials:
     doc:
       type: git


tfoote:
This package looks like a great resource and the cleanup will be valuable to the community. However the name choice is too generic for our naming convention, falling into the catchall category. http://wiki.ros.org/ROS/Patterns/Conventions and http://www.ros.org/reps/rep-0144.html

For the meshes etc I would suggest that according to the REP each sensor should get a package <SENSOR_NAME>_description. And although it's a bit more of a pain, coupling these sensors into a single repository is probably not the best idea since the repository is the unit of release and consequently you're tying the releases of updates for any specific sensor to all the other sensors. 

And for the laser filters if they're generic and useful I'd suggest submitting them to be reviewed and added to the laser_filters package http://wiki.ros.org/laser_filters

JenniferBuehler:
Hi Tully,

thanks for the feedback! It's great to see that packages submitted to ROS get a proper review.

I agree about the naming. I was pondering about which name to choose for a while. Because this package essentially aims at collecting the most commonly used sensors, I stuck to "common sensors" in the end. But the problem with this over time is also: which sensor classifies as "common" to make it into such a package? So I've kept thinking and the only other idea I had so far is "sensors_collection".

As to the separation into several packages: My original aim of creating this package was that I didn't want to always end up having many external package dependencies (to urdf, gazebo, filters, etc.) just because of a couple of very common sensors I am using. I just wanted one package with all common sensors (which we had in the lab and are very common). I'm sure there are other people who would like that. However I agree that it would be nicer to have individual packages, e.g. hokuyo_description, xtion_description, etc. But it would still be nice if only *one* package would need installing, which could then be a meta package which includes all the commonly sensors. What do you think about that?

I see what you mean with the issue of having all of the packages in the same repository. On the other hand, I wouldn't expect updates on any of the packages to happen super frequently. And the packages are not that big, so does it still make sense to have a dedicated repository for just a couple of URDF files?

Yes, I'll see if it's possible to get the laser filters into the ROS package you mentioned. However because it's largely based on code from uni Texas, it should maybe be them to do it. I just wanted to add the code here for convenience (and for not needing yet another package dependency). 
I can email them and ask what they think about that.

Long story short: this package is not of high priority for my other packages which I'll be submitting in the next few days (from [my git repos](https://github.com/JenniferBuehler)). It's only an optional runtime dependency (when using certain launch files). I wanted to use this package as a "practise run" for submitting a package with bloom. So I can start working on all the changes you suggested over time and then re-submit this PR.

tfoote:
Hi Jennifer, 

We do our best to review package submissions. Naming is something that we have the most visibility into. 

If you'd like to create a metapackage to bring in all the resources you'd like to have for your project that would be fine. Just use an appropriate prefix so that others would understand the scope. 

As a test run, this looks like a success. Obviously we didn't try building it yet, but you made it all the way through the release process and submitted a well formed pull request. 

I'll close this now and you can open new pull request for your upcoming package releases.

PR #10592, opened 2016-03-01
Adding behavior_tree to documentation index for hydro and indigo 

miccol:
I would like behavior_tree to be indexed and documented on ros.org

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -616,7 +616,7 @@ repositories:
   behavior_tree:
     doc:
       type: git
-      url: https://github.com/miccol/ROS-Behavior-Tree
+      url: https://github.com/miccol/ROS-Behavior-Tree.git
       version: master
     source:
       type: git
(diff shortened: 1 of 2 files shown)


indigo/distribution.yaml
```diff
@@ -613,6 +613,16 @@ repositories:
       url: https://github.com/AutonomyLab/bebop_autonomy.git
       version: indigo-devel
     status: developed
+  behavior_tree:
+    doc:
+      type: git
+      url: https://github.com/miccol/ROS-Behavior-Tree
```
dirk-thomas: The URL must end with `.git`.

dirk-thomas:
`Hydro` has been EOLed for a long time already and we don't have a build farm for it anymore. Therefore we don't accept changes to the Hydro distribution file anymore. Please remove the hydro related changes from this PR.

miccol:
Hydro repo removed.

Indigo url fixed (sorry I though that the git extension was for the source only and not for the doc)

Michele

PR #10596, opened 2016-03-02
graspit_tools: 0.1.1-0 in 'indigo/distribution.yaml' [bloom]

JenniferBuehler:
Increasing version of package(s) in repository `graspit_tools` to `0.1.1-0`:
- upstream repository: https://github.com/JenniferBuehler/graspit-pkgs.git
- release repository: https://github.com/JenniferBuehler/graspit-pkgs-release.git
- bloom version: `0.5.20`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2798,7 +2798,7 @@ repositories:
   graspit_tools:
     doc:
       type: git
-      url: https://github.com/JenniferBuehler/graspit-pkgs.wiki.git
+      url: https://github.com/JenniferBuehler/graspit-pkgs.git
       version: master
     release:
       packages:
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2795,6 +2795,28 @@ repositories:
       url: https://github.com/mikeferguson/grasping_msgs-gbp.git
       version: 0.3.1-0
     status: developed
+  graspit_tools:
+    doc:
+      type: git
+      url: https://github.com/JenniferBuehler/graspit-pkgs.wiki.git
+      version: master
+    release:
+      packages:
+      - grasp_planning_graspit
+      - grasp_planning_graspit_msgs
+      - grasp_planning_graspit_ros
+      - graspit_tools
+      - jaco_graspit_sample
+      - urdf2graspit
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/JenniferBuehler/graspit-pkgs-release.git
+      version: 0.1.1-0
+    source:
+      type: git
+      url: https://github.com/JenniferBuehler/graspit-pkgs.git
+      version: master
+    status: developed
   grid_map:
     doc:
       type: git


indigo/distribution.yaml
```diff
@@ -2795,6 +2795,28 @@ repositories:
       url: https://github.com/mikeferguson/grasping_msgs-gbp.git
       version: 0.3.1-0
     status: developed
+  graspit_tools:
+    doc:
+      type: git
+      url: https://github.com/JenniferBuehler/graspit-pkgs.wiki.git
```
dirk-thomas: This repo does not contain ROS packages and can therefore not be indexed.

JenniferBuehler:
Hi,

I think I made a mistake with the uri for documentation generation, the correct link
should probably be the source repository: https://github.com/JenniferBuehler/graspit-pkgs.git, and not the wiki repository as it is now?

Please let me know if I should update the PR with this.

On another note, I would like to comment about the fact that several packages are involved here: I know it would be nice to have them all in their own repositories, but for 2 reasons I decided for this setup:

1. I have quite a lot of packages which I would like to add in the next days/weeks. Users who install from source already have to clone several repositories as it is for running my tutorials. If I put every single package in their own repository, it's going to be very inconvenient for users who install from source. It would also be more difficult for me to maintain.
2. I expect most of the packages in grapsit_tools (this PR) to be updated all at the same time, and I'm not planning to re-release at each little change of the sources, but only when major changes have been done, which then most likely includes changes in all packages of this PR.

Regards
Jennifer

JenniferBuehler:
I've fixed the documentation URL. Sorry about this mistake.

dirk-thomas:
Regarding the user experience: ROS provides tools like the `rosinstall_generator`, `wstool` and/or `vcstool` to easily manage distributed sources. With a few commands you should be able to clone hundreds of repos at once. I think this is an argument to split not closely related packages into separate repos without negatively affect the user experience.

Bundling a set of packages in a single repo is common practice and makes total sense if they are closely related to each other as it seems to be in this case. So all good with this PR.

wjwwood:
> I think I made a mistake with the uri for documentation generation, the correct link
should probably be the source repository: https://github.com/JenniferBuehler/graspit-pkgs.git, and not the wiki repository as it is now?

That's correct, this documentation entry is used by our farm to try and generate code documentation for your packages automatically as well as it generates the information in the wiki.ros.org page for your package.

Also, as @dirk-thomas said, you can provide a `.rosinstall` file or give the users a `rosinstall_generator` command which will generate a `.rosinstall` file for your packages. Then they can use `wstool` (or similar) to clone everything in one command. For an example of this workflow, have a look at the from source instructions:

http://wiki.ros.org/jade/Installation/Source#Installation

JenniferBuehler:
Thanks for the feedback and merging my PR. 
I have only worked peripherally with rosinstall and ws_tool a while back... thanks for the hint, I'll definitely look into this.

PR #10602, opened 2016-03-03
open_karto: 1.1.4-0 in 'indigo/distribution.yaml' [bloom]

mikeferguson:
Increasing version of package(s) in repository `open_karto` to `1.1.4-0`:
- upstream repository: https://github.com/ros-perception/open_karto.git
- release repository: https://github.com/ros-gbp/open_karto-release.git
- bloom version: `0.5.20`
- previous version for package: `1.1.3-0`
## open_karto

```
* update build status badges
* Adds LocalizedRangeScanWithPoints range scan
* Contributors: Michael Ferguson, Russell Toris
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -6446,7 +6446,7 @@ repositories:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/ros-gbp/open_karto-release.git
-      version: 1.1.3-0
+      version: 1.1.4-0
     source:
       type: git
       url: https://github.com/ros-perception/open_karto.git


mikeferguson:
@dirk-thomas hudson seems to be kinda upset with me: http://build.ros.org/job/Ibin_uT32__slam_karto__ubuntu_trusty_i386__binary/4/console -- I presume this is not caused by the release, but something wonky on the farm?

wjwwood:
Hmm, looks like the slave-agent is not working on that slave... I triggered again and it still failed: http://build.ros.org/job/Ibin_uT32__slam_karto__ubuntu_trusty_i386__binary/5/console

@tfoote any idea?

dirk-thomas:
@tfoote The same slave seems to also be stuck with two other jobs. Can you please disconnect the misbehaving slave so that the farm can continue to work. But also look into the problems on the "broken" slave before destroying it?

tfoote:
For the record that machine seems to be completely hung. I cannot log into it. I don't even get the challenge response for ssh. 

EC2 monitoring shows it at 95% CPU constantly but zero IO and nominal network. 

I'm destroying it since I cannot get in.

PR #10608, opened 2016-03-03
adding behavior_tree package

miccol:
Hello, 

I would like to add the behavior_tree package. 
Thank you
Michele

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -218,6 +218,16 @@ repositories:
       url: https://github.com/AutonomyLab/bebop_autonomy.git
       version: indigo-devel
     status: developed
+  behavior_tree:
+    doc:
+      type: git
+      url: https://github.com/miccol/ROS-Behavior-Tree.git
+      version: master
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Just an unrelated question: where are the build status badges in your README file coming from?

miccol:
I have manually copied from shield.io after testing them. Is there an automatic way to do that?

dirk-thomas:
If the code is being built on our Jenkins farm you can use the badges from there. Please see the wiki which already makes use of that: http://wiki.ros.org/roscpp (click on the sidebar on "Jenkins jobs")

tfoote:
There's an Embeddable Build Status Icon provided by our jenkins instance for each job. 

It provides documentation for how to embed it in different places such as: http://build.ros.org/view/Jsrc_uT/job/Jsrc_uT__roscpp__ubuntu_trusty__source/badge/

miccol:
Thank you! I will change that!

Michele

PR #10609, opened 2016-03-03
adding missing rules for Xenial and Wily

wjwwood:
This is probably necessary to get any where with releasing Kinetic packages. I discovered these while building Jade from source on Xenial.


wjwwood:
I have more rules, but they are not as straight forward since they deal with Qt, OpenCV, etc...

wjwwood:
@ros/owners can I get additional +1's and/or a merge?

dirk-thomas:
lgtm

PR #10621, opened 2016-03-04
add debian jessie to release platforms for kinetic

tfoote:
This will allow us to turn it on in the future.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3,6 +3,8 @@
 # see REP 143: http://ros.org/reps/rep-0143.html
 ---
 release_platforms:
+  debian:
+  - jessie
   fedora:
   - '23'
   - '24'


wjwwood:
lgtm

dirk-thomas:
This did not have any effect when I released `genmsg`. `bloom` still only created stuff for `wily`, `xenial`, `23` and `24`. I assume that is not sufficient?

wjwwood:
I talked with @tfoote off-line, and we need to change the default rules in `bloom`. Here are some links for reference:

- https://github.com/wjwwood/rviz-release/blob/master/tracks.yaml#L53-L54
- https://github.com/ros-infrastructure/bloom/blob/master/bloom/config.py#L194-L206
- https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/generator.py#L526-L527
- https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/generator.py#L534-L540

I'll help with merging and releasing bloom to get that out there asap.

Issue #9945, opened 2015-12-03
Kinetic Kick-off Checklist


esteve:
This is a checklist of things to do for the Kinetic kick-off (pre development and development):

- [x] Choose distribution name: Kinetic
- [x] commission a logo @tfoote
- [x] Add to REP3 with proposed platforms and minimum requirements @esteve https://github.com/ros-infrastructure/rep/pull/106
- [x] List on ROS wiki with expected release date (http://wiki.ros.org/Distributions)
- [x] EOL distribution N-3 (Hydro)
- [x] Add W+X to bootstrap repository on repos.ros.org  @tfoote W+X supported https://github.com/ros-infrastructure/reprepro-updater/commit/951812014a2f9afef53ffd90fa06a41af3fd8b4f
- [x] Add W+X to buildfarm_deployment_config @tfoote https://github.com/ros-infrastructure/buildfarm_deployment_config/commit/9817752851a4123294342ec0af3a134376940d0c (note that one is not the live one it's for reference only)
- [x] Add W+X to ros_buildfarm/common.py ros-infrastructure/ros_buildfarm#223

----

- [x] Import third party packages
  - [x] `colladadom` (@wjwwood pinged Rosen on Feb. 29th)

----

- [x] Add newly supported Ubuntu platforms to list of platforms for each Python package (in `stdeb.cfg`), in order of dependency (for ROS-L, do this as soon as debs for new Ubuntu are available, in pre-alpha!):
  - [x] rospkg (uploaded, no need to rerelease): ros-infrastructure/rospkg#96
  - [x] catkin_pkg (uploaded, no need to rerelease): ros-infrastructure/catkin_pkg#135
  - [x] rosdistro (uploaded, no need to rerelease): https://github.com/ros-infrastructure/rosdistro/pull/61
  - [x] rosdep (uploaded, no need to rerelease): ros-infrastructure/rosdep#434
  - [x] vcstools (uploaded, no need to rerelease): https://github.com/vcstools/vcstools/pull/111
  - [x] wstool (uploaded, no need to rerelease): https://github.com/vcstools/wstool/pull/84
  - [x] rosinstall (uploaded, no need to rerelease): https://github.com/vcstools/rosinstall/pull/101
  - [x] bloom (uploaded, no need to rerelease): ros-infrastructure/bloom#379
  - [x] rosinstall_generator (uploaded, no need to rerelease): ros-infrastructure/rosinstall_generator#32
[...]

PR #10673, opened 2016-03-10
update collada-dom for xenial

k-okada:
see https://github.com/ros/rosdistro/issues/9945#issuecomment-194642340 for discussion

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -475,7 +475,20 @@ collada-dom:
     portage:
       packages: [media-libs/collada-dom]
   macports: [collada-dom]
-  ubuntu: [collada-dom-dev]
+  ubuntu:
+    lucid: [collada-dom-dev]
+    maverick: [collada-dom-dev]
+    natty: [collada-dom-dev]
+    oneiric: [collada-dom-dev]
+    precise: [collada-dom-dev]
+    quantal: [collada-dom-dev]
+    raring: [collada-dom-dev]
+    saucy: [collada-dom-dev]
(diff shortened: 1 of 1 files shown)


tfoote:
I see libcollada-dom2.4-dep-dev but not libcollada-dom-dev

http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=collada-dom&searchon=names

wjwwood:
I agree with @tfoote this will not work as is.

k-okada:
anyway, I updated the PR with libcollada-dom2.4-dep-dev

wjwwood:
Oh, well I can install that too. I guess the virtual packages don't show up on packages.ubuntu.com?

I think we should change it back to `libcollada-dom-dev`. Sorry for all the back and forth @k-okada.

k-okada:
ok, done

--
◉ Kei Okada

wjwwood:
+1, just to confirm I removed all collada packages and PPA's from my xenial VM and was able to install `libcollada-dom-dev` still.

tfoote:
Sorry unfortunately rosdep cannot support virtual packages. apt-get will install it, but we cannot detect they are installed and rosdep will always try to install it.

wjwwood:
Ok, I have to change my mind again. After talking with @tfoote we can't support the virtual package because of the way we check that it is installed with `rosdep`.

k-okada:
ok, i added https://github.com/ros/rosdistro/pull/10700, please merge whichever you like ;-)

k-okada:
#10700 was merged

PR #10700, opened 2016-03-11
update collada-dom for xenial

k-okada:
another version of #10673


tfoote:
+1 

thanks for iterating

PR #10706, opened 2016-03-11
Added rosdep keys for libsvm3 and libsvm-dev

briancairl:
Package for SVM classifiers (c/cpp, python, matlab, etc.).


dirk-thomas:
+1

PR #10710, opened 2016-03-12
Adding kml_util to documentation index for distro

silviomaeta:
I'd like to have kml_util indexed and documented in ros.org site.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -4226,15 +4226,6 @@ repositories:
       type: git
       url: https://github.com/silviomaeta/kml_util.git
       version: indigo-devel
-    release:
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/silviomaeta/kml_util.git
-      version: 0.1.0
-    source:
-      type: git
-      url: https://github.com/silviomaeta/kml_util.git
-      version: indigo-devel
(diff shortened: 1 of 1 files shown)


silviomaeta:
I'd like to have plot_util indexed and documented in ros.org site.

silviomaeta:
I'd like to have kml_util and plot_util indexed and documented in ros.org site.

indigo/distribution.yaml
```diff
@@ -4221,6 +4221,12 @@ repositories:
       url: https://github.com/kf/kingfisher.git
       version: indigo-devel
     status: maintained
+  kml_util:
+    doc:
+      type: git
+      url: https://github.com/silviomaeta/kml_util.git
+      version: indigo-devel
```
dirk-thomas: There is no such branch name in your repo. This should probably be `master`?

Same below.

silviomaeta: Yes.  You are right.   It is master.  I will fix it and submit the request again. 
Thanks,
Silvio

dirk-thomas: You can also edit the files on the branch of this PR.

dirk-thomas:
After updating the PR you might want to comment in the future. Otherwise nobody is being notified.

PR #10730, opened 2016-03-15
add rosdep key for libgif-dev

mikeferguson:


dirk-thomas:
+1

mikeferguson:
Rebased on top of master (I didn't update my fork initially)

tfoote:
Could you add debian [`libgif-dev`](https://packages.debian.org/sid/libgif-dev) and fedora [`libgif-devel`](https://www.rpmfind.net/linux/rpm2html/search.php?query=giflib-devel) too?

mikeferguson:
Added

tfoote:
We turned on GitHub protection for this branch and it will warn if there are any commits out of date, you probably were well out of date if you hadn't updated but now It's likely still mergeable, we just have to use our admin status to override the warning. Since it's unreasonable to expect people to rebase every 5 minutes.

PR #10737, opened 2016-03-17
update libvtk-java for xenial and wily

wjwwood:
I collapsed it because it's been the same since lucid and we don't support anything older.

See: http://packages.ubuntu.com/xenial/libvtk-java


tfoote:
+1

PR #10738, opened 2016-03-17
libpcl and libpcl-dev for wily and xenial

wjwwood:
See: http://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=libpcl


tfoote:
+1

PR #10740, opened 2016-03-17
python-vtk for wily and xenial

wjwwood:
I flattened the rule since it has been the same since lucid and we don't support anything older.


wjwwood:
http://packages.ubuntu.com/xenial/python-vtk

tfoote:
+1

wjwwood:
Ran into a snag with this one: https://github.com/ros-perception/perception_pcl/issues/44#issuecomment-197653379

PR #10749, opened 2016-03-17
add OpenCV3

vrabaud:


wjwwood:
@vrabaud it would be best if you could pull from an upstream tag or something rather than this:

https://github.com/ros-gbp/opencv3-release/blob/master/tracks.yaml#L68

It's not particularly traceable when the trail of where the code came from ends with a directory on your computer. If you give me the appropriate upstream version I can figure out how to do a normal release with that. (I know it's slower that way, but it is better for transparency).

Thanks.

vrabaud:
Right, that's why it's explained on the wiki: http://wiki.ros.org/opencv3#Building_the_Package
You can also fix https://github.com/ros-infrastructure/bloom/issues/259 . Right now, I would have to pull the full 300M OpenCV repo and that's very slow :(

wjwwood:
Well if you don't mind, I'd like to rebloom it. After all very slow != impossible. I'll just go get a cookie while it's running :cookie:.

Unfortunately I don't have time to look into ros-infrastructure/bloom#259 at the moment since it isn't simply add `--depth 1` to all `git clone` calls (I'd need to check on versions and what not and also test it).

PR #10759, opened 2016-03-18
adding python-anyjson

asmodehn:


rosdep/python.yaml
```diff
@@ -107,6 +107,12 @@ python-amqp:
     vivid: [python-amqp]
     wily: [python-amqp]
     xenial: [python-amqp]
+python-anyjson:
+  ubuntu:
+    trusty: [python-anyjson]
+    vivid: [python-anyjson]
+    wily: [python-anyjson]
+    xenial: [python-anyjson]
```
dirk-thomas: This package seems to be available even in older Ubuntu distro. Therefore I think it makes sense to just list it as:

    ubuntu: [python-anyjson]

wjwwood: +1

tfoote:
This also appears to be in debian: https://packages.debian.org/source/wheezy/python-anyjson and fedora https://admin.fedoraproject.org/pkgdb/package/rpms/python-anyjson/ generically. Could you add those rules too. 

And for the record the ubuntu packages: http://packages.ubuntu.com/trusty/python/python-anyjson

asmodehn:
like that ?

PR #10812, opened 2016-03-20
python-urlgrabber for u-x

k-okada:
http://old-releases.ubuntu.com/ubuntu/pool/main/u/urlgrabber/ for utopic?

http://packages.ubuntu.com/search?keywords=python-usb

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2154,17 +2154,9 @@ python-ujson:
     pip:
       packages: [ujson]
 python-urlgrabber:
+  debian: [python-urlgrabber]
   fedora: [python-urlgrabber]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
+1

tfoote:
The update looks good. However looking at this entry all rules are the same. Could you change this to be a single rule for ubuntu to be more compact since it does not look like there will be a package version change. 

Could you also add a debian rule too, it looks to have the same key: https://packages.debian.org/sid/python-urlgrabber 

Ubuntu package info: http://packages.ubuntu.com/trusty/python-urlgrabber

k-okada:
thanks for comments, updated the PR.

PR #10823, opened 2016-03-21
Adding pid & lyap_control pkgs for kinetic.

AndyZe:


kinetic/distribution.yaml
```diff
@@ -273,6 +273,21 @@ repositories:
...
       version: hydro-devel
     status: maintained
+  lyap_control:
+    doc:
+      type: git
+      url: https://bitbucket.org/AndyZe/lyap_control.git
+      version: master
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
```
dirk-thomas: You can't just use the Indigo release. You need to run bloom for the Kinetic for your repo.

dirk-thomas:
Yoy cannot just rename the release tags. You **need** to run a bloom release for Kinetic.

AndyZelenak:
@dirk-thomas OK. I interpreted from the email to ros-users that rosdistro was the first step:
"Once your package has been added to Kinetic in rosdistro, you can generate
a prerelease command on prerelease.ros.org."

dirk-thomas:
If you want to adda release entry to do a prerelease it should not have a version field.

wjwwood:
I believe you just need the source entry so you can run a prerelease (using the devel option). You can omit the release entry until your first release I think.

PR #10829, opened 2016-03-22
Add script to check for unreleased dependencies

jackieokay:
This is a first pass at a script suggested by @dirk-thomas that invokes the rosdistro Python API to list all dependencies for an unreleased repo that a) were dependencies of the package in the previous distribution and b) have not been released into the current distribution.

Input: A list of repository names to check for unreleased dependencies. If no input is given, the default is to look at all repositories that were released into the previous distribution.
Output: A dictionary keyed on repository names where the values are a list of names of unreleased dependencies.

This feature is particularly useful for new maintainers who might not be familiar with the dependency tree of their packages. I have already used it to check for dependencies that are blocking robot_model.

The default distribution is the latest (Kinetic).

to do:
- [x] Return repositories that are blocking the requested repositories rather than packages.
- [x] Check to see which packages the unreleased inputs are blocking (sort of done; I opted to simply list which packages are blocking other packages for now)
- [x] Make the data available to users in a nicer way. Consider rendering a tree of packages that are released, unreleased, and available to be released

### DIFF
--- a/scripts/check_blocking_repos.py
+++ b/scripts/check_blocking_repos.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python
+
+import argparse
+
+import rosdistro
+from rosdistro.dependency_walker import DependencyWalker
+
+
+def is_released(repo, dist_file):
+    return repo in dist_file.repositories and \
+        dist_file.repositories[repo].release_repository is not None and \
+        dist_file.repositories[repo].release_repository.version is not None
+
+
+parser = argparse.ArgumentParser(
+    description='Get unreleased repos and their dependencies.',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument(
+    '--rosdistro', metavar='ROS_DISTRO',
+    help='The ROS distribution to check packages for')
+
+# If not specified, check for all repositories in the previous distribution
+parser.add_argument(
+    '--repositories',
+    metavar='REPOSITORY_NAME', nargs='*',
+    help='Unreleased repositories to check dependencies for')
+
+parser.add_argument(
+    '--depth',
+    metavar='depth', type=int,
+    help='Maxmium depth to crawl the dependency tree')
+
+args = parser.parse_args()
+
+distro_key = args.rosdistro
+repo_names = args.repositories
+prev_distro_key = None
+
+index = rosdistro.get_index(rosdistro.get_index_url())
+valid_distro_keys = index.distributions.keys()
+valid_distro_keys.sort()
+if distro_key is None:
+    distro_key = valid_distro_keys[-1]
+
+# Find the previous distribution to the current one
+try:
+    i = valid_distro_keys.index(distro_key)
+except ValueError:
+    print('Distribution key not found in list of valid distributions.')
+    exit(-1)
+if i == 0:
+    print('No previous distribution found.')
+    exit(-1)
+prev_distro_key = valid_distro_keys[i - 1]
+
+cache = rosdistro.get_distribution_cache(index, distro_key)
+distro_file = cache.distribution_file
+
+prev_cache = rosdistro.get_distribution_cache(index, prev_distro_key)
+prev_distribution = rosdistro.get_cached_distribution(
+    index, prev_distro_key, cache=prev_cache)
+
+prev_distro_file = prev_cache.distribution_file
+
+dependency_walker = DependencyWalker(prev_distribution)
+
+if repo_names is None:
+    # Check missing dependencies for packages that were in the previous
+    # distribution that have not yet been released in the current distribution
+    # Filter repos without a version or a release repository
+    keys = prev_distro_file.repositories.keys()
+    prev_repo_names = set(
+        repo for repo in keys if is_released(repo, prev_distro_file))
+else:
+    prev_repo_names = set(
+        repo for repo in repo_names if is_released(repo, prev_distro_file))
+
+keys = distro_file.repositories.keys()
+current_repo_names = set(
+    repo for repo in keys if is_released(repo, distro_file))
+
+# Print the repositories that will be eliminated from the input
+eliminated_repositories = prev_repo_names.intersection(
+    current_repo_names)
+if len(eliminated_repositories) > 0:
+    print('Ignoring inputs which have already been released:')
+    print('\n'.join(
+        sorted('\t{0}'.format(repo) for repo in eliminated_repositories)))
+
+repo_names_set = prev_repo_names.difference(
+    current_repo_names)
+
+if len(repo_names_set) == 0:
+    if repo_names is None:
+        print('Everything in {0} was released into the next {1}!'.format(
+            prev_distro_key, distro_key))
+        print('This was probably a bug.')
+    else:
+        print('All inputs are invalid or were already released in {0}.'.format(
+            distro_key))
+    print('Exiting without checking any dependencies.')
+    exit(0)
+
+repo_names = list(repo_names_set)
+
+
+# Get a list of currently released packages
+current_package_names = set(
+    pkg for repo in current_repo_names
+    for pkg in distro_file.repositories[repo].release_repository.package_names)
+
+# Construct a dictionary where keys are repository names and values are a list
+# of the missing dependencies for that repository
+
+blocked_repos = {}
+unblocked_repos = set()
+total_blocking_repos = set()
+
+for repository_name in repo_names:
+    repo = prev_distro_file.repositories[repository_name]
+    release_repo = repo.release_repository
+    package_dependencies = set()
+    packages = release_repo.package_names
+    # Accumulate all dependencies for those packages
+    for package in packages:
+        recursive_dependencies = dependency_walker.get_recursive_depends(
+            package, ['build', 'run', 'buildtool'], ros_packages_only=True,
+            limit_depth=args.depth)
+        package_dependencies = package_dependencies.union(
+            recursive_dependencies)
+
+    # For all package dependencies, check if they are released yet
+    unreleased_pkgs = package_dependencies.difference(
+        current_package_names)
+    # remove the packages which this repo provides.
+    unreleased_pkgs = unreleased_pkgs.difference(packages)
+    # Now get the repositories for these packages.
+    blocking_repos = set(prev_distro_file.release_packages[pkg].repository_name
+                         for pkg in unreleased_pkgs)
+    if len(blocking_repos) == 0:
+        unblocked_repos.add(repository_name)
+    else:
+        # Get the repository for the unreleased packages
+        blocked_repos[repository_name] = blocking_repos
+        total_blocking_repos |= blocking_repos
+
+unblocked_blocking_repos = total_blocking_repos.intersection(unblocked_repos)
+unblocked_leaf_repos = unblocked_repos.difference(unblocked_blocking_repos)
+
+# Double-check repositories that we think are leaf repos
+for repo in unblocked_leaf_repos:
+    # Check only one level of depends_on
+    depends_on = dependency_walker.get_depends_on(package, 'build') | \
+        dependency_walker.get_depends_on(package, 'run') | \
+        dependency_walker.get_depends_on(package, 'buildtool')
+    if len(depends_on) != 0:
+        # There are packages that depend on this "leaf", but we didn't find
+        # them initially because they weren't related to our inputs
+        unblocked_blocking_repos.add(repo)
+unblocked_leaf_repos = unblocked_leaf_repos.difference(
+    unblocked_blocking_repos)
+
+if len(blocked_repos.keys()) > 0:
+    print('The following repos cannot be released because of unreleased'
+          'dependencies:')
+    for blocked_repo_name in sorted(blocked_repos.keys()):
+        unreleased_repos = blocked_repos[blocked_repo_name]
+        print('\t{0}:'.format(blocked_repo_name))
+
+        print('\n'.join(
+            sorted('\t\t{0}'.format(repo) for repo in unreleased_repos)))
+
+if len(unblocked_leaf_repos) > 0:
+    print('The following repos can be released, but do not block other repos:')
+    print('\n'.join(
+        sorted('\t{0}'.format(repo) for repo in unblocked_leaf_repos)))
+
+if len(unblocked_blocking_repos) > 0:
+    print('The following repos can be released, and are blocking other repos:')
+    print('\n'.join(
+        sorted('\t{0}'.format(repo) for repo in unblocked_blocking_repos)))


dirk-thomas:
I can't comment on the output of the script yet since the dict / set is not easy to read. I will wait till you get to the third checkbox (pretty-print). I will add some comments about the code itself for now.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python
+
+import argparse
+
+import rosdistro
+import rosdistro.dependency_walker
+
+index = rosdistro.get_index(rosdistro.get_index_url())
+valid_distro_keys = index.distributions.keys()
+valid_distro_keys.sort()
```
dirk-thomas: This block should happen after `argparse`. So if the arguments are invalid the error is immediate and not after fetching remote data. It also would allow to add a index url argument later.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+
+parser = argparse.ArgumentParser(description='')
```
dirk-thomas: A "real" description what the script is doing would be good here.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+parser.add_argument('--distro', metavar='distribution', type=str,
```
dirk-thomas: For the metavar I think something like `ROS_DISTRO_NAME` would be more helpful. Therefore I would also suggest to rename the option to `--rosdistro`.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+                    help='The distribution to check packages for',
```
dirk-thomas: The **ROS** distribution...

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+                    default=valid_distro_keys[-1])
```
dirk-thomas: To show the default value in the help please use the `ArgumentDefaultsHelpFormatter`.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+
+# If not specified, check for all repositories in the previous distribution
+parser.add_argument('--repositories',
+                    metavar='repositories', type=str, nargs='*',
```
dirk-thomas: For the metavar I think something like `REPOSITORY_NAME` would be more helpful.

I would drop arguments which don't change the default: `type=str`

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+                    help='Unreleased repositories to check dependencies for')
+
+parser.add_argument('--depth',
+                    metavar='depth', type=int,
+                    help='Maxmium depth to crawl the dependency tree',
+                    default=None)
```
dirk-thomas: I would drop all arguments which don't change the default: `default=None`

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+
+args = parser.parse_args()
+
+distribution_key = args.distro
+repository_names = args.repositories
+dependency_walker_depth=args.depth
```
dirk-thomas: These variable name seem to be not shorter then their args field. Why add new variables for them?

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+    # For all package dependencies, check if they are released yet
+    # TODO: List the repos that need to be released, not the packages
+
+    unreleased_packages = package_dependencies.difference(
+        current_package_names)
+    # remove the packages which this repo provides.
+    unreleased_packages = unreleased_packages.difference(packages)
+    blocked_packages[repository_name] = unreleased_packages
+
+print blocked_packages
```
dirk-thomas: Please keep the script Python 3 compatible.

scripts/check_blocking_packages.py
dirk-thomas: Please wrap after the open parenthesis to avoid the content dependent indentation of the following lines.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+
+args = parser.parse_args()
+
+distribution_key = args.distro
+repository_names = args.repositories
+dependency_walker_depth=args.depth
+prev_distribution_key = None
+
+# Find the previous distribution to the current one
+for key, i in zip(valid_distro_keys, range(len(valid_distro_keys))):
```
dirk-thomas: for i, key in enumerate(valid_distro_keys):

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+    if key == distribution_key:
+        assert i > 0
+        prev_distribution_key = valid_distro_keys[i-1]
+        break
```
dirk-thomas: Since you require the value to be in the list what you actually want here is:

    try:
        i = valid_distro_keys.index(distribution_key
    except ValueError:
        # error message that there is no previous distro
    else:
        prev_distribution_key = valid_distro_keys[i - 1]

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+
+distribution = rosdistro.get_cached_distribution(index, distribution_key)
+prev_distribution = rosdistro.get_cached_distribution(
+    index, prev_distribution_key)
+
+distro_file = rosdistro.get_distribution_file(index, distribution_key)
+prev_distro_file = rosdistro.get_distribution_file(index, prev_distribution_key)
```
dirk-thomas: Don't get the distribution files directly. The cache distribution already contains the raw distribution file it was created from.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+
+dependency_walker = rosdistro.dependency_walker.DependencyWalker(
```
dirk-thomas: To shorten this do:

    from rosdistro.dependency_walker import DependencyWalker

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+else:
+    # Check the input arguments and remove entries that were already released.
+    prev_repository_names = set(repository_names)
+
+current_repository_names = set(distro_file.get_data()['repositories'].keys())
+repository_names_set = prev_repository_names.difference(
+    current_repository_names)
+if len(repository_names_set) == 0:
+    if repository_names is None:
+        print "Everything in {0} was released into the next {1}!"
```
dirk-thomas: Single quotes.

Same below.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+        # This repo doesn't have a release entry; skip it
+        continue
+    if not 'version' in repo_data['release']:
+        # No valid version so this repo is technically not released, skip it
+        continue
+    package_dependencies = set()
+    if 'packages' in repo_data['release']:
+        packages = repo_data['release']['packages']
+    else:
+        packages = [repository_name]
```
dirk-thomas: You can get that information from the release repo class directly: https://github.com/ros-infrastructure/rosdistro/blob/2bb3d9b7da0cee9b140f8b97c0a497c8585f6fad/src/rosdistro/release_repository_specification.py#L53-L57

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,127 @@
...
+# Construct a dictionary where keys are repository names and values are a list
+# of the missing dependencies for that repository
+# TODO: Could also check which packages are blocked by the input package
+# since DependencyWalker offers a get_depends_on() function
+
+blocked_packages = {}
+for repository_name in repository_names:
+    # Check all the dependencies for the repository
+    # Get all packages (based on info from the previous repository)
+    repo_data = prev_distro_file.get_data()['repositories'][repository_name]
```
dirk-thomas: Instead of accessing the yaml data directly you should use the already parsed data through the API. In this case:

    repo = distribution_file.repositories[repository_name]
    repo.version ...

scripts/check_blocking_packages.py
jacquelinekay: What about the case where i == 0?

dirk-thomas: I missed that. You can still check for that in the else case.

scripts/check_blocking_packages.py
jacquelinekay: ok, so I should call `get_distribution_cache` and get the `cache.distribution_file` out of that? Then presumably I should also pass the cache to `get_cached_distribution`.

I think that I also don't need to get the Distribution object for the current distribution, only the previous one (to get the DependencyWalker).

dirk-thomas: Yes, `cache.distribution_file`.

No, you don't need to pass `cache=...`. The default parameter `None` will result in the cache being fetched from the server.

jacquelinekay:
Preliminary "pretty print" has been implemented. It follows this format:
```
<unreleased_repo 1>:
    <first repo containing dependency>
    <second repo containing dependency>
<unreleased repo 2>:
    ...
If a repo has no unreleased dependencies, it prints:
<repo name>:
    No unreleased dependencies found! Go release the repository.
```

I was thinking of changing this so that instead of printing the repos with no unreleased dependencies in this way, a list of repositories that can be released is printed at the end.

jacquelinekay:
ok, I've changed the output so that packages that are unblocked (all dependencies were released into the new distro) are printed in a list at the end, in alphabetical order.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,160 @@
...
+    # Check the input arguments and remove entries that were already released.
+    prev_repository_names = set(repository_names)
+
+current_repository_names = set(distro_file.get_data()['repositories'].keys())
+# Print the repositories that will be eliminated from the input
+eliminated_repositories = prev_repository_names.intersection(
+    current_repository_names)
+if len(eliminated_repositories) > 0:
+    print('Ignoring inputs which have already been released:')
+    print('\n'.join('\t{0}'.format(repo) for repo in eliminated_repositories))
```
dirk-thomas: `sorted` please.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,160 @@
...
+        # Get the repository for the unreleased packages
+        blocked_repos[repository_name] = blocking_repos
+
+# TODO Print the inputs with no unreleased dependencies separately, in a list
+for blocked_repo_name in blocked_repos.keys():
+    unreleased_repos = blocked_repos[blocked_repo_name]
+    print('{0}:'.format(blocked_repo_name))
+
+    print('\n'.join(
+        sorted('\t{0}'.format(repo) for repo in unreleased_repos)))
```
dirk-thomas: This loop of output needs a "headline" stating what is printed here.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,160 @@
...
+
+if len(unblocked_repos) > 0:
+    print('\nThe following unreleased repositories are ready to be released:')
+    print('\n'.join(sorted('\t{0}'.format(repo) for repo in unblocked_repos)))
```
dirk-thomas: I think it would be valuable here to add if these repos are blocking anything.

jacquelinekay:
I've implemented sorting of repos that are ready to be released and are blocking other repo. The output is now:
```
Ignoring inputs which have already been released:
    <repos>
The following repos are blocked by other packages:
    <repo name>:
       <blocking repos>
The following repos are unblocked, but do not block other repos:
    <repos>
The following repos are unblocked and are blocking other repos:
(They should be released ASAP!)
    <repos>
```

wjwwood:
That sounds awesome @jacquelinekay! One question is the `<blocking repos>` recursive or just one level deep?

jacquelinekay:
Blocking repos is recursive. You can specify the depth level as a command line argument; it gets passed to `DependencyWalker`.

jacquelinekay:
I could add some more enhancements (count how many packages are blocked by blocking packages, display package dependency graph, etc.) but I think this is ready to squash, merge, and maybe advertise that it exists on ros-release. Then someone can pick up the enhancements at a later date.

scripts/check_blocking_packages.py
dirk-thomas: Still the case.

scripts/check_blocking_packages.py
dirk-thomas: Still the case.

scripts/check_blocking_packages.py
dirk-thomas: Still the case.

scripts/check_blocking_packages.py
dirk-thomas: Still the case.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,187 @@
...
+        package, ['build', 'run', 'buildtool'])
+    if len(depends_on) != 0:
+        # There are packages that depend on this "leaf", but we didn't find
+        # them initially because they weren't related to our inputs
+        unblocked_blocking_repos.add(repo)
+unblocked_leaf_repos = unblocked_leaf_repos.difference(
+    unblocked_blocking_repos)
+
+if len(blocked_repos.keys()) > 0:
+    print("The following repos are blocked by other packages:")
```
dirk-thomas: Single quotes.

scripts/check_blocking_packages.py
dirk-thomas: Still the same.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,187 @@
...
+    prev_repository_names = set(
+        prev_distro_file.get_data()['repositories'].keys())
+else:
+    # Check the input arguments and remove entries that were already released.
+    prev_repository_names = set(repository_names)
+
+current_repository_names = set(distro_file.get_data()['repositories'].keys())
+# Print the repositories that will be eliminated from the input
+eliminated_repositories = prev_repository_names.intersection(
+    current_repository_names)
```
dirk-thomas: This check is not sufficient. It needs to check if the repository has a release section as well as that the release section has a version.

scripts/check_blocking_packages.py
```diff
@@ -0,0 +1,187 @@
...
+            sorted('\t\t{0}'.format(repo) for repo in unreleased_repos)))
+
+if len(unblocked_leaf_repos) > 0:
+    print('The following repos are unblocked, but do not block other repos:')
+    print('\n'.join(
+        sorted('\t{0}'.format(repo) for repo in unblocked_leaf_repos)))
+
+if len(unblocked_blocking_repos) > 0:
+    print('The following repos are unblocked and are blocking other repos:')
+    print('(They should be released ASAP!)')
```
dirk-thomas: I would phrase it: `They can be released now`

scripts/check_blocking_packages.py
dirk-thomas: I think showing the block repositories would be more valuable rather then the blocking packages. It would also make the list shorter hence more readable.

scripts/check_blocking_packages.py
jacquelinekay: on line 130 of the current version of the file I get `package_names` from the ReleaseRepositorySpecification, so I'm not sure what the suggestion is.

scripts/check_blocking_packages.py
jacquelinekay: This message is actually wrong, it is currently printing the names of repositories, not packages.

scripts/check_blocking_packages.py
jacquelinekay: hmm, I guess I'm not using the proper indexing into the struct throughout the file--give me a few minutes to fix it

dirk-thomas: I guess it moved - currently around line 100: https://github.com/ros/rosdistro/pull/10829/files#diff-420c2a5f1d18b206dcdc349a0581244dR100

jacquelinekay:
I made changes for the most recent comments and went a little crazy with list comprehensions. Let me know if there's anything I missed.

dirk-thomas:
Should the script be renamed to `check_blocking_repos.py`?

dirk-thomas:
+1

PR #10830, opened 2016-03-22
Add postgis for Wily and Xenial

cberner:


tfoote:
+1 http://packages.ubuntu.com/search?keywords=postgresql there's only one version per ubuntu distro

@cberner Do these debian packages look equivalent? 

wheezy: https://packages.debian.org/wheezy/postgresql-9.1-postgis
jessie: https://packages.debian.org/jessie/postgresql-9.4-postgis-2.1

wjwwood:
lgtm. It would be nice to get the Debian rules in too.

cberner:
@tfoote, yes I think so

cberner:
Updated to include the Debian packages

PR #10850, opened 2016-03-23
python-pygame for all Ubuntu versions

trainman419:
pygame is necessary for korg_nanokontrol, and it's the same dependency name on all versions of Ubuntu.

**squish**


tfoote:
+1

PR #10863, opened 2016-03-24
add debian rule for python-lxml

tfoote:
https://packages.debian.org/sid/python-lxml

Needed for urdfdom_py @jacquelinekay


jacquelinekay:
+1

PR #10891, opened 2016-03-25
Add UTM pip package

anassinator:
https://pypi.python.org/pypi/utm

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2164,7 +2164,7 @@ python-usb:
     raring: [python-usb]
     saucy: [python-usb]
     trusty: [python-usb]
-python-utm:
+python-utm-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2164,6 +2164,19 @@ python-usb:
     raring: [python-usb]
     saucy: [python-usb]
     trusty: [python-usb]
+python-utm:
```
dirk-thomas: The key should include `pip` to have an upgrade path once this package is being packaged by the platforms.

anassinator: Fixed. Need I squash into one commit?

dirk-thomas: Thanks, no squashing is not necessary.

dirk-thomas:
+1

PR #10892, opened 2016-03-25
Add bitstring pip package

anassinator:
https://pypi.python.org/pypi/bitstring


dirk-thomas:
+1

dirk-thomas:
The latest Ubuntu distros have that package: http://packages.ubuntu.com/xenial/python-bitstring

anassinator:
Oh, sorry I was running 14.04.
Should I add it? And if so, should the `-pip` suffix be removed?

wjwwood:
@anassinator you can have both, the `-pip` suffixed one and the one without.

anassinator:
Alright! I think I fixed it in #10893

PR #10903, opened 2016-03-27
Define bullet system dep for Xenial

jonbinney:


dirk-thomas:
Could this be simplified to the following?

    ubuntu: [libbullet-dev]

Anyway +1 from me.

jonbinney:
Yeah I wasn't sure why the previous ubuntu versions had separate definitions for this, but I just continued the pattern

tfoote:
There was a point when it wasn't packaged upstream (it was a ros package back in something like cturtle era) We probably could collapse it now that it appears all supported platforms have the package available. But this will work for now.

PR #10912, opened 2016-03-28
Update debian rosdep rule for libjpeg-dev

tfoote:
It's a dummy package and the provider changed from libjpeg8-dev to libjpeg62-turbo-dev

https://packages.debian.org/search?searchon=names&keywords=libjpeg
https://packages.debian.org/jessie/libjpeg62-turbo-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1644,7 +1644,9 @@ libjackson-json-java:
   ubuntu: [libjackson-json-java]
 libjpeg:
   arch: [libjpeg-turbo]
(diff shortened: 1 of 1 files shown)


tfoote:
@wjwwood this is causing stage in Kinetic to fail to build on debian: http://54.183.26.131:8080/view/Kbin_arm_dJv8/job/Kbin_arm_dJv8__stage__debian_jessie_arm64__binary/5/console

wjwwood:
+1, thanks for looking into it.

PR #10925, opened 2016-03-30
add python-defusedxml

codebot:
Tested on Ubuntu. I attempted to chase it down on other distros.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -485,6 +485,15 @@ python-debian:
     utopic_python3: [python3-debian]
     vivid: [python-debian]
     vivid_python3: [python3-debian]
+python-defusedxml:
+  arch: [python-defusedxml]
+  debian: [python-defusedxml]
+  fedora: [python-defusedxml]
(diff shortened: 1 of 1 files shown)


tfoote:
+1 packages found at: 

ubuntu: http://packages.ubuntu.com/search?keywords=python-defusedxml
debian: https://packages.debian.org/search?keywords=python-defusedxml
arch: https://www.archlinux.org/packages/extra/any/python-defusedxml/
fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/python-defusedxml/
gentoo: https://packages.gentoo.org/packages/dev-python/defusedxml

tfoote:
Note this is not available on Saucy so it cannot go into indigo until after the next sync when saucy is disabled.

wjwwood:
You need OS X (pip).

tfoote:
It's fine to add the rules at any time. We could do it for Kinetic only but this seems like an improvement that could be considered for backporting.

dirk-thomas:
The package is available in *Saucy*.

PR #10927, opened 2016-03-31
adding python-cookiecutter

asmodehn:


dirk-thomas:
+1

tfoote:
Ubuntu: http://packages.ubuntu.com/search?keywords=python-cookiecutter
Debian: https://packages.debian.org/jessie/python/python-cookiecutter
OSX: looks like either pip or brew https://cookiecutter.readthedocs.org/en/latest/installation.html
Fedora I'm not finding a package quickly 
Could you also look up the package names for debian, osx and fedora? We're trying to keep things

asmodehn:
couldn't find fedora... so I used pip for that as well.

tfoote:
Thanks for filling it in.

PR #10931, opened 2016-03-31
adding python-pip

asmodehn:


dirk-thomas:
Without having pip some of the functionality of rosdep won't work in the first place. So I am not sure why you need to explicitly depend on this in a package.

Beside that lgtm.

asmodehn:
I m just working on a package that explicitely relies on pip ( catkin
extension ). And since python is listed in rosdep as well, I thought it
made sense.

PR #10940, opened 2016-03-31
add rosdep key python-qt5-bindings

dirk-thomas:
Waiting for availability of a custom `PySide2` package.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -290,6 +290,8 @@ python-cairo:
     trusty: [python-cairo]
     utopic: [python-cairo]
     vivid: [python-cairo]
+    wily: [python-cairo]
+    xenial: [python-cairo]
 python-catkin-lint:
   fedora: [python-catkin_lint]
   ubuntu: [python-catkin-lint]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1564,6 +1564,13 @@ python-qt4-gl:
   ubuntu:
     apt:
       packages: [python-qt4-gl]
+python-qt5-bindings:
+  debian:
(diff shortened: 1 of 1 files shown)


wjwwood:
The keys in there look good to me.

tfoote:
+1

dirk-thomas:
We need the following packages as well as their dependencies (as added in the second commit):
* `python-pyside2`
* `libpyside2-dev`
* `shiboken2`
* `libshiboken2-dev`

We can get prerelease packages from the following PPA: https://code.launchpad.net/~thopiekar/+archive/ubuntu/pyside-git

@tfoote Can you please import the necessary packages into our repos. I assume we will just not add the PySide2 dependencies for other platforms then?

tfoote:
@dirk-thomas what are the packages we need. Can you propose a config file to be added here https://github.com/ros-infrastructure/reprepro-updater/tree/master/config with the appropriate regex?

For the other platforms do we gracefully degrade if pyside is not installed?

dirk-thomas:
We need the four packages listed in my previous comment. I don't know what you commonly use to determine all recursive dependencies of such packages or if that is a manual process?

tfoote:
Please submit a PR with your best guess for the config and with the justification for why to add them. The git history on that repos is currently our best database of what's been imported and why.

Usually the recursive dependencies are available in the upstream repos and we don't need to explicitly import them. That would only happen if that PPA depended on another PPA.

dirk-thomas:
ros-infrastructure/reprepro-updater#28

dirk-thomas:
The packages have been imported from the PPA. I also added Wily and Xenial entries for two more Python packages. If there are no further comments I will go ahead and merge this.

PR #10966, opened 2016-04-01
add debian rules for libvlc

tfoote:
Follow up to #10959

Package pages:
- https://packages.debian.org/jessie/libvlc-dev
- https://packages.debian.org/jessie/vlc-nox
- https://packages.debian.org/jessie/libvlc5


ipa-fxm:
+1 (thanks)

PR #11004, opened 2016-04-05
Fixed/renamed rosdep entry for omniorb-idl

meyerj:
I would need a rosdep entry for the Ubuntu package `omniorb-idl`.

The original entry `omniorb4-idl` was added back in 2012 in https://github.com/ros/rosdistro/commit/94dd9ce4c76472d742e8281b4ca9176d99b23109. The package was renamed to `omniorb-idl` since Ubuntu 11.04, which still provided `omniorb4-idl` as a transitional package.

Renaming the key to `omniorb-idl` might break existing packages, but because the Ubuntu package name was wrong anyway after 11.04, I do not think that there are any released packages using that key. Not specifying a version number in the rosdep name is more in line with other rosdep entries like `omniorb`.

I verified that the referenced packages exist in current Debian, Fedora and ArchLinux releases, too.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2931,11 +2931,11 @@ omniorb:
     trusty: [omniorb, omniidl, omniorb-nameserver, libomniorb4-dev]
     utopic: [omniorb, omniidl, omniorb-nameserver, libomniorb4-dev]
     vivid: [omniorb, omniidl, omniorb-nameserver, libomniorb4-dev]
-omniorb4-idl:
+omniorb-idl:
   arch: [omniorb]
   debian: [omniorb-idl]
   fedora: [omniORB-devel]
-  ubuntu: [omniorb4-idl]
+  ubuntu: [omniorb-idl]
 opende:
   arch: [ode]
   fedora: [ode]


tfoote:
+1 can you link to the upstream sources here for future reference?

meyerj:
> +1 can you link to the upstream sources here for future reference?

Upstream sources:
- arch: https://aur.archlinux.org/packages/omniorb/
- debian: https://packages.debian.org/search?keywords=omniorb-idl&searchon=names&suite=all&section=all
- fedora: https://apps.fedoraproject.org/packages/omniORB-devel
- ubuntu: http://packages.ubuntu.com/search?suite=all&searchon=names&keywords=omniorb-idl

Is that what you meant?

In ArchLinux and Fedora the IDL files are not in a separate package like in Debian and Ubuntu. An alternative to this PR would be to simply add `omniorb-idl` as a target package to the `omniorb` key for Debian and Ubuntu, especially because `omniorb-idl` is quite light-weight and not really useful without the other omniorb packages.

tfoote:
Yeah, those are the links I was referring to. If it's more consistent to bundle them together with the same rule, that probably makes more sense then.

meyerj:
Closing this in favor of #11006.

Issue #10914, opened 2016-03-28
ffmpeg is not available on debian jessie


tfoote:
There's a note here: https://wiki.debian.org/ffmpeg
And it's in jessie backports: https://packages.debian.org/search?searchon=names&keywords=ffmpeg

I think this might be a good time to leverage our bootstrap repo and bring in the debian jessie backport since it's available on all versions before and after. 

This is breaking the opencv3 builds: http://54.183.26.131:8080/view/Kbin_dJ64/job/Kbin_dJ64__opencv3__debian_jessie_amd64__binary/16/console

wjwwood:
I guess `libav` doesn't work in this case? It seems like they recommend it over `ffmepg` in the wiki you linked.

I suppose back porting it is ok, whatever you think is appropriate.

tfoote:
It seems that several projects switched to libav but are now in the process of switching back to ffmpeg. See: http://superuser.com/a/941670 and http://stackoverflow.com/a/9477756 I'll look into adding an import of the backport.

tfoote:
I added a config here: https://github.com/ros-infrastructure/reprepro-updater/blob/master/config/ffmpeg_debian_jessie_backport.yaml 

And successfully imported on the buildfarm: http://build.ros.org:8080/job/import_upstream/26/console

They will propogate to packages.ros.org with the next syncs.

PR #11007, opened 2016-04-05
Update ffmpeg for Jessie

vrabaud:
as the ffmpeg package does not exist anymore as mentioned on
https://wiki.debian.org/ffmpeg

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -691,7 +691,10 @@ festival-dev:
   ubuntu: [festival-dev]
 ffmpeg:
   arch: [ffmpeg]
-  debian: [ffmpeg, libavcodec-dev, libavformat-dev, libavutil-dev, libswscale-dev]
+  debian:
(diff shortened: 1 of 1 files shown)


vrabaud:
ffmpeg and libav-tools are problematic: on Ubuntu, libav-tools is a virtual (therefore not handled by rosdep) package that points to ffmpeg on Wily and above. On Debian Jessie, there is no ffmpeg

dirk-thomas:
LGTM

tfoote:
We imported ffmpeg from debian jessie backports because ffmpeg is now again the mainstream one and jessie is the only one with it missing. We should plan to use it going foreward too: The import looks to be missing something: https://github.com/ros-infrastructure/reprepro-updater/issues/29 

We just discussed it in https://github.com/ros/rosdistro/issues/10914 I think we should fix the backport  and continue to use it  such that all platforms will have ffmpeg instead of some having avconv

PR #11009, opened 2016-04-05
add python-oauth2client

minshallj:


rosdep/python.yaml
```diff
@@ -1057,6 +1057,11 @@ python-oauth2:
   debian: [python-oauth2]
   fedora: [python-oauth2]
   ubuntu: [python-oauth2]
+python-oauth2client:
+  debian: [python-oauth2client]
+  fedora: [python-oauth2client]
+  ubuntu:
+    trusty: [python-oauth2client]
```
dirk-thomas: Is the package only available on Trusty?

minshallj: only tested it on trusty, didn't want to put it on everything until I had time to test

minshallj: is there a good way to find out what releases have this package other than spinning up dockers and apt-getting it?

wjwwood: http://packages.ubuntu.com/ ?

minshallj: Nice! Ok, updating http://packages.ubuntu.com/search?keywords=python-oauth2client&searchon=names&suite=all&section=all

dirk-thomas: This page won't show packages in EOL Ubuntu distro which we should still list here if the package is available for them.

minshallj: This package might actually start at 14.10. I just tried to install it inside a 13.10 docker and the package couldn't be found (even though python-oauth2 could be).

minshallj: @dirk-thomas did you see my previous message? I think the distros specified are all the distros that the package exists on.

dirk-thomas: Yes, I have read you comment. And as I mentioned e.g. Utopic (14.10) is not listed on the Ubuntu website anymore since it has been EOLed. So looking at the website won't answer the question if it is available. Either look on old-releases.ubuntu.com or use docker to check.

minshallj: My apologies. I've added Utopic in now.

dirk-thomas:
+1

PR #11011, opened 2016-04-06
Adding drc_hubo to documentation index for distro

JeongsooLim:
I'd like drc_hubo to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1865,6 +1865,15 @@ repositories:
       type: git
       url: https://github.com/ZhuangYanDLUT/dlut_vision.git
       version: indigo-devel
+  drc_hubo:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -1865,6 +1865,15 @@ repositories:
       type: git
       url: https://github.com/ZhuangYanDLUT/dlut_vision.git
       version: indigo-devel
+  drc_hubo:
+    doc:
+      type: git
+      url: https://github.com/JeongsooLim/drc_hubo.git
+      version: indigo-devel
```
dirk-thomas: Your repository doesn't have a branch with that name. I assume you mean `master `?

Same below.

JeongsooLim: I make a new branch named "indigo-devel".
Please check it again.

PR #11022, opened 2016-04-07
Update python-wxgtk for ubuntu and debian

jackieokay:
This is a temporary measure to fix the failing robot_model job, because python-wxgtk2.8 is not available in Jessie or Xenial. The eventual goal is to replace the WxWidgets dependency in joint_state_publisher with Qt.


rosdep/base.yaml
```diff
@@ -3644,14 +3644,32 @@ wx-common:
 wxpython:
   arch: [wxpython]
   centos: [wxPython-devel]
-  debian: [python-wxgtk2.8]
+  debian:
+    jessie: [python-wxgtk3.0]
+    sid: [python-wxgtk2.8]
```
tfoote: Sid should be 3.0 it's the experimental rolling release for debian. https://packages.debian.org/sid/python-wxgtk3.0

tfoote:
Ubuntu versions of 2.8: http://packages.ubuntu.com/search?keywords=python-wxgtk2.8 ends with wily
Ubuntu versions of 3.0: http://packages.ubuntu.com/search?keywords=python-wxgtk3.0

Debian version 2.8 ends with wheezy: https://packages.debian.org/wheezy/python-wxgtk2.8
Debian version for 3.0 starts with Jessie: https://packages.debian.org/sid/python-wxgtk3.0

jacquelinekay:
thanks @tfoote, I'm still getting used to Debian... Does it look correct now?

jacquelinekay:
Going to close this in favor of ros/robot_model#128.

tfoote:
It may not be used for this use case but I think we've converged on the accurate rosdep rules. @jacquelinekay I think this is worth merging to keep correctness in the rules. Thoughts?

jacquelinekay:
Sure, why not. I don't need it anymore but maybe someone else really wants wxWidgets.

PR #11028, opened 2016-04-08
add an option to the blocking packages to compare to a named rosdistro

tfoote:
instead of just the last one

### DIFF
--- a/scripts/check_blocking_repos.py
+++ b/scripts/check_blocking_repos.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+
 import argparse
 
 import rosdistro
 from rosdistro.dependency_walker import DependencyWalker
-
+import sys
 
 def is_released(repo, dist_file):
     return repo in dist_file.repositories and \
@@ -52,15 +54,16 @@ if distro_key is None:
 try:
     i = valid_distro_keys.index(distro_key)
 except ValueError:
-    print('Distribution key not found in list of valid distributions.')
+    print('Distribution key not found in list of valid distributions.', file=sys.stderr)
     exit(-1)
-if i == 0:
-    print('No previous distribution found.')
+if i == 0 and not args.comparison:
+    print('No previous distribution found.', file=sys.stderr)
     exit(-1)
 
 if args.comparison:
     if args.comparison not in valid_distro_keys:
-        print('Invalid rosdistro selected for comparison')
+        print('Invalid rosdistro [%s] selected for comparison.' % args.comparison, file=sys.stderr)
+        print('Valid rosdistros are %s.' % valid_distro_keys, file=sys.stderr)
         exit(-1)
     prev_distro_key = args.comparison
 else:
--- a/scripts/check_blocking_repos.py
+++ b/scripts/check_blocking_repos.py
@@ -124,14 +124,13 @@ if len(eliminated_repositories) > 0:
 
 repo_names_set = prev_repo_names.difference(
     current_repo_names)
+invalid_names = set(repo_names_argument).difference(prev_repo_names)
 
 if len(repo_names_set) == 0:
-    if repo_names_argument is None:
-        print('Everything in {0} was released into {1}!'.format(
-            prev_distro_key, distro_key))
-    else:
-        print('All inputs are invalid or were already released in {0}.'.format(
-            distro_key))
+    print('All inputs are invalid or were already released in {0}.'.format(
+        distro_key))
+    if invalid_names:
+        print('Could no resolve: %s in %s' % (list(invalid_names), prev_distro_key))
     print('Exiting without checking any dependencies.')
     exit(0)
 
@@ -213,3 +212,8 @@ if len(unblocked_blocking_repos) > 0:
     print('The following repos can be released, and are blocking other repos:')
     print('\n'.join(
         sorted('\t{0}'.format(repo) for repo in unblocked_blocking_repos)))
+
+if len(invalid_names):
+    print('Could no resolve the following arguments in %s: ' % prev_distro_key)
+    print('\n'.join(
+        sorted('\t{0}'.format(repo) for repo in invalid_names)))
--- a/scripts/check_blocking_repos.py
+++ b/scripts/check_blocking_repos.py
@@ -41,7 +41,7 @@ parser.add_argument(
 args = parser.parse_args()
 
 distro_key = args.rosdistro
-repo_names = args.repositories
+repo_names_argument = args.repositories
 prev_distro_key = None
 
 index = rosdistro.get_index(rosdistro.get_index_url())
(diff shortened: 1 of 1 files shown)


scripts/check_blocking_repos.py
```diff
@@ -51,7 +57,14 @@ def is_released(repo, dist_file):
 if i == 0:
     print('No previous distribution found.')
     exit(-1)
-prev_distro_key = valid_distro_keys[i - 1]
+
+if args.comparison:
```
dirk-thomas: This case doesn't work if `--rosdistro` is the first distro in the list. Please modify the check from line 57 accordingly.

scripts/check_blocking_repos.py
```diff
@@ -51,7 +57,14 @@ def is_released(repo, dist_file):
...
+    if args.comparison not in valid_distro_keys:
+        print('Invalid rosdistro selected for comparison')
```
dirk-thomas: To `stderr`.

scripts/check_blocking_repos.py
tfoote: I didin't think of comparing to future distros. I also added logic to not run if comparing to itself.

tfoote:
Updated to stderr and more robust checks.

scripts/check_blocking_repos.py
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+
 import argparse
 
 import rosdistro
 from rosdistro.dependency_walker import DependencyWalker
-
+import sys
```
dirk-thomas: System package should be in one block, hence behind line 5 according to PEP8.

scripts/check_blocking_repos.py
```diff
@@ -46,12 +54,24 @@ def is_released(repo, dist_file):
...
+    if args.comparison not in valid_distro_keys:
+        print('Invalid rosdistro [%s] selected for comparison.' % args.comparison, file=sys.stderr)
+        print('Valid rosdistros are %s.' % valid_distro_keys, file=sys.stderr)
+        exit(-1)
+    prev_distro_key = args.comparison
+else:
+    prev_distro_key = valid_distro_keys[i - 1]
+
+if prev_distro_key == distro_key:
+    print('It does not make sense to compare distro [%s] to itself.', file=sys.stderr)
```
dirk-thomas: This check only makes sense within the condition of line 63. So moving it under line 68 would be the right scope.

scripts/check_blocking_repos.py
```diff
@@ -46,12 +54,24 @@ def is_released(repo, dist_file):
...
-if i == 0:
-    print('No previous distribution found.')
+if i == 0 and not args.comparison:
+    print('No previous distribution found.', file=sys.stderr)
+    exit(-1)
+
+if args.comparison:
+    if args.comparison not in valid_distro_keys:
+        print('Invalid rosdistro [%s] selected for comparison.' % args.comparison, file=sys.stderr)
+        print('Valid rosdistros are %s.' % valid_distro_keys, file=sys.stderr)
```
dirk-thomas: This should not include the `--rosdistro` since it is not a valid choice.

scripts/check_blocking_repos.py
dirk-thomas: Having this check in a wider scope while it is only relevant in a smaller scope is imo not following out programming conventions to "use the narrowest scope possible".

tfoote:
Updated from the review comments and added a few more niceties.

dirk-thomas:
The patch seems to contain redundant code blocks around the pending comments.

tfoote:
Fixed the bad merge/rebase. Somehow there were two copies of one of the early commits.

scripts/check_blocking_repos.py
```diff
@@ -1,11 +1,13 @@
...
 
+from __future__ import print_function
+
 import argparse
+import sys
 
 import rosdistro
 from rosdistro.dependency_walker import DependencyWalker
 
-
```
dirk-thomas: Please avoid unrelated whitespace changes - especially when they add a PEP8 warning.

scripts/check_blocking_repos.py
```diff
@@ -46,12 +54,28 @@ def is_released(repo, dist_file):
...
+        print('Invalid rosdistro [%s] selected for comparison to [%s].' % (args.comparison, distro_key),
+              file=sys.stderr)
+        print('Valid rosdistros are %s.' % valid_comparison_keys, file=sys.stderr)
+        exit(-1)
+    prev_distro_key = args.comparison
+else:
+    prev_distro_key = valid_distro_keys[i - 1]
+
+if i == 0 and not args.comparison:
+    print('No previous distribution found.', file=sys.stderr)
```
dirk-thomas: This still seems to be redundant with line 60-62.

scripts/check_blocking_repos.py
```diff
@@ -89,15 +111,13 @@ def is_released(repo, dist_file):
...
-        print('Everything in {0} was released into the next {1}!'.format(
-            prev_distro_key, distro_key))
-        print('This was probably a bug.')
-    else:
-        print('All inputs are invalid or were already released in {0}.'.format(
-            distro_key))
+    print('All inputs are invalid or were already released in {0}.'.format(
+        distro_key))
+    if invalid_names:
+        print('Could no resolve: %s in %s' % (list(invalid_names), prev_distro_key))
```
dirk-thomas: All error messages should go to `stderr`.

scripts/check_blocking_repos.py
```diff
@@ -179,3 +199,8 @@ def is_released(repo, dist_file):
     print('The following repos can be released, and are blocking other repos:')
     print('\n'.join(
         sorted('\t{0}'.format(repo) for repo in unblocked_blocking_repos)))
+
+if len(invalid_names):
+    print('Could no resolve the following arguments in %s: ' % prev_distro_key)
+    print('\n'.join(
+        sorted('\t{0}'.format(repo) for repo in invalid_names)))
```
dirk-thomas: Should this message go to `stderr` too?

tfoote: I was thinking of it as a summary output, but it does make sense to make it an error code and return 1 if it does not resolve all arguments.

dirk-thomas:
The diff would be shorter and therefore easier to read if existing variable wouldn't be renamed since there is no need for it.

Anyway this lgtm.

tfoote:
You can't see it in the diff, but the variable `repo_names` is used lower down in a different context so I changed it to not reuse the variable.

scripts/check_blocking_repos.py
```diff
@@ -1,15 +1,18 @@
...
 
 
-def is_released(repo, dist_file):
-    return repo in dist_file.repositories and \
-        dist_file.repositories[repo].release_repository is not None and \
-        dist_file.repositories[repo].release_repository.version is not None
+def is_released(repository, dist_file):
+    return repository in dist_file.repositories and \
+        dist_file.repositories[repository].release_repository is not None and \
+        dist_file.repositories[repository].release_repository.version is not None
```
dirk-thomas: The same for the variable in this block.

PR #11035, opened 2016-04-09
Change Collada package name for Debian

jackieokay:
I'm pretty sure there's no package called `collada-dom-dev` in Debian; perhaps it was renamed since the rosdep key was first added.

https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=collada-dom


dirk-thomas:
lgtm

PR #11038, opened 2016-04-11
Rosdep python-networkmanager

stonier:
Looks like this is only available on trusty onwards, see [here](http://packages.ubuntu.com/search?keywords=python-networkmanager).

I don't have any packages (nor open source packages) that require this prior to trusty though, so a general key is probably sufficient until individual ones are needed.

If you'd like the distro-specific keys, let me know.


dirk-thomas:
The referenced website does not list EOL distribution. I would assume the package is available in older distributions too.

+1

PR #11077, opened 2016-04-13

meuchel:
I would like my package uavc_v4lctl to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -11306,7 +11306,7 @@ repositories:
     source:
       type: git
       url: https://github.com/meuchel/uavc_v4lctl.git
-      version: 1.0.2
+      version: master
     status: maintained
   ubiquity_launches:
     doc:
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -11298,9 +11298,8 @@ repositories:
       type: git
       url: https://github.com/meuchel/uavc_v4lctl.git
       version: 1.0.0
-    release:
-      tags:
-        release: release/indigo/{package}/{version}
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


vrabaud:
The release property should be for a release repo. You might as well remove it here.

meuchel:
I changed distribution.yaml.
Its better now?

vrabaud:
for doc and source, the version should be checkout-able. Usually people put a branch (probably master) but in your case, the tag 1.0.0 does not exist.
You should probably read http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease to set your version number (catkin_prepare_release is a great tool for that)

meuchel:
Hi vrabaud, I tried again and have also a tag now but it's still failing. What is now wrong? I have no idea.

vrabaud:
@tfoote , we usually set a branch and not a tag for sources but a tag should work according to REP 141: http://www.ros.org/reps/rep-0141.html#distribution-file

The code does not indeed check for tags.

https://github.com/ros/rosdistro/blob/master/test/test_url_validity.py#L125 should have tags_valid=True like for docs no ?

tfoote:
> @tfoote , we usually set a branch and not a tag for sources but a tag should work according to REP 141: http://www.ros.org/reps/rep-0141.html#distribution-file
> 
> The code does not indeed check for tags.
> 
> https://github.com/ros/rosdistro/blob/master/test/test_url_validity.py#L125 should have tags_valid=True like for docs no ?

Jenkins polling as used in devel jobs always triggers every cycle if pointed at a tag. Thus if you use a tag in the source job it will almost continuously run your builds.

PR #11097, opened 2016-04-14
fix VTK for Debian

vrabaud:


vrabaud:
How about Fedora ?

jacquelinekay:
+1

vrabaud:
Fedora on VTK6 so all good: https://apps.fedoraproject.org/packages/vtk

PR #11103, opened 2016-04-15
Adding uavc_v4lctl to documentation index for distro jade

meuchel:
Request adding uavc_v4lctl to documentation index for distro jade

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -5380,6 +5380,21 @@ repositories:
       url: https://github.com/ros-teleop/twist_mux_msgs.git
       version: jade-devel
     status: maintained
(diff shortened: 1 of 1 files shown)


wjwwood:
@vrabaud he has an invalid release entry here, which causes a failing source deb job every 15 minutes. In the future, if someone says they are adding doc only, watch out for it having an invalid release entry as well, as is the case here. For example, the failing job:

http://build.ros.org/job/Jsrc_uT__uavc_v4lctl__ubuntu_trusty__source/35/console

@meuchel I'll have to revert this. Please open a new pull request which excludes the "release" entry. That entry is only valid after doing a release with `bloom-release`.

PR #11136, opened 2016-04-17
Adding uavc_v4lctl to documentation index for distro jade (2)

meuchel:
Second try request adding uavc_v4lctl to documentation index for distro jade

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -5388,6 +5388,21 @@ repositories:
       url: https://github.com/ros-teleop/twist_mux_msgs.git
       version: jade-devel
     status: maintained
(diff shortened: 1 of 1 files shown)


meuchel:
Not sure what really went wrong when looking the log.
I have a release 1.0.3-0.
Any hint?

tfoote:
You need to run bloom for jade on your package. 

The upstream version exists 1.0.3, but you need to run bloom to generate all the necessary tags in the release repo for jade. (It has different target platforms and the packages are named differently. (ros-jade-*) when packaged. 

I suggest you close this and let bloom open a replacement.

meuchel:
Closed and retry

PR #11156, opened 2016-04-19
Add rosdep definitions for pcl and python-crypto on debian

tfoote:
https://packages.debian.org/jessie/python-crypto
https://packages.debian.org/jessie/libpcl-dev
https://packages.debian.org/jessie/libpcl1.7


wjwwood:
lgtm

PR #11175, opened 2016-04-20
add rplidar_ros to documentation index for indigo/jade

kintzhao:
I'd like rplidar_ros to be indexed and documented on ros.org. Because the github addres of rplidar_ros package in the version of indigo have been changed  by a mistake .


kintzhao:
I have fixed it .  rplidar_ros is not the package of cob_substitute, and remove it

dirk-thomas:
Please don't open duplicate tickets or close obsolete tickets. The comment from the duplicate ticket also applies here.

PR #11183, opened 2016-04-20
remove source from jsk_apc and jsk_common

k-okada:


dirk-thomas:
May I ask why you want to remove the source entries?

k-okada:
If I understand correctly, this entry triggers
http://build.ros.org/job/Idev__jsk_apc__ubuntu_trusty_amd64/

This package is failing due to missing dependency (?? Maybe)
And we're resting on Travis , also contribute not to apply heave load on
ros buildfarm

2016年4月21日木曜日、Dirk Thomas<notifications@github.com>さんは書きました:

> May I ask why you want to remove the source entries?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/ros/rosdistro/pull/11183#issuecomment-212515450>
>


-- 
--
◉ Kei Okada

dirk-thomas:
Removing the entry prevent users to compile the latest code from source using e.g. `rosinstall_generator`. You can only opt-out of the devel jobs by setting `test_commits: false` within the source section (see [REP 143](http://www.ros.org/reps/rep-0143.html#distribution-file)). (While the REP says that the default value is already `false` the ROS build farm overrides the default to be `true` in order to provide devel jobs out-of-the-box.

k-okada:
I see, updated to use `test_commits` and `test_pull_request`

dirk-thomas:
Please remove the `test_pull_requests` entries. Their default value is already `false`.

dirk-thomas:
And the `test_commits` entry needs to be in alphabetical order.

k-okada:
I see, fixed

PR #11189, opened 2016-04-20
remove uavc_v4lctl from Kinetic for now

jackieokay:
As discussed in meuchel/uavc_v4lctl#1. The package provided by this repo fails to build on all platforms right now for Kinetic, so I am removing the release entry.


dirk-thomas:
You can just remove the `version` field instead. That keeps all the other information for the next bloom release.

jacquelinekay:
ok, done.

PR #11196, opened 2016-04-21
rplidar_ros: 1.5.2-0 in 'jade/distribution.yaml' [bloom]

kintzhao:
Increasing version of package(s) in repository `rplidar_ros` to `1.5.2-0`:
- upstream repository: https://github.com/robopeak/rplidar_ros.git
- release repository: https://github.com/robopeak/rplidar_ros-release.git
- distro file: `jade/distribution.yaml`
- bloom version: `0.5.21`
- previous version for package: `null`
## rplidar_ros

```
1.5.1 (2016-04-12)
* Release 1.5.1.
* Add the deiver support rplidar A2.
* Contributors: yhZhao, RoboPeak Public Repos
```

PR #11212, opened 2016-04-21
Revert "rplidar_ros: 1.5.2-0 in 'jade/distribution.yaml' [bloom]"

wjwwood:
Reverts ros/rosdistro#11196


dirk-thomas:
There is no need to revert the complete repo. Just removing the version entry from the release section is sufficient.

PR #11235, opened 2016-04-22
Removed ridgeback_robot source.

tonybaltovski:


tfoote:
Why are you removing the source entry? That's important for users who want to build from source.

tonybaltovski:
I wanted to get rid of the failing build.

tfoote:
You can just turn off the devel job using  `test_commits: false`  see [rep 143](http://www.ros.org/reps/rep-0143.html) for the syntax documentation.

tfoote:
See #11183 for an example

tonybaltovski:
Will change but isn't the default suppose to be false according to REP143?

tfoote:
It's the default value, however it's overridden at the build file level to default to true. https://github.com/ros-infrastructure/ros_buildfarm_config/blob/production/kinetic/source-build.yaml#L53

PR #11238, opened 2016-04-22
python-sklearn osx entry for osx

wkentaro:


tfoote:
+1

https://pypi.python.org/pypi/scikit-learn

wkentaro:
Thank you!

PR #11244, opened 2016-04-23
ecl_tools: 0.61.4-0 in 'kinetic/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `ecl_tools` to `0.61.4-0`:
- upstream repository: https://github.com/stonier/ecl_tools.git
- release repository: https://github.com/yujinrobot-release/ecl_tools-release.git
- bloom version: `0.5.20`
## ecl_build

```
* check, but just quietly avoid including cotire if version check fails.
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -362,6 +362,25 @@ repositories:
       url: https://github.com/ros/dynamic_reconfigure.git
       version: master
     status: maintained
+  ecl_tools:
+    doc:
+      type: git
+      url: https://github.com/stonier/ecl_tools.git
+      version: indigo-stable
+    release:
+      packages:
+      - ecl_build
+      - ecl_license
+      - ecl_tools
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/yujinrobot-release/ecl_tools-release.git
+      version: 0.61.4-0
+    source:
+      type: git
+      url: https://github.com/stonier/ecl_tools.git
+      version: indigo-devel
+    status: developed
   ecto:
     release:
       tags:


jacquelinekay:
Hey @stonier, I think the source builds for `ecl_tools` and `sophus` are failing for Debian Jessie because you released using version 0.5.20 (support for Debian in the default actions was added in 0.5.21 https://github.com/ros-infrastructure/bloom/commit/5305e9749e17a2162939c4f29d2400e216fe94cb). I think if you update bloom to 0.5.21 and rerun bloom-release then the failures for Debian on the buildfarm will be fixed.

As far as the binary builds for ecl_tools on Ubuntu, those should be addressed. Not sure why it's pulling the wrong package key, could be due to an older version of rosdistro (0.4.4 instead of 0.4.5).

tfoote:
@stonier I disabled the jobs so they will not send emails every 15 minutes for reference they are here: 

http://build.ros.org/view/Ksrc_dJ/job/Ksrc_dJ__ecl_build__debian_jessie__source/
http://build.ros.org/view/Ksrc_dJ/job/Ksrc_dJ__ecl_license__debian_jessie__source/
http://build.ros.org/view/Ksrc_dJ/job/Ksrc_dJ__ecl_tools__debian_jessie__source/
http://build.ros.org/view/Ksrc_dJ/job/Ksrc_dJ__sophus__debian_jessie__source/

For whoever merges the rerelease please reenable these jobs. They will turn themselves back on during the next nightly reconfigure as well. 

For the binary failure http://build.ros.org/view/Kbin_uW64/job/Kbin_uW64__ecl_build__ubuntu_wily_amd64__binary/2/console#console-section-12

It's oddly looking for **indigo**  aka `ros-indigo-ecl-license` not **kinetic** like it should.

tfoote:
@wjwwood Do you have any thoughts why the control files would have indigo in them? ala https://github.com/yujinrobot-release/ecl_tools-release/blob/debian/ros-kinetic-ecl-tools_0.61.4-0_xenial/debian/control

And surprisingly the indigo is only in the `Depends` field the `Build-Depends` correctly depends on `ros-kinetic-catkin`

The spec files have kinetic correctly in them: https://github.com/yujinrobot-release/ecl_tools-release/blob/rpm/ros-kinetic-ecl-build-0.61.4-0_24/rpm/ros-kinetic-ecl-build.spec

tfoote:
It looks like #11247 also has the problem 

@stonier I've been able to reproduce the same symptoms by injecting a custom rosdep source on my machine for ubuntu, like this:

```
ecl_license:
  ubuntu: [ros-indigo-ecl-license]
```

Can you check that you don't have custom rosdep rules on your system? We can't introspect your sources from the Pull request, [hopefully fixed at some point](https://github.com/ros-infrastructure/bloom/issues/387)

dirk-thomas:
@stonier Please address the repeated build failures (http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__ecl_build__ubuntu_xenial_amd64__binary/) or we will remove all related jobs from the build farm for now.

stonier:
Sorry for the delay. It is flagged high on my todo list, but got haven't found a chance since our field tests started.

ETA: Mon or Tue (2nd/3rd)

dirk-thomas:
Thanks, anytime is fine. Without a comment we are just not sure if you are aware of it.

It looks like you have already addressed the problem in #11341.

dirk-thomas:
The same problem happens for the next ecl jobs: http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__ecl_config__ubuntu_xenial_amd64__binary/3/console

stonier:
Aye... They are coming

PR #11250, opened 2016-04-23
rqt_launchtree: 0.1.1-0 in 'indigo/distribution.yaml' [bloom]

pschillinger:
Increasing version of package(s) in repository `rqt_launchtree` to `0.1.1-0`:
- upstream repository: https://github.com/pschillinger/rqt_launchtree.git
- release repository: https://github.com/pschillinger/rqt_launchtree-release.git
- bloom version: `0.5.21`
## rqt_launchtree
- No changes

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -9790,6 +9790,21 @@ repositories:
       url: https://github.com/osrf/rqt_graphprofiler.git
       version: master
     status: developed
+  rqt_launchtree:
+    doc:
+      type: git
+      url: https://github.com/pschillinger/rqt_launchtree.git
+      version: master
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/pschillinger/rqt_launchtree-release.git
+      version: 0.1.1-0
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Out of curiosity: what is the difference / same compared to `rqt_launch`? Would it make sense to merge these two packages?

pschillinger:
Thanks for your fast feedback! Both plugins are related to `roslaunch`, but that's probably the only thing they have in common.

`rqt_launch` is meant to execute nodes of a launch file via GUI or separately. [What rqt_launch does NOT do](http://wiki.ros.org/rqt_launch#What_rqt_launch_does_NOT_do) is displaying the hierarchy or any structural information.

`rqt_launchtree`, on the other hand, is not meant to execute anything. It is intended to be an inspection tool for launch file configuration and hierarchy. It can be used for navigating through embedded launch files and open each of them as required in an external editor, as well as quickly searching where for example a particular parameter is defined.

I see their purpose clearly distinct and thus, would not merge these two, following your previous comment [here](https://github.com/ros-visualization/rqt/issues/57#issuecomment-13863763) and considering that `rqt_launch` is used for execution and `rqt_launchtree` rather for development.

dirk-thomas:
Thanks

PR #10874, opened 2016-03-24
New rosdep keys for libfcl-dev

scpeters:
`libccd` and `fcl` are now included in upstream package managers (Ubuntu W,X) so we don't need to package them separately for Kinetic.

Should resolve https://github.com/ros-gbp/libccd-release/issues/1

PR #8243, opened 2015-04-29
gazebo5: new rosdep keys

scpeters:
Includes arch, homebrew, and ubuntu.
The latest Fedora version is gazebo4 so that will have to wait.

Needed for gazebo_ros_pkgs jade release (https://github.com/ros-simulation/gazebo_ros_pkgs/issues/318).

@wjwwood

PR #11288, opened 2016-04-25
Add rosdep key for gazebo7 in Xenial

jackieokay:
@scpeters or @j-rivero or whoever--does this look correct to you?

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1512,8 +1512,6 @@ libgazebo5-dev:
   gentoo: [sci-electronics/gazebo]
   ubuntu: [libgazebo5-dev]
 libgazebo7-dev:
-  arch: [gazebo]
-  gentoo: [sci-electronics/gazebo]
   ubuntu: [libgazebo7-dev]
 libgeographiclib-dev:
   fedora: [geographiclib]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -808,6 +808,7 @@ gazebo:
     raring: [gazebo]
     saucy: [gazebo2]
     trusty: [gazebo2]
+    wily: [gazebo5]
     xenial: [gazebo7]
 gazebo5:
   arch: [gazebo]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -808,6 +808,7 @@ gazebo:
     raring: [gazebo]
     saucy: [gazebo2]
     trusty: [gazebo2]
+    xenial: [gazebo7]
 gazebo5:
   arch: [gazebo]
   fedora: [gazebo]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -808,7 +808,7 @@ gazebo:
     raring: [gazebo]
     saucy: [gazebo2]
     trusty: [gazebo2]
-    wily: [gazebo5]
+    wily: [gazebo7]
     xenial: [gazebo7]
 gazebo5:
   arch: [gazebo]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1511,6 +1511,11 @@ libgazebo5-dev:
   fedora: [gazebo-devel]
   gentoo: [sci-electronics/gazebo]
   ubuntu: [libgazebo5-dev]
+libgazebo7-dev:
+  arch: [gazebo]
+  fedora: [gazebo-devel]
+  gentoo: [sci-electronics/gazebo]
+  ubuntu: [libgazebo7-dev]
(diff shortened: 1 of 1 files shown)


wjwwood:
If you do it like this then you also need an entry for Wily.

jacquelinekay:
Mmkay. It looks like gazebo5 is the version in Wily: https://launchpad.net/ubuntu/wily/+source/gazebo

wjwwood:
We've already discussed the fact that a newer version will have to be back ported: https://github.com/ros-infrastructure/rep/pull/106#issuecomment-147841656

jacquelinekay:
ahh, ok thanks. I'll change it

wjwwood:
I think it has to be gazebo7 for Wily as well as Xenial.

scpeters:
We also need `libgazebo7-dev`. Should we make new keys or make the `gazebo` key point to different versions of gazebo? For jade, we made new keys (#8243). Also, I think [wily has gazebo6](http://packages.ubuntu.com/source/wily/gazebo), but @j-rivero is working on back-porting the gazebo7 from xenial to wily.

jacquelinekay:
2f39a49 adds the libgazebo7-dev key.

scpeters:
Fedora doesn't have gazebo7 yet:

* https://apps.fedoraproject.org/packages/gazebo

cc @cottsay @jpgr87

jacquelinekay:
indeed... https://bugzilla.redhat.com/show_bug.cgi?id=1247414

I propose I remove the Fedora entry and add Fedora support later. We need this to start pull request testing gazebo_ros_pkgs on the buildfarm, which doesn't build for Fedora anyway. Does that sound ok?

scpeters:
that's what we did for FCL in #10874, sounds ok to me

jacquelinekay:
Gazebo 7 isn't in Debian Jessie yet, either. Not sure what to do about that (@tfoote, is Gazebo important to you for the Turtlebot stack? I assume you're prioritizing the on-robot software...).

j-rivero:
> Gazebo 7 isn't in Debian Jessie yet,

Jessie is a released distribution so it won't be in the official distribution.

rosdep/base.yaml
```diff
@@ -1509,6 +1511,10 @@ libgazebo5-dev:
   fedora: [gazebo-devel]
   gentoo: [sci-electronics/gazebo]
   ubuntu: [libgazebo5-dev]
+libgazebo7-dev:
+  arch: [gazebo]
+  gentoo: [sci-electronics/gazebo]
```
wjwwood: I'd remove these since we can't ensure they are explicitly Gazebo 7.

wjwwood:
> Jessie is a released distribution so it won't be in the official distribution.

It could be backported or put into a ROS PPA equivalent. I'm not sure what they've done in the past.

rosdep/base.yaml
jacquelinekay: Done.

j-rivero: The Gentoo friends are updated with [gazebo versions](https://packages.gentoo.org/packages/sci-electronics/gazebo). It lives in the unstable amd64 branch (note the ~amd64 mark).

The Arch Linux friends are a bit behind but also using [gazebo7 versions](https://aur.archlinux.org/packages/gazebo/).

These distributions don't add a numbered version to the package but both have gazebo7 available.

wjwwood: Is there a way to ensure it's gazebo7 though? If you depend on `libgazebo7-dev`, then rosdep (and by translation the package manager for the OS) better give you version 7. It would be fine to have them for the `gazebo` key.

j-rivero:
> It could be backported or put into a ROS PPA equivalent. I'm not sure what they've done in the past.

Gazebo is not present into any released Debian distro. it has been present in current testing (will be the next released distro) and unstable (rolling).

rosdep/base.yaml
j-rivero: > Is there a way to ensure it's gazebo7 though?

In Gentoo, just by using the package name in the installation command, probably no. You can run `emerge =sci-electronics/gazebo=7.1.0` but this will be broken as soon as they update the package. Fine grained version election is done in other configuration files.

In Arch the situation is the same AFAIK.

jpgr87:
We can probably get Gazebo 7 in Fedora 24 (due out in early June), and definitely get it in Fedora 25+.  We stuck with Gazebo 6 in the current releases so we didn't get too far ahead of the Ubuntu packages, but if there's a case for Gazebo 7 in the F24 time frame we can try to make that happen.

PR #11298, opened 2016-04-26
add rtmros_common for jade

k-okada:


jade/distribution.yaml
```diff
@@ -4905,6 +4905,21 @@ repositories:
...
+      packages:
+      - hrpsys_ros_bridge
+      - hrpsys_tools
+      - openrtm_ros_bridge
+      - openrtm_tools
+      - rosnode_rtc
+      - rtmbuild
+      - rtmros_common
+      tags:
+        release: release/indigo/{package}/{version}
```
dirk-thomas: You can not copy-n-paste the indigo entry. You need to invoke bloom to create the Jade specific release tags and branches.

k-okada: yes, you're right, fixed and updated the PR

k-okada: since this is the first release on jade and I have already run 'git-bloom-release' manually,

PR #11318, opened 2016-04-27
Add python-googleapi to rosdep

erelson:
This add support for apiclient and can be used instead of python-oauth2client for oauth2client.

Ran the unittests as well.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -666,6 +666,17 @@ python-gdown:
 python-gi:
   debian: [python-gi]
   ubuntu: [python-gi]
+python-googleapi:
+  debian: [python-googleapi]
+  fedora: [google-api-python-client]
+  ubuntu:
+    trusty: [python-googleapi]
+    utopic: [python-googleapi]
+    vivid: [python-googleapi]
+    wily: [python-googleapi]
+    wily_python3: [python3-googleapi]
+    xenial: [python-googleapi]
+    xenial_python3: [python3-googleapi]
 python-gridfs:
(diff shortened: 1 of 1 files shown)


tfoote:
+1 

It looks like fedora has an oddly named version [google-ap-python-client](https://admin.fedoraproject.org/pkgdb/package/rpms/google-api-python-client/) that I think is the same.

erelson:
Will tweak to add that, and explains why a quick search didn't find it initially.

erelson:
Looks like that is based on just and so I think it's missing the oauth2client portion that my other entries have.  Should I still add it, or add a separate entry?

erelson:
Link I meant to include in above: https://github.com/google/google-api-python-client

vs.

http://bazaar.launchpad.net/~ubuntu-branches/ubuntu/trusty/python-googleapi/trusty-proposed/files

(I've never dealt into this stuff before, so I'm only quickly inspecting what I think is the source of each)

jacquelinekay:
It looks like in Ubuntu, python-googleapi depends on python-oauth2client, which means installing python-googleapi will bring in python-oauth2client:
http://packages.ubuntu.com/trusty/python-googleapi

Looking at some recent fedora builds, it looks like their equivalent of this package, google-api-python-client, depends on oauth2client-python:
http://koji.fedoraproject.org/koji/rpminfo?rpmID=7289755

Therefore I think adding the key `fedora: [google-api-python-client]` is fine here.

erelson:
It has been updated. Thanks for double-checking @jacquelinekay, @tfoote !

PR #11327, opened 2016-04-28
add python-qt5-bindings-gl

dirk-thomas:
- http://packages.ubuntu.com/search?keywords=python-pyqt5.qtopengl&searchon=names
- https://packages.debian.org/jessie/python-pyqt5.qtopengl
- https://www.rpmfind.net/linux/rpm2html/search.php?query=python-qt5


jacquelinekay:
+1

PR #11332, opened 2016-04-28
add python-opengl for wily and xenial

dirk-thomas:


dirk-thomas:
Since the PR doesn't add any new rosdep keys and only extents the Ubuntu distros with the same value I will go ahead and merge this. If there is any feedback I am happy to address it afterwards.

PR #11354, opened 2016-04-29
add Qt dev packages on Jessie

dirk-thomas:
This should address the following problems with Jessie (after doing debinc releases of the related repos):
- http://build.ros.org/view/Kbin_dJ64/job/Kbin_dJ64__rqt_image_view__debian_jessie_amd64__binary/3/console
- http://build.ros.org/view/Kbin_dJ64/job/Kbin_dJ64__rqt_rviz__debian_jessie_amd64__binary/2/console

While the `-dev` dependency is not create for the runtime of the resulting package the same is already the case with `libpyside2-dev` on Ubuntu.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1633,7 +1633,7 @@ python-qt4-gl:
       packages: [python-qt4-gl]
 python-qt5-bindings:
   debian:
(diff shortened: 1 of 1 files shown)


wjwwood:
+1

PR #11361, opened 2016-04-30
Add gevent to python.yaml

manacoa:
gevent is a coroutine based library for multitasking Python.

http://www.gevent.org/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -664,15 +664,9 @@ python-gdown:
     pip:
       packages: [gdown]
 python-gevent:
-  fedora:
-    apt:
-      packages: [python-gevent]
-  jessie:
-    apt:
-      packages: [python-gevent]
-  ubuntu:
-    apt:
-      packages: [python-gevent]
+  fedora: [python-gevent]
+  jessie: [python-gevent]
+  ubuntu: [python-gevent]
 python-gi:
   debian: [python-gi]
   ubuntu: [python-gi]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -665,8 +665,14 @@ python-gdown:
       packages: [gdown]
 python-gevent:
   ubuntu:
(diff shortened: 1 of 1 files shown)


tfoote:
It looks like gevent is available via debian package. By default we prefer to use the native package manager's version. 

Ubuntu: http://packages.ubuntu.com/trusty/python-gevent
It's also in Jessie https://packages.debian.org/jessie/python-gevent
And Fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/python-gevent/

I've seen several sources that suggest osx needs a combo of `libevent` from homebrew and `gevent` from pip. Which likely suggests that the generic pip rule on any platform needs libevent as a dependency since that can't be tracked in the pip exclusive dependencies.

manacoa:
Hi Tully,

I've changed the pip installs to apt-get. Looking good?

Simon

tfoote:
Thanks for switching to the apt based packages. Since we're using the default package manager it's better to use the compact representation `codename: [package]` for each. It's more compact which makes it more readable and maintainable.

rosdep/python.yaml
```diff
@@ -663,6 +663,16 @@ python-gdown:
   ubuntu:
     pip:
       packages: [gdown]
+python-gevent:
+  fedora:
+    apt:
```
tfoote: apt is not a valid installer on fedora. If you switch to the compact format in my other comment this problem will disappear.

manacoa:
How does this look?

tfoote:
The compact format looks good. Sorry I missed in the last round, but the key for debian should be `debian`  not `jessie`. (Note it will need to jump up to stay in alphabetical order too.)

manacoa:
We're getting there!

tfoote:
+1

Thanks for iterating!

SimonBirrell:
Thank you, sir!

PR #11397, opened 2016-05-04
make ubuntu rule generic for python-cwiid

tfoote:
Needed to release joystick_drivers in Kinetic for xenial and wily rule.


dirk-thomas:
The `python3*` rules are currently not defined. With this change they will be resolved to the Python 2 version of the package.

tfoote:
Talked with Dirk there's a lot of things we need to do for python3 support. Mostly in xylem but we don't have time to deploy. And this package does not appear to be packaged for python3.

PR #11413, opened 2016-05-05
add kinetic dev job for nav

mikeferguson:
working towards release, want to make sure things compile on the real OS

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1224,6 +1224,12 @@ repositories:
       url: https://github.com/ros/metapackages.git
       version: kinetic-devel
     status: maintained
+  navigation:
+    source:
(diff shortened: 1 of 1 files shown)


mikeferguson:
@dirk-thomas -- does the new farm still only reconfig once every 24 hours? I know @tfoote is trying to get turtlebot out -- but I don't have any 16.04 machines setup to test nav

dirk-thomas:
Release jobs are already being (re-)generated "immediately". devel and doc jobs are currently only regenerated once every night. I just triggered http://build.ros.org/job/Kdev_reconfigure-jobs/82/ manually - so your devel jobs should appear soon.

mikeferguson:
awesome -- thanks!

PR #11431, opened 2016-05-06
Updated libvtk-qt

matlabbe:


matlabbe:
Forgot to open a pull request after editing the file... here it is!

tfoote:
+1 matching the above changes to libvtk

PR #11435, opened 2016-05-06
add python-pyqt5.qtsvg for jessie and ubuntu

tfoote:
Re: https://github.com/ros-visualization/rqt_common_plugins/issues/373

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1638,11 +1638,11 @@ python-qt4-gl:
       packages: [python-qt4-gl]
 python-qt5-bindings:
   debian:
-    jessie: [pyqt5-dev, python-pyqt5, python-sip-dev, qtbase5-dev]
+    jessie: [pyqt5-dev, python-pyqt5, python-pyqt5.qtsvg, python-sip-dev, qtbase5-dev]
   fedora: [python-qt5, sip]
   ubuntu:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1638,11 +1638,11 @@ python-qt4-gl:
       packages: [python-qt4-gl]
 python-qt5-bindings:
   debian:
-    jessie: [pyqt5-dev, python-pyqt5, python-sip-dev, qtbase5-dev]
+    jessie: [pyqt5-dev, python-pyqt5, python-pyqt5.qtsvg, python-sip-dev, qtbase5-dev]
   fedora: [python-qt5, sip]
   ubuntu:
-    wily: [libpyside2-dev, libshiboken2-dev, pyqt5-dev, python-pyqt5, python-pyside2, python-sip-dev, shiboken2]
-    xenial: [libpyside2-dev, libshiboken2-dev, pyqt5-dev, python-pyqt5, python-pyside2, python-sip-dev, shiboken2]
```
dirk-thomas: Is this also a problem on Wily and Xenial? I thought it was only a problem on Jessie?

tfoote: This package is not installed on xenial either. It failed for me on jessie, as I assumed I was the first testing on jessie I filed it as such. But when reviewing the dependencies, it was not installed on xenial either.

PR #11441, opened 2016-05-07
Rosdep keys Update

usiraj:
Fixed rosdep keys for indigo build from source for ubuntu xenial.


vrabaud:
OpenCV3 is now the default in Kinetix/Xenial so we probably don't want to add a rosdep key, cf http://wiki.ros.org/opencv3

usiraj:
Okay. Thanks. Didn't thought about Kinetix.

PR #11443, opened 2016-05-07
Add python-skimage-pip entry to 'rosdep/python.yaml'

wkentaro:
There is already pip: [scikit-image] entry for ubuntu.
So maybe it is better to remain the current installation target.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1983,15 +1983,22 @@ python-six:
     xenial: [python-six]
     xenial: [python-six]
 python-skimage:
+  debian: [python-skimage]
+  fedora: [python-scikit-image]
+  osx:
+    pip:
+      packages: [scikit-image]
+  ubuntu:
+    trusty: [python-skimage]
+    wily: [python-skimage]
+    xenial: [python-skimage]
+python-skimage-pip:
   osx:
     pip:
       packages: [scikit-image]
   ubuntu:
     pip:
       packages: [scikit-image]
-python-skimage-apt:
-  ubuntu:
-    trusty: [python-skimage]
 python-sklearn:
   fedora: [python-scikit-learn]
   osx:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1989,6 +1989,9 @@ python-skimage:
   ubuntu:
     pip:
       packages: [scikit-image]
+python-skimage-apt:
+  ubuntu:
+    trusty: [python-skimage]
 python-sklearn:
   fedora: [python-scikit-learn]
   osx:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2014,12 +2014,8 @@ python-skimage:
       pip:
         packages: [scikit-image]
     trusty: [python-skimage]
-    utopic:
-      pip:
-        packages: [scikit-image]
-    vivid:
-      pip:
-        packages: [scikit-image]
+    utopic: [python-skimage]
+    vivid: [python-skimage]
     wily: [python-skimage]
     xenial: [python-skimage]
 python-skimage-pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2014,6 +2014,12 @@ python-skimage:
       pip:
         packages: [scikit-image]
     trusty: [python-skimage]
+    utopic:
+      pip:
+        packages: [scikit-image]
+    vivid:
+      pip:
+        packages: [scikit-image]
     wily: [python-skimage]
     xenial: [python-skimage]
 python-skimage-pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2004,15 +2004,9 @@ python-skimage:
     precise:
       pip:
         packages: [scikit-image]
-    quantal:
-      pip:
-        packages: [scikit-image]
-    raring:
-      pip:
-        packages: [scikit-image]
-    saucy:
-      pip:
-        packages: [scikit-image]
+    quantal: [python-skimage]
+    raring: [python-skimage]
+    saucy: [python-skimage]
     trusty: [python-skimage]
     utopic: [python-skimage]
     vivid: [python-skimage]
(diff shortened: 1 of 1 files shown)


tfoote:
We've converged on our typical policy is to have FOO  and FOO-pip for the pip version.

We had the `python-scikit` previously pointing to pip, but I'd suggest that it would be worth converging toward our standard. Duplicating the existing rule to be called `python-skimage-pip` and changing the original to use apt. It will potentially effect anyone relying on the newer versions provided by pip, but they could move to the new rule. There's nothing yet released relying on the specific version as you can't release against pip versions. 

For more pure backwards compatibility we could make the new version keyed with the upstream name `python-scikit-image` which also matches the fedora name, but I'd suggest trying to stay as close to the debian name as possible for better long term clarity. 

What do others think?

For reference versions of scikit-image 
pip: https://pypi.python.org/pypi/scikit-image
ubuntu: http://packages.ubuntu.com/trusty/python-skimage
debian: https://packages.debian.org/jessie/python-skimage
fedora: https://apps.fedoraproject.org/packages/python-scikit-image/overview/

wkentaro:
I agree with your idea. I updated the commit.

tfoote:
+1

dirk-thomas:
@tfoote What is the policy once we stop building for EOL Ubuntu distro? Are we reducing the set for bloom releases then too? If yes, this should be fine but this would need to be announced. If no, then the entries for older Ubuntu distro should be added otherwise a package dependending on this can't be released into Jade or older.

tfoote:
It makes sense to allow bloom to continue w/o support for the older platforms. One challenge is that we want to make sure that all the source toolchains continue to work. If we naively strip some of the definitions they may stop working. I would assume that we'd announce this sort of thing with the announcement of stopping the builds. We have several things to work out about how to disable the builds, such as hosting multiple versions of the sourcedebs. 

For this issue it's not relevant since the package was only pip before anyway so it's strictly an improvement.

wkentaro:
Hmm. So how can I fix this issue?
Should I refrain from adding apt repository for a while?

tfoote:
We just need one other maintainer to approve from @ros/rosdistro-maintainers

dirk-thomas:
@tfoote shouldn't entries for the missing Ubuntu distros be added before merging this? Since there was no change since my previous comment I would guess the question still applies.

tfoote:
Yes, it would be good to fill in the other distros to allow releasing into debian packages. This pull request does not change the ability to release since it was not releasable before. @wkentaro if you're interested in releasing packages with these you should add coverage of the relevant target ubuntu distros.

wkentaro:
@dirk-thomas Oh, sorry. I found I didn't understand your point in [the last comment](https://github.com/ros/rosdistro/pull/11443#issuecomment-217936979).
You mean maybe I should consider other Ubuntu distros (ex. precise), did you? I will update the commit.

wkentaro:
@tfoote 

> @wkentaro if you're interested in releasing packages with these you should add coverage of the relevant target ubuntu distros.

What exactly should I do? I don't under stand 'add coverage of ...'.

tfoote:
To be used as a dependency in indigo we need debian based rules for saucy and trusty, and for Jade we need trusty, utopic, and vivid. And Kinetic will need Wily and Xenial. As defined in [REP3](http://www.ros.org/reps/rep-0003.html) You've added the older rules which can be helpful for backwards compatibility, but when I say add coverage I mean fill in the missing rules for those distros.

wkentaro:
>  but when I say add coverage I mean fill in the missing rules for those distros.

Maybe I'm missing something, but what am I missing?

tfoote:
In 824f1cc I don't see utopic or vivid rules.

wkentaro:
> In 824f1cc I don't see utopic or vivid rules.

Oops, I added them. thanks.

dirk-thomas:
Yes, currently you could only use the key to release for Kinetic (Ubuntu W and X). The Trusty entry is not sufficient for Jade or Indigo. So adding S, U and V would be good imo.

dirk-thomas:
Why are the new Ubuntu entries using pip? Are you sure that there are no debian packages in e.g. U and V?

dirk-thomas:
With pip entries you can still not use this key to release into Indigo and Jade.

wkentaro:
> Why are the new Ubuntu entries using pip? Are you sure that there are no debian packages in e.g. U and V? With pip entries you can still not use this key to release into Indigo and Jade.

How can I check whether it is released on apt?
The package is not listed in packages.ubuntu.com, because it is removed.

dirk-thomas:
You can e.g. browse the apt server for these distros and check if the package is available or start a Docker container / VM for these distros and try to install the package.

wkentaro:
In vivid, I found the `python-skimage` on apt. (checked with docker)
But in utopic the packages are removed, so I got 404.
http://archive.ubuntu.com/ubuntu/dists/

dirk-thomas:
The you need to check http://old-releases.ubuntu.com/

wkentaro:
Thanks, I confirmed that there is `python-skimage` for utopic.
And I updated the commit.

dirk-thomas:
What about Saucy and the other Ubuntu distros which currently have PIP entries?

wkentaro:
Checked and updated.

dirk-thomas:
+1 from me (even if imo the older Ubuntu entries before Saucy are not necessary since we don't support this anymore). I will leave it to @tfoote to reconfirm his vote since the patch has changed significantly since then.

tfoote:
Yeah, the older distros are probably unnecessary since they're EOL. But the extra info won't really hurt. 

@wkentaro thanks for iterating with us.

wkentaro:
Thank you.

2016年6月1日水曜日、Tully Foote<notifications@github.com>さんは書きました:

> Merged #11443 <https://github.com/ros/rosdistro/pull/11443>.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/ros/rosdistro/pull/11443#event-677463720>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AEHFk4wWfiU_ie3mndgPGRyu3Q-ZJIQnks5qHIZzgaJpZM4IZh-c>
> .
>


-- 
和田 健太郎 / Kentaro Wada
http://wkentaro.com

PR #11449, opened 2016-05-08
Add OSX pip packages for python-dateutil, requests, and responses

anassinator:
| Key | URL |
| --- | --- |
| `python-dateutil` | https://pypi.python.org/pypi/python-dateutil |
| `python-requests` | https://pypi.python.org/pypi/requests |
| `python-responses-pip` | https://pypi.python.org/pypi/responses |


tfoote:
+1

PR #11456, opened 2016-05-09
camera_pose_calibration: Update branch for indigo release

RonaldEnsing:


indigo/distribution.yaml
```diff
@@ -837,16 +837,16 @@ repositories:
     doc:
       type: git
       url: https://github.com/delftrobotics/camera_pose_calibration.git
-      version: master
+      version: indigo
     release:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/delftrobotics/camera_pose_calibration-release.git
-      version: 0.1.2-0
```
dirk-thomas: The release repo does not have any 0.1.3 related branches and tags. You need to invoke `bloom-release` for that.

RonaldEnsing:
@dirk-thomas When running `bloom-release `, the version gets updated to 0.1.2-2, but it needs to be updated to 0.1.3-0. How can I fix this?

dirk-thomas:
The indigo track in the release repo does not specify a branch name (https://github.com/delftrobotics/camera_pose_calibration-release/blob/cd5e22a6f444a4de64b1b07eff32e825fd45a6df/tracks.yaml#L15). Therefore bloom will look in the default branch of the repo (currently master) where the package version is 0.1.2.

So you should:
* update this PR to only change the versions of doc and source entries
* update the branch name in the tracks.yaml in the release repo to point to the `indigo` branch
* run `bloom-release` which should then pick up the version 0.1.3 from that branch and create a PR which increases the release version

RonaldEnsing:
@dirk-thomas Thank you for the explanation. It's fixed.

PR #11476, opened 2016-05-10
add some xenial rosdeps

codebot:


dirk-thomas:
You should also add entries for *Wily*.

PR #11478, opened 2016-05-10
add some xenial rosdeps

codebot:


dirk-thomas:
LGTM. I hope nobody continues to use these keys in Wily / Xenial instead of switching to the Qt 5 ones.

dirk-thomas:
No, there are no better values. Currently several keys only have entries until Vivid (Qt 4 related) and new keys have entries starting at Wily (Qt 5 related).

tfoote:
yeah, we're kind of hitting the problem of having concurrent parallel releases that want to use different default system libraries on a given platform.

PR #11494, opened 2016-05-12
variant: 0.1.1-0 in 'indigo/distribution.yaml' [bloom]

samuelba:
Increasing version of package(s) in repository `ros-variant` to `0.1.1-0`:
- upstream repository: https://github.com/ethz-asl/ros-variant.git
- release repository: https://github.com/ethz-asl/ros-variant-release.git
- bloom version: `0.5.21`
## variant
- No changes
## variant_msgs

```
* Message variant tests successful
* Refined traits, code cleanups
* 

    * Created test package
    * Moved test binaries to test package

* Contributors: Ralf Kaestner
```
## variant_topic_test

```
* Fixes related to OS X clang compatibility
* Added message data type member access by name
* 

    * Added type name resolution for member definitions neglecting package names
    * Added receipt time to variant subscriber callback

* 

    * Code cleanups
    * Added variant publisher and subscriber
    * Added MD5 sum calculation for variant message types
    * Updated README

* Added convenience serialization methods to message class
* Added method for serializer creation to variants
* Basic serializer testing
* Added numeric value access for built-in types
* Message variant tests successful
* Message data type now separates constant and variable members
* Refined traits, code cleanups
* Removed SharedVariant and friend-relationship between Variant and CollectionVariant
* Renamed some classes, code cleanups
* 

    * Created test package
    * Moved test binaries to test package

* Contributors: Ralf Kaestner
```
## variant_topic_tools

```
* Fixes related to OS X clang compatibility
* Fixed some issues of subscriber
* Fixed bug in MessageDefinition::setMessageType() which was caused by a wrong order of defining the required message types of a message type
* Added message data type member access by name
* 

    * Added type name resolution for member definitions neglecting package names
    * Added receipt time to variant subscriber callback

* 

    * Code cleanups
    * Added variant publisher and subscriber
    * Added MD5 sum calculation for variant message types
    * Updated README

* Added convenience serialization methods to message class
* Added method for serializer creation to variants
* Basic serializer testing
* Added numeric value access for built-in types
* Message variant tests successful
* Message data type now separates constant and variable members
* Reverted templated variant value to represent actual value type
* Refined traits, code cleanups
* Removed SharedVariant and friend-relationship between Variant and CollectionVariant
* Renamed some classes, code cleanups
* 

    * Created test package
    * Moved test binaries to test package

* Contributors: Ralf Kaestner
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8902,6 +8902,26 @@ repositories:
       url: https://github.com/ros/ros.git
       version: indigo-devel
     status: maintained
+  ros-variant:
+    doc:
+      type: git
+      url: https://github.com/ethz-asl/ros-variant.git
+      version: master
+    release:
+      packages:
+      - variant
+      - variant_msgs
+      - variant_topic_test
+      - variant_topic_tools
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/ethz-asl/ros-variant-release.git
+      version: 0.1.1-0
+    source:
+      type: git
+      url: https://github.com/ethz-asl/ros-variant.git
+      version: master
+    status: developed
   rosR:
     doc:
       type: git
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8902,26 +8902,6 @@ repositories:
       url: https://github.com/ros/ros.git
       version: indigo-devel
     status: maintained
-  variant:
-    doc:
-      type: git
-      url: https://github.com/ethz-asl/variant.git
-      version: master
-    release:
-      packages:
-      - variant
-      - variant_msgs
-      - variant_topic_test
-      - variant_topic_tools
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/ethz-asl/variant-release.git
-      version: 0.1.1-0
-    source:
-      type: git
-      url: https://github.com/ethz-asl/variant.git
-      version: master
-    status: developed
   rosR:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


vrabaud:
Though I understand "variant" in programming (like Boost variant), I think it is a bad naming because of what a variant is in the ROS world: http://www.ros.org/reps/rep-0142.html
@samuelba , any feedback ?

samuelba:
I don't see a big naming conflict. Yes variant is use in ROS for something different, however it never appears in any package naming. I think the name is appropriate for this package and does not get mixed up with the ROS variant usage.

If the package name is not acceptable for you at all, what alternative name would you suggest?

samuelba:
Perhaps I should add why we what to add this package to the ROS repository. The reason is our [rqt_multiplot](https://github.com/ethz-asl/ros-rqt-multiplot-plugin) package, which is a nice solution to plot all kind of things and depends on our variant package. We think it would be useful for many ROS developers.

tfoote:
I agree that the ros-variant is a name collision and breaks our conventions re using ROS in the name. http://wiki.ros.org/ROS/Patterns/Conventions#Naming_ROS_Resources Though this is only the repo name. So renaming the repo to variant would resolve that. 

Looking at the content, it's a powerful capability, but I'm a little worried about seeing the `variant_msgs` from a general design point of view, using those will break all of our existing introspection tools and type protection. Could you document how this is designed to be used especially as compared to the approach used in [`topic_tools/ShapeShifter`](http://docs.ros.org/api/topic_tools/html/classtopic__tools_1_1ShapeShifter.html)? The `ShapeShifter` currently does not implement the accessors or serializers at the moment, but that presumably could be added in the same way w/o defining new ROS messages to contain raw ROS messages.

kralf:
ROS currently appears in the repo name to make our organization's github projects searchable for ROS-related code. Renaming it to meet your conventions should not be a problem.

Also, do not worry about the `variant_msgs` package. Those messages were majorly defined and used for debugging purposes but do not represent a "hard dependency" of the other packages. Removing them from the project can be achieved at minor effort, affecting only one interface method and the `relay` tool.

The main improvement of `variant_topic_tools/Variant` over `topic_tools/ShapeShifter` is full runtime message type introspection in C++ which naturally comes at the cost of type protection, but provides similar means of working with ROS messages in C++ as in Python, i.e., symbolic access to message members, their types, and values. For an example, see the `echo` tool which mimics the functionality of `rostopic echo`.

samuelba:
@tfoote I renamed the repository (removed ros-) and updated this pull request, but somehow travis-ci fails now. Should I do the whole bloom release process and pull request again with the renamed repository?

Are there other problems with this pull request?

tfoote:
That works for me. Thanks for updating. This is failing CI because with the rename the inserted section needs to be moved lower in the document to maintain alphabetical ordering.

samuelba:
I corrected the alphabetical order, should be ready to merge.

PR #11510, opened 2016-05-13
Add libjpeg62 as distinct from libjpeg

evenator:


tfoote:
Can you point to the packages and explain why you need libjpeg62 vs libjpeg? And are they compatible/simultaneously installable?

evenator:
I believe libjpeg62-dev causes a conflict, but libjpeg62 does not. The package in question is a proprietary external dependency that is linked against libjpeg62. I've been installing the dependency manually, but I'd like to be able to make use of rosdep.

tfoote:
We could consider adding the runtime library as a rosdep, but an important part of rosdep rules is to pick a single version of a dependency for a platform if they conflict. When we try to support multiple conflicting versions of packages we can actually get ourselves into a deadlock. (See some of the ongoing vtk issues.)

I'd suggest that this might be a good use case for setting up a custom rosdep source for interacting with that package?

PR #11537, opened 2016-05-17
fzi_icl_can: 1.0.9-0 in 'kinetic/distribution.yaml' [bloom]

fmauch:
Increasing version of package(s) in repository `fzi_icl_can` to `1.0.9-0`:
- upstream repository: https://github.com/fzi-forschungszentrum-informatik/fzi_icl_can.git
- release repository: https://github.com/fzi-forschungszentrum-informatik/fzi_icl_can-release.git
- bloom version: `0.5.21`
## fzi_icl_can

```
* reset the project name after including the IcMaker
* Contributors: Felix Mauch
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -635,6 +635,21 @@ repositories:
       url: https://github.com/introlab/find-object.git
       version: kinetic-devel
     status: maintained
+  fzi_icl_can:
+    doc:
+      type: git
+      url: https://github.com/fzi-forschungszentrum-informatik/fzi_icl_can.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/fzi-forschungszentrum-informatik/fzi_icl_can-release.git
+      version: 1.0.9-0
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


tfoote:
This appears to be failing on jessie: http://build.ros.org/view/Kbin_dJ64/job/Kbin_dJ64__schunk_canopen_driver__debian_jessie_amd64__binary/4/console

The underlying issue is a dependency on an invalid rosdep: https://github.com/ros/rosdistro/issues/11567

I'd open a ticket at https://github.com/fzi-forschungszentrum-informatik/fzi_icl_can but there is no issue tracker.

fmauch:
Yes, this has already been noticed. Right now I have no idea on how to address this issue. 

Is there a way to define platform dependent dependencies? This seems to be necessary as debian doesn't provide a generic meta package.

fmauch:
As a workaround so that builds don't always fail, could the debian build for schunk_canopen_driver be deactivated? 

Or is there a way to model optional dependencies in the package.xml The dependency is only necessary to run a certain shell script in which we could check for the kernel-headers being present...

tfoote:
We will be able to blacklist packages more effectively when we merge this: https://github.com/ros-infrastructure/ros_buildfarm_config/pull/41

PR #11541, opened 2016-05-18
add libopenni2-dev for jessie also in stretch

tfoote:
Referencing new openni packages backported on jessie: https://github.com/ros-infrastructure/reprepro-updater/commit/63237cd1b5639a51980ce1244a304b74c4220737

https://packages.debian.org/source/stretch/openni2


tfoote:
@130s @jacquelinekay this should fix openni2_camera

dirk-thomas:
+1

PR #11543, opened 2016-05-18
grasp-synergy: 0.0.1-0 in 'indigo/distribution.yaml' [bloom]

felixduvallet:
Increasing version of package(s) in repository `grasp-synergy` to `0.0.1-0`:
- upstream repository: https://github.com/felixduvallet/grasp-synergy.git
- release repository: https://github.com/felixduvallet/grasp-synergy-release.git
- distro file: `indigo/distribution.yaml`
- bloom version: `0.5.21`
- previous version for package: `null`

PR #11545, opened 2016-05-18
Adds doc and source entries for grasp_synergy package.

felixduvallet:
As suggested in #11543


tfoote:
lgtm. note that these jobs will not get generated until tomorrow night's reconfigure and the doc jobs only runevery 3 days so it may be a little while before you see the results.

dirk-thomas:
The doc and source entry should point to the upstream repo, not the release repo. The master branch of the release repo contains no ROS packages.

PR #11548, opened 2016-05-18
gennodejs: 1.0.2-0 in 'kinetic/distribution.yaml' [bloom]

IanTheEngineer:
Increasing version of package(s) in repository `gennodejs` to `1.0.2-0`:
- upstream repository: https://github.com/RethinkRobotics-opensource/gennodejs.git
- release repository: https://github.com/RethinkRobotics-release/gennodejs-release.git
- bloom version: `0.5.21`
- previous version for package: `1.0.1-0`
## gennodejs

```
* Fixed gennodejs file deployment in install space
* Fixed a python3 incompatibility in generate.py
* Contributors: Chris Smith, Ian McMahon, Maarten de Vries
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -756,7 +756,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/RethinkRobotics-release/gennodejs-release.git
-      version: 1.0.1-0
+      version: 1.0.2-0
     source:
       type: git
       url: https://github.com/RethinkRobotics-opensource/gennodejs.git


dirk-thomas:
Could it be that the new version installs the same file for every message package (`share/gennodejs/find.js`)? That is not going to work since every message message will collide: e.g. http://build.ros.org/job/Kbin_dJ64__rosgraph_msgs__debian_jessie_amd64__binary/4/console

On a first look this might be the problematic change: https://github.com/RethinkRobotics-opensource/gennodejs/commit/ad70ebbc0d8f384fd66f4c3e36a0f93fcb4e1b41

IanTheEngineer:
cc @chris-smith Gotcha. Yes, you are completely correct. This was our workaround for the linked devel space in catkin_tools, and works on devel and install builds. Neither of those fail for me, but it appears the buildfarm has stricter standards for collision (understandably). I'll look into this and get you a fix asap

dirk-thomas:
The problem is that every file you install will end up in the Debian package and multiple Debian packages can't install the same file. So it is a limitation of packaging the code for any platform (same for Fedora, etc.).

PR #11570, opened 2016-05-20
Add python-colorama entry for osx

wkentaro:


tfoote:
+1 https://pypi.python.org/pypi/colorama

Issue #11488, opened 2016-05-11
Resolve Consistent Qt Versions on Kinetic


tfoote:
This is coming out of the ongoing discussion in: https://github.com/ros/rosdistro/pull/11240 however that's a closed PR so can't be reopened. 

Other tickets related include: 
https://github.com/ros/rosdistro/pull/10740
https://github.com/ros-perception/perception_pcl/issues/44#issuecomment-198082216
https://github.com/ros/rosdistro/pull/10747
https://github.com/ros/rosdistro/issues/10752
https://github.com/ros/rosdistro/pull/11095
https://github.com/ros/rosdistro/pull/11097
https://github.com/ros/rosdistro/pull/11240

The high level problem is that we're building libraries which are unrunnable because they're linked against both Qt4 and Qt5. We also have run into problems with mutually exclusive rosdep which cause rosdep to fight back and forth installing conflicting dependencies. 

Complicating this is that based on Xenial we want to have Qt5 be the default version since it's overdue to be deployed and that's what our LTS target supports. 

## Jessie: 
Available: 
* [libqt4-dev ](https://packages.debian.org/jessie/libqt4-dev)
* [qtbase5-dev](https://packages.debian.org/jessie/qtbase5-dev)

Common libraries:
* [libvtk5-dev](https://packages.debian.org/jessie/libvtk5-dev)  depends on [libqt4-dev](https://packages.debian.org/jessie/libqt4-dev)
* [libvtk6-dev](https://packages.debian.org/jessie/libvtk6-dev)  depends on [libqt4-dev](https://packages.debian.org/jessie/libqt4-dev)
* [libpcl-dev](https://packages.debian.org/jessie/libpcl-dev) depends on libvtk5 depends on libqt4-dev
* libgazebo7-dev depends on libqt4-dev

ROS libraries:
* opencv3 Qt5
* rviz Qt5
* rqt Qt5

## Wily:

Very similar to Jessie...
* [vtk[5]](http://packages.ubuntu.com/source/wily/vtk) depends on qt4
* [vtk6](http://packages.ubuntu.com/source/wily/vtk6) depends on qt4
...

## Xenial: 
Available: 
* [libqt4-dev ](http://packages.ubuntu.com/xenial/libqt4-dev)
* [qtbase5-dev](http://packages.ubuntu.com/xenial/qtbase5-dev)

Common libraries:
[...]

PR #11593, opened 2016-05-21
add libqwt-qt5-dev for ubuntu xenial

samuelba:
Rqt in Kinetic is based on QT5, this means rqt plugins using QWT must use the QT5 version of `libqwt` which is `libqwt-qt5`. Otherwise, a `Segmentation fault (core dumped)` occurs at the start of the rqt plugin.

gdb output:

``` bash
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fffadb565a4 in ?? () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
```

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2346,6 +2346,9 @@ libqwt6:
     raring: [libqwt-dev]
     saucy: [libqwt-dev]
     trusty: [libqwt-dev]
+    utopic: [libqwt-dev]
+    vivid: [libqwt-dev]
+    wily: [libqwt-qt5-dev]
     xenial: [libqwt-qt5-dev]
 libqwtplot3d-qt4-dev:
   arch: [qwtplot3d]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2346,6 +2346,7 @@ libqwt6:
     raring: [libqwt-dev]
     saucy: [libqwt-dev]
     trusty: [libqwt-dev]
+    xenial: [libqwt-qt5-dev]
 libqwtplot3d-qt4-dev:
   arch: [qwtplot3d]
   debian:


tfoote:
For Kinetic support please add rules for wily and jessie as well. 

connected to #11488

dirk-thomas:
As far as I know there are no Python Qwt 6.1 packages available. So this won't make the plugins work.

samuelba:
at least C++ rqt plugins using qwt (like our [rqt_multiplot](https://github.com/ethz-asl/rqt_multiplot_plugin/tree/kinetic-devel)) work with libqwt-qt5. I don't know about python rqt plugins. @dirk-thomas there is a python qwt implementation working with PyQt5 -> [PythonQwt](https://github.com/PierreRaybaut/PythonQwt), however, I didn't test it.

dirk-thomas:
I was referring to an Ubuntu package which we could depend on. +1 foradding the C++  only entry (with the comments from @tfoote).

samuelba:
I added the qt5 rule for wily and the qt4 rules for utopic and vivid. However, the libqwt-qt5-dev doesn't exist on jessie, there seems to be a [package for stretch](https://packages.debian.org/stretch/libqwt-qt5-dev) but not for jessie.

dirk-thomas:
I already gave my +1 above. This is pending a second review from the @ros/ros_team.

tfoote:
For jessie then we should put in the older version https://packages.debian.org/jessie/libqwt-dev matching the older ubuntu distros.

jacquelinekay:
Wouldn't that mean that qt plugins on Jessie that use libqwt will segfault?

tfoote:
Yeah, I guess we could either create a separate rule for qt5 vs qt4 qwt or just leave it off for now. Or consider a backport into our repos. Otherwise we'll need to blacklist anything using this for debian.

jacquelinekay:
I'm ambivalent about Jessie since it's not one of our "core" platforms; I'll leave it up to Tully to decide what to do about that. +1 from me to add libqwt-qt5 to Xenial.

tfoote:
Ok, we'll go w/o the rule for now. Any package that depends on this will need to be blacklisted for debian jessie. If there's demand we can consider trying to generate a backport.

PR #11637, opened 2016-05-25
Added pytz and tzlocal python packages to rosdep

juhsuanhua:
This adds two python packages for dealing with time conversion between time zones.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1553,6 +1553,10 @@ python-pytest:
   debian: [python-pytest]
   fedora: [python-pytest]
   ubuntu: [python-pytest]
+python-pytz-pip:
+  ubuntu:
+    pip:
+      packages: [pytz]
 python-pyudev:
   debian: [python-pyudev]
   fedora: [python-pyudev]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1553,6 +1553,10 @@ python-pytest:
   debian: [python-pytest]
   fedora: [python-pytest]
   ubuntu: [python-pytest]
+python-pytz:
+  ubuntu:
+    pip:
+      packages: [pytz]
```
wjwwood: Can you please use the convention where you add `-pip` to the end of the key, since you're specifying a pip dependency on Ubuntu. The reason is so that if a Ubuntu package for `pytz` exists or becomes available the `python-pytz` would be available for it to use.

Thanks.

juhsuanhua:
Comment addressed. Thanks!

wjwwood:
Lgtm, but it needs a second review before merging.

tfoote:
+1

If anyone's interested there are ubuntu packages for these as well: http://packages.ubuntu.com/xenial/python-tz and http://packages.ubuntu.com/xenial/python-tzlocal as of xenial

PR #11653, opened 2016-05-27
Add entry for python-mysqldb.

patrickcjh:


tfoote:
Could you add a rule for fedora `python-mysql`

For reference:
Ubuntu package: http://packages.ubuntu.com/xenial/python-mysqldb
Debian package: https://packages.debian.org/jessie/python-mysqldb
Fedora package: https://admin.fedoraproject.org/pkgdb/package/rpms/python-mysql/

For OSX it looks a little more complicated. I think we need to depend on `mysql` from brew and use the pip `MySQL-python` package  [thread 1](https://stackoverflow.com/questions/7375199/python-brew-and-mysqldb) [thread 2](https://stackoverflow.com/questions/17599830/installing-mysql-python-on-mac) We could leave this for now, but I wanted to comment in case someone found this ticket later and wanted to add support.

commaster90:
Added in the entry for fedora. Thanks for the reference.

tfoote:
Thanks for the fixup.

+1

PR #11699, opened 2016-06-01
opencv_apps: 1.11.13-0 in '{indigo,jade}/distribution.yaml'

k-okada:
@vrabaud can we remove oepncv_apps from this and re-release from new repo? or should we keep indigo/jade as it is and start from kinetic?


k-okada:
humm @vrabaud seems busy these days, I think this pr is ok to merge, remove opencv_apps from vision_opencv and add new entity.

vrabaud:
very busy indeed. It got removed indeed. Thx

Issue #11687, opened 2016-05-31
dev-libs/collada-dom and media-libs/collada-dom in gentoo


benkinooby:
Greetings,

I tried to install ROS kinect on gentoo but decided to deviate from the official install guide [1][2] occasionally. Most notably I decided against using ros-overlay since all relevant packages seem to be present in portage.

After running
`rosinstall_generator desktop --rosdistro kinetic ...`

the next command fails as follows (other packages were installed too but are not displayed here)
`$ rosdep install --from-paths src --ignore-src --rosdistro kinetic`
`executing command [sudo -H emerge -a media-libs/collada-dom]`
`Password:` 
`These are the packages that would be merged, in order:`
`Calculating dependencies... done!`
`emerge: there are no ebuilds to satisfy "media-libs/collada-dom".`
`emerge: searching for similar names...`
`emerge: Maybe you meant any of these: dev-libs/collada-dom, media-libs/opencollada, media-libs/ladspa-sdk?`
`ERROR: the following rosdeps failed to install`
`portage: command [sudo -H emerge -a media-libs/collada-dom] failed`

After some investigation I found out that `media-libs/collada-dom` is contained in ros-overlay (seems it was created at a point of time where no such packages existed in portage) so I wonder if `dev-libs/collada-dom` could be added to base.yaml now that portages provides it and allow an overlay-free install alltogether (unless there is more to the overlay's existence apart from historical reasons).

Kind regards

Additional information:
William Woodall added the code in question [3] by merging it to base.yaml from gentoo.yaml [4]

[1] http://wiki.ros.org/kinetic/Installation/Gentoo
[2]
- setting python's default version to 2.7 is not recommended and should be covered by gentoo's python-wrapper
- it uses layman which is practically deprecated by the extended functionality of portage (via repos.conf)
- it recommends to run pip install as root which will mess with portage

[3] https://github.com/ros/rosdistro/blame/master/rosdep/base.yaml
[4] https://github.com/ros/rosdistro/blame/52a9adacd4cdc62de793ac1c0f58fa98315b3cc9/rosdep/gentoo.yaml

tfoote:
@allenh1 thoughts

allenh1:
[...]

PR #11724, opened 2016-06-02
make rosdistro accept collada-dom from gentoo's portage

benkinooby:
dev-libs/collada-dom is provided by gentoos package management system portage.
media-libs/collada-dom is provided by ros-overlay [1] which is used in the official (although experimental) install instructions for ROS on gentoo [2].

Allowing ros to install without the over will ease/eable the transition of gentoo's ROS-support from overlay to its native repository, especially since the overlay is suspected to be redundant.

relevant issue:
https://github.com/ros/rosdistro/issues/11687

[1] https://github.com/ros/ros-overlay
[2] http://wiki.ros.org/kinetic/Installation/Gentoo (also for older ROS distros)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -494,7 +494,7 @@ collada-dom:
   fedora: [collada-dom-devel]
   gentoo:
     portage:
-      packages: [media-libs/collada-dom]
+      packages: [media-libs/collada-dom, dev-libs/collada-dom]
   macports: [collada-dom]
   ubuntu:
     lucid: [collada-dom-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -494,7 +494,7 @@ collada-dom:
   fedora: [collada-dom-devel]
   gentoo:
     portage:
-      packages: [media-libs/collada-dom, dev-libs/collada-dom]
+      packages: [dev-libs/collada-dom]
   macports: [collada-dom]
   ubuntu:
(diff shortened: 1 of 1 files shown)


benkinooby:
If multiple packages are listed, they are _AND_ related [1] [2], therefore **this pull request should not be accepted at its current state**. Changing to dev-libs/collada-dom without while keeping maximum compatibility to the ros-overlay (if that is favourable) would require
- fixing the version of dev-libs/collada-dom to 2.4.0 (as this is the only version available in the overlay)
- transfer/merge of dependencies RDEPEND [3] as applicable

sources:
[1] http://docs.ros.org/independent/api/rosdep/html/rosdep_yaml_format.html
[2] http://www.ros.org/reps/rep-0111.html
[3]
overlay
`RDEPEND="
    dev-libs/boost
        dev-libs/libxml2
        virtual/opengl
        media-libs/freeglut
        >=dev-util/cmake-2.6.0
        >=media-gfx/nvidia-cg-toolkit-2.0
"
`
portage
`RDEPEND="
        dev-libs/boost:=
        sys-libs/zlib:=[minizip]
        dev-libs/libxml2
        dev-libs/libpcre[cxx]
"
`

tfoote:
@allenh1 FYI

benkinooby:
Closing this pull request since the latest change would break the old/outdated install instructions (which users probably still follow)

PR #11764, opened 2016-06-06
Fix mrpt dep

jlblancoc:
Deletion of special case in old (now unsupported Ubuntu releases) due to a bug in those packages. 
Now all versions should depend on libmrpt-dev only. Simpler = better.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2999,15 +2999,7 @@ mplayer:
 mrpt:
   debian: [libmrpt-dev]
   fedora: [mrpt-devel]
-  ubuntu:
-    oneiric: [libmrpt-dev]
-    precise: [libmrpt-dev]
-    quantal: [libmrpt-dev]
-    raring: [libmrpt-dev, mrpt-apps]
-    saucy: [libmrpt-dev, mrpt-apps]
-    trusty: [libmrpt-dev]
-    utopic: [libmrpt-dev]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Won't this break building on Saucy from source?

jlblancoc:
Yes Dirk. But watching the [timeline graph](https://upload.wikimedia.org/wikipedia/en/timeline/98f453cc0c7950a3382ec330a2941bba.png) of official support by Canonical, it seemed to me that Saucy is sort of deprecated, isn't it? 

I could redo the PR by just adding a new exception for the newer releases, but at some point in the future it will be easier to drop really old distributions for simplicity... it's up to you!

dirk-thomas:
I will let @tfoote decide. Usually we try to not break existing users if possible with reasonable effort.

tfoote:
It is EOL, but from our download statistics that definitely does not mean that people are not using it. And the rosdep rules are even more important then since people will be building from source. (Debian packages have the resolved rules baked into them). 

We don't have an official policy but I'd suggest that we at least keep rules all intermediates between supported LTS releases.

If there was something new or otherwise we could consider it, but since this is just a "cleanup" I'd rather not. The cost of keeping the extra lines in the rules file is pretty low. So I'd suggest keeping the more verbose form for now.

jlblancoc:
:+1:  Closing. Superseded by #11771

PR #11771, opened 2016-06-07
Fix mrpt deps in Ubuntu

jlblancoc:
This PR supersedes #11764


dirk-thomas:
+1

tfoote:
Thanks for iterating!

PR #11772, opened 2016-06-07
enable pull requests for navigation branches

mikeferguson:
I've also added ros-pull-request-builder as a collaborator on the repo -- although apparently someone must accept the invite?

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -6319,6 +6319,8 @@ repositories:
       url: https://github.com/ros-gbp/navigation-release.git
       version: 1.12.10-0
     source:
+      test_commits: false
+      test_pull_requests: true
       type: git
       url: https://github.com/ros-planning/navigation.git
       version: indigo-devel
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -2939,6 +2939,8 @@ repositories:
       url: https://github.com/ros-gbp/navigation-release.git
       version: 1.13.1-0
     source:
+      test_commits: false
+      test_pull_requests: true
       type: git
       url: https://github.com/ros-planning/navigation.git
       version: jade-devel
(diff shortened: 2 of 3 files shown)


mikeferguson:
@tfoote -- you mentioned something about disabling devel jobs -- but it seems that test_commits is default to false based on REP-143 -- do I need to add anything else to this?

mikeferguson:
@130s -- we're hitting the 10k line limit on pretty much all of the navigation Travis builds -- if we can use the ROS buildfarm instead, I will be disabling the Travis CI.

tfoote:
test_commits is false by default in the distro file, but it's overridden in the build file. [here](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/master/kinetic/source-build.yaml#L52)

mikeferguson:
sounds good -- updated

130s:
@mikeferguson at https://github.com/ros/rosdistro/pull/11772#issuecomment-224408921
> we're hitting the 10k line limit on pretty much all of the navigation Travis builds -- if we can use the ROS buildfarm instead, I will be disabling the Travis CI.

Sure. +1 for prioritizing what's provided by ros.org.

There already seems [a PR](https://github.com/ros-planning/navigation/pull/478) that was opened after this got merged but I haven't found the CI result. Where can we see it?

dirk-thomas:
Jenkins wasn't able to establish the webhooks. The log says:

> Couldn't create web hook for repository ros-planning/navigation. Does the user (from global configuration) have admin rights to the repository?

mikeferguson:
Ah -- I had only given it "write" permissions -- just bumped it to "admin" -- is there some way to retrigger the setup and test that it works?

dirk-thomas:
It would happen automatically on nightly reconfiguration. I manually save the job config of one of the PR jobs and the webhook was established correctly: see https://github.com/ros-planning/navigation/settings/hooks

You should be able to use the common phrase to retrigger builds:

> retest this please

This should be possible for all members of the org unit. Usually I mention the ros-pull-request-builder in the beginning so that it is clear what the comment is for.

mikeferguson:
Awesome -- thanks @dirk-thomas -- that is working now

PR #11784, opened 2016-06-09
Added python-flask-cors to rosdep/python.yaml

LeonChambers:


tfoote:
Since this looks like a pip specific package could you append `-pip` to the rosdep key? That will leave the bare name available in the case that it is packaged into upstream packages. 

For others reference the package is here on pypi: https://pypi.python.org/pypi/Flask-Cors

tfoote:
Thanks for the update.

+1 from me

PR #11786, opened 2016-06-10
Updating libboost-random from 1.46 to 1.46.1 for ubuntu precise

jonyal:


dirk-thomas:
Can you please explain why you think this dependency (which only affects EOL distributions of ROS) should be updated.

JonatanAlcocer:
We have a complete architecture running on Ubuntu 12.04 with projects which depends on libboost-random, and they have upgraded from 1.46 to 1.46.1. In order to still using rosdep system to install our dependencies we need to upgrade this one.

We are currently working to upgrade to Ubuntu 14.04 & ROS indigo but we need to pass some tests before.

dirk-thomas:
I will leave it to @tfoote to consider the advantage over the potential risk of breakage.

tfoote:
+1 

I checked that `libboost-random-dev` -> `libboost-random1.46-dev` -> `libboost-random1.46.1`

I think either this was never working or that there was a security update which is why they removed the older version.  

I did some testing in a docker instance. Selected output here: https://gist.github.com/tfoote/4743de227177c41a89f56000c1342316 And validated that the current target is "installable" because apt uses a regex to turn it into the new name. And rosdep fails to verify the install because apt is aliasing to a new package.

JonatanAlcocer:
Yes, that's the reason because I changed the current key of libboost-random, because they have removed the older version and we can't install all of our external dependencies with the rosdep system. I don't think to change the current key of base.yaml could break the installation, furthermore, if we didn't change the current key, this bug will continue to happen for developers who use libboost-random library.

PR #11802, opened 2016-06-11
Add python-fcn-pip entry

wkentaro:


tfoote:
Since this is pip only, for future compatibility please add the suffix `-pip` to the key. That way if it's released upstream in the future we can give maintainers the option to depend on the pip version or they can switch to the packaged version for releasing.

wkentaro:
I added the -pip suffix.

tfoote:
+1

PR #11806, opened 2016-06-12
Update rosdep with Kinetic dependencies

cberner:


rosdep/python.yaml
```diff
@@ -893,7 +897,7 @@ python-mako:
   ubuntu: [python-mako]
 python-mapnik:
   fedora: [mapnik-python]
-  ubuntu: [python-mapnik2]
+  ubuntu: [python-mapnik]
```
jacquelinekay: Does this need to be separated out into the different supported Ubuntu versions? Or is `python-mapnik` actually used on all versions, not just Xenial?

cberner: `python-mapnik` should work for all versions. `python-mapnik2` only exists for Trusty and there it's just a transitional package that depends on `python-mapnik`

jacquelinekay: Great.

jacquelinekay:
+1

tfoote:
lgtm python-mapnik in ubuntu: http://packages.ubuntu.com/trusty/python/python-mapnik

PR #11836, opened 2016-06-15
We no longer support agvs in Hydro

carlos3dx:
We deleted agvs entry in hydro/distribution.yaml cause we no longer suppor it in hydro

### DIFF
--- a/hydro/distribution.yaml
+++ b/hydro/distribution.yaml
@@ -122,16 +122,6 @@ repositories:
       url: https://github.com/atenpas/agile_grasp.git
       version: hydro
     status: maintained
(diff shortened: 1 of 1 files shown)


vrabaud:
Hydro is eol-ed anyway so you change will not be taken into account. Your commit also does not modify the ditribution.yaml

carlos3dx:
Sorry, there was an error and the change was not commited properly,

tfoote:
Our policy is that we do not update EOL'd distros and they are kept as last released for archival purposes. There is no expectation of maintenance of a package in an EOL rosdistro.

PR #11841, opened 2016-06-15
add pyper-pip to rosdep/python.yaml

furushchev:
This Pull Request adds `pyper` provided in `PyPI` to `rosdep/python.yaml`.

`pyper` is a library for python which provides an interface between python and R language.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -48,6 +48,10 @@ paramiko:
     vivid: [python-paramiko]
     wily: [python-paramiko]
     xenial: [python-paramiko]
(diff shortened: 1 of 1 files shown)


tfoote:
We've started using the `-pip` suffix for pip only packages. To allow future addition of upstream packaging w/o conflicting. Could you add that? And also comment here with a link to the package description for others reference?

furushchev:
@tfoote Thank you very much for reviewing! I updated my commit and information of PR.

tfoote:
+1

PR #11843, opened 2016-06-15
parrot_arsdk: remove the released tag, add the devel job

mani-monaj:
There are some issues regarding the release that I need to fix, so I removed the `release` tag and added the `source` repo. Could you please tell me how I can re-release the same version using `bloom` when  problems are fixed (i.e. `3.9.1-0` to `3-9.1-1`)?

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -7263,6 +7263,10 @@ repositories:
         release: release/indigo/{package}/{version}
       url: https://github.com/AutonomyLab/parrot_arsdk-release.git
       version: 3.9.1-0
+    source:
+      type: git
+      url: https://github.com/AutonomyLab/parrot_arsdk.git
+      version: indigo-devel
     status: developed
   patrolling_sim:
     doc:


dirk-thomas:
You don't have to remove the release entry for this. You can keep it and as soon as you make a new release with bloom it will just update the version number.

Since you should **not** modify a tag in your upstream repository you should simply release a 3.9.2 once you have fixed the problem. Changing the code for an already tagged version and moving the tag is considered bad practice.

mani-monaj:
Thank you for your comments.

> You don't have to remove the release entry for this. You can keep it and as soon as you make a new release with bloom it will just update the version number.

I wanted to do this to stop build failures on the farm. If that's not a big issue, I will leave it as is.

> Since you should not modify a tag in your upstream repository you should simply release a 3.9.2 once you have fixed the problem. Changing the code for an already tagged version and moving the tag is considered bad practice.

Fair enough. The issue is I wanted to track the upstream versioning scheme and now that I _pre-maturely_ released this version, I am stuck with it. Any ideas?

dirk-thomas:
Since only the binarydeb fails once a day that is not such a big issue if you plan to fix it. A sourcedeb would fail every 15 minutes and we would unused the version attribute to make it stop.

Well, if you want to track the upstream version than you might just change the tag. Your call. In that case you can just run `bloom-release` after you have updated the tag with the same version in your upstream repo and it will do a so called debian increment (3.9.1-N).

mani-monaj:
Thanks Dirk for your help. I've added back the release tag, squashed the
changes and updated this PR. I am working on a fix.

dirk-thomas:
Great, thanks.

PR #11848, opened 2016-06-15
Revert "Revert "Avoid demanding PCL head.""

wjwwood:
@mikepurvis please let me know when upstream has merged pcl 1.8.0 and then I can merge this one.

Reverts ros/rosdistro#11847

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -265,26 +265,8 @@ libpcl-all:
       depends: [libpcl-all-dev]
 libpcl-all-dev:
   osx:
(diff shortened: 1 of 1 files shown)


mikepurvis:
Upstream is merged.

I've rebuilt indigo desktop full using the bottled vtk (sans `--with-qt`), a local build of what will be the PCL 1.8.0 bottle, and Gazebo 7. All seems sane, so I think you're good to merge.

My next step here is to nuke my whole homebrew install and do some end-to-end testing to see where my brewed python install can be further simplified (https://github.com/mikepurvis/ros-install-osx/pull/41)

PR #11862, opened 2016-06-16
Added two new packages to indigo/distribution.yaml

carlos3dx:


carlos3dx:
Added two new packages to indigo/distribution.yaml

dirk-thomas:
The patch needs to be rebased. Also the change contains a weird change to rosdistro.

rosdistro
```diff
@@ -0,0 +1 @@
+Subproject commit 0a330e0714773b20bf033ea0937596ea8bb9226f
```
tfoote: This part of the commit should not be here. It looks like metadata from an instance of submodule that shouldn't be there.

PR #11905, opened 2016-06-21
[rosdep] update java key for jessie

mikaelarguedas:


tfoote:
openjdk-7-jdk is in debian wheezy -> experimental https://packages.debian.org/experimental/openjdk-7-jdk

openjdk-6-jdk is only in wheezy sid and experimental: https://packages.debian.org/experimental/openjdk-6-jdk

This change upgrades jdk on wheezy. It's never been working on jessie and jessie is the first one we're actively supporting. So overall I think it's worth upgrading for all targets not trying to keep the older version on wheezy. If someone needs the old version we can consider adding back a specific key for openjdk-6-jdk.

+1

mikaelarguedas:
@ros/ros_team can someone else review this please ? Thanks!

PR #11911, opened 2016-06-21
Add debian and fedora rosdep keys for libgeos++

evenator:


dirk-thomas:
+1

tfoote:
It looks like on Fedora it's `geos-devel` not `libgeos-devel`

Debian: https://packages.debian.org/jessie/libgeos++-dev
Fedora: https://apps.fedoraproject.org/packages/geos-devel

tfoote:
This also reports to be unmergeable. I suspect it's based on an out of date master branch and needs a rebase against the latest master.

evenator:
Whoops, I must have been looking at the wrong site. I'll fix and rebase.

PR #11983, opened 2016-06-26
Add nodejs-legacy package to nodejs dep

cberner:
Debian and Ubuntu versions after Precise renamed node binary to
nodejs, and provided the node symlink as a separate package. This fixes
many node packages that rely on the node binary, and makes it compatible
with Fedora

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3057,6 +3057,10 @@ nodejs:
   debian: [nodejs]
   fedora: [nodejs]
   ubuntu: [nodejs]
+nodejs-legacy:
+  debian: [nodejs-legacy]
+  fedora: [nodejs]
+  ubuntu: [nodejs-legacy]
 npm:
   debian: [npm]
   fedora: [npm]


tfoote:
That seems reasonable. Can you link to documentation of this change?

If we're switching to calling out each version we should list all intermediate versions not just trusty, wily, and xenial as we'll break anyone using those intermediate distros.

tfoote:
Toward keeping things simpler it might make sense to introduce a new key with this functionality. Anyone who needs it can depend on both the main version and the legacy version. Presumably at some point the legacy one will be dropped and if we integrate it into our definition then we will have to do another transition at that point. Whereas if it's dropped at some point those packages will know their dependency is no longer there and need to port forward. 

Since it's bundled in fedora the `nodejs-legacy` key can simply resolve to `nodejs`

cberner:
Yep: https://nodejs.org/en/download/package-manager/

"A Node.js package is also available in the official repo for Debian Sid (unstable), Jessie (testing) and Wheezy (wheezy-backports) as "nodejs". It only installs a nodejs binary.
The nodejs-legacy package installs a node symlink that is needed by many modules to build and run correctly. The Node.js modules available in the distribution official repositories do not need it."

cberner:
Sure, I can split it into a separate dependency. I'm not sure it will be dropped though. It sounds like the Debian community decided to change the name of the binary from node to nodejs, but upstream it's still named node.
I'm fine with either though, so just let me know which you prefer to do

cberner:
Updated to have a separate dependency

cberner:
@tfoote let me know if you'd prefer it the other way

tfoote:
Thanks I think this is cleaner. 

+1

PR #11991, opened 2016-06-27
Add dependencies useful for tracing program execution and kernel info

iluetkeb:
This pull request adds dependencies for using the Linux Trace Toolkit next generation (LTTng) for both C++ and Python processes. LTTng enables very low-overhead run-time data gathering.

FYI: We (Bosch) have a ROS package in the process of open source release which uses these.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1799,6 +1799,10 @@ libleptonica-dev:
 liblinear-dev:
   fedora: [liblinear-devel]
   ubuntu: [liblinear-dev]
+liblttng-ust-dev:
+  debian: [liblttng-ust-dev]
+  ubuntu: [liblttng-ust-dev]
+  arch: [lttng-ust]
 libmagick:
   arch: [imagemagick]
   debian: [libmagick++-dev]
@@ -2835,6 +2839,14 @@ log4cxx:
   opensuse: [liblog4cxx-devel]
   rhel: [log4cxx-devel]
   ubuntu: [liblog4cxx10-dev]
+lttng-modules:
+  debian: [lttng-modules-dkms]
+  ubuntu: [lttng-modules-dkms]
+  fedora: [lttng-devel]
+  arch: [lttng-modules]
+lttng-tools:
+  ubuntu: [lttng-tools]
+  arch: [lttng-tools]
 lua-dev:
   arch: [lua]
   debian: [liblua5.1-0-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -262,6 +262,9 @@ avrdude:
     portage:
       packages: [dev-embedded/avrdude]
   ubuntu: [avrdude]
+babeltrace:
+  debian: [babeltrace]
+  ubuntu: [babeltrace]
 bazaar:
   arch: [bzr]
   debian: [bzr]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -264,6 +264,7 @@ avrdude:
   ubuntu: [avrdude]
 babeltrace:
   debian: [babeltrace]
+  fedora: [babeltrace]
   ubuntu: [babeltrace]
 bazaar:
   arch: [bzr]
@@ -1823,6 +1824,7 @@ liblinear-dev:
 liblttng-ust-dev:
   arch: [lttng-ust]
   debian: [liblttng-ust-dev]
+  fedora: [lttng-ust]
   ubuntu: [liblttng-ust-dev]
 libmagick:
   arch: [imagemagick]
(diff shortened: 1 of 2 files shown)


rosdep/base.yaml
```diff
@@ -1817,6 +1820,10 @@ libleptonica-dev:
 liblinear-dev:
   fedora: [liblinear-devel]
   ubuntu: [liblinear-dev]
+liblttng-ust-dev:
+  debian: [liblttng-ust-dev]
+  ubuntu: [liblttng-ust-dev]
+  arch: [lttng-ust]
```
tfoote: arch should be before debian to keep it in alphabetical order.

rosdep/base.yaml
```diff
@@ -2896,6 +2903,14 @@ log4cxx:
   opensuse: [liblog4cxx-devel]
   rhel: [log4cxx-devel]
   ubuntu: [liblog4cxx10-dev]
+lttng-modules:
+  debian: [lttng-modules-dkms]
+  ubuntu: [lttng-modules-dkms]
+  fedora: [lttng-devel]
+  arch: [lttng-modules]
```
tfoote: Same alphabetical issue here and 3 lines below

rosdep/python.yaml
```diff
@@ -892,6 +894,11 @@ python-libpgm-pip:
   ubuntu:
     pip:
       packages: [libpgm]
+python2-lttngust:
```
tfoote: It's not usually useful to have different keys for different platforms. The goal of having these keys is that they give the same thing on different platforms. I assume you did this because it's not available in python3 on arch? Maybe a pip rule would be better?

Can you explain a little more about your use case?

iluetkeb: Unfortunately, babeltrace is not on pypi. I surmise that's because it's built as part of a 'C' project.

It gets worse in that python3-babeltrace is not usually available on Ubuntu either, because it's on a PPA.

I'll just remove it.

iluetkeb: Yeah, a pip rule would probably be better.

rosdep/base.yaml
```diff
@@ -2896,6 +2903,14 @@ log4cxx:
   opensuse: [liblog4cxx-devel]
   rhel: [log4cxx-devel]
   ubuntu: [liblog4cxx10-dev]
+lttng-modules:
+  arch: [lttng-modules]
+  debian: [lttng-modules-dkms]
+  fedora: [lttng-devel]
```
tfoote: I think this rule might be invalid as on Fedora it says you need to compile the the modules from source: https://lttng.org/download/#fedora

tfoote:
I find python3-babletrace in Wily and later: http://packages.ubuntu.com/wily/python3-babeltrace

**Could you add a debian rule for python3-babeltrace.** It's available as of jessie: https://packages.debian.org/jessie/python3-babeltrace

For others looking into this:
babletrace:
debian as of jessie: https://packages.debian.org/jessie/babeltrace
ubuntu all supported platforms: http://packages.ubuntu.com/trusty/babeltrace
Fedora all platforms: https://admin.fedoraproject.org/pkgdb/package/rpms/babeltrace/

**Could you add fedora too for babeltrace**


linblttng-ust-dev: 
ubuntu: all platforms: http://packages.ubuntu.com/trusty/liblttng-ust-dev
debian all platforms: https://packages.debian.org/jessie/liblttng-ust-dev

From: https://lttng.org/download/#fedora
Fedora: lttng-tools and lttng-ust

**Could you fill in the lttng for fedora**

iluetkeb:
Tully, thanks a lot for the feedback and the hints! The three new commits should address your comments.

rosdep/base.yaml
```diff
@@ -2896,6 +2905,14 @@ log4cxx:
   opensuse: [liblog4cxx-devel]
   rhel: [log4cxx-devel]
   ubuntu: [liblog4cxx10-dev]
+lttng-modules:
+  arch: [lttng-modules]
+  debian: [lttng-modules-dkms]
+  ubuntu: [lttng-modules-dkms]
+lttng-tools:
+  arch: [lttng-tools]
+  fedora: [lttng-tools]
```
tfoote: One last comment it looks like we're missing a debian rule here.

iluetkeb:
btw, I've added the Python stuff with a distribution spec now, because it is only available in relatively recent versions of debian and ubuntu. However, this means I'll have to update it every time there is a new version of debian and/or ubuntu out.

is there a way to specify a rule as "for this version of a distribution and all later ones"?

rosdep/python.yaml
```diff
@@ -218,6 +218,14 @@ python-babel:
     vivid: [python-babel]
     wily: [python-babel]
     xenial: [python-babel]
+python-babeltrace:
+  debian:
+    jessie_python3: [python3-babeltrace]
```
tfoote: The `<CODENAME>_python3` targets are mostly an artifact of our testing with a customized environment where python3 is changed to the default interpreter. In this case I think that the top level rule should be `python3-babeltrace` and the CODENAMES should be just the standard targets as the libraries are installable side by side.

tfoote:
+1 with squash

dirk-thomas:
@tfoote You can squash during the merge.

PR #11996, opened 2016-06-27
add python-twisted-core in vivid and utopic

jihoonl:


tfoote:
While we're filling this in could you add utopic too?

jihoonl:
Added. :wink:

dirk-thomas:
+1

PR #11997, opened 2016-06-27
librealsense: 0.9.2-0 in 'kinetic/distribution.yaml' [bloom]

mdhorn:
Increasing version of package(s) in repository `librealsense` to `0.9.2-0`:
- release repository: https://github.com/intel-ros/realsense.git
- bloom version: `0.5.21`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1540,6 +1540,21 @@ repositories:
       url: https://github.com/ros-gbp/libg2o-release.git
       version: 2016.4.24-0
     status: maintained
+  librealsense:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/intel-ros/realsense.git
+      version: 0.9.2-0
+    source:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: master
+    status: developed
   linux_peripheral_interfaces:
     doc:
       type: git


jacquelinekay:
Looks like the binary jobs fail because CMake cannot find `libusb.h`.

mdhorn:
There is a dependency in there got the libusb-dev package.

The problem truly appears to be the build could not find 'pkg-config' which is used in the Third Party Makefile. It is a tool made in backquotes in the Makefile that I missed as a dependency.

I'm trying to upstream a new package.xml which adds:
`<build_depend>pkg-config</build_depend>`

What I'm not clear on, is once this change is upstreamed, will Jenkins automatically try to rebuild the packages? Or is there something I need to manually do to force a build again?

jacquelinekay:
If you make a change to the package.xml, you will have to tag and re-release another version of the package for Jenkins to start building the new package; the buildfarm won't automatically pull changes from your upstream repo.

mdhorn:
Can I rerun bloom-release to build the version in the distribution.yaml
from 
`version: 0.9.2-0`
to 
`version: 0.9.2-1`
or just update the distribution.yaml file manually and submit a PR?

I'm hoping I don't have to ask the third party package owner to tag a new release.

tfoote:
Bloom has imported the tagged version into the release repo. You can apply a patch to the upstream package in the release repo and rerun bloom to do a debian increment.

Though I'd recommend using the kinetic directory and overlaying the corrected one. 

Just put the fixed version of package.xml in a `kinetic` directory on master. 

An example of doing that is here:
reference to use overlay directory: https://github.com/ros-gbp/opencv3-release/blob/master/tracks.yaml#L58
file: https://github.com/ros-gbp/opencv3-release/blob/master/kinetic/package.xml

I believe anything in that directory will be inserted or knockout content in the upstream.

mdhorn:
Currently I've upstreamed the package.xml and the CMakeLists.txt files for the build.

I was following the recommendation of upstreaming the changes from http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty

@tfoote If I were to add the overlay, would that not always override what is in the upstream?
Is the overlay preferred over upstream?

tfoote:
Upstreaming is definitely preferred. Following your preference of not forcing upstream to make a patch release right nowm I was recommending using the overlay in the intermediate time before they make their next release to get the right metadata into the repository. Once they've made their next release with the dependency fixed you can remove the overlay.

PR #12035, opened 2016-06-30
enable pull-request building for turtlebot and turtlebot_apps

tfoote:


130s:
Can anyone answer to [this question](http://answers.ros.org/question/239060/usage-of-test_commit-and-test_pull_requests-in-rosdistro-file/) regarding how to use these tags (that seem so useful!)?

PR #12069, opened 2016-07-05
repository_name: 0.0.1-0 in 'indigo/distribution.yaml' [bloom]

onionsflying:
Increasing version of package(s) in repository `repository_name` to `0.0.1-0`:
- upstream repository: https://github.com/hans-x-lab/hansagv_description
- release repository: https://github.com/hans-x-lab/hansagv_description.git
- distro file: `indigo/distribution.yaml`
- bloom version: `0.5.21`
- previous version for package: `null`
## hansagv_description

```
* version 1.0
* Contributors: Cong Liu
```

PR #12076, opened 2016-07-06
repository_name: 0.0.1-1 in 'indigo/distribution.yaml' [bloom]

onionsflying:
Increasing version of package(s) in repository `repository_name` to `0.0.1-1`:
- upstream repository: ~/github/src/hansagv_description
- release repository: https://github.com/hans-x-lab/hansagv_description.git
- bloom version: `0.5.21`
## hansagv_description

```
* version 1.0
* Contributors: Cong Liu
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8711,7 +8711,7 @@ repositories:
       url: https://github.com/So-Cool/report_card.git
       version: master
     status: maintained
-  repository_name:
+  hansagv_description:
     release:
       packages:
       - hansagv_description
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Same comment as https://github.com/ros/rosdistro/pull/12069#discussion_r69577309 and please close obsolete pull requests.

indigo/distribution.yaml
```diff
@@ -8711,7 +8711,7 @@ repositories:
       url: https://github.com/So-Cool/report_card.git
       version: master
     status: maintained
-  repository_name:
+  hansagv_description:
```
tfoote: This is the right name. When there is only one package with the same name as the repository at the root, you should remove the list of packages as the above conditions are assumed if there is no other information. And if it's assumed we don't want to restate the information. So you should use this version but remove lines 8716 and 8717

PR #12143, opened 2016-07-12
python-flask for xenial

tulku:


tfoote:
Could you add utopic, vivid, and wily in the middle too to fill out the definitions? otherwise it looks good. 

This packages looks to be on all platforms: http://packages.ubuntu.com/xenial/python-flask

tulku:
@tfoote updated with your comments. Do not have any installation of those ubuntu versions to test it though :(

Thanks!

PR #12153, opened 2016-07-13
rb1 common & sim packages added to kinetic 

AliquesTomas:


kinetic/distribution.yaml
```diff
@@ -2597,6 +2597,46 @@ repositories:
...
+      type: git
+      url: https://github.com/RobotnikAutomation/rb1_base_common.git
+      version: indigo-devel
+    release:
+      packages:
+      - rb1_base_common
+      - rb1_base_description
+      - rb1_base_pad
+      tags:
+        release: release/indigo/{package}/{version}
```
dirk-thomas: You can't just copy the entry from Indigo but must run bloom for Kinetic.

PR #12166, opened 2016-07-15
Add python module readchar to ubuntu

bajo:
Adds the ability to get keyboard input without waiting for the enter key being pressed. For details see https://pypi.python.org/pypi/readchar

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1737,6 +1737,10 @@ python-rdflib:
     raring: [python-rdflib]
     saucy: [python-rdflib]
     trusty: [python-rdflib]
+python-readchar:
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


tfoote:
If you're just targeting the pip package please add `-pip` to the key. In the future when/if debian or others packages it we'll want to use the name `python-readchar` and if we use it for pip it's much harder to transition.

bajo:
Sure, that make sense. 
Sorry for not following the convention in the first place.

tfoote:
+1

bajo:
@tfoote Thanks for the +1. Do you know who I should ping to get this merged?

bajo:
@dirk-thomas Thank you for merging my pr.  :)

PR #12183, opened 2016-07-18
Add arduino-core rosdep for Debian.

mikepurvis:


dirk-thomas:
Thanks. Since the PR is only adding the missing Debian entry I will merge it without an additional +1.

PR #12210, opened 2016-07-21
I added the marker_msgs also for the previous ros verions jade and hydro

maxbader:


dirk-thomas:
Hydro has been EOL for a long time and we don't run any jobs for our anymore. Therefore we don't accept changes to the Hydro rosdistro file. Please update this PR to only change Jade.

maxbader:
Thanks for the information, I removed the changed on the hydro distribution.yaml

PR #12233, opened 2016-07-25
Update release tag of parrot_arsdk for Indigo from 3.9.1-5 -> 3.9.1-6

mani-monaj:


wjwwood:
Out of curiosity, why not let bloom open the pull request? did you run into some sort of problem with it?

mani-monaj:
@wjwwood I was using bloom for the first time on a new machine and by mistake answered _No_ when I was asked by bloom to create the `OAUTH` tokens. That's why I had to create the PR manually. For subsequent releases (#12234 and #12235), I fixed the issue and let bloom create the PR.

I think it would have been nicer if bloom could detect this issue earlier in the process or would have offered me to create the PR later.

wjwwood:
Ok. You can always abort the process and rerun bloom with `--pull-request-only` if something happens.

mani-monaj:
Thank you. That is definitely good to know. Unfortunately I was not aware of this feature and the final error (warning) message that was displayed by bloom does not give a hint about this option.

PR #12268, opened 2016-08-03
Added python-tabulate for Xenial

fsuarez6:


tfoote:
Can you link to the upstream package sources for easier review? And why did you only add it for wily and xenial?

fsuarez6:
What do you mean with 'link to upsteam package sources'?
I added it only for Wily and Xenial because are the only ones [available](https://launchpad.net/ubuntu/+source/python-tabulate).

tfoote:
Thanks that's the sort of link we're looking for. Often we use the packages.ubuntu.com since it is easier to find other versions etc. Like http://packages.ubuntu.com/xenial/python-tabulate 

We need references for review. Otherwise reviewers need to search and dig things up. And when and if there's a conflict down the line we want to be able to refer back to this ticket to understand how to resolve issues.

I'll note it's also in debian as of stretch: https://packages.debian.org/stretch/python-tabulate
It also looks to be in Fedora as of 23: https://admin.fedoraproject.org/pkgdb/package/rpms/python-tabulate/

So, I'd recommend adding a fedora rule and a debian stretch rule as well since they're available.

fsuarez6:
> We need references for review. Otherwise reviewers need to search and dig things up. And when and if there's a conflict down the line we want to be able to refer back to this ticket to understand how to resolve issues.

Noted

>So, I'd recommend adding a fedora rule and a debian stretch rule as well since they're available.

Done

tfoote:
+1

PR #12291, opened 2016-08-06
Add rosdep key for xsltproc.

mikepurvis:


mikepurvis:
Context here is wanting to transform URDF at build time in ways that aren't possible for xacro. Some OSes bundle xsltproc in the libxslt package, but Ubuntu has it separate, hence a separate key for Ubuntu and Debian.

The other way to handle this could be just adding xsltproc to the existing libxslt key.

dirk-thomas:
+1

PR #12304, opened 2016-08-09
add iproute2

furushchev:
The package `iproute2` provides `ip` command, which will be standard than `ifconfig`.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1129,6 +1129,13 @@ ipmitool:
     portage:
       packages: [sys-apps/ipmitool]
   ubuntu: [ipmitool]
+iproute2:
+  arch: [iproute2]
+  debian: [iproute2]
+  fedora: [iproute]
+  gentoo: [iproute2]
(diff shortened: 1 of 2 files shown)


tfoote:
For others reference a good article on ifconfig vs ip: https://tty1.net/blog/2010/ifconfig-ip-comparison_en.html also some good usage here: https://serverfault.com/questions/458628/should-i-quit-using-ifconfig

Ubuntu: http://packages.ubuntu.com/xenial/iproute2
Debian: https://packages.debian.org/jessie/iproute2
Arch: https://www.archlinux.org/packages/core/x86_64/iproute2/
Gentoo: https://packages.gentoo.org/packages/sys-apps/iproute2

On fedora it's only iproute: https://admin.fedoraproject.org/pkgdb/package/rpms/iproute/

furushchev:
@tfoote Thank you for reviewing. I updated to use `iproute` for fedora.

tfoote:
+1

PR #12306, opened 2016-08-09
adding python-zmq

asmodehn:


asmodehn:
somehow I didnt see it was already there...

PR #12314, opened 2016-08-10
Adding python-cryptography and python-apparmor

ruffsl:
Adding keys for python packages:
python-cryptography
python-apparmor

Didn't find these packages for fedora though...


tfoote:
+1 

https://packages.debian.org/jessie/python-apparmor
http://packages.ubuntu.com/xenial/python-apparmor


Fedora appears to use SELinux only instead of apparmor 
https://ask.fedoraproject.org/en/question/77230/why-does-fedora-use-selinux-rather-than-apparmor/

That will be a problem for releasing this into core packages without fedora equivalents.

PR #12316, opened 2016-08-10
Adding python-openssl

ruffsl:
Forgot to also add python-openssl with previous PR #12314

https://packages.debian.org/jessie/python-openssl
http://packages.ubuntu.com/xenial/python-openssl

Agian, not sure what to do for fedora:
https://ask.fedoraproject.org/en/question/27845/how-to-install-python-openssl/


tfoote:
+1

PR #12321, opened 2016-08-10
reviewing python-six key for ubuntu

asmodehn:


tfoote:
http://packages.ubuntu.com/xenial/python-six

+1

PR #12341, opened 2016-08-12
[dep][ubuntu] Add pandoc.

130s:
`pandoc` seems to have been consistently called pandoc for at least the last few Ubuntu distros:
- [Trusty](http://packages.ubuntu.com/trusty/text/pandoc)
- [Wily](http://packages.ubuntu.com/wily/pandoc) (EoL but just as fyi)
- [Xenial](http://packages.ubuntu.com/xenial/pandoc)


dirk-thomas:
+1

PR #12343, opened 2016-08-12
adding python-click for xenial

asmodehn:


tfoote:
Debian starting with stretch: https://packages.debian.org/stretch/python-click
Ubuntu starting with xenial only: http://packages.ubuntu.com/xenial/python-click
Fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/python-click/ 

Are you hoping to package against this in Kinetic? If so because of the missing Wily version it might make sense to do the same as #12344 instead of this.

asmodehn:
Oh I didn't realise that missing willy would mean we need a kinetic
package...

I am in the process of releasing a bunch of packages for indigo and jade
now. I jist do the rosdep keys while I am at it.

I ll get a look at kinetic after that then...

asmodehn:
It makes sense to do a third party release for kinetic.
But it also makes sense to have it in rosdep, so that : 
- we can support different version than system version if needed
- in any ROS package using it, we can quickly switch from the thirdparty package dependency to the system package dependency. So we can stop maintaining a release repo as soon as it is no longer necessary.

rosdep/python.yaml
```diff
@@ -373,13 +373,18 @@ python-clearsilver:
...
+    sid: [python-click]
+    stretch: [python-click]
   fedora: [python-click]
   osx:
     pip:
       packages: [click]
   ubuntu:
     pip:
       packages: [click]
+    xenial: [python-click]
```
tfoote: I don't believe rosdep supports mixed level rules. With this change the pip rule should be defined for all the older platforms.

asmodehn:
Actually thinking about this, we probably want the rosdep key for pip being different, like having a `-pip` suffix no ? But that would mean changing the meaning of that `python-click` rosdep key...

I am not quite sure about the status of rosdep pip support, or the extent to which we want to support it... Is there any reference documentation about this ?

Should I define all distros for that python-click key, and declare a pip entry in each of them ? seems to be unneeded redundancy...

tfoote:
There's not a strict policy. Most things are on a case by case basis and our best practices are evoloving. The use of the `-pip` extension we've started doing more recently than the original `python-click` entry. If it was to come in now we'd request it to be done differently. However, since it's already there we'd rather not change it. Using the pip rule for the older versions and the debian for the newer is probably the best path forward. Eventually the pip rules can be phased out once the older platforms are EOL.

tfoote:
+1

dirk-thomas:
Is this key used for any Indigo package? If yes, it should have Saucy and Utopic entries, too.

asmodehn:
Not that I can think of... but how can I be sure ? I there a way to check
all package's package.xml for a distro (without installing them) ?

tfoote:
Unfortuantely there's no good way to search, especially as there's potentially an unknown number of private repos which cannot be searched. 

There is an alternative to use the indigo packaged version: https://github.com/ros/rosdistro/pull/12344/files but it appears to be not [building on saucy](http://build.ros.org/view/Ibin_uS64/job/Ibin_uS64__click__ubuntu_saucy_amd64__binary/) due to catkin_pip [not building due to too low a version of pip](http://build.ros.org/view/Ibin_uS64/job/Ibin_uS64__catkin_pip__ubuntu_saucy_amd64__binary/77/console). 

It's EOL, but we might as well add the rule for Saucy and Utopic and move on. Hopefully we'll develop a deprecation/removal process for rosdeps in the future.

asmodehn:
Moving on then... this should be good to merge now.

rosdep/python.yaml
```diff
@@ -373,13 +373,31 @@ python-clearsilver:
     saucy: [python-clearsilver]
     trusty: [python-clearsilver]
 python-click:
+  debian:
+    sid: [python-click]
```
tfoote: We prefer not to define sid rules since sid's a rolling release they end up out of date very quickly.

tfoote:
+1 thanks for the iterations

asmodehn:
Coming back to this topic after I am trying to release a package that depend on it, for indigo and jade.

I have this package.xml : https://github.com/asmodehn/rostful/blob/python_mutate/package.xml#L21
that declares a dependency on "click".
This currently resolves to a third party package that I released here : https://github.com/asmodehn/click-rosrelease. I am going to release on kinetic a third party package soon, because wily doesn't have it, so it is needed (wily is already EOL btw...)

To release this package on future (kinetic++) ROS releases (or even on kinetic in case of xenial platform) I want to switch to use the system package and not the third party ROS released package anymore. What is the best way to do this ?

- I cannot have a second package.xml declaring a different set of dependencies for kinetic++ without having to maintain multiple development branches.
- I could create a patch for kinetic++, but that means a user getting the source might get dependency problems (rosdep trying to resolve "click" from the unpatched package.xml from source)
- I could create a third party release for kinetic++, but I would duplicate the system dependency without a proper reason (the system version already matches my requirements), and that may create problems (two similar package with 2 different build process - we don't want to duplicate all debs in main|universe|multiverse).
- I cannot declare a "click" package for rosdep without duplicating the entry for the same package (unless there is an "alias" function in rosdep I am not aware of ?)

I am thinking the best way would be arguments in package.xml format, but I couldn't find any for this purpose...

Thanks for the advice !

tfoote:
When you switch to the system dependency the rosdep rule name should be different from the ROS package. Usually 'click' becomes 'libclick-dev' for c++ or 'python-click' for python packages. Trying to use the same name almost always ends up with. 

The recommendation is to fork your repo and if you maintain things in parallel, cherry-picking things to the parallel development branches isn't too hard.

If the package is available upstream it's **strongly** recommended not to also release it as a ros package due to potential conflicts.

asmodehn:
OK I ll go with multiple branches in my repos then. I kept the names different (removed the "python-" prefix) so things should be fine (albeit a bit cumbersome).

Just a small comment about
`If the package is available upstream it's strongly recommended not to also release it as a ros package due to potential conflicts.`

Agreed but packages on old distros are sometimes too old to be used.
And when people need an update they go the simplest way and we end up with something worse : a package installing another one without warning, like this : https://github.com/RobotWebTools/rosbridge_suite/issues/162. I would be surprised if it was an isolated case, but afaik the buildfarm does not check for this...

So I think it would be very useful to have an easy way to specify distributions in dependency declaration in package.xml. It would allow users to use an alternative third party version for specific distros whenever needed, and use upstream system package whenever possible.

tfoote:
Yeah, there are lots of things that could be added to the package.xml. We spent a lot of time stabilizing the current version and and making it as simple as possible. If you'd like to propose an extension we can look at it. But in the end it's going to be a balance between adding complexity to support this use case vs making the whole system more complex for everyone.

PR #12369, opened 2016-08-16
adding python-certifi

asmodehn:
similar to click package : python-certifi is available in ubuntu since xenial.
I am also building third party release for indigo ( and up ...).


tfoote:
Ubuntu: http://packages.ubuntu.com/xenial/python-certifi
Debian has it as of stretch: https://packages.debian.org/stretch/python-certifi it would be good to add that rule.
Fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/python-certifi/

tfoote:
+1

PR #12370, opened 2016-08-16
adding python-tz on ubuntu

asmodehn:
seems to be pytz on fedora


tfoote:
This looks like all the ubuntu tags are the same. Please just collapse that into one non-specific rule. That's much more maintainable.

tfoote:
Also the package looks to be available on all debian targets: https://packages.debian.org/jessie/python-tz

Fedora link: https://admin.fedoraproject.org/pkgdb/package/rpms/pytz/

PR #12371, opened 2016-08-16
certifi: 2015.11.20-0 in 'indigo/distribution.yaml' [bloom]

asmodehn:
Increasing version of package(s) in repository `certifi` to `2015.11.20-0`:
- upstream repository: https://github.com/certifi/python-certifi.git
- release repository: https://github.com/asmodehn/certifi-rosrelease.git
- bloom version: `0.5.21`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1136,6 +1136,13 @@ repositories:
       url: https://github.com/asmodehn/catkin_pip.git
       version: indigo-devel
     status: developed
+  certifi:
+    release:
+      tags:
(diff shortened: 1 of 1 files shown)


tfoote:
This looks good picking the same version as what's packaged instead of latest. Note that we generally don't want to turn into an entire linux distro repackaging everything, but with this as a way to backport while waiting for coverage upstream it seems reasonable.

tfoote:
Note: related PR for the rosdep entry: https://github.com/ros/rosdistro/pull/12369

PR #12394, opened 2016-08-17
remove libgstreamer-plugins-ugly1.0

trainman419:
It doesn't actually exist in Ubuntu Trusty or Xenial: http://packages.ubuntu.com/source/xenial/gst-plugins-ugly1.0


tfoote:
Thanks for the fixup to #12356

PR #12400, opened 2016-08-17
wts_driver: 1.0.1-0 in 'indigo/distribution.yaml' [bloom]

ksatyaki:
Increasing version of package(s) in repository `wts_driver` to `1.0.1-0`:
- upstream repository: https://github.com/ksatyaki/wts_driver.git
- release repository: https://github.com/ksatyaki/wts_driver-release.git
- bloom version: `0.5.21`
## wts_driver

```
* First version
* Contributors: Chittaranjan Srinivas Swaminathan, Chittaranjan Swaminathan Srinivas
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -13545,6 +13545,17 @@ repositories:
       url: https://github.com/RIVeR-Lab/wpi_jaco.git
       version: develop
     status: maintained
+  wts_driver:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/ksatyaki/wts_driver-release.git
+      version: 1.0.1-0
+    source:
+      type: git
+      url: https://github.com/ksatyaki/wts_driver.git
+      version: 1.0.1
+    status: developed
   wu_ros_tools:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


ksatyaki:
Now that one pull request is open, is it possible for me to add the package to other distros (jade, etc) without having to use bloom?

dirk-thomas:
No, `bloom` generates the Debian control files for each specific Ubuntu (and Debian) distribution. Each has potentially different dependency names etc. So for every release to every ROS distribution you need to run it again. But the next run of `bloom` will be a bit easier since it will be smart enough to suggest you information based on the existing entry in Indigo.

dirk-thomas:
The failed sourcedeb jobs were due to a recent change on the buildfarm which didn't work on Saucy. The jobs should pass by now. Sorry for the notification emails.

ksatyaki:
Now should I delete the branch? Also, the build is failing because of some std::cout lines that were meant for debugging but were wrongly left un-commented. I should commit the corrections to the [wts_driver ](https://github.com/ksatyaki/wts_driver)repository and that should solve it right?

dirk-thomas:
The latest commits to the master branch of your repo fixed the *devel* job (http://build.ros.org/job/Idev__wts_driver__ubuntu_trusty_amd64/3/) which is being triggered for new commits.

For the *sourcedeb* job is based on your released version of the package. In order to make that use the latest changes you need to make a new release in your repo (update version number and tag it) and then run bloom again (which should ask much less for a repeated release) which will create a pull request like this which bumps the version number here. Then the *sourcedeb* job will use that new version and will likely pass.

PR #12416, opened 2016-08-19
[rosdep] add new key `python-reportlab` for ubuntu

NikolausDemmel:


tfoote:
You have a syntax error, it looks like you're missing a colon. 

Please link to the upstream packages for contributed rosdep rules: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

Also if we could add rules for debian and fedora at least that would be great.

dirk-thomas:
Closing due to no response.

PR #12437, opened 2016-08-22
transforms3d: 0.3.0-3 in 'kinetic/distribution.yaml' [bloom]

jgoppert:
Increasing version of package(s) in repository `transforms3d` to `0.3.0-3`:
- upstream repository: https://github.com/matthew-brett/transforms3d
- release repository: https://github.com/dronecrew/transforms3d-release.git
- bloom version: `0.5.21`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -4524,6 +4524,13 @@ repositories:
       url: https://bitbucket.org/traclabs/trac_ik.git
       version: no_moveit_plugin-kinetic
     status: developed
+  transforms3d:
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/dronecrew/transforms3d-release.git
+      version: 0.3.0-3
+    status: developed
   turtlebot:
     doc:
       type: git


jgoppert:
This is basically a wrapper python library for transofmrations.py that I am using in a lot of projects and I would like to make sure it is easily deploy-able on my target machines with ROS. Let me know if this is the standard approach or if I am doing something incorrect as this is my first attempt to use bloom.

jgoppert:
The main question I  have is the build_type tag, from the docs I don't see what the expected options are, this is just a standard python packages, so I put python, but that being correct seems doubtful after searching the net a bit more.

https://github.com/dronecrew/transforms3d-release/blob/master/kinetic/package.xml#L16

tfoote:
Bloom has extension points for different build types, but pure python is not one currently implemented. I think this answer is still current: http://answers.ros.org/question/63972/release-non-cmake-3rd-party-libraries-with-bloom/ There are several projects using the ExternalProject_Add successfully.

jgoppert:
@tfoote Thanks for the advice, external project add might work for me.  Here is what I ended up hacking together for now, hopefully upstream will just pull it in.

https://github.com/dronecrew/transforms3d/blob/master/CMakeLists.txt

tfoote:
If they'll take it upstream that would be great. Though I suspect they may not want to take it. In which case you can simply overlay that CMakeLists.txt with the package.xml in the release repository patch directory.

tfoote:
@jgoppert Recently @asmodehn  has been working on packaging python packages. He might be able to help.

asmodehn:
In my case I m packaging pure python packages with catkin_pip, a cmake extension I wrote for catkin.

It is currently still experimental, so I wouldnt advise it for something as basic as transforms just yet... but maybe the next release :-)

However I now know the steps required to get pure python to work with ROS. Not sure though if they should be implemented in catkin or bloom or both, but if someone is interested in working on that, catkin_pip is a good starting point and I can probably advise. Just let me know !

PR #12485, opened 2016-08-24
Added curlpp to base rosdep repository list

rdelfin:


tfoote:
Please link to the upstream packages as per https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

Also is there a Fedora equivalent?
And you have a typo in the debian key.

rdelfin:
Just to confirm, should I add the upstream packages in rosdistro/$DISTRIBUTION/distribution.yaml ? This is the first time I've done this some I'm not entirely certain

rosdep/base.yaml
```diff
@@ -560,6 +560,10 @@ curl:
   opensuse: [libcurl-devel, curl]
   rhel: [libcurl-devel, curl]
   ubuntu: [libcurl4-openssl-dev, curl]
+curlpp:
```
tfoote: Please change this to be libcurlpp-dev instead of just curlpp since we may want to have a future rule for just the runtime with the name curlpp. Our default is to copy the name of the debian package.

tfoote:
This is a rosdep rule, you don't need to change the rosdistro entries. 

I was asking for a link to 
- http://packages.ubuntu.com/xenial/libcurlpp-dev
- https://packages.debian.org/jessie/libcurlpp-dev

Since on fedora I think we actually want: curlpp-devel to match the tags: https://apps.fedoraproject.org/packages/curlpp-devel

Not just the runtime: https://apps.fedoraproject.org/packages/curlpp

rdelfin:
Alright, corrected the name and switched over to curlpp-devel. I'll keep all this in mind if I do this in the future

tfoote:
+1 

Thanks for iterating on this.

PR #12489, opened 2016-08-24
rosdep: updated rosdep key libgsl for newer Debian and Ubuntu distros

meyerj:
The library `libgsl` has been updated to version 2 and renamed from `libgsl0-dev` to `libgsl-dev` in newer Debian and Ubuntu releases:
- Debian: https://packages.debian.org/source/gsl
- Ubuntu: http://packages.ubuntu.com/source/gsl

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1656,13 +1656,25 @@ libgps:
   ubuntu: [libgps-dev]
 libgsl:
   arch: [gsl]
-  debian: [libgsl0-dev]
+  debian:
+    jessie: [libgsl0-dev]
+    stretch: [libgsl-dev]
+    wheezy: [libgsl0-dev]
   fedora: [gsl]
   gentoo:
     portage:
       packages: [sci-libs/gsl]
   macports: [gsl]
-  ubuntu: [libgsl0-dev]
+  ubuntu:
+    precise: [libgsl0-dev]
+    raring: [libgsl0-dev]
+    saucy: [libgsl0-dev]
+    trusty: [libgsl0-dev]
+    utopic: [libgsl0-dev]
+    vivid: [libgsl0-dev]
+    wily: [libgsl0-dev]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -1656,13 +1656,22 @@ libgps:
   ubuntu: [libgps-dev]
 libgsl:
   arch: [gsl]
-  debian: [libgsl0-dev]
+  debian:
+    wheezy: [libgsl0-dev]
+    jessie: [libgsl0-dev]
+    stretch: [libgsl-dev]
+    sid: [libgsl-dev]
```
tfoote: We generally prefer not to put sid rules in since it's a rolling distro and thus is expected to change and we don't want to be trying to keep up.

rosdep/base.yaml
```diff
@@ -1656,13 +1656,22 @@ libgps:
...
   fedora: [gsl]
   gentoo:
     portage:
       packages: [sci-libs/gsl]
   macports: [gsl]
-  ubuntu: [libgsl0-dev]
+  ubuntu:
+    precise: [libgsl0-dev]
```
tfoote: Although the other platforms are EOL, please fill in the intermediate platforms since there are people still using them with ROS starting at least from Saucy since that's Indigo.

tfoote:
Thanks for the update. I have two comments but otherwise it looks good.

rosdep/base.yaml
tfoote: This is also out of alphabetical order which is why it's failing the CI.

meyerj:
I addressed your comments and did a force push.
Note that a few other keys also have rules for Debian sid.

tfoote:
Thanks for the update. 

+1

Yeah the sid rules have crept in occasionally. A bunch of them are pretty old. When we touch those elements we now try to clear them, but we generally haven't taken the time to actively clear them.

PR #12510, opened 2016-08-26
Add libopencv-nonfree-dev

jordi-pages:
This library is needed for packages using nonfree features like SIFT, SURF.

These features are used for non-commercial purposes in the following tutorials:

https://github.com/pal-robotics/tiago_tutorials/tree/master/opencv_tut

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2026,6 +2026,8 @@ libopencv-dev:
     vivid: [libopencv-dev]
     wily: [libopencv-dev]
     xenial: [libopencv-dev]
+libopencv-nonfree-dev:
+  ubuntu: [libopencv-nonfree-dev]    
 libopenni-dev:
   arch: [openni]
   debian: [libopenni-dev]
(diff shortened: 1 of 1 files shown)


tfoote:
I am not familiar with that package and I cannot find that package on the ubuntu repositories: http://packages.ubuntu.com/search?suite=all&arch=any&searchon=names&keywords=libopencv-nonfree-dev

In general I don't think that we want to be bringing in a non-standard version of opencv there are a lot of packages building on opencv which might break if a custom built version is not built in the exact same manner.

jordi-pages:
@tfoote I agree that it is strange that the package is not include din Ubuntu repositories. The way to install it is:

sudo add-apt-repository --yes ppa:xqms/opencv-nonfree
sudo apt-get update
sudo apt-get install libopencv-nonfree-dev

Let me know what do you think about including this. If not possible then we will need to modify the tutorials and skip SIFT and SURF.

Thanks

vrabaud:
You will not need it for Kinetic and above as it is in the ROS package of OpenCV3.
For Jade and below, it is only supported on Trusty so that would not work on the ROS buildfarms.

jordi-pages:
OK, got it. Thanks.

PR #12530, opened 2016-08-30
Added some lost OS X Homebrew dependencies as libx11 --> xquartz or postersql

drmateo:


tfoote:
Can you please link to references for the contents of these packages?

drmateo:
I beg my pardon, I'm new contributor and I don't know how I can link to references.

tfoote:
Please link to documentation for the formulas that you've referenced so it's clear they're publicly available and so they can be confirmed to have the expected content.

wjwwood:
Normally we link to a place where the rules can be checked for correctness. For Ubuntu that's http://packages.ubuntu.com/, but I'm not sure there is an equivalent authority for Homebrew.

These changes look ok to me. +1

drmateo:
Ok! Then I guess that the link to the github repository where are located the formulas is enough.

- libpq-dev: https://github.com/Homebrew/homebrew-core/blob/master/Formula/postgresql.rb
- libspnav-dev & spacenavd: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libspnav.rb
- libx11: https://github.com/caskroom/homebrew-cask/blob/master/Casks/xquartz.rb

PR #12537, opened 2016-08-30
various test improvements

dirk-thomas:
- disable checks for EOL distros
- disable checks for dry metadata which is not updated anymore
- make test order deterministic
- update several message to quote distro names / paths etc.

This reduces the test time from ~63s to ~ 26s and reduces the output in the Travis log which makes it easier to read if something fails.


tfoote:
+1

PR #12621, opened 2016-09-06
adding python-six for debian

asmodehn:


asmodehn:
https://packages.debian.org/search?keywords=python-six

PR #12623, opened 2016-09-06
adding python-tblib for ubuntu and debian

asmodehn:
http://packages.ubuntu.com/search?keywords=python-tblib
https://packages.debian.org/search?keywords=python-tblib&searchon=names&suite=all&section=all


dirk-thomas:
+1

PR #12628, opened 2016-09-07
Rosdep: Markdown added, gcc for Nvidia

nalt:


rosdep/base.yaml
```diff
@@ -3138,7 +3141,7 @@ nvidia-cg:
 nvidia-cuda:
   debian: [nvidia-cuda-toolkit]
   gentoo: [nvidia-cuda-toolkit]
-  ubuntu: [nvidia-cuda-toolkit]
+  ubuntu: [nvidia-cuda-toolkit, g++-4.8]
```
dirk-thomas: g++ 4.8 is certainly not the right choice across all Ubuntu distributions. Can you please reference some docs why you think this is necessary.

nalt:
Cuda does not work with the default g++ on 16.04. g++-4.8 is available across several Ubuntu versions and works with Cuda 7.5 (in Ubuntu 16.04 repo). See http://docs.nvidia.com/cuda/cuda-installation-guide-linux/
I agree that it might not be the best choice to force a certain g++ version here. Would it be better to have an additional rule, like nvidia-cuda-gcc?

dirk-thomas:
It might be available across *several* Ubuntu versions but it isn't available across all. So a blank rule `ubuntu: [g++-4.8]` will fail on those, e.g. Precise.

A separate rule like `nvidia-cuda-gcc` would be ok I guess if it only defined for distros which actually have G++ 4.8. It is just unclear to me what you want to do on the others. They simply can't use `nvidia-cuda` then? How is that currently being used?

tfoote:
What are the combinations the work on other Ubuntu versions? This rule needs to be able to resolve correctly on all versions.

nalt:
Which Ubuntu versions should be supported? I'll add separate rules for each Ubuntu version.
Other Ubuntu versions have different Cuda versions, e.g. 5.5 on precise. Its default compiler (4.6) works for Cuda 5.5.
Recommended g++ versions are in the "Getting started" guides linked on https://developer.nvidia.com/cuda-toolkit-archive . g++-4.7 would be compatible with all Cuda versions since 5.5, and it is available from Ubuntu 12.10 to 16.10. However, it is not the _recommended_ version for more recent Ubuntu versions.

tfoote:
At the moment we need to support at least saucy and newer to support our LTS Indigo: http://www.ros.org/reps/rep-0003.html#indigo-igloo-may-2014

dirk-thomas:
Should this be closed?

tfoote:
I created #13465 to address the markdown aspect. We need more info on the compiler list. In general I'm entirely comfortable with the potential side effects of changing the default compile toolchain if you just ask for a specific package's dependencies. There's a lot of other things that may rely on the default version.

@nalt This needs to resolve for all versions of ubuntu back to the oldest one in a supported rosdistro. Right now our low water mark is Saucy. If you'd like to follow up on this please open a new PR for the gcc support.

PR #12637, opened 2016-09-08
Update geographiclib package for newer Ubuntu releases

kartikmohta:


tfoote:
+1 

For reference: http://packages.ubuntu.com/xenial/libgeographic-dev
http://packages.ubuntu.com/trusty/libgeographiclib-dev

dirk-thomas:
What about entire for U and V?

PR #12677, opened 2016-09-12
Add python-pynfft to rosdep.

nstiurca:
The pynfft package allows non-uniform fourier analysis of signals. The
package is available at https://pypi.python.org/pypi/pynfftls, but is installed
through apt so as to satisfy underlying dependencies such as libfftw and libnfft3.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1613,6 +1613,9 @@ python-pymouse:
   ubuntu:
     pip:
       packages: [pymouse]
+python-pynfft:
+  debian: [python-pynfft]
+  ubuntu: [python-pynfft]
 python-pynmea2:
   ubuntu:
     pip:


tfoote:
Please use the shorthand approach. apt is the default for ubuntu so it doesn't need to be declared explicitly, nor does the "packges" key.

Ubuntu packages: http://packages.ubuntu.com/xenial/python-pynfft
It's also in debian stretch and newer: https://packages.debian.org/stretch/python-pynfft so we should add a debian rule.

It does not look to be available for fedora yet: https://admin.fedoraproject.org/pkgdb/packages/%2Anfft%2A/

nstiurca:
Might want to make it clearer how to add packages somewhere then. There's other instances already using explicit forms, eg.

```
python-pyproj:
  debian:
    apt:
      packages: [python-pyproj]
```

which is what I was pattern matching. Anyway, pushing a new patch.

nstiurca:
Ok, I think that should be better.

nstiurca:
Although I don't have access to debian to test that part on.

tfoote:
Thanks for the update +1

PR #12702, opened 2016-09-15
Update Debian header file package names

mdhorn:
Debian does not have a package named 'linux-headers-generic'
E: Package 'linux-headers-generic' has no installation candidate

Found while testing DKMS functionality as part of ros-<distro>-librealsense package install as dependency of ros-<distro>-realsense-camera in Debian docker images for ARM.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2936,7 +2936,10 @@ libzmqpp3:
   ubuntu: [libzmqpp3]
 linux-headers-generic:
   arch: [linux-headers]
-  debian: [linux-headers-generic]
+  debian:
+    jessie: [linux-headers-3.16.0-4-all]
+    stretch: [linux-headers-4.6.0-1-all]
+    wheezy: [linux-headers-3.2.0-4-all]
   fedora: [kernel-headers]
   ubuntu: [linux-headers-generic]
 log4cxx:


rosdep/base.yaml
```diff
@@ -2936,7 +2936,10 @@ libzmqpp3:
   ubuntu: [libzmqpp3]
 linux-headers-generic:
   arch: [linux-headers]
-  debian: [linux-headers-generic]
+  debian:
+    jessie: [linux-headers-3.16.0-4-all]
+    sid: [linux-headers-3.16.0-4-all]
```
tfoote: We generally don't do rules for sid since it's a rolling release, but it would be good to add stretch: https://packages.debian.org/stretch/linux-headers-4.6.0-1-all

rosdep/base.yaml
```diff
@@ -2936,7 +2936,10 @@ libzmqpp3:
   ubuntu: [libzmqpp3]
 linux-headers-generic:
   arch: [linux-headers]
-  debian: [linux-headers-generic]
+  debian:
+    jessie: [linux-headers-3.16.0-4-all]
```
tfoote: For reference: https://packages.debian.org/jessie/linux-headers-3.16.0-4-all

rosdep/base.yaml
```diff
@@ -2936,7 +2936,10 @@ libzmqpp3:
...
+    sid: [linux-headers-3.16.0-4-all]
+    wheezy: [linux-headers-3.2.0-4-all]
```
tfoote: Reference: https://packages.debian.org/wheezy/linux-headers-3.2.0-4-all

rosdep/base.yaml
tfoote: This seems to be the best reference I've found for the default kernel versions: https://packages.debian.org/search?keywords=linux-source

tfoote (changes requested):
Thanks for digging into this. One requested change from sid to stretch.

mdhorn:
@tfoote I don't see header packages when I search for stretch
https://packages.debian.org/search?suite=stretch&arch=any&searchon=names&keywords=linux-headers-
Am I looking in the wrong place?
I'd be happy to drop 'sid' and/or 'stretch' and leave the two I know work which are jessie and wheezy.

tfoote:
You're in the right place. The search is being too "smart" 
```
Your keyword was too generic.
Please consider using a longer keyword or more keywords.
```
And for some reason it gives everything but stretch when you expand

If you make it a little more specific you get the stretch results: https://packages.debian.org/search?suite=stretch&arch=any&searchon=names&keywords=linux-headers-4

I believe 4.6 is the default kernel so the `stretch: linux-headers-4.6.0-1-all` should be right.

mdhorn:
@tfoote Updated. Thank you.

tfoote (approved):
Thanks for iterating.

mdhorn:
@tfoote Thank you!
Once this is merge I'll submit a new release of librealsense which requires the linux-headers. cheers.

dirk-thomas:
On Ubuntu the exact package name for the linux headers changes with every patch release which happens quite frequently. Isn't that also the case on Debian?

mdhorn:
@dirk-thomas Ubuntu, and other linux, provide a nice meta package name to point to the correct headers for the current kernel; linux-headers-generic.
For some reason, Debian has a meta name for the kernel image, but not one for the linux-headers. Looks like someone assumed there was a meta package, or maybe perviously there was, just not the case in currently.

dirk-thomas:
Doesn't that imply that in a couple of month the rosdep rule will try to install an outdated package?

mdhorn:
@dirk-thomas Maybe.
Debian has meta packages for the linux-headers per architecture type:
linux-headers-armmp, linux-headers-arm64, linux-headers-arm64, etc
but not linux-headers-generic like other distributions.
If the rosdep yaml supported architecture, then a meta package per arch could be used.

tfoote:
It looks like debian does a weird renumbering thing where it actually releases the newer kernels with older version numbers: 
Form https://packages.debian.org/jessie/linux-headers-3.16.0-4-all
`Package: linux-headers-3.16.0-4-all (3.16.36-1+deb8u1) `

> This package depends against all architecture-specific kernel header files for Linux kernel version 3.16, generally used for building out-of-tree kernel modules.

The package 3.16.0-4 actually has 3.16.36-1+deb8u1 inside it and I think will stay up to date for all 3.16 changes.

dirk-thomas:
If the key is stable +1 from me.

PR #12710, opened 2016-09-15
Removed source entries from vigir_footstep_planning stack

astumpf:
Removed source entries from vigir_footstep_planning stack due to Jenkins build errors and we need only documentation to be generated.


tfoote:
In general we recommend keeping the source entries and disabling the jenkins job like this: https://github.com/ros/rosdistro/blob/master/indigo/distribution.yaml#L4421

`test_commits: false`

That will allow your package to be used with the rosinstall generator, but will disable the jenkins jobs.

astumpf:
Thanks for your kind advice! I've just re-added everything in #12743 as proposed.

PR #12828, opened 2016-09-26
[rosdep] add liblinphone-dev

benmaidel:
debian: https://packages.debian.org/jessie/liblinphone-dev
ubuntu: http://packages.ubuntu.com/xenial/liblinphone-dev
fedora: not available

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1860,6 +1860,7 @@ liblinear-dev:
   fedora: [liblinear-devel]
   ubuntu: [liblinear-dev]
 liblinphone-dev:
+  fedora: [linphone-devel]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Thanks for the links to the debian and ubuntu. 

It looks like `linphone-devel` is available in Fedora as of version 23: https://apps.fedoraproject.org/packages/linphone-devel  

Could you add `fedora: linphone-devel` to the entries.

ipa-bnm:
Thanks for the fedora link. I've added it.

tfoote (approved):

tfoote:
Thanks for the quick update.

PR #12862, opened 2016-09-30
Desktop Install for Slackware

nikonikolov:
Defined dependencies for Desktop Install on Slackware Linux

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -69,6 +69,9 @@ apr:
   macports: [apr, apr-util]
   opensuse: [libapr1, libapr-util1]
   rhel: [apr-devel, apr-util]
+  slackware:
+    slackpkg:
+      packages: [apr, apr-util]
   ubuntu: [libapr1-dev, libaprutil1-dev]
 arduino-core:
   debian: [arduino-core]
@@ -105,6 +108,7 @@ assimp:
   macports: [assimp]
   opensuse: [libassimp3]
   rhel: [assimp-devel]
+  slackware: [assimp]
   ubuntu:
     lucid:
       apt:
@@ -157,6 +161,7 @@ assimp-dev:
       packages: [media-libs/assimp]
   opensuse: [libassimp3]
   rhel: [assimp-devel]
+  slackware: [assimp]
   ubuntu:
     lucid:
       apt:
@@ -335,6 +340,9 @@ boost:
   macports: [boost]
   opensuse: [boost-devel]
   rhel: [boost-devel]
+  slackware:
+    slackpkg:
+      packages: [boost]
   ubuntu:
     lucid:
       apt:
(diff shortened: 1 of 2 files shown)


vrabaud:
very nice feature. Please squash at some point but mostly fix build errors:
https://travis-ci.org/ros/rosdistro/builds/163984375
(mostly, respect alphabetical order). Thx!

vrabaud:
Working ! I'd suggest you squash the commits though it's not necessary. It's just cleaner (and having not valid commits is always ugly imho).
I'll wait for @tfoote to validate, not sure if there is anything to do for a new distro.

nikonikolov:
Done, thanks! On a side note, how can I put install instructions onto this page http://wiki.ros.org/Installation/Slackware?

rosdep/base.yaml
```diff
@@ -69,6 +69,9 @@ apr:
   macports: [apr, apr-util]
   opensuse: [libapr1, libapr-util1]
   rhel: [apr-devel, apr-util]
+  slackware:
+    slackpkg:
```
tfoote: Is slackpkg not the default package manager? I see three types of rules default, slackpkg, and pip. Clearly the pip ones needs to be long format, but does the slackpkg also?

tfoote (commented):
Thanks for the contribution. This will be great. Filling out the documentation for Slackware on the install page would also be appreciated. I'd also ask that you subscribe to those pages if you'll have a running system to watch for potential changes. 

I have one question about the slackpkg installer rules but otherwise it's great to fill in the dependencies.

tfoote:
@vrabaud we usually use the "squash and merge" button so the result in the main branch will be clean.

rosdep/base.yaml
nikonikolov: So slackpkg is the default package manager for Slackware. There is a limitation however - it works only with packages which come with Slackware by default. So it is mostly used to update and install (in case you did not install during OS installation) packages from the main Slackware tree. 

I have chosen sbotools as the default installer for the rules in rosdep. Sbotools is installer which can install packages from slackbuilds.org - a third party repository for slackware packages which has been approved by the Slackware team. 

Note that packages which come with the official Slackware OS will not be hosted on slackbuilds.org. Hence some packages in the rules above need to use slackpkg, others - sbotools.

tfoote: Ok, thanks for filling in the context. Then this looks good to me.

tfoote (approved):

nikonikolov:
Thanks! I am struggling to find out how to fill instructions for the install page at http://wiki.ros.org/Installation/Slackware. Where should I submit them?

tfoote:
You need to create a page with a name like: kinetic/Installation/Slackware

Assuming you're working on kinetic I edited the landing page to include that specifically when kinetic is selected. You probably want to start with the structure of the electric page to stay consistent.

nikonikolov:
Thanks, I am working on the install instructions. What else needs to be done so that the pull request is merged? It would be easier for the install instructions.

tfoote:
We just need a 2nd approval/merge from @ros/rosdistro-maintainers

Issue #11626, opened 2016-05-24
rosdep libftdipp-dev not resolving properly for Ubuntu 16.04


koenlek:
It tries to install "libftdipp-dev" but I think it should be "libftdipp1-dev" on xenial. Should the rosdep in https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml
be updated from this:

```
libftdipp-dev:
  fedora: [libftdi-c++-devel]
  ubuntu: [libftdipp-dev]
```

To e.g. this:

```
libftdipp-dev:
  fedora: [libftdi-c++-devel]
  ubuntu: 
    precise: [libftdipp-dev]
    trusty: [libftdipp-dev]
    vivid: [libftdipp-dev]
    wily: [libftdipp-dev]
    xenial: [libftdipp1-dev]
```

wjwwood:
Please open a pull request to address the problem. You might want to @ mention whoever added it originally to see what they think.

Thanks!

koenlek:
Good one. Having some busy days, but I plan to make the pull request when I get the time.

koenlek:
Finally found some time ;)

tfoote:
Resolved in #12866

PR #12866, opened 2016-09-30
libftdipp-dev support for ubuntu xenial

koenlek:
Should fix issue #11626. 

I tested it by temporary updating `/etc/ros/rosdep/sources.list.d/20-default.list` to swap the default base.yaml for mine, which worked great:

```
#yaml https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml
yaml https://raw.githubusercontent.com/koenlek/rosdistro/feature/libftdipp_for_xenial/rosdep/base.yaml
```


tfoote (approved):
Thanks for the follow up. I also inserted rules for saucy, jessie and stretch.

PR #12893, opened 2016-10-04
Add pika package

tomas789:
Pika is a pure-Python implementation of the AMQP 0-9-1 protocol that tries to stay fairly independent of the underlying network support library.

PYPI: https://pypi.python.org/pypi/pika/0.10.0
Macports: https://trac.macports.org/browser/trunk/dports/python/py-pika/Portfile
Ubuntu Xenial: http://packages.ubuntu.com/xenial/python-pika
Ubuntu Trusty: http://packages.ubuntu.com/trusty/python-pika
Ubuntu Precise: http://packages.ubuntu.com/precise/python-pika

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -3817,7 +3817,7 @@ repositories:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/clearpath-gbp/husky_robot-release.git
-      version: 0.2.5-0
+      version: 0.2.6-0
     source:
       type: git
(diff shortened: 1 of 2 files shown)


tomas789:
Note that there is weird commit history. But File changes section is on.

tfoote (approved):
Thanks for linking to the package resources. I added a debian rule too. https://packages.debian.org/stretch/python-pika

lgtm. The history shouldn't be a problem. We'll make sure to squash it on merge.

PR #12895, opened 2016-10-05
added dxfgrabber python package

jabbenseth:
A Python library to grab information from DXF drawings - all DXF versions supported.
https://pypi.python.org/pypi/dxfgrabber/0.8.1

Intended use: Generate (feature)maps from DXF drawings.


tfoote (approved):
Since this is a pip only entry I added `-pip` to the key. This will allow for a transition to a potential upstream version if it gets packaged into upstream packaging.

ipa-mig:
@tfoote thanks. This is fine for us.

PR #12898, opened 2016-10-06
Add python-gst-1.0

mvollrath:


dirk-thomas (approved):

tfoote (approved):
I added a debian rule too. Fedora appears not to host gstreamer due to potential licensing issues. 

https://packages.debian.org/jessie/python/python-gst-1.0
http://packages.ubuntu.com/trusty/python/python-gst-1.0

mvollrath:
Thanks!

PR #12912, opened 2016-10-09
mavlink: 2016.10.10-0 in 'kinetic/distribution.yaml' [bloom]

vooon:
Increasing version of package(s) in repository `mavlink` to `2016.10.10-0`:
- bloom version: `0.5.22`
- previous version for package: `2016.9.9-0`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2123,7 +2123,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/mavlink/mavlink-gbp-release.git
-      version: 2016.9.9-0
+      version: 2016.10.10-0
     source:
       type: git
       url: https://github.com/mavlink/mavlink-gbp-release.git


vooon:
**IMPORTANT**: this release will not include `pymavlink` library. Only C and C++11 libs.
Use `python-pymavlink` rosdep dependency to use latest pypi version.

dirk-thomas:
Just to clarify: was `pymavlink` part of the previous release? If yes, do you expect downstream packages to potentially fail building without it being included anymore?

vooon:
Yes, it is part of previous releases. But i do not see other users of "mavlink" package: http://wiki.ros.org/mavlink

And as i see, rospilot uses `python-pymavlink` https://github.com/ros/rosdistro/blame/master/rosdep/python.yaml#L1574

Main reason to remove pymavlink from that package is conflict with pymavlink from pip. This give more trouble in development, so i decide to not install bundled pymavlink, just use it to generate C and C++11 lib.

dirk-thomas:
It's always a difficult decision imo since only public package dependencies are visible. But I will keep this decision to you and merge the pr as is.

PR #12934, opened 2016-10-12
Rosjava kinetic deps

airballking:


airballking:
Tested on my local machine, following these instructions: http://docs.ros.org/independent/api/rosdep/html/contributing_rules.html

tfoote (approved):
lgtm

dirk-thomas:
The entries for non-LTS versions are missing. Since Kinetic is e.g. targeting Xenial as well as Wily I think all missing ones should be filled in as well.

airballking:
No problem, done.

tfoote:
There were still a few missing intermediate ones. I added a commit that collapses the rules into a non-version specific fule. The only different entry was for lucid which is 1.5 years past it's 5 year server-only EOL. Now the rules won't need to be updated regularly as the names appear to be stable.

airballking:
Even better. Did not occur to me to do this. Thanks @tfoote

PR #13010, opened 2016-10-21
Add fcl support for fedora

richmattes:
fcl is now included in the fedora repositories; added a fedora key to libfcl-dev
https://bugzilla.redhat.com/show_bug.cgi?id=1103555

The fedora fcl packages won't be in the stable repositories for about a week, but I wanted to get the PR created so it doesn't fall off the radar.  You can watch the above bug for status, and I'll comment once I submit them to stable.

This should solve https://github.com/ros-gbp/fcl-release/issues/1

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1490,6 +1490,7 @@ libexpat1-dev:
 libfcl-dev:
   arch: [fcl]
   debian: [libfcl-0.5-dev]
+  fedora: [fcl-devel]
   ubuntu: [libfcl-0.5-dev]
 libfftw3:
   arch: [fftw]


tfoote:
It looks like it's been approved! And I see it in the status page: https://admin.fedoraproject.org/pkgdb/package/rpms/fcl/ but can't find the packages themselves.

jpgr87:
The packages are built and in the testing repositories now:
https://bodhi.fedoraproject.org/updates/?packages=fcl

They have to stay in testing for a week before they can be pushed to stable.  I'll make the request this weekend, and the package should be available Sunday or Monday.
https://fedoraproject.org/wiki/Updates_Policy#All_other_updates

jpgr87:
fcl is now pushed to the stable repositories in all fedora releases.  It should be OK to apply this PR.

wjwwood:
Thanks! @130s and @davetcoleman next time you release moveit for Kinetic this should be available and therefore you hopefully won't have to skip fedora.

PR #13011, opened 2016-10-21
python.yaml: Add python-github-pip

wkentaro:
The installation command is `pip install PyGithub` but for importing,
it is `import github`


dirk-thomas:
Needs to be in alphabetical order. Otherwise +1.

wkentaro:
Thanks!

tfoote (approved):

PR #12774, opened 2016-09-20
class_loader: 0.3.5-0 in 'indigo/distribution.yaml' [bloom]

mikaelarguedas:
Increasing version of package(s) in repository `class_loader` to `0.3.5-0`:
- upstream repository: https://github.com/ros/class_loader
- release repository: https://github.com/ros-gbp/class_loader-release.git
- distro file: `indigo/distribution.yaml`
- bloom version: `0.5.22`
- previous version for package: `0.3.4-0`
## class_loader

```
* Add ClassLoader::createUniqueInstance (#38 <https://github.com/ros/class_loader/issues/38>)
  * Wrap comments on createInstance and friend.
  * Delegate createInstance and createUnmanagedInstance to private impl.
  * Add ClassLoader::createUniqueInstance.
  * MultiLibraryClassLoader: Factor out getClassLoaderForClass.
  * MultiLibraryClassLoader: Add unique_ptr API.
  * Add tests for unique_ptr API.
* Contributors: Maarten de Vries
```

PR #13050, opened 2016-10-25
class_loader: 0.3.6-0 in 'indigo/distribution.yaml' [bloom]

mikaelarguedas:
Increasing version of package(s) in repository `class_loader` to `0.3.6-0`:
- upstream repository: https://github.com/ros/class_loader
- release repository: https://github.com/ros-gbp/class_loader-release.git
- bloom version: `0.5.22`
- previous version for package: `0.3.4-0`
## class_loader

```
* Made changes to two locking mechanisms inside class loader core's loadLibrary() function: 1) I added a lock to the 'addClassLoaderOwnerFor...' function to protect it against a race condition with the unloadLibrary() function. 2) I also raised the loader lock to cover the whole function. Previously the check to see if a library is already loaded and the actual loading of the library was not atomic. Multiple threads could create shared library objects, for example.
* Contributors: Jonathan Meyer
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1233,7 +1233,7 @@ repositories:
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/ros-gbp/class_loader-release.git
-      version: 0.3.4-0
+      version: 0.3.6-0
     source:
       test_pull_requests: true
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
close #12774

PR #13140, opened 2016-11-06
rosdep: Add python-future package

vooon:
This patch add python-future package: https://pypi.python.org/pypi/future

Used by upcoming mavlink (new dependencies: lxml, future).


dirk-thomas (approved):

dirk-thomas:
The package is also not available in Ubuntu < Wily: see #13321.

PR #13158, opened 2016-11-09
ecl_core: 0.61.15-0 in 'kinetic/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `ecl_core` to `0.61.15-0`:

- upstream repository: https://github.com/stonier/ecl_core.git
- release repository: https://github.com/yujinrobot-release/ecl_core-release.git
- bloom version: `0.5.23`
- previous version for package: `0.61.14-0`

## ecl_geometry

```
* pose -> legacy pose
* new pose (function only) library started, should eliminate the need for most type conversions
```

## ecl_statistics

```
* Class added for computing cumulative means and variances
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -799,7 +799,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/yujinrobot-release/ecl_core-release.git
-      version: 0.61.14-0
+      version: 0.61.15-0
     source:
       type: git
       url: https://github.com/stonier/ecl_core.git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -799,11 +799,11 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/yujinrobot-release/ecl_core-release.git
-      version: 0.61.14-0
+      version: 0.61.15-0
     source:
       type: git
       url: https://github.com/stonier/ecl_core.git
-      version: devel
+      version: release/0.61-indigo-kinetic
```
dirk-thomas: Just out of curiosity can you elaborate why you are switching to custom "version-locked" branches? I would like to understand better how / why this is being used.

PR #13167, opened 2016-11-09
repository_name: 1.0.1-1 in 'indigo/distribution.yaml' [bloom]

AndyZe:
Increasing version of package(s) in repository `repository_name` to `1.0.1-1`:

- upstream repository: https://github.com/UTNuclearRoboticsPublic/gauges.git
- release repository: https://github.com/UTNuclearRoboticsPublic/gauges-release.git
- bloom version: `0.5.22`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -9442,6 +9442,15 @@ repositories:
       url: https://github.com/So-Cool/report_card.git
       version: master
     status: maintained
+  repository_name:
+    release:
+      packages:
+      - gauges
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/UTNuclearRoboticsPublic/gauges-release.git
+      version: 1.0.1-1
+    status: maintained
   resource_retriever:
     doc:
       type: git


indigo/distribution.yaml
```diff
@@ -9442,6 +9442,15 @@ repositories:
       url: https://github.com/So-Cool/report_card.git
       version: master
     status: maintained
+  repository_name:
```
dirk-thomas: Please use a reasonable name here, e.g. `gauges`.

Once you have updated the above name you will have to remove the following part (since the package name now equals the repo name which doesn't need to explicitly stated):

```
packages:
- gauges
```

dirk-thomas (changes requested):

indigo/distribution.yaml
AndyZelenak: Never mind, I see.

dirk-thomas:
You can also update the patch of this PR by editing the file in the fork / branch this PR is made from. Then you don't need to create a new PR.

PR #13168, opened 2016-11-09
Adding the gauges package to indigo/jade/kinetic.

AndyZe:


tfoote:
Thanks for fixing the name as mentioned in #13167, however you cannot just duplicate the content to each rosdistro. You need to run bloom-release for each rosdistro because it generates rosdistro specific tags in the release repo to support the different target platforms.

You can either edit this branch to just be for indigo or open a new pull request with `bloom-release` using the `-p` option on indigo again. And you'll need to run `bloom-release` for jade and kinetic as well.

PR #13185, opened 2016-11-11
Add tilestache to rosdep db

evenator:
Tilestache (http://tilestache.org/) is a Python-based server application that can serve up map tiles based on rendered geographic data.


tfoote (approved):
- http://packages.ubuntu.com/yakkety/tilestache
- https://packages.debian.org/jessie/tilestache
- https://admin.fedoraproject.org/pkgdb/package/rpms/python-tilestache/

PR #13186, opened 2016-11-11
Removed source builds from recent commit f3c15d0

kmhallen:


wjwwood:
We can certainly do this, but if the reason is to disable the build farm trying to build each commit you make, you can do that with the `test_commits: false` line inside the `source:` block.

dirk-thomas:
The advantage of the approach @wjwwood mentioned is that users can still easily fetch your code using tools like `rosinstall_generator`.

wjwwood:
Hmm, looks like something went wrong and now you're partially removing the doc sections. You'll either have to undo that or remove the `doc:` line as well (if you're intention is disable doc jobs). You want something like this:

```yaml
  dataspeed_can:
    doc:
      type: hg
      url: https://bitbucket.org/dataspeedinc/dataspeed_can
      version: default
    source:
      test_commits: false
      type: hg
      url: https://bitbucket.org/dataspeedinc/dataspeed_can
      version: default
    status: developed
```

kmhallen:
Fixed, silly mistake...

wjwwood:
No worries, thanks for fixing it up quickly.

PR #13187, opened 2016-11-11
Revert "eigen_stl_containers: 0.1.6-0 in 'indigo/distribution.yaml' [bloom]"

tfoote:
Reverts ros/rosdistro#12842

@wjwwood This is causing a major regression on Indigo Saucy. http://repositories.ros.org/status_page/ros_indigo_default.html?q=REGRESSION 

Upstream ticket https://github.com/ros/eigen_stl_containers/issues/7

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2464,11 +2464,20 @@ repositories:
       version: master
     status: maintained
   eigen_stl_containers:
+    doc:
+      type: git
+      url: https://github.com/ros/eigen_stl_containers.git
+      version: master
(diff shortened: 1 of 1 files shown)


dirk-thomas:
There is no need to revert the doc and source entries.

tfoote:
@dirk-thomas it was in flight the revert button doesn't let me edit before opening the PR

tfoote:
Merging this for now to get saucy working again. It sounds like the upstream issue knows what's necessary for an intermediate solution.

PR #13200, opened 2016-11-14
remove partial duplicate eigen3 rosdep rule

tfoote:
Both of these rules are defined above in the `eigen` tag. The `eigen3` tag looks to have been added in a sweeping update for gentoo and then changed in a sweeping fedora update where it should have been merged with the existing entry.

[From this question](http://answers.ros.org/question/247968/build_depend-for-eigen3-kinetic-in-packagexml/?answer=247981#post-id-247981)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -689,12 +689,7 @@ eigen2:
   opensuse: [libeigen2-devel]
   ubuntu:
     apt:
-      packages: [libeigen2-dev]
-eigen3:
-  fedora: [eigen3-devel]
-  gentoo:
-    portage:
-      packages: [dev-cpp/eigen]
+      packages: [libeigen2-dev
 emacs:
   arch: [emacs]
   debian: [emacs]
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
This will probably break some packages for some users, right (as in: `rosdep` will not be able to resolve their `build_depend` on the `eigen3` key)? Should this be announced, or too minor?

tfoote:
This would only effect purely fedora and gentoo users. And anything released or with CI will have already failed due to the lack of ubuntu and debian rules. Clearly it could break a few, but the fix is easy and relatively easy to discover so I don't think it's worth a large broadcast that won't effect most people.

We should find a way to make these sorts of discussions more visible and with better history.

gavanderhoorn:
A post in the *release* category on the discourse site doesn't seem overly 'broadcast'?

tfoote:
Yeah, that's probably a good location. Maintainers are a good target audience as they're the most likely to be affected.

jack-oquin:
+1; looks good to me

tfoote:
[Here's](http://discourse.ros.org/t/removal-of-eigen3-rosdep-rule/822?u=tfoote) the post I made 3 days ago.

@ros/rosdistro-maintainers for review?

dirk-thomas (approved):

PR #13227, opened 2016-11-16
Adding sbg_driver to documentation index for distro

ThomasLeMezo:


kinetic/distribution.yaml
```diff
@@ -4927,6 +4927,21 @@ repositories:
...
     status: developed
+  sbg_driver:
+    doc:
+      type: git
+      url: https://github.com/ENSTABretagneRobotics/sbg_ros_driver.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/ENSTABretagneRobotics/sbg_ros_driver-release.git
```
dirk-thomas: This repo doesn't seem to exist? Have you followed the bloom tutorials to create a release?

ThomasLeMezo: I'm currently working on it, should be done very soon (probably less than an hour)

dirk-thomas:
This needs to be rebased against master (with the release added in a separate PR) to add the doc and source entries.

ThomasLeMezo:
Ok, thanks for the speed of the support. I will make a new PR with the right info

PR #13228, opened 2016-11-16
Added theano

eruffaldi:


tfoote:
I see python-theano available as a debian package: https://packages.debian.org/unstable/main/python-theano starting with stretch 
And ubuntu: http://packages.ubuntu.com/yakkety/python-theano starting with yakkety. 

Could you fill in the rules for the other distros so that others can use this? Note that you can use the short hand and skip the `packages:` line to be more compact.

tfoote:
It's also in fedora as of 23: https://admin.fedoraproject.org/pkgdb/package/rpms/python-theano/

eruffaldi:
Added fedora,debian,arch, older ubunuts
Thanks,
-E

rosdep/python.yaml
```diff
@@ -2451,6 +2451,36 @@ python-texttable:
...
   fedora: [python-tftpy]
   ubuntu: [python-tftpy]
+python-theano:
+  arch: [python-theano]
+  debian: 
+    jessie:
+      pip:
+        packages: [Theano]
+    stretch: [python-theano]
+    sid: [python-theano]
```
tfoote: Please don't add sid. It's a rolling target so we don't try to define what's in it. This is also failing the CI due to being out of alphabetical order.

rosdep/python.yaml
```diff
@@ -2451,6 +2451,36 @@ python-texttable:
 python-tftpy:
   fedora: [python-tftpy]
   ubuntu: [python-tftpy]
+python-theano:
+  arch: [python-theano]
+  debian:
```
tfoote: There's trailing whitespace on this line.

tfoote (changes requested):
Thanks for the update. I've added 2 comments on the lines that are failing CI.

And it would be great to have wily and xenial support for Kinetic users.

eruffaldi:
The changes are implemented. Right?

tfoote (approved):

tfoote:
Yeah, I added them thanks.

PR #13234, opened 2016-11-17
adding python-blinker

asmodehn:


asmodehn:
Any hint on why travis complains here ?
```
FAIL: rosdep_formatting_test.test
```

rosdep/python.yaml
```diff
@@ -268,6 +268,13 @@ python-bitstring-pip:
   ubuntu:
     pip:
       packages: [bitstring]
+python-blinker:
+  debian: [python-blinker]
+  fedora: [python-blinker]
+  ubuntu:
```
tfoote: There's a trailing space on this line. That's what travis is complaining about.

tfoote (changes requested):
If you expect zesty to have the same version, it's fine to add it now. Though it looks like the version is exactly the same for all the supported ubuntu versions so I'd suggest simply using `ubuntu: [python-blinker]`

PR #13277, opened 2016-11-23
changing tornado debian entry to not use pip

asmodehn:
because bloom tells me : 
```
Key 'python-tornado' resolved to '['tornado']' with installer 'pip', which does not match the default installer 'apt'.
Failed to resolve python-tornado on debian:jessie with: Error running generator: The Debian generator does not support dependencies which are installed with the 'pip' installer.
```

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2524,10 +2524,7 @@ python-tk:
     xenial: [python-tk]
     xenial_python3: [python3-tk]
 python-tornado:
-  debian:
-    jessie: [python-tornado]
-    sid: [python-tornado]
-    stretch: [python-tornado]
+  debian: [python-tornado]
   fedora: [python-tornado]
   osx:
     pip:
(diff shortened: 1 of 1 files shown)


asmodehn:
Not sure if this might surprise some users (since version on jessie will suddenly drop).
Which is why I did https://github.com/asmodehn/tornado-rosrelease

tfoote (commented):
That's great that everything is availble from debian packages. Thanks for taking the time to proactively package it into Jade, but now that it's available upstream, I'd recommend against continuing the 3rd party package releases and just rely on the upstream.

tfoote (approved):
I simplified the patch since the rules are all the same. Please squash merge.

asmodehn:
@tfoote regarding stopping thirdparty release asap, I totally agree, but I have the issue described at https://github.com/ros/rosdistro/pull/12343#issuecomment-262171114

asmodehn:
@tfoote  Ah just saw your reply thanks !

asmodehn:
I can't squash merge, but squashed version is [here](https://github.com/ros/rosdistro/pull/13317) if you need it.

tfoote:
The squash merge comment was for @ros/rosdistro-maintainers for the 2nd review/merge.

PR #13283, opened 2016-11-23
Adds ceres-solver to Indigo.

damonkohler:


tfoote:
This is failing CI because the version needs to be a branch. But i'm not sure this is exactly what you're looking for a source entry here can enable CI of this repo (devel jobs), and documentation generation. But if you want to use it as a dependency debian packages need to be made by running it through the bloom release process.

PR #13300, opened 2016-11-25
added jaguar_robot package for indigo, jade and kinetic

gstavrinos:


gstavrinos:
Added jaguar and jaguar_robot packages for indigo, jade and kinetic.

indigo/distribution.yaml
```diff
@@ -4580,6 +4580,45 @@ repositories:
...
+      url: https://github.com/gstavrinos/jaguar.git
+      version: kinetic-devel
+    release:
+      packages:
+      - jaguar_control
+      - jaguar_description
+      - jaguar_msgs
+      - jaguar_navigation
+      tags:
+        release: release/kinetic/{package}/{version}
```
dirk-thomas: You can't just copy-n-paste for each distribution. You need to run `bloom` for each of them separately since it generates specific files for each supported platform.

gstavrinos: Thanks for pointing that out. The wiki was not clear on this. The stack has been tested on all three distributions, so I will just have to run each distribution in a different Ubuntu version, just to be sure. Thanks again.

dirk-thomas: Please link to the exact instructions you are referring to. Otherwise or is impossible to guess which documentation you followed.

You have to run `bloom` for each ROS distribution, where each ROS distribution covers a set of Ubuntu versions (as well as Debian and Fedora). This mapping is specified in [REP 3](http://www.ros.org/reps/rep-0003.html).

gstavrinos: This is the page that I am following
http://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation

dirk-thomas: That page is not describing releasing a package? It only mentions documentation.

gstavrinos: Ok, I guess I will have to read the http://wiki.ros.org/bloom page too. Thanks.

gstavrinos:
You can close this pull request. I have made a new one, using bloom. Thanks for your help.

PR #13317, opened 2016-11-29
changing tornado debian entry to not use pip

asmodehn:
because bloom tells me :
```
Key 'python-tornado' resolved to '['tornado']' with installer 'pip', which does not match the default installer 'apt'.
Failed to resolve python-tornado on debian:jessie with: Error running generator: The Debian generator does not support dependencies which are installed with the 'pip' installer.
```

simplifying debian rule for python-tornado


tfoote (approved):

tfoote:
You can force push to the branch for #13277 or the maintainers can squash merge it. Closing this for that one.

PR #13359, opened 2016-12-06
rosdep: python-msgpack support for ubuntu xenial

ledmonster:
Add python-msgpack support for ubuntu xenial


dirk-thomas (approved):
Please consider to check and add the package for the Ubuntu distros in between (U, V, W) and also Y.

ledmonster:
@dirk-thomas okay, I'll add aditional distros support.

ledmonster:
@dirk-thomas Add support for ubuntu from U to Y.

tfoote:
Thanks!

PR #13362, opened 2016-12-06
Add python-shapely for wily and xenial

mistoll:
http://packages.ubuntu.com/xenial/python/python-shapely
http://packages.ubuntu.com/xenial/python/python3-shapely

Wily is EOL, but without it this seems incomplete.


dirk-thomas (approved):
Please consider adding the entry for Y too.

mistoll:
There's Y. 
However, I'm not sure, if there's something like yakkety_python3.

PR #13371, opened 2016-12-07
rosdep: Fix and update python-fcn-pip depends

knorth55:
- `leveldb` is no more needed 
- `libhdf5-dev`, `liblapack-dev` is not found on osx

cc. @wkentaro

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -696,7 +696,7 @@ python-falcon:
 python-fcn-pip:
   osx:
     pip:
-      depends: [libhdf5-dev, liblapack-dev, leveldb]
+      depends: [hdf5, gfortran]
       packages: [fcn]
   ubuntu:
     pip:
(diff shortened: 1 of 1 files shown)


tfoote:
This looks reasonable. Could you provide links to the packages and their availability for future reference in case we need to review/update this in the future?

knorth55:
Like this?

fcn: https://github.com/wkentaro/fcn
homebrew of hdf5: https://github.com/Homebrew/homebrew-science/blob/master/hdf5.rb
hdf5: https://support.hdfgroup.org/HDF5/
homebrew of gfortran: https://github.com/ros/homebrew-deps/blob/master/gfortran.rb
gfortran: http://gcc.gnu.org/wiki/GFortran

tfoote (approved):
@knorth55 Yes, thanks!

wkentaro:
+1

PR #13373, opened 2016-12-07
Add rosdep key for libjsoncpp in debian

evenator:
https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=libjsoncpp


tfoote (approved):
lgtm, I added yakkety and zesty rules in a second commit too.

PR #13394, opened 2016-12-12
Update arni release packages

matt3o:
Adds the new arni metapackage which should simplify the installation of the arni packages. Furthermore renames rqt_arni_gui_detail to arni_rqt_detail_plugin and rqt_arni_gui_overview to arni_rqt_overview_plugin


matt3o:
Dependency problem. Will fix locally and retry pull request when fixed

PR #13397, opened 2016-12-13
gdown is missing on debian/fedra

k-okada:


wkentaro:
Thanks.

tfoote:
Extending this looks good.

Since this is a pip-only package and only has partial coverage what do people think of updating the key to use our new standard '-pip' at the end for future compatibility?

k-okada:
ok, added python-gdown-pip -> 701f83f

tfoote:
Thanks @k-okada I was actually proposing replacing not extending. Such that we can leave the old key space available if it becomes system packaged in the future.

k-okada:
oh, do you mean
```
diff --git a/rosdep/python.yaml b/rosdep/python.yaml
index 782df23..0fefb6d 100644
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -798,7 +798,13 @@ python-fysom:
   ubuntu:
     pip:
       packages: [fysom]
-python-gdown:
+python-gdown-pip:
+  debian:
+    pip:
+      packages: [gdown]
+  fedora:
+    pip:
+      packages: [gdown]
   osx:
     pip:
       packages: [gdown]
```
?

tfoote:
Yes

k-okada:
Ok fixed now

tfoote (approved):
LGTM. Thanks for iterating @k-okada

PR #13401, opened 2016-12-13
Adding explore to documentation index for distro

yzrobot:
I'd like explore to be indexed and documented on ros.org.


dirk-thomas:
The first commit tried to change Groovy which is also EOL for a long time. Therefore we don't accept changes to it anymore. Please update the pull request to only change current distros.

dirk-thomas:
Closing due to no response.

PR #13411, opened 2016-12-14
added tele_dir like a teleoperation software.

rdelgadov:
i'd like tele_dir to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -13160,7 +13160,7 @@ repositories:
       type: git
       url: https://github.com/rdelgadov/tele_dir.git
       version: master
-     status: maintained
+    status: maintained
   teleop_tools:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -13146,6 +13146,21 @@ repositories:
...
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/rdelgadov/tele_dir-release.git
+      version: 0.0.2-2
+    source:
+      type: git
+      url: https://github.com/rdelgadov/tele_dir.git
+      version: master
+     status: maintained
```
dirk-thomas: This line is indented incorrectly. Must be one space less.

rdelgadov: Hi, resolve in the second commit

PR #13412, opened 2016-12-14
Add python-pexpect and python3-pexpect rules for ubuntu:xenial

jacobperron:


dirk-thomas:
Please fill in the entries for the Ubuntu distributions in between if available.

tfoote (approved):

PR #13427, opened 2016-12-17
Added rosdep keys for fswebcam, libxft-dev, ntp, tmux, python-gcloud and and python-ipdb

mirzashah:
* fswebcam
* libxft-dev
* ntp
* tmux
* python-gcloud
* python-ipdb

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -820,6 +820,8 @@ ftdi-eeprom:
   debian: [ftdi-eeprom]
   fedora: [libftdi-devel, libftdi-c++-devel]
   ubuntu: [ftdi-eeprom]
+fswebcam:
+  ubuntu: [fswebcam]
 g++-multilib:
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@mirzashah Please do not merge PRs on this repo on your own without review from others. Especially for rosdep rule we want more than one person to review new entries (see https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

@tfoote I think the new entry which only has a pip package should have a `-pip` suffix. Please see #13429 for this. @mirzashah If the follow up PR gets merged you will have to update your usage of that rosdep key accordingly.

mirzashah:
@dirk-thomas sorry, my bad

tfoote:
@mirzashah Also for packaging we strongly recommend adding debian and fedora rules since bloom runs generators for those too by default now.

PR #13447, opened 2016-12-21
Adding CIR-KIT utilities to documentation index for distro

MoriKen254:
I'd like to some utility packages for our autonomous robot to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -3966,6 +3966,16 @@ repositories:
       url: https://github.com/fkanehiro/hrpsys-base.git
       version: master
     status: developed
+  human_detector:
+    doc:
+      type: git
+      url: https://github.com/CIR-KIT/human_detector.git
+      version: master
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


MoriKen254:
Could you tell me when the index will be actually active after this pull request is merged?

130s:
@MoriKen254 [This thread](http://answers.ros.org/question/230803/how-can-i-check-the-status-of-my-package-submission/) should be of your interest.

indigo/distribution.yaml
```diff
@@ -5738,6 +5748,16 @@ repositories:
       url: https://github.com/orocos-toolchain/log4cpp.git
       version: toolchain-2.8
     status: maintained
+  lower_step_detector:
```
dirk-thomas: Shouldn't this be called `remote_monitor`?

MoriKen254:
@dirk-thomas Excuse me, you are right. I fixed the typo.

@130s Thank you. However http://build.ros.org/ doesn't show on a browser because of proxy error now... Is this supposed to be fix?

PR #13449, opened 2016-12-21
trac_ik: 1.4.5-0 in 'kinetic/distribution.yaml' [bloom]

pbeeson:
Increasing version of package(s) in repository `trac_ik` to `1.4.5-0`:

- upstream repository: https://bitbucket.org/traclabs/trac_ik.git
- release repository: https://github.com/traclabs/trac_ik-release.git
- bloom version: `0.5.23`

## trac_ik

```
* Updated CMake file to work with Eigen on 14.04 through 16.04
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5816,6 +5816,16 @@ repositories:
       type: git
       url: https://bitbucket.org/traclabs/trac_ik.git
       version: no_moveit_plugin-kinetic
+    release:
+      packages:
+      - trac_ik
+      - trac_ik_examples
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -5816,6 +5816,16 @@ repositories:
       type: git
       url: https://bitbucket.org/traclabs/trac_ik.git
       version: no_moveit_plugin-kinetic
```
dirk-thomas: These branch name is not valid anymore.

Same for the `source` entry.

PR #13455, opened 2016-12-21
Update base.yaml to point to debian packages for libnlopt

pbeeson:


rosdep/base.yaml
```diff
@@ -2037,9 +2037,11 @@ libnl-dev:
 libnlopt-dev:
   fedora: [NLopt-devel]
   ubuntu: [libnlopt-dev]
+  debian: [libnlopt-dev]
```
dirk-thomas: The `debian` key needs to be inserted in alphabetical order (hence above `fedora`).

Same below.

dirk-thomas (approved):

rosdep/base.yaml
pbeeson: Fixed

PR #13509, opened 2016-12-30
add fedra and debian entry for python-slacker-cli

k-okada:


dirk-thomas:
Entries needs to be in alphabetical order.

k-okada:
sorry, should be fixed

PR #13522, opened 2017-01-03
Update distribution.yaml

zastrix:
Added UM7.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6261,6 +6261,21 @@ repositories:
       url: https://github.com/ros-drivers/um6.git
       version: indigo-devel
     status: maintained
+  um7:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -6261,6 +6261,21 @@ repositories:
...
+  um7:
+    doc:
+      type: git
+      url: https://github.com/ros-drivers/um7.git
+      version: indigo-devel
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/ros-drivers-gbp/um7-release.git
+      version: 0.0.4-0
```
dirk-thomas: This repo seems to only contain version 0.0.3-0?

dirk-thomas:
I will close this due to no reply.

PR #13523, opened 2017-01-03
add haf_grasping to documentation index for kinetic

davidfischinger:


kinetic/distribution.yaml
```diff
@@ -1519,7 +1519,17 @@ repositories:
...
+  haf_grasping:
+    doc:
+      type: git
+      url: https://github.com/davidfischinger/haf_grasping.git
+      version: kinetic
+    source:
+      type: git
+      url: https://github.com/davidfischinger/haf_grasping.git
+      version: kinetic
+    status: maintained
```
dirk-thomas: This line as well as the other `status` line contain trailing whitespaces which need to be removed.

PR #13527, opened 2017-01-04
rosdep: updated and unified ruby keys

meyerj:
Ubuntu:
- [ruby-defaults](http://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=ruby-defaults&searchon=sourcenames) (source package)
  - 1.9.1 has been the [default version](https://launchpad.net/ubuntu/+source/ruby-defaults/+publishinghistory) from Ubuntu Quantal until Trusty.
  - Since Utopic the default ruby version is 2.1 or newer. Packages `*ruby1.9.1*` do not exist in Utopic and Vivid (added in https://github.com/ros/rosdistro/commit/b979cf3547a6139c134961c7f2d3b033549c6fd6).
  - We can simply use the default version from Quantal on without an effect on existing usages (`ruby-dev` instead of `rubyx.y.z-dev`). Two additional meta-packages would be installed.
  - `rubyx.y.z-dev` already  depends on `librubyx.y.z`. No need to list the library package explicitly?
- [ruby-facets](http://packages.ubuntu.com/search?keywords=ruby-facets&searchon=names&exact=1&suite=all&section=all)
  - replaced [libfacets-ruby](https://launchpad.net/ubuntu/+source/libfacets-ruby/+publishinghistory) since Precise
- [ruby-flexmock](http://packages.ubuntu.com/search?keywords=ruby-flexmock&searchon=names&exact=1&suite=all&section=all)
  - replaced [libflexmock-ruby](https://launchpad.net/ubuntu/+source/libflexmock-ruby/+publishinghistory) since Precise
- [ruby-hoe](http://packages.ubuntu.com/search?keywords=ruby-hoe&searchon=names&exact=1&suite=all&section=all)
- [ruby-nokogiri](http://packages.ubuntu.com/search?keywords=ruby-nokogiri&searchon=names&exact=1&suite=all&section=all)
  - replaced [libnokogiri-ruby](https://launchpad.net/ubuntu/+source/libnokogiri-ruby/+publishinghistory) since Precise
- rdoc never existed as a separate package in Ubuntu. The executable is part of the respective [ruby](http://packages.ubuntu.com/search?keywords=ruby&searchon=names&exact=1&suite=all&section=all) interpreter binary package.
- [ruby-sass](http://packages.ubuntu.com/search?keywords=ruby-sass&searchon=names&exact=1&suite=all&section=all) (moved from base.yaml)

Debian:
- [ruby-defaults](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=sourcenames&keywords=ruby-defaults) (source package)
- [ruby-facets](https://packages.debian.org/search?keywords=ruby-facets&searchon=names&exact=1&suite=all&section=all)
- [ruby-flexmock](https://packages.debian.org/search?keywords=ruby-flexmock&searchon=names&exact=1&suite=all&section=all)
- [ruby-hoe](https://packages.debian.org/search?keywords=ruby-hoe&searchon=names&exact=1&suite=all&section=all)
- [ruby-nokogiri](https://packages.debian.org/search?keywords=ruby-nokogiri&searchon=names&exact=1&suite=all&section=all)
- [rake](https://packages.debian.org/search?keywords=rake&searchon=names&exact=1&suite=all&section=all)
- [rake-compiler](https://packages.debian.org/search?keywords=rake-compiler&searchon=names&exact=1&suite=all&section=all)
- [ruby-sass](https://packages.debian.org/search?keywords=ruby-sass&searchon=names&exact=1&suite=all&section=all) (moved from base.yaml)

What is the reason why [ruby.yaml](https://github.com/meyerj/rosdistro/blob/f09d3f0560d023eef9f2e8ad6d2bab41c11773f0/rosdep/ruby.yaml) is separate from [base.yaml](https://github.com/meyerj/rosdistro/blob/f09d3f0560d023eef9f2e8ad6d2bab41c11773f0/rosdep/base.yaml)? Has the split any consequences on key lookup or are they treated the same?  Should all (new) ruby library keys have a `ruby-` prefix, similar to the ones in [python.yaml](https://github.com/meyerj/rosdistro/blob/f09d3f0560d023eef9f2e8ad6d2bab41c11773f0/rosdep/python.yaml)? Should `ruby` itself be moved to [ruby.yaml](https://github.com/meyerj/rosdistro/blob/f09d3f0560d023eef9f2e8ad6d2bab41c11773f0/rosdep/ruby.yaml)?

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3579,9 +3579,6 @@ ruby:
     trusty: [ruby1.9.1-dev, libruby1.9.1, ruby1.9.1]
     utopic: [ruby1.9.1-dev, libruby1.9.1, ruby1.9.1]
     vivid: [ruby1.9.1-dev, libruby1.9.1, ruby1.9.1]
-ruby-sass:
-  fedora: [rubygem-sass]
-  ubuntu: [ruby-sass]
 ruby1.9.3:
   ubuntu:
     precise: [ruby1.9.3]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3559,33 +3559,6 @@ rsync:
   gentoo: [net-misc/rsync]
   opensuse: [rsync]
   ubuntu: [rsync]
-ruby:
-  arch: [ruby]
-  debian: [ruby, ruby-dev]
-  fedora: [ruby, ruby-devel, openssl-devel, rubygems]
-  gentoo:
-    portage:
-      packages: [dev-lang/ruby]
-  macports: [ruby]
-  ubuntu:
-    lucid: [ruby1.8-dev, libopenssl-ruby1.8, rubygems1.8]
-    maverick: [ruby1.8-dev, libruby1.8, rubygems1.8]
-    natty: [ruby1.8-dev, libruby1.8, rubygems1.8]
-    oneiric: [ruby1.8-dev, libruby1.8, rubygems1.8]
-    precise: [ruby1.8-dev, libruby1.8, rubygems1.8]
-    quantal: [ruby, ruby-dev]
-    raring: [ruby, ruby-dev]
-    saucy: [ruby, ruby-dev]
-    trusty: [ruby, ruby-dev]
-    utopic: [ruby, ruby-dev]
-    vivid: [ruby, ruby-dev]
-    wily: [ruby, ruby-dev]
-    xenial: [ruby, ruby-dev]
-    yakkety: [ruby, ruby-dev]
-    zesty: [ruby, ruby-dev]
-ruby1.9.3:
-  ubuntu:
-    precise: [ruby1.9.3]
 sbcl:
   arch: [sbcl]
   debian: [sbcl]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3561,7 +3561,7 @@ rsync:
   ubuntu: [rsync]
 ruby:
   arch: [ruby]
-  debian: [ruby1.8-dev, libruby1.8, rubygems1.8]
+  debian: [ruby, ruby-dev]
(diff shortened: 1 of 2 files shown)


tfoote:
> What is the reason why ruby.yaml is separate from base.yaml?
We wanted to keep the files from growing too big. The separate file helps isolate a little bit. 
> Has the split any consequences on key lookup or are they treated the same? 
The are treated the same. (techincally the first one listed will win, but if there's no overlapping keys they are equal.) 
> Should all (new) ruby library keys have a ruby- prefix, similar to the ones in python.yaml? Should ruby itself be moved to ruby.yaml?
The python keys are all prefixed with python because that's what upstream debian packaging does. Our default key name is the same as the debian/ubuntu package for simplicity. We should do the same for the ruby packages. Mostly we knew that python was a large set of packages that would be seperable. And we thought separating ruby would also make sense. There's a level that we've figured out how to manage the large file, but I think that keeping it separate makes sense since there's a level that it won't continue to scale in one file. (For example the git blame in github's UI is very slow/gives pink unicorns a lot.) 

I think it would make sense to move the ruby definition itself to the separate file since everything else ruby is already there. That way if you're thinking about ruby anything you have just one source. 

Overall this looks good. It looks like a lot of these rules are quite out of date/non-functional at the moment.

meyerj:
> I think it would make sense to move the ruby definition itself to the separate file since everything else ruby is already there. That way if you're thinking about ruby anything you have just one source.

Done.

Most of these keys have been introduced for the release of the [Orocos Toolchain](https://github.com/orocos-toolchain/). Probably there are not many other users from looking at the history.

tfoote (approved):

PR #13553, opened 2017-01-11
update debian entry for python-click

k-okada:
see https://packages.debian.org/search?keywords=python-click


wjwwood (approved):
It is in jessie-backports, but I assume we can't use/depend on that. It isn't in wheezy, but it is in sid/stretch.

PR #13563, opened 2017-01-12
add debian/fedra for fcn-pip

k-okada:


rosdep/python.yaml
```diff
@@ -701,6 +701,12 @@ python-falcon:
   fedora: [python-falcon]
   ubuntu: [python-falcon]
 python-fcn-pip:
+  debian:
+    pip:
+      packages: [fcn]
```
tfoote: Does the debian rule not have the same depends as the ubuntu one? And if it's also a dependency on osx I suspect fedora needs a depends rule too.

k-okada: thank you. I added depends. just to  make sure that these values are not rosdep key, but actual package name.

rosdep/python.yaml
```diff
@@ -701,6 +701,14 @@ python-falcon:
   fedora: [python-falcon]
   ubuntu: [python-falcon]
 python-fcn-pip:
+  debian:
+    pip:
+      depends: [libhdf5-dev, liblapack-dev]
+      packages: [fcn]
+  fedora:
+    pip:
+      depends: [hdf5-devel, lapack-devel]
```
tfoote: @k-okada they should be rosdep key names, they look a lot like the debian names since that's our default, but even for fedora I think it should be the same libhdf5-dev, liblapack-dev ala https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L1798 and https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L1905

k-okada: I see, but why depend names on osx is different from debian @wkentaro 
anyway, fixed to use rosdep key names.

wkentaro: +1

For mac, we does not have installation method for `liblapack-devel` (required for `numpy`), so we need to install requirements via `gfortran`.

tfoote (approved):

PR #13572, opened 2017-01-13
Adding microstrain to documentation to hydro

bsb808:
I'd like to add my hydro package to ros.org


dirk-thomas:
Hydro is EOL for a long time. Since there is no buildfarm anymore we don't accept changes to its distributions file. In order to pass the CI test you need to remove the Hydro change from this PR.

PR #13574, opened 2017-01-14
slackware desktop-full

nikonikolov:
Added package definitions to enable kinetic for slackware desktop-full installation


rosdep/python.yaml
```diff
@@ -1373,6 +1373,9 @@ python-opencv:
   arch: [opencv, python2-numpy]
   debian: [python-opencv]
   fedora: [opencv-python]
+  slackware:
+    pip:
+      packages: [opencv-python]
```
tfoote: It's too bad this one's not packaged otherwise.

tfoote (approved):

rosdep/python.yaml
nikonikolov: My mistake, opencv package for slackware includes the python bindings. I submitted new pull request.

PR #13579, opened 2017-01-15
add libfreetype6-dev

samuelba:
Is the new key `libfreetype6-dev` ok or should I add `libfreetype6-dev` to the existing `libfreetype6` key?


dirk-thomas (approved):

tfoote (approved):
This looks good. If you're going to be making releases, filling out the debian and fedora dependencies would be helpful.

PR #13585, opened 2017-01-16
Added cheetah dep.

tonybaltovski:


dirk-thomas (commented):
All Python dependencies should go into `rosdep/python.yaml`. Other than that the change looks good. If the package has the same name across all supported Ubuntu distros you don't have to enumerate them explicitly.

tonybaltovski:
Updated, thanks!

rosdep/python.yaml
```diff
@@ -381,6 +381,13 @@ python-catkin-tools:
     pip:
       packages: [catkin_tools]
   ubuntu: [python-catkin-tools]
+cheetah:
+  ubuntu:
+    trusty: [python-cheetah]
+    utopic: [python-cheetah]
+    vivid: [python-cheetah]
+    wily: [python-cheetah]
+    xenial: [python-cheetah]
```
dirk-thomas: You can merge these lines (since they are identical) to:

```
  ubuntu: [python-cheetah]
```

rosdep/python.yaml
```diff
@@ -381,6 +381,13 @@ python-catkin-tools:
     pip:
       packages: [catkin_tools]
   ubuntu: [python-catkin-tools]
+cheetah:
```
dirk-thomas: The key should be prefixed with `python-`.

dirk-thomas: And it needs to be inserted in alphabetical order.

tfoote (approved):

PR #13586, opened 2017-01-16
Add rosdep python rules for defer, kombu, and sortedcontainers

jvgomez:
[PyPi kombu](https://pypi.python.org/pypi/kombu/4.0.2)
[PyPi sortedcontainers](https://pypi.python.org/pypi/sortedcontainers/1.5.7)
[PyPi defer](https://pypi.python.org/pypi/defer/1.0.4)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -588,7 +588,7 @@ python-debian:
     utopic_python3: [python3-debian]
     vivid: [python-debian]
     vivid_python3: [python3-debian]
-python-defer:
+python-defer-pip:
   ubuntu:
     pip:
       packages: [defer]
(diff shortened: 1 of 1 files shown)


dirk-thomas (commented):
Packages which are only available through pip should have the `-pip` suffix. That allows to add the non-pip entry in the future when it becomes available.

jvgomez:
@dirk-thomas sorry about that, didn't realize. I have updated the PR.

dirk-thomas (approved):
No problem. Thanks for the fast update.

A second maintainer will merge this (see https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

PR #13591, opened 2017-01-17
add debian/fedra for chainer

k-okada:


rosdep/python.yaml
```diff
@@ -392,6 +392,12 @@ python-certifi:
     xenial: [python-certifi]
     yakkety: [python-certifi]
 python-chainer:
```
tfoote: Since this is pip only, I'd suggest we update it to use the `-pip` extension as we generalize it.

k-okada: ok, updated

k-okada:
@wkentaro please package.xml in our repo.

PR #13613, opened 2017-01-18
Added php to rosdep rules

kmhallen:
* https://packages.debian.org/jessie/php5
* http://packages.ubuntu.com/trusty/php5
* http://packages.ubuntu.com/xenial/php

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3362,7 +3362,6 @@ php:
   arch: [php]
   debian:
     jessie: [php5]
-    sid: [php7.1]
     stretch: [php7.0]
   fedora: [php]
   ubuntu:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3358,6 +3358,18 @@ pcre-dev:
   ubuntu: [libpcre++-dev]
 phantomjs:
   ubuntu: [phantomjs]
+php:
+  debian:
+    jessie: [php5]
+  ubuntu:
+    precise: [php5]
+    raring: [php5]
+    saucy: [php5]
+    trusty: [php5]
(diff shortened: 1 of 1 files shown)


dirk-thomas (approved):

rosdep/base.yaml
```diff
@@ -3358,6 +3358,18 @@ pcre-dev:
...
+    jessie: [php5]
+  ubuntu:
+    precise: [php5]
+    raring: [php5]
+    saucy: [php5]
+    trusty: [php5]
+    utopic: [php5]
+    vivid: [php5]
+    wily: [php5]
+    xenial: [php]
```
tfoote: @kmhallen thanks for these updates. We also try to keep Fedora rules in sync: like here: https://admin.fedoraproject.org/pkgdb/package/rpms/php/

And looking at debian it looks like php7.0 is available on stretch. And like xenial php is provided by php7.0 http://packages.ubuntu.com/xenial/php7.0

so the rules for xenial and stretch should be php7.0 not just php.

Also would you mind adding yakkety and zesty?

tfoote (changes requested):

kmhallen:
* https://www.archlinux.org/packages/extra/x86_64/php/
* https://packages.debian.org/jessie/php5
* https://packages.debian.org/stretch/php7.0
* https://packages.debian.org/sid/php7.1
* https://admin.fedoraproject.org/pkgdb/package/rpms/php/
* http://packages.ubuntu.com/trusty/php5
* http://packages.ubuntu.com/xenial/php7.0

tfoote (approved):

tfoote:
I removed sid and added yakkety and zesty.

PR #13624, opened 2017-01-19
ros_rpc: 0.0.1-0 in 'indigo/distribution.yaml' [bloom]

130s:
Increasing version of package(s) in repository `ros_rpc` to `0.0.1-0`:

- upstream repository: https://github.com/tork-a/ros_rpc.git
- release repository: https://github.com/tork-a/ros_rpc-release.git
- bloom version: `0.5.23`

## ros_rpc

```
* Initial release into ROS public repository.
* Initial commit.
* Contributors: Isaac I.Y. Saito
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -10777,6 +10777,21 @@ repositories:
       url: https://github.com/ros/ros_realtime.git
       version: hydro-devel
     status: maintained
+  ros_rpc:
+    doc:
+      type: git
+      url: https://github.com/tork-a/ros_rpc.git
+      version: master
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/tork-a/ros_rpc-release.git
+      version: 0.0.1-0
+    source:
+      type: git
+      url: https://github.com/tork-a/ros_rpc.git
+      version: master
+    status: developed
   ros_statistics_msgs:
     doc:
       type: git


wjwwood:
Since this is the first release, I'd like to point out the name is a bit generic and breaks the cardinal rule of not prefixing with `ros_`. Please have a look at: http://www.ros.org/reps/rep-0144.html

I would highly suggest changing the name to be organization specific until it is more generic (I wouldn't have merged the pr just yet, but it already has been).

Also, it would be good to have a better description on the upstream package which differentiates it from existing abilities provided by ROS Services.

dirk-thomas:
@wjwwood please feel free to revert the PR for the stated reason. The results are not yet published so no problem reverting it.

130s:
That's a feedback I was waiting for and I appreciate your prompt response. I'll gladly change the package name once I figure out what is better name that describes the functionality aimed to provide, maybe by the next release that I'm planning real soon.

> changing the name to be organization specific until it is more generic

With this approach, in the very fortunate situation where multiple packages that someone else maintain depend upon my package without me knowing, changing name would cause backlash. So I'd rather choose a name that's not too generic (like the current one) but one that doesn't require future change.

130s:
Oh okay, reverting might even better to avoid useless workload on our precious buildfarm. I'll open a PR.

wjwwood:
@130s I'll handle it, GitHub has a button for that: https://github.com/ros/rosdistro/pull/13625

wjwwood:
> I'd rather choose a name that's not too generic (like the current one) but one that doesn't require future change.

That's your choice but it goes against the (relatively new) recommendation of the Package Naming REP:

> The rules to add those words should be followed in order. For prefixes:
> ...
> - if a package is specialized by an entity (lab, company, ...), prepend the name of the entity. Once the package is commonly used, owned and maintained, that name can be dropped
>
> -- http://www.ros.org/reps/rep-0144.html#naming-rules

In the past many people have not followed this guideline, claiming some pretty high level names. For instance I've done this (with `serial`), but given hindsight I probably would have named it something different.

If it is truly meant to be generic, you can name is something very generic, like `rpc` or `rpc_with_services`, but you ought to be prepared to elevate its design to be generic and reusable, which takes a lot more work usually. This is usually high risk (many packages which aim for this fall short, which is ok), that's why we recommended people start with an organization name as a prefix to start, despite the overhead of people needing to change the name they depend on in the future.

PR #13644, opened 2017-01-20
Fixing libuv-dev rosdep rules

beatrizleon:
I was getting an error when running rosdep in kinetic for libuv-dev. Since xenial it is libuv0.10-dev.
http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=libuv-dev&searchon=names

This pr will fix this.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2862,7 +2862,12 @@ libusb-dev:
 libuv-dev:
   fedora: [libuv-devel]
   ubuntu:
+    precise: [libuv-dev]
+    saucy: [libuv-dev]
     trusty: [libuv-dev]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
This generally looks good. 

The package: http://packages.ubuntu.com/zesty/libuv-dev was changed to a virtual package. The implementation is in http://packages.ubuntu.com/xenial/libuv0.10-dev

However we need to fill in the older platforms. At least saucy through current + precise.

beatrizleon:
Thanks @tfoote for looking into this. I added the older platforms. Let me know if that is ok.

tfoote (approved):

PR #13645, opened 2017-01-20
Add rosaria to kinetic/distribution.yaml

reed-adept:
Add rosaria package entry to kinetic/distribution.yaml


reed-adept:
Thanks Dirk!

PR #13683, opened 2017-01-24
yaml_cpp_0_3: 0.3.1-0 in 'kinetic/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `yaml_cpp_0_3` to `0.3.1-0`:

- upstream repository: https://github.com/stonier/yaml_cpp_0_3.git
- release repository: https://github.com/yujinrobot-release/yaml_cpp_0_3-release.git
- bloom version: `0.5.23`

## yaml_cpp_0_3

```
* refactored the actual library namespace (previously just renamed the library)
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6934,6 +6934,13 @@ repositories:
       url: https://github.com/ethz-asl/ethzasl_xsens_driver.git
       version: master
     status: maintained
+  yaml_cpp_0_3:
+    release:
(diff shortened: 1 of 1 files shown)


stonier:
While it looks like ubuntu now has support for the old version, it doesn't let you develop with both versions simultaneously so this is still useful packaged as a ros 3rd party dependency, with appropriate namespacing for the headers.

PR #13686, opened 2017-01-24
Enable test_pull_requests for all uos repos

mintar:


dirk-thomas:
Have you made sure that the buildfarm user can access the org units for these repos (see http://wiki.ros.org/buildfarm/Pull%20request%20testing)? Otherwise this won't work when being merged.

mintar:
I haven't granted administrative rights to @ros-pull-request-builder, but instead I've set up the web hooks as described on that page. Thanks for asking though, because I forgot to also give push and pull rights to @ros-pull-request-builder .

I've just invited @ros-pull-request-builder to our GitHub group and will grant it push + pull rights on the repos as soon as the invitation is accepted.

dirk-thomas:
I just confirmed the invitation for the @ros-pull-request-builder user.

mintar:
Great, I've granted write access to @ros-pull-request-builder for the concerned repos.

PR #13695, opened 2017-01-24
Added python-cheetah as fedora rosdep.

tonybaltovski:


tfoote (changes requested):
It looks like python-cheetah has an rpm available. Unless there's a reason otherwise we should use that.

https://apps.fedoraproject.org/packages/python-cheetah

tonybaltovski:
You are right, I only searched for cheetah which had no results.

tfoote:
Yeah, not the best search interface. I think I found the same results when the other rules were contributed.

tfoote (approved):

PR #13714, opened 2017-01-25
RCLL related packages for Jade and Kinetic

timn:
This pull request adds rcll_ros, rcll_ros_msgs, and rcll_refbox_peer to the Jade and Kinetic distributions (it had been in indigo before). I have fixed an error that prevented proper builds (missing dependencies for message generation).


timn:
That was quick. Thank you!

PR #13716, opened 2017-01-25
python-termcolor for debian

stonier:
So packages can be built on the build farm that have this as a dependency


dirk-thomas (approved):

tfoote:
I removed sid since we don't want to target the rolling distro.

stonier:
Ok, got it.

PR #13717, opened 2017-01-26
add nvidia-cuda-dev to nvidia-cuda

furushchev:
Currently dependency `nvidia-cuda` installs `nvidia-cuda-tool` which has only libraries.
This pull request adds `nvidia-cuda-dev` which includes header files for cuda to `nvidia-cuda`.


tfoote:
We're working toward improving our rule cleanlyness and separating runtime and development/build packages. Could you instead submit these as a new rule nvidia-cuda-dev so that packages can differentiate the runtime and build time dependencies.

furushchev:
@tfoote Thank you for quick review and advice! OK. I'll create a new pull request.

PR #13724, opened 2017-01-26
Added the tensorflow pip package

peci1:
https://pypi.python.org/pypi/tensorflow/0.12.1.

Google's framework for neural networks. This only provides the python interface; the C++ API is trickier to get.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2560,6 +2560,10 @@ python-telegram-bot:
   ubuntu:
     pip:
       packages: [python-telegram-bot]
+python-tensorflow:
+  ubuntu:
+    pip:
+      packages: [tensorflow]
 python-termcolor:
   debian:
     jessie: [python-termcolor]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2560,6 +2560,10 @@ python-telegram-bot:
   ubuntu:
     pip:
       packages: [python-telegram-bot]
+python-tensorflow:
```
tfoote: Please use `-pip` extension for pip only packages. Pip is often well ahead of the system packages when they eventually appear in the future which can cause future breakage.

tfoote (changes requested):
Since it's pip only, it might be helpful for others to add rules for other common platforms such as fedora, debian and osx.

peci1:
Updated.

rosdep/python.yaml
```diff
@@ -2560,6 +2560,19 @@ python-telegram-bot:
   ubuntu:
     pip:
       packages: [python-telegram-bot]
+tensorflow-pip:
```
dirk-thomas: Should the key be `python-tensorflow-pip`?

peci1: Ahh, sorry :)

dirk-thomas (approved):

tfoote (approved):

PR #13727, opened 2017-01-26
Added can-utils for wily and xenial.

tonybaltovski:


dirk-thomas (approved):
Since all names are the same I would suggest to remove the specific Ubuntu distros names to avoid the maintenance effort on the future.

tonybaltovski:
It is missing is precise.  Not sure if it that matters any more.

dirk-thomas:
No non-EOL ROS distro supports "precise" anymore. Therefore I wouldn't spend effort on maintaining this entry but collapse them.

tonybaltovski:
Sounds good, thanks.

tfoote (approved):
I made the debian rule more generic too. Since can-utils is available for stretch and sid too.

https://packages.debian.org/stretch/can-utils

It's good to go once ci passes.

tonybaltovski:
Are packages in testing available by default in Debian? Jessie was the only one with can-utils in stable.

tfoote:
Yes, Stretch itself is "testing" at the moment. Jessie is the "stable" release. Eventually Stretch will transition from "testing" to "stable" and jessie will be eol'd.

tonybaltovski:
Perfect, thanks!

PR #13767, opened 2017-01-31
[Indigo][moveit] Add/update metapackages.

130s:
- Moving existing metapackages from an individual repo to `moveit` repo (for Indigo).
- Adding a new metapackage (IJK).

**Question**: in both cases above, `bloom` has not yet been run (ie. release repo not updated). I'm afraid this PR would fail without those packages being bloomed? (If so I think I'll have to close this).

(Changes here is discussed in https://github.com/ros-planning/moveit_metapackages/issues/11)

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -6391,6 +6391,8 @@ repositories:
       - moveit_controller_manager_example
       - moveit_core
       - moveit_fake_controller_manager
+      - moveit_full
+      - moveit_full_pr2
       - moveit_kinematics
       - moveit_planners
       - moveit_planners_ompl
@@ -6407,6 +6409,7 @@ repositories:
       - moveit_ros_robot_interaction
       - moveit_ros_visualization
       - moveit_ros_warehouse
+      - moveit_runtime
       - moveit_setup_assistant
       - moveit_simple_controller_manager
       tags:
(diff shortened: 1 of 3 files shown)


130s:
Closing. I assume moving packages to another repo still requires release artifact generated in the new release repo and that's why [package.xml is not found](https://github.com/ros-planning/moveit/blob/be24a76ae73a3290f4d5681d727b93b1b1167747/moveit_runtime/package.xml) for the moved packages. I'll bloom shortly.

130s:
tldr; Please review. The change is for indigo only.

The packages listed in the change are already moved to a new devel repository, but `bloom` fails because they are already listed in rosdistro within the old repo. Therefore, in order to complete the move to a new devel repo, they need to be removed from rosdistro. Then `bloom` needs to be run shortly to keep the period where those packages in transition are not available as short as possible.

I removed the change from J and K, which can be done by bloom only (as already done for K https://github.com/ros/rosdistro/pull/13809).

dirk-thomas:
What should be reviewed here? The ticket has been closed.

wjwwood:
@130s this pr is closed?

130s:
Reopen button is disabled so that I can't reopen. Maybe I'll just open a new PR instead of digging further for now.

PR #13768, opened 2017-01-31
stf: 0.1.0-0 in 'indigo/distribution.yaml' [bloom]

jstnhuang:
Increasing version of package(s) in repository `stf` to `0.1.0-0`:

- upstream repository: https://github.com/jstnhuang/stf.git
- release repository: https://github.com/jstnhuang-release/stf-release.git
- bloom version: `0.5.23`

## stf

```
The initial release of stf.
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -13174,6 +13174,21 @@ repositories:
       type: git
       url: https://github.com/srv/stereo_slam.git
       version: indigo
+  stf:
+    doc:
+      type: git
+      url: https://github.com/jstnhuang/stf.git
+      version: 0.1.0
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/jstnhuang-release/stf-release.git
+      version: 0.1.0-0
+    source:
+      type: git
+      url: https://github.com/jstnhuang/stf.git
+      version: 0.1.0
+    status: developed
   straf_recovery:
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -13174,6 +13174,21 @@ repositories:
       type: git
       url: https://github.com/srv/stereo_slam.git
       version: indigo
+  stf:
+    doc:
+      type: git
+      url: https://github.com/jstnhuang/stf.git
+      version: 0.1.0
```
dirk-thomas: This needs to refer to a branch name. I your case `indigo-devel`.

Same for the `source` entry.

tfoote:
This package also doesn't have a descriptive name. Our naming conventions are here: http://wiki.ros.org/ROS/Patterns/Conventions 

From reading the package name 'stf' I had no idea what the package does. Even after I read the full name in the documentation Static Transformations Library, not just the acronym I still couldn't guess what the package did. Static transforms already have meaning in the ROS ecosystem which is also a problem.

Please consider renaming the package to something slightly more descriptive. Looking at the functionality it appears to be a forward kinematics library based on a graph representation. Potentially a name like `graph_kinematics` might make sense. Or `static_graph_kinematics`.

jstnhuang:
Thanks for the feedback. I'll look into changing the name and resubmit.

PR #13769, opened 2017-02-01
add fshap and mono

k-okada:
- add fsharp and mono-complete

cc: @YuyaNagamatsu

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -815,6 +815,9 @@ freetype:
   fedora: [freetype]
   gentoo: [media-libs/freetype]
   macports: [freetype]
+fsharp:
+  debian: [fsharp]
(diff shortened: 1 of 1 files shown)


dirk-thomas (approved):

wjwwood:
It would be nice to have the fedora entries, but after looking, I don't see an obvious matching set that doesn't involve something like PPA's.

tfoote:
@cottsay @jpgr87 any insight?

jpgr87:
There's nothing analogous to [mono-complete](http://packages.ubuntu.com/yakkety/mono-complete), you'd have to list the corresponding [mono packages](https://apps.fedoraproject.org/packages/s/mono) in fedora.

fsharp isn't packaged, and I don't see anyone working on it.

PR #13774, opened 2017-02-01
Added libi2c-dev rosdep rule

amiller27:
This package includes headers that allow for i2c devices to be accessed from userspace programs.

The particular use case that this was necessary for is a ROS node that fetches data from a Lidar-Lite v2 laser rangefinder.

On Arch, the required header is contained in this package:

https://www.archlinux.org/packages/core/x86_64/linux-api-headers/

On Debian and Ubuntu respectively:

https://packages.debian.org/jessie/libi2c-dev

https://packages.ubuntu.com/xenial/libi2c-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1811,6 +1811,10 @@ libhdf5-dev:
   debian: [libhdf5-dev]
   fedora: [hdf5-devel]
   ubuntu: [libhdf5-dev]
+libi2c-dev:
+  arch: [linux-api-headers]
+  debian: [libi2c-dev]
+  ubuntu: [libi2c-dev]
 libicu-dev:
   fedora: [libicu-devel]
   ubuntu: [libicu-dev]


wjwwood (approved):
It would be nice to have the fedora entries for these rules, but what is there looks ok to me.

amiller27:
I looked for the fedora version, and didn't seem to find one.  Most I could find was this old bug report: https://bugzilla.redhat.com/show_bug.cgi?id=484128

rosdep/base.yaml
```diff
@@ -1811,6 +1811,16 @@ libhdf5-dev:
...
+libi2c-dev:
+  arch:
+    pacman:
+      packages: [linux-api-headers]
+  debian:
+    apt:
+      packages: [libi2c-dev]
+  ubuntu:
+    apt:
+      packages: [libi2c-dev]
```
dirk-thomas: The `debian` and `ubuntu` entry doesn't need to call out `apt` and `packages` but can directly list the packages: `[libi2c-dev]` (see other entries for comparison).

I assume for `arch` it is the same but @tfoote will know for sure.

amiller27: I know I saw people doing it without that stuff for arch in other places, I can get rid of it.  I mostly figured it was better to be more explicit

tfoote: We prefer the most compact version since it usually requires the least maintenance and is the most readable. And it's not ambiguous. Every platform has a default package manager.

amiller27:
Fixed the formatting, it's using the compact version now

PR #13780, opened 2017-02-02
Adding ubuntu xenial `fltk` dependency

rdelfin:


dirk-thomas:
Please also add an entry for `wily` otherwise you won't be able to release for Kinetic.

PR #13782, opened 2017-02-03
Add 'nvidia-cuda-ocl'

furushchev:
## tl;dr
This pull request adds `nvidia-cuda-ocl` dependency target.
This is assumed to install `ocl-icd-opencl-dev` at first, then install `nvidia-cuda-toolkit`.

## Description

rosdep has install target `nvidia-cuda` which installs `nvidia-cuda-toolkit` via `apt-get`.
`nvidia-cuda-toolkit` has a conditional dependency to `nvidia-opencl-dev | opencl-dev`.
`nvidia-opencl-dev` package depends on `nvidia-340` which is a legacy driver for nvidia video card and if this is installed all other newer drivers are removed instead.
Since this driver does not support the latest nvidia video cards any more, this causes black screen on reboot on some computers after installation.
The dependency `nvidia-opencl-dev | opencl-dev` means that if `opencl-dev` is already installed before installing `nvidia-cuda-toolkit`, then `nvidia-opencl-dev` is not installed, which means we can keep both `nvidia-cuda-toolkit` and, for example, `nvidia-367` (> 340).
`opencl-dev` is actually a virtual package and is `nvidia-opencl-dev` or `ocl-icd-opencl-dev`.
So to avoid this problem, I'd like to add a dependency that installs `ocl-icd-opencl-dev` first and then installs `nvidia-cuda-toolkit`.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3315,6 +3315,19 @@ nvidia-cuda-dev:
   debian: [nvidia-cuda-dev]
   gentoo: [nvidia-cuda-dev]
   ubuntu: [nvidia-cuda-dev]
+nvidia-cuda-icd:
+  debian:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
+  gentoo:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
+  ubuntu:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
+ocl-icd-opencl-dev:
+  debian: [ocl-icd-opencl-dev]
+  ubuntu: [ocl-icd-opencl-dev]
 omniidl:
   arch: [omniorb]
   debian: [omniidl, omniorb-idl]


furushchev:
refer: https://github.com/tork-a/darknet-release/issues/6

tfoote:
The apt installer does not support depends. It is expected that apt/dpkg will manage dependencies and ordering of installs itself. 

Pip does support depends for cross package manager dependencies. (aka in needs c libraries to build on top of). This should be filed upstream against the nvidia package. if it can cause machines not to boot it should get some attention.

furushchev:
@tfoote Thank you for review and advice! I reported this issue to "Ubuntu Developers Discuss` mailing list and am now awaiting response.
And I'd like to note that installing by `apt-get install nvidia-cuda-toolkit ocl-icd-opencl-dev` solves this issue but when installing with `rosdep`, we cannot specify multiple target packages nor its order, I think.

tfoote:
You can specify multiple packages but not the order of installation. Packages are not supposed to be order dependent.

furushchev:
@tfoote Yes, for example, when we define `package.xml` as:

```xml
<build_depend>pkg_a</build_depend>
<build_depend>pkg_b</build_depend>
```

Current `rosdep` executes:

```bash
sudo apt-get install pkg-a
sudo apt-get install pkg-b
```

or 

```bash
sudo apt-get install pkg-b
sudo apt-get install pkg-a
```

but not

```bash
sudo apt-get install pkg-a pkg-b
```

right?

This might causes problem when a package depends on some packages using `|` operator.

tfoote:
Yes, we install packages individually so you get an error message about the exact package that failed to install. Otherwise if you `sudo apt-get install foo bar baz bim bop boo` you'll get an error that one of them was uninstallable, but it's not clear which one. And we haven't added the complexity of then retrying individually. 

It might cause trouble, but if properly packaged should not. If you have an  `|` operator in the dependencies. It should work for both `a` or `b`, not break your system. If a and b are both in the `|` list then the last one should win. One of the aspects of rosdep rules is that we work to maintain them such that they should never be mutually exclusive with another rosdep rule. If you get that sort of thing you end up with rule a fighting with rule b. And each time you rerun rosdep it switches a to b, or b to a.

k-okada:
@furushchev can you create another PR just contains 
```
+ocl-icd-opencl-dev:
 +  debian: [ocl-icd-opencl-dev]
 +  ubuntu: [ocl-icd-opencl-dev]
```
*

furushchev:
@k-okada OK => #13903

rosdep/base.yaml
```diff
@@ -3315,6 +3315,19 @@ nvidia-cuda-dev:
...
+nvidia-cuda-icd:
+  debian:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
+  gentoo:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
+  ubuntu:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
```
Alessandro-Barbieri: `dev-util/nvidia-cuda-toolkit` if you want be more precise

rosdep/base.yaml
```diff
@@ -3315,6 +3315,19 @@ nvidia-cuda-dev:
   debian: [nvidia-cuda-dev]
   gentoo: [nvidia-cuda-dev]
   ubuntu: [nvidia-cuda-dev]
+nvidia-cuda-icd:
+  debian:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
+  gentoo:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
```
Alessandro-Barbieri: dev-util/nvidia-cuda-toolkit if you want be more precise

rosdep/base.yaml
```diff
@@ -3315,6 +3315,19 @@ nvidia-cuda-dev:
   debian: [nvidia-cuda-dev]
   gentoo: [nvidia-cuda-dev]
   ubuntu: [nvidia-cuda-dev]
+nvidia-cuda-icd:
+  debian:
+    depends: [ocl-icd-opencl-dev]
+    packages: [nvidia-cuda-toolkit]
+  gentoo:
+    depends: [ocl-icd-opencl-dev]
```
Alessandro-Barbieri: this is totally wrong (there is no `ocl-icd-opencl-dev` in gentoo)

```
 * dependency graph for dev-util/nvidia-cuda-toolkit-8.0.61
 `--  dev-util/nvidia-cuda-toolkit-8.0.61  [~amd64 keyword]
   `--  sys-devel/gcc-5.4.0-r3  (>=sys-devel/gcc-4.7) amd64  [cxx]
   `--  x11-drivers/nvidia-drivers-381.22  (>=x11-drivers/nvidia-drivers-375.26) NVIDIA-r2 license(s)  [X uvm]
   `--  sys-libs/libtermcap-compat-2.0.8-r3  (sys-libs/libtermcap-compat) amd64
   `--  sys-libs/ncurses-5.9-r101  (sys-libs/ncurses) amd64  [tinfo]
   `--  virtual/jre-1.8.0-r1  (>=virtual/jre-1.6) amd64
```

Alessandro-Barbieri (changes requested):

rosdep/base.yaml
Alessandro-Barbieri: Maybe one of these?
```
eix -c opencl
[N] app-eselect/eselect-opencl (1.1.0-r1): Utility to change the OpenCL implementation being used
[N] dev-libs/opencl-clhpp [2] (~2.0.10): OpenCL Host API C++ bindings (cl.hpp and cl2.hpp)
[N] dev-python/pyopencl (2014.1): Python wrapper for OpenCL
[N] games-action/openclonk (7.0): A free multiplayer action game where you control clonks
[N] media-gfx/openclipart (0.20): Open Clip Art Library (openclipart.org)
[N] net-wireless/cpyrit-opencl (~0.5.0): A sub-package that adds OpenCL-capability to Pyrit
[N] sci-CRAN/OpenCL [1] (~0.1.3-r1): Interface allowing R to use OpenCL
[N] virtual/opencl (0-r4): Virtual for OpenCL implementations
[1] "R_Overlay" /var/lib/layman/R_Overlay
[2] "science" /var/lib/layman/science

Found 8 matches
```

dirk-thomas:
What is the status of this ticket? Will it be iterated on in the near future or should it be closed for now?

tfoote:
I think correct fixes for this needs to go upstream. I'll close this since I don't think there's more we can do at the moment.

k-okada:
memo : https://github.com/ros-infrastructure/rosdep/pull/545 changed `rosdep install` behavior, now it installes as we write on depends (correct?)

PR #13865, opened 2017-02-11
Add numpy-stl pip package

anassinator:
https://pypi.python.org/pypi/numpy-stl

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1367,6 +1367,13 @@ python-numpy:
     vivid: [python-numpy]
     wily: [python-numpy]
     xenial: [python-numpy]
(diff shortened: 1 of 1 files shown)


tfoote (approved):

rosdep/python.yaml
```diff
@@ -1367,6 +1367,13 @@ python-numpy:
     vivid: [python-numpy]
     wily: [python-numpy]
     xenial: [python-numpy]
+python-numpy-stl:
```
dirk-thomas: The key should have a `-pip` suffix to leave room for this name if the package becomes available through Debian packages in the future.

anassinator: Fixed!

PR #13881, opened 2017-02-13
add python-pyudev for U->Y

k-okada:
see http://packages.ubuntu.com/search?keywords=python-pyudev


dirk-thomas:
The order is not alphabetical and `wily` is misspelled.

k-okada:
Sorry and fixed now

PR #13884, opened 2017-02-14
add entries for python(3)-catkin-pkg-modules

dirk-thomas:


wjwwood:
Should this not use the `_py3` ubuntu distros like some of the other rules? For example look at how `python-yaml` is defined.

You could also include the `python3-caktin-pkg-modules` rule, but adding the `_py3` entries to the `python-catkin-pkg-modules` entry would just make it more useful I think.

wjwwood:
edit, that should be `_python3`, as in `xenial_python3`, rather than `_py3`.

dirk-thomas:
For these new keys I need two explicit keys rather then relying in the distro name (and its potential suffix `_python3`) to choose for me. Otherwise we would need at least three entries: one for Python 2, one for Python 3, and one selecting based on the distro name. Since I don't have a use case for the last case I didn't add it in this PR. I am not sure if it should be named `python2-catkin-pkg-modules` then but that wouldn't match any other key right now.

wjwwood:
Seems to me, that if the prefix is `python-` as all the others ones, then it should behave like them (i.e. it should have the `_python3` distros specified). If you explicitly don't want that, i.e. when someone asks for `python-catkin-pkg-modules` they get the python2 `python-catkin-pkg-modules` debian even when the os is set to `xenial_python3` (for example), then I'd say a different name, like `python2-catkin-pkg-modules` might be appropriate.

dirk-thomas:
I actually only need to reference the Debian package name directly. Since I don't have a use case for the rosdep key and can specify the Debian package instead I will just close this. No need to discuss the `_python3` suffix for this.

PR #13886, opened 2017-02-14
update wxwidgets, from xenial we need to use gtk3.0

k-okada:
update wxwidgets, from xenial we need to use gtk3.0

I decied to switch to gtk 3.0 from xenial, as wxpython entry do
http://packages.ubuntu.com/trusty/libwxgtk3.0-dev


dirk-thomas (approved):
This will affect already released packages depending in this key. They need to be rebloomed to get the new mapping.

k-okada:
yes, and these packages should be failed within the build firm.
http://build.ros.org/job/Kbin_uX64__jsk_rviz_plugins__ubuntu_xenial_amd64__binary/3/console#console-section-12


--
◉ Kei Okada


2017-02-14 10:55 GMT+09:00 Dirk Thomas <notifications@github.com>:

rosdep/base.yaml
```diff
@@ -4160,7 +4160,23 @@ wxwidgets:
   macports: [wxWidgets-python]
   opensuse: [wxGTK-devel]
   rhel: [wxGTK-devel]
-  ubuntu: [libwxgtk2.8-dev]
+  ubuntu:
```
tfoote: This line is is here twice.

tfoote (changes requested):
This should also be updated for debian: It lookslike 3.0 was published as of Jessie: https://packages.debian.org/jessie/libwxgtk3.0-dev
And 2.8 was last in wheezy: https://packages.debian.org/wheezy/libwxgtk2.8-dev

http://packages.ubuntu.com/trusty/libwxbase2.8-dev vs http://packages.ubuntu.com/trusty/libwxgtk3.0-dev

wjwwood:
Also, perhaps a bit off-topic, but why does your `jsk_rviz_plugins` package depend on `wx`?

rosdep/base.yaml
k-okada: @tfoote sorry, fixed now.

k-okada:
> Also, perhaps a bit off-topic, but why does your jsk_rviz_plugins package depend on wx?

@wjwwood Good question, may be we do not need that. I just afraid something bad happens by changing dependency. I'll try to remove that https://github.com/jsk-ros-pkg/jsk_visualization/pull/664
https://github.com/jsk-ros-pkg/jsk_visualization/commit/a7f3f76f86efcc73069eb1f840ccab90f5c12e00

wjwwood:
I mean, do whatever works for you. I don't mind that you're using `wx`, and I think we should fix the rosdep rule. Just curious.

rosdep/base.yaml
tfoote: +1

rosdep/base.yaml
```diff
@@ -4153,14 +4153,32 @@ wxpython:
     xenial: [python-wxgtk3.0]
 wxwidgets:
   arch: [wxgtk]
-  debian: [libwxgtk2.8-dev]
+  debian:
+    jessie: [libwxgtk3.0-dev]
+    stretch: [libwxgtk3.0-dev]
```
tfoote: Thanks for the debian updates. I removed the sid entry since we don't target sid because it's a rolling target.

tfoote (approved):

PR #13912, opened 2017-02-16
add catkin to lunar

dirk-thomas:


dirk-thomas:
Ignoring CI result since it failed due to ros-infrastructure/rosdistro#94.

PR #13936, opened 2017-02-21
Update distribution.yaml

neckutrek:


neckutrek:
I removed the release of the hiqp metapackage for now as it fails to compile on jenkins. It depends on a proprietary package, this dependency will later be removed.

PR #13953, opened 2017-02-21
gennodejs: 2.0.1-0 in 'lunar/distribution.yaml' [bloom]

IanTheEngineer:
Increasing version of package(s) in repository `gennodejs` to `2.0.1-0`:

- upstream repository: https://github.com/RethinkRobotics-opensource/gennodejs.git
- release repository: https://github.com/RethinkRobotics-release/gennodejs-release.git
- bloom version: `0.5.23`

## gennodejs

```
* Fix an issue uncovered in checking fixed size message fields
* Contributors: Chris Smith
```

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -69,7 +69,7 @@ repositories:
         release: release/lunar/{package}/{version}
       url: https://github.com/RethinkRobotics-release/gennodejs-release.git
       version: 2.0.1-0
-    status: developed
+    status: maintained
   genpy:
     doc:
       type: git
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -69,6 +69,11 @@ repositories:
         release: release/lunar/{package}/{version}
       url: https://github.com/RethinkRobotics-release/gennodejs-release.git
       version: 2.0.1-0
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/RethinkRobotics-opensource/gennodejs.git
+      version: kinetic-devel
     status: maintained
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please also add a `source` entry so that users can easily get the sources using the `rosinstall_generator`. If you don't want a devel job for the repo you can add `test_commits: false` to the source entry.

IanTheEngineer:
Yep! Should be all good now. Lemme know if there's anything else we need to do for the release :)

lunar/distribution.yaml
```diff
@@ -59,6 +59,22 @@ repositories:
...
+      type: git
+      url: https://github.com/RethinkRobotics-opensource/gennodejs.git
+      version: kinetic-devel
+    release:
+      tags:
+        release: release/lunar/{package}/{version}
+      url: https://github.com/RethinkRobotics-release/gennodejs-release.git
+      version: 2.0.1-0
+    source:
+      test_pull_requests: true
```
dirk-thomas: Have you made sure the buildfarm has access to your repo / org unit (see http://wiki.ros.org/buildfarm/Pull%20request%20testing)?

IanTheEngineer: Excellent, thanks for that resource! I just added the webhook for issue_comment and pull_request, so we should be good to go!

dirk-thomas: I haven't seen that you gave permission to the @ros-pull-request-builder yet. So the PR integration might not be complete yet.

dirk-thomas:
Thanks for the quick PR.

PR #14005, opened 2017-02-23
volksbot_kinetic: Add doc job for kinetic

mintar:
Don't add source job, because it fails anyway (requires external
libepos2-dev).


tfoote:
Note it's recommended to have the source entry and disable the job using. `test_commits: false` inside the source block. This will allow the source checkout tools to work, but turn off the jenkins job.

mintar:
Ah, good to know, thanks! Done here: https://github.com/ros/rosdistro/pull/14013

PR #14025, opened 2017-02-24
added a few coinor-* packages to base.yaml

ipa-rmb:


dirk-thomas (approved):

rosdep/base.yaml
```diff
@@ -501,6 +501,16 @@ cmake:
...
+coinor-libcbc-dev:
+  ubuntu: [coinor-libcbc-dev]
+coinor-libcbc3:
+  ubuntu: [coinor-libcbc3]
+coinor-libcgl-dev:
+  ubuntu: [coinor-libcgl-dev]
+coinor-libclp-dev:
+  ubuntu: [coinor-libclp-dev]
+coinor-libcoinutils-dev:
+  ubuntu: [coinor-libcoinutils-dev]
```
mikaelarguedas: should we add debian and fedora rules for these too?

tfoote: If these want to be used in a release debian and fedora will need to be added.

PR #14040, opened 2017-02-28
update rule for libjpeg

wjwwood:
See:

- http://packages.ubuntu.com/yakkety/libjpeg-dev
- http://packages.ubuntu.com/zesty/libjpeg-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1901,6 +1901,8 @@ libjpeg:
     vivid: [libjpeg-dev]
     wily: [libjpeg-dev]
     xenial: [libjpeg-dev]
+    yakkety: [libjpeg-dev]
+    zesty: [libjpeg-dev]
 libjson-glib:
   arch: [json-glib]
   debian: [libjson-glib-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1880,6 +1880,7 @@ libjpeg:
   arch: [libjpeg-turbo]
   debian:
     jessie: [libjpeg62-turbo-dev]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@wjwwood could you add debian rules too please?

wjwwood:
I guess that's the right `libjpeg` rule for stretch?

mikaelarguedas (commented):
I think it is: libjpeg62-turbo-dev (https://packages.debian.org/stretch/libdevel/libjpeg62-turbo-dev) that provide the virtual package libjpeg-dev

wjwwood:
But that's the same thing we're doing in Ubuntu:

http://packages.ubuntu.com/yakkety/libjpeg-dev

I don't know which we should prefer, because you could also do `libjpeg8-dev`.

mikaelarguedas:
hmm good question. I guess it comes down to: do we prefer to have keys that are the same across platforms or call out versions.
I'm guessing based on the other rules that you're right we should use `libjpeg-dev` for both. Is the consensus to use the "generic" name as long as it doesn't result in ambiguity to which package provides it? (like it would be the case for `libtlfk-dev` for example provided by both  `libfltk1.3-dev` and `libfltk1.1-dev` )

mikaelarguedas (approved):

wjwwood:
Yeah, but in the case of `libfltk-dev` we explicitly want the non-default version (1.3 rather than 1.1). I don't care what we do here, but I'd want to make it consistent between ubuntu and debian.

wjwwood:
We decided to leave it as proposed here.

PR #14041, opened 2017-02-28
update rule for libfltk-dev

wjwwood:
See:

- http://packages.ubuntu.com/yakkety/libfltk1.3-dev
- http://packages.ubuntu.com/zesty/libfltk1.3-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1580,7 +1580,6 @@ libflann-dev:
   ubuntu: [libflann-dev]
 libfltk-dev:
   arch: [fltk]
-  debian: []
   debian:
     jessie: [libfltk1.3-dev]
     sid: [libfltk1.1-dev]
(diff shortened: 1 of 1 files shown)


tfoote (approved):

mikaelarguedas:
given that all distros we target for kinetic and lunar do provide fltk1.3. Should we add a [stretch](https://packages.debian.org/testing/libfltk1.3-dev) and [jessie](https://packages.debian.org/stable/libfltk1.3-dev) rule for libfltk-1.3-dev so that we target the same version for all supported platforms?

wjwwood:
I guess so. I'm not sure how much it matters, IIRC the reason for the split in ubuntu rules was that 1.1 caused a dependency conflict of some sort that 1.3 addressed. 1.1 is still in ubuntu too:

- http://packages.ubuntu.com/yakkety/libfltk1.1-dev
- http://packages.ubuntu.com/zesty/libfltk1.1-dev

wjwwood:
Ok, I updated it so that Jessie and Stretch uses `libfltk3.3-dev`, though I'm concerned about updating the Jessie rule after it might be in use. I do think this is "better", but we are risking breaking it I think.

wjwwood:
Ok, this should be good to go with 6cf3796.

mikaelarguedas (approved):
lgtm, thanks for iterating on this

PR #14073, opened 2017-03-02
Enable test_pull_requests for phidgets_drivers

mintar:


dirk-thomas:
I am not sure you have followed these instructions to provide the necessary access for the ROS buildfarm GitHub user to the org unit yet?

mintar:
No, I haven't. I was hoping that the Jenkins user already has access to the ros-drivers org. Do you have any way to check?

mintar:
I've added another commit to move phidgets_drivers on indigo to ros-drivers as well. Everything else would get confusing anyway. :)

dirk-thomas:
Looks good. Thanks for iterating.

PR #14077, opened 2017-03-03
Adding rapidjson-dev Ubuntu package

pbeeson:


tfoote (approved):
I added debian and fedora rules. The Fedora one oddly doesn't have a -devel version. 

http://packages.ubuntu.com/xenial/libdevel/rapidjson-dev
https://apps.fedoraproject.org/packages/rapidjson
https://apps.fedoraproject.org/packages/rapidjson/overview/

PR #14092, opened 2017-03-06
Updating release version of move_base_to_manip.

AndyZe:


kinetic/distribution.yaml
```diff
@@ -3038,7 +3038,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/UTNuclearRoboticsPublic/move_base_to_manip-release.git
-      version: 1.0.6-2
+      version: 1.0.9
```
dirk-thomas: The release version must contain the Debian increment.

kinetic/distribution.yaml
```diff
@@ -3038,7 +3038,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/UTNuclearRoboticsPublic/move_base_to_manip-release.git
-      version: 1.0.6-2
+      version: 1.0.9-0
```
dirk-thomas: The release repo does not contain a release with that version number for Kinetic. Please make sure that you have run `bloom`.

AndyZe: I think I'm encountering a problem since splitting 'master' into 'indigo' and 'kinetic' branches.' I just ran bloom again, creating tag 1.0.10 for kinetic. But that tag already exists for indigo. Run it again to create tag 1.0.11? (see attached)
1_0_10_already_exists

dirk-thomas: You should first fix the tracks to point to the correct branch for a specific ROS distro (otherwise it uses the default branch which is wrong for at least one distro):

* https://github.com/UTNuclearRoboticsPublic/move_base_to_manip-release/blob/c247cde96f821d83635dac7a59b177efc9c8c9ee/tracks.yaml#L15
* https://github.com/UTNuclearRoboticsPublic/move_base_to_manip-release/blob/c247cde96f821d83635dac7a59b177efc9c8c9ee/tracks.yaml#L39

Then you can run `bloom-release` again.

dirk-thomas:
The patch looks good now.

We are waiting for another sync before merging this though...

AndyZe:
Closing because it's superseded by PR #14129

PR #14113, opened 2017-03-06
Added python-redis for 14.10 -> 16.10

josephduchesne:
Updated rosdep python package list for ubuntu 14.10 through 16.10 for the python-redis package.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2051,6 +2051,7 @@ python-redis:
     wily: [python-redis]
     xenial: [python-redis]
     yakkety: [python-redis]
+    zesty: [python-redis]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@josephduchesne Is it possible to add the entry for zesty (17.04) as well ?

mikaelarguedas (approved):
thanks @josephduchesne for the quick iteration

dirk-thomas:
Since all the Ubuntu entries are the same it would be easier to just fold it into `ubuntu: [python-redis]` to avoid any future work on this.

dirk-thomas:
Thank you for iterating on this.

PR #14114, opened 2017-03-06
add doc and source entries for metapackages in Indigo and Jade

dirk-thomas:
In order to extract metapackages for the wiki as well as to work with `rosinstall_generator`.


mikaelarguedas (approved):
lgtm

PR #14116, opened 2017-03-06
remove explicit call for apt

mikaelarguedas:
apt being the default for bedian and ubuntu, there is no need to call out apt as the package manager.

Note: this doesn't collapse existing rules when identical but can be added to this PR if deemed relevant


dirk-thomas (approved):

tfoote:
If we want to collapse we can do that separately, and this will make it easier to review too.

PR #14126, opened 2017-03-07
Added pocketsphinx to base.yaml

beatrizleon:
It is available after xenial:
http://packages.ubuntu.com/search?keywords=pocketsphinx+&searchon=sourcenames

https://admin.fedoraproject.org/pkgdb/package/rpms/pocketsphinx/

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3358,6 +3358,14 @@ pkg-config:
 pocketsphinx:
   fedora: [pocketsphinx]
   ubuntu:
+    precise: [libsphinxbase1]
+    quantal: [libsphinxbase1]
+    raring: [libsphinxbase1]
+    saucy: [libsphinxbase1]
+    trusty: [libsphinxbase1]
+    utopic: [libsphinxbase1]
+    vivid: [libsphinxbase1]
+    wily: [libsphinxbase1]
     xenial: [pocketsphinx]
     yakkety: [pocketsphinx]
     zesty: [pocketsphinx]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3355,6 +3355,12 @@ pkg-config:
     slackpkg:
       packages: [pkg-config]
   ubuntu: [pkg-config]
+pocketsphinx:
+  fedora: [pocketsphinx]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The link you posted lists it as being available for Precise and Trusty too. And the non-LTS distros in between likely also contain the package.

beatrizleon:
It didn't exits as such `pocketsphinx`. I get `Unable to locate package pocketsphinx` in trusty. I think it was needed to install it from sources:
https://github.com/ros/rosdistro/blob/master/indigo/distribution.yaml#L8516
which is no longer available in kinetic/distribution.yaml

dirk-thomas:
Which part of the package are you interested in? The shared libraries? They are available as http://packages.ubuntu.com/trusty/libsphinxbase1

beatrizleon:
We have a package on xenial-kinetic that `<build_depend>pocketsphinx</build_depend>` and rosdep is not able to find it: `Cannot locate rosdep definition for [pocketsphinx]`

beatrizleon:
Yes, exactly. That is why we are doing the PR

dirk-thomas:
Ok, but you won't be able to release you package afterwards for Kinetic. Kinetic targets Xenial as well as Wily and there is no mapping for Wily. Therefore I suggest to add mapping for additional Ubuntu distros. The key can map to a different package name for older Ubuntu distros if necessary.

beatrizleon:
Like this?

dirk-thomas (approved):
LGTM

beatrizleon:
Thanks!

k-okada:
@beatrizleon , @dirk-thomas we used package with `  <run_depend>pocketsphinx</run_depend>
` https://github.com/tork-a/ros_seminar/blob/master/ros_seminar_navigation/my_navigation/package.xml#L20, which is inted to install `ros-indigo-pocketsphinx`, but since we add `pocketsphinx` entry to base.yaml, we can not install `pocketsphinx` with `rosdep install` any more

can we rename this  to `libsphinxbase` ?

tfoote:
@k-okada please submit a PR to do that. They shouldn't have a conflicting name if they have different content.

PR #14141, opened 2017-03-10
Update eigen-stl-containers to 0.1.8

clalancette:
This is a "by-hand" pull request because I answered "no" by accident on the last part of "bloom-release".


tfoote:
@clalancette you can rerun bloom-release with the `-p` option to only open the PR next time. And it will do just this. 

We're holding this for a coordinated indigo release.

clalancette:
Ah, thanks Tully, I'll use that next time.

PR #14146, opened 2017-03-10
urdfdom_py: 0.3.3-0 in 'lunar/distribution.yaml' [bloom]

clalancette:
Increasing version of package(s) in repository `urdfdom_py` to `0.3.3-0`:

- upstream repository: https://github.com/ros/urdf_parser_py/
- release repository: https://github.com/ros-gbp/urdfdom_py-release.git
- bloom version: `0.5.24`

## urdfdom_py

```
* Made Chris and Shane the maintainers
* Added python-lxml to the travis build.
* Reverted line break (ros/urdfdom#77 <https://github.com/ros/urdfdom/pull/77>) now that there is a more generic solution. (#5 <https://github.com/ros/urdf_parser_py/issues/5>)
* Added line break to make errors easier to read. (#4 <https://github.com/ros/urdf_parser_py/issues/4>)
* Contributors: Chris Lalancette, Isaac I.Y. Saito
```

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -675,5 +675,21 @@ repositories:
       url: https://github.com/ros/std_msgs.git
       version: groovy-devel
     status: maintained
+  urdfdom_py:
+    doc:
+      type: git
+      url: https://github.com/ros/urdf_parser_py.git
+      version: 0.3.2
+    release:
+      tags:
+        release: release/lunar/{package}/{version}
+      url: https://github.com/ros-gbp/urdfdom_py-release.git
+      version: 0.3.3-0
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/ros/urdf_parser_py.git
+      version: indigo-devel
+    status: maintained
 type: distribution
 version: 2
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -679,7 +679,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros/urdf_parser_py.git
-      version: 0.3.2
+      version: indigo-devel
     release:
       tags:
         release: release/lunar/{package}/{version}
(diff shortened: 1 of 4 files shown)


lunar/distribution.yaml
```diff
@@ -675,5 +675,21 @@ repositories:
       url: https://github.com/ros/std_msgs.git
       version: groovy-devel
     status: maintained
+  urdfdom_py:
+    doc:
+      type: git
+      url: https://github.com/ros/urdf_parser_py.git
+      version: 0.3.2
```
dirk-thomas: Why document a different version than what is released?

clalancette: I was just copying what was in kinetic here (mostly because I wasn't entirely sure what this doc section really meant).  Does bloom not update the doc version when doing a release?

dirk-thomas: No, bloom doesn't automatically update the doc entry when you do a release. Only the release entry. You could specify a branch instead if that contains the latest content.

tfoote: There is a prompt for the doc entry when you run bloom. When it's generating the pull-request. Note that it will seed it with the previous value. Usually it's a development branch so it only needs updating when the development forks.

Note that the upstream uri is malformed. it should end in .git not a slash: https://github.com/ros/urdf_parser_py/  https://github.com/ros-gbp/urdfdom_py-release/commit/b4a45a156e933005685df783c8f6cb2bddb3bf5e#diff-b9cba95a33d64aec1406327c4d43fa8dR115 though it looks like that should be fixed for all the versions. Though there's enough manually done in this package that it appears not to be a problem.

clalancette: Got it about the malformed upstream URI; I'll fix that one over in the release repository.

As far as the documentation, there has been no change to the documentation, and no forks of the development, since Indigo.  So I think I'll change the version on all of them to "indigo-devel", which is the main development branch.  Patch coming.

PR #14158, opened 2017-03-12
Re-releasing turtlebot_arm for kinetic with a new demo

corot:


corot:
:confused:  mmm...  I don't understand the failed check. I just want to add the package to the index for trying a kinetic prerelease.

dirk-thomas:
You are trying to add a release entry for Kinetic. But you release repo only contains content for Indigo. I assume you haven't run `bloom` for a Kinetic release?

corot:
Right; sorry!  I thought Prerelease came first. But no idea how to run a prerelase without adding the repo to the index

dirk-thomas:
You can add the repo with a `source` entry first. That should allow you to run a prerelease. (Not 100% sure, haven't done it for a while).

corot:
Thanks! Anyway, I'll hold back the release a bit, as i have noticed that one package I depend on is not yet released on Kinetic.
Question: with catkin_build_isolated I have many errors when including moveit headers (/opt/ros/kinetic/include/moveit/macros/declare_ptr.h) that use std::shared_ptr; do we compile on Kinetic with C++11 by default? Or do I need to force C++11 use in every package importing moveit headers?

dirk-thomas:
You need to explicitly set it in your package. I am not sure if the moveit package exports the information somehow.

corot:
Ok; just src and doc by now. With this it should appear on the wiki with a kinetic tab, right?
Thanks again!

dirk-thomas:
Yes, every will also be compiled which might fail if one of your dependencies hasn't been released yet.

PR #14161, opened 2017-03-13
Add junit-xml rules (both distribution-wise and PIP)

jvgomez:
This package helps creating test output (in platforms where rostest cannot be easily used).

- [junit-xml](https://pypi.python.org/pypi/junit-xml)

Two different rules have been done because of different versioning. PIP is currently on 1.7, but [Ubuntu versions vary significantly](http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python-junitxml&searchon=names). I am not sure if this is the correct procedure in this case.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1023,10 +1023,6 @@ python-jsonschema-pip:
       packages: [jsonschema]
 python-junitxml:
   ubuntu: [python-junitxml]
-python-junitxml-pip:
-  ubuntu:
-    pip:
-      packages: [junit-xml]
 python-kdtree:
   fedora: [libkdtree++-python]
   ubuntu:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1021,6 +1021,12 @@ python-jsonschema-pip:
   ubuntu:
     pip:
       packages: [jsonschema]
+python-junitxml:
+  ubuntu: [python-junitxml]
+python-junitxml-pip:
+  ubuntu:
+    pip:
+      packages: [junit-xml]
 python-kdtree:
   fedora: [libkdtree++-python]
   ubuntu:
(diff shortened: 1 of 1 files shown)


dirk-thomas (approved):

tfoote:
Do you need the newer pip version specifically for your use case? In general we'd rather stick with the distribution version unless there's a specific reason to jump forward. Code that relies on the pip version is not releasable, and can end up conflicting with the code that depends on the distribution version.

jvgomez:
@tfoote I just found a way around in my application that does not require these rules anymore (I was in fact using the PIP one). I am modifying the PR to include only the distribution version (so the time is not completely lost :) )

tfoote:
That's great. I've also added Debian and Fedora rules to make it generally releaseable.

Note that as of Stretch it's already at 1.7 

Fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/python-junit_xml/
stretch: https://packages.debian.org/stretch/python-junit.xml
jessie: https://packages.debian.org/jessie/python-junitxml


Actually now that I look deeper it seems that there's actually 2 different packages.

As of xenial there's: http://packages.ubuntu.com/search?keywords=python-junit.xml  (v 1.7)
compared to: http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python-junitxml&searchon=names (v 0.6) Which bottom out to completely different projects. Not just different versions. 

https://github.com/kyrus/python-junit-xml  is in the series 1.7 vs https://launchpad.net/pyjunitxml which is the 0.6 series.

jvgomez:
That's as interesting as confusing, thanks for the info!

tfoote:
Since python-junitxml seems not to be available on debian stretch and it's latest commit is from 2011 I'd suggest that the junit.xml package is a better target. It's also what's in pypi. 

It's not as available on older platforms but since you hadn't noticed that already it shouldn't be a problem to define the rules for xenial and newer, stretch, and fedora. So that's what I'd propose to do.

PR #14169, opened 2017-03-14
add python-slackclient to rosdep

benmaidel:


tfoote (changes requested):
Please use the `-pip` extension convention since it's exclusively available via pip.

ipa-bnm:
done

tfoote (approved):
That looks good. Note that you will not be able to release packages against pip targets, but it will work from source.

PR #14163, opened 2017-03-13
add OpenCV3 to lunar

vrabaud:

PR #14170, opened 2017-03-14
split libpng and update rules for latest distros

mikaelarguedas:
follow up on https://github.com/ros/rosdistro/pull/14163#issuecomment-286331857
@ros/ros_team I kept the key name as is (`libpng12-dev`) even if it's not accurate anymore. Would it be better to create a new `libpng-dev` key with all these rules?

@vrabaud FYI

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2196,7 +2196,7 @@ libpng-dev:
     jessie: [libpng12-dev]
     stretch: [libpng-dev]
     wheezy: [libpng12-dev]
-  fedora: [libpng12-devel]
+  fedora: [libpng-devel]
   gentoo:
     portage:
       packages: [media-libs/libpng]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -2192,15 +2192,30 @@ libpng++-dev:
   ubuntu: [libpng++-dev]
 libpng12-dev:
   arch: [libpng]
-  debian: [libpng12-dev]
+  debian:
+    jessie: [libpng12-dev]
+    sid: [libpng12-dev]
```
tfoote: We generally don't add sid since it's a rolling target

tfoote (changes requested):
Since it's transitioning to a new identifier by default, I think it makes sense to create a new key with the new name and let the old one age out.  Otherwise in the future we'll still be using libpng12-dev when were're on a much higher version and it won't make much sense.

mikaelarguedas:
Sounds good, updated to add a new key and leave `libpng12-dev` untouched

tfoote (approved):

tfoote:
I tweaked the fedora rule to use the fedora generic version. https://apps.fedoraproject.org/packages/libpng-devel otherwise lgtm

mikaelarguedas:
thanks @tfoote, I forgot to update fedora

PR #14175, opened 2017-03-15
bump OpenCV3 after bloom fix

vrabaud:


tfoote:
@vrabaud It's still failing on lunar due to not using the new libpng-dev keys: https://github.com/ros/rosdistro/pull/14170

vrabaud:
I saw, can only fix that in 12 hours. In the meantime, Jade and Kinetic and ARM will rebuild, that buys me some time :)

tfoote:
Sounds good. jade and kinetic will keep the farm busy until then.

PR #14176, opened 2017-03-15
adding daemontools for fedora ?

asmodehn:
I am not sure about this... I found the package there : https://pkgs.org/download/daemontools but not there : https://admin.fedoraproject.org/pkgdb/packages/daemontools%2A/
I cannot test on any fedora system.
Anyway I need this https://cr.yp.to/daemontools.html as a dependency on one of my package and it is not declared for fedora.
Please advise...

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -512,6 +512,7 @@ cython:
 daemontools:
   arch: [daemontools]
   debian: [daemontools]
+  fedora: [daemontools]
   gentoo:
     portage:
       packages: [sys-process/daemontools]


tfoote:
For us to add it to the rules set it needs to be in the default upstream repositories. pkgs.org seems to be referencing it from some person's personal repository: http://ftp.gwdg.de/pub/opensuse/repositories/home:/zhonghuaren/Fedora_25/src/ To add it here you'd want to request official Fedora packaging.

asmodehn:
Looks like it is not going to make it in fedora... https://bugzilla.redhat.com/show_bug.cgi?id=480727
I guess I ll have to come up with an alternative (I am only using setlock)...
Lets drop this one.

asmodehn:
Out of curiosity, would packaging daemontools for ROS be an option ?
In that case, how do we name it so that rosdep is not confused about which package is meant as a dependency ?

tfoote:
It's certainly possible, but in general if we package something that is available on the system it should make sure not to collide. Which often includes renamespacing etc. 

Looking at the ticket you found it looks like that was a very inexperienced packager which was not following the standard packaging procedures. And it was closed in 2009 when the package was also pulled from Ubuntu Hardy. But clearly it's been added back to the Debian/Ubuntu world so it likely could be rerequested to be added back to the Fedora world if done up to standards.

However it sounds like you don't need much functionality that's specific to that package and it will likely be lower effort to pick a different locking mechanism.

asmodehn:
Recently found out about `flock`, which is part of util-linux package. Seems to do what I need, and I think installed by default on most distros... So I switched to use that instead. Thanks !

PR #14180, opened 2017-03-15
remove PySide from Zesty for now since upstream doesn't build on it atm

dirk-thomas:
This should let `python_qt_binding` pass on Zesty (with PyQt support only).


dirk-thomas:
Once this is merged it needs a debinc release of `python_qt_binding` and `qt_gui_core` for Lunar to use the updated mapping.

tfoote (approved):
That's unfortunate. Hopefully that will be fixed and we can add it back soon.

PR #14184, opened 2017-03-15
Adding auction_methods_stack to documentation index for groovy

joaoquintas:
I'd like auction_methods_stack to be indexed and documented on ros.org.


dirk-thomas:
ROS Groovy has been EOLed many years ago. We don't operate a buildfarm for it anymore and therefore can't accept any contributions to it anymore. Please consider adding your repository to one of the active ROS distributions instead: http://wiki.ros.org/Distributions

joaoquintas:
Thanks for the quick reply. Ok, I'll do this change.

PR #14186, opened 2017-03-15
Adding auction_methods_stack to documentation index for kinetic

joaoquintas:
I'd like auction_methods_stack to be indexed and documented on ros.org.

### DIFF
--- a/groovy/distribution.yaml
+++ b/groovy/distribution.yaml
@@ -153,6 +153,11 @@ repositories:
       type: git
       url: https://github.com/ethz-asl/asctec_mav_framework.git
       version: master
+  auction_methods_stack:
+    doc:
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
@joaoquintas I think that travis is failing because you are missing a `.git` at the end of your repository url

joaoquintas:
@mikaelarguedas , thank you so much for the help. It seem it passed

joaoquintas:
@mikaelarguedas , thank you so much. It was the missing .git that was indeed causing the problem

mikaelarguedas:
looks good, thanks.
Note: the documentation is running every 3 days so you may have to wait a bit before this shows up on the ros wiki

PR #14206, opened 2017-03-17
Add python-requests-pip

megawac:
The version in the ubuntu distro for instance is approximately 3 years old.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2137,7 +2137,6 @@ python-requests-pip:
   ubuntu:
     pip:
       packages: [requests]
-packages: [autobahn]
 python-responses-pip:
   osx:
     pip:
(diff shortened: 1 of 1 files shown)


tfoote:
In general if there is a system version we strongly recommend making it work. Otherwise if one package requires the new one and one package requires the old one they become mutually exclusive. That's why the distro has picked a specific version, so that everyone will be consistent. 

Is the version available via PIP 100% backwards compatible (extensions only)with the version in apt that it won't break any code? Otherwise I'd strongly recommend against adding this since it will side effect package using the system installation since the pip version will override the apt version.

megawac:
Ah that makes sense, I wasn't considering that installing via pip would override the system installation. Gotcha on preferring the distro versions when available, feel free to close this PR if you see fit.

As far as I can tell there have only been bug fixes and supplementary feature support added to requests since 2.2.1 (rosdistros latest version on ubuntu) https://pypi.python.org/pypi/requests#release-history.

tfoote:
Yeah, sometimes it's frustrating how far behind things get sometimes, but stability has benefits too. Thanks for understanding.

PR #14208, opened 2017-03-17
Revert "Add junit-xml rules (both distribution-wise and PIP)"

tfoote:
Reverts ros/rosdistro#14161

As per the discussion in #14161 this is an inconsistent state bringing in two different libraries, junitxml and juint_xml are different. I'd like to revert this and make sure we have a a single library.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1068,12 +1068,6 @@ python-jsonschema-pip:
   ubuntu:
     pip:
       packages: [jsonschema]
-python-junitxml:
-  debian:
-    jessie: [python-junitxml]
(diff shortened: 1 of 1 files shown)


mikaelarguedas (approved):

mikaelarguedas:
I agree with https://github.com/ros/rosdistro/pull/14161#issuecomment-286280888 and create a key for junit.xml only on recent distros

tfoote:
@jvgomez Could you make an updated PR with the changes mentioned by @mikaelarguedas assuming that works for your use case. (Since you were using that before anyway that seems like it should work.)

PR #14223, opened 2017-03-20
[rosdep] add several keys for openSUSE

zhxt:


rosdep/base.yaml
```diff
@@ -2901,6 +2905,7 @@ libvtk-java:
   gentoo: [sci-libs/vtk]
   slackware: [VTK]
   ubuntu: [libvtk-java]
+  opensuse: [vtk-java]
```
dirk-thomas: Needs to be in alphabetical order.

zhxt: Thanks for reply. Yeah, I will fix that.  : )

zhxt: @dirk-thomas Fixed.
And one more question, do you know is there a way to add key for different openSUSE version?

dirk-thomas: @wjwwood might know the answer to your question. He is maintaining `rosdep`.

tfoote: OpenSuse does have support for versions like the other platforms: https://github.com/ros-infrastructure/rospkg/blob/master/src/rospkg/os_detect.py#L219

zhxt: @dirk-thomas , @tfoote many thanks, good to know.

rosdep/python.yaml
```diff
@@ -2051,6 +2058,7 @@ python-qt5-bindings-gl:
     stretch: [python-pyqt5.qtopengl]
   fedora: [python-qt5]
   gentoo: [dev-python/PyQt5]
+  opensuse: [python3-qt5]
```
tfoote: Python3 vs python2

rosdep/python.yaml
```diff
@@ -620,6 +622,7 @@ python-defusedxml:
   debian: [python-defusedxml]
   fedora: [python-defusedxml]
   gentoo: [dev-python/defusedxml]
+  opensuse: [python3-defusedxml]
```
tfoote: python3 vs 2

rosdep/python.yaml
```diff
@@ -1207,6 +1210,7 @@ python-mock:
   debian: [python-mock]
   fedora: [python-mock]
   gentoo: [dev-python/mock]
+  opensuse: [python3-mock, python-mock]
```
tfoote: 3 and 2 here

rosdep/python.yaml
```diff
@@ -1445,6 +1449,7 @@ python-opencv:
   arch: [opencv, python2-numpy]
   debian: [python-opencv]
   fedora: [opencv-python]
+  opensuse: [python3-opencv,python-opencv]
```
tfoote: 3 and 2

rosdep/python.yaml
```diff
@@ -1757,6 +1762,7 @@ python-pygraphviz:
   arch: [python2-pygraphviz]
   debian: [python-pygraphviz]
   fedora: [graphviz-python]
+  opensuse: [python3-pygraphviz, python-pygraphviz]
```
tfoote: 3 and 2

rosdep/python.yaml
```diff
@@ -509,6 +510,7 @@ python-coverage:
   debian: [python-coverage]
   fedora: [python-coverage]
   gentoo: [dev-python/coverage]
+  opensuse: [python3-coverage, python-coverage]
```
tfoote: 3 and 2

rosdep/base.yaml
```diff
@@ -3531,6 +3541,7 @@ python-pygraphviz:
   gentoo:
     portage:
       packages: [dev-python/pygraphviz]
+  opensuse: [python3-pygraphviz, python-pygraphviz]
```
tfoote: 3 and 2

tfoote (changes requested):
Overall this looks good. However there are several rules that duplicate the python2 and python3 entries. They are both not substitutes as well as generally shouldn't just have both installed if only one of the two is needed.

rosdep/python.yaml
zhxt: Thanks for review.  Compared with other distros, selected python2 as the default.

zhxt: uhmmm, there  is no defusedxml module for python2.

zhxt: Found one after add devel-language-python repo(http://download.opensuse.org/repositories/devel:/languages:/python/openSUSE_Leap_42.2/)

:D

rosdep/python.yaml
zhxt: all fixed with python2 now.

zhxt:
updated with python2 as default.

rosdep/python.yaml
tfoote: Is that available in a default installation? I don't seem to have instructions for OpenSUSE here to test: http://wiki.ros.org/ROS/Installation

zhxt: It seems not in the default installation, I don't see it in an official repo(http://software.opensuse.org/package/python-defusedxml).  And now I remember I have another repo(devel:language:python3) being added.
Yeah, I don't see instructions for openSUSE,too.  I start from the "Source Installation" that for Ubuntu and Debian, and thought that would possible for openSUSE since they are all Linux distorts :P

zhxt: s/distorts/distros/

dirk-thomas (dismissed):
This looks good to me.

@tfoote can you please update your review.

dirk-thomas (approved):
:+1: 

@tfoote can you please update your review.

tfoote (approved):

PR #14239, opened 2017-03-21
Don't test commits for basler_tof + sick_ldmrs_laser

mintar:
* sick_ldmrs_laser depends on libsick_ldmrs, which is only available as
  source package
* basler_tof depends on external binary library


dirk-thomas:
`test_pull_requests: false` is the default for the ROS distros. Please simply remove those lines to avoid redundant information.

mintar:
Done!

PR #13319, opened 2016-11-29
jaguar_robot: 0.1.0-1 in 'kinetic/distribution.yaml' [bloom]

gstavrinos:
Increasing version of package(s) in repository `jaguar_robot` to `0.1.0-1`:

- upstream repository: https://github.com/gstavrinos/jaguar_robot
- release repository: https://github.com/gstavrinos/jaguar_robot-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.5.23`
- previous version for package: `null`

## jaguar_base

```
* Functional Indoors Navigation
* Motor controller drivers are stable
* Contributors: George Stavrinos
```

## jaguar_bringup

```
* Added laser bringup launch file
* Added myahrs bringup launch file (external imu)
* Contributors: George Stavrinos
```

## jaguar_robot

```
* Metapackage for the physical robot (not simulated)
* Contributors: George Stavrinos
```

PR #14249, opened 2017-03-22
reverting #13319

tfoote:
Due to missing branches on release repository.

Reverts: #13319 @gstavrinos FYI


gstavrinos:
I think the problem is the myahrs_driver package which is missing from kinetic.

tfoote:
There should be a lot of branches on the release repo like this with 66: https://github.com/gstavrinos/jaguar-release

Where this release repo only has 8: https://github.com/gstavrinos/jaguar_robot-release

PR #14323, opened 2017-03-26
[rosdep] Add Adafruit BNO055 python library

HALtheWise:
Added python-adafruit-bno055 to rosdep as used by https://github.com/markusgrotz/ros_bno055_driver

Let me know if python-adafruit-bno055-pip is the better name here.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -112,6 +112,16 @@ python:
     yakkety_python3: [python3-dev]
     zesty: [python-dev]
     zesty_python3: [python3-dev]
+python-adafruit-bno055:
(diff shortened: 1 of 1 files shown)


mikaelarguedas (changes requested):
The package being only available via pip, can you  rename the key to `python-adafruit-bno055-pip` please?
Thanks!

Edit: just saw the edit on your comment. Yes indeed using the suffix "-pip" will be better in this case. Thanks

HALtheWise:
Done.

mikaelarguedas (approved):
Thanks

PR #14326, opened 2017-03-27
Package abb indigo release 1.2.1

Levi-Armstrong:


Levi-Armstrong:
@dirk-thomas: I noticed that two packages( abb_resources & abb_irb6640_support) are not shown on the  http://build.ros.org/ website. Any thoughts to why? These were new packages since the last release so not sure if that makes a difference.

dirk-thomas:
@Levi-Armstrong Please see #14360 for the reason. I hope it gets addressed soon.

PR #14332, opened 2017-03-27
random_numbers: 0.3.1-1 in 'lunar/distribution.yaml'

nuclearsandwich:
Releasing random_numbers for lunar.

upstream repository: https://github.com/ros-planning/random_numbers.git
release repository: https://github.com/ros-gbp/random_numbers-release.git
distro file: `lunar/distribution.yaml`
bloom version: `0.5.25`

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -558,6 +558,10 @@ repositories:
       version: kinetic-devel
     status: maintained
   random_numbers:
+    doc:
+      type: git
+      url: https://github.com/ros-planning/random_numbers.git
+      version: master
     release:
       tags:
         release: release/lunar/{package}/{version}
@@ -566,7 +570,7 @@ repositories:
     source:
       type: git
       url: https://github.com/ros-planning/random_numbers.git
-      version: 0.3.1
+      version: master
     status: maintained
   realtime_tools:
     doc:
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -557,6 +557,17 @@ repositories:
       url: https://github.com/ros-visualization/qt_gui_core.git
       version: kinetic-devel
     status: maintained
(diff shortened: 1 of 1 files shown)


lunar/distribution.yaml
```diff
@@ -557,6 +557,17 @@ repositories:
       url: https://github.com/ros-visualization/qt_gui_core.git
       version: kinetic-devel
     status: maintained
+  random_numbers:
```
dirk-thomas: This should also have a `doc` section (same content as the `source` section).

lunar/distribution.yaml
```diff
@@ -557,6 +557,17 @@ repositories:
...
+    release:
+      tags:
+        release: release/lunar/{package}/{version}
+      url: https://github.com/ros-gbp/random_numbers-release.git
+      version: 0.3.1-1
+    source:
+      type: git
+      url: https://github.com/ros-planning/random_numbers.git
+      version: 0.3.1
```
dirk-thomas: You should use the master branch here.

lunar/distribution.yaml
nuclearsandwich: Added in https://github.com/ros/rosdistro/pull/14332/commits/ba9e010c0c8974177cd1a1b3bd9237bedd0e00d2.

mikaelarguedas:
Note: you may want to do the release including the maintainer change commit this way you will be the one receiving emails if the builds fail on the buildfarm and not the previous maintainer

nuclearsandwich:
> Note: you may want to do the release including the maintainer change commit this way you will be the one receiving emails if the builds fail on the buildfarm and not the previous maintainer

whoops, yes. This is the second side effect of targeting the 0.3.1 tag rather than master. 

If I make a new bloom release with the correct maintainer info since this has already merged.

PR #14355, opened 2017-03-28
Update python.yaml with dependecy for gTTS

giusebar:


mikaelarguedas:
the entries need to be listed alphabetically in the file, I think that is the reason travis is failing.
Could you also add a debian rule ? thanks!

dirk-thomas (approved):

PR #14359, opened 2017-03-28
added python-expiringdict & python-xlib for ubuntu/xenial

easymov:


mikaelarguedas:
@easymov can you add python-expiringdict to all ubuntu distributions where this package is available ?
Is it possible to add debian and fedora entries as well ? Thanks!

easymov:
No problem, I will check availability on other distributions.

mikaelarguedas:
if available from apt please use the apt rather than the pip version:
- `python-xlib`
  - ubuntu: available on all not EOL ubuntu distributions: http://packages.ubuntu.com/search?keywords=python-xlib&searchon=names&suite=all&section=all
  - debian: is available on all not EOL debian distribution: https://packages.debian.org/search?keywords=python-xlib&searchon=names&suite=all&section=all
  - fedora: available on all not EOL distros: https://admin.fedoraproject.org/pkgdb/package/rpms/python-xlib/
 
- `python-expiringdict`
  -  ubuntu: is available on ubuntu since vivid: http://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python-expiringdict&searchon=names
  - debian: `python-expiringdict` is available on debian since jessie https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=python-expiringdict
  - not available on fedora: you can use pip for now or just pass on the fedora entry

Thanks!

easymov:
Thanks for the list.
It's updated now.

mikaelarguedas:
Thanks for iterating on this :+1:

mikaelarguedas (approved):
:+1:

rosdep/python.yaml
```diff
@@ -689,6 +689,18 @@ python-evdev:
     vivid: [python-evdev]
     wily: [python-evdev]
     xenial: [python-evdev]
+python-expiringdict:
+  debian:
+    jessie: [python-expiringdict]
+    sid: [python-expiringdict]
```
tfoote: Please don't add rules for sid explicitly, we don't target it because it's a rolling target. 

And since these are all the same it doesn't need to have each version called out individually.

rosdep/python.yaml
```diff
@@ -2827,6 +2839,14 @@ python-wstool:
 python-wtforms:
   fedora: [python-wtforms]
   ubuntu: [python-wtforms]
+python-xlib:
+  debian:
+    jessie: [python-xlib]
+    sid: [python-xlib]
```
tfoote: Ditto here for sid and repeated.

rosdep/python.yaml
```diff
@@ -689,6 +689,18 @@ python-evdev:
...
+  debian:
+    jessie: [python-expiringdict]
+    sid: [python-expiringdict]
+    stretch: [python-expiringdict]
+  fedora:
+    pip:
+      packages: [expiringdict]
+  ubuntu:
+    xenial: [python-expiringdict]
+    yakkety: [python-expiringdict]
```
tfoote: Also here calling out the version is not necessary since it's the same for all versions.

tfoote (changes requested):
I had a few comments to simplify the rules down. Otherwise looks good.

rosdep/python.yaml
easymov: Ok, is [this place](http://www.ros.org/reps/rep-0003.html) the reference to find targeted versions or is there something else to quickly find which rosdep keys should be added ?
Thanks !

rosdep/python.yaml
easymov: I though it was a way to explicitly mention that this package was available for this versions only, isn't it ?

mikaelarguedas: It's my bad I thought we needed to call out versions when it's not available for some of the currently supported platforms (e.g. trusty for this package). can you confirm @tfoote that it's fine to collapse the rules anyway?

mikaelarguedas:
@tfoote can you re-review this please ?

tfoote (approved):
Thanks for iterating. LGTM

PR #14370, opened 2017-03-29
adding python-hypothesis

asmodehn:
I m not sure about distro details for debian and fedora. please check...

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1002,6 +1002,21 @@ python-httplib2:
     saucy: [python-httplib2]
     trusty: [python-httplib2]
     trusty_python3: [python3-httplib2]
+python-hypothesis:
+  debian: 
+    stretch: [python-hypothesis]
+    sid: [python-hypothesis]
+  fedora: 
+    '24': [python-hypothesis]
+    '25': [python-hypothesis]
+    '26': [python-hypothesis]
+  ubuntu:
+    xenial: [python-hypothesis]
+    xenial_python3: [python3-hypothesis]
+    yakkety: [python-hypothesis]
+    yakkety_python3: [python3-hypothesis]
+    zesty: [python-hypothesis]
+    zesty_python3: [python3-hypothesis]
 python-imaging:
   arch: [python2-pillow]
   debian: [python-imaging]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1003,10 +1003,10 @@ python-httplib2:
     trusty: [python-httplib2]
     trusty_python3: [python3-httplib2]
 python-hypothesis:
-  debian: 
-    stretch: [python-hypothesis]
+  debian:
     sid: [python-hypothesis]
-  fedora: 
+    stretch: [python-hypothesis]
+  fedora:
     '24': [python-hypothesis]
     '25': [python-hypothesis]
     '26': [python-hypothesis]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1002,6 +1002,21 @@ python-httplib2:
     saucy: [python-httplib2]
     trusty: [python-httplib2]
     trusty_python3: [python3-httplib2]
+python-hypothesis:
+  debian:
+    sid: [python-hypothesis]
```
dirk-thomas: Please remove `sid`. Since it is a rolling release we don't specify it explicitly. Since all keys on Debian are the same you could also make a generic entry (rather than making it distro specific): `debian: [python-hypothesis]`

rosdep/python.yaml
```diff
@@ -1002,6 +1002,21 @@ python-httplib2:
...
+    stretch: [python-hypothesis]
+  fedora:
+    '24': [python-hypothesis]
+    '25': [python-hypothesis]
+    '26': [python-hypothesis]
```
dirk-thomas: Since all keys on Fedora are the same you could also make a generic entry (rather than making it distro specific): fedora: [python-hypothesis]

rosdep/python.yaml
mikaelarguedas: we usually don't target sid because it's a rolling distro

rosdep/python.yaml
mikaelarguedas: look like this is also available on 23 https://admin.fedoraproject.org/pkgdb/package/rpms/python-hypothesis/

mikaelarguedas (commented):
Are you targeting only lunar with this key and plan on releasing the python package yourself for kinetic (as you did for jade?)

asmodehn:
This is for kinetic and later. I am planning to backport the xenial version as thirdparty release for indigo and jade.

mikaelarguedas (dismissed):

dirk-thomas:
@asmodehn If you target `Kinetic` you will need an entry for `wily`.

mikaelarguedas:
oh good point, thanks @dirk-thomas for double checking

asmodehn:
Well,
- Wily is already EOL. https://wiki.ubuntu.com/Releases I would assume anyone using it should upgrade already.
- Wily does not have the package available.

I would like to use the ubuntu package as early in release timeline as possible (avoid maintenance).
The only possible way I am aware of, to shift from a third party ROS release to a linux distro release, is via a patch on package.xml in the release repo. But these depends on ROS releases.

=> What can I do ? If possible I'd like to avoid having to have a to use a ROS package as dependency when a system package is available. Also AFAIK there is no way to specify in the package.xml a different rosdep key per linux distro...

I personally would vote for "Dont try to support ubuntu distros that canonical doesn't even support" but let me know what you think is best here. Thanks.

dirk-thomas:
Please see https://discourse.ros.org/t/drop-support-for-wily/727 for more information on this topic. For now you either can't release your package into Kinetic (assuming you don't want to create a third-party package for your dependency) or live with the failing wily jobs until building for EOL distros has been disabled on the buildfarm. That being said we are pretty close to be able to disable those jobs. Hopefully after the next sync (within a couple of weeks) we can turn of jobs for EOL Ubuntu distros.

dirk-thomas (approved):

asmodehn:
Thanks for the information.

I will wait a bit before releasing on kinetic and then I will live with the
failing wily jobs, I already have failing saucy jobs for another package
anyway.

Looking forward having a leaner, faster buildfarm :-)

mikaelarguedas:
@asmodehn Jobs are now disabled for all EOL distros (https://discourse.ros.org/t/suspension-of-debian-packaging-for-eold-distros-saucy-utopic-vivid-and-wily/1601) so you should be able to release on kinetic.

Issue #14360, opened 2017-03-28
kinetic job reconfiguration is failing


hrnr:
Hi,

kinetic jobs reconfiguration task `Krel_reconfigure-jobs` is failing since Mar 23 due to circular dependecies between `roch_bringup, roch_capabilities`. This was introduced by #14299. This prevents releasing new packages for kinetic.

Is anybody aware of it?

Cheers,
Jiri

dirk-thomas:
@tfoote is in charge of Kinetic and has received notification email for all of the failing jobs. Please also improve the unit tests in this repo to catch the problem before merging in the future.

dirk-thomas:
The same is obviously the case for Indigo.

tfoote:
Also at issue:
Indigo: https://github.com/ros/rosdistro/pull/14268
Debinc: https://github.com/ros/rosdistro/pull/14306

tfoote:
Kinetic reconfiguring is running, and Indigo is waiting for kinetic to finish to run. I think it should work after kinetic finishes. I'll check in the morning.

tfoote:
Indigo and Kinetic are both reconfiguring successfully now.

PR #14376, opened 2017-03-29
Reverting roch_viz to remove circular dependency

tfoote:
As flagged in #14360 

@SawYer-Robotics FYI

Reverts: https://github.com/ros/rosdistro/pull/14273


tfoote:
retriggering travis

tfoote:
nevermind travis is offline and not accepting any new jobs while they resolve a db issue: https://www.traviscistatus.com/incidents/93wfcv0ksqrz

PR #14392, opened 2017-03-31
add bullet for yakkety and zesty

tfoote:
@mikaelarguedas


mikaelarguedas (approved):
Added debian along the way

tfoote:
Thanks @mikaelarguedas that will be useful shortly.

PR #14397, opened 2017-03-31
Add python-crccheck-pip

anassinator:
https://pypi.python.org/pypi/crccheck


tfoote (approved):
Looks fine. You could consider rules for debian and fedora as well.

anassinator:
Done!

PR #14408, opened 2017-03-31
Add python-pyaudio-pip

thomasbi1:
https://pypi.python.org/pypi/PyAudio

This installs newer versions of PyAudio than with apt.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1740,6 +1740,16 @@ python-pyaudio:
   debian: [python-pyaudio]
   fedora: [pyaudio]
   ubuntu: [python-pyaudio]
+python-pyaudio-pip:
+  debian:
+    pip:
+      packages: [pyaudio]
+  osx:
+    pip:
+      packages: [pyaudio]
+  ubuntu:
+    pip:
+      packages: [pyaudio]
 python-pydot:
   arch: [python2-pydot]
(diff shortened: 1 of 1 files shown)


dirk-thomas (approved):

tfoote:
Since there's already a version packaged upstream we generally prefer not to overlay pip versions as they will effect and override any other version installed on the system. This can relatively easily break any other package that expects the system version that they were packaged against, and can end up getting out of date since the pip version is not tracked by apt and will not auto update. 

Can you explain your reasoning for needing a new package. I only see bugfixes in the recent changelog. https://people.csail.mit.edu/hubert/pyaudio/

thomasbi1:
Apt installs pyAudio v. 0.2.8 which has bugs relating "overflow error handling" which are fixed starting from v. 0.2.9.

A ROS package I am currently developing is dependant on these bugs being fixed, or else it's core cannot run properly.

thomasbi1:
@tfoote Could this potentially be merged, or do you prefer not to, due to your mentioned reasons?

tfoote:
I'd strongly prefer not to diverge from the upstream packaged version. If it's just patch fixes those should be acceptable for release into the debian packages.

dirk-thomas:
I will close this based on the previous feedback.

PR #14409, opened 2017-03-31
Add portaudio2

thomasbi1:
https://packages.debian.org/jessie/libportaudio2
http://packages.ubuntu.com/yakkety/libportaudio2

Provides PortAudio v19

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3426,6 +3426,10 @@ portaudio:
   debian: [libportaudio-dev]
   fedora: [portaudio-devel]
   ubuntu: [libportaudio-dev]
+portaudio19-dev:
+  debian: [libportaudio19-dev]
+  fedora: [libportaudio-devel]
+  ubuntu: [libportaudio19-dev]
 postgresql:
   fedora: [postgresql-server, postgresql-contrib]
   ubuntu: [postgresql, postgresql-contrib]


dirk-thomas (approved):

tfoote:
Looking to fill this in for fedora I noticed that libportaudio2 has the same contents as libportaudio on fedora:

http://packages.ubuntu.com/yakkety/amd64/libportaudio2/filelist
https://apps.fedoraproject.org/packages/portaudio/contents/

So the fedora one should be added. But then trying to compare to libportaudio-dev I found that it's actually been dropped the last version listed is trusty: http://packages.ubuntu.com/trusty/libportaudio-dev so it's not in xenial or newer, It may be in EOL'd distros between. (Just not indexed on the website) 

And digging deeper there's the new libportaudio19-dev http://packages.ubuntu.com/precise/portaudio19-dev which libportaudio2 is the library package http://packages.ubuntu.com/source/precise/portaudio19

Related to that do you really only need the library or do you need the dev package too? 

Overall I'd suggest that we update this to have a rule 
`libportaudio19-dev` which points to the debian and ubuntu version which is generally available (and will implicitly pull in libportaudio2 as a dependency for the library) And on Fedora it would be just libportaudio-devel which has the 19/2 version inside it already.

thomasbi1:
According to your suggestions, I changed my pull request. There is already a key name `portaudio`,  so I named the new key according to the existing one: `portaudio19`. Also the since the fedora lib is already contained in the `portaudio` key, I do not have to add it to `portaudio19`, correct?

rosdep/base.yaml
```diff
@@ -3426,6 +3426,9 @@ portaudio:
   debian: [libportaudio-dev]
   fedora: [portaudio-devel]
   ubuntu: [libportaudio-dev]
+portaudio19:
```
tfoote: This should have the `-dev` extension. Our policy is to by default copy the debian name.

rosdep/base.yaml
```diff
@@ -3426,6 +3426,9 @@ portaudio:
...
+  debian: [libportaudio19-dev]
+  ubuntu: [libportaudio19-dev]
```
tfoote: Please add `fedora: [libportaudio-devel]` which has the same content.

tfoote (changes requested):
Thanks for the update. I had two small comments.

thomasbi1:
Should I change the `portaudio` key to `portaudio-dev` as well, then?

tfoote (approved):

tfoote:
Thanks for the iteration. Since that key is referring to EOL'd targets we should just let it be until it's aged out. Otherwise we'd break anyone currently using it.

PR #14410, opened 2017-03-31
Add cython-pip

thomasbi1:
https://pypi.python.org/pypi/Cython/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1,3 +1,10 @@
+cython-pip:
+  debian:
+    pip:
+      packages: [cython]
+  ubuntu:
+    pip:
+      packages: [cython]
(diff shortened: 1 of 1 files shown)


dirk-thomas (approved):

tfoote:
Cython is packaged upstream. We strongly prefer natively packaged versions. Can you explain why you need the pip version instead of the apt/rpm/brew version: http://packages.ubuntu.com/search?keywords=cython

thomasbi1:
I checked for a Cython key in `python.yaml` and not `base.yaml`. Please excuse this pull request, the apt version will do fine :)

tfoote:
No worries. Discoverability is not that great for these things.

PR #14459, opened 2017-04-05
address removal of old formula from homebrew-core

wjwwood:
Most were migrated to ros/homebrew-deps, see:

https://github.com/ros/homebrew-deps/pull/31


wjwwood:
Do not consider merging before ros/homebrew-deps#31.

wjwwood:
See: https://github.com/ros/rosdistro/pull/14573

PR #14462, opened 2017-04-06
added debian and fedora names of packages needed for homer_robot_face

niyawe:


dirk-thomas (approved):

rosdep/base.yaml
```diff
@@ -1400,13 +1401,15 @@ libdrm-dev:
...
 libesd0-dev:
+  debian: [libesd0-dev]
   fedora: [esound-devel]
   ubuntu: [libesd0-dev]
 libespeak-dev:
   debian: [libespeak-dev]
   fedora: [espeak-devel]
   ubuntu: [libespeak-dev]
 libestools-dev:
+  debian: [libestools2.1-dev]
```
mikaelarguedas: rosdep doesn't support pure virtual packages, can you use the parent package  `libestools-dev` (https://packages.debian.org/stretch/libestools-dev) instead ?

rosdep/base.yaml
```diff
@@ -2481,6 +2485,8 @@ libqt5-widgets:
   slackware: [qt5]
   ubuntu: [libqt5widgets5]
 libqt5x11extras5-dev:
+  debian: [libqt5x11extras5-dev]
+  fedora: [libQt5X11Extras-devel]
```
mikaelarguedas: I think that we use only RPM available on the Fedora package database. Maybe @wjwwood or @rmattes can confirn this

rosdep/base.yaml
niyawe: in debian jessie libestools-dev is a virtual package

mikaelarguedas: In this case we will need to make a subcategory for jessie and for stretch referring to the right non-virtual packages

rosdep/base.yaml
niyawe: It seems like i used the wrong package name. Now I found qt5-qtx11extras-devel (https://apps.fedoraproject.org/packages/qt5-qtx11extras-devel).

mikaelarguedas:
fixed in https://github.com/ros/rosdistro/pull/14464

PR #14486, opened 2017-04-07
catkin_simple: 0.1.2-1 in 'indigo/distribution.yaml' [bloom]

fmina:
Increasing version of package(s) in repository `catkin_simple` to `0.1.2-1`:

- upstream repository: https://github.com/fmina/catkin_simple.git
- release repository: https://github.com/fmina/catkin_simple-release.git
- bloom version: `0.5.23`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1274,6 +1274,21 @@ repositories:
       url: https://github.com/asmodehn/catkin_pip.git
       version: indigo
     status: developed
+  catkin_simple:
+    doc:
+      type: git
+      url: https://github.com/fmina/catkin_simple.git
+      version: master
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/fmina/catkin_simple-release.git
+      version: 0.1.2-1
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@ros/ros_team I am not sure if this should be release. Any opinions?

wjwwood:
I don't think it should be released from a fork. As for releasing it in general, we had reservations about supporting packages which use it on the build farm. I've also laid out in the past what needs to be done before it should be release in the past, and in that post I mentioned that is should get a new name if someone forks it:

> If you're interested, you could take over and release it. Or you could fork it as is and release it under a different name.

-- https://github.com/catkin/catkin_simple/issues/29#issuecomment-193911887

So, I think that if @fmina wants to release it and support it, then the name should be adjusted to `fmina_catkin_simple` (in the `package.xml` not just the repository name here) or something similar.

dirk-thomas:
I agree with @wjwwood on this. Therefore I will close the PR. Please feel free to contribute to improve `catkin_simple` or consider forking and releasing it under a different name (e.g. with a prefix).

PR #14517, opened 2017-04-12
comment out EOL ubuntu distros from distribution files

mikaelarguedas:
This is a follow-up of https://github.com/ros-infrastructure/bloom/issues/422

We will let this soak for a while and wait for feedback on the ros-buildfarm mailing list before merging this

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -7,6 +7,7 @@ release_platforms:
   - '21'
   - heisenbug
   ubuntu:
+  # EOLed
   # - saucy
   - trusty
 repositories:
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Please add a comment line above each commented block to indicate why the line is commented out. Similar to https://github.com/ros-infrastructure/ros_buildfarm_config/pull/80/commits/273bc51d393feca9f632ce3a92be40069480aa1e

dirk-thomas (approved):

mikaelarguedas:
Please comment on the [discourse thread](https://discourse.ros.org/t/stop-blooming-for-eol-distros-and-comment-them-out-from-distribution-files/1652) if you have any feedback about this change

dirk-thomas:
The unit tests don't support comment at the moment. They either need to be removed or the tests need to be updated.

dirk-thomas (approved):

PR #14564, opened 2017-04-18
Add gazebo8 to rosdep

j-rivero:
//cc @mikaelarguedas

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -774,6 +774,7 @@ gazebo:
     trusty: [gazebo2]
     wily: [gazebo7]
     xenial: [gazebo7]
+    yakkety: [gazebo8]
     zesty: [gazebo8]
 gazebo5:
   arch: [gazebo]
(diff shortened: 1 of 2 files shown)


rosdep/base.yaml
```diff
@@ -774,11 +774,19 @@ gazebo:
     trusty: [gazebo2]
     wily: [gazebo7]
     xenial: [gazebo7]
+    zesty: [gazebo8]
```
mikaelarguedas: you'll need a yakkety entry here for completeness

mikaelarguedas (approved):
lgtm thanks @j-rivero

tfoote:
Are these already imported and available in the ROS debian repo? I haven't seen tickets for that.

mikaelarguedas:
Good point I forgot to mention it here, thay have not been pulled in yet. thanks for pointing it out.

mikaelarguedas:
requires https://github.com/ros-infrastructure/reprepro-updater/pull/40 before being merged

tfoote (approved):
I imported into the bootstrap repo, and then into the buildfarm: http://build.ros.org:8080/job/import_upstream/116/console

On Fedora it's gazebo8 by default on F26. So I guess aliasing like this is best.

PR #14566, opened 2017-04-18
[dep][ubuntu/debian] Add geographiclib-tools/libgeographiclib-dev

TSC21:
```bash
$rosdep resolve geographiclib-tools
#apt
geographiclib-tools

$rosdep resolve geographiclib-tools --os=ubuntu:xenial
#apt
geographiclib-tools
```
Main purpose is allowing integration of datasets of GeographicLib to be used by [mavros](http://wiki.ros.org/mavros).

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -826,6 +826,7 @@ geographiclib:
     xenial: [libgeographic-dev]
 geographiclib-tools:
   debian: [geographiclib-tools]
+  fedora: [GeographicLib]
   ubuntu: [geographiclib-tools]
 geos:
   arch: [geos]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -812,6 +812,7 @@ gdal-bin:
   gentoo: [sci-libs/gdal]
   ubuntu: [gdal-bin]
 geographiclib:
+  debian: [libgeographiclib-dev]
   fedora: [GeographicLib-devel]
   ubuntu:
     precise: [libgeographiclib-dev]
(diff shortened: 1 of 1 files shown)


wjwwood (approved):

rosdep/base.yaml
```diff
@@ -823,6 +823,14 @@ geographiclib:
     vivid: [libgeographic-dev]
     wily: [libgeographic-dev]
     xenial: [libgeographic-dev]
+geographiclib-tools:
+  debian: [geographiclib-tools]
+  ubuntu:
+    precise: [geographiclib-tools]
```
tfoote: These look to be the same. Is there any reason not to just use `ubuntu: [geographiclib-tools]`?

TSC21: Not really. Just wanna to complete OS dependencies as much as possible. Should I take it then?

tfoote: As long as the package is generally available the more compact version is better, since it doesn't need updating for every future ubuntu release. And it will work for the intermediate ones that you didn't call out either.

TSC21: updating then. Thanks

TSC21:
@tfoote done. Thank you!

tfoote:
@TSC21 Sorry I wasn't clear about simplifying.

 i was suggesting getting rid of the specific versions inside Ubuntu not dropping debian. Having support for debian is valuable as well as fedora are our standards.  
I see https://apps.fedoraproject.org/packages/GeographicLib for Fedora. Which would be great to have too.

TSC21:
@tfoote makes sense now?

TSC21:
Btw, `fedora: [GeographicLib-devel]` already for `geographiclib` dep, so it was already included (already as the tools to download the datasets).

TSC21 (commented):
Ok after `rosdep update` I saw that that works. I didn't add that because it wasn't working but I was forgetting to update. Thanks

tfoote (approved):
I just filled in the fedora rule. @wjwwood can you rereview with the changes.

wjwwood (approved):

PR #14567, opened 2017-04-19
Add geographiclib python dependency

anassinator:


wjwwood (commented):
Can you please add rules for Debian:

- https://packages.debian.org/stretch/python-geographiclib
- https://packages.debian.org/jessie/python-geographiclib

Also, Fedora if you can find an alternative.

rosdep/python.yaml
```diff
@@ -912,6 +912,21 @@ python-gdown-pip:
   ubuntu:
     pip:
       packages: [gdown]
+python-geographiclib:
+  ubuntu: [python-geographiclib]
+python-geographiclib-pip:
+  debian:
+    pip:
+      packages: [geographiclib]
+  fedora:
```
tfoote: The fedora package is here: https://apps.fedoraproject.org/packages/python-GeographicLib and appears to be available for all recent platforms

tfoote (changes requested):

anassinator:
Done, but not sure if it's right?

wjwwood (approved):
lgtm, thanks!

tfoote:
The new rules for fedora and debian look good. Could you remove the `-pip` rule. If the package exists in upstream packaging the pip install will mask it and can side effect other installations.

anassinator:
What about OSX?

anassinator:
I've removed it as requested regardless.

tfoote (approved):

tfoote:
For OSX if someone needs/wants it we usually look for a homebrew package or pip if necessary and someone needs it on the main rule. The `-pip` separate rule is for things that don't have mainline packages at the moment but we think they might in the future. Using the `-pip` extension make sure they don't collide in the future.

PR #14572, opened 2017-04-19
added python-zbar

niyawe:


mikaelarguedas:
Can you please add a debian entry for this package as well? https://packages.debian.org/search?keywords=python-zbar&searchon=names&suite=all&section=all

mikaelarguedas (approved):
thanks for iterating on this, lgtm

tfoote:
It looks like zbar.so is in `zbar-pygtk` on fedora: https://apps.fedoraproject.org/packages/zbar-pygtk/ It would be great to add that too in parallel with the debian/ubuntu version: http://packages.ubuntu.com/precise/amd64/python-zbar/filelist

tfoote (approved):

PR #14575, opened 2017-04-20
add log_server package to kinetic distribution

lelongg:


lelongg:
Ok, I changed it back.

PR #14598, opened 2017-04-24
Add libpcl-all entries for ubuntu YZ and debian

wkentaro:
For https://github.com/ros-perception/pcl_conversions/issues/25

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2185,48 +2185,8 @@ libpcl-all:
     vivid: [libpcl1.7]
     wily: [libpcl1.7]
     xenial: [libpcl1.7]
-    yakkety:
-      - libpcl-apps1.8
-      - libpcl-common1.8
-      - libpcl-features1.8
-      - libpcl-filters1.8
-      - libpcl-io1.8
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@ros/ros_team this is due to libpcl1.8 not existing on yakkety and newer. I'm fine with the current approach, I don't know if it would be better to create individual rosdep keys for each of the new split PCL packages.

@wkentaro you will need the same thing for the debian rules:
- jessie should use libpcl1.7
- stretch should use 1.8 (the same as yakkety and zesty)

wjwwood (approved):
lgtm, I don't see another way for us to have an aggregate runtime key.

tfoote (changes requested):
I think this is the right approach.

We do need the debian rules as well +1 for @mikaelarguedas 's proposal re jessie vs stretch

mikaelarguedas (approved):
lgtm, thanks for the quick iteration

mikaelarguedas:
@tfoote I'm going to merge this to allow @wkentaro to move forward with the release of pcl packages into lunar

tfoote:
lgtm

PR #14688, opened 2017-04-26
show progress while ordering packages topologically

dirk-thomas:
Show some progress during the test...


mikaelarguedas:
great idea ! I'm always wondering if it's hanging since we added the cyclic dependency test

PR #14704, opened 2017-04-27
Adding pid package to Lunar.

AndyZe:


lunar/distribution.yaml
```diff
@@ -1063,6 +1063,21 @@ repositories:
...
+  pid:
+    doc:
+      type: git
+      url: https://bitbucket.org/AndyZe/pid.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/AndyZelenak/pid-release.git
+      version: 0.0.20-0
```
mikaelarguedas: @AndyZe did you generate this pull request with bloom ?
The version 0.0.20-0 doesn't match the one on the [release repository for lunar](https://github.com/AndyZe/pid-release#pid-lunar---0021-2) (which is 0.0.21-2). That is the reason why all the jobs are failing

lunar/distribution.yaml
```diff
@@ -1063,6 +1063,21 @@ repositories:
...
       version: lunar-devel
     status: maintained
+  pid:
+    doc:
+      type: git
+      url: https://bitbucket.org/AndyZe/pid.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
```
dirk-thomas: This line also references `kinetic`.

Maybe we should extend the tests to detect this?

mikaelarguedas: yeah it looks like it's a copy paste from the [kinetic entry](https://github.com/ros/rosdistro/blob/master/kinetic/distribution.yaml#L4544-L4558).
I think extending the test for this is a good idea! Is there any corner case where the release entry is not of the form `release/<ROSDITRO>/XX` ?

dirk-thomas: I think you might be able to override that but I haven't seen any entry which is different. When implementing the check we will see if any existing entry uses a non-standard entry.

mikaelarguedas:
@AndyZe Would you mind doing the following:
update your packages: `sudo apt-get update && sudo apt-get install python-rosdistro python-bloom`
run bloom to generate a debinc: `bloom-release pid -r lunar -t lunar`
and hit enter bloom prompt you to create a pull request for you ?

Thanks!

lunar/distribution.yaml
mikaelarguedas: yeah I figured we would notice for existing entries, I was just wondering in conceptually there can be a reason to override this. But you're right that if it passes for all existing entries I think it's safe to enforce it and to deal with the corner case if one appears someday.

AndyZe:
Thanks guys. I made the PR manually -- the automatic bloom PR doesn't work since I changed my Github username from AndyZelenak to AndyZe. But I'll correct this.

AndyZe:
The attached picture shows the failure, if you're curious. It should be looking at AndyZe/rosdistro
wrong_repo

mikaelarguedas:
Thanks for updating the rosdistro entry, looks like eveything built fine :tada: 
I think your github username has been kept in ~/.config/bloom. If you delete that file and run bloom again, bloom will prompt you to create a new token for your current github username and it should work next time you make a release.

PR #14832, opened 2017-04-29
add rosdep for cpuburn and lm-sensors

mikeferguson:
Adds two rosdeps for commonly used tools for heat/stress testing robot computers.

This might look like a small bit of duplication as there is a key for libsensors4-dev, which happens to use lm_sensors on arch/fedora, but if you want the command line tools, you really need the full lm_sensors for ubuntu/debian.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3153,6 +3153,7 @@ lm-sensors:
   arch: [lm_sensors]
   debian: [lm-sensors]
   fedora: [lm_sensors]
+  gentoo: [sys-apps/lm_sensors]
   ubuntu: [lm-sensors]
 log4cxx:
   arch: [log4cxx]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -469,6 +469,11 @@ cppunit:
   rhel: [cppunit-devel]
   slackware: [cppunit]
   ubuntu: [libcppunit-dev]
+cpuburn:
+  arch: [cpuburn]
+  debian: [cpuburn]
+  fedora: [cpuburn]
+  ubuntu: [cpuburn]
 crypto++:
   fedora: [cryptopp-devel]
   gentoo: [dev-libs/crypto++]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
lm-sensors seems to be [named lm_sensors](https://admin.fedoraproject.org/pkgdb/package/rpms/lm_sensors/) on fedora
also cpuburn is deprecated and is not released in ubuntu yakkety and xenial as well as in debian stretch (not sure if it exists at all in fedora), I don't know if there is a good maintained equivalent but you may consider other options if every newly released OS don't ship this package anymore.

Not: I haven't checked the arch packages availability

tfoote:
Might stress be a possible alternative to cpuburn? It looks to be genrally available

http://packages.ubuntu.com/xenial/stress
https://packages.debian.org/stretch/stress
https://admin.fedoraproject.org/pkgdb/package/rpms/stress/

http://manpages.ubuntu.com/manpages/zesty/man1/stress.1.html

Alessandro-Barbieri:
`app-benchmarks/cpuburn` is available in gentoo but masked in `amd64/no-multilib` profile
not available at all for non `x86` or `amd64` architectures

rosdep/base.yaml
```diff
@@ -3144,6 +3149,11 @@ linux-headers-generic:
     wheezy: [linux-headers-3.2.0-4-all]
   fedora: [kernel-headers]
   ubuntu: [linux-headers-generic]
+lm-sensors:
+  arch: [lm_sensors]
+  debian: [lm-sensors]
+  fedora: [lm_sensors]
```
Alessandro-Barbieri: add a `gentoo: [sys-apps/lm_sensors]` line

dirk-thomas:
@mikeferguson This needs to be rebased to be mergable.

@tfoote While there might be alternative package I don't think it should prevent this patch to be merged.

tfoote (approved):
+1 after CI I added gentoo and resolved the conflict

mikeferguson:
@tfoote -- thanks for helping on this -- I'm currently on the road.

PR #14842, opened 2017-05-01
add omp dep key

willcbaker:
although omp is installed with gcc compiler, this key is needed to be explicit in deps


mikaelarguedas (approved):
lgtm

PR #14843, opened 2017-05-01
test commits on rviz

wjwwood:


kinetic/distribution.yaml
```diff
@@ -6810,7 +6810,7 @@ repositories:
       url: https://github.com/ros-gbp/rviz-release.git
       version: 1.12.4-0
     source:
-      test_commits: false
+      test_commits: true
```
dirk-thomas: Since this is the default you might want to remove this line.

PR #14859, opened 2017-05-02
Moved collada_parser and collada_urdf to a new repo

sloretz:
This is a manual pull request because collada_parser and collada_urdf have been moved to a new repository.

I ran bloom up to the point where it warned `Failed to open pull request: AssertionError - Duplicate package name 'collada_parser' exists in repository 'robot_model' as well as in re
pository 'collada_urdf'`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1771,6 +1771,25 @@ repositories:
       url: https://github.com/ipa320/cob_substitute.git
       version: indigo_dev
     status: maintained
+  collada_urdf:
+    doc:
+      type: git
+      url: https://github.com/ros/collada_urdf.git
+      version: indigo-devel
+    release:
+      packages:
+      - collada_parser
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -10239,8 +10258,6 @@ repositories:
       version: indigo-devel
     release:
       packages:
-      - collada_parser
-      - collada_urdf
```
dirk-thomas: Is it ensured that future releases of this repo don't try to re-release these two packages?

sloretz: Yes. Should I merge the removal of these two packages from the robot_model repo before this PR is merged? https://github.com/ros/robot_model/pull/197

dirk-thomas: Not necessarily. I just wanted to make sure that it is being handled.

PR #14886, opened 2017-05-04
add Y/Z for wxpython

k-okada:
see http://packages.ubuntu.com/cgi-bin/search_packages.pl?keywords=python-wxgtk&searchon=names&subword=1&version=dapper&release=all

and

https://packages.debian.org/search?keywords=python-wxgtk


mikaelarguedas (approved):
Lgtm

PR #14938, opened 2017-05-08
Revert "gdown: 3.2.6-0 in 'kinetic/distribution.yaml' [bloom]"

wkentaro:
Reverts ros/rosdistro#14932

Sorry, I found https://github.com/ros-infrastructure/ros_release_python and think it must be used for releasing python-gdown.


tfoote:
Note, Please be a little more careful about your revert commits. This had 3 commits inside it, but most of them were empty. When we merge we squash, but it's not good to reference reverting things that you're not actually reverting.

wkentaro:
I see. Thanks.

PR #14959, opened 2017-05-10
Added Python-argh to available packages

yotabits:
This package is needed for the further release of python-watchdog R.O.S. package

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3198,8 +3198,6 @@ xdot:
   fedora: [python-xdot]
   ubuntu: [xdot]
 python-argh:
-  ubuntu:
-    trusty:[python-argh]
-    xenial:[python-argh]
-    yakkety:[python-argh]
-    zesty:[python-argh]
+  debian: [python-argh]
+  fedora: [python-argh]
+  ubuntu:[python-argh]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3197,3 +3197,9 @@ xdot:
   debian: [xdot]
   fedora: [python-xdot]
   ubuntu: [xdot]
+python-argh:
+  ubuntu:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3197,3 +3197,9 @@ xdot:
   debian: [xdot]
   fedora: [python-xdot]
   ubuntu: [xdot]
+python-argh:
+  ubuntu:
```
tfoote: It looks like python-argh is generally available on all versions. Could you change this rule to not call out specific versions: http://packages.ubuntu.com/search?keywords=python-argh

tfoote (changes requested):
If you want to make a release a debian rule will be necessary: https://packages.debian.org/search?keywords=python-argh

As well as a fedora rule: https://admin.fedoraproject.org/pkgdb/package/rpms/python-argh/

And the ubuntu rule can be collapsed to a single line for easier reading and maintenance.

tfoote:
Note that python-watchdog is already available upstream, https://packages.debian.org/search?keywords=python-watchdog. If you want to use that you don't need to add rules for all the dependent packages.

yotabits:
@tfoote Updated :)

tfoote (approved):

tfoote:
This looks better now. However I need to ask how you are planning to release python-watchdog for trusty. We don't support releasing arbitrary generic python packages. They need to be ros packages in a rosdistro.

yotabits:
Actually I already have a functional ROS package of watchdog and one his specific dependency not available as a ROS package for now.

To make it I used [catkin_pip](https://github.com/asmodehn/catkin_pip), for now it works like a charm.

tfoote:
Ok great. We need a 2nd review/merger from @ros/rosdistro-maintainers

PR #14960, opened 2017-05-10
Added PyYAML to available packages

yotabits:
This package is needed for the further release of [python-watchdog](https://github.com/gorakhargosh/watchdog) package

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3197,3 +3197,7 @@ xdot:
   debian: [xdot]
   fedora: [python-xdot]
   ubuntu: [xdot]
+py-yaml-pip:
+  ubuntu:
+    pip:
+      packages: [PyYAML]


tfoote (changes requested):
This will conflict with/override the python-yaml rule already in use. And potentially cause side effects for any user of that package.

I don't think we want to accept this as it can cause issues with the system installed version. Our general policy is to use the system installed version if there is one.

tfoote:
Note that python-watchdog already appears to be packaged as well: https://packages.debian.org/search?keywords=python-watchdog and it uses python-yaml.

yotabits:
> Note that python-watchdog already appears to be packaged as well[...]

@tfoote yes but I need to use it on Ubuntu trusty,
according to this http://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python-watchdog&searchon=names it's not available..

> I don't think we want to accept this as it can cause issues with the system installed version.

Totally agree with, I didn't saw it -_-'

PR #14986, opened 2017-05-14
add asio library as header-only alternative to boost::asio

christianrauch:
asio library for ubuntu, debian, fedora, arch


dirk-thomas (approved):

mikaelarguedas:
Note that (at least for apt) installling asio results in installing and using boost by default. It bit us recently https://github.com/ros2/ci/issues/61

PR #15015, opened 2017-05-21
Add rosdep for msgpack

akio:
This adds [MessagePack](http://msgpack.org/) library package.
MessagePack is a efficient binary serialization format.


mikaelarguedas:
given that all ubuntu rules are identical, can you please collapse then into a single one `ubuntu: [libmsgpack-dev]` ?

akio:
Done.

mikaelarguedas (approved):
lgtm, thanks for quick iteration

PR #15030, opened 2017-05-22
Add smbclient to rosdep/base.yaml

HEllRZA:
**What, the Swiss army knife of the Samba suit is missing?**

We are using this in a project, so it would be great to have it here.
Initially we wanted to use pysmb but had not the best results on some devices.
smbclient just worked where pysmb failed. You might find this self-claimed "ugly hack" useful: https://bitbucket.org/nosklo/pysmbclient/overview

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3853,6 +3853,8 @@ smartmontools:
   debian: [smartmontools]
   fedora: [smartmontools]
   ubuntu: [smartmontools]
+smbclient:
+  ubuntu: [smbclient]
 snappy:
   arch: [snappy]
   fedora: [snappy]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3854,6 +3854,9 @@ smartmontools:
   fedora: [smartmontools]
   ubuntu: [smartmontools]
 smbclient:
+  arch: [smbclient]
+  debian: [smbclient]
+  fedora: [samba-client]
   ubuntu: [smbclient]
 snappy:
   arch: [snappy]


dirk-thomas (approved):

rosdep/base.yaml
```diff
@@ -3853,6 +3853,11 @@ smartmontools:
   debian: [smartmontools]
   fedora: [smartmontools]
   ubuntu: [smartmontools]
+smbclient:
+  arch: [smbclient]
+  debian: [smbclient]
+  fedora: [samba-client]
```
mikaelarguedas: this package doesn't seem to be available on fedora official package database: https://admin.fedoraproject.org/pkgdb/packages/samba*/

HEllRZA:
I must admit, i googled it and found that name. Will install a new Fedora into a VM tomorrow and check if yum install samba-client succeeds without adding any additional package repository.

clalancette:
Just FYI, on my Fedora 25 laptop:

```
$ which smbclient
/usr/bin/smbclient
$ rpm -qf /usr/bin/smbclient
samba-client-4.5.8-1.fc25.x86_64
```

It may not have shown up in the pkgdb search because the source package is "samba", not "samba-client".

mikaelarguedas:
yes it is part as samba as listed in the link I posted above https://admin.fedoraproject.org/pkgdb/packages/samba*/.
The source package does include the server as well that's why I pointed it out. They result in separate binary packages so adding `samba-client` as is is valid :+1:

HEllRZA:
👍

PR #15033, opened 2017-05-22
	add debian python-path.py

rotu:


mikaelarguedas:
@rotu platforms need to be listed in alphabetical order

mikaelarguedas:
@ros/ros_team not sure how we should go from here.
- This rule uses only pip versions of this package. Should we create a new key suffixed with `-pip` rather than adding new rules to this one?
- This package is available upstream on most platforms.[Ubuntu](https://packages.ubuntu.com/search?keywords=python-path&searchon=names&suite=all&section=all) [Debian](https://packages.debian.org/search?keywords=python-path&searchon=names&suite=all&section=all), [Fedora](https://admin.fedoraproject.org/pkgdb/package/rpms/python-path/) should we create a new key targeting this upstream version ?

dirk-thomas:
It depends on if the Ubuntu package version is "new enough" for the ROS packages which currently depend on this key. If yes, I would say change the existing key (and only if there is a need for it also add a `-pip` entry). Even if no it should be considered to add the `-pip` entry and update existing ROS packages to use that one in order to update the non-pip entry to use the upstream packages.

mikaelarguedas:
@dirk-thomas Is there an automated way to find out which packages are using this key ? Given that it's pip I assume only non released packages use it but I don't know how to get a list of such packages

dirk-thomas:
You could search through the repos registered with a source entry. But that doesn't cover any code not listed. So I would say no.

Alessandro-Barbieri:
`dev-python/path-py` in gentoo (no pip)

dirk-thomas:
@rotu This needs to be rebased to be mergable.

dirk-thomas:
Closing due to no response.

PR #15050, opened 2017-05-24
bump linux-headers version for debian stretch

mathias-luedtke:
https://packages.debian.org/stretch/linux-headers-4.9.0-3-all

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3158,7 +3158,7 @@ linux-headers-generic:
   arch: [linux-headers]
   debian:
     jessie: [linux-headers-3.16.0-4-all]
-    stretch: [linux-headers-4.9.0-2-all]
+    stretch: [linux-headers-4.9.0-3-all]
     wheezy: [linux-headers-3.2.0-4-all]
   fedora: [kernel-headers]
   ubuntu: [linux-headers-generic]


dirk-thomas (approved):
Does Debian not have a package like "latest" to avoid the continuous updates?

ipa-mdl:
> Does Debian not have a package like "latest" to avoid the continuous updates?

I haven't found any.. But I am no debian user.
I just want to silence the build farm errors ;)
There are architecture-specific ones, like https://packages.debian.org/stretch/linux-headers-amd64

ipa-mdl:
For my use case `linux-libc-dev` (ubuntu & debian) would work as well.

mikaelarguedas:
It looks like both `linux-headers-generic` and `linux-headers` are considered obsolete and cannot be installed (tested on Jessie and Stretch): 
```
Package linux-headers-generic is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
```
So I don't see any other "generic" package candidate that would prevent us from having to update them manually..

ipa-mdl:
> Edit: we could create another key for the C development version (linux-libc-dev) as suggested by 

Please merge this for now to fix the version problem on debian stretch.
I will open a new PR for `linux-kernel-headers` (virtual package for `linux-libc-dev`)

dirk-thomas:
I already approved this patch. It needs a second maintainer to merge.

PR #15059, opened 2017-05-25
Added xeinial key for python-usb and python-beautifulsoup

terakoji:
This added xenial keys for two packages.
There are keys for prevous ubuntu distributions.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -264,17 +264,7 @@ python-beautifulsoup:
   arch: [python2-beautifulsoup3]
   debian: [python-beautifulsoup]
   fedora: [python-BeautifulSoup]
-  ubuntu:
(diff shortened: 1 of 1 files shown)


dirk-thomas (approved):
It would be great if you could add the entries for the other Ubuntu distros too.

mikaelarguedas:
an alternative to entering each distro by hand is to use the collapsed version of the rules `ubuntu: [python-beautifulsoup]` and `ubuntu: [python-usb]` (given that the rules are the same for all ubuntu distributions)

dirk-thomas (approved):

terakoji:
I don't know which is better.
#. using only ubuntu key.
#. using ubutu key and distribution keys.
I pushed first one.

PR #15070, opened 2017-05-25
use non-virtual package for linux-kernel-headers

mathias-luedtke:
should be reverted as soon as new rosdep version (with support for virtual packages) was released.
(as discussed in https://github.com/ros/rosdistro/pull/15064#issuecomment-304127344)


mikaelarguedas (approved):
lgtm

PR #15078, opened 2017-05-26
more python dependencies for gentoo

Alessandro-Barbieri:


allenh1:
lgtm

Alessandro-Barbieri:
https://github.com/ros/rosdistro/issues/15082 for reference

rosdep/python.yaml
```diff
@@ -2296,6 +2384,7 @@ python-rosdistro:
 python-rosinstall:
   arch: [python2-rosinstall]
   fedora: [python-rosinstall]
+  gentoo: [dev-util/rosinstall]
```
allenh1: Please change this to `dev-python/rosinstall`. We'll fix that one if it's not working properly.

rosdep/python.yaml
```diff
@@ -2311,6 +2400,7 @@ python-rosinstall:
 python-rosinstall-generator:
   arch: [python2-rosinstall-generator]
   fedora: [python-rosinstall_generator]
+  gentoo: [dev-util/rosinstall_generator]
```
allenh1: Please change to `dev-python/rosinstall_generator`

rosdep/python.yaml
```diff
@@ -3105,17 +3227,20 @@ python-websocket:
...
   fedora: [python-werkzeug]
+  gentoo: [dev-python/werkzeug]
   ubuntu: [python-werkzeug]
 python-ws4py-pip:
   ubuntu:
     pip:
       packages: [ws4py]
 python-wstool:
   fedora: [python-wstool]
+  gentoo: [dev-util/wstool]
```
allenh1: please use `dev-python/wstool`

rosdep/python.yaml
Alessandro-Barbieri: Does have something special?

rosdep/python.yaml
Alessandro-Barbieri: It's different from gentoo version?

rosdep/python.yaml
allenh1: I mean, I for one am very strongly opposed to mixing them. We maintain ours, and we don't want the downstream people to be ahead/behind the official repo. If we are behind, please file an issue requesting that the ebuild be updated.

rosdep/python.yaml
allenh1: Maybe not at the moment, but it's not _our_ version.

rosdep/python.yaml
Alessandro-Barbieri: `rosinstall_generator::ros-overlay` it's outdated

Alessandro-Barbieri:
Since these tool are already in gentoo they are to be the default choice

rosdep/python.yaml
allenh1: Then I'll update it

rosdep/python.yaml
Alessandro-Barbieri: Overlay `rosinstall` isn't different from gentoo one

rosdep/python.yaml
Alessandro-Barbieri: And it's isn't different from gentoo one

Alessandro-Barbieri:
Checked  the overlay ebuilds, they don't have anything special apart bad being outdated, dependencies and python issues

allenh1:
@Alessandro-Barbieri while that may be true, my point is that it may cease to be true in the future.

For instance, maybe the Gentoo people decide that, since the official install method has a much larger set of packages and is updated far more frequently, that they will be more lax about maintaining their ebuilds.

I feel it would be far more simple to make the official set of installs self-contained, so it is my opinion that we should refrain from using the downstream version wherever possible.

Alessandro-Barbieri:
Gentoo is about choice so change the category name and let the user choose if using the overlay or not (isn't needed for now since downstream have the tools for a source based install)

Alessandro-Barbieri:
'Now' they have the same tools as here, if in future some ebuilds will need change, you can change it (and if you retain the same category the 'upgrade' will be seamless by just adding the overlay and pull the fixed version)

allenh1:
> 'Now' they have the same tools as here, if in future some ebuilds will need change, you can change it (and if you retain the same category the 'upgrade' will be seamless by just adding the overlay and pull the fixed version)

In hindsight, that would be the solution. But, the same point could be made for those using the `dev-python` versions, which is the method mentioned in the official install guide.

This isn't about a _mine_ or _theirs_ thing, this is a consistency thing. The goal here is to keep everything consistent and without mixing our repo and their repo, regardless of content.

The last reason I have for not changing the name is that our ebuilds actually predate the official ones. I am also not sure why we should go to the `dev-util` category, since these are python api's which I think should belong in the `dev-python` category.

I'm not trying to call you wrong here, as you are by no means wrong. I am simply clarifying that it would be my preference to use the existing category, rather than change to another one.

Alessandro-Barbieri:
The guide can be changed easily, old users can transition without much trouble.

Alessandro-Barbieri:
IMHO they weren't put in the `dev-python` category because these tools aren't python libraries
for example `gentoolkit` is written in python but it's in `app-portage`

Alessandro-Barbieri:
@allenh1 ok? I removed the category so it will work with or without the overlay

allenh1:
So, that doesn't help me with ebuild generation... Portage became very unhappy when dependencies didn't have a category attached to them... Otherwise that would be the best solution.

mikaelarguedas:
As a non Gentoo expert and not familiar with the exact meaning or use of categories, I'd vouch for having a behavior as close as the other systems. For Ubuntu or Debian, ROS packages, even if available upstream are downloaded from our apt repo with our latest released version. So whichever category reproduces this behavior is the one that should be used IMO.
If the Gentoo community wants to support installing ROS using their version of these packages (that's awesome!) but doesn't have to be reflected here.

Alessandro-Barbieri:
There is not 'our' or 'their' version: only upstream version matter and gentoo provides exactly what can be found upstream. Gentoo *does not* modify code of what is found upstream, so why worry? Why provide the same thing in an incompatible way? The ebuilds in the overlay doesn't add *nothing* apart of being written worst than downstream and providing outdated version.

Alessandro-Barbieri:
@allenh1 how to you query portage to generate the ebuilds?

allenh1:
I wrote a custom system (note: not completed): https://github.com/allenh1/ebuild-generator

Alessandro-Barbieri:
Crazy idea here: for example create a fake `dev-util/rosinstall` that rdepends on `dev-python/rosinstall`

allenh1:
While that would work, that's still not my preference.

Alessandro-Barbieri:
@allenh1 

>Otherwise that would be the best solution.

Trys this code (you'll need `eix`)

```python
#check if pkg does not contain the category an then searches in the overlay for the <category>/<name> matching pkg

 if "/" in pkg:
    pkg=subprocess.Popen("eix --exact %s --in-overlay ros-overlay --only-names" %pkg, shell=True,  stdout=subprocess.PIPE, stderr=subprocess.STDOUT).stdout.read().decode('
    ...: ascii')
```

allenh1:
I think we're going to update the existing ones, keeping the `dev-python` prefix. The primary reason here is that I don't want to break existing installations/confuse portage when ebuilds suddenly disappear.

If we change later, that's fine, but I'd rather get the dependencies pulled in and later have a separate discussion about the `dev-python`/`dev-util` issue. @Alessandro-Barbieri is that fair?

Alessandro-Barbieri:
>I don't want to break existing installations/confuse portage when ebuilds suddenly disappear.

You will not break anything, if you use that snippet it will (when you generate the ebuilds) convert, for example, `rosinstall` in `dev-python/rosinstall`
`eix` is super fast (really: `0,211 s` for that command) and will run only for the few packages listed without a category

Alessandro-Barbieri:
Well, do what you want, I want to try to installing using the old method.
Please don't nuke this pull request, it's possible to add a pull request after?

allenh1:
Well, do what you want, I want to try to installing using the old method.
Please don't nuke this pull request, it's possible to add a pull request after?

That would be greatly preferred by me

Alessandro-Barbieri:
@allenh1 since workaround exist I made the changes.
Ready to be merged.

rosdep/python.yaml
Alessandro-Barbieri: done

rosdep/python.yaml
Alessandro-Barbieri: done

rosdep/python.yaml
Alessandro-Barbieri: done

Alessandro-Barbieri (commented):
Time to merge

Alessandro-Barbieri:
@allenh1 can you rewiev the PR? Does it need other changes?

Alessandro-Barbieri:
@allenh1 @mikaelarguedas why is taking so long? It's something missing?

mikaelarguedas (approved):

allenh1:
+1

PR #15086, opened 2017-05-26
Build failed due to misconfigured python.yaml

dingelish:
A bunch of users noticed that ROS won't build on their systems. On my laptop, the rosdep fails on
`webkit_dependency: No definition of [python-qt-bindings-webkit] for OS version [xenial]`
I think this is an acceptable fix for my lapotp (ubuntu 16.04). But we need more lines for other environments.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2122,6 +2122,7 @@ python-qt-bindings-webkit:
     trusty: [python-qt4]
     utopic: [python-qt4]
     vivid: [python-qt4]
+    xenial: [python-qt4]
 python-qt4-gl:
   debian: [python-qt4-gl]
   fedora: [PyQt4]


mikaelarguedas:
ROS Kinetic and later target qt5 (http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021). The key used by Ubuntu Xenial and later is a few lines below: https://github.com/ros/rosdistro/pull/15086/files#diff-701c2f61a9f7ac8fc393ccc6630c8fa5R2159. Could you give more details about the packages that fail to build ? they should be updated to use the new key in their package.xml

dingelish:
Hi there, I'm doing a complete fresh compile in a entirely clean Ubuntu 16.04 virtual machine, following the guide at http://wiki.ros.org/jade/Installation/Source . I used Desktop Install (recommended) commands for set up. Then I met this error:

```
ding@ubuntu:~/ros_catkin_ws$ rosdep install --from-paths src --ignore-src --rosdistro jade -y
to system dependencies:
webkit_dependency: No definition of [python-qt-bindings-webkit] for OS version [xenial]
```
For my own fix, I edited `src/webkit_dependency/package.xml` and removed the dependency line and everything works fine.

wjwwood:
@dingelish Jade doesn't support Ubuntu Xenial.

https://github.com/ros-infrastructure/rep/blob/master/rep-0003.rst#jade-turtle-may-2015---may-2017

We can add the key here, but you're likely to run into other issues. I'd recommend using Kinetic instead of Jade.

dingelish:
My Bad :)

Issue #15082, opened 2017-05-26
Missing Dependencies for Gentoo Linux


allenh1:
Here's the list from my script:

 * [ ] disper
 * [ ] dkms
 * [ ] g++-static
 * [ ] gstreamer0.10-pocketsphinx
 * [ ] ipython
 * [ ] language-pack-de
 * [ ] language-pack-en
 * [ ] leg_detector
 * [ ] libcap-dev
 * [ ] libcoin80-dev
 * [ ] libesd0-dev
 * [ ] libfreenect-dev
 * [ ] mapnik-utils
 * [ ] mrpt
 * [ ] nodejs-legacy
 * [ ] npm
 * [ ] opende
 * [ ] people_velocity_tracker
 * [ ] python-argcomplete
 * [ ] python-argh
 * [ ] python-avahi
 * [ ] python-bloom
 * [ ] python-bs4
 * [ ] python-bson
 * [ ] python-catkin-lint
 * [ ] python-catkin-tools
 * [ ] python-chainer-pip
 * [ ] python-cheetah
 * [ ] python-cherrypy
 * [ ] python-click
 * [ ] python-colorama
 * [ ] python-concurrent.futures
 * [ ] python-couchdb
 * [ ] python-crypto
 * [ ] python-dateutil
 * [ ] python-dlib
 * [ ] python-docutils
 * [ ] python-enum34
 * [ ] python-expiringdict
 * [ ] python-fcn-pip
 * [ ] python-flask
 * [ ] python-gdown-pip
 * [ ] python-gi
 * [ ] python-gst
 * [ ] python-h5py
 * [ ] python-mapnik
 * [ ] python-mechanize
 * [ ] python-oauth2
 * [ ] python-omniorb
 * [ ] python-percol
 * [ ] python-pip
 * [ ] python-progressbar
 * [ ] python-pygame
 * [ ] python-pygithub3
 * [ ] python-pymongo
 * [ ] python-pytest
 * [ ] python-pyudev
 * [ ] python-requests
 * [ ] python-rosdistro
 * [ ] python-rosinstall
 * [ ] python-serial
 * [ ] python-simplejson
 * [ ] python-singledispatch
 * [ ] python-six
 * [ ] python-skimage
 * [ ] python-sklearn
 * [ ] python-slacker-cli
 * [ ] python-sqlalchemy
 * [ ] python-sympy
 * [ ] python-tabulate-pip
 * [ ] python-termcolor
 * [ ] python-texttable
[...]

PR #15090, opened 2017-05-28
ruby dependencies for gentoo

Alessandro-Barbieri:


Alessandro-Barbieri:
https://github.com/ros/rosdistro/issues/15082 for reference

rosdep/ruby.yaml
```diff
@@ -77,7 +85,9 @@ ruby:
 ruby-sass:
   debian: [ruby-sass]
   fedora: [rubygem-sass]
+  gentoo: [dev-ruby/sass]
   ubuntu: [ruby-sass]
 ruby1.9.3:
+  gentoo: []
```
mikaelarguedas: do you know of any package using this ? I think this can be removes given that no released package on the buildfarm uses it

mikaelarguedas (approved):
lgtm, except one comment about adding rules to unused keys

rosdep/ruby.yaml
Alessandro-Barbieri: It's upstream deprecated and no more available on gentoo. I don't know any package that uses it.

mikaelarguedas: feel free to remove this line then

mikaelarguedas:
LGTM

PR #15095, opened 2017-05-29
Added watchdog to jade/distribution.yaml

yotabits:
Having some issues on bloom while doing release PR for watchdog package so doing it by hand.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8389,6 +8389,13 @@ repositories:
       url: https://github.com/AutonomyLab/parrot_arsdk.git
       version: indigo-devel
     status: developed
+  pathtools:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/yotabits/pathtools-release.git
+      version: 0.1.2-0
+    status: maintained
   patrolling_sim:
     doc:
       type: git
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8389,13 +8389,6 @@ repositories:
       url: https://github.com/AutonomyLab/parrot_arsdk.git
       version: indigo-devel
     status: developed
-  pathtools:
-    release:
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/yotabits/pathtools-release.git
-      version: 0.1.2-0
-    status: maintained
   patrolling_sim:
     doc:
       type: git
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Is this the same as the Debian package `python-watchdog`? If yes, should it be named the same?

yotabits:
Hi,

Not exactly the same because the Debian package for `python-watchdog` is only available for [16.04](https://packages.ubuntu.com/search?keywords=python-watchdog&searchon=names&suite=yakkety&section=all). 
I'm actually releasing it for Jade and Indigo as I need to use it on Ubuntu trusty.

dirk-thomas:
So what you are saying is that it is exactly the same package. You only want to make it available in older distros.

In that case it would even make more sense to name the package the same (`python-watchdog`) so that packages depending on it don't have to change their rosdep key when moving from e.g. Trusty to Xenial.

yotabits:
@dirk-thomas I followed your guidance

mikaelarguedas (approved):
lgtm

dirk-thomas:
I was actually suggesting `python-watchdog` (note the dash instead of the underscore). Otherwise it would still require ROS packages to change the rosdep key in the manifest file between ROS distros.

tfoote:
@dirk-thomas we've had trouble in the past trying to transition packages from ros packages to upstream packages in the past. They were c++ packages which ended up with namespace collisions etc since if any supported ubuntu distro for the rosdistro had both the upstream version and the ros version. Maybe this will work for python packages. But in general my suggestion is not to try to mask the changeover from ROS package to rosdep. They generally have different versions, especially once the upstream package is used across multiple Ubuntu distros, it's almost guarenteed to be different versions.

dirk-thomas:
@tfoote The recent relaxation of the package name rules in the ROS REPs 127 and 140 (see ros-infrastructure/rep#126) aim exactly for this use case: to use the same name as a Debian package (e.g. `python-watchdog`) to allow releasing it as a ROS package into "older" Ubuntu distros. That way the rosdep key is the same and other ROS packages don't need to branch / care. That use case was specifically described when the REP change was proposed.

yotabits:
@dirk-thomas for my self I agree with @tfoote even I'm only transforming the package source code into a ROS package I'm already making things slightly different just by doing that.
But I will follow the rules as it has to be done.

dirk-thomas (approved):
Looks good to me. I will leave it up to @tfoote to merge since he was part of the discussion before.

tfoote:
Since this was a use case the relaxed rules were desiged to cover lets try it.

Unfortunately this is now conflicting and needs a rebase before it can be merged @yotabits could you take a look?

PR #15127, opened 2017-06-01
Removing indigo: move_base_to_manip (don't care to maintain it).

AndyZe:


dirk-thomas:
If the existing Indigo package works why not keep it as is? Or is there a specific reason to pull the package?

AndyZe:
There's a kind-of-significant bug that's getting patched for Kinetic.

dirk-thomas:
Your call...

PR #15152, opened 2017-06-05
Added pathtools 

yotabits:
Previously released pathtools for Jade, now doing it for Indigo


mikaelarguedas:
Is this package the same as [`python_pathtools`](https://packages.ubuntu.com/search?keywords=pathtools&searchon=names&suite=all&section=all) ? If so, please rename it to `python_pathtools` to have the same name as the debian package and keep rosdep key the same for all distributions.

dirk-thomas:
If it is not named `python-pathtools` (note the dash instead of the underscore) it would require ROS packages to change the rosdep key in the manifest file between ROS distros.

PR #15153, opened 2017-06-05
add w1thermsensor for debian

jakerye:
[w1thermsensor](https://pypi.python.org/pypi/w1thermsensor) is useful for interfacing with a [ds18b20 temperature sensor](https://www.adafruit.com/product/381) on a raspberry pi


mikaelarguedas:
can you add an ubuntu entry as well ?

dirk-thomas (approved):

PR #15154, opened 2017-06-05
Added all necessary gentoo dependencies for ros lunar.

allenh1:
This just pulls in those necessary to satisfy all deps for currently released ros lunar packages.


mikaelarguedas (approved):
lgtm

PR #15167, opened 2017-06-06
[WIP][test] Check if wiki.ros.org/PKG_NAME exists.

130s:
**Background**
Currently AFAIK there's no mechanism to check if a page on wiki.ros.org for the packages listed on rosdistro exist. Some/many of them are missing because of that.

**Approach**
Check if the corresponding wiki page exists (i.e. not "does not exist").

**Limitation**
At the time of this commit, the added check is commented out, as it does NOT work as intended
(with the commented-out code, test checks all listed packages and accesses wiki.ros.org, most of which result in http status code 503, probably due to web server's regulation).

Checking the packages that are ONLY included in the pull request should improve the situation, but the author couldn't implement a way to do so.

Question-1. Is this feature something we might want to add?
Question-2. Is there already way existing to get package names from the pull request diff?

### DIFF
--- a/test/test_build_caches.py
+++ b/test/test_build_caches.py
@@ -1,5 +1,7 @@
 from collections import OrderedDict
+import httplib
 import os
+from urlparse import urlparse
 
 from catkin_pkg.package import parse_package_string
 from ros_buildfarm.common import topological_order_packages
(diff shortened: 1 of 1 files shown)


test/test_build_caches.py
```diff
@@ -34,6 +54,12 @@ def test_build_caches():
                 else:
                     caches[dist_name] = cache
 
+                #for pkg_name, pkg_xml in cache.release_package_xmls.items():
```
130s: Commented out this block to avoid unnecessary http accesses.

dirk-thomas:
What is the goal here? Do you want to enforce that the wiki page exists? Imo it is not a good idea. It just adds jyet another step we force users to do even if they have no intention of ever adding any information to that page. A generated index like rosindex.github.io is imo a much better approach to ensure that a page with generic package information is available for every package.

tfoote:
> Question-1. Is this feature something we might want to add?

I think this is valuable. However I'm not sure that having it in the CI here is quite the right place. Encouraging people to make the pages is good. And having an automated way to query it is also good. However coupling the wiki availability/connectivity to the cache check is not something I think that we want to do. This would also possibly add a large load to the wiki. There's a reason you're tests are hitting rate limits.

> Question-2. Is there already way existing to get package names from the pull request diff?
There is code for detecting changed sections in https://github.com/ros/rosdistro/blob/master/test/test_url_validity.py

Alessandro-Barbieri:
Travis is stuck here too

dirk-thomas:
The ticket already has comments suggesting that a CI test enforcing the existence of a wiki page might not be desired. So triggering Travis isn't necessary.

130s:
I do appreciate inputs I got and am willing to update my suggestion accordingly when I can. Meanwhile I can leave this closed if this bothers anything.

PR #15175, opened 2017-06-06
bump OpenCV3 in Lunar

vrabaud:
no change, this is to try if that fixed the farm.

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -1322,7 +1322,7 @@ repositories:
       tags:
         release: release/lunar/{package}/{version}
       url: https://github.com/ros-gbp/opencv3-release.git
-      version: 3.2.0-7
+      version: 3.2.0-8
     status: maintained
   openni_camera:
     doc:


mikaelarguedas:
@vrabaud sorry for the misunderstanding, can you try to bump the version number rather than doing a debinc ? I think that if you create a new orig tarball with your binaries files in it it will work.

tfoote:
So it looks like dpkg-source by default rejects binary file changes when computing the deb-inc diff, ala `dpkg-source: error: cannot represent change to doc/tutorials/core/basic_linear_transform/images/Basic_Linear_Transform_Tutorial_gamma.png: binary file contents changed`

Most others are seeing this when they accidentally add things like `.pyc` files or vim `.swp` files by accident.

There is a thread here: https://stackoverflow.com/questions/9677971/having-trouble-building-python-deb-package-complain-about-modified-binary that suggests that there might be an option to relax that with `--include-binaries` as an option. Of course that's not a place where we have configurability in the pipeline. 

There are also less helpful suggestions like putting them into [XPM format](https://lists.debian.org/debian-mentors/2004/12/msg00059.html) or to `uuencode them` [here](https://talk.maemo.org/showthread.php?t=52808)

As @mikaelarguedas mentioned the easiest solution is probably to bump the upstream version.

PR #15183, opened 2017-06-07
Revert Opencv3 debian increments

tfoote:
This is rolling back the OpenCV3 changes that are failing to build. The binary changes(to pngs) could not go though the sourcedeb debian diff generation.

Talking with @vrabaud OpenCV 3.2.1 is coming out in a week. We will roll back and then we can push the new version in a week or so.


tfoote:
This connects to https://github.com/ros-perception/vision_opencv/issues/170

Closing and reopening to retrigger travis

PR #15208, opened 2017-06-12
gentoo dependencies

Alessandro-Barbieri:


rosdep/base.yaml
```diff
@@ -932,6 +933,7 @@ gstreamer0.10-plugins-ugly:
     wily: [gstreamer0.10-plugins-ugly]
 gstreamer0.10-pocketsphinx:
   fedora: [pocketsphinx-plugin]
+  gentoo: [app-accessibility/pocketsphinx]
```
leio: After curious wandering around from esound bug, I shall note that gstreamer0.10 using pocketsphinx will be removed from Gentoo as soon as I finally get around to solving it. gstreamer0.10 is known security vulnerable and it is irresponsible to pull it in.

Alessandro-Barbieri: I only added the gentoo packages names. ROS still use old packages like `gstreamer0.10` or `QT4` and they have to decide to remove them (from ALL distros then) and to block users to depend on outdated packages.

Alessandro-Barbieri: https://github.com/ros/rosdistro/issues/15210

rosdep/base.yaml
```diff
@@ -3294,6 +3297,7 @@ nodejs-legacy:
 npm:
   debian: [npm]
   fedora: [npm]
+  gentoo: [net-libs/nodejs]
```
tfoote: Does the nodejs package really come with npm on gentoo? npm is typically a layer on top of nodejs.

leio: $ qfile /usr/bin/npm
net-libs/nodejs (/usr/bin/npm)

leio: Actually technically the dependency is `net-libs/nodejs[npm]`, that is, the npm USE flag has to be enabled on it. But I don't know if your base.yaml gentoo key value supports any of that

Alessandro-Barbieri: AFAIK you can't `emerge net-libs/nodejs[npm]` (or something not involving the `USE` variable, e.g. some commandline switch for portage) and I don't know if someone want to change `rosdep` in order to support gentoo useflags.

@allenh1 TODO: compile a list of required package/use

allenh1: @Alessandro-Barbieri I, unfortunately, do not know the `USE` flags required for most ROS packages, and don't see it as a worth-while endeavor to go through every pack by hand and determine it... I also can't think of a good way to script it. So I think my standpoint here will be enough to be Ubuntu-esque, and, since Ubuntu has switched to Gnome 3, I think that can be relaxed to "enough to support Gnome 3."I

Alessandro-Barbieri: I'll provide you a list

leio: I can't comment on the usefulness of requiring certain USE flags or not. I think it can be counterproductive for cases where the USE flag controls the building of some feature into it as they normally do (that is, affecting the outcome of a given installed binary or such).
But in case of nodejs, the thing is that when there's no USE=npm, there will be no npm package manager installed at all. Albeit USE=npm is the default, but user could have disabled it and then your rosdep thing doesn't force it back.
So there's USE flags affecting what libraries are linked to (and therefore probably also what features are supported, or what library is used for some feature - e.g ssl), and USE flags that control if some part of the source package is built/installed or not.
Binary distributions simply build the source package and split it into multiple binary packages afterwards, but Gentoo as a source distribution would have to do it at source package level, as that's what we have primarily. Often making the tarball be able to build things separately is too heavy of a maintenance burden to be feasible for the maintainer and then a USE flag is used to control that.

Alessandro-Barbieri: The useflags can't be specified here because of the portage syntax: unlike other package managers, you have to set a variable before calling the executable and this need a rework on how `rosdep` handles the package managers.
see https://bugs.gentoo.org/show_bug.cgi?id=621608

Alessandro-Barbieri: @leio thanks, didn't know of this
This list is used by `rosdep` to emerge the needed packages if you use the manual install and to generate ebuilds for the `ros-overlay`

rosdep/base.yaml
```diff
@@ -715,6 +715,7 @@ g++-static:
   arch: [gcc]
   debian: [g++]
   fedora: [gcc-c++, glibc-devel, glibc-static, libstdc++-devel, libstdc++-static]
+  gentoo: [sys-devel/gcc]
   ubuntu: [g++]
```
allenh1: Does this guarantee that it's static?

rosdep/base.yaml
allenh1: So does the bracket syntax work? I could add this feature to the generator if so, then we can handle the `USE` flags on an as-needed basis.

rosdep/base.yaml
```diff
@@ -1290,12 +1294,12 @@ libcap-dev:
   arch: [libcap]
   debian: [libcap-dev]
   fedora: [libcap-devel]
-  gentoo: [sys-libs/libcap]
+  gentoo: ['=sys-libs/libcap-2*']
```
allenh1: Why is this being restricted to version 2?

rosdep/base.yaml
```diff
@@ -1317,20 +1321,20 @@ libcoin60-dev:
   arch: [coin]
   debian: [libcoin60-dev]
   fedora: [Coin2-devel]
-  gentoo: [media-libs/coin]
+  gentoo: ['=media-libs/coin-3.1.3*']
```
allenh1: Also not sure about this version number -- how did you get from coin60 to coin-3.1.3?

rosdep/base.yaml
```diff
@@ -1317,20 +1321,20 @@ libcoin60-dev:
...
   ubuntu: [libcoin60-dev]
 libcoin80-dev:
   debian: [libcoin80-dev]
   fedora: [Coin2-devel]
   ubuntu: [libcoin80-dev]
 libconfig++-dev:
   debian: [libconfig++-dev]
-  gentoo: [dev-libs/libconfig]
+  gentoo: ['dev-libs/libconfig[cxx]']
```
allenh1: I don't think this `USE` flag is needed.

rosdep/base.yaml
```diff
@@ -3916,7 +3921,6 @@ swi-prolog-http:
   arch: [swi-prolog]
   debian: [swi-prolog-http]
   fedora: [pl]
-  gentoo: [dev-lang/swi-prolog]
```
allenh1: why is this being removed?

rosdep/base.yaml
```diff
@@ -3903,7 +3909,6 @@ swi-prolog-clib:
   arch: [swi-prolog]
   debian: [swi-prolog-clib]
   fedora: [pl-devel]
-  gentoo: [dev-lang/swi-prolog]
```
allenh1: Why is this being removed?

rosdep/base.yaml
Alessandro-Barbieri: https://packages.debian.org/wheezy/libcoin60-dev

rosdep/base.yaml
Alessandro-Barbieri: What means static?

rosdep/base.yaml
Alessandro-Barbieri: Because is the -dev version of libcap2

rosdep/base.yaml
Alessandro-Barbieri: Can't find an useflag that enables http

rosdep/base.yaml
Alessandro-Barbieri: Can't find an useflag that enables clib

rosdep/base.yaml
allenh1: Thank you!

rosdep/base.yaml
allenh1: I think the rule here should be only to specify a version in Gentoo if it's explicitly specified for the Ubuntu version of the package. Otherwise I don't think we should specify.

rosdep/base.yaml
allenh1: But then wouldn't it just be a part of the swi-prolog package?

rosdep/base.yaml
allenh1: I don't think this is reason to remove it, though. I think that just means that it's enabled by default.

rosdep/base.yaml
Alessandro-Barbieri: We must be coherent. In debian -dev lose the version number
You can see here https://packages.debian.org/sid/libcap-dev came from the same source as libcap2

rosdep/base.yaml
Alessandro-Barbieri: `swi-prolog-clib` does not exist in debian

rosdep/base.yaml
Alessandro-Barbieri: `swi-prolog-http` does not exist in debian

rosdep/base.yaml
allenh1: Ok, that's all I needed. Thanks.

rosdep/base.yaml
allenh1: > swi-prolog-http does not exist in debian

That's not good. We should figure out what it's supposed to be and push that in a separate merge request. First we should figure out if any packages are even using this dep anymore, though. It's possible this dep is a very outdated one, in which case we shouldn't worry about it.

allenh1:
Ok, so we had some internal discussion on this and we decided that the versioning and use flags should be done as follows.

* don't specify a `USE` flag that is on by default (npm, for example, is on by default, so we shouldn't enable it here)
* don't specify a version or a slot number unless the rosdep key is also versioned (or it breaks things).

Alessandro-Barbieri:
Some users put `*/* -*` in the first line of `package.use`, we have to make sure to put the appropriate dependencies in this case too.

Alessandro-Barbieri:
Some users put `*/* -*` in the first line of `package.use`, we have to make sure to put the appropriate dependencies in this case too.

rosdep/base.yaml
leio: The bracket syntax would work in generated ebuilds (e.g automatically generated meta ebuilds or whatnot) if they are generated as EAPI-2 or newer (preferably EAPI-6 right now), but not when passed straight to emerge

rosdep/base.yaml
Alessandro-Barbieri: found in `lenny` `woody` and `lucid`

Alessandro-Barbieri:
@leio do you know if `sys-apps/lm_sensors:0` will remain version 4?

leio:
No, I'd say that depends on upstream. Maybe version 5 will just be catching on the version number fever from browsers. Or maybe it breaks everything but isn't parallel installable with current SLOT=0 and so will be a subslot change. Or maybe it breaks too much for ROS purposes even before.

rosdep/base.yaml
allenh1: In this sense, I believe that means that it is statically linked.

Alessandro-Barbieri: This entry needs a total refactoring:

1. The name is wrong, there is no gcc-static in fedora, only glibc-static
2. remove all other distros other than fedora

rosdep/base.yaml
```diff
@@ -819,7 +819,7 @@ gfortran:
   arch: [gcc-fortran]
   debian: [gfortran]
   fedora: [gcc-fortran]
-  gentoo: [sys-devel/gcc]
+  gentoo: ['sys-devel/gcc[fortran]']
```
allenh1: Is this flag enabled by default?

rosdep/base.yaml
```diff
@@ -976,11 +978,13 @@ gstreamer1.0-plugins-ugly:
 gstreamer1.0-tools:
   debian: [gstreamer1.0-tools]
   fedora: [gstreamer1]
+  gentoo: ['media-libs/gstreamer:1.0']
   ubuntu: [gstreamer1.0-tools]
 gstreamer1.0-x:
   arch: [gstreamer]
   debian: [gstreamer1.0-x]
   fedora: [gstreamer1-plugins-base]
+  gentoo: ['media-libs/gst-plugins-base:1.0[X,pango]']
```
allenh1: Should this just be `gstreamer`?

rosdep/base.yaml
```diff
@@ -1470,7 +1475,7 @@ libfltk-dev:
     wheezy: [libfltk1.1-dev]
   fedora: [fltk-devel]
   freebsd: [fltk]
-  gentoo: [=x11-libs/fltk-1*]
+  gentoo: [X11-libs/fltk]
```
allenh1: Is this a typo?

rosdep/base.yaml
```diff
@@ -2090,7 +2096,7 @@ libosmesa6-dev:
   arch: [mesa]
   debian: [libosmesa6-dev]
   fedora: [mesa-libOSMesa-devel]
-  gentoo: [media-libs/mesa]
+  gentoo: ['media-libs/mesa[osmesa]']
```
allenh1: Is this use flag enabled by default?

rosdep/base.yaml
```diff
@@ -2891,7 +2897,7 @@ libvtk-qt:
     stretch: [libvtk6-qt-dev]
     wheezy: [libvtk5-qt4-dev]
   fedora: [vtk-qt]
-  gentoo: [sci-libs/vtk]
+  gentoo: ['sci-libs/vtk[qt4]'] #qt5 is available too
```
allenh1: I don't think we should put a `USE` flag here, should the user have enabled/disabled `qt4` or `qt5`.

rosdep/base.yaml
```diff
@@ -3308,6 +3314,7 @@ nodejs-legacy:
 npm:
   debian: [npm]
   fedora: [npm]
+  gentoo: ['net-libs/nodejs[npm]']
```
allenh1: Is this on by default?

rosdep/base.yaml
Alessandro-Barbieri: yes

rosdep/base.yaml
Alessandro-Barbieri: nope

rosdep/base.yaml
Alessandro-Barbieri: no, see https://packages.debian.org/sid/gstreamer1.0-x

rosdep/base.yaml
Alessandro-Barbieri: the entry is `libvtk-qt`

rosdep/base.yaml
allenh1: So, for a user that doesn't do `*/* -*`, if they use one of the desktop profiles, this flag would not be enabled?

rosdep/base.yaml
allenh1: Excellent. Thanks!

dirk-thomas:
@tfoote Please review the current state of the PR. I don't know enough about Gentoo to make a decision here.

dirk-thomas:
@Alessandro-Barbieri This needs to be rebased to be mergable.

dirk-thomas:
@tfoote ping

rosdep/base.yaml
```diff
@@ -2649,7 +2655,7 @@ libsensors4-dev:
   arch: [lm_sensors]
   debian: [libsensors4-dev]
   fedora: [lm_sensors-devel]
-  gentoo: [sys-apps/lm_sensors]
+  gentoo: ['sys-apps/lm_sensors:0/4.4.0']
```
allenh1: can the sub-version be omitted from the slot here?

rosdep/base.yaml
allenh1: Can an or operator resolve this?

tfoote (approved):

rosdep/base.yaml
Alessandro-Barbieri: no, the rosdep key is ambiguous because doesn't specify the version, is better to fix the rosdep key and leaving qt4 for now

rosdep/base.yaml
Alessandro-Barbieri: yes

rosdep/base.yaml
Alessandro-Barbieri: no because is not default and noone is telling the user to install

rosdep/base.yaml
Alessandro-Barbieri: I'll remove this when will be removed from gentoo

rosdep/base.yaml
allenh1: Ya, I suppose that's the best idea (since wheezy above targets qt4).

allenh1:
lgtm

tfoote (approved):

PR #15222, opened 2017-06-13
useflags for gentoo (mainly python)

Alessandro-Barbieri:


tfoote:
Are these use flags different than the defaults for these packages?

Alessandro-Barbieri:
I don't know, if we need these useflags we need it. Defaults can change (users can remove all defaults from all packages, often done in minimal builds) and being more specific will not break anything.

tfoote:
In general we see to define the targets as minimally as possible to decrease maintenance burden down the line for maintaining the rules. Since any time the available packages changes the rosdep rules must be changed to match. And we don't have any way of tracking that at the moment.

There's not much analogy for use flags in the other distros. Typically what's in a package is what's in a package. And users cannot set flags to enable or disable parts. Supporting the use case of a user who explicitly changes defaults is hard. My analogy for that would be that the user added an Ubuntu PPA with a custom version of a package in our database.

I'm not super familiar with how gentoo overlays these rules. If they've turned off all defaults this will still win? Similarly can use flags conflict?

allenh1:
> If they've turned off all defaults this will still win? 

Because of the way the ebuilds are generated, yes. The issue is that portage would _just_ use the `USE` flags that are specified.

> Similarly can use flags conflict?

Yes, they can definitely conflict (`qt4` and `qt5` generally conflict, for example).

dirk-thomas:
@tfoote Please review the current state of the PR. I don't know enough about Gentoo to make a decision here.

dirk-thomas:
@tfoote ping

tfoote:
Since the targets work as they currently stand with the default use tags. I'd recommend closing this instead of merging. Although these definitions are more precise they will increase our burden of maintenance and take away the delegation of judgement from the upstream maintainers.

Alessandro-Barbieri:
These useflags are not default and ae needed in two ways:

1. to get the right feature enabled on install time when you install ros manually
2. in the ebuilds to install the packages with the right features and keep them installed

tfoote (approved):
lgtm since these are not default.

allenh1 (approved):
+1

PR #15231, opened 2017-06-14
Adding a repository for an Epson G364 IMU driver

jfkeller:
I would like to add a ROS package that is a driver for the Epson G364 IMU.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2889,16 +2889,16 @@ repositories:
   epson_g364_imu_driver:
     doc:
       type: git
-      url: https://kellerj2@bitbucket.org/castacks/epson_g364_imu_driver.git
+      url: https://bitbucket.org/castacks/epson_g364_imu_driver.git
       version: indigo-devel
     release:
       tags:
         release: release/indigo/{package}/{version}
-      url: https://kellerj2@bitbucket.org/castacks/epson_g364_imu_driver.git
+      url: https://bitbucket.org/castacks/epson_g364_imu_driver.git
       version: 1.0.0
     source:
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -2886,6 +2886,21 @@ repositories:
       url: https://github.com/RIVeR-Lab/epos_hardware.git
       version: indigo-devel
     status: maintained
+  epson_g364_imu_driver:
+    doc:
+      type: git
+      url: git@bitbucket.org:castacks/epson_g364_imu_driver.git
```
dirk-thomas: You must use https urls otherwise anonymous users can't fetch the ssh urls.

indigo/distribution.yaml
```diff
@@ -2886,6 +2886,21 @@ repositories:
       url: https://github.com/RIVeR-Lab/epos_hardware.git
       version: indigo-devel
     status: maintained
+  epson_g364_imu_driver:
+    doc:
+      type: git
+      url: https://kellerj2@bitbucket.org/castacks/epson_g364_imu_driver.git
```
dirk-thomas: You need to remove your user name from the URL.

tfoote:
You are indicating that you want it to document and index the `indigo-devel` branch however your repo only has a [master branch](https://bitbucket.org/castacks/epson_g364_imu_driver/branches/) 

And the release section should be a release repository, it should not be the same as the source repository. Typically this is generated by bloom, otherwise you should leave it out. http://wiki.ros.org/bloom

dirk-thomas:
Closing due to no response.

PR #15238, opened 2017-06-14
rosdep: add polyclipping and python-{genshi, tinydb, xmltodict} deps

CodeFinder2:
I would like to add the following package dependencies:
- polyclipping: https://packages.ubuntu.com/xenial/libpolyclipping-dev
- python-genshi: https://packages.ubuntu.com/xenial/python-genshi
- python-xmltodict: https://packages.ubuntu.com/xenial/python-xmltodict
- python-tinydb: https://pypi.python.org/pypi/tinydb

I am using these packages for working on paths polygon as well as XML processing in Python.

Please note that I've followed the guidelines [here](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md) and [here](http://docs.ros.org/independent/api/rosdep/html/contributing_rules.html), and tested my changes accordingly.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3463,10 +3463,12 @@ pocketsphinx:
     xenial: [pocketsphinx]
     yakkety: [pocketsphinx]
     zesty: [pocketsphinx]
-polyclipping:
-  ubuntu:
-    trusty: [libpolyclipping-dev]
-    xenial: [libpolyclipping-dev]
+polyclipping-dev:
+  debian:
+    jessie: [libpolyclipping-dev]
+    stretch: [libpolyclipping-dev]
+    sid: [libpolyclipping-dev]
+  ubuntu: [libpolyclipping-dev]
 portaudio:
   arch: [portaudio]
   debian: [libportaudio-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3463,6 +3463,10 @@ pocketsphinx:
     xenial: [pocketsphinx]
     yakkety: [pocketsphinx]
     zesty: [pocketsphinx]
+polyclipping:
+  ubuntu:
+    trusty: [libpolyclipping-dev]
+    xenial: [libpolyclipping-dev]
 portaudio:
   arch: [portaudio]
   debian: [libportaudio-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3467,7 +3467,6 @@ polyclipping-dev:
   debian:
     jessie: [libpolyclipping-dev]
     stretch: [libpolyclipping-dev]
-    sid: [libpolyclipping-dev]
   ubuntu: [libpolyclipping-dev]
 portaudio:
   arch: [portaudio]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -965,6 +965,10 @@ python-gdown-pip:
     pip:
       packages: [gdown]
 python-genshi:
+  debian:
+    wheezy: [python-genshi]
+    jessie: [python-genshi]
+    stretch: [python-genshi]
   ubuntu: [python-genshi]
 python-geographiclib:
   debian: [python-geographiclib]
@@ -3353,7 +3357,6 @@ python-xmltodict:
   debian:
     jessie: [python-xmltodict]
     stretch: [python-xmltodict]
-    sid: [python-xmltodict]
   ubuntu:
     xenial: [python-xmltodict]
     yakkety: [python-xmltodict]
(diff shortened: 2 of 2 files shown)


rosdep/python.yaml
```diff
@@ -3057,6 +3061,13 @@ python-tilestache:
     heisenbug: [python-tilestache]
     schrödinger’s: [python-tilestache]
   ubuntu: [tilestache]
+python-tinydb:
```
mikaelarguedas: please use `-pip` suffix for packages only available via pip

rosdep/python.yaml
```diff
@@ -3340,6 +3351,9 @@ python-xlib:
   fedora: [python-xlib]
   gentoo: [dev-python/python-xlib]
   ubuntu: [python-xlib]
+python-xmltodict:
+  ubuntu:
+    xenial: [python-xmltodict]
```
mikaelarguedas: This package is available on all ubuntu platform starting with xenial, please add them all. We also usually add the debian key as well

rosdep/python.yaml
```diff
@@ -964,6 +964,10 @@ python-gdown-pip:
   ubuntu:
     pip:
       packages: [gdown]
+python-genshi:
+  ubuntu:
+    trusty: [python-genshi]
+    xenial: [python-genshi]
```
mikaelarguedas: python-genshi is available on every not EOL ubuntu, please make the entry distribution agnostic:
`ubuntu: [python-genshi]`

rosdep/base.yaml
```diff
@@ -3463,6 +3463,10 @@ pocketsphinx:
     xenial: [pocketsphinx]
     yakkety: [pocketsphinx]
     zesty: [pocketsphinx]
+polyclipping:
```
mikaelarguedas: please use `-dev` in the name to avoid confusion with the non dev version `libpolyclipping`

rosdep/base.yaml
```diff
@@ -3463,6 +3463,10 @@ pocketsphinx:
...
+  ubuntu:
+    trusty: [libpolyclipping-dev]
+    xenial: [libpolyclipping-dev]
```
mikaelarguedas: This package is available on all alive ubuntu distribution, please majke this entry distribution agnostic `ubuntu: [libpolyclipping-dev]`.
Also please add the corresponding debian entry

mikaelarguedas (changes requested):

CodeFinder2:
Thanks for your super quick review, @mikaelarguedas! Hopefully, I've addressed all of your comments. ;-)

rosdep/base.yaml
```diff
@@ -3463,6 +3463,12 @@ pocketsphinx:
     xenial: [pocketsphinx]
     yakkety: [pocketsphinx]
     zesty: [pocketsphinx]
+polyclipping-dev:
+  debian:
+    jessie: [libpolyclipping-dev]
+    stretch: [libpolyclipping-dev]
+    sid: [libpolyclipping-dev]
```
mikaelarguedas: We usually don't target rolling distros so you don't have to add sid to this list (same below)

rosdep/python.yaml
```diff
@@ -964,6 +964,8 @@ python-gdown-pip:
   ubuntu:
     pip:
       packages: [gdown]
+python-genshi:
+  ubuntu: [python-genshi]
```
mikaelarguedas: please add a debian entry here as well

mikaelarguedas (commented):
Thanks for the quick iteration, I have a few more comments. The test failure was because of an alphabetical order error on sid so removing the sid entries as per review should also fix the test

CodeFinder2:
All (new) comments addressed in my recent commit. ;-)

(Please note that there are more "sid" entries for Debian. That was the reason why I've added mine.)

mikaelarguedas:
> All (new) comments addressed in my recent commit. ;-)

Thanks :+1: 

> (Please note that there are more "sid" entries for Debian. That was the reason why I've added mine.)

Yeah I know, I've been hesitant to [erform a cleanup and remove them all in case anybody is actually relying on the existing ones. Thanks for being thorough though!

rosdep/python.yaml
```diff
@@ -964,6 +964,12 @@ python-gdown-pip:
   ubuntu:
     pip:
       packages: [gdown]
+python-genshi:
+  debian:
```
mikaelarguedas: these are all the alive debian distros so these rules can be collapsed to: `debian: [python-genshi]` (like you did for ubuntu)

CodeFinder2:
Your last comment should be fixed now as well. ;-) 
Thanks for your detailed comments and for being so patient.  I totally understand that it's necessary to be meticulously in this regard.

Please let me know if there are any other changes required for merging my PR.

mikaelarguedas (approved):

mikaelarguedas:
I made a small fixup and it passes the tests now. Thanks @CodeFinder2 for iterating on this

PR #15239, opened 2017-06-14
Add rosdep keys for some ROS 2 dependencies.

nuclearsandwich:
These keys were added during the packaging of ROS 2 in advance of the beta 2 release. I don't believe any introduce conflicts with existing rosdep keys.

Right now keys have only been added for Ubuntu Xenial since that's the only distro supported by the ROS 2 binaries and I didn't want to add keys that I didn't test for satisfaction.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -351,6 +351,9 @@ checkinstall:
 chromium-browser:
   gentoo: [www-client/chromium]
   ubuntu: [chromium-browser]
+clang-format:
+  ubuntu:
+    xenial: [clang-format]
 cmake:
   arch: [cmake]
   debian: [cmake]
@@ -425,6 +428,9 @@ couchdb:
   fedora: [couchdb]
   gentoo: [dev-db/couchdb]
   ubuntu: [couchdb]
+cppcheck:
+  ubuntu:
+    xenial: [cppcheck]
 cppunit:
   arch: [cppunit]
   debian: [libcppunit-dev]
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please see the failing tests which mention how to update the patch.

rosdep/base.yaml
```diff
@@ -3552,6 +3558,12 @@ pstoedit:
   fedora: [pstoedit]
   gentoo: [media-gfx/pstoedit]
   ubuntu: [pstoedit]
+pydocstyle:
+  ubuntu:
+    xenial: [pydocstyle]
+pyflakes3:
+  ubuntu:
+    xenial: [pyflakes3]
```
dirk-thomas: These Python packages should be in the `python.yaml` file.

nuclearsandwich:
If we do want to expand some of these keys from xenial to any ubuntu, what's the oldest distro the package should be in? Trusty? And none of the ROS 2 dependencies had explicit version constraints provided, but clearly there are some unspecified ones (python3-flake8 came up in offline discussion).

rosdep/base.yaml
```diff
@@ -3552,6 +3558,12 @@ pstoedit:
   fedora: [pstoedit]
   gentoo: [media-gfx/pstoedit]
   ubuntu: [pstoedit]
+pydocstyle:
+  ubuntu:
+    xenial: [pydocstyle]
```
nuclearsandwich: according to rmadison, pydocstyle is present only in xenial+

nuclearsandwich:
@dirk-thomas @mikaelarguedas ready for another :eyes:.

mikaelarguedas:
> If we do want to expand some of these keys from xenial to any ubuntu, what's the oldest distro the package should be in? Trusty?

The buildfarm doesnt build any packages for ubuntu distros older that trusty. So it's safe to use the collapsed `ubuntu: [pkg_name]` for any package available on trusty+ IMO

tfoote (approved):
+1 for now. we'll want to fill it in in the future on other platforms etc.

mikaelarguedas:
Agreed we should revisit soon to at least add Ubuntu Y, Z, A and Debian rules for these new keys

PR #15249, opened 2017-06-15
Add rosdep for adafruit-pca9685

togaen:
This is to add a rosdep key for the Adafruit-PCA9685 raspberry pi controller. Source for the controller is here:

https://github.com/adafruit/Adafruit_Python_PCA9685

PIP package is here:

https://pypi.python.org/pypi/Adafruit-PCA9685/1.0.1


mikaelarguedas (commented):
Can you please use the suffix `-pip` for the key if the package is only available through pip? Thanks!

mikaelarguedas (approved):
lgtm

PR #15257, opened 2017-06-16
remove duplicates keys

k-okada:
c.f. https://github.com/ros-infrastructure/rosdep/pull/520

some of rosdep keys is multiply defined, and that sometime confuse users.

- rename xdot in python.yaml to python-xdot, because xdot is already taken by  ros package, at least until indigo http://wiki.ros.org/xdot

change xdot to python-xdot, this will break backward compatibility....

-  pocketsphinx is already taken by ros  package(http://wiki.ros.org/pocketsphinx) 
    and that different from pocketsphinx system package,
    libshphinxbase is library and pocketsphinx installs some application binaries, so they are different too

  this changes pocketsphinx key in base.yaml to pocketsphinx-bin, this also breaks compatibility, but that was added just a few month ago

   ```

ubuntu-14:~$ dpkg -L libsphinxbase1
/.
/usr
/usr/share
/usr/share/doc
/usr/share/doc/libsphinxbase1
/usr/share/doc/libsphinxbase1/copyright
/usr/share/doc/libsphinxbase1/changelog.Debian.gz
/usr/lib
/usr/lib/libsphinxbase.so.1.1.1
/usr/lib/libsphinxad.so.0.0.1
/usr/lib/libsphinxbase.so.1
/usr/lib/libsphinxad.so.0
ubuntu-14:~$ dpkg -L libpocketsphinx1/.
/usr
/usr/lib
/usr/lib/libpocketsphinx.so.1.1.0
/usr/share
/usr/share/doc
/usr/share/doc/libpocketsphinx1
/usr/share/doc/libpocketsphinx1/copyright
/usr/share/doc/libpocketsphinx1/changelog.Debian.gz
/usr/lib/libpocketsphinx.so.1


ubuntu-16:~$ dpkg -L pocketsphinx
/.
/usr
/usr/bin
/usr/bin/pocketsphinx_mdef_convert
/usr/bin/pocketsphinx_continuous
/usr/bin/pocketsphinx_batch
/usr/share
/usr/share/doc
/usr/share/doc/pocketsphinx
/usr/share/doc/pocketsphinx/copyright
/usr/share/man
/usr/share/man/man1
/usr/share/man/man1/pocketsphinx_batch.1.gz
/usr/share/man/man1/pocketsphinx_continuous.1.gz
/usr/share/man/man1/pocketsphinx_mdef_convert.1.gz
/usr/share/doc/pocketsphinx/changelog.Debian.gz

ubuntu-16:~$ dpkg -L libpocketsphinx3
[...]

PR #15258, opened 2017-06-16
remove python-* in base.yaml, which already defined in python.yaml

k-okada:
rewrited version of https://github.com/ros/rosdistro/pull/15257

some of rosdep keys is multiply defined, and that sometime confuse users.

- This just to remove duplicates entry from base.yaml, no regressions.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3568,48 +3568,14 @@ pyqt5-dev-tools:
   fedora: [python-qt5-devel]
   gentoo: [dev-python/PyQt5]
   ubuntu: [pyqt5-dev-tools]
-python-cairo:
-  fedora: [pycairo]
-  gentoo: [dev-python/pycairo]
-python-nose:
-  fedora: [python-nose]
-  gentoo: [dev-python/nose]
-python-opencv:
-  fedora: [opencv-python]
-  gentoo: [media-libs/opencv]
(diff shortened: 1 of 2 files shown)


tfoote (approved):

ipa-mdl:
To avoid further confusion, what about moving all `python-*` keys into python.yaml by policy?

rosdep/base.yaml
```diff
@@ -3568,48 +3568,14 @@ pyqt5-dev-tools:
...
-  opensuse: [python-pygraphviz]
 python-qrencode:
   debian: [python-qrencode]
   gentoo: [media-gfx/qrencode-python]
   ubuntu: [python-qrencode]
-python-sphinx:
-  fedora: [python-sphinx]
-  gentoo: [dev-python/sphinx]
-python-tornado:
-  arch: [python-tornado]
```
mikaelarguedas: Please add the arch key to the python.yaml file

mikaelarguedas (changes requested):
One rule is removed here and doesn't exist in the python.yaml file.

rosdep/base.yaml
k-okada: thanks for catch! done

mikaelarguedas (approved):
lgtm, thanks for iterating

PR #15262, opened 2017-06-16
Add libxml2-utils key for debian and ubuntu

jlack1987:


tfoote:
This looks ok. Is there a fedora equivalent?

jlack1987:
I think that's a different package. That's for `libxml2` . A key already exists for that. The key added in this PR is `libxml2-utils`. Quick google didn't find a fedora package for `libxml2-utils`.

mikaelarguedas (approved):
lgtm, couldnt find a fedora equivalent after a quick google search

PR #15272, opened 2017-06-19
Add julius-voxforge key

furushchev:
This package provides acoustic model for speech recognition software Julius.
Also found similar package on Arch linux but could not found on others.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1144,6 +1144,7 @@ joystick:
   ubuntu: [joystick]
 julius-voxforge:
   arch: [voxforge-am-julius]
+  fedora: [julius-voxforge]
   ubuntu: [julius-voxforge]
 jython:
   arch: [jython]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1142,6 +1142,9 @@ joystick:
   opensuse: [input-utils]
   rhel: [joystick]
   ubuntu: [joystick]
+julius-voxforge:
+  arch: [voxforge-am-julius]
+  ubuntu: [julius-voxforge]
 jython:
   arch: [jython]
   debian: [jython]


mikaelarguedas (approved):
lgtm: didn't find a debian equivalent either after a quick googling

k-okada:
I think we need empty entry for debian to release using bloom
```
debian: []
```

furushchev:
@k-okada I see.
BTW, I found many entries without for debian distro, why they are ok without it?

tfoote:
You should not add an empty rule, that says that there's nothing required to satisfy the dependency. You will have to skip the debian generators if the dependency is not resolvable.

k-okada:
ok, so how can we skip debian generators? 
```
$ bloom-release --help
usage: bloom-release [-h] [--list-tracks] --track TRACK [--non-interactive]
                     --ros-distro ROS_DISTRO [--new-track] [--pretend]
                     [--no-web] [--pull-request-only]
                     [--override-release-repository-url OVERRIDE_RELEASE_REPOSITORY_URL]
                     [--override-release-repository-push-url OVERRIDE_RELEASE_REPOSITORY_PUSH_URL]
                     [-d] [--version] [--unsafe]
                     repository

Releases a repository which already exists in the ROS distro file.

positional arguments:
  repository            repository to run bloom on

optional arguments:
  -h, --help            show this help message and exit
  --list-tracks, -l     list available tracks for repository
  --track TRACK, -t TRACK
                        track to run
  --non-interactive, -y
  --ros-distro ROS_DISTRO, --rosdistro ROS_DISTRO, -r ROS_DISTRO
                        determines the ROS distro file used
  --new-track, --edit-track, -n, -e
                        if used, a new track will be created before running
                        bloom
  --pretend, -s         Pretends to push and release
  --no-web              prevents a web browser from being opened at the end
  --pull-request-only, -p
                        skips the release actions and only tries to open a
                        pull request
  --override-release-repository-url OVERRIDE_RELEASE_REPOSITORY_URL
                        override the release repository url; the 'Release
                        Repository Push URL' track configuration is ignored
  --override-release-repository-push-url OVERRIDE_RELEASE_REPOSITORY_PUSH_URL
                        override the 'Release Repository Push URL' track
                        configuration; can be used in conjunction with
                        --override-release-repository-url

global:
  -d, --debug           enable debug messages
  --version             prints the bloom version
  --unsafe              Makes bloom faster, but if there is an error then you
                        could run into trouble.
```

or should I remove https://github.com/tork-a/jsk_3rdparty-release/blob/master/tracks.yaml#L34-L35 manually?

Alessandro-Barbieri:
https://packages.gentoo.org/packages/app-accessibility/julius

tfoote:
@Alessandro-Barbieri since this is closed your comment is likely to get lost a new PR would be greatly appreciated. 

@k-okada that would work, I believe you can also skip it in the interactions too.

furushchev:
@Alessandro-Barbieri It looks `julius` itself which is already registered in `rosdistro`.
The package proposed in this PR is accoustic model for English used in `julius` but not a program itself.

PR #15290, opened 2017-06-20
Bump python-pathtools

yotabits:


tfoote:
@yotabits You have trailing whitespace in your commit. See the travis output for details.

Did you try to get bloom to work with the oauth tokens? It should save you a lot of headache.

yotabits:
@tfoote unfortunatelly it was not the problem, I don't know why but bloom is always trying to connect to my github account with my mail address instead of my username...

dirk-thomas:
> which file is used by bloom to find username?

I think it is `~/.config/bloom`.

PR #15299, opened 2017-06-20
Adding python-opencv for xenial ubuntu version

mikramarc:
python-opencv currently not available for xenial ubuntu version. This PR fixes the problem.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1641,6 +1641,7 @@ python-opencv:
   ubuntu:
     saucy: [python-opencv]
     trusty: [python-opencv]
+    xenial: [python-opencv]
 python-opengl:
   arch: [python2-opengl]
   debian: [python-opengl]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1638,10 +1638,7 @@ python-opencv:
   gentoo: [media-libs/opencv]
   opensuse: [python-opencv]
   slackware: [opencv]
-  ubuntu:
-    saucy: [python-opencv]
-    trusty: [python-opencv]
-    xenial: [python-opencv]
+  ubuntu: [python-opencv]
 python-opengl:
   arch: [python2-opengl]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@mikramarc ROS ships with its own opencv3 since Kinetic. According to [REP3](http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021), all ROS distributions supported on Xenial are also targeting OpenCV3.x or higher. You may want to try to port your code to OpenCV3 rather than relying on a non targeted OpenCV version.

jspricke:
@mikaelarguedas I don't see anything wrong with providing the rule, as the package is available on xenial.

Alessandro-Barbieri:
Travis is stuck since four days ...

dirk-thomas:
Retriggering Travis...

mikaelarguedas:
> I don't see anything wrong with providing the rule, as the package is available on xenial.

I agree with that, I was just raising the point than writing packages targeting a version different than the one targeted by the ROS distribution may result in unexpected results and should be considered

mikramarc:
@mikaelarguedas noted, will keep that in mind in the near future, thanks.
Still, any chance this will be merged after all any time soon?

rosdep/python.yaml
```diff
@@ -1641,6 +1641,7 @@ python-opencv:
   ubuntu:
     saucy: [python-opencv]
     trusty: [python-opencv]
+    xenial: [python-opencv]
```
mikaelarguedas: can you use the collapsed version: `ubuntu: [python-opencv]` this way we won't have to modify the key for yakkety, zesty, artful and the following distros ? (noting that starting with artful this will be opencv3 and not opencv2 bindings)

mikramarc:
@mikaelarguedas made changes according to the review

mikaelarguedas (approved):
thanks @mikramarc for the quick iteration.
LGTM once travis done and green

PR #15308, opened 2017-06-20
Depend on generic libcurl-dev

jspricke:
Don't depend on specific openssl version, as this could conflict with an
already installed gnutls version.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -447,7 +447,7 @@ crypto++:
   ubuntu: [libcrypto++-dev]
 curl:
   arch: [curl]
-  debian: [libcurl4-openssl-dev, curl]
+  debian: [libcurl-dev, curl]
   fedora: [libcurl-devel, curl]
   freebsd: [curl]
   gentoo: [net-misc/curl]
@@ -457,7 +457,7 @@ curl:
   slackware:
     slackpkg:
       packages: [curl]
-  ubuntu: [libcurl4-openssl-dev, curl]
+  ubuntu: [libcurl-dev, curl]
 curlpp-dev:
   debian: [libcurlpp-dev]
   fedora: [curlpp-devel]


mikaelarguedas:
@jspricke I couldn't find libcurl-dev on either package.debian.org or packages.ubuntu.com, could you clarify where you found those packages names ?

jspricke:
Here you go:https://packages.debian.org/stretch/libcurl-dev  https://packages.ubuntu.com/de/xenial/libcurl-dev

mikaelarguedas:
oh interesting that the search doesn't return virtual packages o_O. Thanks for the links.
I think that virtual packages are now handled by rosdep so this shouldn't be a blocker for the change. I'll let @tfoote and @dirk-thomas review this though.

dirk-thomas (dismissed):

tfoote:
It does look like this is a problem. I think we resolved the virtual package checks but it doesn't appear that we've resolved the install side issues: 

This is what rosdep would have run:
```
tfoote@snowman:~ Last: [0] (1s Seconds)
$ rosdep install -s resource_retriever --reinstall
#[apt] Installation commands:
  sudo -H apt-get install python-rospkg
  sudo -H apt-get install ros-kinetic-rosconsole
  sudo -H apt-get install libcurl4-openssl-dev
  sudo -H apt-get install curl
  sudo -H apt-get install ros-kinetic-roslib
  sudo -H apt-get install ros-kinetic-catkin
```
I tweaked it to use the new rule and it failed to install. 
```
tfoote@snowman:~ Last: [0] (0s Seconds)
$ sudo -H apt-get install libcurl-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package libcurl-dev is a virtual package provided by:
  libcurl4-openssl-dev 7.47.0-1ubuntu2.2
  libcurl4-nss-dev 7.47.0-1ubuntu2.2
  libcurl4-gnutls-dev 7.47.0-1ubuntu2.2
You should explicitly select one to install.

E: Package 'libcurl-dev' has no installation candidate
```

We've previously tried the same: #57 but had to revert #58

It looks like we haven't resolved being able to install.

I think https://github.com/ros-infrastructure/rosdep/issues/274 needs to be reopened wrt to installing.

ipa-mdl:
I have implemented support for this in https://github.com/ros-infrastructure/rosdep/pull/521 and tested it with this PR: https://gist.github.com/ipa-mdl/401f74bff9bfe64fd359c4593c4e3797

dirk-thomas:

ipa-mdl:
Works with https://github.com/ros-infrastructure/rosdep/pull/585

ipa-mdl:
rosdep should now handle this properly.

dirk-thomas:
@tfoote Cas you please give this another try with the `rosdep` problem resolved.

tfoote (approved):
With the new rosdep this appears to be working correctly. 

https://gist.github.com/tfoote/e84c8d7c3f34515df9fa75433d34a201

dirk-thomas:
Could this change be the reason for the following job starting to fail? http://build.ros.org/view/Mdev/job/Mdev__rc_visard__ubuntu_bionic_amd64/4/console

tfoote:
Yes, that logic block is going to need to become aware of the virtual package logic as it looks to be implementing it in parallel. 
A package from that repo is actually the test case used in https://github.com/ros-infrastructure/rosdep/pull/585/files

dirk-thomas:
What is the status on this?

dirk-thomas:
I went ahead and reverted the change in #18223.

PR #15185, opened 2017-06-07
Revert "Updating distribution.yaml"

tfoote:
Reverts ros/rosdistro#15180

This is failing to build on all platforms. I'm reverting it.

Also @benjayah you are not listed as the maintainer, @martinezjavier is listed as the maintainer and he just pinged me about why he's getting emails. Did you talk to him before forking and releasing the package?

PR #15311, opened 2017-06-20
rodi_robot: 0.0.1-0 in 'kinetic/distribution.yaml'

benjayah:
- Fixed previous problems with the package #15185. It now passes pre-release test.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6040,6 +6040,21 @@ repositories:
       url: https://github.com/robotics-in-concert/rocon_tools.git
       version: release/0.3-kinetic
     status: developed
+  rodi_robot:
+    doc:
+      type: git
+      url: https://github.com/rodibot/rodi_robot.git
+      version: master
+    release:
+      tags:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -6040,6 +6040,24 @@ repositories:
       url: https://github.com/robotics-in-concert/rocon_tools.git
       version: release/0.3-kinetic
     status: developed
+    status: maintained
```
dirk-thomas: This line should not be added.

kinetic/distribution.yaml
```diff
@@ -6040,6 +6040,24 @@ repositories:
...
+  rodi_robot:
+    doc:
+      type: git
+      url: https://github.com/rodibot/rodi_robot.git
+      version: master
+    release:
+      packages:
+      - rodi_robot
```
dirk-thomas: Single package repos with the same name must not enumerate the packages names.

benjayah:
Fixed

PR #15366, opened 2017-06-26
add python-vlc

floweisshardt:
see https://pypi.python.org/pypi/python-vlc/1.1.2


mikaelarguedas:
@ipa-fmw Can you please add the suffix `-pip` to the key so that it's explicit that this dependency is only available via pip ? thanks!

ipa-fmw:
@mikaelarguedas: done.

mikaelarguedas (approved):
@ipa-fmw thanks for the quick iteration.
LGTM once travis is done and green

PR #15367, opened 2017-06-26
Adding pcdfilter to documentation index for indigo/jade/kinetic/lunar

peterweissig:


dirk-thomas:
Please see the error message in the Travis log on how to update your patch to pass. You can also run the tests locally.

PR #15371, opened 2017-06-26
Ds4drv rosdep

cratliff:
Adding a ds4drv rule to rosdep.  The wheel rule is included in this pull request as ds4drv does not install without wheel being installed too.  
I'm using ds4drv to control a robot from a Sony controller and need to be able to install these dependencies.

ds4drv
A Sony DualShock 4 userspace driver for Linux
https://pypi.python.org/pypi/ds4drv/0.5.1

wheel
A built-package format for Python
https://pypi.python.org/pypi/wheel/0.30.0a0

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3356,6 +3356,11 @@ python-werkzeug:
   fedora: [python-werkzeug]
   gentoo: [dev-python/werkzeug]
   ubuntu: [python-werkzeug]
+python-wheel:
+  debian: [python-wheel]
+  fedora: [python-wheel]
+  gentoo: [dev-python/wheel]
+  ubuntu: [python-wheel]
 python-ws4py-pip:
   ubuntu:
     pip:
@@ -3469,11 +3474,6 @@ python3-pep8:
   ubuntu: [python3-pep8]
 python3-setuptools:
   ubuntu: [python3-setuptools]
-wheel:
-  debian: [python-wheel]
-  fedora: [python-wheel]
-  gentoo: [dev-python/wheel]
-  ubuntu: [python-wheel]
 xdot:
   debian: [xdot]
   fedora: [python-xdot]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -8,6 +8,10 @@ adafruit-pca9685-pip:
   ubuntu:
     pip:
       packages: [adafruit-pca9685]
+ds4drv:
```
tfoote: If this is pip only it should have `-pip` at the end.

rosdep/python.yaml
```diff
@@ -3465,6 +3469,10 @@ python3-pep8:
   ubuntu: [python3-pep8]
 python3-setuptools:
   ubuntu: [python3-setuptools]
+wheel:
```
tfoote: Wheel is available from native packages we prefer those than pip versions. 

https://packages.ubuntu.com/xenial/python-wheel
https://packages.debian.org/stretch/python-wheel
https://admin.fedoraproject.org/pkgdb/package/rpms/python-wheel/
https://packages.gentoo.org/packages/dev-python/wheel

tfoote (changes requested):

rosdep/python.yaml
cratliff: I changed the rule to use the packages you listed instead of pip.  Did I miss where these preferences are documented or would it be good to do so?

tfoote: Thanks. Unfortunately it's not well documented. I've added a few simple things here: https://github.com/ros/rosdistro/pull/15373

tfoote (approved):

mikaelarguedas:
What about the `ds4drv` pip entries for debian ? (and other platforms?)

cratliff:
I would be fine to add that if you want.  I noticed a number of other rules only had one OS entry and thought they might just be getting added as needed.

tfoote:
Having the full spread across platforms is less important for pip rules since they cannot be used for packaging. They only need to be available in the expected development environments. But in general they're good to have.

PR #15375, opened 2017-06-26
More updates on rosdep rules guidelines

tfoote:
From thoughts over lunch and thinking back on past reviews.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,10 +38,25 @@ It's valuable to have a record of the packages as submitted and their intended p
 Guidelines for rosdep rules
 ---------------------------
 
- * Native packages are strongly preferred (They are required for packaging and have upgrade and conflict tracking.)
- * Python packages should go in the python.yaml
- * Homebrew into homebrew.yaml
+ * Native packages are strongly preferred. (They are required for packaging and have upgrade and conflict tracking.)
+ * Please use the smallest notation possible.
+   For example if an ubuntu package is the same for all versions only write the generic ubuntu rule, don't call out the same package name for every version of ubuntu.
+   Similarly don't call out indirect dependencies or default options for packages.
+ * Python packages should go in the `python.yaml`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -39,9 +39,20 @@ Guidelines for rosdep rules
 ---------------------------
 
  * Native packages are strongly preferred (They are required for packaging and have upgrade and conflict tracking.)
- * Python packages should go in the python.yaml
- * Homebrew into homebrew.yaml
+ * Please use the smallest notation possible and if something is default or not necessary to specify, don't.
```
wjwwood: I don't think what mean is clear here. The terms `notation` and `is default` aren't well defined imo. Maybe you should include examples of each.

CONTRIBUTING.md
```diff
@@ -39,9 +39,20 @@ Guidelines for rosdep rules
...
+ * Python packages should go in the python.yaml.
```
wjwwood: nitpick: The use of `.` is very inconsistent, you added it here, but on the next line (which you also updated) it doesn't have it. Use it or don't, but being consistent is ideal.

CONTRIBUTING.md
```diff
@@ -39,9 +39,20 @@ Guidelines for rosdep rules
...
+ * Homebrew into osx-homebrew.yaml
  * If a package is only available via pip use the `-pip` extension.
+ * Supported Platforms
+   * Rules can be contributed for any platform. However to be released they must be at least cover the supported platforms in REP 3: http://www.ros.org/reps/rep-0003.html So:
+    * Ubuntu and Debian are top priority.
+    * Fedora and Gentoo have packaging jobs should be best effort.
```
wjwwood: This isn't a sentence, maybe you started to say something and changed or there is a word missing?

dirk-thomas:
@tfoote ping

tfoote:
Comments addressed and rebased

wjwwood (approved):

PR #15388, opened 2017-06-27
Updated name of python-catkin-pkg for Arch.

zootboy:
A bit of minor package renaming is happening for Arch.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -393,7 +393,7 @@ python-catkin-lint:
   fedora: [python-catkin_lint]
   ubuntu: [python-catkin-lint]
 python-catkin-pkg:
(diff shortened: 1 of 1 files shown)


tfoote:
Can you provide a reference for this change so we can see the discussion?

zootboy:
I wasn't personally involved in the change, it seems like it was done as part of making both python2 and python3 packages available. There is some discussion in the comments here:

https://aur.archlinux.org/packages/python-catkin_pkg/

dirk-thomas:
@tfoote @zootboy What is the status on this?

tfoote (approved):
Sorry I missed the comment. Thanks for the reference. lgtm

PR #15389, opened 2017-06-27
Added [python-crypto] & [python3-crypto] keys for OSs

zarecor60:
Added [python-crypto] & [python3-crypto] keys for OSs [yakkety], [artful], & [zesty]

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -613,6 +613,8 @@ python-crypto:
     pip:
       packages: [pycrypto]
   ubuntu:
+    artful: [python-crypto]
+    artful_python3: [python3-crypto]
     lucid: [python-crypto]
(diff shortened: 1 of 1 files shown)


tfoote (approved):

tfoote (changes requested):
This is out of alphabetical order, and looking at it now we're missing zesty in the middle too.

zarecor60:
Ah, didn't realize these were by alphabetical instead of release order. Updated as requested.

mikaelarguedas:
The check fail because of trailing whitespaces on the [xenial_python3](https://github.com/ros/rosdistro/pull/15389/files#diff-701c2f61a9f7ac8fc393ccc6630c8fa5R634) entry

PR #15397, opened 2017-06-28
fix wrong package name for python-vlc

floweisshardt:
did not make it to https://github.com/ros/rosdistro/pull/15366


dirk-thomas (approved):
https://pypi.python.org/pypi/python-vlc/1.1.2

mikaelarguedas:
thanks @ipa-fmw for catching this

PR #15349, opened 2017-06-24
debian dependencies

Alessandro-Barbieri:

PR #15405, opened 2017-06-28
remove old distros and sid

Alessandro-Barbieri:
removed all
```
sid
lenny
woody
squeeze
karmick
lucid
maverick
natty
beefy
spherical
```
entries


dirk-thomas:
While these distros might be EOL we have no policy to remove them from the database at the moment. Some users might still use them to build from source on these platforms.

tfoote (changes requested):
Thanks for the thought to clean up. However as @dirk-thomas mentions we know there are users who for various reasons are still using many of those platforms and we don't want to break them. And keeping the historical data isn't too much overhead for us. Of the cleanups though removing sid would be great as that's something we explicitly discourage.

If we were to clear these rules we would want to provide a backwards compatible way to keep the information available. But we haven't put much time in as the overhead of keeping them in the database isn't that high.

dirk-thomas:
@Alessandro-Barbieri Are you planning on addressing the feedback or should this be closed?

Alessandro-Barbieri:
What should I do?

mikaelarguedas:
I think that the feedback was to remove only sid but none of the EOL distros.

Alessandro-Barbieri:
sid is already gone with https://github.com/ros/rosdistro/pull/15349

dirk-thomas:
@Alessandro-Barbieri if this PR is obsolete can you please close it then.

dirk-thomas:
Closing due to no reply and not being mergable.

PR #15415, opened 2017-06-30
soem: 1.3.0-0 in 'lunar/distribution.yaml' [bloom]

smits:


dirk-thomas:
This seems to be a duplicate of #15414?

smits:
I've released in two steps, apparently both releases got into the same PR. I'll close this one.

PR #15441, opened 2017-07-05
Add python-chainercv-pip entry

knorth55:
add `python-chainercv-pip` in `distro/python.yaml`
- `chainercv`:  https://github.com/chainer/chainercv

cc. @yuyu2172

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -66,6 +66,10 @@ curl:
   osx:
     homebrew:
       packages: []
+cython:
+  osx:
+    homebrew:
+      packages: [cython]
 doxygen:
   osx:
     homebrew:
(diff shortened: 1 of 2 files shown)


dirk-thomas (approved):

mikaelarguedas:
The change looks good to me. Though the package fails to install on my ubuntu machine (16.04) because of missing dependencies

tfoote:
If there are missing dependencies those should be caught in the rosdep rule. @knorth55 do you know what the dependencies are for installing?

knorth55:
Dependencys are `cython` and `python-numpy`.
I added these dependencies and update this PR.

knorth55:
it seems that `python-numpy` is always installed in ros, I removed its dependency.

knorth55:
I made a mistake and close PR.
I reopend, sorry.

knorth55:
I checked with docker image and my osx machine and resolve the dependency problem!

mikaelarguedas (approved):
LGTM, thanks for the follow-up

PR #15460, opened 2017-07-07
added dependency for pcdfilter_pa (parameter_pa) on indigo

peterweissig:


mikaelarguedas:
@peterweissig Package dependencies are captured in the `package.xml` rather than the distribution file. The recommended way to introduce documentation dependency is to use package format2 ([schema](https://github.com/ros-infrastructure/rep/blob/7c404143461d82ac386447385b4043c53deb5570/xsd/package_format2.xsd), [REP](https://github.com/ros-infrastructure/rep/blob/7c404143461d82ac386447385b4043c53deb5570/rep-0140.rst) and [migration guide](http://docs.ros.org/jade/api/catkin/html/howto/format2/migrating_from_format_1.html) ) and to declare your dependency using the `<doc_depend>` tag

dirk-thomas:
Adding dependencies for doc jobs is valid. See [REP 141](http://www.ros.org/reps/rep-0141.html#distribution-file) for this.

The value needs to be a list though and can't be a single name. Please format your patch as in this example: https://github.com/ros/rosdistro/blob/59ee1c91559850af0a38e828b51371449ce56261/kinetic/distribution.yaml#L3990-L3991

mikaelarguedas:
> Adding dependencies for doc jobs is valid. See REP 141 for this.

Oh I look at the wrong REP, thanks for pointing it out. And please discard my previous comment then

PR #15466, opened 2017-07-11
Fix broken pyassimp regression

srsidd:
This is a follow up to comments on the [thread](https://github.com/ros-planning/moveit/issues/86). There is also a separate issue created [here](https://github.com/assimp/assimp/issues/1209). It fixes the broken pyassimp regression by installing libassimp-dev from apt and pyassimp from pip instead of directly installing pyassimp from apt.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2025,7 +2025,11 @@ python-pyassimp:
     utopic: [python-pyassimp]
     vivid: [python-pyassimp]
     wily: [python-pyassimp]
-    xenial: [python-pyassimp]
+    xenial:
+      apt:
+        packages: [libassimp-dev]
+      pip:
+        packages: [pyassimp]
     yakkety: [python-pyassimp]
     zesty: [python-pyassimp]
 python-pyaudio:


tfoote (changes requested):
Unfortunately the rules don't support having two different package managers. And even if we did have support for that (there are ways to approximate that with inter rosdep dependencies), with a pip package as a dependency the rosdep key would no longer be able to be used as a dependency of any debian package release.

jlack1987:
Yeah you'll need 2 keys to fix the issue. One pip key for pyassimp should go [here](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml) and should have a dependency listed on `libassimp-dev`(similar to what is done with the key for `python-chainercv-pip`), and then another key [here](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml) for the `libassimp-dev` key.

I think that'd do it right @tfoote?

tfoote:
That would make rosdep be able to run. But it does not resolve the inability to release based on the pip versions. And we generally do not want pip versions overlaying/overlapping apt versions. It's often up to date once, but once the apt version gets fixed. The pip version will continue to overlay. As this is a confirmed bug upstream on launchpad. I'd recommend pushing hard there for a backport/fix as the long term proper way to fix this. In the mean time you can give developers instructions to work around the issue. If we deploy this via rosdep a lot of people will deploy the workaround without knowing it. And there's no automated way to roll back the workaround.

jlack1987:
Unfortunately I doubt it'll get fixed in 16.04 as the developer doesn't seem very interested in fixing the issue considering it works with pip and in later versions of ubuntu. Also i'm not sure what automated work arounds exist outside of building from source yourself(to me this is not an acceptable solution). That's what has been recommended [for moveit users](https://github.com/ros-planning/moveit/issues/86).

srsidd:
Closing, recommended solution is push for a backport\fix from the original developers.

PR #15474, opened 2017-07-11
cleanup for count_packages script

tfoote:
This is useful for creating nice graphics of the history of releases. And will likely get added to the metrics reports.


scripts/count_rosdistro_packages.py
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import print_function
```
dirk-thomas: This change seems to be unnecessary.

tfoote: We don't need to keep python2 compatibility for this. I was testing it with both. Removed.

dirk-thomas: The change has nothing to do with Python 2 / 3 compatibility. The code should work fine with both versions without that line.

dirk-thomas (approved):

PR #15480, opened 2017-07-11
Add pip key for pylint and deb key for uncrustify

jlack1987:


rosdep/python.yaml
```diff
@@ -93,6 +100,13 @@ pyflakes3:
     stretch: [pyflakes3]
   ubuntu:
     xenial: [pyflakes3]
+pylint-pip:
```
tfoote: We already have python-pylint. If a native packaged version is available we generally use that following our guidelines: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

rosdep/base.yaml
```diff
@@ -4352,6 +4352,9 @@ udhcpc:
   arch: [udhcp]
   debian: [udhcpc]
   ubuntu: [udhcpc]
+uncrustify:
+  debian: [uncrustify]
```
tfoote: Please add: 

Fedora: https://apps.fedoraproject.org/packages/uncrustify
Gentoo: https://packages.gentoo.org/packages/dev-util/uncrustify

tfoote (changes requested):

rosdep/base.yaml
jlack1987: Done, thanks!

tfoote (approved):

PR #15481, opened 2017-07-12
Remove teraranger packages from Indigo

allenh1:
Removes packages `terarangerduo` and `terarangerone` from ROS Indigo. Their release repositories no longer exist.

@Terabee fyi

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -14524,23 +14524,7 @@ repositories:
       url: https://github.com/zlt1991/tensor_field_nav.git
       version: master
     status: maintained
-  terarangerduo-ros:
-    release:
-      packages:
-      - terarangerduo
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/Terabee/terarangerduo-ros-release.git
-      version: 0.1.1-0
-    status: maintained
   terarangerone-ros:
-    release:
-      packages:
-      - terarangerone
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/Terabee/terarangerone-ros-release.git
-      version: 0.1.1-0
     source:
       type: git
       url: https://github.com/Terabee/terarangerone-ros.git


indigo/distribution.yaml
```diff
@@ -14524,28 +14524,6 @@ repositories:
...
-    release:
-      packages:
-      - terarangerone
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/Terabee/terarangerone-ros-release.git
-      version: 0.1.1-0
-    source:
-      type: git
-      url: https://github.com/Terabee/terarangerone-ros.git
```
dirk-thomas: This repo still exists. Why should the entry be removed?

mikaelarguedas:
I think that the idea was to remove only the release repo isnt it ?. The source one should stay.
Did you contact @Terabee beforehand to know if they actually plan on restoring the release repo ?

indigo/distribution.yaml
allenh1: @tfoote Asked me to file a PR to remove it.

Should I have just removed the release section?

mikaelarguedas: @tfoote to confirm what was the expected PR,  but I don't see a reason to nuke the entry completely if the source repo still exists and passed the dev jobs

tfoote: Sorry I was unclear. Yes we should just remove the release section since people can still build from source.

allenh1: Got it

allenh1:
Ok, changes pushed. Let me know if it needs further modification.

dirk-thomas (approved):

tfoote:
@Terabee it would be great to have these packages back if you restored the release repo. Once the release repo is available we can restore these entries.

pl-kabaradjian:
@tfoote @mikaelarguedas @allenh1 
Hello,
I am working in Terabee.
Thank you for reaching out to us and sorry for the delay.
To clear things up, we are not planning on releasing again those repositories (they are gonna be deprecated soon also) as we are releasing new drivers packages(first on kinetic then indigo). 
So unless it is needed to leave these packages in the index for compatibility, they can be removed completely.

Best regards,
PL

PR #15482, opened 2017-07-12
Add key for python sphinx

jlack1987:


mikaelarguedas (changes requested):
python packages should go in the `python.yaml` file. Please see the [contributions guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules)

the python-sphinx rule is already defined there (see https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L2979)

mikaelarguedas:
@jlack1987 I'm going to close this given that the rule already exists [here](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L2979)

PR #15486, opened 2017-07-13
Add python-pylint xenial key

jlack1987:


rosdep/python.yaml
```diff
@@ -2169,6 +2169,7 @@ python-pylint:
     raring: [pylint]
     saucy: [pylint]
     trusty: [pylint]
+    xenial: [pylint]
```
mikaelarguedas: is it possible to add the other ubuntu distributions ? You can either add `yakkety` `zesty`, and `artful` on separate lines or collapse the ubuntu rules into a single `ubuntu:[pylint]` 

Thanks!

tfoote (approved):

mikaelarguedas:
thanks @jlack1987 for the quick iteration

PR #15285, opened 2017-06-19
robot_controllers: 0.5.3-0 in 'kinetic/distribution.yaml' [bloom]

mikeferguson:
Increasing version of package(s) in repository `robot_controllers` to `0.5.3-0`:

- upstream repository: https://github.com/fetchrobotics/robot_controllers.git
- release repository: https://github.com/fetchrobotics-gbp/robot_controllers-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.5.26`
- previous version for package: `0.5.2-0`

## robot_controllers

```
* remove limiter
* Improve cartesian twist controller (#26 <https://github.com/fetchrobotics/robot_controllers/issues/26>)
* Contributors: Hanjun Song, Michael Ferguson
```

## robot_controllers_interface

```
* add error message when pluginlib fails
* fix cmake warnings on kinetic (#28 <https://github.com/fetchrobotics/robot_controllers/issues/28>)
* Contributors: Michael Ferguson
```

## robot_controllers_msgs

- No changes

PR #15504, opened 2017-07-15
Revert "arbotix: 0.10.0-0 in 'kinetic/distribution.yaml' [bloom]"

tfoote:
Reverts ros/rosdistro#15284

Like: #15285 this has indigo dependencies in the generated metadata.


tfoote:
@mikeferguson FYI

PR #15578, opened 2017-07-24
add python-pathtools, python-watchdog on xenial

yichoe:
Only adding ubuntu for now to validate switching from trusty ros-package to kinetic xenial packages.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1869,8 +1869,12 @@ python-path.py:
     pip:
       packages: [path.py]
 python-pathtools:
+  debian: [python-pathtools]
   ubuntu:
     xenial: [python-pathtools]
+    yakkety: [python-pathtools]
+    zesty: [python-pathtools]
+    artful: [python-pathtools]
 python-pcapy:
   arch: [python2-pcapy]
   debian: [python-pcapy]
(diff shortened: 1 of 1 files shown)


asmodehn:
Any hint about why check_duplicates fail ?

We are defining the rosdistro key on kinetic here to be the same as the ros-package on indigo (and the same as the ubuntu package) on purpose, hoping to migrate from a backported ros package on indigo to a system package on kinetic. Does this cause any problem ?

Or maybe we missed something somewhere else ?

FYI @yotabits @tfoote @dirk-thomas 
Ref : https://github.com/ros/rosdistro/pull/15095

dirk-thomas:
@k-okada The new test checking for duplicates seem to not handle this case correctly. Can you please provide a PR to update the logic which works for this case.

mikaelarguedas:
@yichoe I know that this won't pass until the test it fix but is it possible to add these packages for the ubuntu distribution after xenial (yakkety, zesty and artful) as well as a generic debian entry given that these packages are available since jessie ? thanks!

k-okada:
sorry for trouble, https://github.com/ros/rosdistro/pull/15599 should be ok

dirk-thomas:
@yichoe Can you please rebase your patch to the latest master. That will hopefully make CI pass.

mikaelarguedas (approved):
lgtm thanks @yichoe for iterating

PR #15643, opened 2017-07-31
added dependency to libcgal-qt5-dev which is said to fix a bug for CG…

ipa-rmb:
…AL with kinetic and 16.04

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -344,6 +344,8 @@ cgal:
   fedora: [CGAL-devel]
   gentoo: [sci-mathematics/cgal]
   ubuntu: [libcgal-dev]
+cgal-qt5-dev:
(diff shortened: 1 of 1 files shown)


tfoote:
Can you link to the bug you're referencing? And also please look at filling in the other target platforms for this according to the CONTRIBUTING.md.

ipa-rmb:
The bug is described here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=804539 . For us, it appears when compiling our software, which depends on cgal, on Travis for Kinetic with 16.04.

ipa-rmb:
I have added the other dependencies, however, there is only one really new dependency for debian but for the other architectures, the qt part seems to be included in cgal already.

tfoote (approved):
Thanks for the extra details.

PR #15651, opened 2017-07-31
Enable python-qt-bindings-webkit rosdep key for newer Ubuntu distros

meyerj:
The package is still available in newer Ubuntu distros: https://packages.ubuntu.com/search?suite=all&arch=any&searchon=names&keywords=python-qt4

`python-qt-bindings=webkit` is required to build [rqt_web](http://wiki.ros.org/rqt_web) as part of a ROS indigo desktop distribution for Ubuntu Xenial, which actually works almost flawlessly.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2442,11 +2442,7 @@ python-qt-bindings-webkit:
   debian: [python-qt4]
   fedora: [PyQt4]
   gentoo: [dev-python/pyside, dev-python/PyQt4]
-  ubuntu:
-    saucy: [python-qt4]
-    trusty: [python-qt4]
-    utopic: [python-qt4]
-    vivid: [python-qt4]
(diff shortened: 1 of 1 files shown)


tfoote (approved):

mikaelarguedas:
Looks good to me.
@meyerj Please be aware that you may have some conflicts / unexpected behaviour given that some rosdep keys pull in Qt5 on Xenial when pulling Qt4 on previous ubuntu distributions (e.g. [libvtk-qt](https://github.com/ros/rosdistro/blob/3e0ab77128daf88c57566ec0bbe87d21d1ee2cf7/rosdep/base.yaml#L3059) and
[libqwt6](https://github.com/ros/rosdistro/blob/3e0ab77128daf88c57566ec0bbe87d21d1ee2cf7/rosdep/base.yaml#L2604))

PR #15711, opened 2017-08-06
Add dependency to python-naoqi-sdk

fkenghagho:
python-naoqi-sdk is a pure python package to write remote programs for robots' NAOqi-OS.


mikaelarguedas:
@fkenghagho this package is not an official ubuntu package but a package distributed by Softbank Robotics (forerly known as Aldebaran) so it cannot be added to the rosdep database.
@nlyubova @suryaambrose FYI

PR #15724, opened 2017-08-07
Add entry of python-pkg-resources to rosdep/python.yaml

wkentaro:
For https://github.com/ros/catkin/pull/798

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1971,6 +1971,9 @@ python-pip:
   fedora: [python-pip]
   gentoo: [dev-python/pip]
   ubuntu: [python-pip]
+python-pkg-resources:
+  debian: [python-pkg-resources]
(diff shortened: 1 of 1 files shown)


mikaelarguedas (approved):
lgtm, @dirk-thomas to confirm given the referenced PR

dirk-thomas:
I don't think this is required for the PR since `pkg-resources` is afaik always a dependency of `setuptools` but having this rosdep key is ok.

wkentaro:
> I don't think this is required for the PR since pkg-resources is afaik always a dependency of setuptools but having this rosdep key is ok.

Yeah. But should I add it as a dependency for safety?

dirk-thomas:
> should I add it as a dependency for safety?

If the code imports the package it should indeed be listed as a dependency.

PR #15760, opened 2017-08-10
Revert "navigation: 1.14.1-0 in 'kinetic/distribution.yaml' [bloom]"

tfoote:
Reverts ros/rosdistro#15730

Filed upstream at ros-planning/navigation#612

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -4734,7 +4734,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ros-gbp/navigation-release.git
-      version: 1.14.1-0
+      version: 1.14.0-0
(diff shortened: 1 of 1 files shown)


mikeferguson:
@tfoote -- I never got any build failure emails for this, but I am listed as a maintainer. Thoughts?

mikaelarguedas:
My understanding is that navigation builds fine but causes downstream
packages to fail, so if you are not listed as maintainer of any of the
failing downstream packages:
- turtlebot3_navigation
- stdr_server
- multi_map_server
you would not receive an email when they fail

mikeferguson:
@mikaelarguedas -- yeah, I moved the conversation to the navigation ticket, and then noticed and commented there after I noticed the problem is actually downstream.

PR #15767, opened 2017-08-12
point to groovy-devel branch

mikaelarguedas:
Up until lunar the master branch was used for all ros distributions.
Now that there is a `lunar-devel` branch I copied `master` on a `groovy-devel` branch to avoid confusion for users and woud like to remove the master branch.

Should I update all distribution files even for EOL rosdistros to reflect the new branch name ? or should I keep the master branch around and references to it for hydro and older distributions ?

### DIFF
--- a/groovy/distribution.yaml
+++ b/groovy/distribution.yaml
@@ -304,7 +304,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros/bond_core.git
-      version: master
+      version: groovy-devel
     release:
       packages:
       - bond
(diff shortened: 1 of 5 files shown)


tfoote:
It's a little bit of a pain for anyone with a source checkout but not the end of the world. However it should decrease future confusion. A branch named `master` that's not the tip would cause a lot of confusion, so I'd suggest it's better to switch.

mikaelarguedas:
@tfoote I added https://github.com/ros/rosdistro/pull/15767/commits/48da0a130192dcdfb7035ee905ed0b6db339f8a6 to propagate the new-branch name everywhere, which makes travis unhappy. Should we merge it anyway to allow to delete the `master` branch ?

dirk-thomas:
> Should we merge it anyway to allow to delete the master branch ?

Sure.

tfoote:
Yes, wei'll have to override the CI. it doesn't know that this is just a rename not a new release.

PR #15770, opened 2017-08-13
Update indigo/parrot_arsdk to 3.12.61-0

robohacker:
#15769 seems to breaking the source build of this package since it overwrote the release tag.

PR #15771, opened 2017-08-13
Update kinetic/parrot_arsdk to 3.12.61-0

robohacker:
#15768 seems to breaking the source build of this package since it overwrote the release tag.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5421,7 +5421,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/AutonomyLab/parrot_arsdk-release.git
-      version: 3.12.6-1
+      version: 3.12.61-0
     source:
       type: git
       url: https://github.com/AutonomyLab/parrot_arsdk.git


mikaelarguedas:
TL;DR: If you make changes to the source code and not only packaging details (more details below), you should increase the version number (either major, mino and patch e.g. 3.12.6 -> 3.12.7). You can either do it manually or using `catkin_prepare_release`. We recommend doing the steps in [this tutorial](http://wiki.ros.org/bloom/Tutorials/ReleaseCatkinPackage) for releasing new versions of a package.
I would suggest releasing a 3.12.7 rather than a 3.12.61. But we can also merge this as is if you prefer.

What happened in #15768 is that you ran `bloom-release` without increasing the version number, so you asked for a new release of 3.12.6. To do so bloom created a debian increment (3.12.6-0 -> 3.12.6-1). 
Debian increments are used to allow developers to make tweaks to the packaging of a package without altering the source and thus without changing the version number. In your case the changes were not packaging related but source related. This causes the jobs to fail because dpkg assumes that the binary artifacts (library, compressed files etc) of the package are identical if you do a debian increment.

robohacker:
Thanks @mikaelarguedas for your detailed explanation. We have been aware of this issue and have been using the Debian increments as a hack not to change the version number of the package in case we change anything in our packaging scripts. The rationale behind this decision was to keep the catkin package version consistent with the SDK version we are packaging. The upstream SDK version is 3.12.6, but it failed to build for some targets. We investigated this issue and modified the SDK with a patch [1], but released it with the same version number. As you correctly mentioned, using Debian package increments in this case is not correct. Since we can not use 3.12.7, I decided to specify the change as 3.12.61. It is not 100% semantically correct, but I can not think of any other option at the moment. If that's fine with you, please merge this and #15770 as is.

[1] The upstream SDK comes with its own workspace management and build tools/scripts, which are fairly complex. Maybe a better option here could be to re-think our packaging strategy and let bloom add all catkin files, and apply our changes as patches in the release repository. We will look into this.

mikaelarguedas:
@robohacker Sounds good, thanks for the explanation :+1: . Glad to hear that you were aware of it and were leveraging it as expected (and sorry for the unnecessary explanation). Merging this as is then!

The solution mentioned in [1] seems a good strategy to be reconcile SDK versions and ros package version in the future.

dirk-thomas:
You won't be able to release a version 3.12.7 after this since .61 will be considered higher and therefore a future .7 would be considered not newer.

mikaelarguedas:
> You won't be able to release a version 3.12.7 after this since .61 will be considered higher and therefore a future .7 would be considered not newer.

Indeed you will have to wait until the next minor version bump to be able to match the SDK version. Any further 3.12 revision will have to increment the patch version (.61) to be considered newer

Issue #11852, opened 2016-06-15
Enforce version numbers in rosdistro CI jobs follow best practices


tfoote:
With ompl now released there are no packages which don't follow the recommended versioning pattern. We cannot make catkin_pkg more strict for it might break non-standard rules. But we can enforce it in the rosdistro index that if you want to be indexed we can enforce it. 

This should be added to the Pull Request verification checks.

Replaces: https://github.com/ros-infrastructure/catkin_pkg/issues/140

mikaelarguedas:
this has been addressed in #15774

PR #15774, opened 2017-08-14
enforce version conventions

mikaelarguedas:
This is a suggestion to address #11852 

Option 1 Performs the check of the version format in this scripts directly
Option 2 Parses the warnings from catkin_pkg.parse_package_string and add an error if the warning about version conventions is present

Looking for feedback regarding which approach is preferred

### DIFF
--- a/test/test_build_caches.py
+++ b/test/test_build_caches.py
@@ -1,5 +1,4 @@
 from collections import OrderedDict
-# import re
 import os
 
 from catkin_pkg.package import parse_package_string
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Do all package manifests at the moment parse without this warning?

I prefer option 2 since it doesn't duplicate the logic.

test/test_build_caches.py
```diff
@@ -39,7 +40,21 @@ def test_build_caches():
...
+                #     errors.append('%s: %s' % (dist_name, 'Package "%s" does not follow the version conventions. It should not contain leading zeros (unless the number is 0).' % pkg_name))
+
+                # Option 2 test for specific catkin_pkg warning
+                warnings = []
+                pkgs[pkg_name] = parse_package_string(pkg_xml, warnings=warnings)
+                for warning in warnings:
+                    if 'version conventions' in warning:
+                        errors.append('%s: %s' % (dist_name, warning))
+                    else:
+                        print('WARNING: ' + warning)
```
dirk-thomas: In both cases (error and warning) it would be valuable to mention the `pkg_name`. The `dist_name` is already being printed before so it should be clear to which distro the message belong.

mikaelarguedas:
> Do all package manifests at the moment parse without this warning?

Yes, since ompl has updated their version scheme all packages parse without this warning. The only warnigs we get are for naming conventions for a few python packages that have the same name as the upstream one

test/test_build_caches.py
mikaelarguedas: hmm good point, I didnt print the package name because, so far, every warning/error from the parsing function already contain the package name in the message (e.g. https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/package.py#L212) but maybe it's an assumption that won't hold in the future and we should print the name of the package before.
You're right the distribution name doesn't add value here.

Note: Catching the warnings in `parse_package_string` forces us to duplicate the logic for printing them.

dirk-thomas (approved):
lgtm

PR #15784, opened 2017-08-16
[rosdep] Added python-pyro

krukai:
Added rosdep entries for Pyro4, a distributed object middleware for Python (RPC).

Note: the library is called Pyro4 inside the Python ecosystem to preserve the original, discontinued Pyro library. System package managers however provide Pyro4 under pyro. I followed that convention for rosdep.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -130,6 +130,15 @@ pyqt5-dev-tools:
   fedora: [python-qt5-devel]
   gentoo: [dev-python/PyQt5]
   ubuntu: [pyqt5-dev-tools]
+pyro4:
+  arch: [python2-pyro]
+  debian: [python2-pyro4]
+  fedora: [python-pyro]
+  gentoo: [dev-python/pyro]
+  osx:
+    pip:
+      packages: [pyro4]
+  ubuntu: [python2-pyro4]
 pyros-setup-pip:
   ubuntu:
     pip:


mikaelarguedas:
Thanks @krukai for the contribution.
A few remarks based on the [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules):
- We do recommend to use the system packages rather than pip entries (we cannot release ROS packages depending on pip rules). Please use `pyro4` from apt for debian and ubuntu rules
- Given that Pyro4 is the name of the maintained package and the one available on all supported platforms I would call the rosdep key `pyro4` as well

mikaelarguedas:
arch: https://www.archlinux.org/packages/community/any/python2-pyro/
debian: https://packages.debian.org/stretch/pyro4
fedora: https://apps.fedoraproject.org/packages/python-pyro
gentoo: https://packages.gentoo.org/packages/dev-python/pyro
ubuntu: https://packages.ubuntu.com/xenial/pyro4

note: the debian and ubuntu version depends on python2-pyro4

krukai:
I am terribly sorry for missing some of the important details, so thank you for providing the links! I changed the pull request accordingly:

- Renamed rosdep key to `pyro4`, dropping the `python-` prefix, and moving it accordlingly.
- Changed the Debian and Ubuntu package sources to `pyro4`.

mikaelarguedas:
Thanks @krukai for iterating on this.
Could you add a `depends` for the ubuntu and debian rules to reflect that we need the `python2-pyro4` package? (you can find an example [here](https://github.com/krukai/rosdistro/blob/9b3864d30cffabdec95237ca2b3cc1e258d178c8/rosdep/python.yaml#L543))

Without it (at least on Ubuntu Xenial) it looks like apt will default to installing the python3 version:
```
The following NEW packages will be installed:
  pyro4 python3-pyro4 python3-serpent
```

mikaelarguedas:
@krukai Sorry for the confusion. The depends tag cannot be used with apt.
Can you just use python2-pyro4 as the rule for Debian and Ubuntu?
```
debian: [python2-pyro4]
ubuntu: [python2-pyro4]

krukai:
@mikaelarguedas Learned that the hard way, but it's all good. The commit should pop up any second now.

mikaelarguedas (approved):
looks good now, thanks!

tfoote (approved):

PR #15787, opened 2017-08-17
Add Gentoo resolution for bloom and for python-omniorb

allenh1:
After this, I'll be missing only 36 keys for Lunar, Kinetic and Indigo.


mikaelarguedas (approved):
lgtm

@allenh1 Please provide links to the package when requesting to add them to the database.

https://packages.gentoo.org/packages/net-misc/omniORB
https://github.com/ros/ros-overlay/blob/master/dev-python/bloom/bloom-0.5.26.ebuild

PR #15790, opened 2017-08-18
Collapse Ubuntu distributions for python-wstool

jatowler:
All the distributions use `python-wstool` as the name, so they can be collapsed. This also future-proofs it against (in this case) not adding a `xenial` distribution.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3721,13 +3721,7 @@ python-wstool:
   fedora: [python-wstool]
   gentoo: [dev-python/wstool]
   macports: [py27-wstool]
-  ubuntu:
-    precise: [python-wstool]
-    quantal: [python-wstool]
-    raring: [python-wstool]
-    saucy: [python-wstool]
-    trusty: [python-wstool]
-    trusty_python3: [python-wstool]
+  ubuntu: [python-wstool]
 python-wtforms:
   debian: [python-wtforms]
   fedora: [python-wtforms]


dirk-thomas (dismissed):

rosdep/python.yaml
```diff
@@ -3721,13 +3721,7 @@ python-wstool:
...
   gentoo: [dev-python/wstool]
   macports: [py27-wstool]
-  ubuntu:
-    precise: [python-wstool]
-    quantal: [python-wstool]
-    raring: [python-wstool]
-    saucy: [python-wstool]
-    trusty: [python-wstool]
-    trusty_python3: [python-wstool]
+  ubuntu: [python-wstool]
```
mikaelarguedas: unfortunately collapsing them will prevent users from using the python3 version of the rule.
Though trusty_python3 should be `python3-wstool` and not `python-wstool`.
@tfoote I see this comes from #4018. Do you prefer to fix the python3 rule or collapse them ?

tfoote (approved):
I'd rather collapse them. Early in our python3 process we were working on having support for switching the whole system to use python3. We made good progress and a lot of the core can support that. However, it doesn't really scale. We've made a lot of progress on having python2 and python3 things install side by side. And I think that having multiple keys will likely be better than trying to have variants of distros based on the python version.

Also releated: https://github.com/ros-infrastructure/rosdep/pull/542

dirk-thomas:
Independent of the side-by-side installation we still need a way to install the Python 3 specific dependencies. Currently these keys are the only way to do that. Therefore I would rather not remove them.

mikaelarguedas:
Sounds good, collapsing then

PR #15791, opened 2017-08-19
test that source and release repos are different

mikaelarguedas:
fail the test if source and release repo URLs exist and are identical

### DIFF
--- a/scripts/check_rosdistro_urls.py
+++ b/scripts/check_rosdistro_urls.py
@@ -21,9 +21,9 @@ def main(index_url, rosdistro_name):
         sys.stdout.write('.')
         sys.stdout.flush()
         repo = distribution_file.repositories[repo_name]
-        if repo.release_repository and repo.source_repository and repo.release_repository == repo.source_repository:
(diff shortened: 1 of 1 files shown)


tfoote (approved):

dirk-thomas:
The condition seems to compare the repository instance rather than the url. I would expect them to never be equal. Have you tried to make this condition fail?

mikaelarguedas:
You're right. I only confirmed that it wouldnt fail on None instances and was planning on testing before considering merging. because it seemed odd that we didnt have a single violation on all indexed repos.

mikaelarguedas:
This seems to now do the right thing. Not sure how we want to act on the entries that don't comply. Should we contact maintainers individually ?

dirk-thomas:
The entries which list a valid release repo as their source entry might be completely fine as they are. Maybe they add a manifest in the release process. So I consider that a valid use case.

mikaelarguedas:
Given that there is no good way to know if the repository listed in the source entry is a valid release repository I'm going to close this. Maybe we'll think of a better way of testing that later on

PR #15803, opened 2017-08-21
Add xenial key for libpqxx and libpqxx-dev

jatowler:
These are the most up-to-date versions for Xenial according to [the libpqxx page](https://launchpad.net/ubuntu/+source/libpqxx).


mikaelarguedas:
@jatowler Can you please add more recent ubuntu platforms as well?
`libpqxx-dev`  for yakkety zesty and artful (https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libpqxx-dev&searchon=names)
`libpqxx-4.0v5` for yakkety zesty and artful (https://packages.ubuntu.com/search?keywords=libpqxx-4.0v5&searchon=names&suite=all&section=all)

jatowler:
@mikaelarguedas Done. The commit message only lists yakkety and artful, but I included zesty too.

dirk-thomas:
`wily` is still missing.

jatowler:
@dirk-thomas Wily has been EOL for over a year ([announcement](https://lists.ubuntu.com/archives/ubuntu-security-announce/2016-July/003492.html)), but I've added it anyway.

mikaelarguedas (dismissed):
lgtm, thanks @jatowler for iterating

rosdep/base.yaml
```diff
@@ -2389,6 +2389,11 @@ libpqxx:
     trusty: [libpqxx-4.0]
     utopic: [libpqxx-4.0]
     vivid: [libpqxx-4.0]
+    wily: [libpqxx-4.0]
+    xenial: [libpqxx-4.0]
+    yakkety: [libpqxx-4.0v5]
+    zesty: [libpqxx-4.0v5]
+    artful: [libpqxx-4.0v5]
```
dirk-thomas: `artful` needs to be in front of the list (alphabetical order).

Same below.

jatowler: I don't think this makes sense. 17.10 comes after 17.04; it's reasonable for the versions to be in order of release, not in order of name. Otherwise versions will get weirdly interleaved as we move forward, and instead of appending new versions to the end, they will be inserted awkwardly in the middle. For example, we would have Warty in between Vivid and Wily and Hoary in between Hardy and Intrepid, so the version list would do this:

* 15.04
* 4.10
* 15.10
* 16.04
* ...

dirk-thomas: In order to use a custom order the code would need to implement the serialization manually as well as know a list of all platforms (which would need to be continuously updated).

For now the ordering is alphabetical. Please update the patch accordingly in order to be mergable.

jatowler: I've made the change to allow the merge, but order is an odd constraint for a hash.

dirk-thomas: The mandatory order is beneficial to ensure that separated edits don't collide as much as possible as well as. It also allows a human to scan through the hashes to search for a specific one in less than `O(n)`.

tfoote:
I think this is failing CI due to the problem fixed in #15805

mikaelarguedas:
retriggering travis...

mikaelarguedas (approved):
lgtm

PR #15804, opened 2017-08-21
Update keys for PySide

jatowler:
* Collapses keys for `pyside-tools` because the package is identical in all versions of Ubuntu.
* Adds keys for `python-pyside` and `python-pyside.qtuitools` for Wily, Xenial, Yakkety, Zesty, and Artful. Note that these are also identical across all Ubuntu versions, but Python 3 is called out separately.


mikaelarguedas:
pyside depends on qt4 that is not targeted starting with [ros kinetic](http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021). We encourage users to use qt5 and pyside2 for these distributions and I think that's the reason these keys have not been added after vivid

jatowler:
Okay, understood. You can tell I'm working through a list today. I'll close this PR for now, and re-open if it turns out it's useful after all.

PR #15805, opened 2017-08-22
fix to finds duplicate between non overlapping rules

k-okada:
c.f. https://github.com/ros/rosdistro/pull/15800#issuecomment-323850356

```
it appears to be throwing errors about mercurial being multiply defined:

mercurial is multiply defined in
    file:///home/travis/build/ros/rosdistro/rosdep/python.yaml and 
    file:///home/travis/build/ros/rosdistro/rosdep/base.yaml
```

-  use _load_view_dependeencies to avoid merging views
- add more debug view when multiply defined

### DIFF
--- a/scripts/check_duplicates.py
+++ b/scripts/check_duplicates.py
@@ -95,7 +95,7 @@ def check_duplicates(sources, os_name, os_codename):
             ):
                 continue
             if dep_name in db_name_view:
-                print('%s (%s,%s) is multiply defined in\n\t%s and \n\t%s\n' %
+                print('%s (%s, %s) is multiply defined in\n\t%s and \n\t%s\n' %
                       (dep_name, os_name, os_codename, db_name_view[dep_name], view_key))
                 has_duplicates = True
             db_name_view[dep_name] = view_key
--- a/scripts/check_duplicates.py
+++ b/scripts/check_duplicates.py
@@ -79,7 +79,6 @@ def check_duplicates(sources, os_name, os_codename):
     db_name_view = {}
     has_duplicates = False
     # to avoid merge views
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Thanks @k-okada for the quick fix!

This fixes the issue when there is an non overlapping entry in different files.

It seems that this still doesnt check the `osx-homebrew.yaml` file, does it ?
I tried adding
```
mercurial:
  osx:
    pip:
      packages: [mercurial]
```
to `osx-homebrew.yaml` and the test passed. It should be flagged as a duplicate of the same entry in `python.yaml` right?

k-okada:
ok, added osx check

--
◉ Kei Okada


2017-08-22 20:16 GMT+09:00 Mikael Arguedas <notifications@github.com>:

scripts/check_duplicates.py
```diff
@@ -78,7 +78,9 @@ def check_duplicates(sources, os_name, os_codename):
     print("checking duplicates")
     db_name_view = {}
     has_duplicates = False
-    view = lookup.get_rosdep_view(DEFAULT_VIEW_KEY, verbose=None)  # to call init
+    # to avoid merge views
+    #view = lookup.get_rosdep_view(DEFAULT_VIEW_KEY, verbose=None)  # to call init
```
mikaelarguedas: can this line be removed? or do we want to keep it for bookkeeping? if we want to keep it can you please use `# ` (with a space after) to comply with pep8?

mikaelarguedas (commented):
Thanks @k-okada for iterating! the change looks good to me :+1: , I have one minor comment inline

mikaelarguedas:
@ros/rosdistro-maintainers can someone review this please?

dirk-thomas:
@k-okada Thanks for the fix.

Replaced by #15812.

PR #15808, opened 2017-08-22
add joystick rule for fedora 25 and 26

mikaelarguedas:
https://admin.fedoraproject.org/pkgdb/package/rpms/linuxconsoletools/

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1212,17 +1212,7 @@ java:
 joystick:
   arch: [linuxconsole]
   debian: [joystick]
-  fedora:
-    '21': [linuxconsoletools]
-    '22': [linuxconsoletools]
-    '23': [linuxconsoletools]
-    '24': [linuxconsoletools]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
retriggering travis...

dirk-thomas:
Should this rule be collapsed or are there any Fedora distros before 21?

tfoote:
I see builds back to fedora 18 on the above link. We can probably collapse this.

mikaelarguedas:
Yeah I was wondering the same thing. I didn't want to remove existing rules so I defaulted to adding them. But I don't think many people did or are targetting beefy (fedora 17) so I can collapse them if we agree on removing it

tfoote:
+1 for collapsing, It's been EOL since January 2014 and is the 8th oldest EOL fedora release (24 is already EOL). Fedora moves quickly.

tfoote (approved):

PR #15815, opened 2017-08-23
Add python-plyfile

cglwn:
Package is only in PyPI. I am only using it to simplify dependency management and have no plans to make a release with it so it should be alright.

PyPI: https://pypi.python.org/pypi/plyfile

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2029,6 +2029,19 @@ python-planar-pip:
   ubuntu:
     pip:
       packages: [planar]
+python-plyfile-pip:
+  debian:
+    pip:
+      packages: [plyfile]
+  fedora:
+    pip:
+      packages: [plyfile]
+  osx:
+    pip:
+      packages: [plyfile]
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Great, thanks for the link and the rationale!
A few comments based on the [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules):
- Please use the suffix `-pip` for packages available in pip only as per
- Use the smallest notation possible, for example
```
  ubuntu:
    pip:
      packages: [plyfile]
```
rather than listing each ubuntu distribution

cglwn:
Thanks for the review. I've added the changes requested.

mikaelarguedas:
@cglwn can you add the other platforms as well ? 
"Fedora and Gentoo have packaging jobs and should be filled in if they are available.
OSX is also nice to have."
Thanks!

mikaelarguedas (approved):
thanks @cglwn for iterating

mikaelarguedas:
ready for 2nd review

PR #15818, opened 2017-08-24
Adding 'dyn_tune' and 'rqt_dyn_tune' to documentation index for distro

mehdish89:
i'd like 'dyn_tune' and 'rqt_dyn_tune' to be indexed and documented on ros.org. These are the packages of the tool that I created for generic parameter tuning in ROS, during my internship at OSRF.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1571,11 +1571,11 @@ repositories:
   dyn_tune:
     doc:
       type: git
-      url: https://github.com/mehdish89/dyn_tune
+      url: https://github.com/mehdish89/dyn_tune.git
       version: master
     source:
       type: git
-      url: https://github.com/mehdish89/dyn_tune
+      url: https://github.com/mehdish89/dyn_tune.git
       version: master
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -1568,6 +1568,15 @@ repositories:
...
       url: https://github.com/delftrobotics/dr_base-release.git
       version: 1.0.0-0
+  dyn_tune:
+    doc:
+      type: git
+      url: https://github.com/mehdish89/dyn_tune
+      version: master
+    source:
+      type: git
+      url: https://github.com/mehdish89/dyn_tune
```
mikaelarguedas: @mehdish89 the referenced urls must use the full url (in the case of git have the `.git` in the end). You can refer to the travis job output for more details on why the checks fails

mehdish89: @mikaelarguedas i fixed it. i hope thats correct now.

mikaelarguedas (approved):
all good now :+1:

PR #15819, opened 2017-08-25
solving the dependency issue

mehdish89:
'rqt_dyn_tune' dependency on 'dyn_tune' is added in the distribution.yaml

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -7545,6 +7545,8 @@ repositories:
     status: maintained
   rqt_dyn_tune:
     doc:
+      depends:
+      - dyn_tune
       type: git
       url: https://github.com/mehdish89/rqt_dyn_tune.git
       version: master
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1571,11 +1571,11 @@ repositories:
   dyn_tune:
     doc:
       type: git
-      url: https://github.com/mehdish89/dyn_tune
+      url: https://github.com/mehdish89/dyn_tune.git
       version: master
     source:
       type: git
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Does the documentation fail to generate without this ? rosdoc_lite seems to run without error.
@dirk-thomas do you know why an expllicit dependency is necessary here ? It doesnt seem that any other ROS package in kinetic need it

mehdish89:
@mikaelarguedas yes it does. 'rqt_dyn_tune' depends on another package i am indexing 'dyn_tune'. and because 'dyn_tune' is not released, apparently it cannot be resolved as a dependency on Jenkins. @tfoote suggested to use this format.

mikaelarguedas:
Ok so this is necessary because the package is not released and is using package format 1 so cannot declare the doc depend directly in the package.xml. That must be a rare situation expllaining why I didnt see it before. Thanks for clarifying!

tfoote:
@mikaelarguedas I believe you still need the dependency in the rosdistro even if it's format2 if it's unreleased since the rosdistro doesn't have the repo to package indexing available for unreleased packages.

mikaelarguedas:
@tfoote oh interesting, thanks for the info. I need to learn more about how doc jobs work :/.
So if `dyn_tune` was released but `rqt_dyn_param` was not, this would work without adding the depend here ? (we can move the discussion somewhere else, I'm just asking questions to make more correct comments in the future and understand how the other unreleased packages work without this extra step)

dirk-thomas:
> So if `dyn_tune` was released but `rqt_dyn_param` was not, this would work without adding the depend here ?

Correct.

PR #15865, opened 2017-09-03
Add rosbag_metdata as python rosdep

jacobperron:
I've found this tool handy for adding metadata to bag files.

- https://pypi.python.org/pypi/rosbag_metadata
- https://github.com/hordurk/rosbag_metadata

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3859,6 +3859,19 @@ python3-setuptools:
   ubuntu: [python3-setuptools]
 python3-yaml:
   ubuntu: [python3-yaml]
+rosbag-metadata-pip:
+  debian:
+    pip:
+      packages: [rosbag-metadata]
+  fedora:
+    pip:
+      packages: [rosbag-metadata]
+  gentoo:
+    pip:
+      packages: [rosbag-metadata]
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


Alessandro-Barbieri:
Don't use pip on gentoo

dirk-thomas (approved):

dirk-thomas:
@Alessandro-Barbieri Please provide an alternative suggestion.

Alessandro-Barbieri:
The alternative is writing an ebuild for the package, but don't use `pip`.
Running `pip` as root on gentoo will break the system.

see https://wiki.gentoo.org/wiki/Pip

dirk-thomas:
@Alessandro-Barbieri I guess you can't a contributor to write an ebuild so how would you like to proceed on this: should the `gentoo` entry be removed or should the patch keep the pip entry?

mikaelarguedas:
In other PRs we just don't add the pip entry for gentoo to avoid breakage so I'd suggest to do the same here

dirk-thomas (approved):
Sounds good to me.

mikaelarguedas (approved):

mikaelarguedas:
@Alessandro-Barbieri feel free to submit a follow-up PR adding a gentoo entry if an ebuild for this is created for this package

PR #15867, opened 2017-09-04
cogrob_catkin_grpc: 0.0.1-0 in 'indigo/distribution.yaml' [bloom]

BillWSY:
Increasing version of package(s) in repository `cogrob_catkin_grpc` to `0.0.1-0`:

- bloom version: `0.5.22`

## grpc

```
* Pre-release commit. (#1 <https://github.com/CogRob/catkin_grpc/issues/1>)
* Contributors: Shengye Wang
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -1964,23 +1964,6 @@ repositories:
       url: https://github.com/ipa320/cob_supported_robots.git
       version: indigo_dev
     status: developed
-  cogrob_catkin_grpc:
-    doc:
-      type: git
-      url: https://github.com/CogRob/catkin_grpc.git
-      version: master
-    release:
-      packages:
-      - grpc
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/CogRobRelease/catkin_grpc-release.git
-      version: 0.0.1-0
-    source:
-      type: git
-      url: https://github.com/CogRob/catkin_grpc.git
-      version: master
-    status: developed
   collada_urdf:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -1964,6 +1964,23 @@ repositories:
       url: https://github.com/ipa320/cob_supported_robots.git
       version: indigo_dev
     status: developed
+  cogrob_catkin_grpc:
+    doc:
+      type: git
+      url: https://github.com/CogRobRelease/catkin_grpc-release.git
+      version: master
```
dirk-thomas: The master branch of the release repo doesn't contain any ROS packages which could be indexed. You probably want to use the source repo here.

BillWSY: Thanks for the comments. PTAL.

indigo/distribution.yaml
```diff
@@ -1964,6 +1964,23 @@ repositories:
       url: https://github.com/ipa320/cob_supported_robots.git
       version: indigo_dev
     status: developed
+  cogrob_catkin_grpc:
```
dirk-thomas: Since the package name is just `grpc` this name could also be `grpc`. It would make it easier for users calling `rosinstall_generator` since they need to use this name to identify which repos to fetch.

If you update this line the following lines in the `release` section will need to be removed since the single package name matches the repo name now:

```
packages:
- grpc
```

BillWSY: Thanks for the comments. PTAL.

dirk-thomas:
Thanks for iterating on this so quickly.

PR #15870, opened 2017-09-05
airbus_coop: 0.0.2-0 in 'indigo/distribution.yaml' [bloom]

NadiaHG:


indigo/distribution.yaml
```diff
@@ -159,6 +159,24 @@ repositories:
...
+  airbus_coop:
+    release:
+      packages:
+      - agi_docgen
+      - airbus_coop
+      - cobot_gui
+      - plugin_log_manager
+      - plugin_node_manager
+      - plugin_rqt
+      - plugin_rviz
```
dirk-thomas: These packages names seem to be very generic. I think it would be good to rename those to include something more specific.

tfoote: For reference: http://wiki.ros.org/ROS/Patterns/Conventions and http://www.ros.org/reps/rep-0144.html may be useful.

ipa-nhg: Thanks a lot for the suggestions and links. This pull request can be closed, superseded by https://github.com/ros/rosdistro/pull/15876

PR #15873, opened 2017-09-05
[dep][py] paramiko: Collapse entries for all Ubuntu distros.

130s:
Zesty: https://packages.ubuntu.com/zesty/python-paramiko


dirk-thomas:
Since the mapping is the same for all Ubuntu distros it would make sense to collapse them to avoid the need to add new distros in the future .

130s:
@dirk-thomas I see. Done.

dirk-thomas (approved):
Thanks.

PR #15886, opened 2017-09-07
rqt_shell: 0.4.9-0 in 'jade/distribution.yaml' [bloom]

DorianScholz:
Increasing version of package(s) in repository `rqt_shell` to `0.4.9-0`:

- upstream repository: https://github.com/ros-visualization/rqt_shell.git
- release repository: https://github.com/ros-gbp/rqt_shell-release.git
- distro file: `jade/distribution.yaml`
- bloom version: `0.5.26`
- previous version for package: `0.4.8-0`

## rqt_shell

```
* Fixed syntax error when using a startup script with spyder shell
```

PR #15889, opened 2017-09-07
add jade to eol distro list

mikaelarguedas:
to avoid doing it manually https://github.com/ros/rosdistro/pull/15886

### DIFF
--- a/scripts/__init__.py
+++ b/scripts/__init__.py
@@ -1,1 +1,1 @@
-eol_distro_names = ['groovy', 'hydro']
+eol_distro_names = ['groovy', 'hydro', 'jade']


dirk-thomas (approved):

mikaelarguedas:
Food for thoughts: I remember that this information is also stored by the roswiki, known by the ros_buildfarm somehow and likely other places. Would it make sense to consolidate this information in a single human-readable and machine-consumable file ? Maybe an amendment to the index file to define the status of a distro ?

mikaelarguedas:
I'm going to merge this but feel free to comment here post-merge

tfoote:
It would be good to have generally, but we do need to keep things like the wiki from loading remote resources pre-render so sometimes the caching/duplication is important.

mikaelarguedas:
Thanks for the feedback, do you guys think it belongs in the index file ? or that we should specify this somewhere else ?
(slightly related we had the same interrogation when generating dockerfiles for all supported ros distributions and were wondering if it would make sense to have a way to know all active ROSDistros/OSDistros/Architectures supported without having to parse rosdistro + ros_buildfarm_config + custom logic for EOL ROS distros)

PR #15909, opened 2017-09-12
Adding advanced_navigation_driver to documentation index for distro

an-scott:
Please add my repo to the documentation index

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -11,10 +11,10 @@ release_platforms:
   - xenial
   - zesty
 repositories:
-  Advanced_Navigation_driver:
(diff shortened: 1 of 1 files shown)


lunar/distribution.yaml
```diff
@@ -11,6 +11,11 @@ release_platforms:
   - xenial
   - zesty
 repositories:
+  Advanced_Navigation_driver:
```
tfoote: We don't officially have a policy for repository names, but please stick to the lowercase underscoring naming scheme. http://www.ros.org/reps/rep-0144.html

tfoote (changes requested):

PR #15926, opened 2017-09-15
Move joint_state_publisher from robot_model to it's own repository (indigo/kinetic/lunar)

sloretz:
part of ros/robot_model#195

To indigo this releases 1.11.15-0
To kinetic/lunar this releases 1.12.12-0

This moves joint_state_publisher out of the robot_model repository. Bloom was used up to the point of creating this pull request.

```
$ bloom-release --version
0.5.26
```

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -5196,8 +5196,6 @@ repositories:
       url: https://github.com/ros/joint_state_publisher.git
       version: indigo-devel
     release:
-      packages:
-      - joint_state_publisher
       tags:
         release: release/indigo/{package}/{version}
       url: https://github.com/ros-gbp/joint_state_publisher-release.git
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -5190,6 +5190,24 @@ repositories:
       url: https://github.com/JenniferBuehler/general-message-pkgs.git
       version: master
     status: maintained
+  joint_state_publisher:
+    doc:
+      type: git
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
Looks good to me.
Before removing the package from the robot_model source repo, it may be worth checking downstream packages to make sure nobody relies on the previous URI like it happened when we started moving robot model packages

sloretz:
@mikaelarguedas good idea. Looks like this is in the clear. I checked two levels down using the ros wiki `Used by`. All were on github so I searched for `joint_state_publisher` and `robot_model`. Nothing looking like a URL came up.

dirk-thomas:
While `Jade` has just been EOLed you might want to make an effort to maintain the ability to build it from source in case users still want to do so.

mikaelarguedas:
@sloretz @clalancette I just noticed something fishy with EOL distros:
All distros before indigo are fine because they had their own branchs that stayed unmodified and still have all the packages in them.
Now that Jade became EOL it's a bit trickier because it uses the indigo-devel branch like indigo but that branch doesn't have the packages anymore so people cannot build them from source anymore.
Speaking with @tfoote the best way to go would be to create a tag jade-devel on the robot-model repo with the state of the repo before Jade became EOL and update the Jade disctibution file to point to that new tag rather than the indigo-devel branch

sloretz:
@mikaelarguedas PR #15927 updates the jade distribution.yaml to use the tag `jade-devel`. May I ask you to check if I missed anything in it?

PR #15927, opened 2017-09-15
Change robot_model from indigo-devel branch to jade-devel tag

sloretz:
The purpose of this is to not break `rosinstall_generator --upstream-development` for the EOL distro jade once the package `joint_state_publisher` is removed from the `indigo-devel` branch. This changes the version for the repository `robot_model` from the branch `indigo-devel` to a tag `jade-devel`. The tag is set to the state of the repo when [the last jade sync happened](https://discourse.ros.org/t/final-jade-sync-2017-09-03/2559). 

See #15926 
See ros/robot_model#195

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -5280,7 +5280,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros/robot_model.git
-      version: indigo-devel
+      version: jade-devel
     release:
       packages:
       - joint_state_publisher
@@ -5295,7 +5295,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/ros/robot_model.git
-      version: indigo-devel
+      version: jade-devel
     status: maintained
   robot_pose_publisher:
     doc:


mikaelarguedas:
looks good to me, thanks for the follow-up

mikaelarguedas:
Note that the job will fail because we dont allow modifications in eol distros, but it's ok to merge anyway

sloretz:
Note to future self, this failed because the source entry must be a branch and not a tag (thanks @dirk-thomas). I'll delete the jade-devel tag on robot_model and make it a branch. 
```
ERROR: In file '''/home/travis/build/ros/rosdistro/jade/distribution.yaml''': Could not validate source entry for repo robot_model with error [[[Could not validate repository with url https://github.com/ros/robot_model.git and version jade-devel from entry at line '''5294''']]]
```

sloretz:
Failed with only the expected failure:

```
ERROR: In file '''/home/travis/build/ros/rosdistro/jade/distribution.yaml''': There is a change to a doc section of an EOLed distribution. Lines: [5283]
```
so I'm going to squash/merge this.

PR #15959, opened 2017-09-22
Adding hebiros to documentation index for distro

xaxafour:
I'd like hebiros to be indexed and documented on ros.org.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2574,6 +2574,17 @@ repositories:
       url: https://github.com/davidfischinger/haf_grasping.git
       version: kinetic
     status: maintained
+  hebiros:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -2574,6 +2574,17 @@ repositories:
       url: https://github.com/davidfischinger/haf_grasping.git
       version: kinetic
     status: maintained
+  hebiros:
+    doc:
+      type: git
+      url: https://github.com/HebiRobotics/HEBI-ROS.git
+      version: master
+    release:
```
mikaelarguedas: If you only want hte package to be documented, please do not add the release section. If you want to release a package you will need to [run bloom](http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease) and use a separate release repository

mikaelarguedas (approved):
lgtm

PR #15969, opened 2017-09-25
Adding pheeno_ros to documentation index for indigo and kinetic

zmk5:
I also forgot to add `src` info the first time for the indigo release.


zmk5:
Thank you @dirk-thomas!!!

PR #15973, opened 2017-09-27
Update rqt_robot_monitor for Lunar

arne48:
This introduces the newest release of rqt_robot_monitor to the distribution.


mikaelarguedas:
If you want bloom can open the PR for you in the future

arne48:
I will try that again. This time the creation of the PR failed with bloom. Next time I will also create the oauth token maybe not creating one was the culprit.

tfoote:
If you have 2FA setup using an OAuth Token is required. http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization

PR #15974, opened 2017-09-27
libfranka: 0.1.0-0 in 'kinetic/distribution.yaml' [bloom]

fwalch:
Increasing version of package(s) in repository `libfranka` to `0.1.0-0`:

- upstream repository: https://github.com/frankaemika/libfranka.git
- release repository: https://github.com/frankaemika/libfranka-release.git
- bloom version: `0.5.26`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3713,6 +3713,18 @@ repositories:
       url: https://github.com/AutonomyLab/libcreate.git
       version: master
     status: developed
+  libfranka:
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/frankaemika/libfranka-release.git
+      version: 0.1.0-0
+    source:
+      test_commits: false
+      type: git
+      url: https://github.com/frankaemika/libfranka.git
+      version: master
+    status: developed
   libfreenect:
     doc:
       type: git


kinetic/distribution.yaml
```diff
@@ -3713,6 +3713,17 @@ repositories:
...
+  libfranka:
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/frankaemika/libfranka-release.git
+      version: 0.1.0-0
+    source:
+      type: git
+      url: https://github.com/frankaemika/libfranka.git
+      version: master
```
dirk-thomas: Since the source repo doesn't have a `package.xml` file the buildfarm won't be able to build any code from this repo. Therefore you might want to remove the `source` section.

wjwwood: Shouldn't the recommendation be to disable commit testing instead?

Having a source section does not imply it is a catkin package or that it needs a package.xml in my opinion. It simply indicates where to get the original source code, full stop.

fwalch: Thanks :)

I wanted to do a prerelease build to catch these kinds of errors, but couldn't figure out how to do it as the package is not in the `distribution.yaml` yet.
Now I found http://lists.ros.org/lurker/message/20130708.231844.2e867ae7.de.html; is this info still correct?

> If you add it to the index without a version (...) you will be able to run a prerelease of
unreleased stuff.

(I understood this as "index" == `distribution.yaml`)

If yes, then I'll change this PR to not include the `version` for now and then re-add it after a local prerelease build succeeded.

wjwwood: Obviously it would be better to point the source entry to a place which is rolling AND has a package.xml, but it's still useful info to capture even if it doesn't have a `package.xml`.

----

@fwalch you can use http://prerelease.ros.org/kinetic to do a prerelease test, even on unreleased software using the "Add a custom repo" button on the first part of the wizard.

dirk-thomas: > Shouldn't the recommendation be to disable commit testing instead?

> Having a source section does not imply it is a catkin package or that it needs a package.xml in my opinion. It simply indicates where to get the original source code, full stop.

Users fetching the sources with e.g. `rosinstall_generator` won't be able to build the package though. So I personally don't see much use in providing it - but as long as the CI jobs are disabled I don't mind if you think it should stay.

fwalch: @wjwwood I tried that earlier, but something didn't seem to work. The "next" button doesn't proceed to step 3, but hangs on step 2.. the Javascript console shows `TypeError: repositories[repo_name] is undefined`. I feel quite stupid having to ask this, but can you tell me which values I need to put into the URL/branch fields? Trying locally, I also only got up to `The repository 'libfranka' was not found in the distribution file` when trying `generate_prerelease_script.py`.

wjwwood: You could use tuple (`libfranka`, `git`, `https://github.com/frankaemika/libfranka-release.git`, `release/kinetic/libfranka/0.1.0-0`).

fwalch: @wjwwood Thanks, that worked. A local prerelease test was successful.

---

Just let me know if I should change something regarding the `source` entry :)

wjwwood: @fwalch you can either leave it, remove it, or update it to point to a place where a `package.xml` is present. I believe that `test_commit` defaults to False so there won't be an issue as-is, based on http://www.ros.org/reps/rep-0143.html#distribution-file

But you could be explicit and do `test_commits: false`, e.g.: https://github.com/ros/rosdistro/blob/67827f63a944bf111da8e072a87d5e2016360618/kinetic/distribution.yaml#L555

dirk-thomas: `test_commits` defaults to `true` on the ROS buildfarm. Please explicitly disable it if you want to keep the source entry.

fwalch: Added `test_commits: false`.

PR #15990, opened 2017-09-29
add python-rosinstall for additional ubuntu distros

fmessmer:
[xenial](https://packages.ubuntu.com/search?suite=xenial&arch=any&searchon=names&keywords=python-rosinstall)
[zesty](https://packages.ubuntu.com/search?suite=zesty&arch=any&searchon=names&keywords=python-rosinstall)

or do you want me to introduce a line like this to cover all distros?
```
 ubuntu: [python-rosinstall]
```

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2694,6 +2694,7 @@ python-rosinstall:
   gentoo: [dev-python/rosinstall]
   macports: [p27-rosinstall]
   ubuntu:
+    artful: [python-rosinstall]
     lucid: [python-rosinstall]
     maverick: [python-rosinstall]
     natty: [python-rosinstall]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
thanks @ipa-fxm,
Can you also add ubuntu `yakkety` and `artful` for completeness as well as debian `jessie`?

ipa-fxm:
@mikaelarguedas 
or do you want me to introduce a line like this to cover all distros?
```
 ubuntu: [python-rosinstall]
```

mikaelarguedas:
unfortunately these lines cannot be collapsed because there is a rule for the python3 version of the package `trusty_python3: [python3-rosinstall]` that will disappear if we do so. So we are forced to add a new line for each distro for now.

If it's not too cumbersome it would be great if we can fill in all the missing ubuntu distributions even if they are EOL (so ubuntu U, V, W, Y and AA)

ipa-fxm:
@mikaelarguedas :+1:

tfoote (approved):

PR #16013, opened 2017-10-03
python_trep: 1.0.3-1 in 'kinetic/distribution.yaml' [bloom]

jarvisschultz:
Increasing version of package(s) in repository `python_trep` to `1.0.3-1`:

- upstream repository: https://github.com/MurpheyLab/trep.git
- release repository: https://github.com/MurpheyLab/trep-release.git
- bloom version: `0.5.26`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6162,6 +6162,21 @@ repositories:
       url: https://github.com/ros-visualization/python_qt_binding.git
       version: kinetic-devel
     status: maintained
+  python_trep:
+    doc:
+      type: git
+      url: https://github.com/MurpheyLab/trep-release.git
+      version: release/kinetic/python_trep
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/MurpheyLab/trep-release.git
+      version: 1.0.3-1
+    source:
+      type: git
+      url: https://github.com/MurpheyLab/trep.git
+      version: master
+    status: developed
   pyzmp:
     doc:
       type: git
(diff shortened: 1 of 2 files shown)


kinetic/distribution.yaml
```diff
@@ -6162,6 +6162,21 @@ repositories:
...
+      version: release/kinetic/python_trep
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/MurpheyLab/trep-release.git
+      version: 1.0.3-1
+    source:
+      type: git
+      url: https://github.com/MurpheyLab/trep.git
+      version: master
```
mikaelarguedas: This package doesn't seem to be a catkin package. In this case the build farm will not be able to run development jobs on changes in the source repository. I would recommend disabling development jobs by adding `test_commits: false` to the source section

tfoote: Upon review, doing that for the indigo stanza too would be good. Since it's running the devel job, but it's running 0 tests since there are no packages detected.

jarvisschultz: Sorry about that. You're definitely correct that the upstream repo isn't a catkin package. I just added the source section to Indigo and Kinetic, and I guess I didn't do enough research. Is adding `test_commits: false` to all distros sufficient? Would you rather me update this PR or open a new PR for Indigo?

mikaelarguedas: > Is adding test_commits: false to all distros sufficient? 

Yes, for kinetic it will not create the dev job at all, for indigo it will disable the existing one
 
> Would you rather me update this PR or open a new PR for Indigo?

I'm fine either way as long as it's reflected in the PR title or commit message

PR #16030, opened 2017-10-06
Temporarily disable gcloud_speech package because its build fails

BillWSY:


dirk-thomas:
Thank you for being proactive about this!

BillWSY:
I am not sure where to post this question:

Once I fix the problem in dependency (grpc package) and release a new "grpc" package, will the newest version of "grpc" be picked up during the build process, or do I have to wait until "grpc" shows up in packages.ros.org/ros/ubuntu/pool/main?

dirk-thomas:
Once you have released a new version of `grpc` with bloom it can be used by downstream packages. The Debian package doesn't have to be in the public repo. It will be available in the "building" repository immediately after it was build and can be used from there for downstream jobs on the buildfarm.

BillWSY:
Got it. Thank you very much for the information!

PR #16031, opened 2017-10-06
Update python-rosdistro

paulbovbel:
Add more distributions


wjwwood (approved):

mikaelarguedas:
thanks @paulbovbel, can youo also add `artful` for completeness? thanks!

dirk-thomas:
Can you please add `artful` to the top of the list.

wjwwood:
Yeah, that's a fair point, I keep forgetting that there are newer things 😄

PR #16040, opened 2017-10-06
Add logic to help with checking on hooks

tfoote:
This will help us make sure that pull request builds will provision correctly.

It checks if ros-pull-request-builder has appropriate permissions. It requires the environment ROSGHPRB_TOKEN to be set to check. But it will skip the test if that's unset.

### DIFF
--- a/test/hook_permissions.py
+++ b/test/hook_permissions.py
@@ -1,0 +1,155 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2017, Open Source Robotics Foundation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of the Willow Garage, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived from
+#       this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from __future__ import print_function
+
+import argparse
+import os
+import sys
+from github import Github, UnknownObjectException
+
+
+def detect_repo_hook(repo, cb_url):
+    for hook in repo.get_hooks():
+        if hook.config.get('url') == cb_url:
+            return True
+    return False
+
+
+class GHPRBHookDetector(object):
+    def __init__(self, github_user, github_token, callback_url):
+        self.callback_url = callback_url
+        self.gh = Github(github_user, github_token)
+
+    def get_repo(self, username, reponame):
+        try:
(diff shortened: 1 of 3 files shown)


dirk-thomas:
It is great to test this automatically rather on requiring us to manually do so. Is there anything in the output which indicates that the new checks have been run?

tfoote:
It prints a pass notice here: https://github.com/ros/rosdistro/pull/16040/files#diff-2fc07dfe4e98c3fa7008712936aabe4cR91

Maybe it would make more sense to explicitly invoke this as a separate travis test so for the ambiguous cases we know if it's just a hook issue and won't effect local users running tests that they cannot reproduce.

dirk-thomas:
> It prints a pass notice here: https://github.com/ros/rosdistro/pull/16040/files#diff-2fc07dfe4e98c3fa7008712936aabe4cR91

I didn't see the output in the log anywhere. I guess it only gets printed if the patch actually changes a repository entry so it isn't visible for this PR?

tfoote:
Correct, it only checks for hooks inside a changed source entry.

test/hook_permissions.py
```diff
@@ -0,0 +1,140 @@
...
+class GHPRBHookDetector(object):
+    def __init__(self, github_user, github_token, callback_url):
+        self.callback_url = callback_url
+        self.gh = Github(github_user, github_token)
+
+    def get_repo(self, username, reponame):
+        try:
+            repo = self.gh.get_user(username).get_repo(reponame)
+        except UnknownObjectException as ex:
+            # print(ex)
```
dirk-thomas: Should this commented line be removed?

dirk-thomas: Unused variable.

test/hook_permissions.py
```diff
@@ -0,0 +1,140 @@
...
+import argparse
+import os
+import sys
+from github import Github, UnknownObjectException
+
+
+def detect_repo_hook(repo, cb_url):
+    for hook in repo.get_hooks():
+        config = hook.config
+        if 'url' in config and config['url'] == cb_url:
```
dirk-thomas: Maybe use `config.get()` instead. Then the temporary variable `config` would also not be necessary.

test/hook_permissions.py
```diff
@@ -0,0 +1,140 @@
...
+                return False
+            else:
+                errors.append('Warning: Push access detected but unable to verify manual hook '
+                    'configuration for repo [ %s ]. Please visit ' % repo.full_name +
+                    'http://wiki.ros.org/buildfarm/Pull%20request%20testing '
+                    'and make sure hooks are setup.')
+                return True
+
+def check_hooks_on_repo(user, repo, errors, hook_user='ros-pull-request-builder', callback_url='http://build.ros.org/ghprbhook/', token=None, strict=False):
+
```
dirk-thomas: A function shouldn't start with an empty line.

test/hook_permissions.py
```diff
@@ -0,0 +1,140 @@
...
+        args.hook_user,
+        args.callback_url,
+        password)
+    for e in errors:
+        print(e)
+    return result
+
+
+if __name__ == '__main__':
+    sys.exit(main() == False)
```
dirk-thomas: A boolean return from a `main` function is rather uncommon. I would suggest a value which can directly be passed to `sys.exit`.

test/test_url_validity.py
```diff
@@ -107,17 +107,37 @@ def check_git_remote_exists(url, version, tags_valid=False):
...
         return None
 
     version = source['version'] if source['version'] else None
     if not check_git_remote_exists(source['url'], version, tags_valid):
-        return ("Could not validate repository with url %s and version %s from"
-                " entry at line '''%s'''" % (source['url'],
-                                             version,
-                                             source['__line__']))
+        errors.append(
+            "Could not validate repository with url %s and version %s from"
```
dirk-thomas: Single quotes.

Same below.

test/test_url_validity.py
```diff
@@ -107,17 +107,37 @@ def check_git_remote_exists(url, version, tags_valid=False):
...
+            " entry at line '''%s'''" % (source['url'],
```
dirk-thomas: Why `'''`?

test/test_url_validity.py
```diff
@@ -107,17 +107,37 @@ def check_git_remote_exists(url, version, tags_valid=False):
...
+                                         version,
+                                         source['__line__']))
+    test_pr = source['test_pull_requests'] if 'test_pull_requests' in source else None
+    if test_pr:
+        o = urlparse(source['url'])
```
dirk-thomas: Maybe a different variable name than `o`.

tfoote:
Updated for comments

test/hook_permissions.py
```diff
@@ -0,0 +1,144 @@
...
+                'and make sure hooks are setup.')
+            if strict:
+                return False
+            else:
+                errors.append('Warning: Push access detected but unable to verify manual hook '
+                    'configuration for repo [ %s ]. Please visit ' % repo.full_name +
+                    'http://wiki.ros.org/buildfarm/Pull%20request%20testing '
+                    'and make sure hooks are setup.')
+                return True
+
```
dirk-thomas: Needs extra newline.

test/hook_permissions.py
```diff
@@ -0,0 +1,144 @@
...
+        args.repo,
+        errors,
+        args.hook_user,
+        args.callback_url,
+        password)
+    if errors:
+        print('Errors detected:')
+    for e in errors:
+        print(e)
+    if result == True:
```
dirk-thomas: Simply: `if result`

dirk-thomas (commented):
All error messages should be printed to `stderr`.

tfoote:
2 more fixups.

dirk-thomas:
> All error messages should be printed to `stderr`.

This comment is still pending.

tfoote:
Fixed stderr output. I missed that in the general comments.

dirk-thomas (approved):
lgtm

PR #16051, opened 2017-10-09
fixing https://github.com/CPFL/Autoware/issues/810

dejanpan:
See https://github.com/CPFL/Autoware/issues/810 for bug report.

An ultimate fix will be to bring all 92 pkgs in autoware repo into the releasable state, that is being done and tracked in https://github.com/CPFL/Autoware/issues/766.

@amc-nu @130s FYI

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -24,7 +24,7 @@ repositories:
       url: https://github.com/CPFL/Autoware-release.git
       version: 1.4.0-1
     source:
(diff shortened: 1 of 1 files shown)


tfoote:
Thanks for paying attention to the notifications and avoiding unnecessary builds and emails. i trimmed the whitespace to make CI happy. It should pass in a few minutes.

dejanpan:
thx @tfoote and sorry for taking so much time.

PR #16109, opened 2017-10-17
Move some packages from python2 to python3

patrickelectric:
The python3 packages are compatible with python2

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -77,7 +77,7 @@ mercurial:
     pip:
       packages: [mercurial]
 paramiko:
-  arch: [python2-paramiko]
+  arch: [python-paramiko]
   debian: [python-paramiko]
   fedora: [python-paramiko]
   freebsd: [py27-paramiko]
@@ -114,7 +114,7 @@ pyper-pip:
     pip:
       packages: [pyper]
 pyqt4-dev-tools:
-  arch: [python2-pyqt4]
+  arch: [python-pyqt4]
   debian: [pyqt4-dev-tools]
   fedora: [PyQt4-devel]
   gentoo: [dev-python/PyQt4]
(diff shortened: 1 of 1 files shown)


mikaelarguedas (approved):
All these packages indeed have source code compatible with both python2 and python3

wjwwood:
I've hesitate to merge this yet, since I wouldn't do this on Ubuntu, but since it is Arch it might/should be the right thing to do. Anyone else willing to give yet another +1 (adding to @mikaelarguedas's) before merging this?

tfoote:
Maybe another Arch user could review? @zootboy @hauptmech @de-vri-es @nuclearsandwich @bchretien

hauptmech:
Even though rosdep is independent, I think it's fair to say that most will be using it with ROS. So python2 will be the python being used. Adding python3 as a requirement by using the python-* packages is, in my opinion, unnecessary and potentially a burden for resource constrained systems.

de-vri-es:
To be honest it doesn't matter much to me. I maintain a binary pacman repository with ROS packages for myself and my co-workers. For that repository I've chosen to use python3 for *everything*, and it's not giving us much trouble. We don't write a lot of python though, so our testing is mainly limited to the command line tools.

Python 3 is almost 9 years old by now, so I do find it sad that python 2 is still so wide-spread. In that sense I'm for anything that tries to push python 3 in favor of python 2. But I won't notice much of these changes myself if they're merged since I'm already using python 3 anyway.

patrickelectric:
Working with kinetic and lunar "desktop".

hauptmech:
With 30% of the downloads being indigo (http://wiki.ros.org/Metrics) we'd probably want that working as well.

mikaelarguedas:
Coming back to this. I don't mind personally going one way or the other. I originally approved as it looks like we take python3 packages for platforms that use python3 as the default (e.g. gentoo).

If we think it should be python2 for Arch to match the officially targeted version, I'd suggest to close this. Otherwise we can merge it.

Note: the upcoming ROS release will still officially target Python2.7 and we are currently reviewing proposal for officially switching to python3-only support in future ROS distributions (https://github.com/ros-infrastructure/rep/pull/149)

clalancette:
This has been open for a very long time now.  It looks to me like we ended up going a different direction here, where we have specific `python3-` keys for python3 packages.  Because of that, I'm going to close this out.  Please feel free to re-open if you think this was closed in error.

PR #16122, opened 2017-10-18
nerian_sp1 node is now end-of-life and has been replaced by nerian_stereo

nerian-vision:


dirk-thomas:
Maybe adding a `status_description` with the information of the new package would be helpful?

Also I am not sure if it makes sense to update the status for EOL ROS distros? Their status pages etc. won't reflect this since we don't process them anymore.

nerian-vision:
Thanks for pointing out. I added a status description and reverted my changes to the EOL distros.

mikaelarguedas (approved):
lgtm, thanks @nerian-vision for iterating

PR #16140, opened 2017-10-23
Add rendering use flag to vtk

concavegit:
Enables vtk libraries without which opencv would fail to build.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2925,7 +2925,7 @@ repositories:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ros-gbp/household_objects_database_msgs-release.git
       version: 0.1.2-0
-   source:
+    source:
       type: git
       url: https://github.com/ros-interactive-manipulation/household_objects_database_msgs.git
       version: hydro-devel
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2925,7 +2925,7 @@ repositories:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ros-gbp/household_objects_database_msgs-release.git
       version: 0.1.2-0
-    source:
+   rospack source:
       type: git
       url: https://github.com/ros-interactive-manipulation/household_objects_database_msgs.git
       version: hydro-devel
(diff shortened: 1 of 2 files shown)


mikaelarguedas (commented):
lgtm
Not sure why we still target qt4 for this though... The libvtk key gets [qt5 for gentoo](https://github.com/concavegit/rosdistro/blob/8ba3e83294781bed67f2f1c84ef4c68d810b5b55/rosdep/base.yaml#L3062) so I'm not sure why the dev version of the key should get qt4. maybe @tfoote or @allenh1 have an idea (relevant comment https://github.com/ros/rosdistro/pull/15208#discussion_r127595447)

concavegit:
Yeah, I was wondering that as well. I decided that removing the use flag was probably out of my scope.

mikaelarguedas:
>  I decided that removing the use flag was probably out of my scope.

:+1: that's totally fine I just wanted to raise the question here but I'm totally fine having this merged as is

tfoote:
It's fine by me. @allenh1 thoughts?

mikaelarguedas:
@concavegit pushed to the wrong branch ? please open a separate PR for separate issues.

In the case of ROS packages that you don't maintain yourself, if you want a new release please open an issue on the upstream repository so that the maintainer makes a new release when appropriate

concavegit:
I attempted to fix a different issue, but I forgot that pull requests merge like this. Sorry!

Anyways, the version of rospack kinetic uses is incompatible with boost 1.65, but ros/rospack#80 fixed that in a newer version, so I bumped the version up.

concavegit:
I reverted the extraneous commits just now. 'version bump' was a poor choice of words on my part, as the version is already out. Would I still have to open an issue in ros/rospack?

allenh1 (approved):
lgtm, thanks!

allenh1:
> Would I still have to open an issue in ros/rospack?

@concavegit I'll update the ebuilds to have the newer version and tag you in the commit.

allenh1:
@mikaelarguedas I think this should be qt5 personally, but the Debian dependencies are all qt4. Should those _also_ be qt5?

mikaelarguedas:
> I think this should be qt5 personally, but the Debian dependencies are all qt4

I'm not sure I understand, can you clarify what you mean ?
From my understanding, as soon as a distro supports vtk6 (using qt5) we use it. Jessie and wheezy don't have vtk6 so we use vtk5 (that uses qt4) but the more recent versions (Stretch and Buster) use vtk6 + qt5. Same thing for the ubuntu rules, Wily doesnt have vtk6 so we started using vtk6 + qt5 as of Xenial.

mikaelarguedas (approved):
lgtm

PR #15650, opened 2017-07-31
Adding muse_bldc_motor_drive to documentation index for distro

mariiiiiia:
I'd like muse_bldc_motor_drive to be indexed and documented on ros.org.

PR #16143, opened 2017-10-24
Add tracetools package using source

iluetkeb:


tfoote:
trying to retrigger travis

iluetkeb:
@tfoote Is it normal that the check is taking so long?

mikaelarguedas:
It looks like this has been flagged as an abuse by travis (https://travis-ci.org/ros/rosdistro/requests).

@tfoote I see that you "cleared" the account in #15650, can we do the same here ?

indigo/distribution.yaml
```diff
@@ -15062,6 +15062,21 @@ repositories:
...
     status: developed
+  tracetools:
+    doc:
+      type: git
+      url: https://github.com/bosch-robotics-cr/tracetools.git
+      version: devel
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/bosch-robotics-cr/tracetools.git
```
dirk-thomas: You should not use the same repository for the release.

tfoote: Yes, that will make a mess of your source repo.

tfoote:
I've requested to clear the abuse flag with Travis Support

iluetkeb:
All changes are flagged as an abuse by Travis CI. Does anybody have an idea what could cause that?

mikaelarguedas:
It looks like your user and organization never logged into travis so they flagged it as abusive automatically. Is it possible for you to login into travis and then push an empty commit in here ? That should solve the issue.

Otherwise, can you contact travis support (support@travis-ci.com) to get your account/organization cleared ?

PR #16144, opened 2017-10-24
xpp: 0.1.0-0 in 'indigo/distribution.yaml' [bloom]

awinkler:
Increasing version of package(s) in repository `xpp` to `0.1.0-0`:

- upstream repository: https://github.com/leggedrobotics/xpp.git
- release repository: https://github.com/leggedrobotics/xpp-release.git
- bloom version: `0.6.1`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -16407,6 +16407,28 @@ repositories:
       url: https://github.com/wavelab/ximea_camera.git
       version: devel-indigo
     status: maintained
+  xpp:
+    release:
+      packages:
+      - biped_description
+      - hyqb_description
+      - monoped_description
+      - quadrotor_description
+      - xpp_examples
+      - xpp_msgs
+      - xpp_ros_conversions
+      - xpp_states
+      - xpp_vis
+      - xpp_vis_hyq
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/leggedrobotics/xpp-release.git
+      version: 0.1.0-0
+    source:
+      type: git
+      url: https://github.com/leggedrobotics/xpp.git
+      version: master
+    status: developed
   xsens_awinda:
     doc:
       type: git


indigo/distribution.yaml
```diff
@@ -16407,6 +16407,28 @@ repositories:
       url: https://github.com/wavelab/ximea_camera.git
       version: devel-indigo
     status: maintained
+  xpp:
+    release:
+      packages:
+      - biped_description
+      - hyqb_description
+      - monoped_description
+      - quadrotor_description
```
dirk-thomas: These packages names sound very generic. Maybe a `xpp_` prefix would make sense for them too? I will let @tfoote comment on it.

awinkler: The idea was to leave out the xpp from these packages, as they are the standard URDF description packages and don't depend on any of the xpp packages. I was assuming people will install the entire repo, but of course if these packages show up individually in a list, people won't know with what package to use it with. Unsure.

mikaelarguedas: The format of the generated .deb (`ros-indigo-PACKAGE_NAME`) doesnt provide any repository information, so I am not aware of an easy way to "install the entire repo". Except if you provide a metapackage for all the packages in your repository.
In general we recommend to have a consistent prefix to allow people to easily list relevant packages with apt: `apt-get install ros-indigo-xpp_<TAB><TAB>`

tfoote: Our naming conventions are outlined here:

Whether or not they are used with xpp the robot descriptions should be named. Claiming 'quadrotor_description` suggests that this is "the" quadrotor description and there's no need to explain which quadrotor. Looking into the "quadrotor" it doesn't appear to even [have rotors](https://github.com/leggedrobotics/xpp/blob/master/robots/quadrotor_description/urdf/quadrotor.urdf). And there's already examples of things like the hector_quadrotor_description: https://github.com/tu-darmstadt-ros-pkg/hector_quadrotor/tree/kinetic-devel/hector_quadrotor_description showing how to provide a proper name instead claiming the whole namespace. Is that something that is actively used?

The package descriptions clearly state that they are based on hyq. "The URDF file for a one-legged hopper with one HyQ leg." So the packages should be hyq_monoped_description as it's specific to the hyq. Same with the biped and quadroped. And hyq_biped_description and hyq_description since it's a full hyq.

awinkler: Very good points, thanks. I will take out the quadrotor package and focus on visualizing legged robots for now. And adding "xpp_hyq_*" to the other description packages also makes sense. This is the first time I'm trying to add a pkg to ROS, do I know update the github repo, run bloom again and create a new pull request? What's the recommended procedure here? Thank you for your help.

awinkler: Okey, I actually now only have "xpp_hyq" and "xpp_quadrotor" remaining, that encapsulate all the robot specific functionality for these robots to be used within the xpp framework. I also added an xpp-metapackage for grouping. 
https://github.com/leggedrobotics/xpp
http://wiki.ros.org/xpp

I would now create another github release, incrementing all package.xmls and run through the whole bloom procedure again to generate another PR? 
Thanks!

dirk-thomas: > I would now create another github release, incrementing all package.xmls

Yes, `catkin_prepare_release` can do that for you.

> and run through the whole bloom procedure again to generate another PR?

Correct. The second time it should be less questions. After you pass the release repo URL it will find the information from the last run and only ask for confirmation.

dirk-thomas:
@awinkler please close obsolete PR in the future.

PR #16145, opened 2017-10-24
[gentoo] use qt5 use flag for vtk-dev

mikaelarguedas:
Follow up of https://github.com/ros/rosdistro/pull/16140#pullrequestreview-71367010

gentoo uses vtk with the qt5 use flag for the key [`libvtk`](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L3062). So this uses qt5 for `libvtk-qt` as well.

@allenh1 @Alessandro-Barbieri @concavegit FYI

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3108,7 +3108,7 @@ libvtk-qt:
     stretch: [libvtk6-qt-dev]
     wheezy: [libvtk5-qt4-dev]
   fedora: [vtk-qt]
-  gentoo: ['sci-libs/vtk[qt4,rendering]'] #qt5 is available too
+  gentoo: ['sci-libs/vtk[qt5,rendering]']
   opensuse: [vtk-qt]
   slackware: [VTK]
   ubuntu:


allenh1:
I'm very much in favor of this change, I ran into a lot of collisions after adding the qt4 restriction to vtk (restricts it to version 6, and a lot of things depend on 7).

Alessandro-Barbieri:
These useflags are required by some ROS packages?

mikaelarguedas:
I think the general question is what ROS Distribution do we expect to support on gentoo. So far there is a single gentoo rule per key in rosdep. In the case of qt, Indigo targets qt4 and Kinetic and Lunar target qt5 ([see REP3](https://github.com/ros-infrastructure/rep/blob/master/rep-0003.rst) for version targeted for each distro), so any rule that selects explicitly one version can only support a subset of all active ROS distros.
Looking at other rules it looks like gentoo is in general trying to match the latest. In this case it would make sense to target dependency versions for the latest ROS Distributions.

Alessandro-Barbieri:
Oh, I see that a qt useflag was already present, it's ok for me.

tfoote (approved):

PR #16157, opened 2017-10-25
readd pyside2 and shiboken to python qt bindings

mikaelarguedas:
Now that the packages have been imported for zesty and artful: https://github.com/ros-infrastructure/ros_bootstrap_dependencies/issues/5#issuecomment-338821581

PR #16216, opened 2017-10-30
remove pyside2 from zesty and artful again

dirk-thomas:
Partially revert #16157.

The upstream packages on these platforms only support Python 3.


mikaelarguedas (approved):
lgtm

PR #16229, opened 2017-11-01
Contribute rosdep rule for libgpgme11-dev

jwon02:
This is required for encryption/decryption support for `Bag` class: https://github.com/ros/ros_comm/pull/1206.

Web links to packages:
Arch: https://www.archlinux.org/packages/core/x86_64/gpgme/
Debian: https://packages.debian.org/search?keywords=libgpgme11-dev
Fedora: https://apps.fedoraproject.org/packages/gpgme-devel
Gentoo: https://packages.gentoo.org/packages/app-crypt/gpgme
Ubuntu: https://packages.ubuntu.com/search?keywords=libgpgme11-dev


mikaelarguedas (dismissed):
`libgpgme11-dev` has been discontinued in ubuntu as of yakkety in favor of `libgpgme-dev` so we need to explicitly call out Ubuntu distributions, everything up to xenial should use `libgpgme11-dev`  and yakkety and later will use `libgpgme-dev`.
Same for Debian: jessie and before will have `libgpgme11-dev` and stretch and later will use `libgpgme-dev`

jwon02:
Thank you @mikaelarguedas.  Updated the PR as suggested.

tfoote:
Thanks @jwon02 for the quick update. Could you add ubuntu artful and bionic too since we're looking forward toward melodic already. (Note keep alphabetical order so they'll be at the top of the list.)

jwon02:
Thank you @tfoote.  Added artful and bionic.  I also removed digits `11` from the key to have `libgpgme-dev`.

mikaelarguedas (approved):
lgtm, thanks @jwon02 for the quick iteration

PR #16231, opened 2017-11-01
[rosdep] Add optipng.

130s:
For Fedora and Gentoo I can't find how we want to specify per distro so just left one line per each.

Debian: https://packages.debian.org/search?keywords=optipng
Fedora: https://apps.fedoraproject.org/packages/optipng
Gentoo: https://packages.gentoo.org/packages/media-gfx/optipng
Ubuntu: https://packages.ubuntu.com/trusty/optipng

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3665,6 +3665,11 @@ openssl:
   fedora: [openssl]
   gentoo: [dev-libs/openssl]
   ubuntu: [openssl]
+optipng:
+  debian: [optipng]
+  fedora: [optipng]
+  gentoo: [optipng]
+  ubuntu: [optipng]
 osm2pgsql:
   debian: [osm2pgsql]
   fedora: [osm2pgsql]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3668,7 +3668,7 @@ openssl:
 optipng:
   debian: [optipng]
   fedora: [optipng]
-  gentoo: [optipng]
+  gentoo: [media-gfx/optipng]
   ubuntu: [optipng]
 osm2pgsql:
   debian: [osm2pgsql]


rosdep/base.yaml
```diff
@@ -3665,6 +3665,19 @@ openssl:
   fedora: [openssl]
   gentoo: [dev-libs/openssl]
   ubuntu: [openssl]
+optipng:
+  debian:
```
tfoote: Since these are all the same we shouldn't call out each version.

Note we also have a policy not to call out `sid` if we do have to call out versions.

rosdep/base.yaml
```diff
@@ -3665,6 +3665,19 @@ openssl:
...
+    buster: [optipng]
+    jessie: [optipng]
+    sid: [optipng]
+    stretch: [optipng]
+    wheezy: [optipng]
+  fedora: [optipng]
+  gentoo: [optipng]
+  ubuntu:
```
tfoote: Also all the same so should be collapsed, especially since this listing doesn't cover all our supported distros.

tfoote (changes requested):
Please collapse the debian and ubuntu sections for a more compact listing.

mikaelarguedas:
Can you please collapse the ubuntu and debian rules to be single line ones? This will allow to not have to add existing distros (like yakkety artful and bionic) or future ones by hand

130s:
Thanks, I didn't really know that is the policy. Now collapsed I think.

mikaelarguedas:
For future references, guidelines for rosdep rules are available [here](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules)

mikaelarguedas (approved):
thanks @130s for the quick iteration and for providing all the links!

PR #16232, opened 2017-11-02
drop iproute2 for old ubuntu distro

furushchev:
The package `iproute2` is not found on ubuntu raring and earlier.
This pull request drops `iproute2` on old ubuntu distros which has no key for this package.


mikaelarguedas:
@furushchev this is removing a recently added key, can you please rebase the branch?

mikaelarguedas:
the iproute2 key was added way after Hydro was dropped and there shouldn't be any package released in hydro depending on it.
What is the use case you are trying to cover?
If you have a package that declare this as a dependency on a platform older than saucy, this PR will not fix it because the rule will not exist anymore.
A proper solution would be to declare that rule as empty for the platforms that dont have this package `precise: []`.

furushchev:
@mikaelarguedas Thank you for comment. I see, we are currently releasing the same packages for all distros (without dividing branches) and our test on hydro env stopped on `rosdep install` because of this error.
https://github.com/jsk-ros-pkg/jsk_common/pull/1570

But it seems this approach for fix was not so good according to your comment...
If you'd like to merge this, it's fine, otherwise just close this issue please.

mikaelarguedas:
Can you please add artful and bionic to the list (in alphabetical order)? 

> If you'd like to merge this, it's fine, otherwise just close this issue please

I'm fine with merging this (now that the empty rules have been added) if this covers your use case. 
I'll leave it up to another rosdistro maintainer to comment on the approach / close or merge accordingly

wjwwood:
+1 with @mikaelarguedas' request to have the new ubuntu distros represented.

furushchev:
@mikaelarguedas @wjwwood Got it, updated!

mikaelarguedas (dismissed):
lgtm once ci green, thanks @furushchev for the quick iteration

tfoote (changes requested):
The empty list `[]` implies that the dependency is already provided (for example it's built-in) which is not the case for the older distros. We should not encode that this dependency is resolved without installing anything if in fact it's not available.

This cannot be a required dependency if the package was not provided on those targets, and this is embedding a specific use cases optional resolution into a generic rosdep key.

The "correct" thing for this would be to have no entries for those distros as that dependency is not resolvable. We've generally held that having the generic keys that fail to install on no longer supported platforms is ok for the sake of compactness. as the rule currently is defined.

wjwwood (changes requested):
> This cannot be a required dependency if the package was not provided on those targets, and this is embedding a specific use cases optional resolution into a generic rosdep key.

Actually that's true, I agree with @tfoote. The `[]` implies "implicitly installed" which is not right in this case.

mikaelarguedas:
ok my bad, I wasnt sure that's why I wanted another set of eyes on this.

If I read @tfoote's comment properly it would actuallly be preferred to close this PR given that @furushchev's team will stop supporting hydro regardless of this getting merged. And not merging this will allow the definition to stay compact and maintainable more easily in the future. Is that correct?

tfoote:
+1 for closing as leaving as is to keep the compact format. If someone wants to skip this during development they can use the `--skip-key` option for rosdep.

mikaelarguedas:
closing then, sorry @furushchev for the unnecessary iteration

furushchev:
@mikaelarguedas Completely no problem!
Thank you all for comments!! :+1:

PR #16235, opened 2017-11-02
Clean PR for tracetools

iluetkeb:
Old one was broken due to bad merge.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -15095,6 +15095,16 @@ repositories:
       url: https://bitbucket.org/traclabs/trac_ik.git
       version: master
     status: developed
+  tracetools:
+    doc:
+      type: git
+      url: https://github.com/bosch-robotics-cr/tracetools.git
+      version: devel
+    source:
+      type: git
+      url: https://github.com/bosch-robotics-cr/tracetools.git
+      version: devel
+    status: developed
   transform_graph:
     doc:
       type: git
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -15100,6 +15100,11 @@ repositories:
       type: git
       url: https://github.com/bosch-robotics-cr/tracetools.git
       version: devel
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/bosch-robotics-cr/tracetools.git
+      version: 0.1.0
     source:
       type: git
       url: https://github.com/bosch-robotics-cr/tracetools.git
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -15062,6 +15062,16 @@ repositories:
       url: https://bitbucket.org/traclabs/trac_ik.git
       version: master
     status: developed
+  tracetools:
+    doc:
+      type: git
+      url: https://github.com/bosch-robotics-cr/tracetools.git
+      version: devel
+    source:
+      type: git
+      url: https://github.com/bosch-robotics-cr/tracetools.git
+      version: devel
+    status: developed
   transform_graph:
     doc:
       type: git
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -15103,7 +15103,7 @@ repositories:
     release:
       tags:
         release: release/indigo/{package}/{version}
-      url: https://github.com/bosch-robotics-cr/tracetools.git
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
This fails because there's no package.xml on the release repository. Looking at it, it looks like the release repository is completely empty. Did you forget to run bloom-release ? (Useful links http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease, http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty)

iluetkeb:
I have not done a release, yet. I was setting up to do one, when bloom complained that it could not find a release repo in the rosdistro.

You know, this would all go a lot easier for first-timers if some things were classified as warnings instead of errors, because of the cyclic dependencies between bloom and rosdistro.

mikaelarguedas:
> I have not done a release, yet. I was setting up to do one, when bloom complained that it could not find a release repo in the rosdistro.

Bloom looks at rosdistro to get as much information as possible to avoid users to have to type them in. It should indeed have classified the absence of a release repository as a warning and have offered you to provide a release repository manually. Maybe the message from bloom was not clear and if so we'd definitely consider a PR to make this message more explicit.

On the rosdistro side, I think it makes sense that an invalid or empty release repository is flagged as an error because if it gets merged as is the buildfarm will try to build packages from that invalid release repository and fail.

wjwwood:
Unless I'm mistaken the message includes this line:

> You can continue the release process by manually specifying the location of the RELEASE repository.

(bloom's capitalization not mine)

So, I think it should be prompting you to enter the address of the release repository, not add it here. If there is a confusing message please copy-paste the output here or suggest an improvement on the bloom issue tracker as @mikaelarguedas said.

iluetkeb:
That's the way I'm doing it now. It's balked at the release tag (which, somehow, it expects it to match to version directly, instead of being release/<distro>/<version>, as I had thought).

Anyway, gotta run to another thing now, so it's gonna be a few hours. Don't wait for me ;-)

btw, this whole process is a bit icky, but some of it is because it's so configurable. Maybe the whole rails-idea of "convention over configuration" would be something to go for. I'll think about it.

wjwwood:
> It's balked at the release tag (which, somehow, it expects it to match to version directly, instead of being release//, as I had thought).

You can configure that, see:

https://github.com/ros-infrastructure/bloom/blob/d3349cc63837f0b71e264d42876d0a2728277d4e/bloom/config.py#L94-L104

Which should have been printed to you during configuration.

> btw, this whole process is a bit icky, but some of it is because it's so configurable. Maybe the whole rails-idea of "convention over configuration" would be something to go for. I'll think about it.

For the 99% use case (an actual ROS package), all of the settings are default (other than which repository and what ros distro). The problem is that your repository is unconventional, i.e. using something other than just the version number as the release tag, which is the convention for ROS repositories.

iluetkeb:
Well, I had *thought* that release/.../ *was* the default, because some other tool expected that and didn't get it. I believe it was the nosetests for rosdistro. Whatever ;-)

iluetkeb:
Ah... and the nosetests complained, because I took my cue from the "turtlebot" repository in the rosdistro, which, apparently, is also not-standard

tfoote:
The turtlebot is standard where the tags are all just the version numbers: https://github.com/turtlebot/turtlebot/releases

The only place I'm familiar with the release/distro/package/version is in the release repos that bloom generates like https://github.com/turtlebot-release/turtlebot-release/releases?after=rpm%2Fros-indigo-turtlebot-bringup-2.3.14-0_heisenbug The tags defined here: https://github.com/ros/rosdistro/blob/master/kinetic/distribution.yaml#L9636 are the tags to which bloom will output package specific artifacts inside the release repository.

iluetkeb:
Alright, looking good here. Thanks for the help @tfoote @mikaelarguedas !

mikaelarguedas:
Thanks for iterating. You will need to run bloom-release on kinetic as well, currently the release repository has only the ungido branches

iluetkeb:
@mikaelarguedas In https://github.com/bosch-robotics-cr/tracetools-release I see branches for kinetic.

mikaelarguedas:
You're right, looks like github on smartphone hides stuff from me. Looks good then!

mikaelarguedas (approved):

mikaelarguedas:
Note: The buildfarm is pretty busy right now so it may take a while for these debs to be created

PR #16253, opened 2017-11-03
remove geographiclib from gentoo

Alessandro-Barbieri:
only available in the pchrist overlay as an ebuild with a banned (or soon to be) EAPI
hope to see in main tree https://bugs.gentoo.org/636338

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1680,7 +1680,6 @@ libgeographiclib-dev:
   debian:
     wheezy: [libgeographiclib-dev]
   fedora: [geographiclib]
-  gentoo: [sci-geosciences/geographiclib]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
For reference this is the PR @allenh1 is referring to https://github.com/ros/ros-overlay/pull/269

allenh1:
It _would_ be super great, however, if you could add the definition for `geographiclib` under the same name (since the ebuild is there now). It's blocking mavros on Lunar.

mikaelarguedas:
@Alessandro-Barbieri @allenh1 If there is no objection, I'm going to close this as it appears not to be necessary anymore.
Feel free to open a different PR for adding the `libgeographiclib` gentoo rule

PR #16266, opened 2017-11-03
Revert "bond_core: 1.8.1-0 in 'kinetic/distribution.yaml' [bloom]"

mikaelarguedas:
Reverts ros/rosdistro#16244

Thought that https://github.com/ros/ros_comm/pull/1205 was already merge, reverting until steady timer feature is available


dirk-thomas:
I would say don't worry about it. Once it syncs to trying and the PR passes I will release it and it should be fine.

mikaelarguedas:
My concern was that this may prevent enough packages from building that there won't be enough packages to sync to testing. I'll merge this just to be sure and synchronize my release with yours

PR #16301, opened 2017-11-07
Adding eigenpy to documentation index for indigo

nim65s:
Eigenpy gives bindings between numpy and eigen using boost::python.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2901,6 +2901,16 @@ repositories:
       url: https://github.com/ros/eigen_stl_containers.git
       version: master
     status: maintained
+  eigenpy:
+    doc:
+      type: git
+      url: https://github.com/stack-of-tasks/eigenpy.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/nim65s/eigenpy.git
+      version: ros
+    status: maintained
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -2901,6 +2901,16 @@ repositories:
...
       version: master
     status: maintained
+  eigenpy:
+    doc:
+      type: git
+      url: https://github.com/stack-of-tasks/eigenpy.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/nim65s/eigenpy.git
```
mikaelarguedas: Is it on purpose that the documentation comes from a repo and the source code comes from a fork of that repo?

nim65s: Yes, I need some time to understand the ROS ecosystem and to adapt our code for catkin & bloom.
So if possible I'd like to use my fork as a starting point while it's not fully operationnal, before pushing the modifications to the upstream.

But I guess I should also use my fork for the doc, as the upstream doesn't have a `package.xml` file for now.

mikaelarguedas (approved):
sounds good, thanks for iterating

nim65s:
Thanks :)

PR #16302, opened 2017-11-07
find_object_2d: 0.6.2-0 in 'lunar/distribution.yaml' [bloom]

matlabbe:
Increasing version of package(s) in repository `find_object_2d` to `0.6.2-0`:

- upstream repository: https://github.com/introlab/find-object.git
- release repository: https://github.com/introlab/find_object_2d-release.git
- bloom version: `0.6.1`

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -635,6 +635,21 @@ repositories:
       url: https://github.com/ros/filters.git
       version: hydro-devel
     status: maintained
+  find_object_2d:
+    doc:
+      type: git
+      url: https://github.com/introlab/find-object.git
+      version: lunar-devel
+    release:
+      tags:
+        release: release/lunar/{package}/{version}
+      url: https://github.com/introlab/find_object_2d-release.git
+      version: 0.6.2-0
+    source:
+      type: git
+      url: https://github.com/introlab/find-object.git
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@matlabbe FYI: If you release from a new branch you will need to bump the minor version number as only one version of the code can exist on the upstream repository for a given tag. Otherwise catkin_prepare_release will fail next time you prepare a release for both kinetic and lunar with the same version number.
(see [ros_tutorials](http://repositories.ros.org/status_page/compare_indigo_jade_kinetic_lunar.html?q=rospy_tutorial) for an example of package releasing from a different branch every ROS distro and bumping the version number accordingly)

matlabbe:
Thx for the warning. I use the same version of the standalone (no ROS) version of [Find-Object](https://introlab.github.io/find-object/) to be easier to track which version is actually used in ROS (if I have issues opened on a particular version of Find-Object, standalone or ROS).

I do the same with [RTAB-Map](http://introlab.github.io/rtabmap/), the ros package uses the version of the standalone version to track the issues instead of the ROS distros. 

I didn't have problems in the past with bloom doing so though, I'll see the next release if it is too problematic. If it fails, I'll know why!

mikaelarguedas (approved):
Sounds good

PR #16309, opened 2017-11-08
Add ruamel.yaml in 'rosdep/python.yaml'

barryridge:
[ruamel.yaml](https://pypi.python.org/pypi/ruamel.yaml) provides much improved YAML parsing over pyyaml.

Use case: it is becoming close to a necessity for continuing work on a new ROS package I've been working on for meta-scripting, templating and code generation of state machine prototypes called [SMACHA](https://github.com/ReconCell/smacha).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2836,7 +2836,16 @@ python-rrdtool:
     saucy: [python-rrdtool]
     trusty: [python-rrdtool]
 python-ruamel.yaml:
-  ubuntu: [python-ruamel.yaml]
+  debian:
+    buster: [python-ruamel.yaml]
+    jessie: [python-ruamel.yaml]
+    stretch: [python-ruamel.yaml]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
This package doesnt seem to be available to ubuntu distributions before xenial. Given that we still have an activ ROS distribution (indigo) targeting trusty, can you change the rule to enumerate ubuntu distributions xenial and later?
Can you add debian rules as well ? We require at least debian and ubuntu rules to merge a new key, providing fedora as well is prefered.

barryridge:
Done.

I have also enumerated the Debian rules, since there is no package for Wheezy.

There is no Python3 package for Fedora, but I have included a rule for the Python2 package.

wjwwood (approved):

tfoote (approved):

mikaelarguedas (approved):
Lgtm, thanks @barryridge for iterating

PR #16288, opened 2017-11-06
Add python-couchdb for additional ubuntu distros.

achim-k:
See also <https://packages.ubuntu.com/search?keywords=python-couchdb>

Could be simplified to 
```yaml
ubuntu: [python-couchdb]
```
?

PR #16312, opened 2017-11-08
add pyxhook and sphinxcontrib-bibtex for osx as well

mikaelarguedas:
follow-up of #16288 and #16305

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2483,6 +2483,9 @@ python-pyxhook-pip:
   fedora:
     pip:
       packages: [pyxhook]
+  osx:
+    pip:
+      packages: [pyxhook]
   ubuntu:
     pip:
       packages: [pyxhook]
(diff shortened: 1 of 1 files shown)


wjwwood (approved):
+1 for adding them, but note that it's not strictly required unless people using the key want to support macOS.

mikaelarguedas:
> but note that it's not strictly required unless people using the key want to support macOS.

:+1: I just figured that it's easy enough when it's pip to just add the entry for the 4 distros to avoid needing a follow up PR to add them when someone face the issue. Is it preferred to make this opt-in and for mac users to submit follow-up PRs if they want to ?

wjwwood:
I would say opt-in is fine. If we're going to use it in desktop-full then I'd say we should require it for mac, otherwise opt-in is fine imo.

mikaelarguedas:
Sounds good, thanks for clarifying

PR #16326, opened 2017-11-10
Disable automatic builds for ensenso_driver

saierd:
The package needs an external library that is currently not available in rosdep, so I'd like to disable the builds for now.


mikaelarguedas (approved):
Sounds good, thanks @saierd

PR #16331, opened 2017-11-10
Add pysnmp

peci1:


peci1:
@kubelvla so that you're happy :)

mikaelarguedas:
Thanks @peci1  for contributing to rosdep, however please have a look at the Contribution guildelines (https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules), as there are a few information missing from this PR:

> For convenience in reviewing please link to the web listings of packages such as on http://packages.ubuntu.com, http://packages.debian.org, and https://admin.fedoraproject.org/pkgdb/packages/ or if a pip package pypi.python.org.

> Please also briefly describe the package being added and what use case you want to use it for. It's valuable to have a record of the packages as submitted and their intended purpose for clarity in the future so that if there's a conflict there's information to fall back on instead of speculation about the original use cases.

> Native packages are strongly preferred. (They are required for packaging and have upgrade and conflict tracking.)
> If a package is only available via pip use the -pip extension.

Looks like this package exist in ubuntu and debian https://packages.ubuntu.com/search?keywords=python-pysnmp&searchon=names, do you need the one from pip ? or would the upstream one be ok?

tfoote:
Since there is a stable one upstream we should use that not the pip version. It's available trusty through bionic. https://packages.ubuntu.com/artful/python-pysnmp4 Pip installations are not easily introspectable removable and will potentially side effect anyone else who has the debian dependency.

peci1:
Okay, I'm sorry for missing the contribution guidelines, I just remembered that submitting PRs to rosdep is quite easy (so now I now it's not super-easy, but still reasonably easy).

Thanks for your comments.

We plan to use pysnmp to read out signal strength values from Ubiquiti Bullet wireless routers.

mikaelarguedas:
Debian: https://packages.debian.org/search?keywords=python-pysnmp4&searchon=names&suite=all&section=all
Gentoo: https://packages.gentoo.org/packages/dev-python/pysnmp
Fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/pysnmp/
Ubuntu: https://packages.ubuntu.com/search?keywords=python-pysnmp4&searchon=names

mikaelarguedas (approved):
lgtm

PR #16335, opened 2017-11-11
Added mkdocs.

peci1:
A library for generating end-user documentation and its skins.

https://pypi.python.org/pypi/mkdocs
https://pypi.python.org/pypi/mkdocs-bootstrap
https://pypi.python.org/pypi/mkdocs-bootswatch
https://packages.ubuntu.com/search?keywords=mkdocs&searchon=names&suite=all&section=all
https://admin.fedoraproject.org/pkgdb/package/rpms/mkdocs
https://admin.fedoraproject.org/pkgdb/package/rpms/mkdocs-bootstrap
https://admin.fedoraproject.org/pkgdb/package/rpms/mkdocs-bootswatch/
https://packages.debian.org/search?searchon=names&keywords=mkdocs
https://packages.gentoo.org/packages/dev-python/mkdocs
https://packages.gentoo.org/packages/dev-python/mkdocs-bootstrap
https://packages.gentoo.org/packages/dev-python/mkdocs-bootswatch

PR #16342, opened 2017-11-11
Merge python in base

mikaelarguedas:
This came up yesterday when reviewing #16335 that introduced a python package in base.yaml (note that the name of the package is not prefixed with `python-`).
Discussing with @tfoote we thought that the current split is pretty inconsistent (some files are splitting platforms, some package managers and some are splitting content) and not necessarily justified anymore (my understanding is that this was mostly for historical reason and maintainership split).

So this PR aims to merge python.yaml into base.yaml in order to have all the content in a single place.
@ros/rosdistro-maintainers I'd like to have feedback on the idea to know if this makes sense or if I'm missing context. Thanks!


wjwwood (approved):

dirk-thomas:
Since both files (`base.yaml` as well as `python.yaml`) are already huge 4.8k vs 4k lines I would rather keep them separate in order to avoid a single ~9k lines long file. At least I don't see a strong enough reason why they should be merged. We have been using this splitting for a long time. Maybe a simple (re-)clarification what should go where is sufficient.

mikaelarguedas:
> Maybe a simple (re-)clarification what should go where is sufficient.

I think one question that came up is that (given that some python packages don;t get prepended with `python-` when getting released for the first time in Debian/Ubuntu), where should these packages go ?
should they go in base.yaml or python.yaml?

If we say they should go in `python.yaml` we should make sure that reviewers confirm that keys added to `base.yaml` are non-python packages.

The split seemed pretty artificial to me as I didnt see a strong reason to have them separate. And keeping the content consistent across these 2 files seemed to me to be more overhead without a strong rationale. If the rationale is that we don't want to have a very long file, that's a fair point. Is there any other reason for keeping them separated?

tfoote (approved):
I think it makes sense to merge them, right now it's a kind of disjoint grouping with semi-ambiguous classification. 

Although the files are getting long, I think it would be easier to search in one list rather than 2. And I'm not aware of any editors that we are anywhere near the limit of being able to open the file. Once you're beyond one screen height you're going to be using searches and jumps and in my opinion doing that within one file is actually easier.

The many files for rosdep inherits from the distant past when we were approaching it as a distributed database that different packages contributed keys to. But through experience we've changed it to be just a central location. Keeping it split into 2 based on the type of package doesn't seem to make much sense. If we are really worried about the file size it would seem to make much more sense to split it on something simpler like the alphabet, A-M and-N-Z. And if we picked a max desired file size we could split it into more shards in the future. But until we reach a point of something not working, I don't see much point in keeping it split apart. Having the two files split as they are now is a common cause of misunderstandings for new users/contributors.

mikaelarguedas:
@wjwwood would you like to provide feedback on this ?

wjwwood:
I don't mind one way or the other. Having a single source makes it easier to search, but having a larger file might be frustrating (especially being reminded of early 2000's internet speeds this morning on our backup network).

dirk-thomas:

mikaelarguedas:
Waiting for a decision with the entire ROS team during next rosdep discussion as we didnt reach consensus here.

PR #15000, opened 2017-05-17
Use gazebo7 instead of gazebo8 in Lunar

j-rivero:
Revert pull request #14564 and use gazebo7 for ROS Lunar. Details: https://discourse.ros.org/t/lunar-to-target-gazebo-7/1840/1

PR #16363, opened 2017-11-14
Add gazebo8 rosdeps for ubuntu

paulbovbel:


mikaelarguedas:
I dont think gazebo8 is part of any upstream repository (Ubuntu or Debian) at the moment.
Maybe @j-rivero can give some insight as if gazebo8 will be in upstream Debian or Ubuntu in the future distributions?

j-rivero:
> Maybe @j-rivero can give some insight as if gazebo8 will be in upstream Debian or Ubuntu in the future distributions?

It is not present in the official Debian or Ubuntu repositories but fully available, supported and developed from packages.osrfoundation.org and with support from `gazebo_ros_pkgs`. We can consider it as not officially present in any official release of Ubuntu/Debian/ROS but supported for alternative cases.

If it does not imply a troubles and people find a use case for it, I think that we can consider add the keys for rosdep.

tfoote:
@j-rivero There is a set of rosdep rules you can add to your rosdep sources if you add packages.osrfoundation.org to your apt sources right? I can't find the link at the moment.

mikaelarguedas:
@j-rivero Ok I guess I got confused, and the fact that it was not available upstream was the reason we removed it: (#14564 #15000).
Is this supposed to be used only for source builds ?

tfoote:
The keys in the default rosdep rules should represent only what's available from upstream + the ROS repo.  We don't accept keys that refer to custom PPAs (ubuntu hosted or not) as that would cause rosdep say that things are resolved that are not downloadable. For rosdep to work with content in a PPA the PPA mainatiners should maintain a rosdep rules file that defines the resolutions that they provide and then when you add the PPA to your sources, you also add a rosdep source to your system.

paulbovbel:
That make sense, thank you. If you do find rosdep rules for packages.osrf.org, please let me know!

j-rivero:
> @j-rivero There is a set of rosdep rules you can add to your rosdep sources if you add packages.osrfoundation.org to your apt sources right?

The ones I use when changing `gazebo_ros_pkgs` from using the official version (i.e: gazebo7 in Kinetic) to an alternative one (i.e: gazebo8 in Kinetic) are here: https://github.com/osrf/osrf-rosdep/ 

> The keys in the default rosdep rules should represent only what's available from upstream + the ROS repo.

ACK.

tfoote:
@paulbovbel  FYI ^^

@j-rivero It would be great to add instructions for inserting those to this tutorial: http://gazebosim.org/tutorials/?tut=ros_wrapper_versions

PR #16386, opened 2017-11-16
Use pip for `python-oauth2`

furushchev:
Since ubuntu from wily no apt package `python-oauth2` is available, this package should be installed via `pip`.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1820,10 +1820,18 @@ python-numpydoc:
     zesty: [python-numpydoc]
 python-oauth2:
   arch: [python-oauth2]
-  debian: [python-oauth2]
+  debian:
+    wheezy:  [python-oauth2]
   fedora: [python-oauth2]
   gentoo: [dev-python/oauth2]
-  ubuntu: [python-oauth2]
+  ubuntu:
+    precise: [python-oauth2]
+    quantal: [python-oauth2]
+    raring: [python-oauth2]
+    saucy: [python-oauth2]
+    trusty: [python-oauth2]
+    utopic: [python-oauth2]
+    vivid: [python-oauth2]
 python-oauth2client:
   debian: [python-oauth2client]
   fedora: [python-oauth2client]


rosdep/python.yaml
```diff
@@ -1823,7 +1823,9 @@ python-oauth2:
   debian: [python-oauth2]
   fedora: [python-oauth2]
   gentoo: [dev-python/oauth2]
-  ubuntu: [python-oauth2]
+  ubuntu:
+    pip:
+      packages: [oauth2]
```
dirk-thomas: This kind of change is not feasible since there might be packages released into e.g. Indigo using this key. With the switch to `pip` those would not be releasable anymore. Please change the patch to enumerate the Ubuntu distros which contain the Debian package. For newer ROS distros maybe there is a package with a different name?

mikaelarguedas:
Looks like `python-oauth2` was a fork of `python-oauth` that has been dropped by Debian/Ubuntu since trusty (https://packages.ubuntu.com/trusty/python-oauth2). There is still a `python-oauth2client` being releasaed in all distros as well as `python-oauth`. I don't know if they provide the same functionality though

mikaelarguedas (dismissed):
marking this as changes requested until comments  are addressed

furushchev:
@dirk-thomas @mikaelarguedas Thank you for comments. Unfortunately `python-oauth2` is for OAuth1.0 (regardless of its name) but not for OAuth2.0 (supported by `python-oauth2client`)
Maybe the right way to fix is just dropping this key for wily and later ubuntu distros (where this package is no more available)?

mikaelarguedas:
@furushchev what about `python-oauth` doesnt it provide OAuth1.0 ?

Regarding `python-oauth2`, yes we should change the rule to cover only the distributions where it's available (for Ubuntu and Debian)

furushchev:
@mikaelarguedas Got it! Updated.
About `python-oauth`, I found the reason that we should use v2 package: https://github.com/joestump/python-oauth2#history

mikaelarguedas (approved):
lgtm, thanks @furushchev for iterating

PR #16412, opened 2017-11-16
[rosdep] Add python-graphviz

evenator:
Adds the PyPi graphviz package (https://pypi.python.org/pypi/graphviz)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1184,6 +1184,19 @@ python-googleapi:
     wily_python3: [python3-googleapi]
     xenial: [python-googleapi]
     xenial_python3: [python3-googleapi]
+python-graphviz:
+  debian:
+    pip:
+      packages: [graphviz]
+  fedora:
+    pip:
+      packages: [graphviz]
+  gentoo:
+    pip:
+      packages: [graphviz]
+  ubuntu:
+    pip:
+      packages: [graphviz]
 python-gridfs:
   debian: [python-gridfs]
   fedora: [python-pymongo-gridfs]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Since this is pip only please use the `-pip` extension convention for the key.

dirk-thomas:
Is the existing rosdep key `python-pygraphviz` not sufficient for your use case?

mikaelarguedas:
Gentoo doesn't support pip. But has an official package for it: https://packages.gentoo.org/packages/dev-python/graphviz.

I see that there is a `python-pygraphviz` in Ubuntu and Debian, is this similar to what you're looking for ?

Also it looks like newer Ubuntu and Debian will have a python package called python-graphviz but it's only a source package for now:
Ubuntu
```
 python-graphviz | 0.5.2-1        | zesty/universe  | source
 python-graphviz | 0.5.2-1ubuntu1 | artful/universe | source
 python-graphviz | 0.5.2-1ubuntu1 | bionic/universe | source
```
Debian:
```
python-graphviz | 0.5.2-1       | testing    | source
python-graphviz | 0.5.2-1       | unstable   | source
```

evenator:
> Since this is pip only please use the -pip extension convention for the key.

Yeah, I can do that.

> Is the existing rosdep key python-pygraphviz not sufficient for your use case?

pygraphviz is a different package.

> Gentoo doesn't support pip. But has an official package for it: https://packages.gentoo.org/packages/dev-python/graphviz.

I'll update accordingly.

> Also it looks like newer Ubuntu and Debian will have a python package called python-graphviz but it's only a source package for now:

Good to know. It's too bad there's no binary package yet.

mikaelarguedas (approved):
lgtm, thanks @evenator for iterating

tfoote (approved):

PR #16443, opened 2017-11-21
Use wxpython instead of python-wxtools on Arch

alkersan:
There is no package named `python-wxtools` in the Arch repository, neither in the AUR. Instead it's called [wxpython](https://www.archlinux.org/packages/community/x86_64/wxpython/) and provides similar tools as [Ubuntu's package](https://packages.ubuntu.com/xenial/all/python-wxtools/filelist) (e.g. `img2xpm`, `xrced`, `pywxrc`)


mikaelarguedas (approved):
lgtm, thanks @alkersan for providing the links and checkign the content of the package

PR #16445, opened 2017-11-21
Add python-virtualenv to rosdep

paulbovbel:


mikaelarguedas:
Ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python-virtualenv&searchon=names
Debian: https://packages.debian.org/search?keywords=python-virtualenv&searchon=names&suite=all&section=all
Fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/python-virtualenv/
Gentoo: https://packages.gentoo.org/packages/dev-python/virtualenv

mikaelarguedas:
@paulbovbel I didnt see python3-virtualenv in the fedora package database. Can you provide a link for it please ?
Can you also add the Gentoo rule `dev-python/virtualenv` ? thanks

@tfoote do we take `python3` keys in rosdep or should users provide <distro>_python3 rules instead to kep the key identical regardless of the python version? Are the python3 prefixed ones only the ones we added for ROS 2 beta release?

paulbovbel:
re: gentoo, added
re: fedora, my mistake, removing

tfoote:
The `_python3` rules were an attempt to allow switching to python3 using the os override functionality. But with [REP 149](http://www.ros.org/reps/rep-0149.html#dependencies-across-ros-1-and-ros-2) in process that should allow us to have those conditional keys in the package.xml. In particular the conditional dependencies on the python version will let us leverage the `python-*` vs `python3-*` appropriately.

Gentoo is I believe by default python3 so the rules should be the same? @allenh1 ?

For fedora 

For Ubuntu I also find `python3-venv` https://packages.ubuntu.com/xenial/python3-venv (though from the below docs this was recommend in 3.3 and 3.4 but is deprecated in 3.6)

And to add confusion venv is also built into python3 https://docs.python.org/3/library/venv.html

I think that it's likely that we should just recommend the built in venv module and thus to have the capability this rule would just need to require the core python3 package.

paulbovbel:
Thank you for clarifying. It's not immediately clear how the linked content would satisfy resolving to different system dependencies, but it looks like I should be able to get by just using python 3's internal module.

Would I be clear to add a `python3` rosdep pointing to python3-dev?

tfoote (approved):

mikaelarguedas (approved):
lgtm, squashing and merging

PR #16471, opened 2017-11-27
Update naoqi_libqi version

suryaambrose:


mikaelarguedas:
Thanks @suryaambrose for the quick fix :+1: .
The debinc going from 5 to 1 is surprising though... Looking at the release repository it seems to be the good one for lunar :+1: , it's hard to know how you populated the branches (the README's last update is 25 Mar 2016). I'd recommend using bloom-release in the future, it will update the README and generate the rosdistro PR automatically with a description making it easier to review.

suryaambrose:
The doc about thrid-party packages is more about git-bloom-release than bloom-release on its own.

I had some issues with the pre-existing patches, so I had to start over on all the branches (and the thing is yesterday I accidentally created the first lunar package with an increment at 5 instead of 0.. hence its decrease ^^

mikaelarguedas:
> I had some issues with the pre-existing patches, so I had to start over on all the branches (and the thing is yesterday I accidentally created the first lunar package with an increment at 5 instead of 0.. hence its decrease ^^

Haha makes sense. Based on my understanding of the way we handle apt repos: it should be fine; as the higher version has never hit the main repo. The only risk is that people who downloaded the version that built and made it to the testing repository (ros-shadow-fixed) will not get the new version by running apt upgrade as the new version is lower that the old one.

suryaambrose:
Yeah you're right. I hope noone had a chance to download it during that one-day window.

It still does not compile though.. I can't keep testing the build on the ROS buildfarms, I will mount a VM with yakkety and fix this. But I won't be available for a week, so if the farms complain too much, feel free to blacklist the package for a while ^^

mikaelarguedas:
If you want to test in a faster manner than setting up a VM manually, you can check out [prerelease.ros.org](http://prerelease.ros.org). This will generate docker containers for the selected platform, download the dependencies and build your repo with `catkin_make_isolated`.

>  so if the farms complain too much, feel free to blacklist the package for a while

It's fine to have packages failing for a little bit. I may need to remove this repo completely from rosdistro though as I'm [planning on syncing](https://discourse.ros.org/t/preparing-lunar-sync-2017-11-29/3268) the packages to the main repo later this week

PR #16528, opened 2017-12-03
Remove catkin_virtualenv test packages (indigo)

paulbovbel:


dirk-thomas:
You also need to remove the following lines since the release now contains exactly one package which has the same name as the repo:

```
packages:
   catkin_virtualenv
```

You should also update the release repo to ignore the test packages in order to avoid that future release reintroduce these packages.

paulbovbel:
Thanks, I'll take care of that in the upcoming release

PR #16534, opened 2017-12-04
simplify mrpt pkg names in Ubuntu

jlblancoc:
The need for the special cases for saucy & raring is no longer needed, since those distros fell long, long time ago into EOL (!)


mikaelarguedas (approved):
lgtm,
I'll let the second reviewer comment if we actually want to keep the rule accurate for raring and saucy.
I'm fine with it as they're both EOL for more that 3 years and were not LTS

tfoote (approved):
lgtm, raring and saucy are well out the back of our support lifetime.

PR #16535, opened 2017-12-04
pose_cov_ops: 0.1.6-0 in 'kinetic/distribution.yaml' [bloom]

jlblancoc:
Increasing version of package(s) in repository `pose_cov_ops` to `0.1.6-0`:

- upstream repository: https://github.com/mrpt-ros-pkg/pose_cov_ops.git
- release repository: https://github.com/mrpt-ros-pkg-release/pose_cov_ops-release.git
- bloom version: `0.6.1`
- previous version for package: `0.1.5-0`

## pose_cov_ops

```
* Fix catkin dependencies
* Use C++14
* Fix build against MRPT 1.5 and 2.0
* Contributors: Javier G. Monroy, Jose Luis Blanco Claraco, Jose Luis Blanco-Claraco
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6292,7 +6292,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/mrpt-ros-pkg-release/pose_cov_ops-release.git
-      version: 0.1.5-0
+      version: 0.1.6-0
     source:
       type: git
       url: https://github.com/mrpt-ros-pkg/pose_cov_ops.git


mikaelarguedas:
@jlblancoc I see c++14 in the changelog. ROS Kinetic targets C++11 as it's latest standard as most platforms don't have full compiler support for c++14. Does the code compile on platforms that don't have c++14 support ?

jlblancoc:
TL;DR: No, no C++14 features are required for now in the package code.

@mikaelarguedas thanks for pointing it out! Actually, yes, it builds: it's tested on empty docker containers in which ROS kinetic is installed from scratch, etc. 
The reason of adding the CMake C++14-enable variable is because the same sources of this package build against two version of the underlying MRPT library: 
* the old one (currently in the official Ubuntu repos) uses good-old C++98 only, and 
* the "under development" version, that will require C++14 (since language features are used in public headers). But that newer version will not be released until, at least, Ubuntu 18.04 (probably later). So existing ROS distros will remain building against the current "old" library version. 

Though, I want to keep just one version of the ROS package source code so it builds against all new and old versions of its dependencies, that's why it was updated: for the compiler not to go mad **if** it finds C++14 features in `#include`d  headers.

mikaelarguedas:
> Though, I want to keep just one version of the ROS package source code so it builds against all new and old versions of its dependencies, that's why it was updated: for the compiler not to go mad if it finds C++14 features in #included headers.

sounds good :+1: , thanks for clarifying.

PR #16541, opened 2017-12-06
Enable pull request testing for nmea_msgs

evenator:


jade/distribution.yaml
```diff
@@ -3758,6 +3758,7 @@ repositories:
       url: https://github.com/ros-drivers-gbp/nmea_msgs-release.git
       version: 1.0.0-0
     source:
+      test_pull_requests: true
```
dirk-thomas: The ROS distro `Jade` has been EOLed a while ago. We don't run a buildfarm for it anymore. Therefore we don't accept changes to it anymore. Please revert this part of the PR.

evenator:
Are any other changes needed?

PR #16544, opened 2017-12-06
python: add gpiozero as pip dependency

kyrofa:
Ubuntu is the only platform I have to test, and it's not packaged.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -947,10 +947,6 @@ gperftools:
   fedora: [gperftools, gperftools-devel]
   gentoo: [dev-util/gperf]
   ubuntu: [google-perftools, libgoogle-perftools-dev]
-gpiozero:
-  ubuntu:
-    pip:
-      packages: [gpiozero]
 gpsd:
   debian: [gpsd]
   fedora: [gpsd]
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Please use the `-pip` suffix when adding pip only keys.
And add python packages in the python.yaml file rather that base.yaml.

Even if you dont test with it can you please add a debian entry as well?

Note for the future: this package will be available in apt soon:
- ubuntu arm since zesty no amd64 yet
- debian since buster

mikaelarguedas:
same comments apply to #16545

kyrofa:
Sure thing, done. Fixed #16545 as well.

tfoote:
I see ubuntu and debian packages for this too: 

https://packages.ubuntu.com/zesty/python-gpiozero
https://packages.debian.org/source/buster/gpiozero

kyrofa:
Alright, updated.

tfoote:
Thanks for iterating. Two slight requests, add yakkety which was skipped and CI needs the keys in alphabetical order.

you can validate locally with `nosetests -s` or specifically for rosdep `nostests -s test/test_rosdep_formatting.py`

kyrofa:
Whoops, not sure how I missed Y! Done and done.

mikaelarguedas:
Can you please add ubuntu bionic as well? and rename the key to python-gpiozero

kyrofa:
Ah yes of course. Renamed and added bionic.

tfoote:
Sorry, also top level alphabetization here.

kyrofa:
Fixed here as well.

tfoote (approved):

mikaelarguedas (approved):

PR #16545, opened 2017-12-06
python: add RPi.GPIO as pip dependency

kyrofa:
I didn't see any other examples in base.yaml that capitalizes the dependency names, so I kept it lower-case, even though the pip package itself is not. Please let me know if that should be changed.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4006,13 +4006,38 @@ rosbag-metadata-pip:
   ubuntu:
     pip:
       packages: [rosbag-metadata]
-rpi.gpio-pip:
+rpi.gpio:
   debian:
-    pip:
-      packages: [RPi.GPIO]
+    buster: [python-rpi.gpio]
+    jessie:
+      pip:
+        packages: [RPi.GPIO]
+    stretch:
+      pip:
+        packages: [RPi.GPIO]
   ubuntu:
-    pip:
-      packages: [RPi.GPIO]
+    artful: [python-rpi.gpio]
+    zesty: [python-rpi.gpio]
+    xenial:
+      pip:
+        packages: [RPi.GPIO]
+    wily:
+      pip:
+        packages: [RPi.GPIO]
+    vivid:
+      pip:
+        packages: [RPi.GPIO]
+    utopic:
+      pip:
+        packages: [RPi.GPIO]
+    trusty:
+      pip:
+        packages: [RPi.GPIO]
+  fedora:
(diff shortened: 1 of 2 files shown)


tfoote:
This looks to be in ubuntu as of zesty: https://packages.ubuntu.com/zesty/python-rpi.gpio and debian as of buster: https://packages.debian.org/source/buster/rpi.gpio we should target those versions and keys. Also Fedora as of 26: https://admin.fedoraproject.org/pkgdb/package/rpms/python-rpi-gpio

kyrofa:
Updated.

tfoote:
Same yakkety and alphabetic requests here too.

kyrofa:
Fixed, thank you!

tfoote (approved):

tfoote (changes requested):
Sorry I missed the key should be python-rpi.gpio here too.

mikaelarguedas:
bionic should be added here too

kyrofa:
Renamed and added bionic. Thanks you two!

rosdep/python.yaml
```diff
@@ -4006,6 +4006,42 @@ rosbag-metadata-pip:
   ubuntu:
     pip:
       packages: [rosbag-metadata]
+python-rpi.gpio:
```
tfoote: Sorry one more tweak, with the rename it's out of alphabetical order now.

kyrofa:
Ugh, duh, fixed, so sorry.

tfoote (approved):

mikaelarguedas (approved):

mikaelarguedas:
Apparently these apt packages are not available on non-arm platform.
https://packages.ubuntu.com/bionic/python-rpi.gpio
This makes rosdep fail on amd if a workspace contains a package declaring a dependency on this key.
<details><summary> Error message </summary>

```
executing command [apt-get install python-rpi.gpio]
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package python-rpi.gpio
E: Couldn't find any package by glob 'python-rpi.gpio'
E: Couldn't find any package by regex 'python-rpi.gpio'
  apt: command [apt-get install python-rpi.gpio] failed
```
</details>

Pip seems happy to install regardless of the architecture.

@tfoote @kyrofa WDYT ? should we change this key to be a pip only key ? (I doubt any package is released using this as it would fail on all non-arm archs)

kyrofa:
I'm still only using the pip version, so I don't have strong feelings either way.

PR #16549, opened 2017-12-07
Add rosdep keys for pyparsing and pytest.

nuclearsandwich:
pytest and pyparsing are needed for ROS 2 development.

pytest is the new recommended test runner for ament_python packages (as far as I understand).

pyparsing is used in the current implementation of the conditional dependencies [[1]] for package.xml format 3

[1]: https://github.com/ros-infrastructure/rep/issues/143

Package references:

* pyparsing
    * arch https://www.archlinux.org/packages/extra/any/python-pyparsing/
    * debian https://packages.debian.org/stable/python3-pyparsing
    * fedora https://apps.fedoraproject.org/packages/python3-pyparsing
    * gentoo https://packages.gentoo.org/packages/dev-python/pyparsing
    * ubuntu
        * trusty https://packages.ubuntu.com/trusty/python-pyparsing
        * xenial https://packages.ubuntu.com/xenial/python-pyparsing
        * artful https://packages.ubuntu.com/artful/python-pyparsing
        * bionic https://packages.ubuntu.com/bionic/python-pyparsing

* pytest
    * arch https://www.archlinux.org/packages/community/any/python-pytest/
    * debian https://packages.debian.org/stable/python3-pytest
    * fedora (2)  https://apps.fedoraproject.org/packages/python3-pytest/sources/
    * gentoo https://packages.gentoo.org/packages/dev-python/pytest
    * ubuntu
        * trusty https://packages.ubuntu.com/trusty/python-pytest
        * xenial https://packages.ubuntu.com/xenial/python-pytest
        * artful https://packages.ubuntu.com/artful/python-pytest
        * bionic https://packages.ubuntu.com/bionic/python-pytest

(2) Fedora's python3-pytest is an EPEL-only package.
I'm not sure if it should be included.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4055,6 +4055,8 @@ python3-pyparsing:
   ubuntu: [python3-pyparsing]
 python3-pytest:
   arch: [python-pytest]
+  ubuntu:
+    bionic: [python3-pytest]
 python3-pytest-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4055,6 +4055,8 @@ python3-pyparsing:
   ubuntu: [python3-pyparsing]
 python3-pytest:
   arch: [python-pytest]
+  debian: [python3-pytest]
+  gentoo: [dev-python/pytest]
   ubuntu: [python3-pytest]
 python3-ruamel.yaml:
   debian:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4057,7 +4057,6 @@ python3-pytest:
   arch: [python-pytest]
   debian: [python3-pytest]
   fedora: [python3-pytest]
-  gentoo: [dev-python/pytest]
   ubuntu: [python3-pytest]
 python3-ruamel.yaml:
   debian:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4047,6 +4047,18 @@ python3-pep8:
 python3-pkg-resources:
   debian: [python3-pkg-resources]
   ubuntu: [python3-pkg-resources]
+python3-pyparsing:
+  arch: [python-pyparsing]
+  debian: [python3-pyparsing]
+  fedora: [python3-pyparsing]
+  gentoo: [dev-python/pyparsing]
+  ubuntu: [python3-pyparsing]
+python3-pytest:
+  arch: [python-pytest]
+  debian: [python3-pytest]
+  fedora: [python3-pytest]
+  gentoo: [dev-python/pytest]
+  ubuntu: [python3-pytest]
 python3-ruamel.yaml:
   debian:
     buster: [python3-ruamel.yaml]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4057,16 +4057,6 @@ python3-pytest:
   arch: [python-pytest]
   ubuntu:
     bionic: [python3-pytest]
-python3-pytest-pip:
-  debian:
-    pip:
-      packages: [pytest]
-  fedora:
-    pip:
-      packages: [pytest]
-  ubuntu:
-    pip:
-      packages: [pytest]
 python3-ruamel.yaml:
   debian:
     buster: [python3-ruamel.yaml]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4055,8 +4055,7 @@ python3-pyparsing:
   ubuntu: [python3-pyparsing]
 python3-pytest:
   arch: [python-pytest]
-  ubuntu:
-    bionic: [python3-pytest]
+  ubuntu: [python3-pytest]
 python3-ruamel.yaml:
   debian:
     buster: [python3-ruamel.yaml]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4055,9 +4055,16 @@ python3-pyparsing:
   ubuntu: [python3-pyparsing]
 python3-pytest:
   arch: [python-pytest]
-  debian: [python3-pytest]
-  fedora: [python3-pytest]
-  ubuntu: [python3-pytest]
+python3-pytest-pip:
+  debian:
+    pip:
+      packages: [pytest]
+  fedora:
+    pip:
+      packages: [pytest]
+  ubuntu:
+    pip:
+      packages: [pytest]
 python3-ruamel.yaml:
   debian:
     buster: [python3-ruamel.yaml]


dirk-thomas:
I don't think this is sufficient since we need a newer version of `pytest` (see ros2/ci#111).

nuclearsandwich:
ooh I hadn't realized we needed 3.2. 

pytest 3.2 is available but masked in gentoo and 3.3 is the default on arch.

We could just drop the keys for fedora, debian, and ubuntu or move them out to a python3-pytest-pip rosdep key.

nuclearsandwich:
Trying not to let this wither. I've removed the key for platforms where pytest is too old to be used for ROS 2 and added a separate `python3-pytest-pip` key.

Bionic [lists a suitable python3-pytest](https://packages.ubuntu.com/bionic/python3-pytest) but I'm not sure whether now is the time to add bionic keys or not.

I also apparently had this pull request mistargted and it is now correctly pointing at master.

rosdep/python.yaml
```diff
@@ -4047,6 +4047,24 @@ python3-pep8:
...
   ubuntu: [python3-pkg-resources]
+python3-pyparsing:
+  arch: [python-pyparsing]
+  debian: [python3-pyparsing]
+  fedora: [python3-pyparsing]
+  gentoo: [dev-python/pyparsing]
+  ubuntu: [python3-pyparsing]
+python3-pytest:
+  arch: [python-pytest]
+python3-pytest-pip:
```
dirk-thomas: Which package(s) is/are going to use this key?

dirk-thomas:
> Bionic lists a suitable python3-pytest but I'm not sure whether now is the time to add bionic keys or not.

Since the code name is fixed you can add entries for it anytime.

rosdep/python.yaml
nuclearsandwich: I have no explicit use case for this key, I was attempting to follow observed convention that keys using pip were separate from the regular rosdep key.

It's possible that a platform or OS variable could be used with Package format 3 dependency conditions to select the pip dependency on platforms where it is not available from the default provider.

dirk-thomas: I would suggest not to add rosdep keys which are not needed / used by any package but wait with such an addition until that is the case.

nuclearsandwich: Does it matter that dropping this would mean that no rosdep dependency key for ubuntu xenial (the supported platform) satisfies the test dependencies for ROS 2?

dirk-thomas: Before you said the key isn't used anywhere. Under that assumption I recommended to not add it (since it would not be used anywhere).

Can you please clarify which package currently uses the key?

nuclearsandwich: > Can you please clarify which package currently uses the key (or is going to be updated to use this key)?

No packages currently use the key. `ament_cmake_pytest` depends on https://github.com/ament/ament_cmake/blob/master/ament_cmake_pytest/package.xml#L14 `python3-pytest` but as you previously mentioned, apparently the version in apt isn't sufficient. Since tests don't abort buildfarm builds, this did not block the release.

There's also a significant number of packages which have `test_requires=['pytest']` in the setup.py but no entry for this in the package.xml.

I don't have much stake in whether the solution is using the pip provider in the "generic" key or switching to use the pip key knowing that our supported platform can't be supported without it. The only way I can think to have cake and eat it too is by trying to use the new dependency conditions to apply platform specific dependencies.

dirk-thomas: > The only way I can think to have cake and eat it too is by trying to use the new dependency conditions to apply platform specific dependencies.

Can you please make a concrete example. Currently I can't follow you how this would look like.

nuclearsandwich:
```
<depend condition="$OS_CODENAME == bionic">python3-pytest</depend>
<depend condition="$OS_CODENAME != bionic">python3-pytest-pip</depend>
```

dirk-thomas: That would imply that the package can't be released with bloom into distributions other than `bionic`.

nuclearsandwich: Right, I forgot that bloom hard-stops on pip dependencies.

There was a proposed change to allow them to be skipped instead. https://github.com/ros-infrastructure/bloom/pull/412

tfoote: Skipping them breaks the expectation that all dependencies are resolved when installing a debian package.

nuclearsandwich: > Skipping them breaks the expectation that all dependencies are resolved when installing a debian package.

It isn't my aim to revisit that PR discussion here. I had to remind myself of bloom's behavior and cross-linked that PR as it was part of that investigation.

If this key isn't useful. I'll remove it.

rosdep/python.yaml
```diff
@@ -4047,6 +4047,16 @@ python3-pep8:
...
+python3-pyparsing:
+  arch: [python-pyparsing]
+  debian: [python3-pyparsing]
+  fedora: [python3-pyparsing]
+  gentoo: [dev-python/pyparsing]
+  ubuntu: [python3-pyparsing]
+python3-pytest:
+  arch: [python-pytest]
+  ubuntu:
+    bionic: [python3-pytest]
```
dirk-thomas: I would suggest to just use `ubuntu: [python3-pytest]` in order to not need to maintain this in the future. The specific version requirement of `ament_tools` is just an arbitrary line in the sand. Any other package might have a different threshold.

nuclearsandwich: This is contrary to your first piece of feedback on this pull request.

dirk-thomas: My first comment was in the context of your proposal to add the rosdep keys in order to be used for the ROS 2 packages. In that context I mentioned that this will not work since we require a newer version.

During this discussion we have established that the ROS 2 packages can't use this new rosdep key since there is no Debian package available which is new enough. And in the context of just adding these rosdep keys for other packages I think it should not be limited to bionic.

dirk-thomas (approved):

rosdep/python.yaml
```diff
@@ -4047,6 +4047,15 @@ python3-pep8:
...
   ubuntu: [python3-pkg-resources]
+python3-pyparsing:
+  arch: [python-pyparsing]
+  debian: [python3-pyparsing]
+  fedora: [python3-pyparsing]
+  gentoo: [dev-python/pyparsing]
+  ubuntu: [python3-pyparsing]
+python3-pytest:
+  arch: [python-pytest]
+  ubuntu: [python3-pytest]
```
mikaelarguedas: We encourage to add rules for debian and not only ubuntu.
arch fedora and gentoo are optional but apreciated

mikaelarguedas:
Ubuntu links in description point to python2 version of the packages. Here are the links for the pythn3 packages:

python3-pyparsing: https://packages.ubuntu.com/search?keywords=python3-pyparsing&searchon=names&exact=1&suite=all&section=all
python3-pytest: https://packages.ubuntu.com/search?keywords=python3-pytest&searchon=names&exact=1&suite=all&section=all

@nuclearsandwich should the debian and gentoo rules for pytest be readded? Or have they been left out on purpose?

nuclearsandwich:
> @nuclearsandwich should the debian and gentoo rules for pytest be readded? Or have they been left out on purpose?

Now that I'm not trying to add rules solely to satisfy ROS 2, I think they can be re-added. The Fedora package was both very old and in EPEL rather than the primary Fedora repos so I don't think it should be included.

mikaelarguedas:
> The Fedora package was both very old and in EPEL rather than the primary Fedora repos so I don't think it should be included.

agreed :+1: that's why I scoped my comment to Debian and Gentoo

tfoote (approved):

PR #16551, opened 2017-12-07
tango_ros: 2.0.0-0 in 'kinetic/distribution.yaml' [bloom]

meyerj:
Increasing version of package(s) in repository `tango_ros` to `2.0.0-0`:

- upstream repository: https://github.com/Intermodalics/tango_ros.git
- release repository: https://github.com/Intermodalics/tango_ros-release.git
- bloom version: `0.6.1`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9509,6 +9509,10 @@ repositories:
       version: master
     status: developed
   tango_ros:
+    doc:
+      type: git
+      url: https://github.com/Intermodalics/tango_ros.git
+      version: master
     release:
       packages:
       - tango_ros_messages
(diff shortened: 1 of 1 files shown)


meyerj:
The [dev job for this repo](http://build.ros.org/view/Kdev/job/Kdev__tango_ros__ubuntu_xenial_amd64/1/console) fails to build because
- the repo contains submodules which are required to build
- the `tango_ros_streamer` package requires an installation of the Android SDK and only builds an APK, so it probably cannot be easily tested on the ROS build farm. The package is ignored for binary releases.

Is there a way to disable the dev job? Or should we just remove the `source` section again? It was added by bloom on request but I did not think about the consequences.

dirk-thomas:
> Is there a way to disable the dev job?

Yes, you can add `test_commits: false` to the `source` section to skip the devel jobs but still allow e.g. `rosinstall_generator` to fetch the sources.

PR #16557, opened 2017-12-08
Adding ROSRemote to documentation index for distro

alysonmp:


tfoote:
The hydro distro is EOL for quite a while now. http://wiki.ros.org/Distributions We no longer accept updates. Please consider submitting it to one of the current releases, Indigo, Kinetic, Lunar.

PR #16558, opened 2017-12-08
roll back image_pipeline for kinetic sync

tfoote:
https://discourse.ros.org/t/preparing-for-kinetic-sync-2017-11-29/3275/3

This is rolling back the known regression in https://github.com/ros-perception/image_pipeline/pull/299 pending a new release with the patch. There's a proposed fix here: https://github.com/ros-perception/image_pipeline/pull/311 but while that's not merged and released I'm rolling this back to unblock the release. 

Re: #16276

@vrabaud FYI

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3161,7 +3161,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ros-gbp/image_pipeline-release.git
-      version: 1.12.21-0
+      version: 1.12.20-0
     source:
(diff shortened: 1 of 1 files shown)


tfoote:
This rollback failed to build. The new version of opencv deprecated a method we're using that was resolved by setting a flag here: https://github.com/ros-perception/image_pipeline/compare/1.12.20...1.12.21#diff-d18a1b026774c185d315d5a023df04d0R43

https://docs.opencv.org/3.3.1/d0/d3a/classcv_1_1DataType.html

PR #16559, opened 2017-12-08
Remove failing hebiros packages

tfoote:
Avoid the build: https://github.com/HebiRobotics/HEBI-ROS/issues/18

Partial rollback of #16522


xaxafour:
https://answers.ros.org/question/277490/how-do-you-determine-the-architecture-for-release-builds/

tfoote:
Note that is more related to https://github.com/HebiRobotics/HEBI-ROS/issues/17 than the above 18

PR #16562, opened 2017-12-09
Add new package, rdl, to the distribution list

jlack1987:
This is my first time going through this process and releasing a package via bloom to rosdistro. I believe I have done it correctly, but please let me know if any problems are found and i'll redo.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6834,6 +6834,26 @@ repositories:
       url: https://github.com/timn/ros-rcll_ros_msgs.git
       version: master
     status: developed
+  rdl:
+    doc:
+      type: git
+      url: https://gitlab.com/jlack/rdl.git
+      version: master
+    release:
+      packages:
+      - rdl
+      - rdl_benchmark
+      - rdl_cmake
+      - rdl_dynamics
+      - rdl_urdfreader
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://gitlab.com/jlack/rdl_release.git
+      version: 0.9.1-1
+    source:
+      type: git
+      url: https://gitlab.com/jlack/rdl.git
+      version: master    
   realsense_camera:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


jlack1987:
Not sure why that failed. The details just say some identical file check failed. I'll go through the release process again and see if I messed anything up. Any suggestions are appreciated!

kinetic/distribution.yaml
```diff
@@ -6834,6 +6834,26 @@ repositories:
...
+      - rdl_dynamics
+      - rdl_urdfreader
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://gitlab.com/jlack/rdl_release.git
+      version: 0.9.1-1
+    source:
+      type: git
+      url: https://gitlab.com/jlack/rdl.git
+      version: master
```
dirk-thomas: The error message also mentions:

> see folded block 'block6' for details

And in block6 it shows a diff in this line. It seems to have trailing whitespaces which should not be there. If you edit the PR and remove those the Travis test will likely pass.

mikaelarguedas:
@jlack1987 quick follow-up of https://discourse.ros.org/t/newly-released-package-not-building-on-travis/3373/2:

- packages released on ros buildfarm should have all their dependencies satisfied with apt as debian packages cannot embed pip dependencies. In other words, a package with dependencies available via pip cannot be released
- the error you are seeing is because "cppcheck-junit" is not an existing apt package on ubuntu xenial.
- looking at the [release repository](https://gitlab.com/jlack/rdl_release) it looks like none of the packages had their keys successfully resolved (I don't see any branch `debian/kinetic/OS_NAME/PACKAGE_NAME`) So I suspect bloom failed and should have thrown warnings telling you that it could not generate any of the control files (for ubuntu debian or fedora).

jlack1987:
hmm ok. It's not really a true code dependency but I use it in the GitLab CI job to lint my `CMakeLists.txt` files. I guess I can remove it from the dependency list and just install it directly in my CI dockers so it'll still be there for my CI jobs on GitLab. Thanks for following up as i've never worked with Travis nor released a ros package before so the info is greatly appreciated. I'll come up with a fix for this in the next few days and release again.

PR #16579, opened 2017-12-12
Add nvidia-ml-py as pip dependency

rsmallegoor:
Adds [nvidia-ml-py](https://pypi.python.org/pypi/nvidia-ml-py/7.352.0) which can be used to retrieve information from an NVIDIA GPU.

Required by this [PR](https://github.com/ros/diagnostics/pull/70) which add a GPU monitor to the diagnostic tools.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1866,9 +1866,6 @@ python-numpydoc:
     xenial: [python-numpydoc]
     zesty: [python-numpydoc]
 python-nvidia-ml-py-pip:
-  osx:
-    pip:
-      packages: [nvidia-ml-py]
   ubuntu:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Thanks @rsmallegoor for the links and the context.

Can you please add a debian rule for this package as well?

Note that any package depending on pip keys (or any non apt key) cannot be released as a debian package. So if the linked PR gets accepted, `diagnostic_common_diagnostics` cannot be released in ROS anymore

rsmallegoor:
@mikaelarguedas Based on your comments and [this discussion](https://github.com/ros/diagnostics/pull/70), I've decided to create a standalone package.

PR #16580, opened 2017-12-12
rdl: 0.9.3-0 in 'kinetic/distribution.yaml' [bloom]

jlack1987:
Increasing version of package(s) in repository `rdl` to `0.9.3-0`:

- upstream repository: https://gitlab.com/jlack/rdl.git
- release repository: https://gitlab.com/jlack/rdl_release.git
- bloom version: `0.6.1`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6849,10 +6849,12 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://gitlab.com/jlack/rdl_release.git
+      version: 0.9.3-0
     source:
       type: git
       url: https://gitlab.com/jlack/rdl.git
       version: master
+    status: developed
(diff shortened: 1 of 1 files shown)


jlack1987:
Ok I removed `cppcheck-junit-pip` from the keys and have my CI docker install it manually for my unit testing purposes and then did a new release which I believe went much better this time. There was still an issue with `rpm` not having a key for `cppcheck`, but that's again another test dependency and keys for it do exist for other linux distros so I think it's ok.

mikaelarguedas:
Yeah the release repository does look much better now.

The ROS buildfarm doesn't build RPMs, only debs. The fact that the key is missing means that people running their own fedora buildfarm won't be able to build that package.
I'm going to merge this as is, please consider contributing the missing fedora key before your next release.

Thanks!

PR #16604, opened 2017-12-16
rdl: 0.9.5-0 in 'kinetic/distribution.yaml' [bloom]

jlack1987:
Increasing version of package(s) in repository `rdl` to `0.9.5-0`:

- upstream repository: https://gitlab.com/jlack/rdl.git
- release repository: https://gitlab.com/jlack/rdl_release.git
- bloom version: `0.6.1`
- previous version for package: `0.9.4-0`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6863,7 +6863,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://gitlab.com/jlack/rdl_release.git
-      version: 0.9.4-0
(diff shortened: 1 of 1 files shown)


jlack1987:
Hoping this gets all the ubuntu packages passing on the buildfarm. There's something different about how it's building on the ROS buildfarm that's slightly different from how I build it on my CI server that's causing a dependency to not always get installed. Tried running the buildfarm locally but couldn't get it to work. Think this'll fix it, but again thanks for the patience.

PR #16651, opened 2017-12-25
Adding nonpersistent voxel layer package for Kinetic

SteveMacenski:


kinetic/distribution.yaml
```diff
@@ -5487,6 +5487,20 @@ repositories:
       url: https://github.com/ros/nodelet_core.git
       version: indigo-devel
     status: maintained
+  non_persistent_voxel_layer:
```
dirk-thomas: I would suggest to use the same name as the repository here.

kinetic/distribution.yaml
```diff
@@ -5487,6 +5487,20 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/SteveMacenski/nonpersistent_voxel_layer.git
+      version: master
+    release:
+      packages:
+      - nonpersistent_voxel_layer
+      url: https://github.com/SteveMacenski/nonpersistent_voxel_layer.git
```
dirk-thomas: The release repository should be a separate repository. Please see the `bloom` tutorial for details.

kinetic/distribution.yaml
SteveMacenski: I assume this name have to be the same as the ros answers URL? I have created the page non_persistent_voxel_layer already. If its an issue I can delete the page and make a new page with the repo name.

dirk-thomas: The wiki page must match the package name [nonpersistent_voxel_layer](https://github.com/SteveMacenski/nonpersistent_voxel_layer/blob/9feec7616edf155382a24e1d12883a3496dc4849/package.xml#L2). Since the package name as well as the repository name are `nonpersistent_voxel_layer` I would suggest to use the same name here since otherwise user will be confused why they have to use `non_persistent_voxel_layer` when calling e.g. the `rosinstall_generator`.

Also I didn't find the wiki page with either of the two names. If you have already created a wiki page it would be good to link it here to avoid ambiguity.

SteveMacenski: wiki page is [here](http://wiki.ros.org/nonpersistent_voxel_grid), I renamed page from `non_persistent_voxel_grid` to `nonpersistent_voxel_grid`. 

I'll edit this file right now! Thanks for the help and let me know if there's anything else off.

PR #16657, opened 2017-12-26
Adding trac_ik to documentation index for lunar

pbeeson:
I'd like trac_ik to be indexed on ros.org in preparation for a release on lunar


pbeeson:
Accidentally kept kinetic in the release info in distribution.yaml instead of changing to lunar.  Realized after deleting my fork, so closing and opening a new PR.

PR #16666, opened 2017-12-27
bus_server: 0.1.1-0 in 'kinetic/distribution.yaml' [bloom]

rohbotics:
Increasing version of package(s) in repository `bus_server` to `0.1.1-0`:

- upstream repository: https://github.com/UbiquityRobotics/bus_server.git
- release repository: https://github.com/UbiquityRobotics-release/bus_server-release.git
- bloom version: `0.6.1`

## bus_server

```
* Added minor travis stuff
* Contributors: Rohan Agrawal
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -686,6 +686,12 @@ repositories:
       url: https://github.com/voxel-dot-at/bta_tof_driver.git
       version: master
     status: maintained
+  bus_server:
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/UbiquityRobotics-release/bus_server-release.git
+      version: 0.1.1-0
   calibration:
     doc:
       type: git


tfoote:
This name doesn't really follow our naming conventions: http://www.ros.org/reps/rep-0144.html http://wiki.ros.org/ROS/Patterns/Conventions `bus` and `server` are very ambiguous. I'd like to suggest that this be renamed to be a little more clearly scoped.

rohbotics:
@waynegramlich How about a more specific name like `makerbus_server` or `ubiquity_bus_server`? Any other ideas?

@tfoote How would be go about the rename? I could do a bloom release for the package and remove `bus_server` from the rosdistro, does that sound correct?

tfoote:
Yeah, after updating the source you can just run bloom agian with the new name, and remove the entry added here.

waynegramlich:
Since Freya is no longer in active development, I nominate renaming it to `loki_node` or `loki_robot` or something more descriptive.  I would use the same node name as for Magni, just substituting in Loki instead.

PR #16678, opened 2018-01-03
Update python.yaml

reinzor:
Added pypozyx pip candidate

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2428,7 +2428,7 @@ python-pynmea2:
   ubuntu:
     pip:
       packages: [pynmea2]
-python-pypozyx:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
I see the package here: https://pypi.python.org/pypi/pypozyx

Since it's pip only please add the `-pip` suffix to the rosdep key.  Do you expect it to work on platforms other than ubuntu if so adding rules for them might make sense too.

reinzor:
I added the -pip suffix. I have not tested the library on other platforms .

reinzor:
ping

wjwwood (approved):

tfoote (approved):

PR #16679, opened 2018-01-03
Update the FreeBSD rosdistro information

NorwegianRockCat:
These match the current pkg names instead of the older port names. It
is using some pip packages that need to become FreeBSD ports. This is
a chicken and egg problem, so we have to start somewhere. At least
with these changes it is possible to install ROS on FreeBSD.


tfoote (approved):
This is great to fill in. In general we ask that contributions link to online package listings for contributions for easier review. (like packages.ubuntu.com or packages.debian.org ) With this many it's harder to link to them all. But is there a good place to check for freebsd packages?

NorwegianRockCat:
Yes, the more or less canonical place is [freshports.org](https://www.freshports.org/).

NorwegianRockCat:
The more or less canonical place is [ Freshports](https://www.freshports.org/).

NorwegianRockCat:
Hmm… OK. I see that an open request gathers other things I push to master. So, for the moment I'll close this and come back when I get more packages added for the desktop stuff. Sorry for creating the mess and thanks for the earlier approval.

NorwegianRockCat:
*Doh* I apologize. OK. I guess you can't merge if I close the request. I'm reopening and letting this sit. Sorry for the notification spam. I'll just create a branch to stop the extra pushes in my fork.

tfoote:
No problem. In general we recommend creating a feature branch with the specific changes for a PR. That way you can specifically update that branch in response to reviews, but other development will be isolated.

rosdep/base.yaml
```diff
@@ -595,6 +596,7 @@ eigen:
     stretch: [libeigen3-dev]
     wheezy: [libeigen3-dev]
   fedora: [eigen3-devel]
+  freebsd: [eigen]
```
tfoote: I don't see plain eigen listed: https://www.freshports.org/search.php?stype=name&method=match&query=eigen&num=10&orderby=category&orderbyupdown=asc&search=Search&format=html&branch=head I see eigen2 and eigen3

rosdep/base.yaml
```diff
@@ -2986,6 +3014,7 @@ liburdfdom-dev:
   arch: [urdfdom]
   debian: [liburdfdom-dev]
   fedora: [urdfdom-devel]
+  freebsd: [ros-urdfdom]
```
tfoote: I don't see ros-urdfdom in the index: https://www.freshports.org/search.php?stype=name&method=match&query=ros-urdfdom&num=10&orderby=category&orderbyupdown=asc&search=Search&format=html&branch=head

tfoote (changes requested):
I did some random checks and found a few that don't seem to be listed in the ports index.

rosdep/base.yaml
NorwegianRockCat: Actually the field you want to look at from freshports is **PKGNAME**. That's the name the package program uses. It doesn't necessarily line up with the port name. I'm not sure why this can diverge, but I'm sure there are historical reasons (FreeBSD ports were around long before the new pkg system, and they serve as the backbone for most of pkg) 

In the case above, you'll see that eigen3's package name is, indeed, eigen. So, that should be OK.

I have removed the ros-urdfdom changes for the moment as I would need to push those ports forward. I've got the basics done. I just need to ask to get them added.

NorwegianRockCat:
@tfoote Thanks. I'll keep my future pull requests in a feature branch.

rosdep/base.yaml
NorwegianRockCat: I have removed the ros-urdfdom changes for the moment as I would need to push those ports forward. I've got the basics done. Just need to push them forward. Without ros-urdfdom, I think you get the base communication system up and running.

tfoote (approved):
Thanks for the info. lgtm

rosdep/base.yaml
```diff
@@ -1462,6 +1469,7 @@ libconsole-bridge-dev:
   arch: [console-bridge]
   debian: [libconsole-bridge-dev]
   fedora: [console-bridge-devel]
+  freebsd: [ros-console_bridge]
```
nuclearsandwich: Since this appears to be a speculative key rather than one representing an existing port, do you mean to mix dashes and underscores in this port name? I know there's no restriction on doing so but it seems the common convention is to use dashes as the standard separator.

NorwegianRockCat: This is actually in process as I've submitted a [new port for this](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=224896). I agree the port name is not great, but I'm hoping that this is something that can be discussed/changed when the port is actually committed. I'll be following this and sending an update as soon as I know more.

nuclearsandwich: Makes sense. Thanks for clarifying.

nuclearsandwich (approved):
lgtm

PR #16687, opened 2018-01-04
Add melodic.

clalancette:
add melodic distribution file + add melodic to index


clalancette:
Travis fails because the tests are looking for melodic in the `index.yaml`, which is added by this PR :).

dirk-thomas:
This should be ok to be merged. It should also be easy to update the catch this corner case make it pass.

dirk-thomas:
Please only put the line which can actually raise in the `try` / `catch` block.

dirk-thomas (approved):

clalancette:
Thanks!  Merging.

PR #16755, opened 2018-01-10
Disable test pull request for roswww

furushchev:
Since this is already tested on travis, it is not necesary to test it also on buildfirm.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -8264,7 +8264,6 @@ repositories:
       url: https://github.com/ros-gbp/roswww-release.git
       version: 0.1.10-0
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/tork-a/roswww.git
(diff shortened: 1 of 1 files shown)


furushchev:
@dirk-thomas Thank you very much for merging!
It looks that CI on buildfirm is still enabled on that repo:
https://github.com/tork-a/roswww/pull/43
Is there any other procedure to be done?

dirk-thomas:
The change will only be applied during the nightly reconfiguration of the buildfarm. If the "problem" still exists tomorrow please comment here again.

mikaelarguedas:
the buildfarm is reconfigured every night (PST time) so it may take up to 24hours for the job to be deleted

furushchev:
@dirk-thomas @mikaelarguedas Thank you for clarifying! I confirmed that testing on buildfirm is now properly disabled. Thank you again!

PR #16759, opened 2018-01-11
Add azure-iothub-device-client pip package

anthturner:
This PR provides the azure-iothub-device-client Python pip package to ROS. The package provides a Python SDK for device clients connecting to the Microsoft Azure IoT Hub service. This would be a dependency for any projects making use of the IoT Hub service within a rospy package.

**PyPi Package Description:** https://pypi.org/project/azure-iothub-device-client/
**Source Repository:** https://github.com/Azure/azure-iot-sdk-python/tree/master/device

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8,6 +8,16 @@ adafruit-pca9685-pip:
   ubuntu:
     pip:
       packages: [adafruit-pca9685]
+azure-iothub-device-client-pip:
+  debian:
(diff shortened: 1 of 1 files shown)


wjwwood (approved):

tfoote (approved):
Looks fine. Though note that since it's pip only it will only work for source checkouts and cannot be used in a package to be released onto the buildfarm.

PR #16767, opened 2018-01-13
disable commit testing and enable pr testing for serial projects

wjwwood:


wjwwood:
Anyone have an idea why this fails? The tests pass for me locally.

dirk-thomas:
Line 5616 and following shows that it fails to validate the permission test for the pull request hooks:

> ERROR: Not enough permissions to setup pull request builds for repo [ wjwwood/serial_utils ] Please see http://wiki.ros.org/buildfarm/Pull%20request%20testing

I assume those tests are being skipped when you run the tests locally (@tfoote might be able to confirm this).

I noticed a request for the @ros-pull-request-builder user earlier but the link from the invite didn't work. Maybe try to invite the user again. After that is accepted the test should pass.

dirk-thomas:
@tfoote The error message for the permission problem is currently hidden in the folded output of the previous section. I think it would make sense to update the test to create a separate folding section for the it and make it visible to the user (similar as for the other tests).

wjwwood:
I accepted the invite. However, you cannot give admin writes to a repository owned by a user, so I did the alternative method on both.

It really sucks that you cannot see the error without expanding all of the sections...

Sorry I restarted the test on accident (on travis).

wjwwood:
The only difference I think is that I made it a collaborator on my wjwwood/serial repository when I was trying the normal method.

wjwwood:
Looks like it's giving me a soft-pass on wjwwood/serial:

> Warning: Push access detected but unable to verify manual hook configuration for repo [ wjwwood/serial ]. Please visit http://wiki.ros.org/buildfarm/Pull%20request%20testing and make sure hooks are setup.

I do have the hooks setup (correctly I believe) on both.

wjwwood:
I don't think you can check webhooks on a repository for which you're not an admin (or an admin of the org). Fundamentally, I don't think it's possible to make this check for personal repositories. How should we proceed? (I'm not really interested in making an org just for these two repositories)

dirk-thomas:
> It really sucks that you cannot see the error without expanding all of the sections...

Absolutely, the test should be "fixed" to show the error message / mention the folding group to look for the info.

> How should we proceed?

Since @tfoote implemented this I will let him respond.

wjwwood:
Bump, @tfoote should we remove this test until it can be fixed or make it disable itself when the repository in question is not part of a GitHub organization?

tfoote:
We can just disable it. It's only active when the PR is submitted by a committer. Otherwise, the GHPRB_TOKEN credentials necessary are firewalled away. I've removed it from the travis job config and the job is passing now.

wjwwood:
Thanks, should I create an issue to reenable it later? or is it not worth it?

tfoote:
Conceptually we can't reenable it without having a better firewall/credential management system that protects the credentials from the users code. Travis prevents the credentials from being used in a PR from non-colaborators. Otherwise they could just echo the environment variable, or email it to themselves etc in malicious code. If we hosted it on Jenkins we could use the credentials from a code path that is not available for the user to change. And then it can only positively validate if it has org admin access.

PR #16780, opened 2018-01-16
Add python-rx-pip package to python.yaml

jorgenfb:
Provide RxPy as a pip package to ROS. This package provides reactive extensions to python. (RxPy)[https://github.com/ReactiveX/RxPY] is _a library for composing asynchronous and event-based programs using observable collections and LINQ-style query operators in Python_.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2999,6 +2999,16 @@ python-ruamel.yaml:
     xenial: [python-ruamel.yaml]
     yakkety: [python-ruamel.yaml]
     zesty: [python-ruamel.yaml]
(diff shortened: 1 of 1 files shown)


tfoote:
This looks reasonable. It's failing CI because the keys are out of alphabetical order. And note that you will not be able to add dependencies on this from a released package because pip is not a native dependency for the package managers on any system.

tfoote:
@jorgenfb do you want to update this to pass CI so we can accept it?

jorgenfb:
Updated! Thank you!

mikaelarguedas (approved):
lgtm

PR #16785, opened 2018-01-19
Added pyrebase to rosdep/python.yaml

rdelfin:
This package provides a firebase client on python. It is useful for robots that use firebase as external storage.
Package: [https://pypi.python.org/pypi/Pyrebase](https://pypi.python.org/pypi/Pyrebase)


tfoote (changes requested):
If this is pip only please use `-pip` as a suffix for the key.

Also note that it will not be available for packaging and release into debian packages as a dependency.

rdelfin:
Understood. The key name should be fixed now

tfoote (approved):

mikaelarguedas (approved):
lgtm

PR #16807, opened 2018-01-22
add keys for catkin dependencies in melodic

dirk-thomas:


rosdep/python.yaml
```diff
@@ -497,6 +501,8 @@ python-catkin-pkg:
     pip:
       packages: [catkin-pkg]
   ubuntu:
+    artful: [python-catkin-pkg]
+    bionic: [python-catkin-pkg]
```
mikaelarguedas: should these have a `_python3` equivalent like it was done for the kinetic ? (same for mock and nose below)

rosdep/python.yaml
```diff
@@ -909,6 +915,8 @@ python-empy:
     pip:
       packages: [empy]
   ubuntu:
+    artful: [python-empy]
+    bionic: [python-empy]
```
mikaelarguedas: these should be collapsed to avoid having to add them manually in the future

rosdep/python.yaml
dirk-thomas: Since that depends on how we want Python 3 testing to happen I didn't add them at this time.

PR #16812, opened 2018-01-23
add python-gtk2 keys for missing ubuntu distributions

fmessmer:
see https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=python-gtk2

would this be a candidate to collapse the whole ubuntu entry to just `ubuntu: [python-gtk2]`?

PR #16813, opened 2018-01-23
collapse several ubuntu keys

fmessmer:
follow-up of #16812 

I guess I found several "collapse"-able keys...

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -112,13 +112,19 @@ pydocstyle:
     buster: [pydocstyle]
     stretch: [pydocstyle]
   ubuntu:
+    artful: [pydocstyle]
+    bionic: [pydocstyle]
     xenial: [pydocstyle]
+    zesty: [pydocstyle]
 pyflakes3:
   debian:
     buster: [pyflakes3]
     stretch: [pyflakes3]
   ubuntu:
+    artful: [pyflakes3]
+    bionic: [pyflakes3]
     xenial: [pyflakes3]
+    zesty: [pyflakes3]
 pyper-pip:
   ubuntu:
     pip:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I only looked at a few but they didn't seem to be collapsible. E.g. both `pydocstyle` and `pyflakes3` don't exist in earlier Ubuntu distros and therefore collapsing them would result in an incorrect mapping on older Ubuntu distros.

mikaelarguedas:
:+1: to @dirk-thomas' comment.
Another note: the keys that haven't been updated for a long time may not be used by any package anymore so having this information will be valuable the day we want to audit the unused keys. With that in mind I recommend updating only the rules that are needed for your development

ipa-fxm:
ok, so in case a package does not exist in earlier distro's (even if EOL'ed), it has to be specified per distribution for all future distributions - even if it is the same starting with the distro where it was introduced?

ist there no better mechanism to catch this?
I mean adding the keys for every new distro (not only ubuntu)...

mikaelarguedas:
We've been discussing several ways to mitigate that, for example a `<DISTROX>_and_newer` type of tag or distro comparators `distro >= trusty` to be able to collapse part of the rules while preserving the behavior for older distros.
As this require some significant changes in rosdep as well as a knwoledge of the order of the distros for each platform we didnt get around implementing it. See also https://xylem.readthedocs.io/en/latest/spec.html#any-version-and-version-ranges for design document exploring various syntax options.

Although there is no plan to spend time on this refactoring in the near future, we try to make incremental improvements as time permits (e.g. https://github.com/ros-infrastructure/rosdep/pull/573)

dirk-thomas:
> so in case a package does not exist in earlier distro's (even if EOL'ed), it has to be specified per distribution for all future distributions - even if it is the same starting with the distro where it was introduced?

Yes, that is not the case for keys you collapsed though. E.g. for the two keys I mentioned the package is not available on Trusty which is not EOL yet.

> is there no better mechanism to catch this?
> I mean adding the keys for every new distro (not only ubuntu)...as soon as the new distro gets release?

This has been a long standing missing feature. Recent I created ros-infrastructure/rosdep#573 to address this but the PR still requires an update to the REP before it can be merged.

ipa-fxm:
thanks for the explanations

I modified my commit
 - only collapse keys where [ubuntu packages](https://packages.ubuntu.com/) lists support for `trusty`, `xenial`, `zesty`, `artful` and `bionic`
 - added distributions where I did not collapse but distributions were missing

Feel free to merge or close unmerged - in case is not conform with the "audit"-explanation given by @mikaelarguedas

tfoote (approved):
With collapsing only rules that extend to at least the currently supported distros this won't change any behavior on currently supported distros and will only cause false positives on EOL'd distros which would have failed in the past so +1

mikaelarguedas (approved):
confirmed that all these are available for the supported platfoms and archs.
Thanks @ipa-fxm for doing this!

PR #16814, opened 2018-01-23
Adding apriltags2_ros to documentation index for kinetic

dmalyuta:


mikaelarguedas:
Notes:
- that the documentation jobs are ran every 72 hours so it may take that long for the information to be available.
- the already existing wiki page will not be updated automatically. To include information generated by the documentation job, the wiki page will need to include e.g PackageHeader macro `<<PackageHeader(MY_PACKAGE_NAME)>>`

PR #16824, opened 2018-01-25
Add python-rosrx-pip package to python.yaml

jorgenfb:
Provide ROSRx as a pip package. ROSRx provides reactive extensions to
ROS making it easier to work with ROS and reactive data streams.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2732,6 +2732,16 @@ python-rospkg:
     yakkety_python3: [python3-rospkg]
     zesty: [python-rospkg]
     zesty_python3: [python3-rospkg]
+python-rosrx-pip:
+  debian:
+    pip:
+      packages: [rosrx]
+  osx:
+    pip:
+      packages: [rosrx]
+  ubuntu:
+    pip:
+      packages: [rosrx]
 python-rpi.gpio:
   debian:
     buster: [python-rpi.gpio]


tfoote:
Might this make more sense to release through the standard ROS release pipeline since it's a ROS package? http://wiki.ros.org/bloom/Tutorials Doing so will allow much greater reuse as a dependency in other packages as well as providing a lot more infrastructure for documentation and indexing.

jorgenfb:
Yes, it seems reasonable to release this through ROS release pipeline, at least somewhere down the line when I get a change to look into the procedure you are linking to. Is a reason not to add it here as well?

tfoote:
In general we don't recommend releasing things through multiple channels. If you do that you can end up with inconsistent versions on different installations based on how they or someone else installed something in the past. In particular pip versions of things override deb versions of python packages, but the pip versions don't auto-update when the rest of the debian packages do so. Also switching people between a system dependency and a standard ROS dependency can be tricky as they may end up with both installed and on their path and then you need to worry about them conflicting/colliding depending on how the system is setup. Python's better at this than C++ but there are still corner cases.

In general we don't recommend releasing ROS packages on pypi because their dependencies cannot be fully satisfied through pip dependencies. Thus if you'd consider making the release as a full ROS package in the future, I'd recommend against adding it here to avoid future conflicts.

jorgenfb:
I do have an external dependency on a python library not available as a ROS-package or system dep, only through PIP. What is the best way to handle that?

mikaelarguedas:
Generally releasing the dependency as a ROS package and contacting the maintainers for releasing them into debian is the preferred way to go.
This way as soon as the dependency is available upstream you can stop releasing it and declare a rosdep key on it instead.

dirk-thomas:
What is the status of this ticket? Can it be closed?

PR #16946, opened 2018-02-08
[dep][py] Add GitPython.

130s:
- debian http://neuro.debian.net/pkgs/python-git.html
- fedora https://apps.fedoraproject.org/packages/GitPython/overview/
- gentoo https://packages.gentoo.org/packages/dev-python/git-python
- ubuntu https://packages.ubuntu.com/search?keywords=python-git

Looks like `pip` version is already added in https://github.com/ros/rosdistro/pull/13583. What are we supposed to do in these case?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1134,6 +1134,12 @@ python-gi:
   fedora: [pygobject3]
   gentoo: [dev-python/pygobject]
   ubuntu: [python-gi]
+python-git:
+  arch: [python2-gobject]
+  debian: [python-git]
+  fedora: [GitPython]
+  gentoo: [dev-python/git-python]
+  ubuntu: [python-git]
 python-github-pip:
   osx:
     pip:


dirk-thomas (approved):
> Looks like `pip` version is already added in https://github.com/ros/rosdistro/pull/13583. What are we supposed to do in these case?

I don't think anything needs to be done for the existing pip entry.

rosdep/python.yaml
```diff
@@ -1134,6 +1134,12 @@ python-gi:
   fedora: [pygobject3]
   gentoo: [dev-python/pygobject]
   ubuntu: [python-gi]
+python-git:
+  arch: [python2-gobject]
+  debian: [python-git]
+  fedora: [GitPython]
+  gentoo: [git-python]
```
nuclearsandwich: Shouldn't this package name be prefixed with `dev-python/` per https://packages.gentoo.org/packages/dev-python/git-python (currently offline)

130s: I did not know that, thanks. Updated.

mikaelarguedas (approved):

PR #16954, opened 2018-02-11
add rc_visard to indigo

flixr:
manual pull request for first release of rc_visard via bloom...

Side-note/question:
I have two-factor authentication enabled, so creation of the oauth token didn't work.
But I couldn't find any info on which scopes are needed when I manually create a new personal access token for bloom.


dirk-thomas:
> Side-note/question:
I have two-factor authentication enabled, so creation of the oauth token didn't work.

Please ask you question on answers.ros.org or fill a ticket in the bloom repository.

PR #16976, opened 2018-02-13
Add Gentoo Missing Gentoo dependencies

allenh1:
Filed in favor of ros2/rosdistro#71

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3900,6 +3900,7 @@ python3-venv:
   debian: [python3-venv]
   ubuntu: [python3-venv]
 python3-yaml:
+  debian: [python3-yaml]
   gentoo: [dev-python/pyyaml]
   ubuntu: [python3-yaml]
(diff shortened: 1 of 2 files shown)


rosdep/python.yaml
```diff
@@ -3890,11 +3894,13 @@ python3-ruamel.yaml:
     zesty: [python3-ruamel.yaml]
 python3-setuptools:
   debian: [python3-setuptools]
+  gentoo: [dev-python/setuptools]
   ubuntu: [python3-setuptools]
 python3-venv:
   debian: [python3-venv]
   ubuntu: [python3-venv]
 python3-yaml:
+  gentoo: [dev-python/pyyaml]
```
mikaelarguedas: Can you please add the missing debian entry here as well?

allenh1: Yep. Will do.

mikaelarguedas (approved):
lgtm

PR #16982, opened 2018-02-14
Remove robotican from indexing.

tfoote:
There are forked copies of packages masking released packages.

Indigo: https://github.com/robotican/robotican/issues/21
Kinetic: https://github.com/robotican/armadillo2/issues/1


tfoote:
This reverts: #11059 and #16676 They original submitter is now ghost.

FYI @RoboTiCan

The documentation will be restored within the next 3 days as the doc jobs cycle through. It was a race condition as to which jobs ran most recently to get the correct link. With it removed they will return within one cycle.

PR #16992, opened 2018-02-15
Adding yoctopuce_altimeter to documentation index for kinetic

anja-sheppard:
I'd like my repo yoctopuce_altimeter to be indexed and documented on ros.org.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -11694,14 +11694,9 @@ repositories:
       type: git
       url: https://github.com/amstrudy/yoctopuce_altimeter.git
       version: kinetic
-    release:
-      tags:
-        release: release/kinetic/{package}/{version}
-      url: https://github.com/amstrudy/yoctopuce_altimeter.git
-      version: 0.1.0-1
     source:
       type: git
-      url: https://github.com/amstrudy/yoctopuce_altimeter.git
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git  
       version: kinetic
     status: maintained
   youbot_description:
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -11689,6 +11689,21 @@ repositories:
       url: https://github.com/yujinrobot/yocs_msgs.git
       version: kinetic
     status: developed
+  yoctopuce_altimeter:
+    doc:
+      type: git
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git
+      version: kinetic
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git
+      version: kinetic
+    status: maintained
   youbot_description:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -11689,6 +11689,21 @@ repositories:
...
     status: developed
+  yoctopuce_altimeter:
+    doc:
+      type: git
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git
+      version: kinetic
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git
```
tfoote: This is not a release repository. It doesn't look like you have made a release. If you want to have this indexed and documented, but haven't made a release you should skip the `release` section here.

kinetic/distribution.yaml
```diff
@@ -11689,6 +11689,21 @@ repositories:
...
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git
+      version: kinetic
```
tfoote: This is not a valid branch name. The only branch is `master`.

kinetic/distribution.yaml
```diff
@@ -11689,6 +11689,21 @@ repositories:
       url: https://github.com/yujinrobot/yocs_msgs.git
       version: kinetic
     status: developed
+  yoctopuce_altimeter:
+    doc:
+      type: git
+      url: https://github.com/amstrudy/yoctopuce_altimeter.git
+      version: kinetic
```
tfoote: This is not a valid branch name. The only branch is `master`.

tfoote (dismissed):

tfoote:
For reference the tutorial for adding a package to the index: http://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation

The release section is added by bloom when you make a release: http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

kinetic/distribution.yaml
amstrudy: Sorry, I'm a bit new to this. I see what you're saying about the release section, and I fixed that. I don't understand what you mean when you say "This is not a valid branch name. The only branch is master." Could you please elaborate on that? Thanks again.

mikaelarguedas: The github repository listed in the previous line has only one branch, named `master`. And the version entry should be a valid branch of the repository. So it shouldn't be `kinetic` but `master` instead

amstrudy:
Thank you for the revisions. I have fixed all of the above issues and updated the `distribution.yaml` file and committed again.

kinetic/distribution.yaml
amstrudy: Thanks! I ended up just adding a kinetic branch.

dirk-thomas:
One of the lines has trailing spaces which are not valid. Please see the Travis output for details.

Just out of curiosity: did Travis send you an email about the failed CI? According to the configuration file it is supposed to do so and I would like to minimize our need for comments like that (just repeating what was already reported by the notification).

amstrudy:
@dirk-thomas Thanks for the heads-up - I didn't get an email from Travis so I wouldn't have seen that otherwise. I fixed a trailing whitespace and hopefully Travis will pass this time around!

dirk-thomas:
@amstrudy I was wrongly assuming notification emails would be sent but as it turns out that is not the case for any PR builds on Travis. So please check on PR you created yourself to ensure the CI builds passed successfully.

PR #16998, opened 2018-02-16
Add More Gentoo Deps

allenh1:
I created the `dev-python/pydocstyle` ebuild personally, and it now exists in the ROS Overlay. Also, the `ros-pkg-resources` module is just a part of setuptools on Gentoo.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3805,6 +3805,8 @@ openni-dev:
   fedora: [openni-devel]
   gentoo: [dev-libs/OpenNI]
   ubuntu: [libopenni-dev]
+opensplice:
+  gentoo: [sci-libs/opensplice]
 openssh-client:
   debian: [openssh-client]
   fedora: [openssh-clients]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -115,6 +115,7 @@ pydocstyle:
   debian:
     buster: [pydocstyle]
     stretch: [pydocstyle]
+  gentoo: [dev-python/pydocstyle]
   ubuntu:
     artful: [pydocstyle]
     bionic: [pydocstyle]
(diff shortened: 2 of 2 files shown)


allenh1:
The only odd change here is the opensplice addition. This is needed for `superflore` to resolve the key correctly.

If preferred, I could hard-code the resolution, but I would really rather not.

tfoote:
If opensplice is available upstream in an appropriate version we might as well leverage that.

nuclearsandwich:
> The only odd change here is the opensplice addition. This is needed for superflore to resolve the key correctly.

This is only odd because I messed up and forgot to upstream the addition of an opensplice key from the should-be-defunct ros2 rosdep tree https://github.com/ros2/rosdistro/commit/1871b9e565b9c9e6d3a846a671ee6c64f712b623
I think the reason I hesitated is that the package is in the OSRF and ROS 2 repositories rather than upstream. Should we add the ubuntu key as well?

allenh1:
> Should we add the ubuntu key as well?

I don't think it's unreasonable to add it. While it's not safe to assume that rosdep being installed => they have set up the OSRF / ROS 2 repositories, I don't think we're necessarily responsible for maintaining the other installation methods. That's the approach I've been going with for the Gentoo installs, at least.

nuclearsandwich:
In any case I think we can resolve that issue separate from getting Gentoo operational. I can PR that separately for discussion.

nuclearsandwich (approved):

mikaelarguedas:
> I think the reason I hesitated is that the package is in the OSRF and ROS 2 repositories rather than upstream. Should we add the ubuntu key as well?

If we expect people to be able to create / release packages that depend on it if should either:
- be available upstream
- be imported in the ROS repo and distributed by us

Without this it cannot be added to the rosdep database

mikaelarguedas:
> If opensplice is available upstream in an appropriate version we might as well leverage that.

@allenh1 the gentoo packages website is still down and I can't find this package. Can you provide a link and a version number?

nuclearsandwich:
> Can you provide a link and a version number?

That's a good point. Ardent targets opensplice 6.7.

allenh1:
> the gentoo packages website is still down and I can't find this package. Can you provide a link and a version number?

@mikaelarguedas Sorry for the delay in response, had to take an exam.

There is no upstream Opensplice for Gentoo, only the one I wrote in the overlay. It's for 6.7 as well.

PR #17020, opened 2018-02-21
Add keys for cpprest

jlack1987:
Add key for cpprest. Was not able to find package names for fedora, gentoo, or arch. `cpprest` is just C++ REST library and we use it for an `HTTP` server

Ubuntu: https://packages.ubuntu.com/xenial/libdevel/libcpprest-dev
Debian: https://packages.debian.org/search?keywords=libcpprest-dev
Fedora: https://copr.fedorainfracloud.org/coprs/c72578/cpprest/

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1497,13 +1497,13 @@ libconsole-bridge-dev:
 libcpprest-dev:
   debian:
     buster: [libcpprest-dev]
-    sid: [libcpprest-dev]
     stretch: [libcpprest-dev]
   fedora: [cpprest-devel]
   ubuntu:
     artful: [libcpprest-dev]
     bionic: [libcpprest-dev]
     xenial: [libcpprest-dev]
+    yakkety: [libcpprest-dev]
     zesty: [libcpprest-dev]
 libdbus-dev:
   arch: [dbus-core]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1494,6 +1494,17 @@ libconsole-bridge-dev:
   opensuse: [libconsole_bridge0]
   slackware: [console_bridge]
   ubuntu: [libconsole-bridge-dev]
+libcpprest-dev:
+  debian:
+    buster: [libcpprest-dev]
+    sid: [libcpprest-dev]
+    stretch: [libcpprest-dev]
+  fedora: [cpprest-devel]
+  ubuntu:
+    artful: [libcpprest-dev]
+    bionic: [libcpprest-dev]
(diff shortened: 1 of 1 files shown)


mikaelarguedas (changes requested):
@jlack1987 your description mentions that you couldnt find a package name for fedora but it adds a fedora rule. Did you find the package after all ?

As per [guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions), please provide a brief description of the package and what you use it for as well as links to the package for the distributions you add.

Can you please rename the key to match the debian package name? As this is a dev package it should contain the `-dev` suffix, this way if users want to use cpprest as a run dependency they can provide a cpprest key containing only the binaries.

jlack1987:
oh whoops, didn't mean to put that I couldn't find it for fedora. I did find it for fedora, just not arch or gentoo. `cpprest` is just C++ REST library and we use it for an `HTTP` server.

mikaelarguedas:
Sounds good, do you have a link for the fedora package?

Please use libcpprest **-dev** as the key name.
Also this package doesnt seem to be available to ubuntu prior to xenial (and debian prior to Stretch). So you'll need to specify the exect distro names this package is available on.

Links for ubuntu and debian:
Ubuntu: https://packages.ubuntu.com/search?keywords=libcpprest-dev&searchon=names&exact=1&suite=all&section=all
Debian: https://packages.debian.org/search?keywords=libcpprest-dev&searchon=names&exact=1&suite=all&section=all

jlack1987:
Sorry yeah it should all be good to go now.

mikaelarguedas:
if this is a dev package, the fedora rule should have the `-devel` suffix

jlack1987:
I haven't ever used fedora myself so I have no clue. I was just going off the link I found [here](https://copr.fedorainfracloud.org/coprs/c72578/cpprest/packages/)

mikaelarguedas:
Fedora: looks like `cpprest-devel` is what we want here (http://rpmfind.net/linux/rpm2html/search.php?query=cpprest-devel&submit=Search+...&system=Fedora&arch=)

Ubuntu/Debian: The previous comment about listing only the distributions where the package is available has not been addressed yet. Please call out specific Ubuntu / Debian versions where this package is available.

(almost there :) )

jlack1987:
Ok try again, thx for the feedback

tfoote (approved):
I removed sid and added yakkety for you. Now it looks good to me.

mikaelarguedas (approved):
thanks @jlack1987 for iterating

PR #17045, opened 2018-02-27
Adding realsense_ros_camera to documentation index for Kinetic and Indigo distro

icarpis:
I'd like `realsense_ros_camera` to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -10638,6 +10638,12 @@ repositories:
       url: https://github.com/timn/ros-rcll_ros_msgs.git
       version: master
     status: developed
+  realsense2_camera:
+    doc:
+      type: git
+      url: https://github.com/intel-ros/realsense.git
+      version: development
+    status: maintained
   realsense_camera:
     doc:
       type: git
@@ -10653,12 +10659,6 @@ repositories:
       type: git
       url: https://github.com/intel-ros/realsense.git
       version: indigo-devel
-    status: maintained
-  realsense_ros_camera:
-    doc:
-      type: git
-      url: https://github.com/intel-ros/realsense.git
-      version: development
     status: maintained
   realtime_tools:
     doc:
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -10654,6 +10654,12 @@ repositories:
       url: https://github.com/intel-ros/realsense.git
       version: indigo-devel
     status: maintained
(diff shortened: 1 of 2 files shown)


indigo/distribution.yaml
```diff
@@ -10654,6 +10654,12 @@ repositories:
       url: https://github.com/intel-ros/realsense.git
       version: indigo-devel
     status: maintained
+  realsense_ros_camera:
+    doc:
+      type: git
+      url: https://github.com/intel-ros/realsense.git
```
tfoote: This looks to be a duplicate of: https://github.com/ros/rosdistro/pull/17045/files#diff-4bbb8e35ff5dc5e7f2d7b4d00374533fR10644 with a different branch. Do you want to change the documentation branch? If so then changing lin 10645 would be the right thing to do.

tfoote (changes requested):
It looks like realsense_ros_camera is already submitted and indexed. Can you explain what you're looking for to be added?

tfoote:
It looks like the driver has been renamed in the new branch?
Related to that in general our naming convention is to not include 'ros' in the ROS package name as it's redundant. http://www.ros.org/reps/rep-0144.html http://wiki.ros.org/ROS/Patterns/Conventions

icarpis:
The [realsense_camera](http://wiki.ros.org/realsense_camera) and [realsense_ros_camera](http://wiki.ros.org/realsense_ros_camera) are independent packages. The first one works with [librealsense1](http://wiki.ros.org/librealsense) and the second one works with [librealsense2](http://wiki.ros.org/librealsense2). They are both located at the same [repository](https://github.com/intel-ros/realsense) but in different branches. The `realsense_camera` is located at the [indigo-devel](https://github.com/intel-ros/realsense/tree/indigo-devel) branch and the `realsense_ros_camera` is located at the [development](https://github.com/intel-ros/realsense/tree/development) branch.

dirk-thomas:
> The first one works with librealsense1 and the second one works with librealsense2.

Maybe choose a different repository name then? The additional `ros` doesn't express the difference. Maybe `realsense2_camera`?

icarpis:
Ok, I'll rename the repository name and apply the change to this PR.

dirk-thomas:
Please see the Travis log for errors. The new entry needs to be inserted before `realsense_camera`.

tfoote (approved):

PR #17064, opened 2018-03-01
add kaitaistruct for pip

deadundead:


dirk-thomas (approved):

mikaelarguedas:
This package is available in recent [Debian](https://packages.debian.org/search?keywords=python-kaitaistruct&searchon=names&exact=1&suite=all&section=all) and [Ubuntu](https://packages.ubuntu.com/search?keywords=python-kaitaistruct&searchon=names&exact=1&suite=all&section=all).
What platform are you using?
Is it possible to add the regular APT key for these platforms?

tfoote:
If there are native packages available we should use them not the pip installer. https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

deadundead:
I'm using ubuntu 16.04 LTS
`sudo apt install python-kaitaistruct `
gives 

> unable to locate package

And it works like a charm installing it from pip.

mikaelarguedas:
> I'm using ubuntu 16.04 LTS
sudo apt install python-kaitaistruct
gives

>    unable to locate package

Yes, as pointed out in the above comment the package is available only in the most recent ubuntus (artful and bionic) and debian (buster).
So I would recommend to add a `python-kaitaistruct` key that uses the pip package for distributions from `trusty` to `zesty` and the apt version for `artful` and `bionic`.
Same thing for debian: use pip for `jessie` and `stretch` and apt for `buster`

mikaelarguedas:
@deadundead gentle :bellhop_bell:: do you think you can address the pending comments so that we can get this merged?

mikaelarguedas:
Closing this due to no response. @deadundead feel free to open another PR once this branch is updated to address the review comments

PR #17079, opened 2018-03-05
Adding python-voluptuous

mr337:
PR to add the JSON validation library [Voluptuous](https://github.com/alecthomas/voluptuous). We use this to validate JSON commands coming from external sources (laptop, cloud etc).

Ubuntu: https://packages.ubuntu.com/search?keywords=python-voluptuous&searchon=names&suite=all&section=all
Fedora: https://admin.fedoraproject.org/pkgdb/packages/volup%2A/
Debian: https://packages.debian.org/search?keywords=voluptuous&searchon=names&suite=stable&section=all
Gentoo: https://packages.gentoo.org/packages/dev-python/voluptuous

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3674,6 +3674,11 @@ python-vlc-pip:
   ubuntu:
     pip:
       packages: [python-vlc]
+python-voluptuous:
+  debian: [python-voluptuous]
+  fedora: [python-voluptuous]
+  gentoo: [dev-python/voluptuous]
+  ubuntu: [python-voluptuous]
 python-vtk:
   arch: [vtk]
   debian: [python-vtk]


dirk-thomas (approved):

rosdep/python.yaml
```diff
@@ -3674,6 +3674,17 @@ python-vlc-pip:
   ubuntu:
     pip:
       packages: [python-vlc]
+python-voluptuous:
+  debian:
+    jessie:
```
mikaelarguedas: This package is available in [all versions of Debian](https://packages.debian.org/search?keywords=python-voluptuous&searchon=names&exact=1&suite=all&section=all), can you please use the apt version everywhere?

rosdep/python.yaml
```diff
@@ -3674,6 +3674,17 @@ python-vlc-pip:
...
+      pip:
+        packages: [voluptuous]
+    stretch: [python-voluptuous]
+  fedora:
+    pip:
+      packages: [voluptuous]
```
mikaelarguedas: This package is also available in the fedora package database: https://admin.fedoraproject.org/pkgdb/package/rpms/python-voluptuous/

mikaelarguedas (changes requested):
Thanks @mr337 for the description and the links!
It looks that this package is available upstream for all the targeted platforms so it would be better to use the upstream version rather than the pip one

mr337:
@mikaelarguedas Fixed. Little frustrated I'm still learning how to search packages properly on Fedora and Debian.

mikaelarguedas:
> mikaelarguedas Fixed

Thanks! :+1: 

> I'm still learning how to search packages properly on Fedora and Debian

Yeah it can get a bit tricky, in general for debian and ubuntu I select: distribution "any" to get the result for all distributions and not only the currently active one.
For fedora: They use the `*` wildcard for matching package names. So in the case of `python-voluptuous` you need to search for either `*python-voluptuous*` or `*voluptuous*`. As you searched for `volup*` it didnt return any match but `*volup*` should have returned all the packages matching that regex.

Hope this helps

mikaelarguedas (approved):

PR #17080, opened 2018-03-05
Added libimage-exiftool-perl

HEllRZA:
We are using libimage-exiftool-perl, to get extended exif data from images,


dirk-thomas (approved):

mikaelarguedas:
@HEllRZA CI fails because of trailing whitespaces in this PR.
This package is available on gentoo as [`media-libs/exiftool`](https://packages.gentoo.org/packages/media-libs/exiftool). Can you please add it to this PR as well?

mikaelarguedas (approved):
thanks @HEllRZA for iterating

PR #17110, opened 2018-03-12
Modernize Python 2 code to get ready for Python 3

cclauss:
Make the minimal, safe changes required to convert the repo's code to be syntax compatible with both Python 2 and Python 3.  There may be other changes required to complete a port to Python 3 but this PR is a minimal, safe first step.

Run: [futurize --stage1 -w **/*.py](http://python-future.org/futurize_cheatsheet.html)

See Stage 1: "safe" fixes http://python-future.org/automatic_conversion.html#stage-1-safe-fixes

* __print()__ is a function in Python 3.
* __file()__ was removed in Python 3 so this PR substitutes __with open() as__ to ensure file handles are closed and promptly garbage collected on all implementations of Python.
* Added Python 3.6 in __allow_failures__ mode as a parallel run in the Travis CI tests.
* Added [__flake8__](http://flake8.pycqa.org) to the Travis CI tests to find __syntax errors__ and __undefined names__.

@mikepurvis @dirk-thomas

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 language: python
 
-python: "2.7"
+python:
+  - "2.7"
+  - "3.6"
+
+matrix:
+  allow_failures:
+    - python: "3.6"
 
 sudo: false
 
--- a/test/test_url_validity.py
+++ b/test/test_url_validity.py
@@ -11,7 +11,10 @@ import os
 import subprocess
 import sys
 import unittest
-from urlparse import urlparse
+try:
+    from urllib.parse import urlparse  # Python 3
+except ImportError:
+    from urlparse import urlparse      # Python 2
 
 import rosdistro
 from scripts import eol_distro_names
--- a/doc/scripts/distro_to_rosinstall.py
+++ b/doc/scripts/distro_to_rosinstall.py
@@ -1,5 +1,6 @@
 #! /usr/bin/env python
 
+from __future__ import print_function
 import os
 import sys
 import yaml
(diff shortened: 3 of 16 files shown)


clalancette (commented):
This looks generally fine to me.  The one place we may need to discuss is the flake8 calls in `.travis.yml`.  I'm fine with the addition of flake8, and even the warnings for full flake8 compliance, but others may feel differently about that.

cclauss:
Removed __flake8__ tests.

.travis.yml
```diff
@@ -1,6 +1,12 @@
 language: python
 
-python: "2.7"
+python:
+  - "2.7"
+  - "3.6"
+
+matrix:
+  allow_failures:
+    - python: "3.6"
```
dirk-thomas: Rather than allowing failures I would prefer to have the tests passing for Python 3.

test/test_url_validity.py
```diff
@@ -11,7 +11,10 @@
 import subprocess
 import sys
 import unittest
-from urlparse import urlparse
+try:
+    from urllib.parse import urlparse  # Python 3
+except ImportError:
+    from urlparse import urlparse      # Python 2
```
dirk-thomas: Please use two space before the comment for pep8 compliance.

dirk-thomas (changes requested):
Neither of the future imports of the print function seem necessary. Therefore I would prefer to not add them in the first place.

.travis.yml
cclauss: I have seen this mistake made too many times.  Fixing all Python 3 issues in a single PR is a mistake because it takes too long to test the whole in Python 2 (See the Facebook presos at PyCon).  For example, this PR already touches 16 files and is so complex to review that it has not progressed in the three months since it would open.

I would highly recommend that we take an _incremental_ approach to Python 3 compatibility using this sequence of actions:
1. Get the whole codebase to work on Python 2 with no Python 3 syntax errors or undefined names (we need to prove that Python 2 test pass even if Python 3 tests still fill.)
2. Get the whole codebase to work on both Python 2 and Python 3.  (All tests pass.)

We are still on step 1 and will not start on step 2 until _after_ step 1 is complete.

dirk-thomas: > For example, this PR already touches 16 files and is so complex to review ...

That is simply not the case. As commented several of the changes are unnecessary (print future) or unrelated to Python 3 (scoped file handle). The remaining changes are trivial parenthesis for print statements as well as one conditional import. The patch is extremely easy to review.

I am sorry for the delay. The delay was mostly due to one person starting the review but then not following up on it.

Please continue to address the remaining Python 3 problems step by step.

PR #17118, opened 2018-03-13
dartsim: 6.3.1-0 in 'kinetic/distribution.yaml' [bloom]

mxgrey:
Increasing version of package(s) in repository `dartsim` to `6.3.1-0`:

- upstream repository: https://github.com/dartsim/dart.git
- release repository: https://github.com/dartsim/ros-dartsim-release.git
- bloom version: `0.6.2`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1633,7 +1633,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/dartsim/dart.git
-      version: v6.3.1
+      version: release-6.3
   dataspeed_can:
     doc:
       type: hg
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1623,6 +1623,17 @@ repositories:
       url: https://github.com/leggedrobotics/darknet_ros.git
       version: master
     status: developed
+  dartsim:
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/dartsim/ros-dartsim-release.git
+      version: 6.3.1-0
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/dartsim/dart.git
+      version: v6.3.1
   dataspeed_can:
     doc:
       type: hg


mxgrey:
Travis seems to be failing with the following complaint:

```
AssertionError: [u"In file '''/home/travis/build/ros/rosdistro/kinetic/distribution.yaml''': Could not validate source entry for repo dartsim with error [[[Could not validate repository with url https://github.com/dartsim/dart.git and version v6.3.1 from entry at line 1632]]]"] is not false
```

The url https://github.com/dartsim/dart.git seems to have no trouble linking to the DART source code repo, and we've created a [tag with the name v6.3.1](https://github.com/dartsim/dart/releases/tag/v6.3.1), so I'm not really sure why this error would be occurring.

mxgrey:
I'm not a python expert, but it looks like the error is because [on line 146 of test_url_validity.py](https://github.com/ros/rosdistro/blob/b3c371f7d6a80ce738de6050436f64dd62b274ea/test/test_url_validity.py#L146), the function `check_source_repo_entry_for_errors(repo['source'])` is called with the default value for the second argument `tags_valid=False`.

Because of this, the script doesn't check for any tags with the name `v6.3.1`. Is this the expected behavior? The `bloom-release` instructions seemed to indicate that tags can be used to mark releases, but this script seems to be hard-coded against that, as far as I'm able to tell.

mxgrey:
After thinking about it further, I guess the `source` information in this case is referring to the source code development branch rather than the source code of the specific release in the distro.

I can change the entry to point to our `release-6.3` branch instead of the specific release tag.

mikaelarguedas:
Yes the source entry expects a branch and not a tag, as it's used to test new commits on the development branch.

For `test_pull_requests: true` : Have you followed the [instructions](http://wiki.ros.org/buildfarm/Pull%20request%20testing) to provide access to this repo or the org unit to the @ros-pull-request-builder user? If not, please either remove this line or provide the necessary access as described in the wiki page.

mxgrey:
Yep, I invited @ros-pull-request-builder to be a collaborator with write access, and the invitation is currently awaiting a reply. I also added the webhook for http://build.ros.org/ghprbhook/.

It didn't occur to me initially that the "source" was tied to the PR testing; I was imagining that it was just a reference to the source code for the specific release that we were putting into the distro.

PR #17141, opened 2018-03-16
Add graphviz-dev to base.yaml

miguelprada:
Graphviz development package. Needed by a not yet publicly available catkinized version of [rfsmTools](https://github.com/robotology/rfsmTools).

- https://packages.ubuntu.com/xenial/libgraphviz-dev
- https://packages.debian.org/stretch/libgraphviz-dev

Package exists in all Ubuntu and Debian versions listed in packages.ubuntu.com and packages.debian.org respectively, ~~but I've not checked all supported Ubuntu versions (i.e. haven't checked utopic, vivid, wily, yakkety and zesty)~~. Is there any easy way to check other than manually going through the releases in http://old-releases.ubuntu.com?

I also could not find the package in Fedora, Gentoo or OSX, but I'm not very familiar with those platforms and I may be missing something.

**Edit:** I ended up manually checking those, but the question stands. Is there an easier way?

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1003,6 +1003,7 @@ graphviz:
   ubuntu: [graphviz]
 graphviz-dev:
   debian: [libgraphviz-dev]
+  fedora: [graphviz-devel]
   ubuntu: [libgraphviz-dev]
 gringo:
   debian: [gringo]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1001,6 +1001,9 @@ graphviz:
   rhel: [graphviz]
   slackware: [graphviz]
   ubuntu: [graphviz]
+graphviz-dev:
+  debian: [libgraphviz-dev]
+  ubuntu: [libgraphviz-dev]
 gringo:
   debian: [gringo]
   fedora: [gringo]


mikaelarguedas:
@miguelprada There is no easier way that I know of. For new keys, it is not necessary to check EOL ubuntu distributions as no package was using these keys when the distributions were active.

Regarding the other platforms:
I think the equivalent is gentoo is: [media-gfx/graphviz](https://packages.gentoo.org/packages/media-gfx/graphviz)
and in fedora: [graphviz](https://admin.fedoraproject.org/pkgdb/package/rpms/graphviz/)

miguelprada:
`graphviz` in fedora [doesn't contain headers](https://apps.fedoraproject.org/packages/graphviz/contents), which are [part of `libgraphviz-dev` in Ubuntu](https://packages.ubuntu.com/xenial/amd64/libgraphviz-dev/filelist), so I'm not sure they're equivalent.

The link to the [gentoo `graphviz` package files](http://www.portagefilelist.de/site/query/listPackageVersions/?category=media-gfx&package=graphviz&do#result) currently returns an error.

mikaelarguedas:
> graphviz in fedora doesn't contain headers, which are part of libgraphviz-dev in Ubuntu, so I'm not sure they're equivalent.

Sorry that was not a very precise explanation. `graphviz` is the package with the binaries (lib, exe), if you want the development version (with headers, shared files etc) in fedora the suffix `-devel` is used. So in this case [`graphviz-devel`](https://apps.fedoraproject.org/packages/graphviz-devel/contents) will be the "equivalent" package.

> The link to the gentoo graphviz package files currently returns an error.

Looks like another bug of the gentoo website. Feel free to not add gentoo then

miguelprada:
> Sorry that was not a very precise explanation. graphviz is the package with the binaries (lib, exe), if you want the development version (with headers, shared files etc) in fedora the suffix -devel is used. So in this case graphviz-devel will be the "equivalent" package.

Noted for the future. Thanks for the help @mikaelarguedas!

mikaelarguedas (approved):
lgtm, thanks @miguelprada for the description, the links and the quick iterations :+1:

miguelprada:
Thanks!

PR #17157, opened 2018-03-16
change mapping of python-imaging since the package was removed from bionic

dirk-thomas:
To address the failing CI from ros/ros_comm#1350 as well as the now failing binary job (http://build.ros.org/view/Msrc_uB/job/Mbin_uB64__rosbag__ubuntu_bionic_amd64__binary/3/).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1399,7 +1399,7 @@ python-imaging:
       packages: [python-pillow]
   ubuntu:
     artful: [python-imaging]
-    bionic: [python-imaging]
+    bionic: [python-pil]
     lucid: [python-imaging]
     maverick: [python-imaging]
     natty: [python-imaging]


mikaelarguedas:
As `python-pil` is available on all Melodic targeted platforms, should we update the other ones ? maybe just artful as if doesn;'t share any platform with the other ROS distros ?

dirk-thomas:
Please feel free to propose related changes. I just want this specific change to get merged since it blocks building melodic packages.

dirk-thomas:
Please feel free to propose related changes. I just want this specific change to get merged since it blocks building melodic packages.

I don't see why any other mapping needs to change though.

mikaelarguedas:
IDK it seems strange to target a different package for only one platform while the newly targeted package is available on all active Ubuntu and Debian distributions.
If you're confident that `python-pil` in bionic provides exactly the same behavior as `python-imaging` on other platforms and that using `python-pil` everywhere would break things I guess that's fine to merge as is

tfoote:
Looking at python-imaging: https://packages.ubuntu.com/artful/python-imaging it appears to have just been a basic wrapper package that depends on python-pil for the full implementation.

`python-pil` has been available as far back as trusty: https://packages.ubuntu.com/artful/python-pil

We could consider switching all of the rules to python-pil but I think this is the more conservative approach without deeper research.

PR #17166, opened 2018-03-19
python-pathos: add package in python.yaml

Minipada:
This is a standard pip package, located [here](https://pypi.python.org/pypi/pathos).

pathos is used to provide interfaces to do distributed computing and to do computations in parallel. We use it for the latter using the pool interface.

From the docs:
`The high-level pool.map interface, yields a map implementation that hides the RPC internals from the user. With pool.map, the user can launch their code in parallel, and as a distributed service, using standard python and without writing a line of server or parallel batch code.`

I am conscious it is not much but we use it for less than 5 LOC in our package for now.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1998,7 +1998,7 @@ python-path.py:
   ubuntu:
     pip:
       packages: [path.py]
-python-pathos:
+python-pathos-pip:
   debian:
     pip:
       packages: [pathos]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1998,6 +1998,13 @@ python-path.py:
   ubuntu:
     pip:
       packages: [path.py]
+python-pathos:
+  debian:
+    pip:
+      packages: [pathos]
+  ubuntu:
+    pip:
+      packages: [pathos]
 python-pathtools:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
If the package is only available via pip, please use the `-pip` suffix for the key as per [Contribution Guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules).

Can you please also [`briefly describe the package being added and what use case you want to use it for`](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions)

Thanks!

Minipada:
Thanks for the feedback. I added a new commit related to the `-pip` extension, I didn't want to squash so the conversation is kept clean but squashing may be better. 

I also edited the description. I know it's short, but we use less than 5 lines of code of this library for now

mikaelarguedas:
thanks @Minipada,
I can't see the edited description, did you forget to submit the change ?

Minipada:
It seems so, sorry about it. It's done

tfoote (approved):

mikaelarguedas (approved):
lgtm, thanks @Minipada

PR #17232, opened 2018-03-22
google-cloud-bigquery: add package in python.yaml

Minipada:
Add [google-cloud-bigquery](https://pypi.python.org/pypi/google-cloud-bigquery) pip package. It contains library to communicate with google big query.

Google big query us used to store and do analytics of large datasets. It provides the hardware, infrastructure, UI and command line tools.

We use it to optimize map generation parameters. There are also third-party tools to interact with BigQuery, such as visualizing the data or loading the data.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1194,6 +1194,13 @@ python-gnupg:
   fedora: [python-gnupg]
   gentoo: [dev-python/python-gnupg]
   ubuntu: [python-gnupg]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Pypi link: https://pypi.python.org/pypi/google-cloud-bigquery

mikaelarguedas (approved):
lgtm, thanks @Minipada

PR #17243, opened 2018-03-22
Update Gentoo gmock dependency

allenh1:
It seems `gmock` was absorbed by `gtest`.

```
# equery belongs /usr/lib64/libgmock.so
 * Searching for /usr/lib64/libgmock.so ... 
dev-cpp/gtest-1.8.0-r1 (/usr/lib64/libgmock.so)
```

I'm not exactly sure how the other distributions are affected, but the Ubuntu gtest version seems to be up to date (at `1.8.0`).

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -962,7 +962,7 @@ google-mock:
   debian: [google-mock]
   fedora: [gmock-devel]
   freebsd: [googlemock]
-  gentoo: [dev-cpp/gmock]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
My understanding it that the 2 projects merged in v1.8.0.
now `google-mock` and `libgtest-dev` are just placeholders for backward compatibility that are wrapping the `googletest`package that contains both

mikaelarguedas (approved):

PR #17247, opened 2018-03-23
python-sklearn-pip: add package in python.yaml

Minipada:
Add [python-scikit-learn](http://scikit-learn.org) pip package. It contains libraries and tools to data mining and data analysis. It's built on NumPy, SciPy, and matplotlib.

It is mainly used for machine learning and features various classification, regression and clustering algorithms.

We use it in our machine learning process to optimize map generation parameters.

Pypi link: https://pypi.python.org/pypi/scikit-learn/0.19.1

PR #17248, opened 2018-03-23
python-sklearn-deap-pip: add package in python.yaml

Minipada:
Add [python-sklearn-deap](https://github.com/rsteca/sklearn-deap) pip package. It contains libraries to use evolutionary algorithms instead of gridsearch in scikit-learn

This allows you to exponentially reduce the time required to find the best parameters for your estimator. Instead of trying out every possible combination of parameters, evolve only the combinations that give the best results.

We use it in our machine learning process to optimize map generation parameters.

Pypi link: https://pypi.python.org/pypi/sklearn-deap

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3198,6 +3198,13 @@ python-sklearn:
     vivid: [python-sklearn]
     wily: [python-sklearn]
     xenial: [python-sklearn]
+python-sklearn-deap-pip:
+  debian:
+    pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Like #17247 please use the apt packages and take a look for other platforms. 

https://packages.ubuntu.com/trusty/python-sklearn

Minipada:
I will do, thanks

PR #17249, opened 2018-03-23
python-fastdtw-pip: add package in python.yaml

Minipada:
Add [python-fastdtw-pip](https://github.com/slaypni/fastdtw) package. It contains python implementation of the FastDTW algorithm:
http://cs.fit.edu/~pkc/papers/tdm04.pdf

It is an approximate Dynamic Time Warping (DTW) algorithm that provides optimal or near-optimal alignments with an O(N) time and memory complexity.

We use it to find similarities between two temporal sequences in different paths.

Pypi link: https://pypi.python.org/pypi/fastdtw

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -963,6 +963,13 @@ python-falcon:
   fedora: [python-falcon]
   gentoo: [dev-python/falcon]
   ubuntu: [python-falcon]
+python-fastdtw-pip:
+  debian:
+    pip:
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This one doesn't appear to have an upstream packaged version so the pip rule is fine.

However please be aware that the pip rule can only be used in a source build and cannot be released on top of.

mikaelarguedas (approved):

PR #17267, opened 2018-03-25
Add qglviewer-qt4 libs for artful and bionic

ahornung:


rosdep/base.yaml
```diff
@@ -2591,6 +2593,8 @@ libqglviewer-qt4-dev:
   fedora: [libQGLViewer-devel]
   gentoo: ['x11-libs/libQGLViewer:0/qt4-2']
   ubuntu:
+    artful: [libqglviewer-dev-qt4]
+    bionic: [libqglviewer-dev-qt4]
```
dirk-thomas: This line has trailing spaces which fails the tests.

PR #17279, opened 2018-03-26
Add werkzeug to base.yaml

tdenewiler:
Web development package used for creating simple web-based operator interfaces.

Package exists for arch, debian, fedora, gentoo, and ubuntu. I tested with `rosdep resolve werkzeug --os=<OS:version>` for all of those distributions. The `nosetests` all pass.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4721,12 +4721,6 @@ vrep:
     trusty:
       source:
         uri: 'https://raw.githubusercontent.com/peci1/vrep_ros_bridge/rdmanifest/vrep.rdmanifest'
-werkzeug:
-  arch: [python-werkzeug]
-  debian: [python-werkzeug]
-  fedora: [python-werkzeug]
-  gentoo: [dev-python/werkzeug]
-  ubuntu: [python-werkzeug]
 wget:
   arch: [wget]
   debian: [wget]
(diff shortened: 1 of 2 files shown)


tdenewiler:
Arch: https://www.archlinux.org/packages/community/any/python-werkzeug/
Debian: https://packages.debian.org/search?keywords=python-werkzeug
Fedora: https://www.tenable.com/plugins/nessus/104941
Gentoo: https://packages.gentoo.org/packages/dev-python/werkzeug
Ubuntu: https://packages.ubuntu.com/search?keywords=python-werkzeug

I am not familiar with, and could not find packages for, the following:
osx, freebsd, opensuse, cygwin, slackware, rhel

mikaelarguedas (changes requested):
Thanks @tdenewiler for providing the links and descroption.

As this is a Python package, it should be added to the python.yaml file and the name of the key should match the debian name aka `python-werkzeug`

tdenewiler:
I was under the wrong impression that `python.yaml` was for pip packages only. The `werkzeug` packages were listed already, except for the `arch` OS. I have added that one only now.

mikaelarguedas (approved):
lgtm, thanks @tdenewiler !

tfoote (approved):

PR #17273, opened 2018-03-26
Added libb64-dev

HEllRZA:
Added a (DAMN fast[sic!]) c++ lib for encoding/decoding base64

arch: https://www.archlinux.org/packages/community/x86_64/libb64/
debian: https://packages.debian.org/sid/libb64-dev
fedora: https://fedora.pkgs.org/25/rpm-sphere/libb64-devel-1.2.1-9.1.i686.rpm.html
gentoo: https://packages.gentoo.org/packages/dev-libs/libb64
ubuntu: https://packages.ubuntu.com/xenial/libb64-dev

PR #17286, opened 2018-03-26
[rosdep][py] Adds svgwrite & svgpathtools

moriarty:
Adds python-svgwrite and svgpathtools for ubuntu.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3332,6 +3332,10 @@ python-support:
     pip:
       packages: []
   ubuntu: [python-support]
+python-svgpathtools:
+  ubuntu: [python-svgpathtools]
+python-svgwrite:
+  ubuntu: [python-svgwrite]
 python-svn:
   debian: [python-svn]
   fedora: [pysvn]


mikaelarguedas:
Please provide a description of the package and your use case as well as rules and links to the packages for all required platforms (Debian and Ubuntu) and if possible for the recommended platforms (Gentoo, Fedora). More information can be found in our [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

PS: keys must be listed in alphabetical order, this is the reason CI is currently failing on this PR

moriarty:
Sorry I overlooked the alphabetical order. 
I followed these guidelines. http://docs.ros.org/independent/api/rosdep/html/contributing_rules.html

1. I have no access to test for other platforms, just ubuntu.
2. I have amended the commit to include the description of the packages (from pypi.python.org )
3. Use case: ``` rosdep install --from-paths src --ignore-src --rosdistro indigo -y```
   some internal packages depend on python-svgwrite and python-svgpathtools

mikaelarguedas:
> I have no access to test for other platforms, just ubuntu.

Providing links from the package databases of the other platforms is enough. As we build binaries for Debian and Ubuntu we require at least rules for these platforms in order to add a new key (otherwise users using these keys will not be able to release their packages). In general the rule is the same as Ubuntu gets its packages from Debian.

Here is an example of previous PR with description and links for the packages: https://github.com/ros/rosdistro/pull/17273.

HTH

moriarty:
I will re-open this when I get a chance

PR #17328, opened 2018-04-03
Additions for ubuntu:bionic and melodic

leolchat:


mikaelarguedas:
@leolchat please provide the entries in alphabetical order to make CI happy.

As you are adding bionic rules please consider providing the artful rules as well as both are required for users to release packages depending on these keys.

leolchat:
I do not have a artful system to check if the packages install well and rosdesktop build, so I would let you/ROS maintenance team check it... on my side, it seems to allow building from source without missing dependencies.

mikaelarguedas:
Please add the rules anyway, as all the rules you're adding here are all the same as in Zesty, it's extremely unlikely that the name is different for artful.

leolchat:
If you insist, please go ahead and add that change, that pull request should be editable by maintainers.

wjwwood:
@leolchat for future reference, they can be looked up (without a running Artful system) here:

https://packages.ubuntu.com/artful/

wjwwood:
I'll add the artful entries eventually, since I need at least `assimp` myself to do the `rviz` release. It might be a few days before I catch up on rviz issues, however. @leolchat if you have time in the meantime we would appreciate the contribution and merge it sooner.

leolchat:
I guess if it breaks on artful, you will catch it early enough.

wjwwood (approved):
Yeah, we'll notice, thanks for adding them. I'll touch them up if needed when we start using them on our released packages.

mikaelarguedas (approved):
lgtm

PR #17355, opened 2018-04-05
rosdep: adding python-parso-pip entry

machinekoder:
Adds a rodep entry for [Parso: A Python Parser](http://parso.readthedocs.io/en/latest/).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2017,6 +2017,17 @@ python-paramiko:
   rhel: [python-paramiko]
   slackware: [paramiko]
   ubuntu: [python-paramiko]
+python-parso:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
This package has debs for Ubuntu [Bionic](https://packages.ubuntu.com/bionic/python-parso) and [Debian Buster](https://packages.debian.org/buster/python-parso).
Can you please drop the `-pip` suffix from the key and update the rules to reflect it so that we can use the deb for these disctributions?
Thanks!

machinekoder:
Okay, let's use pip for the last LTS.

tfoote (approved):

PR #17366, opened 2018-04-06
Added python-poster key to rosdep python.

ghost:
Using poster to handle submitting login information for a web api.

https://packages.ubuntu.com/trusty/python-poster
https://packages.debian.org/stretch/python-poster


tfoote (approved):
+1 I couldn't find a fedora or gentoo equivalent from a quick search

PR #17376, opened 2018-04-07
rc_visard -> rc_visard_ros to match repository name

mikaelarguedas:
@flixr FYI

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -10721,7 +10721,7 @@ repositories:
       url: https://github.com/roboception/rc_genicam_api.git
       version: master
     status: developed
-  rc_visard:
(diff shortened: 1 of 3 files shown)


flixr:
I named it `rc_visard` as that is the name of the meta-package... But if it helps or is the convention to name it as the repo, I don't mind either...

mikaelarguedas:
I noticed it as I have scripts assuming that the keys to the `repositories` dict were repository names. This one was the first one not complying so I reported it to know if it was on purpose. I don't think it's the only entry not matching and I have now updated my scripts to not assume the names are the same so, as you confirmed it's on purpose, I'm going to close this.

Thanks for the feedback :+1:

PR #17434, opened 2018-04-12
Fixed package names for heron_desktop and warthog_simulator.

tonybaltovski:
Sorry guys, I used dashes instead of underscores.  I re-bloomed, thus the version change.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -4515,7 +4515,7 @@ repositories:
       url: https://github.com/heron/heron.git
       version: indigo-devel
     status: developed
-  heron-desktop:
+  heron_desktop:
     doc:
       type: git
       url: https://github.com/heron/heron_desktop.git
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Arg this will need to be reordered alphabetically then

tonybaltovski:
@mikaelarguedas is that something I can do?  They are in alphabetically order currently.

mikaelarguedas:
If you look at the Travis log it should point you to the command you can run locally to rearrange your distro file alphabetically (and print the diff that needs to be applied). I'm responding from a phone so I can't really look at the log myself.

indigo/distribution.yaml
```diff
@@ -16541,7 +16541,7 @@ repositories:
       url: https://github.com/warthog-cpr/warthog.git
       version: indigo-devel
     status: developed
-  warthog-simulator:
+  warthog_simulator:
```
mikaelarguedas: This should be after `warthog_desktop`

tonybaltovski:
Thanks for that @mikaelarguedas :beers:

mikaelarguedas (approved):
:+1:

PR #17452, opened 2018-04-14
adding python-aniso8601

asmodehn:
Note this package had been backported on indigo http://repositories.ros.org/status_page/ros_indigo_default.html?q=aniso (and jade), and from ubuntu xenial is available as a package which I want to use.
Let me know if there is any renaming needed anywhere...

Ref : https://pkgs.org/download/aniso8601

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -266,6 +266,7 @@ python-amqp:
   gentoo: [dev-python/py-amqp]
   ubuntu: [python-amqp]
 python-aniso8601:
+  centos: [python-aniso8601]
   debian: [python-aniso8601]
   fedora: [python2-aniso8601]
   ubuntu: [python-aniso8601]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
The `pkgs.org` link provided has information only for slackware. Does that mean the you want to add this for slackware as well?

Looks like it's also available for gentoo as well: [dev-python/aniso8601](https://packages.gentoo.org/packages/dev-python/aniso8601)  

Regarding the naming, it looks like in the past you've been using the same name for the backported version in ROS and the new versions added in rosdep `python-PKG_NAME` (e.g. python-watchdog).

Is there a motivation for not doing the same for this package?
If not, I suggest renaming the ROS package to match the rosdep key, this should remove the need to modify the dependency declaration in the `package.xml` from one distro to the next.

asmodehn:
Ah good point I had forgotten the python-watchdog case. I will follow the same and rename the indigo pacakgeto match the rosdistro key.

The pkgs.org link I meant to send is actually this one : https://pkgs.org/download/python-aniso8601 to be used as a reference checking if I missed anything...

asmodehn:
https://github.com/ros/rosdistro/pull/17458

mikaelarguedas (approved):
lgtm

PR #17461, opened 2018-04-16
python-flask-restful: 0.3.4-4 in 'indigo/distribution.yaml' [bloom]

asmodehn:
Increasing version of package(s) in repository `python-flask-restful` to `0.3.4-4`:

- upstream repository: https://github.com/flask-restful/flask-restful.git
- release repository: https://github.com/pyros-dev/flask-restful-rosrelease.git
- bloom version: `0.6.4`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -10147,6 +10147,13 @@ repositories:
       url: https://github.com/pyros-dev/aniso8601-rosrelease.git
       version: 0.8.3-5
     status: maintained
+  python-flask-restful:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/pyros-dev/flask-restful-rosrelease.git
+      version: 0.3.4-4
+    status: maintained
   python-ftputil:
     release:
       tags:
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -3469,13 +3469,6 @@ repositories:
       url: https://github.com/asmodehn/flask-cors-rosrelease.git
       version: 3.0.2-2
     status: developed
-  flask_restful:
-    release:
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/asmodehn/flask-restful-rosrelease.git
-      version: 0.3.4-3
-    status: maintained
   flask_reverse_proxy:
     release:
       tags:


mikaelarguedas:
This branch has conflicts, can you rebase it please?

asmodehn:
rebased.

mikaelarguedas (approved):
lgtm

asmodehn:
It seems we now have duplicates ?

```
python-flask-restful (ubuntu, trusty) is multiply defined in
	file:///home/travis/build/ros/rosdistro/indigo/distribution.yaml and 
	file:///home/travis/build/ros/rosdistro/rosdep/python.yaml
```

Just to be clear, this is intended to get implicit transition from python ROS package (indigo) to system package (kinetic).

Should the rosdep key for trusty be dropped ?

tfoote:
We shouldn't be providing the same named thing through two different mechanisms. That will lead to confusion and errors. I'm going to roll this back for now until we understand what you want to do.

asmodehn:
The rosdep key is a pip key (up to xenial), I thought this was not really suitable (can't remember exactly why...) ?
In that case I would think the ros backport is a better choice.

So I can then remove all pip keys from rosdep for this package ?

tfoote:
That's also sort of working, but again might break anyone using the keys already. 

In general though we shouldn't be trying to use the same names unless we're expecting them to be exactly the same. That's why we use `-pip` as a suffix so that when we switch to the debian it's not a collision. And in general we don't want to be backporting debian packages with the same symbols because at some point in the future when a package becomes available upstream it will clobber the ros packaged version on some specific platform.

asmodehn:
> And in general we don't want to be backporting debian packages with the same symbols because at some point in the future when a package becomes available upstream it will clobber the ros packaged version on some specific platform.

I agree with that statement, but what about the python-watchdog example (rospkg indigo #15095 and rosdep key #15578)  ? I thought this was the example to follow when one wants to have dependency on one backported package transition to a system/rosdep package.

Also note, that I can side-step this problem by making distro specific patches in release repos for all packages (which I already started doing for some of my python packages anyway). Extra work, but probably the solution that will give me the most freedom, and allow me to chose dependencies names for each distro. Should that be the advised solution ?

Thanks for letting me know which way would be best.

mikaelarguedas:
My understanding was that we encourage people to release the ython packages they need as ROS packages ather than relying on rosdep keys. In the hope that once they become available upstream we can just stop releasing them as  a ROS package without requiring any change on the consuming packages.
@tfoote what should we recommend users to do in that case in your opinion? create packages with different names (implying changing keys in consumers manifests once it's available)? Or releasing under the same name as the upstream package not yet available?

In the case of flask-restful: the issue happens because we had an "hybrid" set of rules with some pip for the platform that don;t have the package and some apt for the more recent ones. I'm not sure what is the best practice in this case? prevent people from adding pip rules in non `-pip` suffixed keys ?

The second case `python-aniso8601`. Is actually that the rule was wrong and providing a wildcard for a package that was not available on all active distros. So the rule needs to be fixed anyway

I opened #17466 with a suggested fix for both violations. I can split it if we think only the second one makes sense

asmodehn:
Worth reminding here that this problem remains : https://github.com/ros-infrastructure/rosdep/issues/539

So the added python-flask-restful ROS package is [present](http://repositories.ros.org/status_page/ros_indigo_default.html?q=flask), but the rosdep key has been dropped,
And rosdep doesn't find the package on indigo : 

```
to system dependencies:
rostful: No definition of [python-flask-restful] for OS version [trusty]
```
Build : https://travis-ci.org/asmodehn/rostful/builds/367544893

PR #17460, opened 2018-04-16
yp-spur: 1.15.0-0 in 'kinetic/distribution.yaml' [bloom]

at-wat:
Increasing version of package(s) in repository `yp-spur` to `1.15.0-0`:

- upstream repository: https://github.com/openspur/yp-spur.git
- release repository: https://github.com/openspur/yp-spur-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.6.4`
- previous version for package: `null`

PR #17462, opened 2018-04-16
yp-spur: 1.15.0-1 in 'kinetic/distribution.yaml' [bloom]

at-wat:
Increasing version of package(s) in repository `yp-spur` to `1.15.0-1`:

- upstream repository: https://github.com/openspur/yp-spur.git
- release repository: https://github.com/openspur/yp-spur-release.git
- bloom version: `0.6.4`
- previous version for package: `1.15.0-0`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -12483,7 +12483,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/openspur/yp-spur-release.git
-      version: 1.15.0-0
+      version: 1.15.0-1
     source:
       type: git
       url: https://github.com/openspur/yp-spur.git


at-wat:
Package manifest was wrong in #17460. 
I hope this fixes the build on the buildfarm. (I'm not very sure on it since the pre-release test didn't reproduce the error...)

mikaelarguedas:
> Package manifest was wrong in #17460.

If your package is pure cmake. You shouldnt need to depend on catkin, only cmake.
In that case you should use `<buildtool_depend>cmake</buildtool_depend>`.
Some examples can be found at http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty

> I hope this fixes the build on the buildfarm

You will need to create a new release (1.15.1) that integrates that change. [the package.xml used for the current release](https://github.com/openspur/yp-spur/blob/1.15.0/package.xml) (tag 1.15.0) doesnt include the change.

> I'm not very sure on it since the pre-release test didn't reproduce the error...

Yes this is one of the corner cases that can't be caught by the prerelease jobs as they use `catkin_make_isolated` to build the workspace, so CMake is already installed. You would need to [run the release jobs locally](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/jobs/release_jobs.rst#run-the-release-job-locally) to catch this kind of error.

I'm going to mark this as "changes requested" as a new release needs to be made for the package.xml fix to take effect

mikaelarguedas (changes requested):
new release required

at-wat:
@mikaelarguedas Thank you. I'll release 1.15.1 later.

wjwwood:
Actually the tutorial recommends a run depend on catkin, otherwise the setup file isn’t there when only install the third party package, see http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty#Adding_a_Package.xml_to_the_master_Branch

at-wat:
@wjwwood So, is the below correct for package format version 2?
```
  <depend>catkin</depend>
  <buildtool_depend>cmake</buildtool_depend>
```

mikaelarguedas:
@at-wat as you are using format 2 in your package the `run_depend` will be an `exec_depend` in your use case
```
  <buildtool_depend>cmake</buildtool_depend>
  <exec_depend>catkin</exec_depend>
```

PR #17468, opened 2018-04-16
Drop odva_ethernetip from Melodic.

mikepurvis:
Efforts from @clalancette notwithstanding, this isn't building on Bionic right now.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -772,11 +772,6 @@ repositories:
       type: git
       url: https://github.com/ros-drivers/odva_ethernetip.git
       version: indigo-devel
-    release:
-      tags:
-        release: release/melodic/{package}/{version}
-      url: https://github.com/ros-drivers-gbp/odva_ethernetip-release.git
-      version: 0.1.2-0
     source:
       type: git
       url: https://github.com/ros-drivers/odva_ethernetip.git


melodic/distribution.yaml
```diff
@@ -772,11 +772,6 @@ repositories:
       type: git
       url: https://github.com/ros-drivers/odva_ethernetip.git
       version: indigo-devel
-    release:
-      tags:
-        release: release/melodic/{package}/{version}
-      url: https://github.com/ros-drivers-gbp/odva_ethernetip-release.git
-      version: 0.1.2-0
```
dirk-thomas: I would suggest to just drop the `version` line but keep the rest of the release entry.

clalancette:
@mikepurvis FYI; I think the reason that the binary jobs are still failing is that the whole thing is still trying to use version 0.1.2 of the code, which does not have my latest fixes.  It may be enough to do another odva_ethernetip release and re-bloom the whole thing to make it work.

mikepurvis:
@clalancette Oh boo, so you're right. I actually did tag and bloom another release (see https://github.com/ros-drivers-gbp/odva_ethernetip-release), but must have got interrupted before the PR was generated to update this.

mikepurvis:
PR for 0.1.3 in #17470 17470.

PR #17472, opened 2018-04-16
Add pacifica-dbw-ros to documentation index for distro

rborchert-ne:
Please include pacifica-dbw-ros to be indexed and documented on ros.org

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6538,7 +6538,7 @@ repositories:
       test_commits: false
       type: git
       url: https://github.com/NewEagleRaptor/pacifica-dbw-ros.git
-      version: default
+      version: master
     status: developed
   packml:
     doc:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -6529,6 +6529,17 @@ repositories:
       url: https://github.com/allenh1/p2os.git
       version: master
     status: maintained
+  pacifica_dbw_ros:
+    doc:
+      type: git
+      url: https://github.com/NewEagleRaptor/pacifica-dbw-ros.git
+      version: default
```
mikaelarguedas: This needs to be a valid branch name, In this case: `master`

rborchert-ne: Thank you.  I realized that at the same time I saw you post the comment.

mikaelarguedas (approved):
lgtm

PR #17492, opened 2018-04-17
Renaming robot_process to robot_activity and adding to kinetic,lunar,melodic

asiron:
Unfortunately I had to rename the package as it turned out to collide with one of our old internal pkgs

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -11133,6 +11133,17 @@ repositories:
       url: https://gitlab.com/nasa-jsc-robotics/robodyn_utilities.git
       version: develop
     status: maintained
+  robot_activity:
+    doc:
+      type: git
+      url: https://github.com/snt-robotics/robot_activity.git
+      version: master
+    source:
+      test_pull_requests: true
+      type: git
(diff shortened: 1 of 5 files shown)


mikaelarguedas:
Jade as been EOL for almost a year now and we don't run any infrastructure for it anymore, so we don't take any modification of the distribution file for that distro

indigo/distribution.yaml
```diff
@@ -11133,6 +11133,17 @@ repositories:
       url: https://gitlab.com/nasa-jsc-robotics/robodyn_utilities.git
       version: develop
     status: maintained
+  robot_activity:
+    doc:
+      type: git
+      url: https://github.com/snt-robotics/robot_activity.git
+      version: master
+    source:
+      test_pull_requests: true
```
mikaelarguedas: I guess the webhook has been created on the new repository as well ?

asiron: I changed the name of the repo and the webhook is still there, so it should be fine, otherwise I can generate a new one

mikaelarguedas (approved):

PR #17494, opened 2018-04-17
Update distribution.yaml

scarrionv:


tfoote:
Due to the large number of branches and tags that git-buildpackage requires (and bloom creates) it's highly recommended to use a separate release repository from the upstream code repository. http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease#Creating_a_Release_Repository Also I'd highly recommend adding at least a source and doc entry for your package. If you let bloom create the pull request for you it will help you do that.

PR #15259, opened 2017-06-16
rename xdot in python.yaml to python-xdot, becase xdot is already tak…

k-okada:
…en by ros package, atleast until indigo http://wiki.ros.org/xdot

rewrited version of #15257

some of rosdep keys are multiply defined, and that sometime confuse users.

rename xdot in python.yaml to python-xdot, becase xdot is already taken by ros package, atleast until indigo http://wiki.ros.org/xdot
change xdot to python-xdot, this will break backword compatibility....

PR #17553, opened 2018-04-20
Add bson

awesomebytes:
`bson` is a dependency for `rosbridge_library`, needed for Gentoo (as seen in https://github.com/ros/ros-overlay/issues/560).

~~`xdot` is a run depend of [ecto](https://github.com/plasmodic/ecto/blob/master/package.xml#L17).~~

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -496,6 +496,7 @@ python-bs4:
 python-bson:
   debian: [python-bson]
   fedora: [python-bson]
+  gentoo: [dev-python/bson]
   osx:
     pip:
       packages: [bson]


tfoote (changes requested):
There's already a key `python-xdot` https://github.com/ros/rosdistro/blob/c075a3427f3797beb517b27da18f5b68e96b6731/rosdep/python.yaml#L3922 that should be extended/reused instead of creating a new one.

mikaelarguedas:
After a bit of archaeology:
This key has been renamed from `xdot` to `python-xdot` in https://github.com/ros/rosdistro/pull/15259. This is mostly because rosdep doesnt support/allow collisions between released ros package names and rosdep keys (https://github.com/ros-infrastructure/rosdep/issues/539). But we didn't want to keep releasing xdot as a ROS package now that it's available upstream.

At the time of the renaming, issues have been opened on the repos using it to notify of the change:
https://github.com/srv/srv_tools/issues/16
https://github.com/plasmodic/ecto/issues/297
But the packages don;t seem to have been updated since.

If you want to compile these packages from source you will need to rename the key in their package.xml. If you can submit PRs on these repos with the change that would be valuable as well :smile: 

The bson change can stay in this PR though :+1: but the xdot rule should be removed.

Thanks!

awesomebytes:
@mikaelarguedas Done, removed the commit about xdot. Sorry for the noise! I didn't find that information at the moment. Interesting. I'll need to patch that for the Gentoo packages building.

mikaelarguedas (approved):
lgtm, thanks @awesomebytes

awesomebytes:
@tfoote All good now?

tfoote (approved):

tfoote:
@awesomebytes Please actually submit the patch upstream so that the code uses the correct rosdep keys. It will be needed in the next release anyway and it will be better to have it upstream instead of just in gentoo.

awesomebytes:
@tfoote Done:
* srv_tools: https://github.com/srv/srv_tools/pull/21
* ecto: https://github.com/plasmodic/ecto/pull/299

PR #17573, opened 2018-04-22
add python-pydbus to rosdep

cratliff:
python-pydbus

https://pypi.org/project/pydbus/

Pythonic DBus library.

This package is used to get information from and interact with systemd in your python module.

This package is available as an apt package in artful, but not prior.  So I didn't use the -pip extension, but am still installing it with pip.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2321,6 +2321,18 @@ python-pyaudio:
   fedora: [pyaudio]
   gentoo: [dev-python/pyaudio]
   ubuntu: [python-pyaudio]
+python-pydbus:
+  ubuntu:
+    artful: [python-pydbus]
+    bionic: [python-pydbus]
+    trusty:
+      pip:
+        depends: [python-gi]
+        packages: [pydbus]
+    xenial:
+      pip:
+        depends: [python-gi]
+        packages: [pydbus]
 python-pydot:
   arch: [python2-pydot]
   debian: [python-pydot]


mikaelarguedas:
> This package is available as an apt package in artful, but not prior. 

It looks like python-dbus is availabe on Ubuntu for a while (currently available on [all active distros](https://packages.ubuntu.com/search?keywords=python-dbus&searchon=names&exact=1&suite=all&section=all)).
Are you sure python-dbus is the same package as the pydbus project ?

> So I didn't use the -pip extension, but am still installing it with pip.

If the package exist on recent distributions, please provide the apt rule for the distributions it;s available on. If there are only pip rules for all platforms, please use the `-pip` suffix.

mikaelarguedas:
Note: Apparently the debian name of the pydbus library is [python-pydbus](https://packages.ubuntu.com/search?keywords=python-pydbus&searchon=names&exact=1&suite=all&section=all) (artful and bionic).

:+1: to @tfoote comment, if python-dbus satisfies your need I'd recommend using that instead

cratliff:
I changed the listing to pyhton-pydbus and added the apt listings for artful and bionic.

python-dbus doesn't work for what we are doing.  Also, it's development status is listed as inactive on pypi and the last update to the code is over 2 years ago.

cratliff:
@mikaelarguedas Thanks, pip is only used for Xenial and Trusty now.

mikaelarguedas (approved):
lgtm, thanks @cratliff

PR #17590, opened 2018-04-24
omronsentech_camera update distribution.yaml

ose-support-ros:
Our package depends on a third-party library which is distributed separately. To avoid build-failure in Jenkins, at this moment we exclude the source code indexing.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6311,10 +6311,6 @@ repositories:
       type: git
       url: https://github.com/ose-support-ros/omronsentech_camera.git
       version: master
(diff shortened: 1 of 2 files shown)


tfoote (changes requested):
Instead of removing it please just add a tag `test_commits: false` that way it can be checked out by people, but won't be attempted to be built on the farm.

http://www.ros.org/reps/rep-0143.html

tfoote (approved):

tfoote:
Great, the devel job will be removed in the next reconfigure cycle tomorrow evening.

PR #17617, opened 2018-04-25
add scratch4robots:0.0.1-0 in 'kinetic/distribution.yaml'

scarrionv:


mikaelarguedas (approved):
sync happened https://discourse.ros.org/t/new-package-for-kinetic-2018-04-25/4620

scarrionv:
Hi, should my package be already available? I do not see it in the packages added in last synchronization. I have also received many error messages in Jenkins Jobs. Something has gone wrong? 
Thanks

mikaelarguedas:
> should my package be already available?

As this PR was merged *after* the sync, the packages were not part of the sync. If everything builds fine it will be part of the next sync.

> I have also received many error messages in Jenkins Jobs. Something has gone wrong?

This means that your package cannot be built and installed and needs to me modified.
A good way to catch most of these problem is to build your package locally with `catkin_make_isolated --install`

HTH

tfoote:
In the emails you were sent you'll be able to get the console output. The install is failing to find `src/scratch4ros` which I suspect is from the default install rules for these settings: 

https://github.com/JdeRobot/Scratch4Robots/blob/f88f9dac06a3ca8ee3f049addb9e0c0ae960e252/setup.py#L11-L12

You don't appear to have any manually defined installation rules which means that the resultant debian packages will likey have very little inside of them.

PR #17620, opened 2018-04-25
Remove ridgeback_simulator from kinetic builds due most packages failing to build

tfoote:
Ticketed at: https://github.com/ridgeback/ridgeback_simulator/issues/3

mechanum_gazebo_plugin does appear to build wihtout error but doesn't actually build the plugin so removing it too.


tfoote:
@tonybaltovski FYI this is rolling back #17413

PR #17622, opened 2018-04-25
removing camera1394stereo from kinetic due to it not building

tfoote:
Partial rollback of #17001 @miquelmassot

Ticketed upstream at: https://github.com/srv/camera1394stereo/issues/8


miquelmassot:
Related to https://github.com/ros/common_msgs/issues/97

jamessergeant:
@tfoote The [ROS wiki](http://wiki.ros.org/camera1394stereo) still lists kinetic supported for camera1394stereo. Should be updated.

tfoote:
Yeah, we don't automatically delete content. I've manually cleared the content.

PR #17633, opened 2018-04-26
[dep] Add python-sh.

130s:
- Debian https://packages.debian.org/sid/python/python-sh
- Fedora https://admin.fedoraproject.org/pkgdb/package/rpms/python-sh/
- Gentoo https://packages.gentoo.org/packages/dev-python/sh
- Ubuntu https://packages.ubuntu.com/search?keywords=python-sh
  - It's tricky to me that the apt pkg seems to have become available since Xenial and on Trusty it's only available via `pip`. Assuming it'll remain available onward, it there a way to describe the rule cleanly so that we won't have to populate for each distro?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3119,6 +3119,15 @@ python-setuptools:
     xenial_python3: [python3-setuptools]
     yakkety: [python-setuptools]
     zesty: [python-setuptools]
+python-sh:
+  debian: [python-sh]
+  fedora: [python-sh]
+  gentoo: [dev-python/sh]
+  ubuntu:
+    '*': [python-sh]
+    trusty:
+      pip:
+        packages: [sh]
 python-shapely:
   debian: [python-shapely]
   fedora: [python-shapely]


mikaelarguedas:
REP111 has been [updated to support os wildcards](https://github.com/ros-infrastructure/rep/pull/153) to cover this type of scenario. And O think it has been implemented (https://github.com/ros-infrastructure/rosdep/pull/573). So we could try using that here.

rosdep/python.yaml
```diff
@@ -3119,6 +3119,17 @@ python-setuptools:
     xenial_python3: [python3-setuptools]
     yakkety: [python-setuptools]
     zesty: [python-setuptools]
+python-sh:
+  debian: [python-sh]
+  fedora: [python-sh]
+  gentoo: [sh]
```
awesomebytes: For Gentoo it should be `dev-python/sh` please.

130s:
@mikaelarguedas @awesomebytes Thanks, updated.

mikaelarguedas:
thanks @130s for the update, looks like ci is failing due to trailing whitespaces and alphabetical order (the wildcard should be before any alphabetical character)

130s:
Wow, I couldn't tell from the error output. Thanks @mikaelarguedas, now CI passed.

mikaelarguedas:
:+1: 

> I couldn't tell from the error output. 

At the end of the [log](https://travis-ci.org/ros/rosdistro/builds/371659588) you can see `AssertionError: see folded block 'block2' for details`
Then you can go up and click on the mentionned block and it prints:
```
Checking rosdep file: rosdep/python.yaml
checking for trailing spaces...
  ERR: trailing space line 3125
checking for incorrect indentation...
checking for non-bracket package lists...
checking for item order...
  ERR: list out of alphabetical order line 3130.  '*' should come before 'trusty'
building yaml dict...
there were errors, please correct the file
```

130s:
> `AssertionError: see folded block 'block2' for details`

I see. I didn't know block2 here is the folded block in the CI output (but I can't think of a good way to improve this either).

tfoote (approved):

rosdep/python.yaml
```diff
@@ -3119,6 +3119,15 @@ python-setuptools:
     xenial_python3: [python3-setuptools]
     yakkety: [python-setuptools]
     zesty: [python-setuptools]
+python-sh:
+  debian: [python-sh]
+  fedora: [python-sh]
+  gentoo: [dev-python/sh]
+  ubuntu:
+    '*': [python-sh]
```
dirk-thomas: This wildcard will also match Ubuntu distros earlier than Trusty. And for those the resulting mapping will be invalid. Therefore imo this should enumerate the older Ubuntu distros explicitly with a mapping to `None`.

mikaelarguedas: That's what we've been doing for new keys for a while: Keep rule simple and not adding explicitly null entries for every distro that was already EOL when the key was added.
Did that change ?

dirk-thomas: When not using a wildcard the missing entry just means rosdep can't resolve the dependency on that platform. That was just fine. But when using `'*'` the impact is very different because not existing mappings will fall back to the wildcard which is wrong in that case. Therefore when using the wildcard "unsupported" distros should be explicitly listed as "not supported" to avoid that wrong mapping.

mikaelarguedas: Isnt that the exact same thing as adding `ubuntu: [foo]` ? it will be resolved for all ubuntu platforms even the EOL ones. And that;s what we've been doing for a while as long as it was defined for all non EOL distros.

I dont think expecting people to list explicitly every EOL distros when adding a new key is desired.

tfoote: It has been our policy that if a key is added it's ok to use the generic matches all if all non-EOL distros support that specific key. This will cause EOL distros to incorrectly resolve the keys on the older distros. However that will not break any existing users, because the key never existed. So by definition it's not being used on the EOL distros, thus we will not break anyone on the EOL distros. (except for a name collision with a private package or rosdep keys but that's on them and easily resolvable.)

Our goal for keeping the EOL'd keys around is to not break people using them. This continues to not break existing users. It's just something that if you're actively developing or backporting to an EOL platform you may run into resolutions that are not expected, but you're going outside of our supported envelope and some work may be required.

PR #17670, opened 2018-04-28
Melodic release: rqt_runtime_monitor, rqt_nav_view

ablasdel:
Got a "KeyError - u'gmail" error when trying to auto open PRs via bloom (maybe my 2 factor?)

Either way here is a copy of the patch text from both rqt_runtime_monitor & rqt_nav_view

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -1605,6 +1605,21 @@ repositories:
       url: https://github.com/ros-visualization/rqt_msg.git
       version: master
     status: maintained
+  rqt_nav_view:
+    doc:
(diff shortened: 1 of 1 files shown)


tfoote:
@ablasdel If you're using 2FA you will want to use an oauth token. http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization

ablasdel:
I did that bit to get as far as I did. The error was only in the final PR stage so I figured it wasn't worth worrying about. 

Possible it is related to my setup. If it happens next release I'll look deeper.

PR #17699, opened 2018-05-02
Add a rosdep key for wxwidgets for artful and bionic.

clalancette:


mikaelarguedas (approved):

tfoote:
Why did we put this instead of #17689 ?

clalancette:
My fault; I didn't realize #17689 was open when I opened this PR.  I don't care which way we go, we just need these keys for melodic.

mikaelarguedas:
My bad, I didnt see this was the same key as the other PR when reviewing and the change was valid so I went ahead.
Do you want me to open a new PR with the commits from #17689 ?

PR #17750, opened 2018-05-04
Add rosdep keys for gazebo9

j-rivero:


rosdep/base.yaml
```diff
@@ -792,6 +794,13 @@ gazebo7:
   gentoo: [sci-electronics/gazebo]
   slackware: [gazebo]
   ubuntu: [gazebo7]
+gazebo9:
```
tfoote: These rules are a little over broad. Gazebo9 isn't available on all versions of these distros. Can we make this a little bit more specific so that we don't have rosdep trying to install gazebo9 on unsupported platforms?

tfoote: Also where is gazebo9 available from? My machine only shows as available from packages.osrfoundation.org (xenail) and rmadison only shows it for bionic and cosmic. If it's not in the default sources + ROS we shouldn't have it here.

rosdep/base.yaml
```diff
@@ -781,6 +781,8 @@ gazebo:
     xenial: [gazebo7]
     yakkety: [gazebo7]
     zesty: [gazebo7]
+    artful: [gazebo9]
+    bionic: [gazebo9]
```
mikaelarguedas: CI is failing because these are not in alphabetical order

tfoote (commented):
Definitely updating the "gazebo" rosdep key makes sense. However I'm a little concerned about adding the gazebo9 key. See my inline comments about availability, but it's also generally been our goal to not have conflicting dependencies. If one package calls out gazebo7 and one package calls out gazebo9 they won't be able to be installed on the same system, and we'll end up with packages in the same rosdistro that are no longer compatible. Part of the role of our distro management is to specifically pick a single version that we know is compatible and enforce that. 

We can merge this as is, but it's basically leaving a lot of corner cases for our users to foul up, and it's not just localized to them, if we get separate sets of packages with the conflicting dependencies, it can bifrucate the rosdistro(s).

rosdep/base.yaml
j-rivero: Thanks Mika. Fixed.

rosdep/base.yaml
j-rivero: Ouch, Gazebo9 is not available in all the distributions listed here, you are right. It is available for Xenial from packages.osrfoundation.org, and Ubuntu servers for Bionic. In Debian, it is available in Unstable and Testing. In Gentoo from the official repo. In Arch from the User repositories.

j-rivero:
> If one package calls out gazebo7 and one package calls out gazebo9 they won't be able to be installed on the same system, and we'll end up with packages in the same rosdistro that are no longer compatible. Part of the role of our distro management is to specifically pick a single version that we know is compatible and enforce that.
> 
> We can merge this as is, but it's basically leaving a lot of corner cases for our users to foul up, and it's not just localized to them, if we get separate sets of packages with the conflicting dependencies, it can bifrucate the rosdistro(s).

Sound good to me this vision. I see two options: we remove the definitions to `gazebo9`and `libgazebo9` and add a generic `libgazebo-dev` key in the same way that we have `gazebo`. Or we keep the xxx9 keys and I refine/fix the support platforms (I found this a bit redundant and see no real gain to be honest).

mikaelarguedas:
> Sound good to me this vision. I see two options: we remove the definitions to gazebo9and libgazebo9 and add a generic libgazebo-dev key in the same way that we have gazebo.

How would that work for Debian Stretch ? (for Lunar it should resolve to gazebo7 and for melodic it should resolve to gazebo9). My guess was that introducing the gazebo9 key would allow gazeb_ros_pkgs to be released in melodic and using gazebo9 as a key to be sure to pick that version on Debian Stretch.

Gazebo 9 is available in the ros repositories for Ubuntu Artful and Debian Stretch (http://packages.ros.org/ros/ubuntu/pool/main/g/gazebo9/). I agree this doesnt resolve the situaton for the other platforms though.

j-rivero:
> How would that work for Debian Stretch ? (for Lunar it should resolve to gazebo7 and for melodic it should resolve to gazebo9).

Ummm, that is probably the reason why it was created in the first place and the reason why I found it redundant while it is not. Good point.

rosdep/base.yaml
```diff
@@ -792,6 +794,16 @@ gazebo7:
   gentoo: [sci-electronics/gazebo]
   slackware: [gazebo]
   ubuntu: [gazebo7]
+gazebo9:
+  debian:
+    buster: [gazebo9]
```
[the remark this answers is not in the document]
j-rivero: ok.

rosdep/base.yaml
```diff
@@ -792,6 +794,16 @@ gazebo7:
...
+    sid: [gazebo9]
```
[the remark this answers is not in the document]
j-rivero: done.

rosdep/base.yaml
```diff
@@ -792,6 +794,16 @@ gazebo7:
...
+  gentoo:
+    portage:
```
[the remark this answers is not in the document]
j-rivero: removed.

mikaelarguedas (approved):
Thanks Jose for iterating on this! this looks good to me now :+1: 
I'll leave it up to @tfoote to give the final review / merge

j-rivero:
> Thanks Jose for iterating on this! this looks good to me now +1

Thanks for tips Mika. I'll wait for Tully no problem.

PR #17751, opened 2018-05-04
[dep] build-essential.

130s:
- debian, opensuse, Ubuntu can be found via https://pkgs.org/download/build-essential

For some Linux distros that look commonly registered, such as gentoo, fedora, this pkg isn't found. Also confirmed individually:
- gentoo https://packages.gentoo.org/packages/search?q=build-essential
- fedora https://apps.fedoraproject.org/packages/s/build-essential

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -307,6 +307,10 @@ box2d-dev:
   debian: [libbox2d-dev]
   fedora: [Box2D-devel]
   ubuntu: [libbox2d-dev]
+build-essential:
+  debian: [build-essential]
+  opensuse: [build-essentials-packages]
+  ubuntu: [build-essential]
 bullet:
   arch: [bullet]
   debian: [libbullet-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -309,6 +309,8 @@ box2d-dev:
   ubuntu: [libbox2d-dev]
 build-essential:
   debian: [build-essential]
+  fedora: [development-tools]
+  gentoo: []
   opensuse: [build-essentials-packages]
   ubuntu: [build-essential]
 bullet:


tfoote:
Fedora looks like `development-tools` : https://unix.stackexchange.com/questions/1338/what-is-the-fedora-equivalent-of-the-debian-build-essential-package

Maybe @allenh1 or @awesomebytes could suggest a gentoo equivalent?

allenh1:
So, in the case of gentoo, that will be there for sure. we could, I suppose, introduce a virtual package that would pull in those deps (if it is needed for some package), but the build essential is most equivalent to the stage3 you start with.

allenh1:
@130s what is the reason for adding this department rule?

130s:
@allenh1 I have a corporate internal pkg that builds Debian and it depends on this pkg.

tfoote:
If those tools are expected to already be installed on gentoo, we can just give it an empty list to say that nothing needs to be installed `gentoo: []`

awesomebytes:
It just came to my mind that on a Gentoo install based on binary packages it isn't necessarily true that build-essentials stuff will be there (in fact, I just came up with not having CMake in such a system).

Maybe gentoo could depend on:

```
gentoo: [sys-devel/gcc, sys-libs/glibc, sys-devel/make, sys-devel/binutils, app-arch/bzip2, app-arch/xz-utils, dev-lang/perl, dev-libs/gmp, dev-libs/mpfr, dev-libs/mpc, sys-devel/gcc-config, sys-libs/ncurses, sys-libs/readline, app-shells/bash, app-arch/xz-utils, sys-apps/sed, sys-apps/baselayout-prefix, dev-libs/libffi, sys-devel/m4, sys-devel/flex, sys-apps/diffutils, sys-devel/bison, sys-devel/patch, sys-devel/binutils-config, sys-apps/coreutils, sys-apps/findutils, app-arch/gzip, app-arch/tar, sys-apps/grep, sys-apps/gawk, sys-apps/file]
```

I came up with this list by checking what does build_essentials in Ubuntu provide + what is bootstrapped on Gentoo on the first stage of bootstrapping (https://gitweb.gentoo.org/repo/proj/prefix.git/plain/scripts/bootstrap-prefix.sh). Maybe there is something missing, or something extra.

There may be a better way, some meta-package that already brings all the necessary, but I don't know it and haven't found it.

130s:
Thanks for the feedback. I'm not sure if I understand all feedback, so I appreciate if each of you could review @allenh1 @awesomebytes @tfoote

allenh1:
That looks good as far as `gentoo` tag is concerned (in my opinion), but I defer to @tfoote's opinion.

nuclearsandwich:
Archlinux has the `base-devel` package group which is roughly equivalent to build-essential, although it includes the packaging utilities for Arch rather than dpkg-dev.

However, I'm not sure that we necessarily want to try and incorporate non-debian-derived distros into this key by assuming that a package which depends on build-essential would function adequately on another distro with its packaging tools.

The Debian Policy manual also [explicitly recommends against](https://www.debian.org/doc/debian-policy/#package-relationships) adding build dependencies on build-essential unless restricting the dependency to a specific version. Depending on build-essential without a version restriction is [a lintian error](https://lintian.debian.org/tags/build-depends-on-build-essential-package-without-using-version.html).

allenh1:
I'm fine with leaving out the non-deb people, since I'm not aware of any packages getting blocked by this one.

We'll make a key & a virtual package for gentoo if it becomes an issue.

130s:
> Archlinux has the base-devel package group which is roughly equivalent to build-essential, although it includes the packaging utilities for Arch rather than dpkg-dev.
> 
> However, I'm not sure that we necessarily want to try and incorporate non-debian-derived distros into this key by assuming that a package which depends on build-essential would function adequately on another distro with its packaging tools.

I'm not sure either. As a user, I would expect exactly the same software (despite version difference between distro) if the rosdep key is the same. But I'm not sure, as @tfoote suggested a package for Fedora that is not exactly the same software https://github.com/ros/rosdistro/pull/17751#issuecomment-386731757, that might be ok in ROS land?

@nuclearsandwich

> The Debian Policy manual also explicitly recommends against adding build dependencies on build-essential unless restricting the dependency to a specific version. Depending on build-essential without a version restriction is a lintian error.

Thanks for the pointer. Are you suggesting to not add a rosdep key for `build-essential`, or is it just a warning when one uses this key? Personally I still like to add this key as usage should be up to users (developers).

nuclearsandwich:
> Are you suggesting to not add a rosdep key for build-essential

Thank you for asking and giving me the opportunity to voice my concerns. I think it has a high potential to cause confusion and cross-platform headaches if incorporated in the general rosdep database. I would like to share two alternatives which I think keep this rosdep database optimized for approachable cross-platform use:
* Identifying what members of build-essential are required by the internal package and depending on them directly. Contributing those rosdep keys upstream.
* Including build-essential in an additional list of rosdep keys used internally by your organization.

I recognize that both require extra effort for you and your team but I think that it can save effort across the community preventing dependence on a virtual package with a lot of cross-platform nuance where I cannot see a generically suitable resolution.

> Personally I still like to add this key as usage should be up to users (developers).

I hear you. Flexibility is paramount to successful community and it is not my intention to exercise anything like a power of veto, even if I had such a thing. I also don't want to discourage contributing rosdep definitions upstream. However, I think this case in particular, and platform-specific virtual packages / package groups in general, are worth the additional deliberation we've had to consider whether their inclusion has a net benefit. As I understand it, rosdep's architecture enables you to augment and modify the rosdep database by including additional sources on your machines so we are not preventing your usage of the key but rather preventing its use by default.

tfoote:
Since build-essential is a debianism that both doesn't have equivalents on other platforms and even by debian standards is generally not recommended as a dependency due to it's virtual nature. I think it does make sense to not put this in as a rosdep rule and recommend that the more specific dependencies that are actually needed. (It has some significantly different dependencies, from multiple compilers to runtime libraries and build systems. for which several of them already have rules available) There is a level of flexibility we want to leave to the users to choose. But if we cannot keep a consistent resolution across all platforms that will devalue the rosdep database as it won't reliably be able to resolve across platforms. We also have to limit choices that cause conflicts for better usability.

130s:
I see. I just did not think of such a complication this PR would inject. Case closed.

> I would like to share two alternatives which I think keep this rosdep database optimized for approachable cross-platform use:

Noted. Thanks for the suggestion.

PR #17761, opened 2018-05-05
Drop test_pull_requests: true from perception_pcl

wkentaro:
## Why?

- It is already tested on Travis.
- It fails because of compiler warnings in PCL code, and they cannot
be solved in the ROS package.

It is described in https://github.com/ros-perception/perception_pcl/pull/197#issuecomment-386824502.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -1113,7 +1113,6 @@ repositories:
       url: https://github.com/ros-gbp/perception_pcl-release.git
       version: 1.6.0-0
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros-perception/perception_pcl.git
       version: melodic-devel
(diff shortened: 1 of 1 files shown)


wkentaro:
@paulbovbel Do you agree?

paulbovbel:
Sounds good to me, although our 'devel' jobs will continue faililng. It may be worthwhile to ask the maintainers of PCL to backport to 1.8.x

tfoote:
@wkentaro  @paulbovbel  You can also set "test_commits: false" to disable the devel jobs

http://www.ros.org/reps/rep-0143.html

wkentaro:
It seems that the default value of test_commits is `false`, and it does not make any affects.
Is that right?

dirk-thomas:
While the `ros_buildfarm` package has some default settings the configuration files of the official buildfarm have their own default values: https://github.com/ros-infrastructure/ros_buildfarm_config/commit/64b1fa17f64095b948c5d80bb2607f71bc25d906#diff-4579fee430150a132a64f47e9ea55f97R53

So for build.ros.org commits are tested by default while pull request testing had to be enabled explicitly.

wkentaro:
@dirk-thomas Thanks for the comment. I updated the commits.

mikaelarguedas (approved):
lgtm

PR #17770, opened 2018-05-07
Add python-gitlab package to python.yaml

maralbus:
Add **Python Gitlab** ([github](https://github.com/python-gitlab/python-gitlab), [read-the-docs](http://python-gitlab.readthedocs.io/en/stable/index.html)) to `python.yaml`

> `python-gitlab` is a Python package providing access to the GitLab server API.
> It supports the v3 and v4 APIs of GitLab, and provides a CLI tool (gitlab).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4083,7 +4083,7 @@ python3-gitlab:
     '*': [python3-gitlab]
     jessie:
       pip:
-        packages: [python3-gitlab]
+        packages: [python-gitlab]
   ubuntu:
     '*': [python3-gitlab]
     trusty:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1227,8 +1227,22 @@ python-github-pip:
     pip:
       packages: [PyGithub]
 python-gitlab:
-  debian: [python-gitlab]
-  ubuntu: [python-gitlab]
+  debian:
+    buster: [python-gitlab]
+    jessie:
+      pip:
+        packages: [python-gitlab]
+    stretch: [python-gitlab]
+  ubuntu:
+    artful: [python-gitlab]
+    bionic: [python-gitlab]
+    cosmic: [python-gitlab]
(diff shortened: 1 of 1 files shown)


mikaelarguedas (dismissed):
`python-gitlab` is [available in Ubuntu](https://packages.ubuntu.com/search?keywords=python-gitlab&searchon=names&exact=1&suite=all&section=all) since Xenial. Please use the system version rather than the pip version.

New keys need to cover at least Ubuntu and Debian so that they can be used to release packages.
More information can be found in the [Contribution Guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions)

ipa-flg-ma:
Thank you.
I updated my changes... Is it now OK?

mikaelarguedas:
@ipa-flg-ma I made a few follow-up changes:
As `python(3)-gitlab` is not available on all active distributions, we need to call out the distributions explicitly:
  - For Ubuntu: available on all active distributions but Trusty, so I added a pip entry for `trusty` and the APT entries for all the other distributions
  - For Debian: available on all active distribution except Jessie, so I added a pip entry for `jessie` and the APT entries for all the other ones

mikaelarguedas (approved):
lgtm

tfoote:
@mikaelarguedas Would it not make sense to use the new `'*'` syntax for this to make it less verbose and future proof? We can keep the pip rules explicitly for trusty and jessie but every other supported distro and expected future distro will match the wildcard rule.

mikaelarguedas:
Yeah that's fine by me :+1: .
That was my original approach but I reconsidered as it seemed we were not all on the same page on which context we should use it last time

tfoote:
Ok, we're going to need to start trying it out and learning if it works. I've updated to use the wildcard syntax.

mikaelarguedas (approved):
lgtm :+1:

PR #17775, opened 2018-05-08
mrpt1: 1.5.7-0 in 'kinetic/distribution.yaml' [bloom]

jlblancoc:
Increasing version of package(s) in repository `mrpt1` to `1.5.7-0`:

- upstream repository: https://github.com/MRPT/mrpt.git
- release repository: https://github.com/mrpt-ros-pkg-release/mrpt1-release.git
- bloom version: `0.6.4`
- previous version for package: `1.5.6-0`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5523,7 +5523,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/mrpt-ros-pkg-release/mrpt1-release.git
-      version: 1.5.6-0
+      version: 1.5.7-0
     source:
       type: git
       url: https://github.com/mrpt/mrpt.git


jlblancoc:
Thanks for merging!

BTW: This newer version builds OK from sources in the ROS build farm but still fails to generate .deb packages, as described in [this ROS answer question](https://answers.ros.org/question/290505/jenkins-error-dpkg-shlibdeps-cant-find-libs-building-non-catkin-pkg/). 
Sorry for irrupting into this PR with this call for help, but I definitively run out of ideas and hopefully someone has found a similar issue in the past. It seems to me that @vrabaud addressed a similar issue with OpenCV years ago. Any help would be greatly appreciated!

tfoote:
Looking at the logs. I noticed that it's erroring looking for the 1.5 versions of the libraries. But it's only processing the 1.5.7 libraries which I would have expected to have the same content, thus be indistinguishable.

jlblancoc:
Thanks for taking a look!
There are even symlinks from lib-x.y -> lib-x.y.z created as part of the standard Debian packing procedure, so that shouldn't be a problem. 

I found a [great guide](https://gist.github.com/awesomebytes/196eab972a94dd8fcdd69adfe3bd1152) by @awesomebytes on locally building the deb package using bloom, let's see if that allow me to do further findings...

awesomebytes:
Wohoo it was useful for someone else, great! @jlblancoc thanks :)

PR #17807, opened 2018-05-11
Added ros_ips to package index of kinetic distribution.yaml

christianarndt0:
I would like my ROS package (ros_ips) to be indexed to run prerelease tests and eventually release it as a binary.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9207,6 +9207,15 @@ repositories:
       url: https://github.com/jstnhuang/ros_explorer.git
       version: kinetic-devel
     status: developed
+  ros_ips:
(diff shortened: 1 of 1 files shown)


tfoote:
@chrarndt Please consider renaming this package to follow our naming conventions. http://wiki.ros.org/ROS/Patterns/Conventions

From looking at the package, I'd recommend `metratec_ips` as the package name since that's an concise non-ambiguous description of the content. ROS in the name is redundant, and IPS is not obviously interpretable and too short to be searchable.

tfoote:
That looks about right. It looks like you haven't actually run bloom yet, so it will basically just be 1 through 3.

PR #17816, opened 2018-05-11
Adding dmu_ros to documentation index for distro

lgruda:
I would like to add dmu_ros to be indexed and documented on ros.org


dirk-thomas:
If PR #17810 is obsolete please go ahead and close it. Thanks.

dirk-thomas:
Nmv, you already did. Thank you.

PR #17835, opened 2018-05-13
Adding odometry_serialcom to documentation index for distro

wlfws:
I'd like odometry_serialcom to be indexed and documented on ros.org

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -8583,6 +8583,15 @@ repositories:
       url: https://github.com/xaxxontech/oculusprime_ros.git
       version: master
     status: developed
+  odometry_serialcom:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


indigo/distribution.yaml
```diff
@@ -8583,6 +8583,15 @@ repositories:
       url: https://github.com/xaxxontech/oculusprime_ros.git
       version: master
     status: developed
+  odometry_serialcom:
+    doc:
+      type: git
+      url: https://github.com/wlfws/odometry_serialcom.git
+      version: indigo-devel
```
mikaelarguedas: the `version` here should be a valid branch name. Looking at your repository there is a single branch named `master`. you likely want to use `master` instead of `indigo-devel` here (Same comment below for the source entry)

mikaelarguedas:
@wlfws friendly ping

PR #17841, opened 2018-05-13
Add dependencies for Melodic that rospilot needs

cberner:
Description: PostGIS adds support for geographic objects to the PostgreSQL object-relational
database.

Links:
* https://admin.fedoraproject.org/pkgdb/package/rpms/postgis/
* https://packages.debian.org/stretch/postgresql-9.6-postgis-2.3
* https://packages.ubuntu.com/bionic/postgresql-10-postgis-2.4

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4079,10 +4079,10 @@ postgresql:
   ubuntu: [postgresql, postgresql-contrib]
 postgresql-9.x-postgis:
   debian:
-    buster: [postgresql-9.6-postgis-2.3]
-    jessie: [postgresql-9.4-postgis-2.1]
-    stretch: [postgresql-9.6-postgis-2.3]
-    wheezy: [postgresql-9.4-postgis-2.1]
+    buster: [postgresql-9.6-postgis-2.3, postgresql-contrib, postgresql-server-dev-all]
+    jessie: [postgresql-9.4-postgis-2.1, postgresql-contrib, postgresql-server-dev-all]
+    stretch: [postgresql-9.6-postgis-2.3, postgresql-contrib, postgresql-server-dev-all]
+    wheezy: [postgresql-9.4-postgis-2.1, postgresql-contrib, postgresql-server-dev-all]
   fedora: [postgis]
   gentoo: [dev-db/postgis,'=dev-db/postgresql-9*']
   ubuntu:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1907,15 +1907,10 @@ libgphoto-dev:
   fedora: [libgphoto2-devel]
   gentoo: [media-gfx/gphoto2]
   ubuntu:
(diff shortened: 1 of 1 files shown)


mikaelarguedas (changes requested):
Please provide a description as well as links to the packages you add for all OSs covered by this PR as per [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions)

cberner:
updated. Let me know if there's anything else!

rosdep/base.yaml
```diff
@@ -1906,16 +1906,7 @@ libgphoto-dev:
...
-    lucid: [libgphoto2-dev]
-    precise: [libgphoto2-2-dev]
-    raring: [libgphoto2-2-dev]
-    saucy: [libgphoto2-6-dev]
-    trusty: [libgphoto2-dev]
-    utopic: [libgphoto2-dev]
-    vivid: [libgphoto2-dev]
-    wily: [libgphoto2-dev]
-    xenial: [libgphoto2-dev]
+  ubuntu: [libgphoto2-dev]
```
mikaelarguedas: these rules were not all identical, e.g. precise raring and saucy mapped to a different package.
We could use the wildcard syntax to avoid having to list out every new distro but keeping the mapping for the previous distros that had different rules:
```
  ubuntu:
    '*': [libgphoto2-dev]
    precise: [libgphoto2-2-dev]
    raring: [libgphoto2-2-dev]
    saucy: [libgphoto2-6-dev]
```

cberner:
updated. Thanks for the comments!

cberner:
@mikaelarguedas any other changes you'd like to see?

rosdep/base.yaml
```diff
@@ -4104,6 +4099,18 @@ postgresql-client:
   fedora: [postgresql-client]
   gentoo: [dev-db/postgresql]
   ubuntu: [postgresql-client]
+postgresql-postgis:
```
mikaelarguedas: this new key seem very similar to `postgresql-9.x-postgis` above.
Though bionic is the first version using 10.X instead of 9.X, so this new key is version agnostic :+1:

@tfoote how do you feel about this ? should we add the EOL rules from https://github.com/ros/rosdistro/blob/8bb80c8ce3bcbb570acf3f9d77e3358c486d64cc/rosdep/base.yaml#L4098 here ? To have one key that works for all distributions even outdated ones ?
Or are you ok merging this as is ?

@cberner can you please apply the fix of the missing packages (contrib and server) for debian to the existing `postgresql-9.x-postgis` key as well? Or confirm that the current rules are equivalent in debian and ubuntu ? (https://github.com/ros/rosdistro/pull/10830)

cberner: done

mikaelarguedas (approved):

PR #17846, opened 2018-05-14
update python-gitlab

maralbus:
@mikaelarguedas 
I had to make several changes to the `python-gitlab` packages, because the package name is ambiguous depending on the release version (e.g. **xenial**: `python-gitlab` --> `pyapi-gitlab` (7.5.0) [link](https://github.com/pyapi-gitlab/pyapi-gitlab) but **cosmic**: `python-gitlab` --> `python-gitlab` (1.3.0) [link](https://github.com/python-gitlab/python-gitlab)).
The same problem occurs with debian.
Changes:
- `python-gitlab`/`python3-gitlab` native if possible / pip otherwise
- new pkg `pyapi-gitlab`/`pyapi3-gitlab` native if possible / pip otherwise

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -135,30 +135,6 @@ pika:
     pip:
       packages: [pika]
   ubuntu: [python-pika]
-pyapi-gitlab:
-  debian:
-    '*':
-      pip:
-        packages: [pyapi-gitlab]
-    stretch: [python-gitlab]
-  ubuntu:
-    '*':
-      pip:
-        packages: [pyapi-gitlab]
-    artful: [python-gitlab]
-    xenial: [python-gitlab]
-pyapi3-gitlab:
-  debian:
-    '*':
-      pip:
-        packages: [python-gitlab]
-    stretch: [python3-gitlab]
-  ubuntu:
-    '*':
-      pip:
-        packages: [python-gitlab]
-    artful: [python3-gitlab]
-    xenial: [python3-gitlab]
 pydocstyle:
   debian:
     buster: [pydocstyle]
@@ -1254,17 +1230,15 @@ python-github-pip:
       packages: [PyGithub]
 python-gitlab:
   debian:
-    '*':
+    jessie:
       pip:
         packages: [python-gitlab]
     buster: [python-gitlab]
-    sid: [python-gitlab]
   ubuntu:
+    '*': [python-gitlab]
     artful:
       pip:
         packages: [python-gitlab]
-    bionic: [python-gitlab]
-    cosmic: [python-gitlab]
     trusty:
       pip:
         packages: [python-gitlab]
@@ -4114,15 +4088,8 @@ python3-flake8:
   ubuntu: [python3-flake8]
 python3-gitlab:
   debian:
-    '*':
-      pip:
-        packages: [python3-gitlab]
     buster: [python3-gitlab]
-    sid: [python3-gitlab]
   ubuntu:
-    '*':
-      pip:
-        packages: [python3-gitlab]
     bionic: [python3-gitlab]
     cosmic: [python3-gitlab]
 python3-nose:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1230,10 +1230,13 @@ python-github-pip:
       packages: [PyGithub]
 python-gitlab:
   debian:
+    '*': [python-gitlab]
+    stretch:
+      pip:
+        packages: [python-gitlab]
     jessie:
       pip:
         packages: [python-gitlab]
-    buster: [python-gitlab]
   ubuntu:
     '*': [python-gitlab]
     artful:
@@ -4088,10 +4091,24 @@ python3-flake8:
   ubuntu: [python3-flake8]
 python3-gitlab:
   debian:
-    buster: [python3-gitlab]
+    '*': [python3-gitlab]
+    jessie:
+      pip:
+        package: [python-gitlab]
+    stretch:
+      pip:
+        package: [python-gitlab]
   ubuntu:
-    bionic: [python3-gitlab]
-    cosmic: [python3-gitlab]
+    '*': [python3-gitlab]
+    artful:
+      pip:
+        package: [python-gitlab]
+    trusty:
+      pip:
+        package: [python-gitlab]
+    xenial:
+      pip:
+        package: [python-gitlab]
 python3-nose:
   debian: [python3-nose]
   gentoo: [dev-python/nose]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1230,15 +1254,23 @@ python-github-pip:
       packages: [PyGithub]
 python-gitlab:
   debian:
-    '*': [python-gitlab]
-    jessie:
+    '*':
       pip:
         packages: [python-gitlab]
+    buster: [python-gitlab]
+    sid: [python-gitlab]
```
mikaelarguedas: Please don;t add sid as per contributions guildelines: "However please don't target 'sid' as it's a rolling target and when the keys change our database gets out of date."

rosdep/python.yaml
```diff
@@ -1230,15 +1254,23 @@ python-github-pip:
...
   ubuntu:
-    '*': [python-gitlab]
```
mikaelarguedas: this can stay to match bionic and cosmic.

rosdep/python.yaml
```diff
@@ -4082,15 +4114,17 @@ python3-flake8:
   ubuntu: [python3-flake8]
 python3-gitlab:
   debian:
-    '*': [python3-gitlab]
-    jessie:
+    '*':
```
mikaelarguedas: the wildcard should match the apt package as this will be the default used for any new packages. So widlcards should not resolve to pip rules.

rosdep/python.yaml
```diff
@@ -4082,15 +4114,17 @@ python3-flake8:
...
       pip:
-        packages: [python-gitlab]
+        packages: [python3-gitlab]
```
mikaelarguedas: I don;t think the pip package python3-gitlab exists, same below

rosdep/python.yaml
```diff
@@ -4082,15 +4114,17 @@ python3-flake8:
...
+    buster: [python3-gitlab]
+    sid: [python3-gitlab]
```
mikaelarguedas: please remove sid

rosdep/python.yaml
```diff
@@ -4082,15 +4114,17 @@ python3-flake8:
...
   ubuntu:
-    '*': [python3-gitlab]
-    trusty:
+    '*':
```
mikaelarguedas: wildcard should resolve to apt rule like it was before

rosdep/python.yaml
```diff
@@ -1230,15 +1254,23 @@ python-github-pip:
       packages: [PyGithub]
 python-gitlab:
   debian:
-    '*': [python-gitlab]
-    jessie:
+    '*':
```
mikaelarguedas: wildcard should resolve to apt rule like it was before

rosdep/python.yaml
```diff
@@ -135,6 +135,30 @@ pika:
     pip:
       packages: [pika]
   ubuntu: [python-pika]
+pyapi-gitlab:
```
mikaelarguedas: waiting for feedback on the necessity to add these keys.
If we need to add them, all the commencts below apply here as well

mikaelarguedas (changes requested):
@ipa-flg-ma do you use both pyapi-gitlab and python-gitlab?

akaAlbo:
@mikaelarguedas No, I only need python-gitlab... but probably someone needs it in the future? Shall I apply all comments on `pyapi-gitlab`, too? Or do you want to wait until someone has need for it?

ipa-flg-ma:
@mikaelarguedas No, I only need python-gitlab... but probably someone needs it in the future? Shall I apply all comments on `pyapi-gitlab`, too? Or do you want to wait until someone has need for it?

rosdep/python.yaml
```diff
@@ -1230,15 +1230,21 @@ python-github-pip:
       packages: [PyGithub]
 python-gitlab:
   debian:
-    '*': [python-gitlab]
     jessie:
       pip:
         packages: [python-gitlab]
+    buster: [python-gitlab]
```
mikaelarguedas: there is now no rule for `stretch`.
Something like this should do the job (explicit rule for jessie and stretch, wildcard for buster and releases that will come after)
```
   debian:
     '*': [python-gitlab]
     jessie:
       pip:
         packages: [python-gitlab]
     stretch:
       pip:
         packages: [python-gitlab]
```

rosdep/python.yaml
```diff
@@ -4082,15 +4088,10 @@ python3-flake8:
...
-      pip:
-        packages: [python-gitlab]
+    buster: [python3-gitlab]
   ubuntu:
-    '*': [python3-gitlab]
-    trusty:
-      pip:
-        packages: [python-gitlab]
+    bionic: [python3-gitlab]
+    cosmic: [python3-gitlab]
```
mikaelarguedas: Same here, we need rule for all active ubuntu distributions, with this change there is no rule for `trusty`, `xenial` and `artful`.
Can you please apply the same logic as suggested in the comments above?

rosdep/python.yaml
ipa-flg-ma: oh, OK, so the wildcard is used for coming releases too...

mikaelarguedas: Yes, sorry if it was unclear. The main goal of introducing wildcards is to allow us to not have to add one new entry everytime a new OS distro comes out. So ideally we would have specific rules for anything that diverges from what is in the latest distro, and use the wildcard for the latest and upcoming distributions

ipa-flg-ma: I had a misunderstanding for the wildcards, thank you for claryfing.

rosdep/python.yaml
```diff
@@ -1231,14 +1231,23 @@ python-github-pip:
 python-gitlab:
   debian:
     '*': [python-gitlab]
+    stretch:
```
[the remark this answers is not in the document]
ipa-flg-ma: damn, hopefully my programming skills are better than my knowledge of the alphabet

mikaelarguedas (approved):
lgtm, thanks @ipa-flg-ma for catching this weird mix-n-match between `python-gitlab` and `pyapi-gitlab` :+1: 

And thanks for the multiple iterations on the PR!

ipa-flg-ma:
I have to thank you @mikaelarguedas for the kind support and guidance!
Btw, why is it `packages` and not the singular?

mikaelarguedas:
> Btw, why is it `packages` and not the singular?

For a given platform we provide a list `[]`, the list can be extended with other packages. As the list can contain 0 to many packages we use `packages`.

ipa-flg-ma:
Thanks for the explanation

tfoote (approved):

PR #17864, opened 2018-05-15
py_trees: 0.6.0-0 in 'melodic/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `py_trees` to `0.6.0-0`:

- upstream repository: https://github.com/stonier/py_trees.git
- release repository: https://github.com/stonier/py_trees-release.git
- bloom version: `0.6.4`

## py_trees

```
* Python 2/3 compatibility
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -1517,6 +1517,22 @@ repositories:
       url: https://github.com/ros/pluginlib.git
       version: melodic-devel
     status: maintained
+  py_trees:
+    doc:
+      type: git
+      url: https://github.com/stonier/py_trees.git
+      version: release/0.6-melodic
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/stonier/py_trees-release.git
+      version: 0.6.0-0
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/stonier/py_trees.git
+      version: devel
+    status: maintained
   python_qt_binding:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -1517,6 +1517,22 @@ repositories:
...
+      type: git
+      url: https://github.com/stonier/py_trees.git
+      version: release/0.6-melodic
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/stonier/py_trees-release.git
+      version: 0.6.0-0
+    source:
+      test_pull_requests: true
```
mikaelarguedas: Sanity check: Have you followed the [instructions](http://wiki.ros.org/buildfarm/Pull%20request%20testing) to provide access to this repo to the @ros-pull-request-builder user? As this is a personnal organization I think you will have to setup the webhook manually. Otherwise this line should be removed as it will create a job on the farm that will never get triggered

stonier: Aye, have it open right now. Was wondering about the 'organization' specific mention.

I'll setup the manual webhook. Still needs the test_pull_requests flag though right?

mikaelarguedas: > I'll setup the manual webhook. Still needs the test_pull_requests flag though right?

Yep :+1: the buildfarm [defaults to false](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/ffce76e9d5ed6d1c5c0d6c5b07b2ff181f7edba9/melodic/source-build.yaml#L56-L57) so we need to explicitly have the line here

mikaelarguedas: Can you please comment here once the webhook is setup ? thanks !

stonier: The webhook is good to go.

PR #17874, opened 2018-05-17
Adding hls-lfcd-lds2-driver to documentation index for distro

HLDS-GIT:
I'd like hls-lfcd-lds2-driver to be indexed and documented on ros.org.


tfoote (changes requested):
This PR now conflicts with master and needs to be rebased before travis can run and we can consider merging.

HLDS-GIT:
I'd like hls-lfcd-lds2-driver to be indexed and documented on ros.org.

tfoote:
Replaced by: #17890

PR #17883, opened 2018-05-17
remove map_merge from lunar

hrnr:
I have decided to do binary release only for upcoming melodic via #17882. This removes the doc job from lunar to avoid confusion on the wiki.

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -1651,16 +1651,6 @@ repositories:
       url: https://github.com/hrnr/m-explore.git
       version: lunar-devel
     status: developed
-  map_merge:
-    doc:
-      type: git
-      url: https://github.com/hrnr/map-merge.git
-      version: lunar-devel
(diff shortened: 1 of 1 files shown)


tfoote:
In general we recommend having the doc and source entries there in case users want to use packages from source even if you're not going to release binaries. We work to differentiate the availability of binary packages from just having documentation.

hrnr:
Sounds reasonable.

The trouble is that while this package works with ROS lunar, it also needs PCL>=1.8. So basically it works only with Debian Stretch ROS Lunar combination. This causes confusion to users (https://github.com/hrnr/map-merge/issues/1). I think it is more clear to only have this available in melodic, which have PCL>=1.8 available on every supported distro.

I'm sorry for introducing this package to lunar at the first place, but was working on Debian and I only realised the PCL problem later.

tfoote:
Ok, that makes sense.

PR #17917, opened 2018-05-21
Adding  my_repo "dlut_path_planner" to documentation index for distro

DS921020:
I'd like to my_repo "dlut_path_planner" to be indexed and documented on ros.org

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -2699,15 +2699,6 @@ repositories:
       type: git
       url: https://github.com/ZhuangYanDLUT/dlut_laser.git
       version: indigo-devel
-  dlut_path_planner:
-    doc:
-      type: git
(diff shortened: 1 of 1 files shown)


DS921020:
I'd like to my_repo "dlut_path_planner" to be indexed and documented on ros.org

DS921020:
a new way of building really time local map

dirk-thomas:
The [package name](https://github.com/DS921020/Hexagon_map/blob/f4bb965e73093bcf87226483eb421e6a7a35ed6f/package.xml#L3) doesn't follow the current naming conventions. You might want to use lower case only instead.

Also please use a [valid email address](https://github.com/DS921020/Hexagon_map/blob/f4bb965e73093bcf87226483eb421e6a7a35ed6f/package.xml#L10) for the maintainer. Otherwise the buildfarm can't notify you of any issues.

PR #17920, opened 2018-05-21
extract rosconsole

dirk-thomas:
Holding for new `ros_comm` release...


dirk-thomas:
Merging the patch into #17921 to get both changes in at once without conflicts.

PR #17931, opened 2018-05-22
Add svg.path to rosdep/python.yaml

d4n1elchen:
* ubuntu: https://packages.ubuntu.com/search?keywords=python-svg.path
* debian: https://packages.debian.org/sid/python/python-svg.path
* fedora: https://admin.fedoraproject.org/pkgdb/package/rpms/python-svg-path

Add this package for implementing some feature about picture
drawing which using svg files as input.

There's some spacing modification by `scrips/clean_rosdep_yaml.py`.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1377,8 +1377,8 @@ python-gridfs:
 python-gst:
   arch: [gstreamer0.10-python]
   debian:
-    jessie:  [python-gst0.10]
-    wheezy:  [python-gst0.10]
+    jessie: [python-gst0.10]
+    wheezy: [python-gst0.10]
   fedora: [gstreamer-python]
   gentoo: [dev-python/gst-python]
   ubuntu:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3449,6 +3449,14 @@ python-support:
     pip:
       packages: []
   ubuntu: [python-support]
+python-svg.path:
+  debian: [python-svg.path]
```
mikaelarguedas: As this is not available on Debian Jessie we should add a pip rule for Jessie

mikaelarguedas (commented):
Thanks for providing the links and description!

The change looks good to me except for debian where we need to add special handling for debian jessie

tfoote (dismissed):
As Mikael mentioned The same wildcard arrangement for jessie as trusty would make this complete.

team6612:
Updated!

mikaelarguedas (approved):
lgtm, thanks @team6612

tfoote (approved):

PR #17933, opened 2018-05-22
Adding ftm_msgs to documentation index for distro

guykhazma:
I would like to be indexed and documented on ros.org


mikaelarguedas:
@guykhazma not that if you want API documentation to be generated automatically you will need to add a `doc` entry as well

guykhazma:
@mikaelarguedas I have updated the entry and added a doc entry.
Thanks!

mikaelarguedas:
Yes you should create a new pull request for the addition of the doc section.
Thanks!

PR #17923, opened 2018-05-22
actionlib: 1.11.14-0 in 'melodic/distribution.yaml' [bloom]

mikaelarguedas:
Increasing version of package(s) in repository `actionlib` to `1.11.14-0`:

- upstream repository: https://github.com/ros/actionlib.git
- release repository: https://github.com/ros-gbp/actionlib-release.git
- distro file: `melodic/distribution.yaml`
- bloom version: `0.6.4`
- previous version for package: `1.11.13-0`

## actionlib

```
* Complete the full set of Ptr typedefs (#106 <https://github.com/ros/actionlib/issues/106>)
* Change boost::posix_time::milliseconds init to int64_t (#105 <https://github.com/ros/actionlib/issues/105>)
* Added ROS_ERROR message for Release code when asserts are ignored (#94 <https://github.com/ros/actionlib/issues/94>)
* fix typos. (#102 <https://github.com/ros/actionlib/issues/102>)
* Contributors: Bence Magyar, Patrick Beeson, Tobias Fischer, csukuangfj
```

PR #17934, opened 2018-05-22
Revert "actionlib: 1.11.14-0 in 'melodic/distribution.yaml' [bloom]"

clalancette:
Reverts ros/rosdistro#17923 - This caused breakage with realtime_tools which will also be reverted.


mikaelarguedas:
Code change looks good but commit list is fishy, we should squash and merge and remove the mention to other commits that are not actually reverted by this PR

PR #17924, opened 2018-05-22
realtime_tools: 1.12.0-0 in 'melodic/distribution.yaml'

mikaelarguedas:
Depends on https://github.com/ros/rosdistro/pull/17923

@bmagyar FYI

PR #17935, opened 2018-05-22
Revert "realtime_tools: 1.12.0-0 in 'melodic/distribution.yaml'"

clalancette:
Reverts ros/rosdistro#17924.  This caused breakage with downstream consumers of these packages.


mikaelarguedas:
same comment as https://github.com/ros/rosdistro/pull/17934#issuecomment-391035129

PR #17937, opened 2018-05-22
Fix clean rosdep yaml

### LINKED ISSUE #17930: scripts/clean_rosdep_yaml.py doesn't work properly with wildcards
I was running scripts/clean_rosdep_yaml.py because I failed the `nosetests`, but the clean-up scripts removed all colons that wrap the wildcard expressions, which caused errors when I run `nosetests` command again with the modified version.

```diff
496c496
<   gentoo: ['=dev-python/beautifulsoup-4*']
---
>   gentoo: [=dev-python/beautifulsoup-4*]
1034,1037c1034,1037
<     '21': [python-flake8]
<     '22': [python-flake8]
<     '23': [python-flake8]
<     '24': [python-flake8]
---
>     21: [python-flake8]
>     22: [python-flake8]
>     23: [python-flake8]
>     24: [python-flake8]
1233c1233
<     '*': [python-gitlab]
---
>     *: [python-gitlab]
1241c1241
<     '*': [python-gitlab]
---
>     *: [python-gitlab]
```

dirk-thomas:
Fixes #17930.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -831,7 +831,7 @@ gazebo:
 gazebo5:
   arch: [gazebo]
   fedora: [gazebo]
-  gentoo: ['=sci-electronics/gazebo-5*']
+  gentoo: [=sci-electronics/gazebo-5*]
   ubuntu: [gazebo5]
 gazebo7:
   arch: [gazebo]
@@ -840,7 +840,7 @@ gazebo7:
     jessie: [gazebo7]
     stretch: [gazebo7]
   fedora: [gazebo]
-  gentoo: ['=sci-electronics/gazebo-7*']
+  gentoo: [=sci-electronics/gazebo-7*]
   slackware: [gazebo]
   ubuntu: [gazebo7]
 gazebo9:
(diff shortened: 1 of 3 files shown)


dirk-thomas:
There are already many rules which don't have quotes (search for "gentoo: [=" in the `base.yaml`, e.g. https://github.com/ros/rosdistro/blob/a180372527178e97654a4156cc4f4292ef18da38/rosdep/base.yaml#L740). So I don't see how the once I changed in this patch could be any different.

mikaelarguedas:
> There are already many rules which don't have quotes

The rule you linked to is the only one not having the quotes so I wouldn't call it "many".

If you tested it on gentoo and are confident it works I'm fine with merging as is :+1: . Otherwise I would feel more comfortable having confirmation from gentoo users.

mikaelarguedas:
As the rules using use flags e.g. https://github.com/ros/rosdistro/pull/17937/files#diff-acba1c75aeaa27ae4af7a5e0a5b78c08R4190 or nested `[` are not modified by this PR I guess it's find to merge as is.

tfoote (approved):
I think this should be fine with gentoo, as the stringification is part of the yaml spec. There are possibly places that need the string quoted to be disambiguated but I think we're safe with the current changes.

awesomebytes:
Bit late to the party, but yeah, it's more of a yaml thing than gentoo I
guess (if the rosdep backend deals with lists, ofc).
All good.

allenh1:
Lgtm -- sorry for the late reply

PR #17948, opened 2018-05-23
add pyqrcode from pip

fmessmer:
as available here: https://pypi.org/project/PyQRCode/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2529,6 +2529,26 @@ python-pyproj:
     pip:
       packages: [pyproj]
   ubuntu: [python-pyproj]
+python-pyqrcode:
+  arch: [python-qrcode]
+  debian:
+    buster: [python-pyqrcode]
+    jessie:
+      pip:
+        packages: [PyQRCode]
+    stretch:
+      pip:
+        packages: [PyQRCode]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Is this the same package as https://packages.ubuntu.com/bionic/python-pyqrcode ?

If so please use the apt version rather than pip for the disctributions where it's available

ipa-fxm:
I also added the packages for
arch: https://www.archlinux.org/packages/?sort=&q=qrcode&maintainer=&flagged=
debian: https://packages.debian.org/search?suite=default&section=all&arch=any&lang=de&searchon=names&keywords=python-qrcode
gentoo: https://packages.gentoo.org/packages/dev-python/pyqrcode

mikaelarguedas:
@ipa-fxm thanks for iterating.
Please see review of https://github.com/ros/rosdistro/pull/17950 for details about:
- how to add keys in a future proof way
- not adding entries for EOL distros
- providing pip keys for debian distributions that don't have the package

ipa-fxm:
I hope everything looks good now

mikaelarguedas (approved):
lgtm, thanks @ipa-fxm for iterating :+1:

PR #17950, opened 2018-05-23
add pypng from pip

fmessmer:
as available here: https://pypi.org/project/pypng/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2516,6 +2516,28 @@ python-pynmea2:
   ubuntu:
     pip:
       packages: [pynmea2]
+python-pypng:
+  arch:
+    pip:
+      packages: [pypng]
+  debian:
+    buster: [python-png]
+    jessie:
+      pip:
+        packages: [pypng]
+    stretch:
+      pip:
+        packages: [pypng]
+  fedora: [python-pypng]
+  gentoo: [dev-python/pypng]
+  ubuntu:
+    '*': [python-png]
+    trusty:
+      pip:
+        packages: [pypng]
+    xenial:
+      pip:
+        packages: [pypng]
 python-pypozyx-pip:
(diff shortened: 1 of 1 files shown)


ipa-fxm:
thanks for the review...I updated the PR

package is available for
gentoo: https://packages.gentoo.org/packages/dev-python/pypng

but not for the other variants :point_right: keeping pip

rosdep/python.yaml
```diff
@@ -2516,6 +2516,37 @@ python-pynmea2:
...
+  debian: [python-png]
+  fedora: [python-pypng]
+  gentoo: [dev-python/pypng]
+  ubuntu:
+    artful: [python-png]
+    bionic: [python-png]
+    trusty:
+      pip:
+        packages: [pypng]
+    utopic:
```
mikaelarguedas: We usually don add new keys for EOL distributions. Can you please keep only trusty and xenial here?

For future proofness, we can use the new wildcard syntax to cover upcoming distributions as wel like cosmic:
```
ubuntu:
  '*': [python-png]
  trusty:
    pip:
      packages: [pypng]
  xenial:
    pip:
      packages: [pypng]
```

rosdep/python.yaml
```diff
@@ -2516,6 +2516,37 @@ python-pynmea2:
   ubuntu:
     pip:
       packages: [pynmea2]
+python-pypng:
+  arch:
+    pip:
+      packages: [pypng]
+  debian: [python-png]
```
mikaelarguedas: This is only available on buster, we need to provide pip entries for the active distributions that don;t have that package: `jessie` and `stretch`

ipa-fxm:
I hope everything looks good now

mikaelarguedas (approved):
lgtm, thanks @ipa-fxm !

PR #18012, opened 2018-05-28
[rosdep] fix setserial gentoo rosdep rule

mikaelarguedas:
Adds the missing `sys-apps` category to the `setserial` rosdep gentoo rule.
Gentoo: https://packages.gentoo.org/packages/sys-apps/setserial
Follow up of https://github.com/ros/rosdistro/pull/18006


mikaelarguedas:
Merging without second review as I see this PR as the second review of #18006

PR #18112, opened 2018-06-05
Added rosdeps for glfw3.

remod:
This adds a rosdep key for [glfw3](http://www.glfw.org/).

- debian: https://packages.debian.org/jessie/libglfw3-dev
- ubuntu: https://packages.ubuntu.com/xenial/libdevel/libglfw3-dev

I am using this dependency to compile [librealsense](https://github.com/IntelRealSense/librealsense) as a catkin package. Tested on Ubuntu 16.04.

I am not sure whether the key shall be `glfw` or `glfw3`. On Ubuntu 16.04, the following packages are available:

- libglfw2
- libglfw3
- libglfw3-dev
- libglfw-dev (installs libglfw2)

For compiling librealsense, I need libglfw3-dev, therefore `glfw3` would more specific. However, on some the distributions I googled (e.g. gentoo: https://packages.gentoo.org/packages/media-libs/glfw), it seems that it is not possible to select the version to install and you just get the latest.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -965,12 +965,6 @@ gksu:
 glc:
   arch: [glc]
   gentoo: [media-libs/quesoglc]
-glfw3:
-  debian: [libglfw3-dev]
-  gentoo: [media-libs/glfw]
-  ubuntu:
-    '*': [libglfw3-dev]
-    trusty: null
 glut:
   arch: [freeglut]
   debian: [freeglut3-dev]
@@ -1843,6 +1837,12 @@ libglew-dev:
     xenial: [libglew-dev]
     yakkety: [libglew-dev]
     zesty: [libglew-dev]
+libglfw3-dev:
+  debian: [libglfw3-dev]
+  gentoo: [media-libs/glfw]
+  ubuntu:
+    '*': [libglfw3-dev]
+    trusty: null
 libglib-dev:
   arch: [glib2]
   debian: [libglib2.0-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -967,7 +967,9 @@ glc:
   gentoo: [media-libs/quesoglc]
 glfw3:
   debian: [libglfw3-dev]
-  ubuntu: [libglfw3-dev]
+  ubuntu:
+    '*': [libglfw3-dev]
+    trusty: null
 glut:
   arch: [freeglut]
   debian: [freeglut3-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -965,6 +965,9 @@ gksu:
 glc:
   arch: [glc]
   gentoo: [media-libs/quesoglc]
+glfw3:
+  debian: [libglfw3-dev]
+  ubuntu: [libglfw3-dev]
 glut:
   arch: [freeglut]
   debian: [freeglut3-dev]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
> I am not sure whether the key shall be glfw or glfw3

The key should match the ubuntu / debian package name. If you want `libglfw3-dev` that's the name of the key we should pick.

Unfortunately `libglfw3`(-dev) does not exist on Ubuntu Trusty, that is one of our targeted active Ubuntu distribution.
We will need to add a special case for trusty to say that this key cannot be satisfied on that platform.
Something like:
```
libglfw3-dev:
  debian: [libglfw3-dev]
  ubuntu:
    '*': [libglfw3-dev]
    trusty: null
```

remod:
Thank you for your comment.

I updated the PR accordingly. As I haven't seen this syntax somewhere else in the file, can rosdep handle this?

mikaelarguedas:
> (e.g. gentoo: https://packages.gentoo.org/packages/media-libs/glfw), it seems that it is not possible to select the version to install and you just get the latest.

Yes Gentoo sometimes provides only the latest. It's fine as gentoo is a rolling release. Please add it to the list of rules as, currently, `media-libs/glfw` points to version 3.

> As I haven't seen this syntax somewhere else in the file, can rosdep handle this?

Both the wildcard and null syntax have been added "recently" https://github.com/ros-infrastructure/rep/pull/153. While we used the wildcard in a couple of places, I think this is the first time we use the "null". I copied the example from https://github.com/ros-infrastructure/rep/blob/master/rep-0111.rst#allow-package_argument-to-be-explicitly-null . According to rosdep's changelog, support for both has been added in 0.12.0: https://github.com/ros-infrastructure/rosdep/blob/master/CHANGELOG.rst#0120-2018-02-07

remod:
That's great, thanks. I added the Gentoo package.

rosdep/base.yaml
```diff
@@ -965,6 +965,12 @@ gksu:
 glc:
   arch: [glc]
   gentoo: [media-libs/quesoglc]
+glfw3:
```
mikaelarguedas: please rename the key to `libglfw3-dev` as per initial review (note that the block will need to be moved to the appropriate location to respect alphabetical order)

remod: Done, sorry for the misunderstanding.

mikaelarguedas:
thanks @remod for iterating.
While rosdep supports the new syntax, the rosdistro tests don't know about it ATM.
I opened https://github.com/ros/rosdistro/pull/18113 to add support for it.

I tested this new key in a trusty container and I get what I think is the expected behavior:
```
root@a468953e6c30:/# rosdep resolve glfw3

ERROR: [glfw3] defined as "not available" for OS version [trusty]

[glfw3] defined as "not available" for OS version [trusty]
	rosdep key : glfw3
	OS name    : ubuntu
	OS version : trusty
	Data: debian:
- libglfw3-dev
ubuntu:
  '*':
  - libglfw3-dev
  trusty: null
```

mikaelarguedas (approved):
lgtm, CI is now happy

bluryi:
This change is breaking `rosdep db` on Ub 14.04. 
`ERROR: Rosdep experienced an error: rosdep OS definition for [libglfw3-dev:ubuntu] must be a dictionary, string, or list: None`

IMO we don't need to make trusty as a special case, that if any package requires libglfw3-dev in Ub 14.04, just let the build fail.

dirk-thomas:
@bluryi You likely haven't installed the latest version of the `python-rosdep` package. Please update that package and the current rosdep database should be readable without a problem.

bluryi:
@dirk-thomas  thanks!

PR #18098, opened 2018-06-03
raspi_temperature: 0.1.1-0 in 'kinetic/distribution.yaml' [bloom]

bberrevoets:
Increasing version of package(s) in repository `raspi_temperature` to `0.1.1-0`:

- upstream repository: https://github.com/bberrevoets/raspi_temperature.git
- release repository: https://github.com/bberrevoets/raspi_temperature-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.6.4`
- previous version for package: `null`

## raspi_temperature

```
* Release
* Corrected a little error
* Contributors: B.N.Berrevoets
```

PR #18117, opened 2018-06-05
Adding raspi_temperature to documentation index for kinetic

bberrevoets:
I'd like raspi_temperature to be indexed and documented on ros.org.


dirk-thomas:
This seems to be a duplicate of #18098 which has already been merged.

Also your branch has diverged from master and shows as currently not mergable (probably because it conflicts with the already merged patch?).

PR #18123, opened 2018-06-05
added pyqt5.qsci and sysv-ipc packages as ubuntu xenial rule

okanasik:
We are working on JdeRobot/VisualStates (https://github.com/JdeRobot/VisualStates) project to release it as a ros package. In the project, we have two python dependencies that are not part of rosdep.
- `python-qt5.qsci` is a python bindings of QScintilla source code editor library. We use this package to provide a source code editor to the users.
- `python-sysv-ipc` is a python package to provide system v inter process communication functionality. We use this package to communicate python and c++ programs.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2812,6 +2812,9 @@ python-qt5-bindings-gl:
     xenial: [python-pyqt5.qtopengl]
     yakkety: [python-pyqt5.qtopengl]
     zesty: [python-pyqt5.qtopengl]
+python-qt5-bindings-qsci:
+  ubuntu:
+    xenial: [python-pyqt5.qsci]
 python-qt5-bindings-webkit:
   arch: [python2-pyqt5]
   debian:
@@ -3574,6 +3577,9 @@ python-sympy:
     wily_python3: [python3-sympy]
     xenial: [python-sympy]
     xenial_python3: [python3-sympy]
+python-sysv-ipc:
+  ubuntu:
+    xenial: [python-sysv-ipc]
 python-tablib:
   debian: [python-tablib]
   fedora: [python-tablib]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3580,7 +3580,7 @@ python-sympy:
     xenial_python3: [python3-sympy]
 python-sysv-ipc:
   debian: [python-sysv-ipc]
-  fedora: [python-sysv-ipc]
+  fedora: [python2-sysv-ipc]
   ubuntu: [python-sysv-ipc]
 python-tablib:
   debian: [python-tablib]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2813,8 +2813,9 @@ python-qt5-bindings-gl:
     yakkety: [python-pyqt5.qtopengl]
     zesty: [python-pyqt5.qtopengl]
 python-qt5-bindings-qsci:
-  ubuntu:
-    xenial: [python-pyqt5.qsci]
+  debian: [python-pyqt5.qsci]
+  fedora: [python2-qscintilla-qt5]
+  ubuntu: [python-pyqt5.qsci]
 python-qt5-bindings-webkit:
   arch: [python2-pyqt5]
   debian:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
thanks @okanasik for the contribution.

However I encourage you to refer to the [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

### Provide links to the packages:

"For convenience in reviewing please link to the web listings of packages such as on http://packages.ubuntu.com, http://packages.debian.org, and https://admin.fedoraproject.org/pkgdb/packages/ or if a pip package pypi.python.org."

### Package description:

"Please also briefly describe the package being added and what use case you want to use it for. It's valuable to have a record of the packages as submitted and their intended purpose for clarity in the future so that if there's a conflict there's information to fall back on instead of speculation about the original use cases."

### Platforms that should be covered:

"Rules can be contributed for any platform. However to be released they must be at least cover the supported platforms in REP 3: http://www.ros.org/reps/rep-0003.html" so please provide at least Ubuntu and Debian rules for all active distributions

rosdep/python.yaml
```diff
@@ -2812,6 +2812,9 @@ python-qt5-bindings-gl:
     xenial: [python-pyqt5.qtopengl]
     yakkety: [python-pyqt5.qtopengl]
     zesty: [python-pyqt5.qtopengl]
+python-qt5-bindings-qsci:
+  ubuntu:
+    xenial: [python-pyqt5.qsci]
```
tfoote: This looks to be availabe on xenial and newer: https://packages.ubuntu.com/xenial/amd64/python-pyqt5.qsci I'd rather see a non-versioned rule for it so we don't have to updated it in the future, and it will work for newer platforms too.

tfoote: It's also available on stretch and newer: https://packages.debian.org/sid/python-pyqt5.qsci A key for debian would be good.

tfoote: The generic fedora and gentoo rules could also be copied in here.

rosdep/python.yaml
```diff
@@ -3574,6 +3577,9 @@ python-sympy:
     wily_python3: [python3-sympy]
     xenial: [python-sympy]
     xenial_python3: [python3-sympy]
+python-sysv-ipc:
+  ubuntu:
+    xenial: [python-sysv-ipc]
```
tfoote: This is available on jessie and newer and should have a rule: https://packages.debian.org/jessie/python-sysv-ipc

Also it's on xenial and newer too: https://packages.ubuntu.com/xenial/python-sysv-ipc

Fedora and gentoo rules would be appreciated as well.

tfoote (changes requested):

tfoote (approved):
It looks like it might be cleaner to use the newer fedora package `python2-sysv-ipc` https://fedora.pkgs.org/26/fedora-x86_64/python2-sysv_ipc-0.7.0-4.fc26.x86_64.rpm.html it obsoletes `python-sysv-ipc` So I bumped it.

rosdep/python.yaml
```diff
@@ -2812,6 +2812,10 @@ python-qt5-bindings-gl:
     xenial: [python-pyqt5.qtopengl]
     yakkety: [python-pyqt5.qtopengl]
     zesty: [python-pyqt5.qtopengl]
+python-qt5-bindings-qsci:
+  debian: [python-pyqt5.qsci]
```
mikaelarguedas: This package is not available on `jessie`, we should provide either a pip rule or a `null` rule for Debian Jessie

mikaelarguedas (changes requested):
@tfoote in your previous review you mentionned propagating gentoo rules to these keys ?

rosdep/python.yaml
tfoote: I think we can ignore that for jessie as it's EOL'd today.

tfoote:
I thought that this was a subpart of the generic qt5 package on fedora, but it's actually an editor so there might be a gentoo rule specific for it but I don't think the generic one will necessarily cover it without more research so I'd suggest not blocking on it being added.

mikaelarguedas (approved):
Jessie is EOL

mikaelarguedas:
Thanks @okanasik for iterating and providing the descriptions :+1:

PR #18135, opened 2018-06-06
Adding topics_rviz_plugin to documentation index for melodic

VictorLamoine:
I have also changed the status of the package to `maintained` on `kinetic`.


mikaelarguedas:
Looks like there is only a doc entry for kinetic and no source entry. And this PR adds a source entry for melodic but no doc entry. Is it intentional?

VictorLamoine:
No this is an error on my side, thanks for catching it! I have fixed my pull request.

PR #18151, opened 2018-06-07
Remove EOL Release platforms Debian Jessie, Fedora 23 and 24

tfoote:
This is a follow on to https://github.com/ros-infrastructure/ros_buildfarm_config/pull/118

Holding until we've confirmed jessie is shutdown cleanly.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3,11 +3,6 @@
 # see REP 143: http://ros.org/reps/rep-0143.html
 ---
 release_platforms:
-  debian:
-  - jessie
-  fedora:
-  - '23'
-  - '24'
   ubuntu:
   - xenial
 repositories:


mikaelarguedas (approved):
:+1:

dirk-thomas (approved):

tfoote:
Since it's been close to a week and there's been no known issues I will merge this.

dirk-thomas:
@tfoote Some jobs seem to not be disabled yet on build.ros.org - I noticed e.g. the PR job for Debian Jessie.

ipa-mdl:
This change breaks rosdep for jessie:

```
$ rosdep resolve catkin

ERROR: No definition of [catkin] for OS [debian]

No definition of [catkin] for OS [debian]
        rosdep key : catkin
        OS name    : debian
        OS version : jessie
        Data: _is_ros: true
osx:
  homebrew:
    packages:
    - ros/kinetic/catkin
ubuntu:
  xenial:
    apt:
      packages:
      - ros-kinetic-catkin
```
Not sure if this was intended.

tfoote:
It's not directly intended but is actually basically correct. Since the rosdistro hasn't diverged much right now it would mostly keep working. However in general it should not be expected to keep working. New packages would incorrectly resolve etc. We likely should take a snapshot of the current state of the repository and export that as a jessie specific rosdep rules file that users can add to their rosdep sources.

ipa-mdl:
> We likely should take a snapshot of the current state of the repository and export that as a jessie specific rosdep rules file that users can add to their rosdep sources.

To keep the maintenance easy: perhaps just a `jessie_eol` tag and a sources list that references these EOL distros with a higher priority?
(Haven't had a look yet at how rosdep handles this internally..)

ipa-mdl:
I just stumbled upon this because rosdep still works for hydro on 12.04. Users might be confused that `rosdep update` breaks their set-up on jessie.

PR #18153, opened 2018-06-08
Add equivalent libturbojpeg rules for Artful, Bionic, and Cosmic

cberner:


rosdep/base.yaml
```diff
@@ -3113,7 +3113,11 @@ libturbojpeg:
   fedora: [turbojpeg-devel]
   freebsd: [libjpeg-turbo]
   gentoo: [media-libs/libjpeg-turbo]
-  ubuntu: [libturbojpeg, libjpeg-turbo8-dev]
+  ubuntu: 
+    '*': [libturbojpeg, libjpeg-turbo8-dev]
+    artful: [libturbojpeg, libjpeg-turbo8-dev, libturbojpeg0-dev]
```
mikaelarguedas: wouldn't just listing `libturbojpeg0-dev` be enough ? it looks like it depends on the 2 other packages

rosdep/base.yaml
```diff
@@ -3113,7 +3113,11 @@ libturbojpeg:
   fedora: [turbojpeg-devel]
   freebsd: [libjpeg-turbo]
   gentoo: [media-libs/libjpeg-turbo]
-  ubuntu: [libturbojpeg, libjpeg-turbo8-dev]
+  ubuntu: 
+    '*': [libturbojpeg, libjpeg-turbo8-dev]
```
mikaelarguedas: for future-proof-ness we may want to inverse the logic here to keep the wildcard for "any future distro". This will allow us to no have to add a new rule everytime a new distro comes out

rosdep/base.yaml
```diff
@@ -3113,7 +3113,11 @@ libturbojpeg:
   fedora: [turbojpeg-devel]
   freebsd: [libjpeg-turbo]
   gentoo: [media-libs/libjpeg-turbo]
-  ubuntu: [libturbojpeg, libjpeg-turbo8-dev]
+  ubuntu:
```
mikaelarguedas: the `debian` rule needs to be updated as well, it currently lists `libturbojpeg1-dev` that is not available on Stretch and newer.
We should update it to use [`libturbojpeg1-dev` for jessie](https://packages.debian.org/jessie/libturbojpeg1-dev) and [`libturbojpeg0-dev` for all other distros](https://packages.debian.org/stretch/libturbojpeg0-dev)

mikaelarguedas (changes requested):
ci is failing due to trailing whitespaces.

Several suggested changes / improvements inline below

mikaelarguedas (approved):
lgtm

mikaelarguedas:
Ubuntu Artful and newer: https://packages.ubuntu.com/bionic/libturbojpeg0-dev
Debian Stretch and newer: https://packages.debian.org/stretch/libturbojpeg0-dev

PR #18178, opened 2018-06-11
rosdep/python: add python-pytest-qt

machinekoder:
Add the [pytest-qt](https://pypi.org/project/pytest-qt/) pytest extension.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2672,6 +2672,16 @@ python-pytest:
   fedora: [python-pytest]
   gentoo: [dev-python/pytest]
   ubuntu: [python-pytest]
+python-pytest-qt-pip:
+  debian:
+    pip:
+      packages: [pytest-qt]
+  fedora:
+    pip:
+      packages: [pytest-qt]
+  ubuntu:
+    pip:
+      packages: [pytest-qt]
 python-pytz-pip:
   ubuntu:
     pip:


rosdep/python.yaml
```diff
@@ -2672,6 +2672,25 @@ python-pytest:
   fedora: [python-pytest]
   gentoo: [dev-python/pytest]
   ubuntu: [python-pytest]
+python-pytest-qt:
+  debian:
+    buster: [pytest-qt]
+    jessie:
+      pip:
+        [pytest-qt]
+    wheezy:
```
mikaelarguedas: this is out of alphabetical order

rosdep/python.yaml
```diff
@@ -2672,6 +2672,25 @@ python-pytest:
   fedora: [python-pytest]
   gentoo: [dev-python/pytest]
   ubuntu: [python-pytest]
+python-pytest-qt:
+  debian:
+    buster: [pytest-qt]
```
mikaelarguedas: there is no binary deb available on buster yet.

mikaelarguedas (changes requested):
This package doesn't have binary packages available for our targeted platforms (Ubuntu and Debian).
So users cannot release packages depending on them

Can you please make this key a pip key (with the suffix `-pip`) ?

rosdep/python.yaml
machinekoder: You are right. There is one for python3 though: https://packages.debian.org/buster/python3-pytestqt

rosdep/python.yaml
```diff
@@ -2672,6 +2672,18 @@ python-pytest:
   fedora: [python-pytest]
   gentoo: [dev-python/pytest]
   ubuntu: [python-pytest]
+python-pytest-qt:
```
mikaelarguedas: We can remove that key altogether as it doesn't have any resolution for the platforms we require support for.

We will consider adding it back once it's available for all platforms targeted by at least one ROS distribution

machinekoder: sure

rosdep/python.yaml
mikaelarguedas: > There is one for python3 though: https://packages.debian.org/buster/python3-pytestqt

:+1:
Currently there is no ROS distribution targeting Python3. Though that will likely change in the future https://github.com/ros-infrastructure/rep/pull/149

mikaelarguedas (approved):
lgtm

PR #18190, opened 2018-06-12
Add libpython-dev & python-numpy

jlblancoc:


mikaelarguedas:
@jlblancoc python packages should be listed in the `python.yaml` file as per [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules).

Apparently the key `python-yaml` is already defined [there](https://github.com/ros/rosdistro/blob/cd95fbc592b2487413e0449b9f920e4dbe357eb9/rosdep/python.yaml#L1977) did rosdep fail to resolve it ?.

jlblancoc:
oh, no, rosdep didn't fail, it's just that I missed `python.yaml`, thanks for the pointer.
On `libpython-dev`, it seems strange to me that it's not already covered somewhere, but couldn't find it... 
If you know `libpython-dev` is already somewhere, please let me know, otherwise I'll close and open a new clean PR with just that addition.
Thanks.

mikaelarguedas:
> On `libpython-dev`, it seems strange to me that it's not already covered somewhere, but couldn't find it...

We have a [generic key called `python`](https://github.com/ros/rosdistro/blob/cd95fbc592b2487413e0449b9f920e4dbe357eb9/rosdep/python.yaml#L201) that pulls in `python-dev` (and transitively `libpython-dev`)

jlblancoc:
Great! Then, nothing to add here 👍

PR #18199, opened 2018-06-12
Updated libusb-1.0 rosdep keys for fedora

kmhallen:
https://apps.fedoraproject.org/packages/libusbx


rosdep/base.yaml
```diff
@@ -3208,6 +3208,10 @@ libusb-1.0:
     '22': [libusbx]
     '23': [libusbx]
     '24': [libusbx]
+    '25': [libusbx]
+    '26': [libusbx]
+    '27': [libusbx]
+    '28': [libusbx]
```
mikaelarguedas: @tfoote @dirk-thomas How do you feel about just dropping the beefy rule here ? (beefy being EOL since 2013)
This would allow us to just have: 

```
   fedora: [libusbx]
```

dirk-thomas: I am fine with that. Using `*` would solve the same problem without any "downside" though.

mikaelarguedas: Yeah I considered it as well.

@kmhallen the suggestion is to use the new wildcard syntax to use libusbx on all distros but beefy. It looks like:

```
   fedora:
     '*': [libusbx]
     beefy: [libusb1]
```

mikaelarguedas (approved):
lgtm, thanks @kmhallen

PR #18231, opened 2018-06-15
Add clang-format for all Ubuntu distros

paulbovbel:
clang-format is available xenial-onwards. Is it acceptable to provide it like this, or does each distro post xenial need to be specified independently?


tfoote (approved):
This looks good to me. In general if we add keys that overmatch on older distros that's ok. If people start using it on Trusty, the only older still supported distro it will fail to install instead of fail to resolve.

But since I expect it to be continuously available for all future platforms the more compact rule makes sense for long term maintenance.

PR #18238, opened 2018-06-18
Added pyshark and tshark to python.yaml rosdep rules

jack-digilabs:
Added pyshark and tshark to the rosdep rules

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2656,6 +2656,8 @@ python-pyramid:
   fedora: [python-pyramid]
   gentoo: [dev-python/pyramid]
   ubuntu: [python-pyramid]
+python-pyshark:
+  ubuntu: [python-pyramid]
 python-pyside:
   arch: [python2-pyside]
   debian: [python-pyside]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3897,6 +3897,8 @@ python-trep:
   ubuntu:
     pip:
       packages: [trep]
+python-tshark:
+  ubuntu: [tshark]
 python-twisted-bin:
   arch: [python2-twisted]
   debian: [python-twisted-bin]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2657,7 +2657,9 @@ python-pyramid:
   gentoo: [dev-python/pyramid]
   ubuntu: [python-pyramid]
 python-pyshark:
-  ubuntu: [python-pyramid]
+  ubuntu:
+    apt:
+      packages: [pyshark]
 python-pyside:
   arch: [python2-pyside]
   debian: [python-pyside]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2658,7 +2658,7 @@ python-pyramid:
   ubuntu: [python-pyramid]
 python-pyshark:
   ubuntu:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2656,6 +2656,10 @@ python-pyramid:
   fedora: [python-pyramid]
   gentoo: [dev-python/pyramid]
   ubuntu: [python-pyramid]
+python-pyshark:
```
tfoote: If this is pip only it should have a `-pip` ending.

rosdep/python.yaml
```diff
@@ -3893,6 +3897,8 @@ python-trep:
   ubuntu:
     pip:
       packages: [trep]
+python-tshark:
+  ubuntu: [tshark]
```
tfoote: Please contribute at least debian rules too. Fedora macOS and Gentoo would be appreciated as well.

tfoote (changes requested):
This would be great to add. Please see the contribution guidelines https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

There are a few more definitions that would be necessary and a few more would appreciated. 

Also please link to the package names/references as well for easier verification of the package names.

jack-digilabs:
I added Debian to tshark but I'm not certain about any platforms for tshark. I added Debian, fedora, gentoo, and osx to the pyshark rule and the -pip suffix.

rosdep/python.yaml
```diff
@@ -2656,6 +2656,22 @@ python-pyramid:
...
+      packages: [pyshark]
+  gentoo:
+    pip:
+      packages: [pyshark]
+  ubuntu:
+    pip:
+      packages: [pyshark]
+  osx:
+    pip:
+      packages: [pyquery]
```
tfoote: This is surprising to have a different pip name on osx. Can you link to a resource that explains why this is different?

rosdep/python.yaml
```diff
@@ -2656,6 +2656,22 @@ python-pyramid:
...
+      packages: [pyshark]
+  fedora:
+    pip:
+      packages: [pyshark]
+  gentoo:
+    pip:
+      packages: [pyshark]
+  ubuntu:
+    pip:
+      packages: [pyshark]
```
tfoote: Since these are pip wrappers for the underlying too. I suspect that they need tshark installed to be useful. If so these entries should have a depend key as well onto tshark.

tfoote (changes requested):
For reference:

Ubuntu: https://packages.ubuntu.com/xenial/net/tshark
Debian: https://packages.debian.org/stretch/tshark

Other keys I found for tshark
Fedora: `wireshark` https://fedora.pkgs.org/25/fedora-x86_64/wireshark-2.1.1-2.fc25.x86_64.rpm.html  wireshark includes tshark: https://stackoverflow.com/a/8237588

Gentoo: `net-analyzer/wireshark` https://packages.gentoo.org/packages/net-analyzer/wireshark

OSX/homebrew: `wireshark` http://brewformulas.org/Wireshark

mikaelarguedas:
@jack-digilabs friendly :bellhop_bell:

dirk-thomas:
@jack-digilabs  Closing due to no response.

PR #18251, opened 2018-06-21
[dep] Add debtree.

130s:
Debian https://packages.debian.org/search?keywords=debtree
Ubuntu https://packages.ubuntu.com/xenial/utils/debtree


tfoote (approved):
Debian specific so it makes sense not to only have those targets.

PR #18260, opened 2018-06-25
Adding tensorflow_ros_cpp (doc + source) to indigo, kinetic and melodic.

peci1:


peci1:
A [related jenkins job](http://build.ros.org/view/Idev/job/Idev__tensorflow_ros_cpp__ubuntu_trusty_amd64/lastBuild/console) has failed.

I only expected doc jobs to be run, but this is a dev job. What's its purpose? Should I care about the build failure, @tfoote ?

It fails because this package depends on tensorflow_catkin, which is an unindexed ROS package (however, it is available on github). But that package is an optional dependency, and W. J. Wood [recommends](https://answers.ros.org/question/202251/policy-on-run_depends-for-external-optional-binary-packages/?answer=202260#post-id-202260) "having all optional dependencies listed as run_depend dependencies." If I'd remove the dependency from package.xml, build order would be almost 100% wrong (because building tensorflow_catkin can take up to a few hours).

I see a few options:

 - remove the dependency and instruct users to build tensorflow_catkin in a different workspace
 - create a jenkins branch in tensorflow_ros_cpp that would "track" master but would always remove the line with the dependency (so that users get the dependency, but Jenkins doesn't)
 - try to convince the author of tensorflow_catkin to index his package

What's the preferred option?

Note (maybe irrelevant): neither tensorflow_ros_cpp, nor tensorflow_catkin make sense to be ever released as binary packages. Their role is to detect and utilize the specific environment the developer has on his machine.

tfoote:
Yes, we shouldn't leave the devel job hanging.

For the doc jobs we have a `depends` tag you can add to pull additional repos into your overlay before the documentation job is run http://www.ros.org/reps/rep-0141.html.

But the devel jobs are expected to test based on the declared dependencies explicitly. If you only wanted docs just add a `test_commits: false` to your source stanza and the devel job will go away. It's on by default in the main distro.

peci1:
By REP 143, test_commits should be false by default... Or has this changed?

And for the depends tag: it seems I'd need to add also the dependency to distribution.yaml, am I right? What section is important then? Just the source section, or does it also have to have both source and doc sections? I'll try to contact the maintainer and help him getting his package to the index.

tfoote:
test_commits at that level defaults to false, but it's overridden by a higher level config to turn them on by default https://github.com/ros-infrastructure/ros_buildfarm_config/blob/master/kinetic/source-build.yaml#L52

PR #18268, opened 2018-06-25
Added python-boltons

agdespopoulos:


tfoote (changes requested):
Because this has been packaged upstream we should target that version and possibly use pip for older distros only or explicitly backport it.

Ubuntu since Artful (also found in zesty manually)
https://packages.ubuntu.com/artful/python-boltons
Debian since Stretch
https://packages.debian.org/stretch/python-boltons

agdespopoulos:
Added packages for OS distributions that have it. Let me know if that's better.

tfoote (approved):

mikaelarguedas (approved):

PR #18269, opened 2018-06-25
Added python-networkx-pip

agdespopoulos:
Added `python-networkx-pip` as the OS package manager `python-networkx` packages are all on the 1.x release series, whereas pip is on the 2.x series, which came out last September.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1964,6 +1964,25 @@ python-networkx:
     yakkety_python3: [python3-networkx]
     zesty: [python-networkx]
     zesty_python3: [python3-networkx]
+python-networkx-pip:
+  arch:
+    pip:
+      packages: [networkx]
+  debian:
+    pip:
+      packages: [networkx]
+  fedora:
+    pip:
+      packages: [networkx]
+  gentoo:
+    pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
As stated in our contributions guidelines we strongly prefer the system packaged versions. https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

This is a long standing package around since at least wheezy and trusty
https://packages.debian.org/wheezy/python-networkx
https://packages.ubuntu.com/trusty/python-networkx

And we already have a rule for them.

If we put this in, it will break any package that depends on the system installed version.

And as of buster debian has moved to the 2.x branch

Since upstream is actively moving forward with the new version I think we should stick with the system packaged version for our official rosdep index.

If you need newer rules for your project and want to backport the packages or use the pip rules I'd recommend making your own rosdep rules that you can use locally without potentially disrupting all the other users of the distro.

agdespopoulos:
> If we put this in, it will break any package that depends on the system installed version.

For some reason I had though just having the rosdep rule be a different name would avoid that, but that's clearly not the case, you're right. I'll look at other solutions. Thanks.

PR #18282, opened 2018-06-27
pybind11_catkin: 2.2.3-1 in 'indigo/distribution.yaml' [bloom]

wxmerkt:
Increasing version of package(s) in repository `pybind11_catkin` to `2.2.3-1`:

- upstream repository: https://github.com/ipab-slmc/pybind11_catkin.git
- release repository: https://github.com/wxmerkt/pybind11_catkin-release.git
- bloom version: `0.6.5`

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -10275,6 +10275,12 @@ repositories:
       url: https://github.com/clearpathrobotics/puma_motor_driver.git
       version: master
     status: maintained
+  pybind11_catkin:
+    release:
+      tags:
+        release: release/indigo/{package}/{version}
+      url: https://github.com/wxmerkt/pybind11_catkin-release.git
+      version: 2.2.3-1
   pyros:
     doc:
       type: git


tfoote:
Rebuilding after #18283

Note that this can go into indigo and kinetic but not melodic as pybind11 is available as of artful: https://packages.ubuntu.com/artful/pybind11-dev

tfoote:
Sorry apparently this needs a rebase on the above changes to pass CI. I thought it would test post merge.

wxmerkt:
Thanks @tfoote about the note regarding the release in Ubuntu Artful and later. We are adding extra CMake macros on top of the external project (cf. [here](https://github.com/ipab-slmc/pybind11_catkin/blob/master/cmake/pybind.cmake.in#L18-L28)) so may consider releasing just those without the source checkout for melodic if that's ok?

I also rebased on master.

tfoote:
If you're just adding cmake_macros that would be reasonable. You could also consider submitting them to https://github.com/ros/cmake_modules

PR #18283, opened 2018-06-27
Pin version of PyYAML to fix rosdistro checks.

nuclearsandwich:
[PyYAML 4.1](https://pypi.org/project/PyYAML/4.1/) was released today and it looks like there have been some API changes.
I haven't taken the time to investigate them but pinning the last 3.x version will at least let CI continue for now.


wjwwood (approved):
lgtm as a temporary work around

PR #18291, opened 2018-06-27
Add python-parse, python-construct

ejalaa12:
- added parse package for parsing strings using  a specification based on the Python format() syntax (https://pypi.org/project/parse/)
- added the construct library for parsing and building binary data (https://pypi.org/project/construct/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -763,12 +763,7 @@ python-concurrent.futures:
   gentoo: [virtual/python-futures]
   ubuntu: [python-concurrent.futures]
 python-construct:
-  debian:
-    pip:
-      packages: [construct]
-  ubuntu:
-    pip:
-      packages: [construct]
+  ubuntu: [python-construct]
 python-control-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@ejalaa12 please use `-pip` suffix for packages only available via pip.
In this particular case, `python-construct` exists in ubuntu since [at least trusty](https://packages.ubuntu.com/trusty/python-construct), so we strongly recommend using the apt package rather than the pip version.

tfoote (changes requested):
Also python-parse is available from upstream as well: https://packages.ubuntu.com/trusty/python-parse

Following our guidelines please update this to use the system packages. 
https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

tfoote:
Thanks for the update, just one more request. Please add the debian rule too, support for both is our minimum requirement in the contributing guideline.

I've verified they're both available
https://packages.debian.org/stretch/python-construct
https://packages.debian.org/stretch/python-parse

ejalaa12:
Sure! I'm adding them.

tfoote (approved):
Thanks for iterating!

PR #18302, opened 2018-06-28
rosdep: Added ruby-dev as a separate key

meyerj:
For now it's a copy of key `ruby`, without the specializations for Ubuntu EOL distros.
In the future we would like to split ruby dependencies in run-time and build-time.

Furthermore some Orocos packages (e.g. [typelib](https://github.com/orocos-toolchain/typelib), see also https://github.com/orocos-toolchain/typelib/pull/109) would require the `ruby-dev` key for compatibility with the [AutoProj](https://www.rock-robotics.org/documentation/autoproj/index.html) [package set](https://github.com/rock-core/package_set) used by the Rock community. The `manifest.xml` file was originally supposed to be compatible to ROS/rosdep and AutoProj.

### DIFF
--- a/rosdep/ruby.yaml
+++ b/rosdep/ruby.yaml
@@ -84,11 +84,11 @@ ruby:
     zesty: [ruby, ruby-dev]
 ruby-dev:
   arch: [ruby]
-  debian: [ruby, ruby-dev]
-  fedora: [ruby, ruby-devel, openssl-devel, rubygems]
+  debian: [ruby-dev]
+  fedora: [ruby-devel, openssl-devel, rubygems]
   gentoo: [dev-lang/ruby]
   macports: [ruby]
-  ubuntu: [ruby, ruby-dev]
+  ubuntu: [ruby-dev]
 ruby-sass:
   debian: [ruby-sass]
   fedora: [rubygem-sass]
--- a/rosdep/ruby.yaml
+++ b/rosdep/ruby.yaml
@@ -82,6 +82,13 @@ ruby:
     xenial: [ruby, ruby-dev]
     yakkety: [ruby, ruby-dev]
     zesty: [ruby, ruby-dev]
+ruby-dev:
+  arch: [ruby]
+  debian: [ruby, ruby-dev]
+  fedora: [ruby, ruby-devel, openssl-devel, rubygems]
+  gentoo: [dev-lang/ruby]
+  macports: [ruby]
+  ubuntu: [ruby, ruby-dev]
 ruby-sass:
   debian: [ruby-sass]
   fedora: [rubygem-sass]


rosdep/ruby.yaml
```diff
@@ -82,6 +82,13 @@ ruby:
     xenial: [ruby, ruby-dev]
     yakkety: [ruby, ruby-dev]
     zesty: [ruby, ruby-dev]
+ruby-dev:
+  arch: [ruby]
+  debian: [ruby, ruby-dev]
+  fedora: [ruby, ruby-devel, openssl-devel, rubygems]
+  gentoo: [dev-lang/ruby]
+  macports: [ruby]
+  ubuntu: [ruby, ruby-dev]
```
tfoote: If this is to be a future only dev version should we not include the full ruby dependency here?

tfoote (commented):
It would make sense to split this out if the dependencies are noteably different.

We will need to have a fully planned out migration since there are likely a lot of packages currently relying on the build and runtime dependencies provided together.

meyerj:
> We will need to have a fully planned out migration since there are likely a lot of packages currently relying on the build and runtime dependencies provided together.

I did not change the `ruby` key yet and only added a new key `ruby-dev`, so no risk to break existing packages, no?

With `ruby-dev` defined packages can start to declare a build-dependency to `ruby-dev` and a run-time dependency to `ruby`. In a later step the development packages could be removed from `ruby`.

> If this is to be a future only dev version should we not include the full ruby dependency here?

If I understand correctly you are asking to remove the full ruby dependency and only depend on `ruby-dev`? That's indeed possible and would work for `typelib`. At least for Ubuntu `ruby-dev` indirectly depends on `ruby` anyway and not only on `libruby2.x`, but that might have been different in the past.

tfoote:
>> We will need to have a fully planned out migration since there are likely a lot of packages currently relying on the build and runtime dependencies provided together.
> I did not change the ruby key yet and only added a new key ruby-dev, so no risk to break existing packages, no?
> With ruby-dev defined packages can start to declare a build-dependency to ruby-dev and a run-time dependency to ruby. In a later step the development packages could be removed from ruby.

Yes, this won't break anything in itself, but I'd like to have a general plan for how to make the future migration. If we don't plan it out we can back ourselves into a corner.

I think in this case since artful and bionic are not yet defined and artful is going EOL this month. We could take the opportunity to switch that platform over.  However, Bionic is closed to stretch so we'd likely want to switch it over too. However stretch is already defined and being used for both Lunar and Melodic. 

Luckily as we're strictly removing dependencies we could set the policy that on buster and newer `ruby` will resolve to runtime only. And for older platforms it will resolve to both. And for Melodic and newer if you want the development files use the rosdep `ruby-dev`. 

Packages on stretch that depend on `ruby` will get `ruby-dev` included by default too.

So I'd suggest that in conjunction with adding the currently proposed `ruby-dev` We add for `ubuntu` `'*': [ruby]` that will effectively target all newer platforms than currently defined zesty.

And for Debian explicitly call out `[ruby, ruby-dev]` for jessie and stretch for backwards compatibility and user `'*': [ruby-dev]` to catch all newer platforms starting with buster.

This should allow all users to make the transition with the simple requirement that they make sure to use  `ruby-dev` key as of Melodic. (It's defined back onto older platforms so it can be used in older releases and not require forking development.)

mikaelarguedas (approved):
lgtm

PR #18308, opened 2018-06-29
Move some release repos to ros-gbp

mikeferguson:
Not sure why I originally created these in fetchrobotics-gbp, since the source repos are personal repos.


mikaelarguedas (approved):
lgtm, note that the tracks will need to be updated to reflect the new release repository URL

mikeferguson:
Good point on the tracks, thanks for mentioning that!

PR #18342, opened 2018-07-03
Bloom fluent logger ros 0

akio:
Second attempt to release `fluent_logger_ros`.
----

Increasing version of package(s) in repository `fluent_logger_ros` to `0.1.0-0`:

- upstream repository: https://github.com/akio/fluent_logger_ros.git
- release repository: https://github.com/akio/ros_fluent_logger-release.git
- bloom version: `0.6.1`

## fluent_logger_ros

```
* Initial release
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2633,7 +2633,7 @@ repositories:
     source:
       type: git
       url: https://github.com/akio/fluent_logger_ros.git
-      version: 0.1.0
+      version: kinetic-devel 
     status: developed
   follow_waypoints:
     doc:
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2624,6 +2624,17 @@ repositories:
       url: https://github.com/ros-drivers/flir_ptu.git
       version: master
     status: maintained
+  fluent_logger_ros:
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/akio/ros_fluent_logger-release.git
+      version: 0.1.0-0
+    source:
+      type: git
+      url: https://github.com/akio/fluent_logger_ros.git
+      version: 0.1.0
+    status: developed
   follow_waypoints:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


tfoote:
This looks reasonable. However, I'd ask that you rename your package to not have `_ros` in the name as that's redundant as it's a ROS package.  Please see the naming conventions: http://wiki.ros.org/ROS/Patterns/Conventions

akio:
The convention recommends to avoid using prefix `ros` but it doesn't mention the suffix `_ros`.

In my understanding, the suffix `_ros` is used in order to indicate that the package provides ROS interfaces of a existing library.
As far as I know there are many examples of this convention from small packages to popular packages:

| Library | ROS interfaces/utilities|
|----------|-------------------|
|cartographer|cartographer_ros|
|gazebo|gazebo_ros|
|smach|smach_ros|
|moveit|moveit_ros|
|stage|stage_ros|
|ypspur|ypspur_ros|
|apriltag2|apriltag2_ros|
|aruco|aruco_ros|
|hiqp|hiqp_ros|
|imagezero|imagezero_ros|
|julius|julius_ros|
|octomap|octomap_ros|
|pcl|pcl_ros|
|py_trees|py_trees_ros|
|rtabmap|rtabmap_ros|
|towr|towr_ros|
|visp|visp_ros|

Since there is the original `fluent_logger` library and this package provides ROS interfaces/utilities of that, I would like to follow the convention that the many forerunners used.

What are your thoughts on my idea? (Especially the convention that other libraries follow)

tfoote:
A large fraction of those packages actually have two versions. Both the ROS wrapper package as well as the pure library which is why they differentiate. For example: http://wiki.ros.org/rtabmap as the core library as a ROS package and rtabmap_ros has the ROS interface for it. This is true for visp, tf2, towr, py_trees, (pcl for used to be a ROS package, it's kept for legacy reasons), octomap, julius, smach, stage, gazebo and cartographer. 

If you want to split out the core library and the ROS wrapper that would make sense, but if you're only creating one package you shouldn't use the suffix. It's only recommended to be used if there needs to be differentiation.

PR #18354, opened 2018-07-03
Define ruby for Ubuntu 18.04

lnnx2006:


nuclearsandwich:
@lnnx2006 this needs a fix for alphabetic ordering.

There's an ongoing discussion about the introduction of a separate ruby-dev key in https://github.com/ros/rosdistro/pull/18302 but I believe the intention there is not to change the existing definitions for the ruby key.

tfoote:
Since this has a high overlap with #18302 wrt to adding `ruby-dev` I'm going to close this as a duplicate. Adding artful and bionic rules in a separate PR could be merged even as that discussion is pending though.

PR #18356, opened 2018-07-03
remove pybind11 from indigo

tfoote:
Removing the version will effectively remove it

Reverts #18282 due to not building https://github.com/ipab-slmc/pybind11_catkin/issues/1

@wxmerkt FYI


wxmerkt:
Ok, will submit again once building. I ran pre-release tests using the ros-industrial docker ci image and it passed there. Apologies for the errors - didn't get to resolving them right now and will get back to this soon.

PR #18371, opened 2018-07-05
Adding choreo to ducumentation index for distro.

yijiangh:
I'd like choreo to be indexed and documented on ros.org.


mikaelarguedas (changes requested):
CI is failing because the url are missing the extension `.git` at the end,

You can see it from the [travis job log](https://travis-ci.org/ros/rosdistro/builds/400576391?utm_source=github_status&utm_medium=notification): 
```
Repository 'choreo' with url 'https://github.com/yijiangh/Choreo' should end with `.git` but does not.
```

PR #18374, opened 2018-07-05
Python 3 compatible syntax

dirk-thomas:
Redo of #17110 since the original author closed the PR.

PR #18378, opened 2018-07-06
update url validity test for Python 3

dirk-thomas:
Follow up of #18374.


dirk-thomas:
I will go ahead and merge this so that upcoming PRs are passing the Python 3 CI.

PR #18392, opened 2018-07-09
Update python.yaml

reinzor:
Added rosbag_pandas via pip

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2999,6 +2999,10 @@ python-responses-pip:
   ubuntu:
     pip:
       packages: [responses]
+python-rosbag-pandas-pip:
+  ubuntu:
+    pip:
+      packages: [rosbag_pandas]
 python-rosdep:
   arch: [python2-rosdep]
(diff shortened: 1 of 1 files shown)


mikaelarguedas (changes requested):
releasing the package only via pip will prevent people from releasing packages depending on it.
Any reason for not releasing it as a ROS package ?

We require at least debian and ubuntu rules to accept a new key. So if the pip route is the final one Debian needs to be added as well

tfoote:
This should definitely be a ROS package. It's importing `rospy` and `rosbag` which means that it can't satisfy it's dependencies via pip and is potentially rosdistro specific.

I'd suggest talking @aktaylor08 about releasing this as a ROS package? If he's not interested maybe you could help.

Rayman:
Added a PR to convert this package to ROS here: https://github.com/aktaylor08/RosbagPandas/pull/9

tfoote:
Since that PR was accepted and it can be released as a ROS package now I'm going to close this.

PR #18430, opened 2018-07-10
Re-add pybind11_catkin to indigo (Revert #18356)

wxmerkt:
The build issues documented in https://github.com/ipab-slmc/pybind11_catkin/issues/1 should now be fixed.

This reverts commit 71d6005f8d46ba0b71919b79614b727e592d705c.


tfoote:
That's the same version, how was the build issue resolved?

wxmerkt:
I updated the CMake logic to distinguish between catkin devel and install spaces (which works for catkin tools workflows) and force-updated the tag in the source repository (hence the same version). Do I need to run bloom again if it's the same tag?

PR #18440, opened 2018-07-11
Simplify python-docopt rule

tfoote:
It's available on all supported ubuntu platforms: https://packages.ubuntu.com/trusty/python-docopt
And on all debian too: https://packages.debian.org/jessie/python-docopt

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -970,14 +970,10 @@ python-dlib:
   ubuntu:
     pip: [dlib]
 python-docopt:
-  debian:
-    buster: [python-docopt]
-    jessie: [python-docopt]
-    stretch: [python-docopt]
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
This means that this rule will stop resolving for precise and other distros, is this on purpose? If yes, PR LGTM

tfoote:
That's not intended, I only checked the currently supported distros. If someone's actively using it on an EOL distro I'd be fine adding back the pip rule using the wildcard approach. But since those are EOL I'd suggest we wait for there to be a demand for supporting the EOL distros before we use the more complex markup.

mikaelarguedas (approved):
lgtm, based on last comment about rules being droped for some EOL distros

PR #18474, opened 2018-07-13
Add minimal alpine rosdeps for ros_core

at-wat:
With https://github.com/ros-infrastructure/rospkg/pull/148 and https://github.com/ros-infrastructure/rosdep/pull/616,
all packages required by ros_core can be installed by rosdep on Alpine Linux (edge version)

Tested with https://github.com/at-wat/alpine-ros/tree/master/melodic-ros-core-alpine

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2992,7 +2992,7 @@ libssh2-dev:
   gentoo: [net-libs/libssh2]
   ubuntu: [libssh2-1-dev]
 libssl-dev:
-  alpine: [openssl1.0-dev]
+  alpine: [libressl-dev]
   arch: [openssl]
   debian: [libssl-dev]
   fedora: [openssl-devel]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -992,7 +992,7 @@ golang-go:
 google-mock:
   alpine:
     source:
-      uri: 'https://github.com/at-wat/alpine-ros-sources/raw/master/googletest/googletest.rdmanifest'
+      uri: 'https://github.com/at-wat/alpine-ros-sources/raw/master/googlemock/googlemock.rdmanifest'
   arch: [gmock]
   debian: [google-mock]
   fedora: [gmock-devel]
@@ -1138,9 +1138,7 @@ gstreamer1.0-x:
   gentoo: ['media-libs/gst-plugins-base:1.0[X,pango]']
   ubuntu: [gstreamer1.0-x]
 gtest:
-  alpine:
-    source:
-      uri: 'https://github.com/at-wat/alpine-ros-sources/raw/master/googletest/googletest.rdmanifest'
+  alpine: [gtest-dev, gtest]
   arch: [gtest]
   debian: [libgtest-dev]
   fedora: [gtest-devel]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -54,6 +54,7 @@ apache2-mpm-prefork:
   gentoo: ['www-servers/apache[apache2_mpms_prefork]']
   ubuntu: [apache2-mpm-prefork]
 apr:
+  alpine: [apr, apr-util]
   arch: [apr, apr-util]
   cygwin: [libapr1, libaprutil1]
   debian: [libapr1-dev, libaprutil1-dev]
@@ -266,6 +267,7 @@ bluez-hcidump:
     wily: [bluez-hcidump]
     xenial: [bluez-hcidump]
 boost:
+  alpine: [boost-dev]
   arch: [boost]
   cygwin: [libboost-devel, libboost1.40]
   debian:
@@ -330,6 +332,7 @@ bullet:
     yakkety: [libbullet-dev]
     zesty: [libbullet-dev]
 bzip2:
+  alpine: [bzip2-dev]
   arch: [bzip2]
   cygwin: [bzip2]
   debian: [libbz2-dev]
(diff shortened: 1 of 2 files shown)


rosdep/base.yaml
```diff
@@ -986,6 +990,9 @@ golang-go:
   gentoo: [dev-lang/go]
   ubuntu: [golang-go]
 google-mock:
+  alpine:
+    source:
+      uri: 'https://github.com/at-wat/alpine-ros-sources/raw/master/googletest/googletest.rdmanifest'
```
tfoote: I see https://pkgs.alpinelinux.org/package/edge/main/x86/gtest-dev which looks like it should resolve the dependency.

rosdep/base.yaml
```diff
@@ -1531,6 +1541,9 @@ libconfig-dev:
   gentoo: ['dev-libs/libconfig[cxx]']
   ubuntu: [libconfig-dev]
 libconsole-bridge-dev:
+  alpine:
+    source:
+      uri: 'https://github.com/at-wat/alpine-ros-sources/raw/master/console-bridge/console-bridge.rdmanifest'
```
tfoote: Related, it we can probably go ahead and merge this as is if it's working. But for greater maintainability and better use experience it would be great to submit this upstream for packing.

rosdep/base.yaml
```diff
@@ -3543,6 +3559,9 @@ lm-sensors:
   gentoo: [sys-apps/lm_sensors]
   ubuntu: [lm-sensors]
 log4cxx:
+  alpine:
+    source:
+      uri: 'https://github.com/at-wat/alpine-ros-sources/raw/master/log4cxx/log4cxx.rdmanifest'
```
tfoote: We've found that source rosdeps generally end up being more work than their worth. I thought we'd officially deprecated them but I can't find a reference. But either way they're strongly discouraged. If you've tested them as working we can merge them for now until a better solution is available.

at-wat:
Currently, released versions (non 'edge' version) of alpine repository lack a lot of necessary packages.

e.g. poco-dev was registered at April 2017 and only available at edge/testing repository at now. I have to understand package life cycle in Alpine well.
(openssl1.0-dev, sbcl and py-coverage are also only in edge.)

So, I think I should define package names for each alpine release versions.

rosdep/base.yaml
at-wat: This gtest-dev package doesn't contain gmock, and gmock is not provided in alpine package repository.

Since [gmock requires to be built with gtest source](https://github.com/google/googletest/blob/b4d4438df9479675a632b2f11125e57133822ece/googlemock/CMakeLists.txt#L15-L19), I think it's not easy to coexist source built gmock with binary installed gtest.
(Patching CMakeLists.txt of gmock might enable to build it using binary installed gtest. I would like to try it.)

at-wat (commented):
I commented package availability information of the packages which is not provided in Alpine v3.5 (oldest currently supported release).

rosdep/base.yaml
at-wat: I agree that using source rosdeps lacks maintainability pretty much.
Registering them to the upstream package repository would be necessary. But until completing them, keeping source rosdeps would be helpful for testing.

at-wat:
> So, I think I should define package names for each alpine release versions.

Since `FdoDetect` in rospkg uses full version including patch version, to specify package name for each Alpine release, `rosdep/base.yaml` should be like:
```yaml
libssl-dev:
  alpine:
    3.8.0: [openssl1.0-dev]  # 'edge' has same os_version with latest release.
                             # **it works on edge but not on 3.8**
    3.7.0: [openssl-dev]
    3.6.2: [openssl-dev]
    3.6.1: [openssl-dev]
    3.6.0: [openssl-dev]
    ...
```
I guess there is no nice way to support non-edge Alpine version at this moment. (any idea?)
I would like to keep os_version not specified for alpine.

rosdep/base.yaml
at-wat: @tfoote I changed it to use alpine package version of 'gtest' and source rosdep version of 'google-mock'.
(It would be better to register google-mock to upstream apk repository in near future.)

rosdep/base.yaml
```diff
@@ -2979,6 +2994,7 @@ libssh2-dev:
   gentoo: [net-libs/libssh2]
   ubuntu: [libssh2-1-dev]
 libssl-dev:
+  alpine: [openssl1.0-dev]
```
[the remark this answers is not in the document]
at-wat: Alpine looks recommending to use libressl instead of openssl (https://bugs.alpinelinux.org/issues/4970).
Using openssl may cause conflict with libressl.

at-wat:
@tfoote I think it is ready as a starting point for discussing and testing ROS on Alpine Linux.
Please take another look.

tfoote (approved):
To get started we could only support the current edge and newer for testing. And once edge becomes a release it could be the low water mark. I'm assuming that coverage will grow.

Note that there are things like `'*'` syntax you can use to make different version simpler, but in general I'd suggest the low water mark will be a better approach in the long term until things change above that point there's no need to define support for older versions that are known not supported.

I'd also encourage you to push hard on getting upstream to package the things that are currently installed from source too. The source rosdeps will make things much more painful in the long run.

at-wat:
@tfoote Thanks for your comments. I'm learning the packaging system of Alpine. I'll start working to register them to the upstream. Also, I'd like to start a discussion at the discourse toward enabling all ROS features on Alpine.

I have found one mistake that `python` must be `python2-dev`, not `python2`.
I'll quickly fix it.

tfoote (approved):

PR #17464, opened 2018-04-16
Revert "python-flask-restful: 0.3.4-4 in 'indigo/distribution.yaml' [bloom]"

tfoote:
Reverts ros/rosdistro#17461

PR #18483, opened 2018-07-14
added libmongoclient-dev 

bbferka:
C++ mongo client  legacy version added to base.yaml for Ubuntu bionic; As far as I am aware this is new  package in bionic, not available in earlier distros. This package is not to be conuised with libmongo-client-dev which is an unofficial C driver;

https://packages.ubuntu.com/bionic/libmongoclient-dev

Currently this is wrapped and released as libmongocxx_ros for kinetic and indigo through the [mongodb_store](https://github.com/strands-project/mongodb_store.git) stack. This can now be replaced by a system dependency. 

We are also planning on releasing our [perception framework](https://github.com/RoboSherlock/robosherlock) later this year through bloom which will depend on this.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2188,8 +2188,12 @@ libmodbus-dev:
   fedora: [libmodbus-devel]
   ubuntu: [libmodbus-dev]
 libmongoclient-dev:
+  debian: [libmongoclient-dev]
+  fedora: [mongo-cxx-driver]
   ubuntu:
+    artful: [libmongoclient-dev]
     bionic: [libmongoclient-dev]
+    cosmid: [libmongoclient-dev]
 libmotif-dev:
   arch: [lesstif]
   debian: [libmotif-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2190,10 +2190,7 @@ libmodbus-dev:
 libmongoclient-dev:
   debian: [libmongoclient-dev]
   fedora: [mongo-cxx-driver]
-  ubuntu:
-    artful: [libmongoclient-dev]
-    bionic: [libmongoclient-dev]
-    cosmid: [libmongoclient-dev]
+  ubuntu: [libmongoclient-dev]
 libmotif-dev:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Please add at least artful and cosmic. At some level I'd like to actually just add the generic rule not calling out the specific version as it's not going to break anyone on an older platform and will be future compatible. 

Following the contributing guide please add debian and fedora rules if you're planning to make a release in the future. 
I see https://packages.debian.org/stretch/libmongoclient-dev

bbferka:
Done. 

arful: https://packages.ubuntu.com/artful/libmongoclient-dev
cosmic: https://packages.ubuntu.com/cosmic/libmongoclient-dev

debian: https://packages.debian.org/stretch/libmongoclient-dev
fedora: https://apps.fedoraproject.org/packages/mongo-cxx-driver

Should I merge the ubuntu packages into as single entry then?

tfoote:
If you could make the ubuntu ones use the simpler rule that would be great. There's already a typo in the cosmic line.

tfoote (approved):
We don't keep an authoritative set of versions, we just syntax check them to make sure that the formatting is right.

mikaelarguedas:
@tfoote while this is released as a ros package on indigo and kinetic. Won't we have conflicts between rosdep keys and ros package if the package has a match in both?

I think we faced this in the past:
https://github.com/ros/rosdistro/pull/17461
https://github.com/ros/rosdistro/pull/17464

So my understanding is that the ubuntu rule added here should be modified to not match trusty or xenial

dirk-thomas:
> while this is released as a ros package on indigo and kinetic. Won't we have conflicts between rosdep keys and ros package if the package has a match in both?

@mikaelarguedas  Can you where you see the conflict? The ROS distros don't contain a package with `libmongoclient` in the name.

dirk-thomas (approved):

dirk-thomas:
I will go ahead and merge this as is.

@mikaelarguedas Please feel free to comment after the fact regarding the above question.

PR #18494, opened 2018-07-16
remove ira_photonfocus_driver Indigo

pietrocolombo:
remove ira_photonfocus_driver for compiling error


tfoote:
Are you removing this due to getting emails from the build farm? If so you can turn them off by adding

`test_commits: false`

More info: 
https://answers.ros.org/question/236406/default-bool-value-for-test_commits-and-test_pull_requests-in-rosdistro-format/

http://www.ros.org/reps/rep-0143.html#distribution-file

PR #18538, opened 2018-07-18
Adding python-imaging-imagetk

Minipada:
Here is a link to ubuntu: packages:
https://packages.ubuntu.com/trusty/python-pil.imagetk

The Imagetk module provides functions to convert PIL images to the format that is need to display the images in a GUI with tkinter. We use it to display dynamic images in a simple GUI that we use to control our robot.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1600,7 +1600,8 @@ python-imaging:
     zesty: [python-imaging]
     zesty_python3: [python3-imaging]
 python-imaging-imagetk:
-    ubuntu: [python-pil.imagetk]
+  debian: [python-pil.imagetk]
+  ubuntu: [python-pil.imagetk]
 python-impacket:
   debian: [python-impacket]
   fedora: [python-impacket]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1599,6 +1599,13 @@ python-imaging:
     yakkety_python3: [python3-imaging]
     zesty: [python-imaging]
     zesty_python3: [python3-imaging]
+python-imaging-imagetk:
+  ubuntu:
+      artful: [python-pil.imagetk]
```
tfoote: This indentation is incorrect. And please use the generic `ubuntu: [python-pil.imagetk]` since all versions are the same.

tfoote (changes requested):
Following the Contributing guideline: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

Please also add the debian rules it's also available on all targets: https://packages.debian.org/stretch/python-pil.imagetk

If you have a moment to look for Fedora or Gentoo that would be great too.

tfoote:
It should only be 2 indents per layer. Also please add the debian rule too.

tfoote (approved):

tfoote:
Thanks for iterating! lgtm

Minipada:
Sorry for the errors ... I don't know where my mind was ...

PR #18543, opened 2018-07-19
Adding spinnaker-ros to documentation index for kinetic

ghost:


tfoote:
Also looking at the name of the project it doesn't meet our naming requirements. Please don't use `_ros` in the ROS package name it's redundant. 

http://www.ros.org/reps/rep-0144.html

tfoote:
@shahvi Thanks for the updated naming.

shahvi:
@tfoote I see that the package information has been updated, but I don't understand why the documentation at http://wiki.ros.org/spinnaker_camera_driver shows incorrect package summary.  It is pointing to the flir_camera_driver instead.

Also, the Jenkins build has failed as I had expected.  Its because these drivers use external Spinnaker SDK package from Pt Grey.  These packages are not available in a repo and have to be downloaded and installed as .deb files.  How do I configure the build job so that it builds correctly with an external dependency?

mhosmar-cpr:
@tfoote @shahvi This is because I released the meta-package flir_camera_driver with a standalone package which serves as a spinnaker camera driver spinnaker_camera_driver a few days ago. The link points to that repo because the generated docs were already created.

tfoote:
Well it looks like multiple people want the same thing. 

@shahvi It looks like that package name is already taken. As it looks like you're duplicating work, might I suggest that you look at what's already there and see if you can contribute any functionality you have into the already existing driver instead of creating a new one?

shahvi:
@tfoote our code is different in that it uses the SDK, but it is optimized for multiple cameras and hardware triggering etc. So I'd prefer to keep it separate. Is there a convention to use here?  Would spinnaker_multicam_driver be reasonable?

tfoote:
That would work or spinnaker_sdk_camera_driver if it's specific to the sdk. Keeping `camera_driver` as a substring I think would be recommended.

Note that the other driver says they'd like to switch to use the sdk so there might be a good opportunity to coverge in the long run.

shahvi:
@tfoote  So that brings me to my earlier question, how do I setup the build job?  This package depends on an external Spinnaker SDK package from Pt Grey. The SDK packages are not available in a repo and have to be downloaded and installed as .deb files. How do I configure the build job so that it builds correctly with an external dependency?

tfoote:
You need it to be able to build in your workspace. Sometimes you can download and install resources in the CMakeLists but a deb package will be a problem as you need to keep all operations in your workspace.

PR #18591, opened 2018-07-21
add ntpdate for fedora

fmessmer:
seems available for fedora, too: https://apps.fedoraproject.org/packages/ntpdate


fmessmer:
I added `ntp` for arch (https://www.archlinux.org/packages/extra/x86_64/ntp/) to both rosdep keys `ntp` and `ntpdate` (as done for gentoo)...let me know whether this is ok...

mikaelarguedas (approved):
lgtm

PR #18635, opened 2018-07-23
Adding teleop_twist_keyboard_cpp

methylDragon:
Wrote a C++ implementation of the teleop_twist_keyboard_node.

It's very useful for integrating with other protocols and frameworks that use C++.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -12077,6 +12077,16 @@ repositories:
       url: https://github.com/ros-teleop/teleop_twist_keyboard.git
       version: master
     status: maintained
+  teleop_twist_keyboard_cpp:
+    doc:
+      type: git
+      url: https://github.com/methylDragon/teleop_twist_keyboard_cpp.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/methylDragon/teleop_twist_keyboard_cpp.git
+      version: master
+    status: maintained
   tensorflow_ros_cpp:
     doc:
       type: git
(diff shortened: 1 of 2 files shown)


tfoote:
Thanks for submitting this. Would you consider submitting it to be in the teleop_twist_keyboard repository next to the telop_twist_keyboard python implementation. The repo could be reorganized to have them be side by side packages to keep the dependencies minimal.

methylDragon:
Alrighty! So I send in a pull request to the existing teleop repo, then update the rosdistro .yamls?

tfoote:
If it gets accepted there, in the next release we should remove this entry as it will be indexed over there.

methylDragon:
Ok, I'm a little unsure about how exactly I should go about including it in the original repo though, as I'm unsure about the best practices with regards to this. (Sorry for the trouble!)

Should I
a) Add it into a subdirectory in the repo as its own package (teleop_twist_keyboard_cpp)
b) Add it into a subdirectory in the repo with the same package name (teleop_twist_keyboard) on the package.xml
c) Reorganise the repo, and package it into two folders (python) and (cpp) both with the package.xmls denoting that the package name is teleop_twist_keyboard
or d) Edit the existing CMakeLists and package.xml on the original repo so that my code will be an executable under the teleop_twist_keyboard package, executable alongside the python implementation?

tfoote:
a and b won't work due to packages not nesting.

Since d will add dependencies on roscpp and rospy it won't be separable. So I'd suggest you propose c in a PR, but the packages shouldn't have the same name. I'd suggest keeping your current name so that they can coexist side by side on a system and the deployer can choose to use one or both.

PR #18662, opened 2018-07-27
Renaming spinnaker_camera_driver to spinnaker_sdk_camera_driver as per #18543

ghost:
as per #18543

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -344,21 +344,6 @@ repositories:
       url: https://github.com/RIVeR-Lab/apriltags_ros.git
       version: indigo-devel
     status: maintained
-  ar_sys:
-    doc:
-      type: git
-      url: https://github.com/Sahloul/ar_sys.git
-      version: indigo-devel
-    release:
-      tags:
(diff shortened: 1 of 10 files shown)


tfoote:
@shahvi The updated urls look ok. but there's extra whitespace that is causing the parsing to fail.

From your comments and those of @mhosmar-cpr  in #18543 it seems like their driver might be a little bit more mature as it already has the SDK download and continuous integration setup and passing as well as what I'm guessing is active use and testing at Clearpath. I might suggest that collaborating on that driver would be more effective than working to duplicate the effort already expended getting the sensor and release integration all working.

shahvi:
@tfoote fixed the extra space.  I believe that our package can do multi-camera reasonably well.  I'd like to leave the package out there for now...

tfoote:
Ok, but please keep in mind that the other camera driver appears to be more mature including automated downloads and unit test with operational continuous integration. As well as integration as a nodelet. It looks like you have support for more parameters but I'd highly encourage you to look at how to converge the drivers so we don't get fragmentation in the community.

PR #18663, opened 2018-07-27
Remove ar_sys

tfoote:
That users's account appears to have been deactivated. None of the repositories are reachable and the user's account homepage is also a 404


wjwwood (approved):

tfoote:
@bmagyar FYI you're listed as a maintainer of this package that's now unavailable. http://wiki.ros.org/ar_sys 

I've directly emailed the other listed maintainers too.

bmagyar:
Thanks Tully. I definitely wasn't aware of this package, definitely not maintaining it.

PR #18686, opened 2018-07-31
Add python3 opencv dependency for ubuntu

gaoethan:
Ubuntu: https://packages.ubuntu.com/bionic/python3-opencv

Now only add for ubuntu firstly to use, because it's already on the way for the other Linux flavors while they're finally ready and stable, e.g Debian: https://packages.debian.org/search?keywords=python3-opencv

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4441,10 +4441,8 @@ python3-numpy:
 python3-opencv:
   debian:
     buster: [python3-opencv]
-    stretch: null
   ubuntu:
     bionic: [python3-opencv]
-    xenial: null
 python3-pep8:
   debian:
     buster: [python3-pep8]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4443,9 +4443,7 @@ python3-opencv:
     buster: [python3-opencv]
   ubuntu:
     bionic: [python3-opencv]
-    xenial:
-      pip:
-        packages: [opencv-python]
+    xenial: null
 python3-pep8:
   debian:
     buster: [python3-pep8]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4439,7 +4439,13 @@ python3-numpy:
   gentoo: [dev-python/numpy]
   ubuntu: [python3-numpy]
 python3-opencv:
-  ubuntu: [python3-opencv]
+  debian:
+    buster: [python3-opencv]
+  ubuntu:
+    bionic: [python3-opencv]
+    xenial:
+      pip:
+        packages: [opencv-python]
 python3-pep8:
   debian:
     buster: [python3-pep8]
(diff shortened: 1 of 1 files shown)


tfoote:
Please go ahead and add the debian rule. Since it's already in buster and that's no longer a rolling release we can target dependencies in it.

mikaelarguedas:
This package is not available on ubuntu xenial that is targeted by 2 active ROS 2 distributions.
So the rule should be modified to target bionic only or add an explicit `null` entry for the other active ubuntu distributions

gaoethan:
The support to debian buster has already been enabled @tfoote 
For Ubuntu Xenial, even though it has no python3-opencv, it has pip support via `pip3 install opencv-python` and I added its pip enabling instead of `null`, but not sure whether it will invoke **pip3 but not pip** with the following rule @mikaelarguedas  

```
xenial:
      pip:
        packages: [opencv-python]
``` 

thanks !

mikaelarguedas:
> but not sure whether it will invoke pip3 but not pip with the following rule

It will invoke pip and not pip3. Support for pip3 is not available in rosdep at the moment.
It also looks like this package is not working out of the box at the moment because of missing dependencies:
```
# pip3 install opencv-python
# python3
>>> import cv2
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/cv2/__init__.py", line 3, in <module>
    from .cv2 import *
ImportError: libgthread-2.0.so.0: cannot open shared object file: No such file or directory
```

gaoethan:
okay. thanks for clarification, I've already set it to `null` for ubuntu:xenial

mikaelarguedas (approved):
lgtm,

added a `null` rule for debian stretch as well in https://github.com/ros/rosdistro/pull/18686/commits/f0b85671710223c6853a5aebfbabc4bfc157d3ae

mikaelarguedas:
@ros/rosdistro-maintainers can I get a second review on this please?

dirk-thomas:
I don't understand what the `null` entries are for? There is no usage of a wildcard so what is the purpose of the `null` entries compared to them not being present?

mikaelarguedas:
oh right, we added the null entry originally because these were global rules to start with (`ubuntu: [python3-opencv]`) now that distros are listed explicitly the null entries are not necessary anymore

mikaelarguedas:
Updated in https://github.com/ros/rosdistro/pull/18686/commits/c4025f1a15b8a71a70e82e5e9797f9c80cc98862

dirk-thomas (approved):
LGTM

Should Artful / Cosmic be added to?

mikaelarguedas:
As this was a key expected for ROS 2 usage we didnt bother (we added only distros targeted by active ROS2 distributions).

But it doesn't hurt to add them if we think it's better. I don't feel storngly either way as we currently don't have guidelines for python 3 packages

dirk-thomas:
Fine with me. Can merge as is.

PR #18700, opened 2018-08-01
added libopenblas-dev to rosdep

wzli:
https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=libopenblas-dev
https://packages.debian.org/search?suite=all&searchon=names&keywords=libopenblas-dev
https://www.rpmfind.net/linux/rpm2html/search.php?query=openblas-devel&submit=Search+...&system=&arch=

another common linear algebra dependency

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2014,6 +2014,9 @@ libhdf5-dev:
   gentoo: [sci-libs/hdf5]
   ubuntu: [libhdf5-dev]
 libhidapi-dev:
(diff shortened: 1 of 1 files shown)


mikaelarguedas (approved):
lgtm
Didnt find a gentoo equivalent after a quick search there is  the blas library but not the openblas in the sci-libs category

mikaelarguedas:
@ros/rosdistro-maintainers can I get a second review on this please?

PR #18718, opened 2018-08-03
added entry for python-can

daniel-dsouza:
added entry for python-can

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -603,6 +603,13 @@ python-cairo:
     slackpkg:
       packages: [pycairo]
   ubuntu: [python-cairo]
+python-can:
+  debian:
+    pip:
+      packages: [python-can]
+  ubuntu:
+    pip:
+      packages: [python-can]
 python-cantools-pip:
   debian:
     pip:


mikaelarguedas (changes requested):
Thanks @daniel-dsouza for contributing.

I don't know if you've seen the [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions) but we require several information in order to merge new rosdep keys:

- description of the package
- description of the use case
- use upstream packages rather than pip where available
- if no upstream version is available add the suffix `-pip` to the key name
- link to the web listings of packages such as on http://packages.ubuntu.com, http://packages.debian.org, and https://admin.fedoraproject.org/pkgdb/packages/ or if a pip package pypi.python.org.

tfoote:
I see ubuntu and debian packages for this: https://packages.debian.org/stretch/python-can https://packages.ubuntu.com/bionic/python-can

dirk-thomas:
@daniel-dsouza Closing due to no response.

PR #18722, opened 2018-08-03
Add mrl_ipcamera package for documentation and continuous integration indexing

alireza-hosseini:


mikaelarguedas:
what does `mrl` stand for ?
Could this package just be `ipcamera`?
In case its useful, the package naming rules can be found [here](http://www.ros.org/reps/rep-0144.html#package-naming)

alireza-hosseini:
@mikaelarguedas Thanks for your review, 
MRL stands for the "Mechatronic Research Laboratory". I implemented this package when I was a member of this research lab, so I just didn't think of dropping the **mrl** prefix.
I took a look on the ROS existing packages and apparently there is no package called **ipcamera** so I think I can use it, however, shouldn't it be **ipcamera_driver**?

alireza-hosseini:
@mikaelarguedas I'm closing this PR as I created a new one using bloom which has the release tag and its name is renamed to ipcamera_driver:
https://github.com/ros/rosdistro/pull/18742

mikaelarguedas:
:+1:

PR #18751, opened 2018-08-07
Adding hr_msgs to documentation index for kinetic distribution.yaml

wenwei-dev:
I'd like hr_msgs to be indexed on ros.org.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3688,11 +3688,19 @@ repositories:
       version: hydro-devel
     status: end-of-life
   hr_msgs:
+    doc:
+      type: git
+      url: https://github.com/hansonrobotics/hr_msgs.git
+      version: master
     release:
       tags:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -3687,6 +3687,21 @@ repositories:
...
     status: end-of-life
+  hr_msgs:
+    doc:
+      type: git
+      url: https://github.com/hansonrobotics/hr_msgs.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/hansonrobotics/hr_msgs.git
```
tfoote: This is your main repository not a release repository. This whole 'release' section should likely not be there if you're just looking to be documented and indexed. 

The release section is for making releases to be built into binary packages.

wenwei-dev: Now I see. Thank you. Will remove it.

PR #18752, opened 2018-08-07
Adding vulkan loader library and dev files to base.yaml

ethanfowler:
We use the [vulkan loader library](https://github.com/KhronosGroup/Vulkan-Loader) and associated dev files in a VR application. It would make our CI happy if it were in rosdep.

Links to the distro-specific packages added:
[debian](https://packages.debian.org/sid/libvulkan-dev)
[fedora](https://apps.fedoraproject.org/packages/vulkan-devel#)
[gentoo](https://packages.gentoo.org/packages/media-libs/vulkan-loader)
[ubuntu](https://packages.ubuntu.com/xenial/libvulkan-dev)

Many thanks,
Ethan

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3409,9 +3409,7 @@ libvtk-qt:
     yakkety: [libvtk6-qt-dev]
     zesty: [libvtk6-qt-dev]
 libvulkan-dev:
-  debian:
-    buster: [libvulkan-dev]
-    stretch: [libvulkan-dev]
+  debian: [libvulkan-dev]
   fedora: [vulkan-devel]
   gentoo: [media-libs/vulkan-loader]
   ubuntu:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3408,6 +3408,17 @@ libvtk-qt:
     xenial: [libvtk6-qt-dev]
     yakkety: [libvtk6-qt-dev]
     zesty: [libvtk6-qt-dev]
+libvulkan-dev:
+  debian:
+    buster: [libvulkan-dev]
+    stretch: [libvulkan-dev]
+  fedora: [vulkan-devel]
+  gentoo: [media-libs/vulkan-loader]
+  ubuntu:
+    artful: [libvulkan-dev]
+    bionic: [libvulkan-dev]
+    cosmic: [libvulkan-dev]
+    xenial: [libvulkan-dev]
 libwebp-dev:
   debian: [libwebp-dev]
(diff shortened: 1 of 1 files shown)


wjwwood (approved):

rosdep/base.yaml
```diff
@@ -3408,6 +3408,17 @@ libvtk-qt:
     xenial: [libvtk6-qt-dev]
     yakkety: [libvtk6-qt-dev]
     zesty: [libvtk6-qt-dev]
+libvulkan-dev:
+  debian:
```
mikaelarguedas: as this is available on all debian active distros, this can be simplified to: `debian: [libvulkan-dev]`

rosdep/base.yaml
```diff
@@ -3408,6 +3408,17 @@ libvtk-qt:
...
+    buster: [libvulkan-dev]
+    stretch: [libvulkan-dev]
+  fedora: [vulkan-devel]
+  gentoo: [media-libs/vulkan-loader]
+  ubuntu:
```
mikaelarguedas: I think that we require rosdep package availability on all active ubuntu distributions.
This package is not available on ubuntu trusty.
It should either be backported to trusty and imported on the farm or be released as a ROS package in ROS Indigo. @tfoote @wjwwood @dirk-thomas to confirm

wjwwood: Well, as long as the key is enumerated like this, I don't see why we wouldn't allow it. If you try to use this key in a package being released into Indigo it will fail still (as before). I'll let @dirk-thomas or @tfoote reply.

ethanfowler: That was my thinking in exhaustively listing the ubuntu distributions; trusty users will see the same behaviour they saw before this change.

dirk-thomas: The opposite approach would just not require any future maintenance:

```
ubuntu:
  '*': [libvulkan-dev]
  trusty: null
```

I don't mind either or.

rosdep/base.yaml
ethanfowler: Fixed in 56ff3d5.

rosdep/base.yaml
ethanfowler: Changed to the wildcard/exclusion approach in ef1113a.

Hoping this can be squashed before merging to avoid polluting your history.

dirk-thomas: Thanks for the updates.

> Hoping this can be squashed before merging to avoid polluting your history.

Absolutely.

PR #18780, opened 2018-08-08
Add rosdep keys for libssh

hogabrie:
I have a project where I use libssh instead of libssh2 which is already present in the rosdep keys. Different than the name might imply libssh2 is not the second version of libssh. libssh is actually more advanced than libssh2, see here for a comparison of libssh and libssh2: https://www.libssh2.org/libssh2-vs-libssh.html

Package links for different distributions:
Ubuntu: https://packages.ubuntu.com/bionic/libssh-dev
Debian: https://packages.debian.org/sid/libssh-dev
Fedora: http://rpmfind.net/linux/rpm2html/search.php?query=libssh-devel

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3012,6 +3012,11 @@ libsrtp0-dev:
   fedora: [libsrtp-devel]
   gentoo: ['net-libs/libsrtp:0']
   ubuntu: [libsrtp0-dev]
+libssh-dev:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Gentoo link: https://packages.gentoo.org/packages/net-libs/libssh

mikaelarguedas (approved):
lgtm

PR #18792, opened 2018-08-09
Adding xsens_official to documentation index for kinetic

Steven-GH:


kinetic/distribution.yaml
```diff
@@ -13875,6 +13875,16 @@ repositories:
       url: https://github.com/ethz-asl/ethzasl_xsens_driver.git
       version: master
     status: maintained
+  xsens_official:
```
tfoote: For consistency please have the name match the repository name so it's easier to find and match.

kinetic/distribution.yaml
```diff
@@ -13875,6 +13875,16 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/xsens/xsens_mti_ros_node.git
```
tfoote: Related i'll note that this repository name is a little misleading as it's more than just the ROS node.

tfoote (changes requested):
Thanks for submitting this. Please at least update the top level entry so it matches the repository name. It's not necessary, but I would suggest considering a different name for the repository, something like `xsense_support` or possibly `xsense_driver_official` then make the top level key match that updated name.

kinetic/distribution.yaml
StevenXsens: Thank you for the feedback. Should it be named driver instead of node? Or do you suggest something else?

tfoote: Often the repository will match the package name, but that's generally only done if there's only one package in the repository. I'd suggest `xsens_support` or something else similar. Or `xsens_ros_driver` maybe.

dirk-thomas:
@StevenXsens Closing due to no response.

PR #18807, opened 2018-08-11
[python-pyudev] [ubuntu] wildcard for python2 and added python3 for >trusty

veimox:
- https://packages.ubuntu.com/bionic/python-pyudev
- https://packages.ubuntu.com/bionic/python3-pyudev

- [x] rosdistro_reformat
- [x] nosetests

PR #18808, opened 2018-08-11
[python-pep8] [ubuntu] wildcard for python2 where possible and added python3 for >trusty

veimox:
- https://packages.ubuntu.com/bionic/python-pep8
- https://packages.ubuntu.com/bionic/python3-pep8

- [x] rosdistro_reformat
- [x] nosetests

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2299,6 +2299,8 @@ python-pep8:
     pip:
       packages: [pep8]
   ubuntu:
+    bionic: [python-pep8]
+    bionic_python3: [python3-pep8]
     precise: [pep8]
     quantal: [pep8]
     raring: [pep8]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2299,21 +2299,21 @@ python-pep8:
     pip:
       packages: [pep8]
   ubuntu:
-    bionic: [python-pep8]
+    '*': [python-pep8]
+    artful_python3: [python3-pep8]
     bionic_python3: [python3-pep8]
+    cosmic_python3: [python3-pep8]
     precise: [pep8]
     quantal: [pep8]
     raring: [pep8]
     saucy: [pep8]
     trusty: [pep8]
+    trusty_python3: [python3-pep8]
     utopic: [pep8]
     vivid: [pep8]
     wily: [pep8]
-    xenial: [python-pep8]
     xenial_python3: [python3-pep8]
-    yakkety: [python-pep8]
     yakkety_python3: [python3-pep8]
-    zesty: [python-pep8]
     zesty_python3: [python3-pep8]
 python-percol:
   debian:


veimox:
@mikaelarguedas I have done three PRs (#18809 #18808 #18807) like this for Ubuntu Bionic. The objective is to have this dependencies resolved for Melodic in Bionic. I was wondering if you prefer the PRs to tackle the whole Melodic distro [Artful, Bionic, Stretch and Arch](http://wiki.ros.org/melodic) instead of just Bionic.

mikaelarguedas:
Thanks @veimox for the contribution.
Yes we prefer PRs adding support for all platforms targeted by a given ROS distribution.

mikaelarguedas:
Just to follow-up on this: in general we ask to provide rules for all currently active ubuntu and debian distributions (even if they are not targeted by a specific ROS distribution, e.g. Ubuntu Cosmic or Debian Buster).
Adding gentoo / fedora / arch is preferred but not required.

veimox:
Currently there are two tags that are overlapping:

- [Here](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L2293) python-pep8 is defined and also some of the <distro>_python3. In this case is defined per release of ubuntu (I guess due to the *_python3 is not available in all of them)
- [Here](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L4429) python3-pep8 is defined. In this case is defined for the all the ubuntu distributions.

I was wondering what is better... On one hand, from a users perspective is probably simpler to just have one tag (e.g. python-pep8) without specifying if it is python3 or 2 and that being defined in an lower level. On the other hand, this creates maintenance problems because I am basically adding the same instruction for the new Ubuntu releases even though the have the same name (and will have in future ones). 

If we take the second approach, wouldn't make sense to have just a 

```yaml
python-pep8:
  ubuntu: [python-pep8]
  ...
```
like in python3-pep8??

PD: Sorry if this is already answered, i couldn't find it :(
PD2: This is also applicable for #18807 and #18807

mikaelarguedas:
@veimox These are good questions.

TL;DR: We should update the `python-<PKG_NAME>` keys.
We are aware of the maintenance burden of having to maintain the `<OS>` and `<OS_python3>` rules and are exploring ways of streamlining it as well as providing better Python 3 support for ROS (https://github.com/ros-infrastructure/rep/pull/149)

---

The keys used in ROS are the `python-<PKG_NAME>` ones. Currently all ROS distributions target Python 2 by default. We provide the `<OS>_python3` versions to allow users to build with python3 without having to change the `package.xml` of all packages.

The `python3-<PKG_NAME>` have been introduced for ROS 2.0 that is targeting Python 3 only and are currently not used by any ROS 1 package.

---

A way to simplify the rules with what is available today, is to use the wildcard syntax for all the python2 rules (we still have to add the `<OS>_python3` rules by hand for now).

For example for #18807:
```
 python-pyudev:
  debian: [python-pyudev]
  fedora: [python-pyudev]
  gentoo: [dev-python/pyudev]
  ubuntu:
    artful: [python-pyudev]
    artful_python3: [python3-pyudev]
    bionic: [python-pyudev]
    bionic_python3: [python3-pyudev]
    cosmic: [python-pyudev]
    cosmic_python3: [python3-pyudev]
    lucid: [python-pyudev]
    maverick: [python-pyudev]
    natty: [python-pyudev]
    oneiric: [python-pyudev]
    precise: [python-pyudev]
    quantal: [python-pyudev]
    raring: [python-pyudev]
    saucy: [python-pyudev]
    trusty: [python-pyudev]
    trusty_python3: [python3-pyudev]
    utopic: [python-pyudev]
    vivid: [python-pyudev]
    wily: [python-pyudev]
    xenial: [python-pyudev]
    xenial_python3: [python3-pyudev]
    yakkety: [python-pyudev]
```

can be simplified to:
```
 python-pyudev:
  debian: [python-pyudev]
  fedora: [python-pyudev]
  gentoo: [dev-python/pyudev]
  ubuntu:
    '*': [python-pyudev]
    artful_python3: [python3-pyudev]
    bionic_python3: [python3-pyudev]
    cosmic_python3: [python3-pyudev]
    trusty_python3: [python3-pyudev]
    xenial_python3: [python3-pyudev]
```

veimox:
@mikaelarguedas 
As suggested, is using wildcard for >xenial distros. The rest are still defined as `pep8`.

Links:
- [python-pep8](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python-pep8&searchon=names)
- [python3-pep8](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python3-pep8&searchon=names)

mikaelarguedas (approved):
lgtm, thanks @veimox for iterating :+1:

wjwwood (approved):

PR #18809, opened 2018-08-11
[python-termcolor] [ubuntu] wildcard for python2 and trusty-cosmic for python3

veimox:
- https://packages.ubuntu.com/bionic/python-termcolor
- https://packages.ubuntu.com/bionic/python3-termcolor

- [x] rosdistro_reformat
- [x] nosetests

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3812,6 +3812,8 @@ python-termcolor:
     pip:
       packages: [termcolor]
   ubuntu:
+    bionic: [python-termcolor]
+    bionic_python3: [python3-termcolor]
     trusty: [python-termcolor]
     trusty_python3: [python3-termcolor]
     utopic: [python-termcolor]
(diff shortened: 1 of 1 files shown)


tfoote:
Same here please add zesty, artful and cosmic.

rosdep/python.yaml
```diff
@@ -3812,13 +3812,13 @@ python-termcolor:
     pip:
       packages: [termcolor]
   ubuntu:
-    trusty: [python-termcolor]
+    '*': [python-pyudev]
```
mikaelarguedas: Copy-n-paste: this should be `python-termcolor`

mikaelarguedas (changes requested):

mikaelarguedas:
> is just a coincidence or is there something against debian unstable?

As `sid` is a moving target it makes keeping the database up-to-date challenging.
From [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules):

`If there's a new distro in prerelease adding a rule for it is fine. However please don't target 'sid' as it's a rolling target and when the keys change our database gets out of date.`

mikaelarguedas (approved):
lgtm thanks @veimox for iterating

PR #18833, opened 2018-08-14
Removing kvaser_interface due to missing prereqs.

JWhitleyWork:
Building this package requires that a Kvaser-specific library be built on the target system. Since that library is not available in the standard Ubuntu repos (though I do have a PPA for it), the build farm build continues to fail. Removing for now.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -6192,16 +6192,6 @@ repositories:
       type: git
       url: https://github.com/uos/kurt_navigation.git
       version: indigo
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
A way to work around the fact that the dependency is not available upstream (Ubuntu or Debian) would be to release the dependency as a ROS package. You will then be able to use it as a dependency of this package.

I'll merge this in the meantime

JWhitleyAStuff:
I would but it's a kernel module.

PR #18840, opened 2018-08-15
Added python-pika-pip

reinzor:
pika is already available as rosdep but installed via `dpkg`, we required the newer version available via `pip`. I just added the entry since I do not want to break anything. What's the best practice here?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2363,6 +2363,13 @@ python-pexpect:
     wily_python3: [python3-pexpect]
     xenial: [python-pexpect]
     xenial_python3: [python3-pexpect]
+python-pika-pip:
+  debian:
+    pip:
+      packages: [pika]
+  ubuntu:
+    pip:
+      packages: [pika]
 python-pip:
   arch: [python2-pip]
   debian: [python-pip]


tfoote (changes requested):
In general if there is a "native" package manager version we default to using that version. It's mentioned in the Contributing guide: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

The version in the debian packages has been tested and QAd against all the other dependent packages. If you override it with a pip based version you have the potential to break anything already using the debian version since the pip installation will become the default overall. Generally upgrades are ok, but sometimes not. We've also had issues that a pip version of something is installed, then a newer or patched version of the debian is installed but the pip version lags and masks out the critical fixes that are in the debian package. So although it's a bit frustrating, I'm going to recommend that we decline this.

Part of being a larger ecosystem is to walk in sync with everyone else. I'm sure that the newer version of pika will be available in future versions of Ubuntu but until then we recommend using the system version for your system. There are tons of libraries that we'd love to jump forward on but we have to make sure to target the low water mark for our supported platforms.

If you really need it you could create your own custom rosdep rules to override our system maintained ones that you could use privately on your systems. But we don't want to add this override for all users who may or may not be experts in managing their system.

dirk-thomas (approved):
Adding this entry as `python-pika-pip` is valid in my opinion. We already allow that for numerous other keys  - that is why it has the `-pip` suffix as recommended.

@tfoote Please re-review.

tfoote:
On platforms where there's not a native package manager we have been adding rules for installing via pip. We added the convention to use `-pip` as an extension for clarity in the rule name as well as to avoid conflicts in the future when the package potentially becomes packaged in the native package manager.

We could have every python package have a pip rule in additoin to the mainline rule. But then we would have package breaking all the time because one package depends on the pip rule and upgrades a system dependency that's already been QAd by the debian team. The rosdeps need to be a consistent set of packages and as per our policy if there's a native version of the package you should use that. If you want to add an overlay or other version you need to fully isolate it so that it won't interfere with other packages. 

There's also issues with pip packages not auto upgrading when apt packages upgrade which also traps users. I stand by my statement that we should not merge a duplicate rule specifically for upgrading a version. We work very hard to work with the system that we're given. There are a ton of things we'd love to upgrade on the system that has newer features than what's in the distro. But we stick to what's in the distro to be a good member of the community and maintain compatibility with everyone else. 

As I've said this can be easily setup on their specific systems if they want to diverge from the distro, but if we add this rule here we are basically diverging the entire ROS community from the debian and ubuntu distros by overriding this dependency and upgrading it. 

Another way to look at this is would be be willing to backport the new version into a debian package, verify that it doesn't break anything and host it in the ROS repos? This has a similar effect, but even more subtlly it only will happen if you install a package with this as a dependency. So you could have a fully running system, add something to your source tree, rosdep install, and you've potentially broken anything depending on the older version of pika.

tfoote:
I'll update the contribution guide to clearlly state this. My suggestion is that  we should state that pip entries that did exist but then become available upstream should be switched to null values so that rosdep will not resolve them when they are packaged into the native package manager on a platform.

This will protect the naive user from accidentally overriding their system installation using the default rules. And for anyone who's working from source and really wants the pip rules they can create a local custom rosdep rule that resolves it to pip on their systems.

ros-discourse:

https://discourse.ros.org/t/should-we-warn-new-users-about-difficulties-with-python-3-and-alternative-python-interpreters/3874/14

PR #18852, opened 2018-08-15
Updated sdformat for xenial

dniewinski:


tfoote:
Thanks, would you mind filling in the other versions of Ubuntu for completeness so this rule can be used more generally.

mikaelarguedas (changes requested):
In bionic and cosmic the package `libsdformat4-dev` doesn't exist, it is `libsdformat6-dev` instead:
Bionic: https://packages.ubuntu.com/bionic/libsdformat6-dev
Cosmic: https://packages.ubuntu.com/cosmic/libsdformat6-dev

As ROS Melodic ships libsdformat6-dev on artful I think we should use that package on artful as well: http://packages.ros.org/ros/ubuntu/pool/main/s/sdformat6/

dniewinski:
@mikaelarguedas Could have sworn I checked that 4 was available on bionic.  It is changed now.  I left Artful using 4 since 6 hasn't been released for it yet

mikaelarguedas:
> I left Artful using 4 since 6 hasn't been released for it yet

I'm afraid this will conflict with the fact that ROS packages released in Artful (ROS Melodic) are assuming `libsdformat6-dev` ([because assuming gazebo9](http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023)), this is why the ROS apt repositories provide libsdformat6-dev for Artful (http://packages.ros.org/ros/ubuntu/pool/main/s/sdformat6/).
So I think we can go 2 ways here:
- use libsdformat6-dev for artful in this PR
- not add artful in this PR as artful is already EOL and will stop being supported by the ROS infrastructure [as of today](https://discourse.ros.org/t/preparing-for-melodic-sync-2018-08-20/5773/3)

dniewinski:
I'll leave it up to you which option we go with..  I have no preference

mikaelarguedas:
Changed to `libsdformat6-dev` for Artful in https://github.com/ros/rosdistro/pull/18852/commits/b3e946a75c8477d9e0844a4dde7a46c97e1fb27d

mikaelarguedas (approved):
lgtm thanks @dniewinski for iterating

PR #18853, opened 2018-08-15
Adding ros_anomaly_detector to documentation index for distro

narayave:
I'd like my repo, ros_anomaly_detector, to be indexed and documented on ros.org

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9778,6 +9778,12 @@ repositories:
       url: https://gitlab.com/InstitutMaupertuis/ros_additive_manufacturing.git
       version: kinetic
     status: developed
+  ros_anomaly_detector:
+    doc:
+      type: git
+      url: https://github.com/narayave/ros_anomaly_detector.git
+      version: master
+    status: maintained
   ros_canopen:
     doc:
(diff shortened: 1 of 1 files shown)


tfoote:
Thanks for the submission. Browsing your package though I noticed that you have ros in your package name which does not follow our naming conventions: http://www.ros.org/reps/rep-0144.html and http://wiki.ros.org/ROS/Patterns/Conventions would you mind renaming before we start generating documentation information so we don't end up with stray documentation for the old name.

narayave:
@tfoote I was just going over the conventions, and I have a quick question. Would it be better to call it anomaly_detector or mh5_anomaly_detector.
The package so far will work for the MH5 robot, but can be easily modified to work for other robots and use cases.

tfoote:
If it's only going to work for the mh5 please scope it to that. In the future once it's made more generic you can change the name to be more generic as well. Or fork it to not break the old package users, but likely take the opportunity to restructure some of it to be more generic and build on lessons learned in the API.

narayave:
I've renamed the package appropriately, and it's reference in the distribution.yaml file.
I will close this PR and open a new one.

PR #18869, opened 2018-08-17
Flake8 for xenial

SteveMacenski:


rosdep/python.yaml
```diff
@@ -1038,6 +1038,7 @@ python-flake8:
   ubuntu:
     trusty: [python-flake8]
     trusty_python3: [python3-flake8]
+    xenial: [python-flake8]
```
mikaelarguedas: as this package is available on [all active ubuntu distributions](https://packages.ubuntu.com/trusty/python-flake8), can you please update this to cover them all?
```yaml
  ubuntu:
    '*': [python-flake8]
    trusty_python3: [python3-flake8]
```
should do the trick

SteveMacenski:
done

mikaelarguedas (approved):
lgtm, thanks @SteveMacenski

rosdep/python.yaml
mikaelarguedas: Python3 rules are added upon request from users using them. As it;s not been requested for 4 years I'm not planning on adding them here

tfoote (approved):

PR #18914, opened 2018-08-21
Add python subprocess32 package

bobhenz-jabil:
https://packages.debian.org/stretch/python-subprocess32
https://apps.fedoraproject.org/packages/python-subprocess32
https://packages.gentoo.org/packages/dev-python/subprocess32
https://packages.ubuntu.com/bionic/python-subprocess32

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3647,6 +3647,17 @@ python-sqlalchemy:
     wily: [python-sqlalchemy]
     xenial: [python-sqlalchemy]
     xenial_python3: [python3-sqlalchemy]
+python-subprocess32:
+  debian:
+    '*': [python-subprocess32]
+    jessie:
+      packages: []
+  fedora: [python-subprocess32]
+  gentoo: [dev-python/subprocess32]
+  ubuntu:
+    '*': [python-subprocess32]
+    trusty:
+      packages: []
 python-support:
   debian: [python-support]
   fedora: [python]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3650,14 +3650,12 @@ python-sqlalchemy:
 python-subprocess32:
   debian:
     '*': [python-subprocess32]
-    jessie:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3647,6 +3647,11 @@ python-sqlalchemy:
     wily: [python-sqlalchemy]
     xenial: [python-sqlalchemy]
     xenial_python3: [python3-sqlalchemy]
+python-subprocess32:
+  debian: [python-subprocess32]
+  fedora: [python-subprocess32]
+  gentoo: [dev-python/subprocess32]
+  ubuntu: [python-subprocess32]
```
mikaelarguedas: This package is not available on trusty, so we should explicitely set the rule to `null` for that platform

rosdep/python.yaml
```diff
@@ -3647,6 +3647,11 @@ python-sqlalchemy:
     wily: [python-sqlalchemy]
     xenial: [python-sqlalchemy]
     xenial_python3: [python3-sqlalchemy]
+python-subprocess32:
+  debian: [python-subprocess32]
```
mikaelarguedas: This package is not available on jessie, so we should explicitely set the rule to `null` for that platform

mikaelarguedas (changes requested):
@bobhenz-jabil thanks for contributing. Can you please provide a description of the package as well a the use cases you need it for as per [contribution guildelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

bobhenz-jabil:
Brief Description: subprocess32 is a backport of the Py3 stdlib subprocess module for Py2. 

Use case: One of the very valuable features of this package is the ability to add a "timeout" parameter to various subprocess methods preventing calls to the system from hanging nodes indefinitely.

mikaelarguedas (approved):
lgtm thanks @bobhenz-jabil for iterating

PR #18926, opened 2018-08-22
trac_ik: 1.5.0-0 in 'melodic/distribution.yaml' [bloom]

pbeeson:
Increasing version of package(s) in repository `trac_ik` to `1.5.0-0`:

- upstream repository: https://bitbucket.org/traclabs/trac_ik.git
- release repository: https://github.com/traclabs/trac_ik-release.git
- bloom version: `0.6.2`

## trac_ik

```
* Switch to C++-11 threads and pointers instead of Boost
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4427,6 +4427,27 @@ repositories:
       url: https://github.com/ethz-adrl/towr.git
       version: master
     status: developed
+  trac_ik:
+    doc:
+      type: git
+      url: https://bitbucket.org/traclabs/trac_ik.git
+      version: master
+    release:
+      packages:
+      - trac_ik
+      - trac_ik_examples
+      - trac_ik_kinematics_plugin
+      - trac_ik_lib
+      - trac_ik_python
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/traclabs/trac_ik-release.git
+      version: 1.5.0-0
+    source:
+      type: git
+      url: https://bitbucket.org/traclabs/trac_ik.git
+      version: master
+    status: maintained
   tracetools:
     doc:
       type: git


pbeeson:
I’m getting spammed hardcore from the buildfarm about melodic branches not existing on hgithib since I ran bloom to track melodic.

Like many dozens of emails that won’t stop coming.

pbeeson:
http://build.ros.org/job/Msrc_uB__trac_ik_python__ubuntu_bionic__source/1/display/redirect

mikaelarguedas:
Source jobs on the buildfarm are ran up to 5 times before being disabled automatically.

All the failing jobs are disabled disabled automatically after 5 tries, so you won't receive any new emails until you make a new release of this repository.

pbeeson:
Yeah but the issue doesn’t seem related to my package at all. It’s looking
for a branch on the github release repo that never got set up. Seems like a
scripting issue.

dirk-thomas:
Please include a link to a failing build / the error message.

pbeeson:
I did. It’s above your initial reply

dirk-thomas:
> you won't receive any new emails until you make a new release of this repository.

Every night the job will be reenabled automatically to try again. If the problem is still the same the job will disable itself again (until the next 24h).

For a permanent solution (until you can fix the release) you can remove [this line](https://github.com/ros/rosdistro/pull/18926/files#diff-7d7a32deea3f312a3b8b22a74f23eb08R4445) through a new PR.

mikaelarguedas:
As of why the jobs are failing, it looks like you don't have any `debian/melodic*` branches on your release repository. This means that bloom failed to generate the release.

Do you happen to still have the console output of when you ran `bloom-release`?

The most common reason for this type of failure is that you were missing rosdep keys and that bloom failed with a loud warning and when it asked if you were "absolutely sure" you wanted to continue.

mikaelarguedas:
> Every night the job will be reenabled automatically to try again. If the problem is still the same the job will disable itself again (until the next 24h).

Correct, my bad, they will be retried once every 24hours

> For a permanent solution (until you can fix the release) you can remove this line through a new PR.

:+1:

pbeeson:
It did yell that there wasnt a pr2_description (why?). But the output only
said to continue if you were absolutely sure that the dependency did not
exist. It does not exist in melodic, so I continued. That’s not a super
well written message.

I guess I can take out my examples that used pr2 to show usage but it seems
dumb that OSRF can’t move simple URDF packages that have been in ROS since
C-Turtle and that they created (or at least willow did) into newer versions
of ROS. Pr2_description doesn’t even compile.

dirk-thomas:
> it seems
dumb that OSRF can’t move simple URDF packages that have been in ROS since
C-Turtle and that they created (or at least willow did) into newer versions
of ROS

Nobody at OSRF is a maintainer of `pr2_description`. The package manifest of that package even states explicitly that the package is orphaned: https://github.com/PR2/pr2_common/blob/6f96993793230c81737478729c4e324c3574424a/pr2_description/package.xml#L8

So in order to make it available in new ROS distros someone has to take on the responsibility and release it (and potentially address problem if any are present).

mikaelarguedas:
> It did yell that there wasnt a pr2_description (why?).

Recommended workflow in ROS is to leverage the ["blocked_releases" status page](http://repositories.ros.org/status_page/blocked_releases_melodic.html) to see if all the dependencies of your packages are satisfied. (Or rely on bloom to tell you that they are missing)
If some dependencies are missing, open an issue on the repositories of your dependencies to request a new release.

> But the output only said to continue if you were absolutely sure that the dependency did not
exist. It does not exist in melodic, so I continued.  That’s not a super well written message.

I understand the confusion, if you have an idea of wording for the error message, please consider opening a PR on [ros-infrastructure/bloom](https://github.com/ros-infrastructure/bloom) with the suggested modification.

pbeeson:
Also I'm just re-running bloom for melodic with that run depend removed.

mikaelarguedas:
yes, the branches were not created as the generator was skipped for that platform (I guess all platforms if you said if you said yes for each of them). https://github.com/ros-infrastructure/bloom/blob/a7bc25e999e09621d759807e118a0ee72e58542c/bloom/commands/git/release.py#L291-L295. So the branches cannnot be cloned because they were never created and pushed.

We could make the error message more explicit in bloom to clarify what the consequences of "skipping a generator" are.

wjwwood:
@mikaelarguedas but we don't know which are required and which are not. Perhaps instead bloom should not offer to open a pr if no platforms were generated?

mikaelarguedas:
> but we don't know which are required and which are not. 

@wjwwood Can you clarify what `which are required` refers to ? dependencies ? platforms?

> Perhaps instead bloom should not offer to open a pr if no platforms were generated?

That sounds good to me. :+1:

We could go further and point to how to blacklist package for a platform. Likely not in bloom itself though as it introduces unwanted coupling between the tools. But maybe on the wiki so that people can figure it out with a google search.

wjwwood:
Platforms, there's a list of expected ones somewhere, that's why you don't get a follow up question if a dependency is missing on Fedora.

PR #18927, opened 2018-08-22
See if we can release newer branch

mikeferguson:
In response to https://github.com/ros-perception/sparse_bundle_adjustment/issues/5#issuecomment-415067011

If this does work, I'll have additional PRs to update the documentation/release.


mikaelarguedas:
thanks! :+1:

PR #18959, opened 2018-08-27
Add clang-tidy rosdep keys for debian and ubuntu

jlack1987:
Adding the following keys for `clang-tidy`:

- Ubuntu: https://packages.ubuntu.com/search?keywords=clang-tidy
- Debian Stretch: https://packages.debian.org/stretch/clang-tidy
- Debian Buster: https://packages.debian.org/buster/clang-tidy

There's a gentoo package named `clang`, but i'm very unfamiliar with gentoo and was unable to determine if it has `clang-tidy` contained within, so I left it out. If I need to add it back in let me know. Didn't see anything for Fedora.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -400,6 +400,13 @@ clang-format:
     stretch: [clang-format]
   gentoo: [sys-devel/clang]
   ubuntu: [clang-format]
+clang-tidy:
+  debian:
+    buster: [clang-tidy]
+    stretch: [clang-tidy]
+  ubuntu:
+    '*': [clang-tidy]
+    trusty: null
 cmake:
   alpine: [cmake]
   arch: [cmake]


wjwwood (changes requested):
It looks like `clang-tidy` is new in Xenial 16.04, but we require rosdep rules to work on Trusty 14.04 as well.

Can you please use the `'*'` mechanism and specifically provide `trusty: null`?

For example: https://github.com/ros/rosdistro/blob/79242ea09a3703b0f715b5742eea3d7861e103c3/rosdep/base.yaml#L3436-L3438

jlack1987:
Interesting, didn't know that `'*'` functionality existed. Fixed and thanks!

wjwwood (approved):

wjwwood:
Thanks for iterating. I have to wait for a second review because this is a rosdep rule change, but someone will get to it soon hopefully.

PR #18960, opened 2018-08-27
change lunar branch for open_karto

mikeferguson:
We ended up releasing the melodic-devel version of sparse_bundle_adjustment, migrating open/slam_karto one at a time to match

### DIFF
--- a/lunar/distribution.yaml
+++ b/lunar/distribution.yaml
@@ -2654,7 +2654,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-perception/open_karto.git
-      version: indigo-devel
+      version: melodic-devel
     release:
       tags:
         release: release/lunar/{package}/{version}
@@ -2663,7 +2663,7 @@ repositories:
     source:
       type: git
       url: https://github.com/ros-perception/open_karto.git
-      version: indigo-devel
+      version: melodic-devel
     status: maintained
   open_street_map:
     doc:


wjwwood (commented):
I'm fine with this, if you are @mikeferguson, but it's a bit confusing since the latest version on `melodic-devel` is `1.2.0` but the latest version in Lunar is still `1.1.4`. So if someone got "lunar from source" they wouldn't just get additional unreleased changes on top of 1.1.4, they'll also get additional versions. Is that what you intended?

mikeferguson:
@wjwwood -- I am planning to do a release once I see that things build in the dev job -- did the same thing for SBA (and will do the same thing for slam_karto once open_karto is out in debs).

mikeferguson:
(and we actually caught a problem with the SBA dev job, that we fixed before doing the debian release)

wjwwood:
Sounds good.

mikaelarguedas:
@mikeferguson it looks like `open_karto` built successfully for lunar. I guess the next step is to release 1.2.X in lunar in order to get `slam_karto` compiling again ?

mikeferguson:
Slam_karto failed dev test after changing branches - I haven’t had a chance yet to fix (should be able to do so early this week)

mikaelarguedas:
> Slam_karto failed dev test after changing branches

Oh I missed that, thanks for clarifying :+1:

> I haven’t had a chance yet to fix (should be able to do so early this week)

looks like C++11 types are used in the headers but the dependant package doesn't set the CMAKE_CXX_STANDARD/compiler flags accordingly.

mikeferguson:
Yep -- I knew what it was -- just didn't get a chance to do it -- fixed and released now -- PR: https://github.com/ros/rosdistro/pull/19108

PR #19012, opened 2018-08-30
Added libdlib-dev To The rosdep base.yaml File

garethellis0:
Rational For Adding Package:
Needed for it's global optimization functions, ie. `find_max_global` and `find_min_global`

Package web listings:
- Debian: https://packages.debian.org/search?keywords=libdlib-dev
- Ubuntu: https://packages.ubuntu.com/xenial/libdlib-dev
- Gentoo: https://packages.gentoo.org/packages/sci-libs/dlib

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1375,7 +1375,6 @@ libdevil-dev:
   ubuntu: [libdevil-dev]
 libdlib-dev: 
   debian: [libdlib-dev]
-  fedora: [dlib-devel]
   gentoo:
     portage:
       packages: [sci-libs/dlib]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1378,7 +1378,9 @@ libdlib-dev:
   gentoo:
     portage:
       packages: [sci-libs/dlib]
-  ubuntu: [libdlib-dev]
+  ubuntu: 
+    '*': [libdlib-dev]
+    trusty: null
 libdmtx-dev:
   arch: [libdmtx]
   debian: [libdmtx-dev]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -1373,6 +1373,12 @@ libdevil-dev:
     portage:
       packages: [media-libs/devil]
   ubuntu: [libdevil-dev]
+libdlib-dev: 
+  debian: [libdlib-dev]
+  gentoo:
+    portage:
+      packages: [sci-libs/dlib]
+  ubuntu: [libdlib-dev]
```
wjwwood: This isn't available in trusty, so this will not work, please use the `'*'` syntax and explicitly disable it for trusty, see: https://github.com/ros-infrastructure/rep/blob/master/rep-0111.rst#allow-package_argument-to-be-explicitly-null

wjwwood (changes requested):
If there's a fedora rule I would ask that you put it in, but I don't know why you remove it with the follow up commit.

garethellis0:
I took a guess at what the fedora rule would be initially, but after looking into it, it didn't look like there was a corresponding package.

rosdep/base.yaml
garethellis0: Ah, thank you, changed.

wjwwood (approved):
The changes look ok, but CI is failing (you have trailing whitespace some where).

garethellis0:
@wjwwood - looks like we should be good to go now!

garethellis0:
@wjwwood - sorry to ping you about this again, but my team has some changes (namely a migration from kinetic to melodic) blocked on this

PR #19015, opened 2018-08-31
Python boto3

HEllRZA:
We use [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) (that's the Amazon Web Services API) in one of our projects. Source code [here](https://github.com/boto/boto3) Would be nice to have it via rosdep.

[arch](https://www.archlinux.org/packages/community/any/python-boto3/)
[debian](https://packages.debian.org/search?searchon=names&keywords=boto3)
[fedora](https://apps.fedoraproject.org/packages/python2-boto3)
[gentoo](https://packages.gentoo.org/packages/dev-python/boto3)
[opensuse](https://software.opensuse.org/package/python-boto3)
[ubuntu](https://launchpad.net/ubuntu/+source/python-boto3)

Thank You!

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -576,7 +576,9 @@ python-boto3:
   fedora: [python2-boto3]
   gentoo: [dev-python/boto3]
   opensuse: [python-boto3]
-  ubuntu: [python-boto3]
+  ubuntu:
+    '*': [python-boto3]
+    trusty: null
 python-bottle:
   debian: [python-bottle]
   fedora: [python-bottle]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -570,6 +570,13 @@ python-boltons:
     zesty:
       pip:
         packages: [boltons]
+python-boto3:
+  arch: [python-boto3]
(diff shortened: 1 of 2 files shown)


rosdep/python.yaml
```diff
@@ -570,6 +570,13 @@ python-boltons:
     zesty:
       pip:
         packages: [boltons]
+python-boto3:
+  arch: [python-boto3]
+  debian: [python-boto3]
+  fedora: [python2-boto3]
+  gentoo: [dev-python/boto3]
+  opensuse: [python-boto3]
+  ubuntu: [python-boto3]
```
wjwwood: Looks like this is new in Xenial, but we require support for Trusty still. Please use the `'*'` syntax and `null` for trusty, see:

https://github.com/ros-infrastructure/rep/blob/master/rep-0111.rst#allow-package_argument-to-be-explicitly-null

wjwwood (changes requested):

wjwwood (approved):

wjwwood:
Lgtm, thanks for iterating. Just have to wait on a second review.

PR #19058, opened 2018-09-05
add python-more-itertools-pip

benmaidel:
https://pypi.org/project/more-itertools/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1913,6 +1913,19 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
+python-more-itertools-pip:
+  arch:
+    pip:
+      packages: [more-itertools]
+  debian:
+    pip:
+      packages: [more-itertools]
+  fedora:
+    pip:
+      packages: [more-itertools]
+  ubuntu:
+    pip:
+      packages: [more-itertools]
 python-msgpack:
   arch: [python2-msgpack]
   debian: [python-msgpack]


wjwwood (approved):

nuclearsandwich:
I noticed gentoo was absent but it appears that none of the `-pip` keys have a gentoo entry so that seems fine.

Timple:
Any particular reason the apt-version is not used for ubuntu?

tfoote:
It looks like it was not available back when this was submitted. We should switch over now that it's available as of bionic and buster. 

https://packages.ubuntu.com/bionic/python-more-itertools
https://packages.ubuntu.com/groovy/python3-more-itertools


https://packages.debian.org/buster/python3-more-itertools

If you have a moment to submit a PR to add those rules and sunset this rule that would be appreciated.

Timple:
Sunsetting this rule means that could break existing installations which depend on the `-pip` prefix right? 
I could simply add the new rules without `-pip`?

tfoote:
Maintaining parallel rules with `-pip` and regular leads to many more problems because the pip version will mask the debian based one. This means that if any package on the system asks for the pip version all packages on the system will get the pip version, even if they asked for the apt version. 

My default suggestion would be to create the apt rules for all existant platforms. And leave the pip rule for only those platforms without the apt based systems as that is our policy. If there's an known large userbase using the pip rule from source a slightly more conservative approach is to say that all future platforms the pip rule is null and only add the apt rules on those platforms. (The better way to write it is to explicitly list all currently supported platforms and not write any others, but functionally you achieve preventing resolution on future platforms.) The balance here is that noone can use the apt based systems on currently supported platforms, it will only be available in the next release. I usually bias towards switching to having apt resolve because that enables the useage of the key in released packages and if you need the newer pip version you can create your own rosdep override if you're working from source.

PR #19070, opened 2018-09-06
Fixed branch for kinetic

maxbader:
I changed the branch name for kinetic to prevent conflicts with melodic


maxbader:
I also added now my packages to melodic for indexing

PR #19112, opened 2018-09-12
[visualstates] Fixing version mismatch error

okanasik:


mikaelarguedas:
It seems that the last version bloomed on the relese repository is 0.1.1-0. This is why CI is failing.
Can you clarify how you faced a version mismatch?

okanasik:
@mikaelarguedas I might be missing something. When I released the package, all the checks are passed and the package is merged. Then ros build system reported build errors. When I check the logs of the build system, I see that package.xml version does not match setup.py version. Therefore, I fixed the error, remove the old version tag and add new version tag with same name, but with the latest fix. After this change I see that jobs starting with `Kdev_`, `Kdoc_` and `Ksrc` is built successfully. However, I still get the same error from the jobs starting with `Kbin_`. Actually, I do not know how to fix this problem what those namings stands for. Recently, I thought I need to create pull request to signal the bloom to get the latest version from the development repo and for that purpose I create this pull request. 

To fix this problem, is it enough to update `release_inc` to 1. More importantly, will this solve version mismatch causing build failure `Kbin_`

I am really sorry for all these confusion. I tried to follow the instructions, but I am totally new to this package release process. Best,

mikaelarguedas:
The Kbin jobs are the ones building binary packages (.deb) that users will be able to install via apt at the end of the day.

> To fix this problem, is it enough to update release_inc to 1. More importantly, will this solve version mismatch causing build failure Kbin_

As you made changes to the sources of your package, you will need to rerun `bloom-release` for the code on the 0.1.1 tag of your source repository to be pushed to the release repository.
Though the recommended workflow is to not modify existing tags but to release a new version (0.1.2 in your case).

Regarding the original issue of the version mismatch between the package.xml and the setup.py, this should not happen. It can be avoided by removing the [hardcoding of the version in the setup.py](https://github.com/JdeRobot/VisualStates/blob/952d6e2aa721d3c7dd42cc70c04212c1bd057d8c/setup.py#L7) and let ROS set the version for you (you can see [rospy's setup.py](https://github.com/ros/ros_comm/blob/c6ca45d448a197a0013f573e4f746f264f15528e/clients/rospy/setup.py) for reference)

okanasik:
Thank your for the information. I think the eternal solution is to remove the version number from setup.py :) . We make a new big release that should also fix this problem.

PR #19116, opened 2018-09-13
capitalization fixed

pholthaus:
In Gentoo, there is only OpenNI, not OpenNi. I don't know if that changed only recently or the spelling here was a typo.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2269,7 +2269,7 @@ libopenni-dev:
   arch: [openni]
   debian: [libopenni-dev]
   fedora: [openni-devel]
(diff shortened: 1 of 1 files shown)


tfoote:
Thanks for the attention. Is there a link you can share to show the correct package for reference?

nuclearsandwich:
https://packages.gentoo.org/packages/dev-libs/OpenNI

nuclearsandwich (approved):

pholthau:
> Thanks for the attention. Is there a link you can share to show the correct package for reference?

Sorry for the missed out reply. Somehow I didn't notice the e-mail...

PR #19154, opened 2018-09-20
add python-google-cloud-texttospeech-pip

fmessmer:
https://pypi.org/project/google-cloud-texttospeech/


tfoote (approved):
lgtm, there doesn't seem to be any natively packaged versions.

PR #19156, opened 2018-09-21
Added libevdev, a wrapper system library for interacting with evdev devices

Vuwij:
libevdev is a library for handling evdev kernel devices. It abstracts the evdev ioctls through type-safe interfaces and provides functions to change the appearance of the device.

https://www.freedesktop.org/software/libevdev/doc/latest/

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1685,7 +1685,7 @@ libev-dev:
 libevdev-dev:
   arch: [libevdev]
   debian: [libevdev-dev]
-  fedora: [libevdev]
+  fedora: [libevdev-devel]
   ubuntu: [libevdev-dev]
 libexpat1-dev:
   arch: [expat]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1682,6 +1682,8 @@ libev-dev:
   fedora: [libev-devel]
   gentoo: [dev-libs/libev]
   ubuntu: [libev-dev]
+libevdev-dev:
+  ubuntu: [libevdev-dev]
 libexpat1-dev:
   arch: [expat]
   debian: [libexpat1-dev]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -1682,6 +1682,11 @@ libev-dev:
   fedora: [libev-devel]
   gentoo: [dev-libs/libev]
   ubuntu: [libev-dev]
+libevdev-dev:
+  arch: [libevdev]
+  debian: [libevdev-dev]
+  fedora: [libevdev]
```
tfoote: I believe the equivalent package in fedora is libevdev-devel

https://rpmfind.net/linux/rpm2html/search.php?query=libevdev-devel&submit=Search+...&system=&arch=

tfoote (changes requested):
Ubuntu: https://packages.ubuntu.com/xenial/libevdev-dev
Debian: https://packages.debian.org/stretch/libevdev-dev
Fedora: https://rpmfind.net/linux/rpm2html/search.php?query=libevdev-devel&submit=Search+...&system=&arch=
Arch: https://www.archlinux.org/packages/extra/x86_64/libevdev/

Please update the Fedora rule.

I also found libev in brew for osx? https://brewinstall.org/Install-libev-on-Mac-with-Brew/ which I'm not sure if it's equivalent

rosdep/base.yaml
Vuwij: Hi, I've fixed the changes. libev is different from libevdev and probably should go in the libev rules rather than the libevdev rules

tfoote (approved):

mikaelarguedas:
Added gentoo: https://packages.gentoo.org/packages/dev-libs/libevdev

mikaelarguedas (approved):
:+1: lgtm

PR #19168, opened 2018-09-23
[odb] added odb in base.yaml

veimox:
Added for Ubuntu, Debian and Fedora.  

- [X] rosdistro_reformat
- [X] nosetests

- [Ubuntu](https://packages.ubuntu.com/search?keywords=odb)
- [Debian](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=odb)
- [Fedora](https://apps.fedoraproject.org/packages/s/odb)
- [Just in case](https://pkgs.org/download/odb)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4038,7 +4038,9 @@ ocl-icd-opencl-dev:
 odb:
   debian: [odb]
   fedora: [odb]
-  ubuntu: [odb]
+  ubuntu:
+    '*': [odb]
+    trusty: null
 omniidl:
   arch: [omniorb]
   debian: [omniidl, omniorb-idl]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

mikaelarguedas (changes requested):
As this package is not available on Ubuntu Trusty (platform targeted by ROS Indigo that is still an active ROS distribution), we need to make a null entry for that platform as the dependency cannot be satisfied

veimox:
@mikaelarguedas I used wildcard for ubuntu and null trusty as in [here](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L407)

tfoote (approved):

mikaelarguedas (approved):
lgtm, thanks @veimox for iterating :+1:

PR #19184, opened 2018-09-26
grpc: 0.0.10-0 in 'kinetic/distribution.yaml' [bloom]

BillWSY:
Increasing version of package(s) in repository `grpc` to `0.0.10-0`:

- bloom version: `0.6.7`
- previous version for package: `0.0.9-0`

## grpc

```
* Follow upstream grpc 1.15.1 (#33 <https://github.com/CogRob/catkin_grpc/issues/33>)
  * Sync grpc to upstream 1.15.1
  * Typo fix
* Contributors: Shengye Wang
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3411,7 +3411,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/CogRobRelease/catkin_grpc-release.git
-      version: 0.0.9-0
+      version: 0.0.10-0
     source:
       type: git
       url: https://github.com/CogRob/catkin_grpc.git


tfoote:
As of at least xenial there is grpc available in ubuntu packages: https://packages.ubuntu.com/source/xenial/grpc Can you comment on the compatibility between this package and the upstream one. We shouldn't release packages that mask/overshadow system dependencies in most cases.

I'll merge this as the package is already there, but a plan should be made to switch to use the system version.

BillWSY:
This package provides a `generate_proto` function, which I don't think the system package provides.

I think a reasonable plan is to use this package to provide `generate_proto` only on the platforms with the system package available.

PR #19190, opened 2018-09-27
Python3 version of pydot

stonier:
Wondering if '*_python3' works and is a better solution?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4538,7 +4538,9 @@ python3-pkg-resources:
   gentoo: [dev-python/setuptools]
   ubuntu: [python3-pkg-resources]
 python3-pydot:
+  arch: [python-pydot]
   debian: [python3-pydot]
+  gentoo: [dev-python/pydot]
   ubuntu: [python3-pydot]
 python3-pyparsing:
   arch: [python-pyparsing]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2646,7 +2646,7 @@ python-pydot:
     pip:
       packages: [pydot]
   slackware: [pydot]
-  ubuntu: 
+  ubuntu:
     '*': [python-pydot]
     bionic_python3: [python3-pydot]
 python-pyexiv2:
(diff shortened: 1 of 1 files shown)


stonier:
Ok, so I tried to do some validation here and ended up yak shaving. I've got a python3 version of rosdep now pointed at `python.yaml` of a local clone before the default version and `ROSDISTRO_INDEX_URL` discovering the bionic bouncy packages. 

However, for the life of me, I cannot work out how to get it to resolve to the `bionic_python3` versions of anything. It doesn't use `ROS_PYTHON_VERSION` and I've crawled the rosdep sources and there doesn't seem to be any redirection of this sort happening at all. The only way I can get it to trigger is if I call `rosdep --os ubuntu:bionic_python3`.

This is all in the name of getting the rosdeps lined up so I can bloom a python package for bouncy. I may be entering the fray too early?

nuclearsandwich:
> This is all in the name of getting the rosdeps lined up so I can bloom a python package for bouncy. I may be entering the fray too early?

I'd recommend adding a new key `python3-pydot`. For most existing ROS 2 packages we're creating separate python3-LIBNAME keys and using the explicitly python3 keys for ROS 2. The `_python3` distribution codenames were part of an python3 compatibility effort. To use them requires the passing of a custom distribution name when running rosdep (e.g. `rosdep resolve --os=ubuntu:xenial_python3 ...`). To the best of my knowledge there's no settled discussion on having rosdep choose an appropriate python version for a package so the two separate keys approach will be recommended for Bouncy and Crystal.

There's more ongoing discussions in https://github.com/ros2/design/pull/139, https://github.com/ros-infrastructure/rosdep/pull/607, https://github.com/ros-infrastructure/rep/pull/149

stonier:
Updated with the new key `python3-pydot`. I'll keep an eye on how things progress there.

clalancette:
Hm, weird.  Let me close and reopen to see if this was a temporary Travis thing.

nuclearsandwich:
I've added keys for Archlinux and Gentoo which aren't required but are nice to have.

Gentoo: https://packages.gentoo.org/packages/dev-python/pydot
Archlinux: https://www.archlinux.org/packages/community/any/python-pydot/

pydot does not appear to be packaged for Fedora or Alpine Linux at this time.

nuclearsandwich (approved):

PR #19197, opened 2018-09-28
Add python-sympy-pip

lalten:
apt provides an outdated version of sympy. pip has the most recent

Example where this is a problem: https://answers.ros.org/question/304466/rosdep-pip-instead-of-apt/


tfoote (changes requested):
Please see the discussion in #18840 about overriding system dependencies with pip.

clalancette:
No response in over a month, and we don't want to be adding new pip keys to override system packages this way (from the discussion in https://github.com/ros/rosdistro/pull/18840).  I'm going to close this out for now, but if you'd like to discuss this more, feel free to keep commenting or re-open.

PR #19200, opened 2018-09-28
vmrc: 0.3.0-1 in 'kinetic/distribution.yaml'

caguero:
Initial release:

* Upstream repository: https://bitbucket.org/osrf/vmrc
* Release repository: https://bitbucket.org/osrf/vmrc-release
* distro file: `kinetic/distribution.yaml`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -13820,6 +13820,11 @@ repositories:
       url: https://bitbucket.org/osrf/vmrc
       version: default
     release:
+      packages:
+      - usv_gazebo_plugins
+      - vmrc_gazebo
+      - wamv_description
+      - wamv_gazebo
       tags:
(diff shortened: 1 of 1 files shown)


clalancette:
@caguero Looking at the logs, it looks like this had a few different failures:

1.  vmrc is in the wrong "place" in the rosdistro file, so it is not in alphabetical order: https://travis-ci.org/ros/rosdistro/jobs/434817043#L732
1.  It couldn't verify the remote type of hg for at least part of it at: https://travis-ci.org/ros/rosdistro/jobs/434817043#L7436.  I think this led to the rest of the failures, though I'm not sure about that.

Rather than trying to fix these by hand, my suggestion here is to close this out and try to use `bloom-release` to do the releases here.  That should do more of what you want in an automated way.

PR #19205, opened 2018-09-30
Revert leap_motion package back to 0.0.11-0

nowittyusername:
Make the buildfarm use version 0.0.11 again. Reverts pull requests https://github.com/ros/rosdistro/pull/19173 https://github.com/ros/rosdistro/pull/19172


tfoote:
This is failing because 0.11 was not released for melodic #19173 adds leap_motion and does not upgrade it. You need to resolve that so this can pass CI before this can be merged.

PR #19235, opened 2018-10-05
Added xdot package to Kinetic distribution.yaml

yotabits:
v 2.0.1-0

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -14234,8 +14234,8 @@ repositories:
     release:
       tags:
         release: release/kinetic/{package}/{version}
-       url: https://github.com/yotabits/xdot-release.git
-       version: 2.0.1-0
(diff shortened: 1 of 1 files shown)


clalancette:
@yotabits So the "old" release of xdot in Indigo was released from https://github.com/jbohren/xdot-release.git (with the sources at https://github.com/jbohren/xdot.git).  How does this package relate to that one?

tfoote (changes requested):
There's also an xdot system package that this looks like it might collide with: https://packages.ubuntu.com/xenial/xdot Maybe a rodep rule would be a better solution?

yotabits:
Sorry I forget to make those basic checks, I was in hurry unfortunately...

yotabits:
Sorry I forget to make those basic checks, I was in hurry unfortunately...

PR #19255, opened 2018-10-09
rosdep: adding python-fysom-pip

machinekoder:
Adds the pip version for Fysom.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1355,6 +1355,16 @@ python-fysom:
   ubuntu:
     pip:
       packages: [fysom]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
As of xenial there's an apt based package we should be using that now: https://packages.ubuntu.com/xenial/python-fysom We already have debian packages defined: https://github.com/ros/rosdistro/pull/19255/files#diff-701c2f61a9f7ac8fc393ccc6630c8fa5R1347

Like discussed in #18840 we should prioritize the apt version and not override it with the pip version.

PR #19291, opened 2018-10-15
rosdep: add libzstd-dev

xqms:
needed for nimbro_network (see https://github.com/AIS-Bonn/nimbro_network)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3643,12 +3643,7 @@ libzstd-dev:
   debian: [libzstd-dev]
   fedora: [libzstd-devel]
   gentoo: [app-arch/zstd]
-  ubuntu:
-    artful: [libzstd-dev]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
lgtm. I simplified the ubuntu logic as it's available on all currently supported platforms.

https://packages.ubuntu.com/bionic/libzstd-dev

xqms:
Hmm, I selected libzstd1-dev (note the 1) for xenial since it's slightly newer, with features that I need. Is that possible?

Sorry, should have pointed that one out.

tfoote:
Sorry I missed that it was differrent. In that case we can do a '*' rule with a specific callout for xenial, could you update it that way?

xqms:
Updated with '*' syntax. Thanks for the hint, wasn't aware of this feature :-)

mjcarroll (approved):

PR #19292, opened 2018-10-16
rosdep: added ubuntu pip dependency for python gRPC framework 'grpcio'

atiderko:
see https://grpc.io/blog/installation for install details of the package

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1412,21 +1412,6 @@ python-gpiozero:
       pip:
         packages: [gpiozero]
     zesty: [python-gpiozero]
-python-grpcio:
-  debian:
-    '*': [python-grpcio]
-    jessie:
-      pip: [grpcio]
-    stretch:
-      pip: [grpcio]
-  ubuntu:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Looking ahead I see that grpcio has been packaged for debian as of buster: https://packages.debian.org/source/buster/python-grpcio and ubuntu as of cosmic: https://packages.ubuntu.com/source/cosmic/python-grpcio

Since we want to prioritize the native package manager versions we should create a rule that will select those by default and manually provide the pip rules on the older platforms. And not use the pip postfix for the name.

Making the ubuntu rules look more like this:
```
'*': [python-grpcio]
xenial:
  pip: [grpcio]
yakkety:
  pip: [grpcio]
...
```
With Xenial through Bionic defined for pip. Likewise with Jessie and Stretch using pip and buster and newer represented by the wildcard.

This will give us future compatibility with upcoming distros and support existing distros.

tfoote (approved):
I fixed the alphabetization it lgtm now.

PR #19299, opened 2018-10-17
Add the orb_slam2_ros package to the index

lennarthaller:


kinetic/distribution.yaml
```diff
@@ -7581,6 +7581,15 @@ repositories:
       type: git
       url: https://github.com/evocortex/optris_drivers.git
       version: kinetic-devel
+  orb_slam2_ros:
+    doc:
+      type: git
+      url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
+    release:
```
tfoote: release sections require having run the bloom release tool or otherwise having created a git buildpackage compliat repository.

I think that you meant to create a `doc` and `source` entries.

kinetic/distribution.yaml
```diff
@@ -7581,6 +7581,15 @@ repositories:
...
+      tags:
+      url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
+      version: 0.0.1
```
tfoote: The version for a source entry needs to be a branch name.

kinetic/distribution.yaml
```diff
@@ -7581,6 +7581,15 @@ repositories:
       type: git
       url: https://github.com/evocortex/optris_drivers.git
       version: kinetic-devel
+  orb_slam2_ros:
+    doc:
+      type: git
+      url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
```
tfoote: You're missing a version tag here. Usually you put the development branch so it gets updated on new changes.

tfoote (changes requested):
This has several formatting issues that need to be resolved. I suggest looking at some of the other entries without a release tag for reference.

kinetic/distribution.yaml
```diff
@@ -7586,10 +7586,6 @@ repositories:
       type: git
       url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
       version: master
-    source:
```
tfoote: If you want to disable CI please keep the source entry but turn off ci using `test_commits: false` like https://github.com/ros/rosdistro/blob/bfaaa5c587d42900befecda8b4cbb983028ce7af/kinetic/distribution.yaml#L2496 instead of removing the whole entry. The source entry allows it to be automatically generated into source checkouts.

tfoote:
If you want to disable CI please keep the source entry but turn off ci using test_commits: false like

See https://github.com/ros/rosdistro/pull/19299#pullrequestreview-166241624

tfoote (approved):

tfoote:
Thanks for iterating!

lennarthaller:
Thanks!

PR #19300, opened 2018-10-18
Added ipython-notebook

reinzor:
For analysis of bag data

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -114,6 +114,8 @@ intelhex-pip:
   ubuntu:
     pip:
       packages: [intelhex]
+ipython-notebook:
+  ubuntu: ipython-notebook
 ipython:
   debian: [ipython]
   fedora: [python-ipython]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -134,9 +134,14 @@ ipython:
     vivid: [ipython]
     wily: [ipython]
     xenial: [ipython]
-ipython-notebook:
-  debian: [ipython-notebook]
-  ubuntu: [ipython-notebook]
+jupyter-notebook:
+  debian:
+    '*': [jupyter-notebook]
+    jessie: [ipython-notebook]
+  ubuntu:
+    '*': [jupyter-notebook]
+    trusty: [ipython-notebook]
+    xenial: [ipython-notebook]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
This is not passing CI due to formatting. Please also add links for the package listed online. We also need at least a debian rule as well.

For guidance see https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

reinzor:
Added debian rule, cannot really figure out what is wrong with the formatting ..

tfoote:
It looks like the CI ran on the old commit again. Likely part of github's data consistency problems 
image

I've retriggered the build: https://status.github.com/messages

tfoote:
I looked up the packages in the database and noticed that these names are only on older distros:

https://packages.ubuntu.com/search?keywords=ipython-notebook
https://packages.debian.org/search?keywords=ipython-notebook

The project has been renamed to jupyter and has been released as `jupyter-notebook` on all the newer platforms.
https://packages.ubuntu.com/search?keywords=jupyter-notebook
https://packages.debian.org/sid/jupyter-notebook

To that end we should use that key and add backwards compatibility for the older distros like this:
```
jupyter-notebook:
  ubuntu:
    '*': [jupyter-notebook]
    trusty: [ipython-notebook]
    xenial: [ipython-notebook]
```

On debian the legacy platform would similarly be jessie.

tfoote (approved):
Thanks for the update, lgtm.

reinzor:
Can this be merged?

cottsay:
If this key specifically targeting the executables and not the python library, please add:
```yaml
  fedora: [python3-notebook]
```
per https://apps.fedoraproject.org/packages/python3-notebook

rosdep/python.yaml
```diff
@@ -134,6 +134,14 @@ ipython:
     vivid: [ipython]
     wily: [ipython]
     xenial: [ipython]
+jupyter-notebook:
+  debian:
+    '*': [jupyter-notebook]
+    jessie: [ipython-notebook]
+  ubuntu:
```
```suggestion
  fedora: [python3-notebook]
  ubuntu:
```

cottsay (changes requested):

cottsay (approved):

PR #19327, opened 2018-10-23
Change dependency to libqwt6 for Ubuntu Xenial

yotabits:
This change is made in order to use same Qt version (eg: qt4) than [artful bionic cosmic trusty] distros
https://packages.ubuntu.com/xenial/libqwt-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2983,7 +2983,7 @@ libqwt6:
     utopic: [libqwt-dev]
     vivid: [libqwt-dev]
     wily: [libqwt-qt5-dev]
-    xenial: [libqwt-qt5-dev]
+    xenial: [libqwt-dev]
 libqwtplot3d-qt4-dev:
   arch: [qwtplot3d]
   debian: [libqwtplot3d-qt4-dev]


tfoote (commented):
This looks like a more consistent version. One thing though is that wily should also probably be set to qt4  too.

I note that it's used by rqt_multiplot https://index.ros.org/d/libqwt6/ Can you look to see if that would be effected?

yotabits:
@tfoote unfortunately rqt_multiplot cannot be build on 16.04/Kinetic using `libqwt-dev` instead of `libqwt-qt5-dev`, it looks like we need a second key for solving my specific problem, something like `libqwt6-qt4`
What is you opinion ?

fyi @pgorczak

tfoote:
Since this key is qt4 on all other platforms I would suggest that this remain qt4 and create a new one for qt5 and convert rqt_multiplot to use the new qt5 version of the key.

This all assumes that they're side by side installable. (??)

yotabits:
@tfoote I agree, I will create an issue in rqt_multiplot.
I will check for side-by-side install.

pgorczak:
Side by side install works in bionic! I'll update my PR for bionic.

tfoote:
Since this and #19329 are adjusting the same key. I'm going to close this one and we can iterate on that one.

PR #19329, opened 2018-10-23
Add libqwt6 package for ubuntu/bionic

pgorczak:
Not sure about using *libqwt-dev* as suggested in #19327 since ROS melodic uses Qt5.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2975,7 +2975,7 @@ libqwt6:
   fedora: [qwt-devel]
   gentoo: ['x11-libs/qwt:6']
   ubuntu:
-    bionic: [libqwt-qt5-dev]
+    bionic: [libqwt-dev]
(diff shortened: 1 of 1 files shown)


tfoote:
Hmm this needs to be homogenized with #19327 do we need a second key? @yotabits fyi

pgorczak:
Changed for backwards compatibility.

tfoote (changes requested):
Thanks this looks good. Could you subsume #19327 switch to just use 'libqwt-dev' in a generic rule. Since all versions will be the same. It will make this much simpler overall.  And the qt5 version can be added as a separate key for side by side installation.

pgorczak:
Done that. Just linking back to 11593—that initially added the keys—for future reference.

tfoote (approved):

clalancette (approved):

PR #19348, opened 2018-10-25
Add rosdep rule for python-math3d

jonazpiazu:
Added a rosdep rule to resolve the [python-math3d](https://github.com/mortlind/pymath3d) package using [pip](https://pypi.org/project/math3d/) for installation.


tfoote (changes requested):
If this is pip only it should have the `-pip` extension.

jonazpiazu:
Fixed the `-pip` extension and squashed the commits.

tfoote (approved):

PR #19377, opened 2018-10-29
Add python-cbor package

mvollrath:


tfoote:
I see this for debian and ubuntu, could you link to the fedora and gentoo packages?

https://packages.debian.org/search?keywords=python-cbor
https://packages.ubuntu.com/search?keywords=python-cbor

mvollrath:
Fedora:
https://copr.fedorainfracloud.org/coprs/hvenev/experiments/package/python-cbor/
(I must admit I have no idea if this will install)

Gentoo:
https://packages.gentoo.org/packages/dev-python/cbor

tfoote (changes requested):
> Fedora:
https://copr.fedorainfracloud.org/coprs/hvenev/experiments/package/python-cbor/
(I must admit I have no idea if this will install)
Yeah that's what I found too. copr is the equivalent of ubuntu ppas so we shouldn't include that entry as a rule as it won't be installable by default.

mvollrath:
Don't intend to use this anymore.

dirk-thomas (approved):

mvollrath:
@tfoote I've excluded the uninstallable Fedora package.

dirk-thomas:
@ros/ros_team Needs a second review and merge.

tfoote (approved):

PR #19382, opened 2018-10-30
Add python-pytorch-pip to python.yaml

Tobias-Fischer:
This adds torch/torchvision to the `python.yaml` file. This is for example used by https://github.com/jsk-ros-pkg/jsk_recognition/pull/2041 and seeing the popularity of pyTorch probably in more repositories in the future. Since the old pull request for pyTorch (https://github.com/ros/rosdistro/pull/14398), the packages have become available in the standard pip.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2974,6 +2974,19 @@ python-pytides-pip:
   ubuntu:
     pip:
       packages: [pytides]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
With this in pypi's database this looks ok. Though we will definitely want to switch to an apt based dependency when that becomes available.

PR #19389, opened 2018-11-01
Add boto3-pip entry

r-meagher:
PyPI project URL: https://pypi.org/project/boto3/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -49,6 +49,13 @@ azure-iothub-device-client-pip:
   ubuntu:
     pip:
       packages: [azure-iothub-device-client]
+boto3-pip:
+  debian:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
We already have a boto3 rule https://packages.ubuntu.com/search?keywords=python-boto3 that this will conflict with.

murphm8:
Hi Tully, how can we depend on the latest version of boto3 in a ROS package? The version in Xenial (1.2.2) is far behind pip (1.9.35) and missing essential features.

tfoote:
If you need it you can locally create your own rosdep rules. But we don't want to distribute that to the community due to it potentially breaks anything that depends on the system packaged version.

PR #19412, opened 2018-11-06
Added python-scp

Vuwij:
For python to be able to secure copy

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3452,11 +3452,11 @@ python-scipy:
     yakkety_python3: [python3-scipy]
     zesty: [python-scipy]
     zesty_python3: [python3-scipy]
-python-scp:
+python-scp: 
   debian: [python-scp]
   fedora: [python-scp]
   ubuntu: [python-scp]
-python3-scp:
+python3-scp: 
   debian: [python3-scp]
   fedora: [python3-scp]
   ubuntu: [python3-scp]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3452,6 +3452,18 @@ python-scipy:
     yakkety_python3: [python3-scipy]
     zesty: [python-scipy]
     zesty_python3: [python3-scipy]
+python-scp:
+  debian: [python-scp]
+  fedora:
+    pip:
+      packages: [scp]
+  ubuntu:
+    xenial: [python-scp]
+    xenial_python3: [python3-scp]
+    yakkety: [python-scp]
+    yakkety_python3: [python3-scp]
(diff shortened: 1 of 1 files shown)


130s:
I think following [the contribution guideline](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md) would help review process. I'm citing what's the most relevant for rosdep key:

> # rosdep rules contributions
> :
> For convenience in reviewing please link to the web listings of packages such as on http://packages.ubuntu.com, http://packages.debian.org, and https://admin.fedoraproject.org/pkgdb/packages/ or if a pip package pypi.python.org.

You might want to list links for some distros you're adding the keys for. See https://github.com/ros/rosdistro/pull/19380 as an example.

rosdep/python.yaml
```diff
@@ -3452,6 +3452,18 @@ python-scipy:
     yakkety_python3: [python3-scipy]
     zesty: [python-scipy]
     zesty_python3: [python3-scipy]
+python-scp:
+  debian: [python-scp]
+  fedora:
+    pip:
+      packages: [scp]
+  ubuntu:
+    xenial: [python-scp]
```
tfoote: Since these are all the same they don't need to be called out by specific version.

rosdep/python.yaml
```diff
@@ -3452,6 +3452,18 @@ python-scipy:
...
+    xenial_python3: [python3-scp]
```
tfoote: If you want a python3 version please add another key `python3-scp` that will resolve to it. The `_python3` distro suffixes were an experiment that we aren't using on newer platforms.

tfoote (changes requested):
A few comments on formatting and as @130s mentioned links to the packages would be appreciated too.

rosdep/python.yaml
```diff
@@ -3452,6 +3452,14 @@ python-scipy:
     yakkety_python3: [python3-scipy]
     zesty: [python-scipy]
     zesty_python3: [python3-scipy]
+python-scp:
+  debian: [python-scp]
+  fedora: [python-scp]
+  ubuntu: [python-scp]
+python3-scp:
```
nuclearsandwich: Thanks for iterating @Vuwij. Since we keep keys sorted by lexical order, the python3 keys end up in another section of the file. Could you move them there?

tfoote: @Vuwij Please update this to be in alphabetical order so it will pass the linters, then we can merge it.

clalancette:
There's been no response on this for a while, so I'm going to close it out.  If you'd still like to get this in, feel free to reopen with the requested changes.

PR #19416, opened 2018-11-06
Index v4 test

dirk-thomas:
Follow up of ros/rosdistro#19415.

The v4 index is a copy-n-paste of the existing v3 index file plus the following two keys as well as version bumps:
* `distribution_status`
* `distribution_type`


wjwwood (approved):
I'm not sure about the context of the new yaml file, I assume it's correct and has the required fields.

The test lgtm.

PR #19444, opened 2018-11-11
python-imaging: 'python-imaging' is not available in Debian buster

kazuki0824:
AFAIK the package 'python-imaging' is deprecated and not supported in Debian/buster and future releases.
See also: https://packages.debian.org/ja/stretch/python-imaging

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1722,7 +1722,10 @@ python-hypothesis:
 python-imaging:
   alpine: [py-imaging]
   arch: [python2-pillow]
-  debian: [python-imaging]
+  debian:
+    buster: [python-pil]
+    stretch: [python-imaging]
+    jessie: [python-imaging]
   fedora:
     '21': [python-pillow, python-pillow-qt]
     '22': [python-pillow, python-pillow-qt]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Thanks this looks correct to switch to `python-pil`

https://packages.debian.org/buster/python-pil I see it comes from the same source package pillow: https://packages.debian.org/source/buster/pillow

Note that this is failing CI due to being out of alphabetical order. 

We should also update this for Ubuntu as python-imaging appears to only be valid through artful. https://packages.ubuntu.com/search?keywords=python-imaging

While python-pil is available on all platforms: https://packages.ubuntu.com/search?keywords=python-pil

To make things easier going forward we should use the wildcard notation for the current name and list all the older platforms explicitly with the old name. This will mean that we won't need to update the rule for each new Ubuntu or Debian release but can preserve the backwards compatibility for all existing platforms.

tfoote (approved):
I updated this to use the wildcard notation for forward compatibility and preserved the old behavior for jessie and stretch.

nuclearsandwich (approved):

PR #19477, opened 2018-11-15
Adding crane_x7_ros to documentation index for kinetic

Tiryoh:
I'd like to add crane_x7_ros for ROS Kinetic to be doc'd and indexed.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1776,7 +1776,7 @@ repositories:
       url: https://github.com/clearpathrobotics/cpr_multimaster_tools.git
       version: kinetic-devel
     status: developed
-  crane_x7_ros:
+  crane_x7:
     doc:
       type: git
       url: https://github.com/rt-net/crane_x7_ros.git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1776,6 +1776,12 @@ repositories:
       url: https://github.com/clearpathrobotics/cpr_multimaster_tools.git
       version: kinetic-devel
     status: developed
+  crane_x7_ros:
+    doc:
+      type: git
+      url: https://github.com/rt-net/crane_x7_ros.git
+      version: master
+    status: developed
   crazyflie:
     doc:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Quickly looking at this I noticed that you have ros in at least one of your package names which is redundant. `crane_x7_ros` Please see: http://wiki.ros.org/ROS/Patterns/Conventions and http://www.ros.org/reps/rep-0144.html

While I was there I looked at the license. And your package.xml files list Apache 2.0 as the license, but the LICENSE file lists a non-commercial license which is not Apache 2.0 these should be consistent.

rt-net:
Dear Mr. Tfoote

Thanks for your your note pointing out that packages should named in compliance with ROS naming conventions. We will rename the files accordingly. 

As for the non-commercial license agreement, it does incorporate by reference all licenses relevant to open source materials including Apache 2.0, which is set out in the manner prescribed by the Apache Software Foundation (http://www.apache.org/licenses/LICENSE-2.0). We appreciate if you could elaborate on your earlier comment about inconsistency, as we will be amending as required.

Thanks and best regards.

tfoote (commented):
I just noticed that the declared licenses in the package.xml was inconsistent with what the licenses seemed to be in the code. The updates like [this](https://github.com/rt-net/crane_x7_ros/commit/7003c04a569e25fbc7c2b9bce9f66b676d052f66) resolve that.

The one package using ros in the name is [here](https://github.com/rt-net/crane_x7_ros/blob/0dd24c1bdd901954811f6f7df90050996c841b5f/crane_x7_ros/package.xml#L2)
Please comment back here when it's resolved and we can add it.

Tiryoh:
Dear @tfoote,
Thank you for pointing them out.
We've updated the package name.  Would you please check it out?
https://github.com/rt-net/crane_x7_ros/commit/f576354f9c17f8f8073e1a0e6ea71d910f7d5170

tfoote (approved):
Thanks for the updates. It looks good to me now.

PR #19478, opened 2018-11-15
rosdep: added dependency 'python-grpc-tools' for gRPC code generating plugin

atiderko:
For current debian and ubuntu versions this is a pip dependency `grpcio-tools`.
For buster (debian) and cosmic (ubuntu):
https://packages.debian.org/source/buster/python-grpc-tools
https://packages.ubuntu.com/source/cosmic/python-grpc-tools


nuclearsandwich (approved):
Since this follows the same conventions as #19292 it looks good to me.

tfoote:
Thanks for the complete PR following the same approach as grpcio.

PR #19505, opened 2018-11-20
add python3-lark-parser key for Ubuntu

dirk-thomas:
Required for ros2/rosidl#326.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4631,6 +4631,8 @@ python3-gitlab:
     xenial:
       pip:
         packages: [python-gitlab]
+python3-lark-parser:
+  ubuntu: [python3-lark-parser]
 python3-mock:
   debian: [python3-mock]
   fedora: [python3-mock]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -4631,6 +4631,8 @@ python3-gitlab:
     xenial:
       pip:
         packages: [python-gitlab]
+python3-lark-parser:
+  ubuntu: [python3-lark-parser]
```
nuclearsandwich:
```suggestion
  ubuntu:
    bionic: [python3-lark-parser]
```

This package is only available in bionic.

nuclearsandwich (changes requested):
The package isn't available for all supported ubuntu versions and so cannot elide the distro codename.

rosdep/python.yaml
dirk-thomas: @tfoote suggested to add it as is. I don't mind either or.

nuclearsandwich (approved):

PR #19515, opened 2018-11-21
Update rosdep rule guidelines for python 3 style.

nuclearsandwich:
We've had a handful of contributions come in using the `_python3`-suffixed distro codenames rather than the separate python3 rosdep keys. This updates our rosdep contribution guidelines with our current preferred python3 style.

[:scroll: Rendered preview](https://github.com/nuclearsandwich/rosdistro/blob/3d1e919cdc3f8e4d2301d724d3ceaa9fe56401cc/CONTRIBUTING.md#python-3-rules) of the modified guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,26 @@ Guidelines for rosdep rules
        However please don't target 'sid' as it's a rolling target and when the keys change our database gets out of date.
   * Keep everything in alphabetical order for better merging.
   * No trailing whitespace.
-   
+
+### Python 3 rules
+
+When adding rules for python 3 packages, create a separate entry prefixed with `python3-` rather than `python`
+For example:
+
+```yaml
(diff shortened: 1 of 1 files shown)


tfoote (approved):

ipa-lth:
Hi, here my personal short feedback:
I welcome having two separated entry for python2 and python3, however the alphabetical ordering separates these two entries from each other quite a lot. Having the entries after each other may lead to better coverage, since the editor does not need to look for it. Maybe we should rethink the rules for ordering.

nuclearsandwich:
Thanks for the feedback @ipa-lth I've merged as is since updating the linter and desired format is a bit outside the scope of improving the documentation for the current desired format. But I appreciate the suggestion. I think making a special case for ordering python3 keys could be more confusing than immediately helpful but we'll discuss it with the rest of the ROS team.

PR #19519, opened 2018-11-22
BVHBroadcaster to documentation index for distro

mingfeisun:


tfoote (changes requested):
Please update your package to follow our naming conventions. In particular it's using uppercase letters at the moment.

http://www.ros.org/reps/rep-0144.html

mingfeisun:
sorry it's my mistake. Thanks for pointing out. I just rename it according to the naming rule

tfoote (approved):
thanks for updating the name!

PR #19523, opened 2018-11-23
added bionic to python list of python-pymongo

frederico-klein:
Single line added to support bionic.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3925,6 +3925,7 @@ mongodb-dev:
   gentoo: [dev-db/mongodb]
   macports: [mongodb]
   ubuntu:
+    bionic: [libmongoclient-dev, libmongoclient0, mongodb-dev]
     lucid: [libmongo-client-dev, mongodb-dev]
     maverick: [libmongo-client-dev, mongodb-dev]
     natty: [libmongo-client-dev, mongodb-dev]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2845,6 +2845,7 @@ python-pymongo:
     pip:
       packages: [pymongo]
   ubuntu:
+    bionic: [python-pymongo]
     lucid: [python-pymongo]
     maverick: [python-pymongo]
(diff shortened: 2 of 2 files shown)


frederico-klein:
added mongodb-dev as well

tfoote (changes requested):
Please link to the packages that you're adding and since they're different packages demonstrate how they're providing the same functionality.

Related it would be good to fill in the other versions inbetween xenial and bionic too.

frederico-klein:
Actually I suppose it adds a bit more functionality because I am also including https://packages.debian.org/stretch/libmongoclient0 which is the  mongo-cxx-driver-legacy (see https://github.com/ros-planning/warehouse_ros_mongo ) 
A bit of googling showed me that this is only available from bionic onwards https://packages.ubuntu.com/bionic/libmongoclient-dev (bionic, cosmic, disco), so I suppose this only makes things easier for newer distros. 
For yakkety, zesty and artful the lines should probably be the same as xenial, but I haven't tested it

frederico-klein:
Oh, https://packages.ubuntu.com/bionic/libdevel/python-pymongo seems it would be alright, but mongodb-dev mongodb-dev probably would not be so straightforward...

tfoote:
My point is that 
https://packages.ubuntu.com/bionic/libmongoclient-dev
and 
https://packages.ubuntu.com/bionic/libmongo-client-dev

Appear to be completely different projects thus a package that depends on one will not be expected to compile or work against the other.

I see that https://github.com/algernon/libmongo-client is obsolete and likely should be transitioned off of, but the new replacement that's not a drop in replacement should be a new key not an evolution of the current key. Especially since the current key is still valid and is released as far forward as cosmic https://packages.ubuntu.com/cosmic/libmongo-client-dev

tfoote:
Since this is a new/different package we can add it but we should use a different key to make sure that people don't expect the same content. I'm going to close this in it's current form. Please update the branch and we can reopen this or create a new PR.

PR #19527, opened 2018-11-26
Releasing packages for integrating with Amazon Web Services

AAlon:


tfoote (commented):
The source entries require a branch name and not a tag due to vagaries of the Jenkins git plugin we use. And in general we actually recommend running documentation off of the development branch as well. This gives you the affordance to update documentation and not be required to make a release before the documentation online is updated. 

Looking through your repositories it looks like most of those should just be master for now.

PR #19546, opened 2018-11-27
Add rosdep keys for libopensplice69.

nuclearsandwich:
@allenh1 I took the gentoo key out because I couldn't find any opensplice packages in sci-libs, let alone tone for 6.9. Is there a separate overlay where the gentoo packages live?

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2492,6 +2492,10 @@ libopensplice67:
   ubuntu:
     bionic: [libopensplice67]
     xenial: [libopensplice67]
+libopensplice69:
+  ubuntu:
+    bionic: [libopensplice69]
+    xenial: [libopensplice69]
 libopenvdb:
   arch: [openvdb]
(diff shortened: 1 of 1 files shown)


dirk-thomas (approved):

allenh1:
> Is there a separate overlay where the gentoo packages live?

Yes, they live in the ROS-Overlay: https://github.com/ros/ros-overlay/tree/master/sci-libs/opensplice.

nuclearsandwich:
@allenh1 thanks for the link. If you're interested in creating an updated ebuild for OpenSplice 6.9 that would help a lot for Crystal on Gentoo. Otherwise I can try to test it out in a container or VM but I won't have work time to dedicate to it and my personal time is stretched around the upcoming winter holidays.

allenh1:
@nuclearsandwich I'll do that when I get a few. It should _just work_ when I change the version number, but... Not sure.

PR #19565, opened 2018-11-30
wall_follow package

engr-sas:
Creating a new public package


clalancette:
Thanks for your contribution, but we don't support or build for hydro anymore.  If you'd like to submit your package for one of our [currently supported releases](https://wiki.ros.org/Distributions), however, we would be glad to consider it.

PR #19572, opened 2018-12-01
Adding wall_follow to documentation index for kinetic  

engr-sas:
please add wall_follow to be indexed on ros.org


clalancette:
The CI failed here because this references release 0.1.0, but the release repository (https://github.com/engr-sas/wall_follow-release) only has 0.0.1 available.  I'm not exactly sure which one you mean, so please update either the release or the metadata here.

For future reference, it is easier to allow `catkin_prepare_release`/`bloom` to do all this work for you; see http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease for instructions on how to do that.

PR #19598, opened 2018-12-04
rosdep: python-msgpack support for ubuntu bionic

garyservin:
Add python-msgpack support for ubuntu bionic


tfoote (approved):
I simplified the rule since python-msgpack is around in all distros. Following new python3 formulas https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-3-rules

I expect it to be stable similar to the debian version. https://packages.ubuntu.com/bionic/python-msgpack

The python3 version can be contributed if it's wanted in the future.

PR #19600, opened 2018-12-04
Add bionic key for python-qt-bindings

pjreed:
https://packages.ubuntu.com/bionic/libpyside-dev
https://packages.ubuntu.com/bionic/libshiboken-dev
https://packages.ubuntu.com/bionic/python-pyside
https://packages.ubuntu.com/bionic/python-qt4
https://packages.ubuntu.com/bionic/python-qt4-dev
https://packages.ubuntu.com/bionic/python-sip-dev
https://packages.ubuntu.com/bionic/shiboken


nuclearsandwich (approved):

tfoote (changes requested):
It would be great to use the wildcard syntax to get all the recent versions and potentially future versions for this change too. See #19614 for a recent discussion and syntax so we can get better coverage from this rule.

pjreed:
Is there anything else that needs to be done for this and #19607?

tfoote (approved):

PR #19607, opened 2018-12-05
Add bionic key for python-alembic

pjreed:
https://packages.ubuntu.com/bionic/python-alembic


nuclearsandwich (approved):

tfoote (changes requested):
This would be a good candidate for the '*' syntax like the discussion in #19614 with the exception being trusty. Then it will work for the intermediate platforms and future platforms too.

tfoote (approved):

PR #19614, opened 2018-12-05
Add keys for libsystemd-dev

mleclair73:
- debian: https://packages.debian.org/stretch/libsystemd-dev
- ubuntu: https://packages.ubuntu.com/bionic/libsystemd-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3133,6 +3133,9 @@ libswscale-dev:
   fedora: [ffmpeg-devel]
   gentoo: [virtual/ffmpeg]
   ubuntu: [libswscale-dev]
+libsystemd-dev:
+  debian: [libsystemd-dev]
+  ubuntu: [libsystemd-dev]
 libtclap-dev:
   arch: [tclap]
   debian: [libtclap-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3135,7 +3135,9 @@ libswscale-dev:
   ubuntu: [libswscale-dev]
 libsystemd-dev:
   debian: [libsystemd-dev]
-  ubuntu: [libsystemd-dev]
+  ubuntu:
+    '*': [libsystemd-dev]
+    trusty: null
 libtclap-dev:
   arch: [tclap]
   debian: [libtclap-dev]


rosdep/base.yaml
```diff
@@ -3133,6 +3133,9 @@ libswscale-dev:
   fedora: [ffmpeg-devel]
   gentoo: [virtual/ffmpeg]
   ubuntu: [libswscale-dev]
+libsystemd-dev:
+  debian: [libsystemd-dev]
+  ubuntu: [libsystemd-dev]
```
nuclearsandwich:
```suggestion
  ubuntu:
    artful: [libsystemd-dev]
    bionic: [libsystemd-dev]
    cosmic: [libsystemd-dev]
    xenial: [libsystemd-dev]
```

This package isn't available in Trusty and so can't use a shorthand ubuntu rule. @tfoote is there a more compact representation for this?

nuclearsandwich (changes requested):
Thanks for contributing. This package isn't available in Ubuntu Trusty and so it can't use the shorthand ubuntu rule. I've suggested an edit which I think is the most compact correct expansion.

rosdep/base.yaml
JWhitleyAStuff: From entries for other projects, it looks like the format would be:
```suggestion
  ubuntu:
    '*': [libsystemd-dev]
    trusty: null
```
See `xulrunner-1.9.2` or `libglfw3-dev` as examples.

tfoote: It's a balancing act. For adding new rules we sometimes let it slide that we add a generic rule that won't be supported on older platforms. (aka it will incorrectly resolve) Since it's a new key, we know that it won't break anyone on the older system as they cannot already be using it. If they do try to use it on an older system it will fail at install time. Trusty is just on the edge, it's still officially supported but only for another 5 months. But since the wildcard technique is relatively compact i'd suggest going with that as @JWhitleyAStuff suggests. We can then drop the trusty rule in a purge at some later date.

mleclair73: Thanks for the feedback. Sounds good!

nuclearsandwich (approved):

PR #19625, opened 2018-12-07
the python docker api

ct2034:
I need it to automate docker containers in https://github.com/ct2034/dockeros
https://pypi.org/project/docker/

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -14194,7 +14194,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ros-drivers-gbp/velodyne-release.git
-      version: 1.5.0-0
+      version: 1.5.1-0
     source:
       type: git
       url: https://github.com/ros-drivers/velodyne.git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2253,11 +2253,6 @@ repositories:
       type: git
       url: https://github.com/ct2034/dockeROS.git
       version: master
-    release:
-      tags:
-        release: release/kinetic/{package}/{version}
-      url: https://github.com/ct2034/dockeros-git.git
-      version: 1.0.0-0
     source:
       type: git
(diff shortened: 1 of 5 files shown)


ct2034:
Okay, I added the release of my package now, too.

kinetic/distribution.yaml
```diff
@@ -2248,6 +2248,21 @@ repositories:
...
     status: developed
+  dockeros:
+    doc:
+      type: git
+      url: https://github.com/ct2034/dockeROS.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/ct2034/dockeros-git.git
```
tfoote: This url is incorrect. I found a release repo at dockeros-release.git but that repository does not appear to have had any pushes from running bloom-release so even with the updated url it would still fail to pass CI.

tfoote (changes requested):
I'd suggest rolling back your last commit and letting bloom automatically add the release section when you run it: http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

ct2034:
Sorry, this is now only about the added python library

kinetic/distribution.yaml
ct2034: Sorry, this is now only about the added python library

ct2034:
@tfoote Can you please review again

tfoote:
As there's a well established python-docker package we should use that not the pip version: https://packages.ubuntu.com/search?keywords=python-docker

fedora: python-docker-py https://apps.fedoraproject.org/packages/python-docker-py
debian: https://packages.debian.org/search?keywords=python-docker

ct2034:
ok, these are the ones:
arch: https://www.archlinux.org/packages/community/any/python-docker/
fedora: https://apps.fedoraproject.org/packages/python-docker/
ubuntu: https://packages.ubuntu.com/xenial/python-docker
debian: https://packages.debian.org/jessie/python-docker

ct2034:
I'll reopen ...

PR #19684, opened 2018-12-11
Add python3-lark-parser for gentoo

allenh1:
Added ebuild to ROS Overlay in commit 98f9546c61f6d14a49efa3f66c2960d7f4f9d5c2.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4619,6 +4619,7 @@ python3-lark-parser:
   fedora:
     '*': [python-lark-parser]
     '28': null
+  gentoo: [dev-python/lark]
   ubuntu:
     bionic: [python3-lark-parser]
 python3-matplotlib:


nuclearsandwich:
@allenh1 did you catch that we need a specific pre-released version is required https://github.com/ros2/ci/pull/220#issuecomment-439213675 ?

The key looks good to me as long as the ebuild contains a sufficient version.

allenh1:
> did you catch that we need a specific pre-released version?

Uh -- nope. No I did not. I'll definitely update my ebuild in the near future to link to that version.

Thanks for catching that!

nuclearsandwich (approved):
Approved with the caveat that until the ebuild is updated this key won't be sufficient to enable ROS 2 on Gentoo. I'll leave it to additional reviewers and @allenh1 to decide if we should wait to merge until the ebuild is updated.

allenh1:
I don't think it's necessary to wait, since the key will still be the same. There's a long way to go yet wrt ROS 2 on gentoo.

PR #19696, opened 2018-12-12
rotors_simulator: 2.2.3-0 in 'melodic/distribution.yaml' [bloom]

michaelpantic:
Increasing version of package(s) in repository `rotors_simulator` to `2.2.3-0`:

- upstream repository: https://github.com/ethz-asl/rotors_simulator.git
- release repository: https://github.com/ethz-asl/rotors_simulator-release.git
- bloom version: `0.6.9`

## rotors_comm

- No changes

## rotors_control

- No changes

## rotors_description

- No changes

## rotors_evaluation

- No changes

## rotors_gazebo

- No changes

## rotors_gazebo_plugins

```
* Added dependencies to gazebo_plugins
* Fixes issues with build on buildfarm and old ubuntu systems (added MAVROS dependencies, version checking, protobuf-dev etc)
* Contributors: michaelpantic
```

## rotors_hil_interface

```
* Merge branch 'feature/hil_mavros_fix' into feature/gazebo9-autobackport
* fix not including message types
* Contributors: Zachary Taylor, michaelpantic
```

## rotors_joy_interface

- No changes

## rotors_simulator

- No changes

## rqt_rotors

- No changes

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4639,7 +4639,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ethz-asl/rotors_simulator.git
-      version: 2.2.2
+      version: master
     release:
       packages:
       - rotors_comm
@@ -4655,11 +4655,11 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/ethz-asl/rotors_simulator-release.git
-      version: 2.2.2-0
+      version: 2.2.3-0
     source:
       type: git
       url: https://github.com/ethz-asl/rotors_simulator.git
-      version: 2.2.2
+      version: master
   rplidar_ros:
     doc:
       type: git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4639,7 +4639,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ethz-asl/rotors_simulator.git
-      version: master
+      version: feature/gazebo9-autobackport
     release:
       packages:
       - rotors_comm
@@ -4659,7 +4659,7 @@ repositories:
     source:
       type: git
       url: https://github.com/ethz-asl/rotors_simulator.git
-      version: master
+      version: feature/gazebo9-autobackport
   rplidar_ros:
     doc:
       type: git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4635,6 +4635,31 @@ repositories:
       url: https://github.com/ros-drivers/rosserial.git
       version: melodic-devel
     status: maintained
+  rotors_simulator:
+    doc:
+      type: git
+      url: https://github.com/ethz-asl/rotors_simulator.git
+      version: 2.2.2
+    release:
+      packages:
+      - rotors_comm
+      - rotors_control
+      - rotors_description
+      - rotors_evaluation
+      - rotors_gazebo
+      - rotors_gazebo_plugins
+      - rotors_hil_interface
+      - rotors_joy_interface
+      - rotors_simulator
+      - rqt_rotors
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/ethz-asl/rotors_simulator-release.git
+      version: 2.2.2-0
+    source:
+      type: git
+      url: https://github.com/ethz-asl/rotors_simulator.git
+      version: 2.2.2
   rplidar_ros:
     doc:
       type: git


michaelpantic:
CI error:

> ERROR: In file '''/home/travis/build/ros/rosdistro/melodic/distribution.yaml''': Could not validate source entry for repo rotors_simulator with error [[[Could not validate repository with url https://github.com/ethz-asl/rotors_simulator.git and version 2.2.2 from entry at line 4659]]]

Repository and version are correct - do  I need to add it somewhere else?

clalancette:
> Repository and version are correct - do I need to add it somewhere else?

Hm, a brief look shows it to be right.  There were some Github issues overnight elsewhere; I'm going to close this PR and reopen to re-kick travis and see if that fixes it.

melodic/distribution.yaml
```diff
@@ -4635,6 +4635,31 @@ repositories:
...
+      - rotors_simulator
+      - rqt_rotors
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/ethz-asl/rotors_simulator-release.git
+      version: 2.2.2-0
+    source:
+      type: git
+      url: https://github.com/ethz-asl/rotors_simulator.git
+      version: 2.2.2
```
clalancette: Oh, right.  The problem is that we don't generally consider tags to be valid for source/doc entries, only branches.  I'll suggest you retarget this to master.

clalancette (dismissed):

melodic/distribution.yaml
tfoote: I believe that doc is actually ok with a tag but is not recommended, but source does need a branch.

michaelpantic:
@tfoote @clalancette 
Will update this as soon package is building for indigo (to iron out potential build issues and not have them  on multiple tracks simultaneously) - will leave PR open

michaelpantic:
@clalancette @tfoote 
Updated this PR. Currently our Gazebo9 version is still sitting in a branch, thus the confusion with the tags/branches.

FYI We're currently ironing out a lot of build-related issues with compatibility with different Gazebo versions. If all goes well, we sync all the indigo/kinetic/releases soon and bump version to 3.1.0 before the end of the year for all distros.

melodic/distribution.yaml
michaelpantic: fixed - thanks much for your hint!

clalancette:
> FYI We're currently ironing out a lot of build-related issues with compatibility with different Gazebo versions. If all goes well, we sync all the indigo/kinetic/releases soon and bump version to 3.1.0 before the end of the year for all distros. (for some reason someone once released a 3.0.0 for kinetic, so we'll have to jump a few versions ;-) )

@michaelpantic OK, that's fine.  Should we merge this PR (at 2.2.3) in the meantime, or should we just wait for the new version?  Either is fine with me.

michaelpantic:
@clalancette  Yes would be nice to merge this one, just to make sure that everything works right now.

michaelpantic:
@clalancette 
oh, build failed, see http://build.ros.org/job/Mbin_ds_dS64__rotors_gazebo_plugins__debian_stretch_amd64__binary/1/console

Any idea why it tries to install gazebo7? I expected it to  use gazebo9 which is default in bionic w/melodic. 
Also rosdep specifies gazebo9 for key gazebo.

tfoote:
The build that failed is on stretch which is gazebo7 by default.

michaelpantic:
@tfoote 
The ARM build also fails with an error of the underlying ignition transport library when checking SDF files with gz sdf (see http://build.ros.org/job/Mbin_ubhf_uBhf__rotors_gazebo__ubuntu_bionic_armhf__binary/8/console)

Command
```cd /tmp/binarydeb/ros-melodic-rotors-gazebo-2.2.3 && gz sdf -p /tmp/binarydeb/ros-melodic-rotors-gazebo-2.2.3/models/rotors_description/urdf/iris_base.urdf >> /tmp/binarydeb/ros-melodic-rotors-gazebo-2.2.3/models/iris/iris.sdf```

fails with 

```
09:43:54 Unsupported setsockopt level=0 optname=32
09:43:54 Error setting socket option (IP_MULTICAST_IF).
09:43:54 Unsupported setsockopt level=0 optname=32
09:43:54 Error setting socket option (IP_MULTICAST_IF).
09:43:54 Did you set the environment variable IGN_IP with a correct IP address? 
09:43:54   [172.30.1.33] seems an invalid local IP address.
09:43:54   Using 127.0.0.1 as hostname.
09:43:54 Unsupported setsockopt level=0 optname=32
09:43:54 Error setting socket option (IP_MULTICAST_IF).
09:43:54 terminate called after throwing an instance of 'std::out_of_range'
09:43:54   what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
```

which seems to be a known issue in ignition when used in Linux for Windows subsystem, but the cause might be similar on ARM Linux Kernels (not being aple to set IP_MULTICAST_IF on that socket - because it seems to be the wrong socket, not local loopback). See also https://bitbucket.org/osrf/gazebo/issues/2351/gazebo-issues-on-windows-subsystem-for

But I guess that is something build-system related that doesn't have much to do with our code, or whats your take on this?

In any case, thanks much for your support and merry christmas ;-)

michaelpantic:
@ffurrer  FYI

tfoote:
The gazebo installation on Stretch issue us that you need to use the straight gazebo version. Gazebo9 is not packaged for stretch. I've submitted a PR to remove the invalid alias from the rosdep database: https://github.com/ros/rosdistro/pull/19843 you should use the default `gazebo` depdency not specifically `gazebo9`. 

---

The socket issue is likely related to our builds being run inside of docker with limited permissions. There might be something to do to make ignition-transport more robust to that, but in this case it looks like it shouldn't even be opening a socket if it's just converting a urdf to sdf. 

@caguero Where's the right place to report this on gazebo?

caguero:
Probably here:

https://bitbucket.org/osrf/gazebo9-release/issues?status=new&status=open

@j-rivero

tfoote:
Thanks ticketed at: https://bitbucket.org/osrf/gazebo9-release/issues/2/gz-sdf-converter-fails-due-to-socket

michaelpantic:
@tfoote 
Thanks for having a look at these things.

Concerning the gazebo vs gazebo9 dependency:
As far as I am aware, we do not have any "gazebo9" dependency - neither in CMakeLists.txt nor package.xml. Our package only depends on "gazebo", "gazebo_plugins" and "gazebo_ros". 

The "gazebo" dependency results in trying to install gazebo7 while "gazebo_ros" resolves to 
"ros-melodic-gazebo-dev" which depends on gazebo9.

AFAIK we need both dependencies to be compatible on other versions, so there's no way we can resolve this for debian stretch currently, correct?

tfoote:
looking at [REP 3](http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023) it looks like we've explicitly chosen to backport gazebo9 for stretch. But I guess that didn't get imported into the repositories. I suspect that this is because gazebo7 was still specified for stretch in [Lunar](http://www.ros.org/reps/rep-0003.html#lunar-loggerhead-may-2017-may-2019) which included stretch. 

@clalancette Were you involved with the discussions with @mikaelarguedas around release time for Melodic?

mikaelarguedas:
@tfoote I replied here with my recollection of the discussion: https://github.com/ros/rosdistro/pull/19843#issuecomment-452395275


> AFAIK we need both dependencies to be compatible on other systems

@michaelpantic as both `gazebo_ros` and `gazebo_plugins` export the dependency on gazebo_dev, you should be able to remove the dependency on `gazebo` in your package.xml and it should be compatible with all ROS versions. Do you know on what systems removing the `gazebo` dependency poses problems ?

clalancette:
@mikaelarguedas I had forgotten about that discussion, but you are totally right that that should fix it.  Thanks for the comment!

PR #19712, opened 2018-12-13
Adding iirob_filters to documentation index for distro

GDwag:


tfoote:
This is replaced by #19713 please try not to submit overlapping changes, or if you do close out the older ones so we don't try to review them. 

Also this looks like you have a bunch of unrelated changes that it would be better not to have in the history.

PR #19715, opened 2018-12-13
Added canopen-pip.

tonybaltovski:
https://pypi.org/project/canopen/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -49,6 +49,16 @@ azure-iothub-device-client-pip:
   ubuntu:
     pip:
       packages: [azure-iothub-device-client]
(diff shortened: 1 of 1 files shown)


tfoote:
This a ros package with this inside how does that compare? Are there potential conflicts?

http://wiki.ros.org/ros_canopen
@ipa-mdl

ipa-mdl:
@tfoote: I don't see any conflicts. The pip package provides generic CANopen support in python. `ros_canopen` targets C++ and provides ROS bridges.

tfoote (approved):
Since there doesn't seem to be any upstream packaging yet we can add the pip rule.

PR #19721, opened 2018-12-14
Updating rr_openrover_basic package

shoemakerlevy9:
Adding package to melodic, and updating release repo for kinetic

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -11925,7 +11925,7 @@ repositories:
     source:
       type: git
       url: https://github.com/RoverRobotics/rr_openrover_basic.git
-      version: master
+      version: kinetic-devel
     status: developed
     status_description: pre-release-version
   rr_swiftnav_piksi:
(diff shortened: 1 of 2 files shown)


clalancette:
Thanks for the contribution.

It will be much better if you split this into two separate pull requests.  The basic reason is that the sync schedules for Kinetic and Melodic are not synced, so it may be the case that this gets held up because of one or the other of them.  The other reason is that it is good to keep the PRs confined to a single thing so we can easily look at the history and/or revert later if needed.

Additionally, this PR looks like it failed CI.  It failed because the release repository at https://github.com/RoverRobotics-release/rr_openrover_basic-release.git isn't set up properly.  In order to setup that release repository, you'll have to follow the steps at http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease .  I'm going to close this PR for now, but once you have fixed these things, please open new PRs (bloom will do this for you).

shoemakerlevy9:
Will do, thanks for the advice

PR #19728, opened 2018-12-14
Adding to Gentoo qtbase5-dev keys widgets & qttest

awesomebytes:
As arised in https://github.com/ros/ros-overlay/issues/725 `qt_gui_cpp` expects Qt to have the widgets and qttest modules.
But in Gentoo `dev-qt/qtcore` does not come with much, we need to add the other two keys to have all dependencies satisfied automatically.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4415,7 +4415,7 @@ qtbase5-dev:
   debian: [qtbase5-dev]
   fedora: [qt5-qtbase-devel]
   freebsd: [qt5-core]
-  gentoo: ['dev-qt/qtcore:5']
+  gentoo: ['dev-qt/qtcore:5', 'dev-qt/widgets:5', 'dev-qt/qttest:5']
   opensuse: [libqt5-qtbase-common-devel, libqt5-qtbase-devel]
   slackware: [qt5]
   ubuntu: [qtbase5-dev]


nuclearsandwich (approved):

tfoote (approved):

rosdep/base.yaml
```diff
@@ -4415,7 +4415,7 @@ qtbase5-dev:
   debian: [qtbase5-dev]
   fedora: [qt5-qtbase-devel]
   freebsd: [qt5-core]
-  gentoo: ['dev-qt/qtcore:5']
+  gentoo: ['dev-qt/qtcore:5', 'dev-qt/widgets:5', 'dev-qt/qttest:5']
```
allenh1: This should be `dev-qt/qtwidgets:5`, and I don't think `dev-qt/qttest:5` is needed unless they have the `test` `USE` flag enabled...

@awesomebytes Could you please fix this typo?

awesomebytes: @allenh1 good catch on the typo. But about qttest, it is actually needed even with the disabled test USE flag by higher level packages (as I arised in this issue: https://github.com/ros/ros-overlay/issues/592). It is just expected to be there in qt by further ROS packages as the other OS install it all together.

awesomebytes: PR about the typo: https://github.com/ros/rosdistro/pull/19791
Sorry about it.

PR #19748, opened 2018-12-17
 Adding ati_force_torque, force_torque_sensor and ipr_extern to melodic documentation

GDwag:


clalancette:
This looks like it is removing the release for kinetic, and then adding source entries for Melodic.  I'm not entirely sure what you intended, but that will tend to break things for Kinetic users.  If you just want to add doc to Melodic, please rebase, squash, and make this PR only modify the melodic distributions.yaml file.  Thanks.

PR #19776, opened 2018-12-18
Module venv is baked into python3 in fedora

paulbovbel:
Only ubuntu/debian requires installing separate python3-venv package.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4779,6 +4779,7 @@ python3-termcolor:
   ubuntu: [python3-termcolor]
 python3-venv:
   debian: [python3-venv]
+  fedora: []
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Yeah, it's odd that debian actually splits it out of python as it's actually part of python3's standard library.

cottsay (approved):
I would *rather* see this key still require the package that supplies the files, even if it seems improbable that it couldn't *not* be installed. It could save someone some googling later. However, the change you proposed should work.

The actual package that supplies the module is `python3-libs`.

paulbovbel:
Sounds reasonable to me

cottsay (approved):

PR #19779, opened 2018-12-18
Add virtualenv key

paulbovbel:
ubuntu: https://packages.ubuntu.com/xenial/virtualenv
debian: https://packages.debian.org/stretch/virtualenv

fedora is a weird one: it ships the `/usr/bin/virtualenv` script as part of python3-virtualenv, verified on a clean docker container. Not sure if there's a preferred approach...

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4821,6 +4821,10 @@ tilestache:
   debian: [tilestache]
   fedora: [python-tilestache]
   ubuntu: [tilestache]
+virtualenv:
+  debian: [virtualenv]
+  fedora: [python3-virtualenv]
+  ubuntu: [virtualenv]  
 wxpython:
   arch: [wxpython]
   centos: [wxPython-devel]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4823,8 +4823,8 @@ tilestache:
   ubuntu: [tilestache]
 virtualenv:
   debian: [virtualenv]
-  fedora: [python3-virtualenv]
-  ubuntu: [virtualenv]  
+  fedora: [virtualenv]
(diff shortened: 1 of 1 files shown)


tfoote:
The ubuntu package is also just a thin wrapper to python3 + python3-virtualenv Are you looking for python2 or python3 usage with this key? It think it might make sense to call out the python2 and python3 keys for clarity instead of this odd redirector package: https://packages.ubuntu.com/xenial/python-virtualenv and https://packages.ubuntu.com/xenial/python3-virtualenv though I would also strongly recommend using the python3 standard `venv` instead of virtualenv for python3 applications.

The virtualenv space is a bit of a mess, but there's a good summary here:
https://stackoverflow.com/questions/41573587/what-is-the-difference-between-venv-pyvenv-pyenv-virtualenv-virtualenvwrappe

paulbovbel:
I'm looking specifically for the `virtualenv` command line utility, which is only available via the `virtualenv` package. This package normally gets pulled in as a recommends of python-virtualenv, unless of course, you're using the --no-install-recommends flag in apt...

I'm kind of torn about that one because we've started using it in our docker-based builds since it's considered 'good practice'. So far, lack of `virtualenv` is the only thing that is explicitly broken.

tfoote (approved):
Since we have `python3-venv` in #19776 I'd suggest we make this `python-virtualenv` since it's specifically python2. But if the command line tool is actually in the virtualenv package we shouldn't rely on recommends and I think we need to have the actual package then. It's a pain that it hard depends on python3, but I guess this is how they're breaking their dependency cycle.

So I guess that's a long winded way to say lgtm.

cottsay (approved):
Glad to see the virtual package support in Fedora put to use.

PR #19807, opened 2018-12-20
Added bionic rosdep for python-pymongo

hawesie:
Needed for melodic release of mongodb-store


tfoote (approved):
I switched this to use wildcard syntax for better future compatibility and coverage.

cottsay (approved):

PR #19823, opened 2018-12-21
Remove flexbe_app version number from distros

pschillinger:
Sorry for the back-and-forth! Jenkins is failing for some reason which I cannot really explain, thus the flexbe_app entry should better be removed for now.

Both the prerelease test and my own CI worked without errors and I can also locally create a debian package using `bloom-generate rosdebian --ros-distro melodic` and `fakeroot debian/rules binary` which I then installed and used successfully for several days. So I'm not sure what causes the Jenkins fail...

Thank you!

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -3635,21 +3635,6 @@ repositories:
       url: https://github.com/team-vigir/flexbe_behavior_engine.git
       version: master
     status: developed
-  flexbe_app:
-    doc:
-      type: git
-      url: https://github.com/FlexBE/flexbe_app.git
-      version: master
-    release:
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/FlexBE/flexbe_app-release.git
-      version: 2.1.3-0
-    source:
-      type: git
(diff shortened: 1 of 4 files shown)


tfoote:
It looks like your devel jobs are running: http://build.ros.org/job/Mdev__flexbe_app__ubuntu_bionic_amd64/

It appears to only be an issue with the binary builds. 

I'd suggest that you just remove the version number which will avoid the attempts to build binaries but the source checkout and devel jobs will still work for people.

For the issue with the binary build a quick look I see shared objects being installed into `share` which I think that debian's linters are pruning and likely the shared objects are not getting onto your path. Its likely that until you use the stricter debian rules and fully isolated builds you might not see it or resolve the libraries from another source.  For example search for libnw.so

pschillinger:
I updated the PR toonly remove the version number now, as suggested.

Also, thank you for looking at the issue, @tfoote! Unfortunately, I have no experience with the debian installation system regarding such issues. The required libraries are downloaded during the build and are installed via the catkin share destination, but I have no idea why this is a problem here.

PR #19826, opened 2018-12-25
Adding ros_indy to documentation index for distro

thachdd88:
I would like ros_indy to be indexed and documented on ros.org.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -12741,8 +12741,18 @@ repositories:
     doc:
       type: git
       url: https://github.com/thachdd88/ros_indy.git
-      version: master
+      version: indigo
     release:
+      packages:
+      - indy5_description
+      - indy5_moveit_config
+      - indy7_description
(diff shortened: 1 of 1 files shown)


clalancette (approved):
This looks OK to me, though I'd like another opinion on whether the `ros_indy` name is appropriate.

One thing I'll note with this is that this will not only add the documentation, but will also release packages into indigo.  @thachdd88 , is that what you intended?

tfoote:
The repository name is only really used here so we don't have as strict naming convention. Often we have a metapackage that has the same name as the repository but if one of those is wanted it would probably just be indy instead of ros_indy to follow naming conventions.

I'm going to assume the release is intentional just added later.

thachdd88:
@clalancette : Yes, I would like to release the packages into indigo.
@tfoote  : I will change the repository name and make a metapackage to follow naming convention if that is the correct way to release my packages.

Thank you for your support, @clalancette @tfoote

PR #19828, opened 2018-12-26
Adding racecar for documentation index for distro

tianb03:
I'd like racecar to be indexed and documented on ros.org.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9371,6 +9371,29 @@ repositories:
       url: https://github.com/ros-visualization/qwt_dependency.git
       version: kinetic-devel
     status: maintained
+  racecar:
+    doc:
+      type: git
+      url: https://github.com/tianbot/tianbot_racecar.git
+      version: master
+    release:
+      packages:
+      - racecar_bringup
+      - racecar_core
+      - racecar_description
+      - racecar_navigation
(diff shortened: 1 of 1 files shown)


clalancette:
It looks like CI is failing because the release repository hasn't been properly setup.  Did you follow the instructions on blooming a package for the first time [here](http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease)?  That should setup the release repository correctly.

Additionally, I'd suggest calling the package something other than `racecar`, which is really quite generic.  Something like `tianbot_racecar` might be a better idea.

tianb03:
Thanks @clalancette ! First time commit to rosdistro I just didn't release...

I will change the name to tianbot_racecar. Does it matter that I keep the packages name unchanged just like racecar_bringup or better change it to tianbot_racecar_bringup also?

clalancette:
> I will change the name to tianbot_racecar. Does it matter that I keep the packages name unchanged just like racecar_bringup or better change it to tianbot_racecar_bringup also?

I think it is better to rename the packages to `tianbot_racecar_bringup`; otherwise, it is unclear for end users what they are getting when they `apt-get install ros-kinetic-racecar-bringup`.

tfoote:
+1 for prefixing all the packages with tianbot. It's much clearer.

PR #19829, opened 2018-12-28
Adding new repository for indexing: raspicam_node

ghost:
I used the wizard on https://index.ros.org/contribute/add_repo/ but the YAML description it generated doesn't contain all the information.
I tried with Chrome and Firefox. I hope it's ok.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9431,11 +9431,6 @@ repositories:
       url: https://github.com/ros-planning/random_numbers.git
       version: master
     status: maintained
-  raspicam_node:
-    doc:
-      type: git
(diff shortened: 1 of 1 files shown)


clalancette:
This failed CI because `raspicam_node` comes after `raspi_temperature` alphabetically.  If you fix this, we should be able to merge.  Thanks!

clalancette (approved):
This looks generally OK to me, though I'd highly suggest to add a `LICENSE` file at the top-level of the source repository; it makes it easier for others to figure out what the license of the code in the repository is.

leclercb:
Actually, it's not my repository. I'm only using it but it would be much more convenient to have it directly in the ros package list.

PR #19832, opened 2019-01-01
vapor_master: 0.1.0-0 in 'melodic/distribution.yaml' [bloom]

NickZ:
Increasing version of package(s) in repository `vapor_master` to `0.1.0-0`:

- upstream repository: https://github.com/roshub/vapor_master.git
- release repository: https://github.com/roshub/vapor_master-release.git
- bloom version: `0.6.9`
- previous version of package: `null`

### vapor_master
```
* initial release
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6166,6 +6166,21 @@ repositories:
       url: https://github.com/ros-drivers/usb_cam.git
       version: develop
     status: unmaintained
+  vapor_master:
+    doc:
+      type: git
+      url: https://github.com/roshub/vapor_master.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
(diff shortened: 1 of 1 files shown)


clalancette (approved):
This looks OK to me, though I'd like to get another opinion before merging.

One suggestion I have is to add license information to the top of every source file in the source repository (like https://github.com/roshub/vapor_master/blob/master/src/config.js).  I don't think it is strictly necessary (the LICENSE file at the top should suffice), but it is standard practice.

tfoote:
Including the header would be good though I know that js is often optimized to avoid content like that. But since the project license is clearly and consistently indicated at the top level it's good for us to merge.

tfoote:
The stretch builds are fialing due to npm not being in stretch: See #19845

PR #19843, opened 2019-01-02
libgazebo9 is not packaged for stretch

tfoote:
Follow up to #17750

@clalancette  this will effect melodic packages depending on gazebo they have to use the generic gazebo tag not the version specific one.
@j-rivero FYI

Related to this I'm wondering if we really want the gazebo9 tags in the main rosdep database at all. Like any other system dependency we expect to use the system version on all targeted platforms and as such we should just have one key that everyone uses to avoid this issue of packaging things that are uninstallable due to conflicting apt rules.

From: https://packages.debian.org/search?searchon=names&keywords=libgazebo
```
Package libgazebo7
stretch (stable) (libs): Open Source Robotics Simulator - shared library 
7.3.1+dfsg-3: amd64 arm64 armhf i386 mips mips64el mipsel
sid (unstable) (libs): Open Source Robotics Simulator - shared library 
7.8.1+dfsg-4+b1 [debports]: m68k sh4 
7.5.0+dfsg-1 [debports]: hppa
Package libgazebo7-dbgsym
sid (unstable) (debug): debug symbols for libgazebo7 
7.8.1+dfsg-4+b1 [debports]: m68k sh4 
7.5.0+dfsg-1 [debports]: hppa
Package libgazebo7-dev
stretch (stable) (libdevel): Open Source Robotics Simulator - Development Files 
7.3.1+dfsg-3: amd64 arm64 armhf i386 mips mips64el mipsel
sid (unstable) (libdevel): Open Source Robotics Simulator - Development Files 
7.8.1+dfsg-4+b1 [debports]: m68k sh4 
7.5.0+dfsg-1 [debports]: hppa
Package libgazebo9
buster (testing) (libs): Open Source Robotics Simulator - shared library 
9.4.1+dfsg-1+b1: amd64 i386
sid (unstable) (libs): Open Source Robotics Simulator - shared library 
9.4.1+dfsg-1+b1: amd64 i386
Package libgazebo9-dev
buster (testing) (libdevel): Open Source Robotics Simulator - Development Files 
9.4.1+dfsg-1+b1: amd64 i386
sid (unstable) (libdevel): Open Source Robotics Simulator - Development Files 
9.4.1+dfsg-1+b1: amd64 i386
```

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -881,15 +881,16 @@ gawk:
 gazebo:
   arch: [gazebo]
   debian:
-    buster: [gazebo7]
+    buster: [gazebo9]
     jessie: [gazebo7]
-    stretch: [gazebo7]
+    stretch: [gazebo9]
   fedora: [gazebo]
   gentoo: [sci-electronics/gazebo]
   slackware: [gazebo]
   ubuntu:
     artful: [gazebo9]
     bionic: [gazebo9]
+    cosmic: [gazebo9]
     precise: [gazebo]
     quantal: [gazebo]
     raring: [gazebo]
@@ -923,6 +924,7 @@ gazebo9:
   ubuntu:
     artful: [gazebo9]
     bionic: [gazebo9]
+    cosmic: [gazebo9]
 gcc-arm-none-eabi:
   arch: [gcc-arm-none-eabi]
   debian: [gcc-arm-none-eabi]
@@ -1977,6 +1979,7 @@ libgazebo9-dev:
   ubuntu:
     artful: [libgazebo9-dev]
     bionic: [libgazebo9-dev]
+    cosmic: [libgazebo9-dev]
 libgconf2:
   arch: [gconf]
   debian: [libgconf-2-4]


clalancette:
Hm, I thought we had solved this back when we released Melodic.  @j-rivero ended up packaging gazebo9 for Debian Stretch, and it is hosted at http://packages.osrfoundation.org/gazebo/debian-stable/.  Is that not enough for it to resolve in the rosdep database?

tfoote:
If it's not in the ROS repo or upstream it shouldn't be in the rosdep database. I believe there's an alternative rosdep source for packages.osrfoundation.org that you can optionally add if you add packages.osrfoundation.org to your apt sources.

tfoote:
THis is a bit of an issue as looking at REP 3 it looks like we've explicitly chosen to backport gazebo9 for stretch in melodic, but gazebo7 was still specified for stretch in Lunar which included stretch.

http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023

mikaelarguedas:
@tfoote yeah that's an issue we discussed before making the decision of using gazebo9 across the board for Melodic.

TL;DR
- users should depend on `gazebo_dev` or `gazebo_ros` instead of the generic `gazebo` key. This allows to not pin down versions in package.xml while targetting the right gazebo version based on the ROS distro.
- Once Lunar is EOL we could consider changing the `gazebo` rosdep key to point to `gazebo9` instead of `gazebo7`

---

My recollection in more details:
- It was decided to use the same version gazebo (9) on all targeted platforms for Melodic. Making Lunar and Melodic not side by side installable on Debian Stretch: (https://github.com/ros-infrastructure/rep/pull/160).
- Gazebo9 was back-ported to Artful and Stretch and imported in the ROS repos (http://build.ros.org/job/import_upstream/201).
- rosdep keys for gazebo9 have been added and are used by gazebo_ros_pkgs (#17750).

I believe the issue user faced is that the rosdep key for `gazebo` [points to gazebo 7](https://github.com/ros/rosdistro/blob/3768fdc68db0ca7639e300ab1cc25574fb1a0ac7/rosdep/base.yaml#L850) for Stretch.
This has been done purposefully to not change the version under Lunar user's feet.
This is similar to what we did for Jade on trusty:
- Jade targeted gazebo 5
- Indigo targeted gazebo 2
- Keys for gazebo 5 were added and used for gazebo_ros and friends. the generic `gazebo` key was not updated and still points to gazebo2.

dpkoch:
Should this (changing the debian stretch `gazebo` key to point to `gazebo9`) be revisited now that Lunar is EOL? Based on the second bullet in the previous comment.

tfoote:
Thanks for the ping back. I think it's time to switch the key forward. I've refactored this PR to do that.

clalancette (approved):

PR #19844, opened 2019-01-02
Adding python3 dependencies for rqt_bag

brawner:
This adds python3 versions of python-cairo and Pillow (the currently maintained fork of PIL). These are dependencies required by rqt_bag.

Links to distro packages
**python3-cairo**
https://www.archlinux.org/packages/extra/x86_64/python-cairo/
https://packages.debian.org/search?keywords=python3-cairo
https://rpmfind.net/linux/rpm2html/search.php?query=pkgconfig(py3cairo)
https://www.freshports.org/graphics/py3-cairo
https://packages.gentoo.org/packages/dev-python/pycairo
https://software.opensuse.org/package/python3-cairo
https://slackbuilds.org/repository/14.1/python/py3cairo/
https://packages.ubuntu.com/search?keywords=python3-cairo

**python3-imaging**
https://pkgs.alpinelinux.org/package/edge/main/x86/py3-pillow
https://www.archlinux.org/packages/community/x86_64/python-pillow/
https://packages.debian.org/jessie/python3-pil
https://rpmfind.net/linux/rpm2html/search.php?query=python3-Pillow
https://fedora.pkgs.org/28/fedora-x86_64/python3-pillow-qt-5.1.0-1.fc28.x86_64.rpm.html
https://www.freshports.org/graphics/py3-pillow
https://software.opensuse.org/package/python3-Pillow
https://slackbuilds.org/repository/14.1/libraries/python3-pillow/
https://packages.ubuntu.com/search?keywords=python3-pil&searchon=names

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4584,6 +4584,17 @@ python3-babeltrace:
   ubuntu:
     wily: [python3-babeltrace]
     xenial: [python3-babeltrace]
+python3-cairo:
+  arch: [python-cairo]
+  debian: [python3-cairo]
+  fedora: [py3cairo]
+  freebsd: [py3-cairo]
+  gentoo: [dev-python/pycairo]
+  opensuse: [python3-cairo]
+  slackware:
+    slackpkg:
+      packages: [py3cairo]
+  ubuntu: [python3-cairo]
 python3-catkin-pkg-modules:
   debian: [python3-catkin-pkg]
   fedora: [python3-catkin_pkg]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -4629,6 +4640,22 @@ python3-gitlab:
     xenial:
       pip:
         packages: [python-gitlab]
+python3-imaging:
```
tfoote: Our default is to use the debian package name for the key so I'd suggest we use python3-pil here as the key to be clear. python3-imaging is the old name and since this is a new key we might as well pick the newer name.

brawner: Ok, switched. ```pil``` is also the old name. As you can see most distros use pillow in the name to differentiate the newer fork from the older project. Maybe ```python3-pillow```?

brawner:
I also revised Alpine's package. Updated it to python3-pillow.

tfoote (approved):
Yeah, although pillow is the more accurate upstream name python3-pillow is a virtual package: https://packages.ubuntu.com/cosmic/python3-pillow and we still have a few issues on the buildfarm relating to those.

clalancette (approved):

clalancette (changes requested):
Arg.  Just noticed CI is failing.  It looks like because python3-pil is not in alphabetical order.  Once that is done, I'm happy to approve and merge this.

brawner:
Whoops. I definitely would need to reorder it after changing its name. CI should pass now.

clalancette (approved):

PR #19881, opened 2019-01-08
Remove 'ros_indy' from the distro

thachdd88:


thachdd88:
I got the error message "AssertionError: Duplicate package name 'indy5_description' exists in repository 'ros_indy' as well as in repository 'indy'"

Actually, I don't know how to remove the **_ros_indy_** repo and replace it with a new one (**_indy_** repo) to follow the naming conventions.

clalancette:
So if you want to rename from `ros_indy` to `indy`, you'll have to do it all in a single PR.  That is, you'll have to remove the `ros_indy` section, then add a new `indy` section into the YAML.

That being said, I think in https://github.com/ros/rosdistro/pull/19826#issuecomment-450952942 we determined that there is no explicit reason to rename it.  @thachdd88 , what is your reasoning for renaming the top-level key here?

thachdd88:
Hi @clalancette 

It's good to hear from you.

Indeed, I would like to follow the standard. The Indy is our series of industrial manipulators (like UR or ABB). So, I hope our customers to have the same experience (like ABB) when they work with the Indy robots on ROS. 

In the indigo/distribution.yaml, I removed the `ros_indy` and added `indy` section. Then, I made this pull request to merge it.

Please let me know if there is anything I have to do more.

Many thanks.

clalancette:
> In the indigo/distribution.yaml, I removed the `ros_indy` and added `indy` section. Then, I made this pull request to merge it.

This PR, though, only adds in the new section, which is why CI is complaining about a duplicate entry.  To make this pass CI and be considered for merging, you'll have to remove the `ros_indy` section (starting at https://github.com/ros/rosdistro/blob/master/indigo/distribution.yaml#L12757), and add in `indy`, all in this PR.  Let me know if that makes sense to you.

thachdd88:
I got your explanation.

In summary, I should make 2 pull requests. The 1st one is for removing `ros_indy` and the 2nd one is for adding `indy`.

If my understanding is correct, please remove this pull request (or I will remove if I can).

Thank you.

clalancette:
> In summary, I should make 2 pull requests. The 1st one is for removing `ros_indy` and the 2nd one is for adding `indy`.

That's one way to do it, but I think it is easier to just have a single pull request that simultaneously removes it as `ros_indy` and then adds it as `indy`.  That is, the final patch would have one section that removes ros_indy, and a different section that adds indy.

thachdd88:
I did remove 'ros_indy' and added 'indy' in this Pull Request but the CI kept reporting the duplicate issue.

So, for me, this PR is better for removing `ros_indy` only.

tfoote:
The diff attached to this PR appears to be zero length with the same content before and after. I think you might want to try starting with a new branch based off of master and then make your changes and put it to this branch with a force push to this branch. Make sure to pull before you make your new branch.

PR #19914, opened 2019-01-10
Added pip packages: breathe and exhale

veimox:
Added breathe and exhale: extensions of Sphinx to generate C++ documentation.

### Tests
[X] nosetests

### Breathe
Ubuntu: python-breathe https://packages.ubuntu.com/xenial/python-breathe
Debian: python-breathe https://packages.debian.org/buster/python-breathe
Same for Fedora: https://src.fedoraproject.org/rpms/python-breathe/

### Exhale
* [Pypi](https://pypi.org/project/exhale/)
* [Docs](https://exhale.readthedocs.io/en/latest/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -49,6 +49,16 @@ azure-iothub-device-client-pip:
   ubuntu:
     pip:
       packages: [azure-iothub-device-client]
+breath-pip:
+  debian:
+    pip:
+      packages: [breath]
+  fedora:
+    pip:
+      packages: [breath]
+  ubuntu:
+    pip:
+      packages: [breath]
 canopen-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -49,16 +49,16 @@ azure-iothub-device-client-pip:
   ubuntu:
     pip:
       packages: [azure-iothub-device-client]
-breath-pip:
+breathe-pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Following our policy of using native packages when they exist please switch to use debs for breathe

Ubuntu: `python-breathe` https://packages.ubuntu.com/xenial/python-breathe
Debian: `python-breathe` https://packages.debian.org/buster/python-breathe

Same for Fedora: https://src.fedoraproject.org/rpms/python-breathe/

Since exhale is still a "Pre Alpha" https://pypi.org/project/exhale/ it's probably not packaged yet so the pip rule makes sense for now.

veimox:
@tfoote I have added `python-breathe` and I think it does not hurt to leave `breathe-pip`. I see the use case because the version in xenial (4.1) is quite old compare with the pypi (4.11). Still I am not sure if simplicity is more important than functionality in this repo. Shall I remove the `breathe-pip`?

tfoote:
Please do remove the overlapping pip entry. We'd rather not have duplicate rules for pip vs apt. If anyone uses the pip rules they will override the apt installation system wide and persist which is the common cause of unexpected side effects.

veimox:
@tfoote done

tfoote (approved):

cottsay (approved):

PR #19915, opened 2019-01-10
Add moviepy

peci1:
Add moviepy for video processing in python.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2128,6 +2128,16 @@ python-more-itertools-pip:
   ubuntu:
     pip:
       packages: [more-itertools]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
lgtm, I don't think there's any native packages for this.

peci1:
The only "problem" I encountered is they require some newer ffmpeg binary. On Trusty, it is named avconv, and when I forced moviepy to use the system-installed avconv, it told me that it is too old and missing some features important for moviepy.

However, moviepy knows about this situation and offers the user to automatically download a newer ffmpeg static binary via imageio. Then it uses this binary and works well.

cottsay (approved):

PR #19921, opened 2019-01-11
Adding tinydir_vendor

AAlon:
As discussed @ https://github.com/ros2/rcl/pull/332.

### DIFF
--- a/crystal/distribution.yaml
+++ b/crystal/distribution.yaml
@@ -1714,6 +1714,21 @@ repositories:
       url: https://github.com/ros2/teleop_twist_keyboard.git
       version: ros2
     status: maintained
+  tinydir_vendor:
+    doc:
+      type: git
+      url: https://github.com/aws-robotics/vendored-tinydir.git
+      version: master
+    release:
+      tags:
+        release: release/crystal/{package}/{version}
+      url: https://github.com/aws-gbp/tinydir_vendor-release.git
+      version: 1.0.0-0
+    source:
+      type: git
+      url: https://github.com/aws-robotics/vendored-tinydir.git
+      version: master
+    status: maintained
   tinyxml2_vendor:
     doc:
       type: git
--- a/crystal/distribution.yaml
+++ b/crystal/distribution.yaml
@@ -1717,16 +1717,16 @@ repositories:
   tinydir_vendor:
     doc:
       type: git
-      url: https://github.com/aws-robotics/vendored-tinydir.git
+      url: https://github.com/ros2/tinydir_vendor.git
       version: master
     release:
       tags:
         release: release/crystal/{package}/{version}
-      url: https://github.com/aws-gbp/tinydir_vendor-release.git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Holding for crystal sync and ament_lint release.

crystal/distribution.yaml
```diff
@@ -1714,6 +1714,21 @@ repositories:
       url: https://github.com/ros2/teleop_twist_keyboard.git
       version: ros2
     status: maintained
+  tinydir_vendor:
+    doc:
+      type: git
+      url: https://github.com/aws-robotics/vendored-tinydir.git
```
clalancette: If this is going to be depended on by `rcl`, then I think it would make more sense to host this under the `ros2` organization.  @ros2/team, thoughts?

nuclearsandwich: That seems reasonable to me. If we move it we should also move the release repository under ros2-gbp and make sure the maintainers still have access to both.

dirk-thomas: We talked about this internally and came to the conclusion that it makes sense to host all repositories containing packages for the core of ROS 2 under the `ros2` / `ros2-gbp` org units.

@AAlon Can you move both repositories to my GitHub user and then I will move them into the ROS 2 org units as well as grant you full access to both repos again. If there are any other GitHub users which I should add initially please let me know - otherwise any admin can add them later.

dirk-thomas: New location: https://github.com/ros2-gbp/tinydir_vendor-release

chapulina: > Regarding permissions, if you could add everyone under https://github.com/orgs/aws-robotics/people that would be great

The list of users is not publicly accessible: "This organization has no public members."

chapulina:
CI using `aws-robotics/vendored-tinydir`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6048/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2541/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4987/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5917/)

crystal/distribution.yaml
dirk-thomas: > if could add everyone under https://github.com/orgs/aws-robotics/people that would be great.,

@AAlon you should have the permissions to add others as collaborators to the repo.

nuclearsandwich:
@AAlon you should be able to use the `-p` flag on bloom-release to create a new pull request using the updated release and source repositories. If you'd like me to do so instead just say the word.

dirk-thomas (approved):

crystal/distribution.yaml
AAlon: Could you also grant me the permissions for the source repo? (https://github.com/ros2/tinydir_vendor)
Thanks!

PR #19944, opened 2019-01-15
Added numpy-quaternion to python.yaml 

tgaspar:
Added pip package to the numpy implementation of quaternions.
numpy-quaternion adds quaterni9on dtype to NumPy (https://github.com/moble/quaternion).

To the best of my knowledge it is only available via Pip.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2338,6 +2338,13 @@ python-numpy-stl-pip:
   ubuntu:
     pip:
       packages: [numpy-stl]
+python-numpy-quaternion-pip:
+  osx:
+    pip:
+      packages: [numpy-quaternion]
+  ubuntu:
+    pip:
+      packages: [numpy-quaternion]
 python-numpydoc:
   arch: [python2-numpydoc]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Please don't change unrelated sections of the file it makes the historical audit of the code much harder and can lead to conflicts for others. Also the explicit null entries are on purpose. https://github.com/ros-infrastructure/rep/blob/master/rep-0111.rst#allow-package_argument-to-be-explicitly-null

Otherwise this looks fine. It does look like a young enough project that there's no upstream packaging. Note that you will not be able to make releases against a pip entry but you can use it when building from source.

tgaspar:
Thank you for your review. 

I must admit I was surprised to read "don't change unrelated sections" because I do not recall doing that. But when I checked my commit diff I right away noticed what you mean. I think those changes came after running `clean_rosdep_yaml.py`. I will clean this mess up and push the updated version.

tfoote (approved):

cottsay (approved):

PR #19950, opened 2019-01-15
rosdep: add golang-1.10-go specific rule

Vuwij:
There is no debian package that as 1.10. Gentoo's one seems to be the latest

https://packages.ubuntu.com/cosmic/golang-1.10
https://packages.debian.org/search?keywords=golang-1.&searchon=names&suite=stable&section=all
https://packages.gentoo.org/packages/dev-lang/go

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1018,7 +1018,6 @@ gnuplot:
   ubuntu: [gnuplot]
 golang-1.10-go:
   ubuntu: [golang-1.10-go]
-  gentoo: [dev-lang/go]
 golang-go:
   debian: [golang-go]
   gentoo: [dev-lang/go]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1016,6 +1016,9 @@ gnuplot:
   fedora: [gnuplot]
   gentoo: [sci-visualization/gnuplot]
   ubuntu: [gnuplot]
+golang-1.10-go:
+  ubuntu: [golang-1.10-go]
+  gentoo: [dev-lang/go]
 golang-go:
   debian: [golang-go]
   gentoo: [dev-lang/go]


tfoote:
Can you explain your use case for why golang-go key is not good enough for your use case? golang-1.10-go is not available on all versions of ubuntu so it will need a much more limited rule than all of ubuntu. And I suspect that gentoo is going to move forward off of 1.10 soon too. In general we prefer to use the generic key that will keep working over time rather than explicitly calling out the version which takes a lot more management and the developer needs to know what versions are available by default on the target platforms.

Note that you're failing CI due to being out of alphabetical order.

Vuwij:
Hi @tfoote . I am working on a project that is using a go backend for web server. The web server has some packages that are only available with go version 1.7 and up. For this, I would normally use snap install go to get the latest version. However go 1.10 for ubuntu is latest one I can find.

I notice some packages like gstreamer and libusb have the version number included such as libusb-1.0. So I'm not sure if this is possible

tfoote:
What platform are you trying to target? I'm guessing Xenial because it only has 1.6 installed by default. https://packages.ubuntu.com/xenial/golang-go and you need 1.7. Unfortunately the 1.10 packing that you've created a rule for is only valid for bionic and newer, unless you have backports or updates enabled which we do not assume to be enabled. And on bionic and newer 1.10 is the default.

As such I think that this rule is not going to achieve what you're looking for of installing a newer backport of golang onto your system. As such I'd suggest that we decline to specifically call out the version in a parallel rule.

Note that libusb vs libusb1.0 are basically completely different libraries and are side by side installable but have the same name and the version number is effectively baked into the name.

Vuwij:
There is a xenial backport here so I am able to install it on xenial
https://packages.ubuntu.com/xenial-backports/golang-1.9

I see, so basically I would need my system to be on bionic to automatically install golang-2.0. So there really isn't any solution using rosdep i guess.

tfoote:
rosdep isn't designed to do backports or modify the packages available. It's designed to facilitate installing ad resolving what's already available across different platforms.

If you want to turn on backports on your system, you can create a custom rosdep rule to add to your rosdep sources which will resolve to use the backported version of golang that you need. We just can't add that to our public database.

PR #19980, opened 2019-01-18
Update indy repository for the distro

ThachDoNRMK:


sloretz:
@thachdd88 it looks like this PR no longer releases from your repo, but instead releases from a new one where the git history has been rewritten such that all commits are from @ThachDoNRMK. Is that intentional?

thachdd88:
Hi @sloretz 

Those are my accounts. "thachdd88" is my personal one and "ThachDoNRMK" is for my work.

I am sorry for making you confused.

PR #19994, opened 2019-01-21
Some frameworks for django

Vuwij:
https://packages.ubuntu.com/disco/python-djangorestframework
https://packages.ubuntu.com/disco/python-django-cors-headers
https://packages.ubuntu.com/disco/python-django-extensions

https://packages.debian.org/stretch/python-django-cors-headers
https://packages.debian.org/stretch/python-django-extensions
https://packages.debian.org/stretch/python-djangorestframework


cottsay (approved):
Looks like Fedora only packages Django with Python 3, so we shouldn't add those keys here.

PR #19978, opened 2019-01-17
rosjava_build_tools: 0.3.3-1 in 'kinetic/distribution.yaml' [bloom]

jubeira:
Increasing version of package(s) in repository `rosjava_build_tools` to `0.3.3-1`:

- upstream repository: https://github.com/rosjava/rosjava_build_tools.git
- release repository: https://github.com/rosjava-release/rosjava_build_tools-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.6.9`
- previous version for package: `0.3.2-0`

## rosjava_build_tools

```
* Fixed problem to find gradlew when cross-compiling.
* Gradle 2.14.1 --> 4.10.2.
* Fix for genjava ignoring most packages in standalone mode.
* Contributors: Johannes Meyer, Juan Ignacio Ubeira, Julian Cerruti, ivanpauno
```

PR #20018, opened 2019-01-22
Revert rosjava_build_tools request

tfoote:
This reverts #19978 which is causing a regression of rosjava_core and others.

@jubeira FYI

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -11274,7 +11274,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/rosjava-release/rosjava_build_tools-release.git
-      version: 0.3.3-1
+      version: 0.3.2-0
     source:
       type: git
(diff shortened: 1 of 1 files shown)


tfoote:
Reverting this did not fix the build issues:

Initial build failure due to the release: http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__rosjava_core__ubuntu_xenial_amd64__binary/38/

Rebuilding failed with the rollback: 
http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__rosjava_core__ubuntu_xenial_amd64__binary/44/

I'm going to revert the revert and I've ticked it upstream https://github.com/rosjava/rosjava_core/issues/288

jubeira:
@tfoote this is because Gradle will look for a script that is in the upstream repository, even if it's not yet released. We will need to update Gradle to 4.x in `rosjava_core`. and make the release there; I'll send a patch soon.

PR #20056, opened 2019-01-24
Try adding codeowners for automatic review requests.

nuclearsandwich:
I've been curious to try this out for some time. Mostly I hope this will make it easier to triage rosdistro PRs by escalating things that are important to me.

https://help.github.com/articles/about-code-owners/

I'm happy to serve as the sole pilot but I will also invite @tfoote and @clalancette

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# This file allows changes in specific paths to automatically request review from individuals.
+# See https://help.github.com/articles/about-code-owners/ for details.
+
+bouncy/* @nuclearsandwich
+crystal/* @nuclearsandwich
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,3 +3,6 @@
 
 bouncy/* @nuclearsandwich
 crystal/* @nuclearsandwich
+indigo/* @tfoote
+kinetic/* @tfoote
+rosdep/* @tfoote
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,4 +5,6 @@ bouncy/* @nuclearsandwich
 crystal/* @nuclearsandwich
 indigo/* @tfoote
 kinetic/* @tfoote
+lunar/* @clalancette
+melodic/* @clalancette
 rosdep/* @ros/rosdeputies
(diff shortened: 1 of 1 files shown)


clalancette:
I'm not really sure how I feel about this.  I think it is good that the current system spreads the review/merge of these PRs around, and by assigning the reviews to the distro maintainers, I'm afraid it may end up discouraging others from reviewing/merging PRs.

nuclearsandwich:
> and by assigning the reviews to the distro maintainers

This doesn't assign the review, it has the same effect as requesting someone's review via the GitHub web interface which bumps up the "priority" of the notification. The reason I want that is that my natural workflow for the rosdistro repositories is to look for crystal and bouncy PRs first and then move on to others and it would be convenient if the notification level aided that process.

tfoote:
That's a great way to help triage these things and for the distro maintainers to get more active notifications. It seems like a good thing for the ros boss to be on for each distro. And then have a team for the rosdep keys. That should help keep us responsive to new issues.

nuclearsandwich:
I took the liberty of changing the rosdep db to a team mention and inaugurated the \@ros/<b>rosdep</b>uties team to be the target. @cottsay I dragged you along in that team for your Fedora wisdom with humble apologies. External tools using the GitHub API can classify team mentions differently. @tfoote if you want to go back to the individual mention on rosdep I beg your pardon for the churn.

tfoote (approved):
lgtm

@clalancette do you want lunar and melodic notifications?

nuclearsandwich:
I think I figured out why the rosdeputies team wasn't being added for review. I was getting grumpy about that and couldn't even add the team as a review request manually so I scoured the documentation and found that the team needs at least explicit read access to the repo before it is eligible for review request regardless of the access already afforded to individual team members. I've made it so and I expect that to start working so @ros/rosdeputies can expect higher profile notifications as a result.

tfoote:
Thanks @nuclearsandwich that will be great.

nuclearsandwich:
Still no joy :rage: https://github.com/ros/rosdistro/pull/20505 was opened without any review request triggered.

tfoote:
According to this thread the team actually needs write access: https://github.community/t5/How-to-use-Git-and-GitHub/CODEOWNERS-works-with-users-but-not-teams/td-p/4986

tfoote:
Success in #20515 adding team write permission!

For future reference the team also has to be public/not secret.

PR #20103, opened 2019-01-29
added homer_tts

niyawe:
needed dependency for homer_navigation

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -4128,6 +4128,13 @@ repositories:
       url: https://gitlab.uni-koblenz.de/robbie/homer_robot_face-release.git
       version: 1.0.18-1
     status: developed
+  homer_tts:
+    release:
+      tags:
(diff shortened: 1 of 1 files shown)


clalancette:
@niyawe I know I merged this, but I just realized that the source code for this repository doesn't have a license (there is no LICENSE file, the `<license>` tag in the package.xml is empty, and none of the sources have a license header).  Could you add a license?  Otherwise, I'll have to revert since we can't guarantee that this code is open source.  Thank you.

niyawe:
There is a `<license>` tag: https://gitlab.uni-koblenz.de/robbie/homer_tts/blob/master/package.xml#L16

dirk-thomas:
> it makes our job of verifying that it is open source much easier that way

I don't think that we "verify" the license of registered packages / repositories. It is up to the maintainer to choose whatever they see fit. For obvious reasons an OSI approved license is preferred.

PR #20106, opened 2019-01-29
rosdep: added libstdc++6

niyawe:
The homer_map_manager package fails to build on arm and arm64 because the libstdc++5 package is missing. So I need to add the libstdc++6 package, since it exists for Ubuntu arm.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3136,6 +3136,11 @@ libstdc++5:
   gentoo: [virtual/libstdc++]
   opensuse: [libstdc++33]
   ubuntu: [libstdc++5]
+libstdc++6:
+  arch: [gcc-libs]
+  debian: [libstdc++6]
+  fedora: [libstdc++]
+  ubuntu: [libstdc++6]
 libsvm-dev:
   debian: [libsvm-dev]
   fedora: [libsvm-devel]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Ubuntu: https://packages.ubuntu.com/xenial/libstdc++6
Debian: https://packages.debian.org/stretch/libstdc++6

Fedora: https://fedora.pkgs.org/29/fedora-x86_64/libstdc++-8.2.1-2.fc29.x86_64.rpm.html this looks to have the so.6 in it so I think it's the same.

sloretz:
> Fedora: https://fedora.pkgs.org/29/fedora-x86_64/libstdc++-8.2.1-2.fc29.x86_64.rpm.html this looks to have the so.6 in it so I think it's the same.

Both rosdep keys `libstdc++5` and `libstdc++6` keys map to `libstdc++` for fedora. It looks like fedora packages `libstdc++5-...` and `libstdc++6` packages list `libstdc++` under provides ( https://rpmfind.net/linux/RPM/mandriva/2007.0/extra/i586/libstdc++5-3.3.6-3mdk.i586.html  and https://rpmfind.net/linux/RPM/mandriva/2011/x86_64/media/main/updates/lib64stdc++6-4.6.1-3.3.x86_64.html ). Maybe the rosdep keys should map to `libstdc++5` and `libstdc++6` for fedora instead?

niyawe:
fixed the fedora part

tfoote (approved):

PR #20111, opened 2019-01-30
Adding doc and source to documentation index for distro

apalomino-edgebrain:
Adding a source and doc entry for earth_rover_localization package.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2417,7 +2417,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/earthrover/earth_rover_localization.git
-      version: 1.0.0-1
+      version: master
     release:
       tags:
         release: release/kinetic/{package}/{version}
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2414,11 +2414,20 @@ repositories:
       version: master
     status: maintained
   earth_rover_localization:
+    doc:
+      type: git
+      url: https://github.com/earthrover/earth_rover_localization.git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -2414,11 +2414,20 @@ repositories:
       version: master
     status: maintained
   earth_rover_localization:
+    doc:
+      type: git
+      url: https://github.com/earthrover/earth_rover_localization.git
+      version: 1.0.0-1
```
tfoote: This supports tags but we strongly recommend using a branch here so that you don't have to make a release to update the docs. 
```suggestion
      version: master
```

Note that this is not a valid tag either the only tags is [`1.0.0`](https://github.com/earthrover/earth_rover_localization/tree/1.0.0)

kinetic/distribution.yaml
```diff
@@ -2414,11 +2414,20 @@ repositories:
...
     release:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/earthrover/earth_rover_localization-release.git
       version: 1.0.0-1
+    source:
+      type: git
+      url: https://github.com/earthrover/earth_rover_localization.git
+      version: 1.0.0-1
```
tfoote:
```suggestion
      version: master
```

Our CI requires a branch here so that it can check for updates when you make new commits. I see there's also a `develop` branch too. I don't know what your process is but am guessing master is the best one to use.

tfoote (changes requested):

tfoote (approved):

PR #20128, opened 2019-01-31
Add pycryptodome

maximeconnolly:
Add pycryptodome which is need by https://github.com/ros/ros_comm/pull/1609


tfoote (changes requested):
If this is targeted for ros_comm it must have native packages on all platforms and not pip packages. The pip dependencies cannot be used when being built into native packages, only when building from source.

Also please link to the packages online in the body of the PR for easier review.

dirk-thomas:
Since the referenced PR in ros_comm is not going to be merge we could just close this PR (if there is no other use case this is intended for).

PR #20185, opened 2019-02-05
Added python-ftdi1

peci1:
This is a (hidden) dependency of adafruit-gpio-pip when you want to use its FT232 subpackage.

I haven't found any binary packages for Fedora and OSX.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1386,6 +1386,11 @@ python-freezegun-pip:
   ubuntu:
     pip:
       packages: [freezegun]
+python-ftdi1:
+  debian: [python-ftdi1]
+  fedora: [python2-libftdi]
+  gentoo: [dev-embedded/libftdi]
+  ubuntu: [python-ftdi1]
 python-future:
   debian: [python-future]
   fedora: [python-future]


rosdep/python.yaml
```diff
@@ -1386,6 +1386,13 @@ python-freezegun-pip:
   ubuntu:
     pip:
       packages: [freezegun]
+python-ftdi1:
+  debian: [python-ftdi1]
+  gentoo: [dev-embedded/libftdi]
+  ubuntu:
```
tfoote: Please collapse this to a single line rule. It is available on all currently supported distros since xenial there's no need to call out specific versions. 
https://packages.ubuntu.com/xenial/python-ftdi1 (The website only lists supported distros, intermediate ones are certainly there. And likely some of the slightly older ones.) It's not available on Trusty but since the rule doesn't exist, it's known that no-one is using it on trusty, so having an invalid rule won't break anyone.

tfoote (changes requested):
Debian package for the record: https://packages.debian.org/sid/python-ftdi1

Gentoo: https://packages.gentoo.org/packages/dev-embedded/libftdi

In general when submitting rosdep rules please link to the package resources. 

I found a package libftdi-python which looks like it for fedora: https://centos.pkgs.org/7/epel-x86_64/libftdi-python-1.1-4.el7.x86_64.rpm.html

peci1:
Found the Fedora package: https://apps.fedoraproject.org/packages/python2-libftdi .

rosdep/python.yaml
peci1: Okay, if the non-existing rule for trusty is not a problem...

tfoote (approved):

rosdep/python.yaml
nuclearsandwich: > It's known that no-one is using it on trusty, so having an invalid rule won't break anyone.

@tfoote This seems contrary to our general position on rosdep validity. For other recent changes we've used the

```yaml
key:
  ubuntu:
    '*': [debname]
    trusty: null
```   

construct as we do for [libdlib-dev](https://github.com/ros/rosdistro/blob/bacb693a5f092c7d844cfda522321075a957d409/rosdep/base.yaml#L1635-L1637). Has our general policy changed with Trusty's imminent retirement or was I misinterpreting things previously?

nuclearsandwich (commented):
Holding off on merging since I've asked a clarifying question in https://github.com/ros/rosdistro/pull/20185#discussion_r254408226

rosdep/python.yaml
tfoote: The `'*'` syntax is a little bit more correct. I'd be fine with either. With trusty EOL in a few months, saving a future touch of the file to remove the older rule at some point has some benefits. The wildcard rule will still fail for recently EOL'd distros. Anything not completely specific is prone to overmatch for some elements.

nuclearsandwich (approved):

PR #20261, opened 2019-02-09
add commits_valid for check_git_remote_exists

k-okada:
Can we use commits (not tag, nor branch) for source version? 
for example, `bfl` repository uses specific commit for ros release due to backward compatibility, but it seems failing on nose tests (https://api.travis-ci.org/v3/job/490889110/log.txt) if we allow this for source entry, we need to fix test code.
If this is not allowed for doc/source (although allowed for release), then it's ok. I'll fix PR #20260

https://github.com/ros/rosdistro/pull/20260/files
```
    source:
	      type: git
	      url: https://github.com/toeklk/orocos-bayesian-filtering.git
	      version: f21791e4d35f7343b06819e0507a21b83edf8f99
```

### DIFF
--- a/test/test_url_validity.py
+++ b/test/test_url_validity.py
@@ -10,6 +10,7 @@ except ImportError:
 import os
 import subprocess
 import sys
+import tempfile
 import unittest
 try:
     from urllib.parse import urlparse
@@ -76,7 +77,7 @@ def detect_lines(diffstr):
     return resultant_lines
 
 
-def check_git_remote_exists(url, version, tags_valid=False):
+def check_git_remote_exists(url, version, tags_valid=False, commits_valid=False):
     """ Check if the remote exists and has the branch version.
     If tags_valid is True query tags as well as branches """
     cmd = ('git ls-remote %s refs/heads/*' % url).split()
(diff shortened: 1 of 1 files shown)


clalancette:
My understanding is that a branch is required for the source entry because of some requirements of Jenkins (though I don't know the details; @tfoote may).  I believe the doc entry can point to either a branch or a tag, though we usually recommend a branch.

nuclearsandwich:
For development and pull requests jobs to work the source entry must be a branch. I haven't audited but I'd wager there are at least a few places that make that assumption between the rosdistro, ros_buildfarm, and bloom codebases.

Is there any reason not to create a branch that will just always point to the specific commit ID? Additional branches cost almost nothing in Git and it would allow all tools to work as expected in the ROS ecosystem.

k-okada:
Thanks for feedback, since I not a developer of the original package so I do not have access to the repository to create new tag/branch.
I think this is a very special case, so close for now.

tfoote:
Yes, this is something that jenkins cannot handle either. I added documentation of this in the CONTRIBUTING.md in #20286

nuclearsandwich:
FYI a situation has actually arisen in ROS 2 where we need similar functionality. I've added a commit in #20287 which incorporates these changes and allows a source version to be a commit (or tag) if both test_commits and test_pull_requests are explicitly false. @k-okada if you would like to re-introduce a source entry to https://github.com/ros/rosdistro/pull/20260 after that PR merges then you should be able to do so by setting test_commits and test_pull_requests to false.

PR #20287, opened 2019-02-12
Update url validity check with better messages and tag priority

tfoote:
This is capturing the requirements documented in #20286 more clearly in the unit tests.

### DIFF
--- a/test/test_url_validity.py
+++ b/test/test_url_validity.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
     from io import StringIO
 import os
+import shutil
 import subprocess
 import sys
 import tempfile
@@ -130,6 +131,9 @@ def check_git_remote_exists(url, version, tags_valid=False, commits_valid=False)
             return (True, '')
         except:
             return (False, 'No commit found matching %s' % version)
+        finally:
+            shutil.rmtree(tmpdir)
+
     
     return (False, 'No branch found matching %s' % version)
     
--- a/test/test_url_validity.py
+++ b/test/test_url_validity.py
@@ -10,6 +10,7 @@ except ImportError:
 import os
 import subprocess
 import sys
+import tempfile
 import unittest
 try:
     from urllib.parse import urlparse
@@ -76,7 +77,7 @@ def detect_lines(diffstr):
     return resultant_lines
 
 
-def check_git_remote_exists(url, version, tags_valid=False):
+def check_git_remote_exists(url, version, tags_valid=False, commits_valid=False):
     """ Check if the remote exists and has the branch version.
     If tags_valid is True query tags as well as branches """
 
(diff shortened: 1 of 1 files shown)


nuclearsandwich (dismissed):

nuclearsandwich:
Per the ongoing discussion in #20309 I've updated this PR to allow source entries to contain tags when test_commits and test_pull_requests are both explicitly false. I've also added changes based on https://github.com/ros/rosdistro/pull/20261 (being sure to give co-authorship to @k-okada) which allow commits to be tested as well.

test/test_url_validity.py
```diff
@@ -76,50 +77,77 @@ def detect_lines(diffstr):
...
+    if not version:
+        # If the above passed assume the default exists
+        return (True, '')
+
+    if 'refs/heads/%s' % version in output:
+        return (True, '')
+
+    if commits_valid:
+        try:
+            tmpdir = tempfile.mkdtemp()
```
tfoote: This tmpdir needs to get cleaned up.

test/test_url_validity.py
```diff
@@ -76,50 +77,77 @@ def detect_lines(diffstr):
...
-def check_source_repo_entry_for_errors(source, tags_valid=False):
+    try:
+        output = subprocess.check_output(cmd).decode('utf-8')
+    except subprocess.CalledProcessError as ex:
+        return (False, 'subprocess call %s failed: %s' % (cmd, ex))
+    if not version:
+        # If the above passed assume the default exists
+        return (True, '')
+
+    if 'refs/heads/%s' % version in output:
```
tfoote: This is technically incorrect to test before the commits. However testing the commits is very expensive. I'd suggest that we only test the commits if the `version` matches the appropriate regex of 40 characters and `[0-9a-f]`

tfoote (commented):
Sorry I got half way through this and my review comments dissappeared. Apparently I can only comment on this now too. Edit: Oh because it's my PR originally ;-)

test/test_url_validity.py
nuclearsandwich: I took a quick look in the tempfile library docs but on the off chance it's there and I didn't find it, is there a context manager based way to mange temp directories a la

```
with tempfile.mkdtemp() as tmpdir
  ....
```

test/test_url_validity.py
nuclearsandwich: 610d309

test/test_url_validity.py
nuclearsandwich: e8dc20a

tfoote (commented):
I added a commit updating the order of checking.

nuclearsandwich:
It looks like Tully and I both think this is ready. Anyone else want to give it a :mag: ?

k-okada:
thanks!, LGTM

PR #20309, opened 2019-02-14
Update source and doc branches for Crystal.

nuclearsandwich:
Created `crystal` branches and updated the release repository crystal track for the following repositories.

https://github.com/ament/ament_cmake/tree/crystal
https://github.com/ament/ament_index/tree/crystal
https://github.com/ament/ament_lint/tree/crystal
https://github.com/ament/ament_package/tree/crystal
https://github.com/ament/googletest/tree/crystal
https://github.com/ament/uncrustify_vendor/tree/crystal
https://github.com/osrf/osrf_testing_tools_cpp/tree/crystal
https://github.com/ros2/ament_cmake_ros/tree/crystal
https://github.com/ros2/common_interfaces/tree/crystal
https://github.com/ros2/console_bridge_vendor/tree/crystal
https://github.com/ros2/depthimage_to_laserscan/tree/crystal
https://github.com/ros2/example_interfaces/tree/crystal
https://github.com/ros2/examples/tree/crystal
https://github.com/ros2/geometry2/tree/crystal
https://github.com/ros2/joystick_drivers/tree/crystal
https://github.com/ros2/kdl_parser/tree/crystal
https://github.com/ros2/launch/tree/crystal
https://github.com/ros2/libyaml_vendor/tree/crystal
https://github.com/ros2/message_filters/tree/crystal
https://github.com/ros2/orocos_kinematics_dynamics/tree/crystal
https://github.com/ros2/pcl_conversions/tree/crystal
https://github.com/ros2/poco_vendor/tree/crystal
https://github.com/ros2/rcl_interfaces/tree/crystal
https://github.com/ros2/rcl_logging/tree/crystal
https://github.com/ros2/rcutils/tree/crystal
https://github.com/ros2/robot_state_publisher/tree/crystal
https://github.com/ros2/ros1_bridge/tree/crystal
https://github.com/ros2/ros2cli/tree/crystal
https://github.com/ros2/rosidl_dds/tree/crystal
https://github.com/ros2/rosidl_defaults/tree/crystal
https://github.com/ros2/rosidl/tree/crystal
https://github.com/ros2/rosidl_python/tree/crystal
https://github.com/ros2/rosidl_typesupport_connext/tree/crystal
https://github.com/ros2/rosidl_typesupport_fastrtps/tree/crystal
https://github.com/ros2/rosidl_typesupport/tree/crystal
https://github.com/ros2/ros_workspace/tree/crystal
https://github.com/ros2/sros2/tree/crystal
[...]

PR #20318, opened 2019-02-15
Do not merge: Testing commit checks

nuclearsandwich:
This is a fusion of #20287 and #20309 to test if the changes to the former allow the latter to pass CI.


tfoote:
@nuclearsandwich can we close this out?

PR #20340, opened 2019-02-19
robot_navigation: 0.2.5-1 in 'melodic/distribution.yaml' [bloom]

DLu:
Increasing version of package(s) in repository `robot_navigation` to `0.2.5-1`:

- upstream repository: https://github.com/locusrobotics/robot_navigation.git
- release repository: https://github.com/DLu/robot_navigation-release.git
- bloom version: `0.7.2`
- previous version for package: `0.2.4-0`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4557,8 +4557,8 @@ repositories:
       - robot_navigation
       tags:
         release: release/melodic/{package}/{version}
-      url: https://github.com/locusrobotics/robot_navigation-release.git
-      version: 0.2.4-0
+      url: https://github.com/DLu/robot_navigation-release.git
+      version: 0.2.5-1
     source:
       test_pull_requests: true
       type: git


melodic/distribution.yaml
```diff
@@ -4557,8 +4557,8 @@ repositories:
       - robot_navigation
       tags:
         release: release/melodic/{package}/{version}
-      url: https://github.com/locusrobotics/robot_navigation-release.git
-      version: 0.2.4-0
+      url: https://github.com/DLu/robot_navigation-release.git
```
tfoote: Is this url change intentional?

DLu: Yes. Conversation here: https://answers.ros.org/question/315581/bloom-same-version-number-different-tag/

Ultimately I prefer to have the different branches on the same version number.

tfoote: Ok, this will work for this, but only because we happen to prefix the sourcedebs with the rosdistro name. In general packaging toolchains expect XXX version Y.Y.Y to be the same whether packaged for one distro or another. If a user finds a reference to 0.2.5 it will be hard for them to understand which version of 0.2.5 they're supposed to use as tags are generally expected to be unique.

We can make it happen but generally strongly recommend against abusing the system like this. The standard thing to do is have separate version series for the forked development paths and versions are expected to be unique.

PR #20382, opened 2019-02-26
grid_map: 1.6.0-0 added to 'melodic/distribution.yaml'

remod:
Releasing grid_map (previously released for indigo, jade and kinetic) now for melodic.


remod:
We will use bloom for this.

PR #20289, opened 2019-02-12
industrial_core: 0.7.0-0 in 'melodic/distribution.yaml' [bloom]

Levi-Armstrong:
Increasing version of package(s) in repository `industrial_core` to `0.7.0-0`:

- upstream repository: https://github.com/ros-industrial/industrial_core.git
- release repository: https://github.com/ros-industrial-release/industrial_core-release.git
- distro file: `melodic/distribution.yaml`
- bloom version: `0.7.2`
- previous version for package: `null`

## industrial_core

```
* No changes
```

## industrial_deprecated

```
* No changes
```

## industrial_msgs

```
* msgs: fix minor typo in RobotStatus comment. Fix #186 <https://github.com/ros-industrial/industrial_core/issues/186>.
* Contributors: G.A. vd. Hoorn, Levi Armstrong, Shaun Edwards
```

## industrial_robot_client

```
* Decrease idle wait time.  250 ms is too slow. Fix #214 <https://github.com/ros-industrial/industrial_core/issues/214>.
* robot_client: add roslaunch testing. Fix #208 <https://github.com/ros-industrial/industrial_core/issues/208>.
* Removed robot_state_visualization.launch, as that would introduce a dependency on rviz. Fix ros-industrial/ros_industrial_issues#50 <https://github.com/ros-industrial/ros_industrial_issues/issues/50>.
* robot_client: add missing dependency on RSP. Fix #209 <https://github.com/ros-industrial/industrial_core/issues/209>.
* Updated the libraries install tags- follow the official documentation. Fix #193 <https://github.com/ros-industrial/industrial_core/issues/193>.
* Added missed test dependencies to rosunit. Fix #205 <https://github.com/ros-industrial/industrial_core/issues/205>.
* Using the 'doc' attribute on 'arg' elements. Added doc strings to all launch file args.
* Merge branch 'kinetic-devel' into urdfdom_headers_fix
* robot_client: make indenting consistent.
* client: build unit tests conditionally
* Contributors: Dmitry Rozhkov, G.A. vd. Hoorn, Harsh Deshpande, Levi Armstrong, Nadia Hammoudeh García, Patrick Beeson, Shaun Edwards, gavanderhoorn
```

## industrial_robot_simulator

```
* Added in better simluation when velocities are provided for trajectory
* robot_simulator: review and fix buildscript and manifest. Fix #207 <https://github.com/ros-industrial/industrial_core/issues/207>.
[...]

PR #20385, opened 2019-02-26
Revert "industrial_core: 0.7.0-0 in 'melodic/distribution.yaml' [bloom]"

gavanderhoorn:
As per title, and reverts ros/rosdistro#20289.

I've opened this here, not to directly merge this, but to request some input on a situation we need to resolve in `ros-industrial/industrial_core`.

---

In ros/rosdistro#20289, a Melodic release of `industrial_core` was merged. That release was Bloom-ed from a newly created `melodic` release branch in `ros-industrial/industrial_core`. The release has also been built by the farm, binaries are available in `shadow-fixed`. So far so good.

Only: the release should have been done from the `kinetic` release branch: all changes for Melodic are bw-compatible with Kinetic and there was no need to branch (we purposefully worked towards such a situation).

In order to reduce maintenance overhead, we'd like to update the Melodic release such that it points to the `kinetic` release branch.

The current plan is to:

 1. delete the `kinetic` release branch
 1. rename the current `melodic` release branch to `kinetic` (the only difference between the two is the changelogs for the `0.7.0` release and the tagging commit)
 1. perform a `0.7.0` release into Kinetic reusing the changelogs and tag already present on the (renamed) `kinetic` release branch

Question: should the changes to the `industrial_core` release repository made by Bloom during the Melodic release be reverted? Alternatively, the release track for Melodic in the `industrial_core` release repository could be manually edited to point to the `kinetic` release branch.

I'm not sure what the best / least invasive approach is, so open to suggestions.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -1942,30 +1942,6 @@ repositories:
       url: https://github.com/ccny-ros-pkg/imu_tools.git
       version: melodic
     status: developed
-  industrial_core:
-    doc:
-      type: git
-      url: https://github.com/ros-industrial/industrial_core.git
-      version: melodic
-    release:
-      packages:
-      - industrial_core
-      - industrial_deprecated
-      - industrial_msgs
-      - industrial_robot_client
-      - industrial_robot_simulator
-      - industrial_trajectory_filters
-      - industrial_utils
-      - simple_message
-      tags:
-        release: release/melodic/{package}/{version}
-      url: https://github.com/ros-industrial-release/industrial_core-release.git
-      version: 0.7.0-0
-    source:
-      type: git
-      url: https://github.com/ros-industrial/industrial_core.git
-      version: melodic
-    status: developed
   interactive_markers:
     doc:
       type: git


gavanderhoorn:
@Levi-Armstrong fyi.

melodic/distribution.yaml
```diff
@@ -1942,30 +1942,6 @@ repositories:
...
-      - industrial_utils
-      - simple_message
-      tags:
-        release: release/melodic/{package}/{version}
-      url: https://github.com/ros-industrial-release/industrial_core-release.git
-      version: 0.7.0-0
-    source:
-      type: git
-      url: https://github.com/ros-industrial/industrial_core.git
-      version: melodic
```
tfoote: If you do the rename you can just change the branch name here instead of reverting.

tfoote (changes requested):
My suggestion would be to go ahead and make the branch rename. Then tag 0.7.1 

Then you can tell bloom to update the target branch in the melodic track at run time or manually change it beforehand to point to the kinetic branch upstream. 

As long as you're releasing a new upstream version there's no worry about conflicts in the release repository that need to be reverted.

And you can push 0.7.1 to kinetic too.

If there's no need for another release you can do the branch rename and release 0.7.0 into kinetic. And next time you need to make a release into Melodic just change the upstream branch. (You can just edit the track now so you don't forget too.)


The upstream branch is just metadata in the track file so it doesn't effect anything else so there's no need to revert this release.

gavanderhoorn:
Thanks @tfoote: you've confirmed what I suspected.

I've just submitted #20392 which releases `0.7.0` into `kinetic`.

I'll manually update the release track in the release repository for Melodic and then delete the `melodic` release branch (as I've already fast-forward `kinetic` release branch to it).

melodic/distribution.yaml
gavanderhoorn: Yes, I understand.

As I wrote above: the PR was more a discussion PR than a real attempt at getting the Melodic release removed.

gavanderhoorn:
Closing this as it was never a "real" PR.

thanks again @tfoote.

PR #20398, opened 2019-02-27
pilz_industrial_motion: 0.4.1-0 in melodic

martiniil:
Increasing version of package(s) in repository `pilz_industrial_motion` to `0.4.1-0`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3831,7 +3831,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/PilzDE/pilz_industrial_motion.git
-      version: kinetic-devel
+      version: melodic-devel
     release:
       packages:
       - pilz_extensions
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -3843,11 +3843,11 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/PilzDE/pilz_industrial_motion-release.git
-      version: 0.4.0-0
+      version: 0.4.1-0
     source:
       type: git
       url: https://github.com/PilzDE/pilz_industrial_motion.git
-      version: kinetic-devel
+      version: melodic-devel
```
clalancette: If we are changing the source branch here, should we also be changing the doc branch above?

martiniil: @clalancette Yes! Thanks for reading carefully.

PR #20415, opened 2019-03-01
Revert "pilz_industrial_motion: 0.4.1-0 in melodic (#20398)"

clalancette:
This reverts commit 7467c8204377d77e8c505c93968d5d42710bb7c6.

Because we reverted moveit back to 0.10, we also need to revert pilz (see http://build.ros.org/job/Mbin_uB64__pilz_trajectory_generation__ubuntu_bionic_amd64__binary/43/console, where it fails to build because the API to moveit has changed).  @martiniil FYI

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3888,7 +3888,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/PilzDE/pilz_industrial_motion.git
-      version: melodic-devel
+      version: kinetic-devel
     release:
       packages:
       - pilz_extensions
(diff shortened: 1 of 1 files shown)


clalancette:
Well, that didn't work.

It looks like we need one of the fixes from release 0.4.1 (https://github.com/PilzDE/pilz_industrial_motion/commit/9c5aa39c95a3b6da3b036064e8fb81b80e3f3f43), but not the API change for moveit 1.0 (https://github.com/PilzDE/pilz_industrial_motion/commit/ae4c0b069486e1f711b318823c04cdd40c4b7a1b).  I'll open a PR to pilz to revert back to the 0.10 API of Moveit instead.

martiniil:
@clalancette Thanks for your effort. We already noticed, that moveit was reverted. But we did not know about your sync plans. We will release asap.
fyi @agutenkunst @hslusarek

clalancette:
> @clalancette Thanks for your effort. We already noticed, that moveit was reverted. But we did not know about your sync plans. We will release asap.

I know this problem is not directly in your package, so I appreciate the help and the understanding.

PR #20426, opened 2019-03-03
Add 'cccc' key.

gavanderhoorn:
As per subject.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -362,6 +362,10 @@ can-utils:
   debian: [can-utils]
   fedora: [can-utils]
   ubuntu: [can-utils]
+cccc:
+  debian: [cccc]
+  gentoo: [dev-util/cccc]
+  ubuntu: [cccc]
 cdk:
   debian: [libcdk5]
   fedora: [cdk]


tfoote (approved):
Ubuntu package: https://packages.ubuntu.com/hu/trusty/cccc
Debian package: https://packages.debian.org/sid/cccc

In the future please link to packages in the description to save time of reviewers doing so after you've already done the search to generate the rules.

gavanderhoorn:
> In the future please link to packages in the description to save time of reviewers doing so after you've already done the search to generate the rules.

My apologies. Seems I missed that bit in [CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

gavanderhoorn:
@nlimpert: FYI.

nuclearsandwich (changes requested):
There is no official Archlinux package for cccc although there is one in the [AUR](https://aur.archlinux.org/packages/cccc/). It also doesn't appear to be packaged in Fedora according to https://apps.fedoraproject.org/packages/s/cccc

Forgot to provide a requested action. For this to be merged, I think those entries have to be removed.

gavanderhoorn:
> There is no official Archlinux package for cccc although there is one in the [AUR](https://aur.archlinux.org/packages/cccc/). It also doesn't appear to be packaged in Fedora according to https://apps.fedoraproject.org/packages/s/cccc

All true. I did an rpm search for `cccc` for Fedora and concluded it would be available based on that.

The Arch stuff I wasn't sure about, but it looked like it was available from AUR. If that is not acceptable then it shouldn't be added.

I've removed them and squashed.

nuclearsandwich:
> it was available from AUR. If that is not acceptable then it shouldn't be added.

Archlinux is a moving target as some packages in the Community repository may lose adoption and drop back into AUR and AUR packages may get adopted by a Trusted User and become packages in the Community repository. I am comfortable being overridden by an alternative criterion but I think that we should strive for keys to be valid and installable by rosdep at the time they're added to the database. There have been a [few](https://github.com/ros-infrastructure/rosdep/pull/276) [attempts](https://github.com/ros-infrastructure/rosdep/issues/560) to add AUR installer support to rosdep although none has been pushed across the finish line.

nuclearsandwich (approved):
Thanks for iterating.

PR #20429, opened 2019-03-04
Exact match tags, not just substrings.

tfoote:
Fixes issues discovered in #20423 #20421 #20420

### DIFF
--- a/test/test_url_validity.py
+++ b/test/test_url_validity.py
@@ -93,7 +93,8 @@ def check_git_remote_exists(url, version, tags_valid=False, commits_valid=False)
     except subprocess.CalledProcessError as ex:
         return (False, 'subprocess call %s failed: %s' % (cmd, ex))
 
-    if 'refs/tags/%s' % version in tag_list:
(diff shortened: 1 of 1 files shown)


test/test_url_validity.py
```diff
@@ -93,7 +93,8 @@ def check_git_remote_exists(url, version, tags_valid=False, commits_valid=False)
     except subprocess.CalledProcessError as ex:
         return (False, 'subprocess call %s failed: %s' % (cmd, ex))
 
-    if 'refs/tags/%s' % version in tag_list:
+    tags = [t for _, t in (l.split(None, 1) for l in tag_list.splitlines())]
```
nuclearsandwich: Wasn't familiar with a split separator of None and had to [look it up](https://docs.python.org/3.7/library/stdtypes.html#str.split).

nuclearsandwich (approved):

cottsay (approved):

test/test_url_validity.py
tfoote: Yeah, it's the default. I wanted to just use the maxsplit 2nd argument, but it doesn't take keyword arguments so I found that None is the same as default.

PR #20475, opened 2019-03-05
Ffmpeg or libav

peci1:
This PR adds key `ffmpeg-or-libav-bin` which tries to address the mess in ffmpeg/libav binaries. Some systems have only ffmpeg (bionic via ffmpeg package), some have only libav (trusty via libav-tools package).

Programs around internet learned that they have to cope with both, try to detect which one is available on runtime, and act accordingly.

I'd like ROS packages to also gain this freedom by depending on just a single rosdep key for all releases. Currently, to always have at least one of them, requires depending on `ffmpeg` and `libav` rosdep keys.

I understand this PR might look a bit weird and I wouldn't be surprised if it didn't get merged...


tfoote (commented):
I'd rather not bother with adding this compatibility layer. The only supported platform on which ffmpeg is not available is trusty which is approaching it's EOL. After it's gone the presence of ffmpeg can be relied upon and I think should be the recommended way to get this functionality. Until there's a driving use case I'd rather not add a rule based on speculative needs.

peci1:
Okay, I'm closing this for now.

PR #20479, opened 2019-03-06
ouster: 0.1.3-0 in 'kinetic/distribution.yaml' [bloom]

amc-nu:
Increasing version of package(s) in repository `ouster` to `0.1.3-0`:

- upstream repository: https://github.com/CPFL/ouster.git
- release repository: https://github.com/CPFL/ouster-release.git
- bloom version: `0.7.2`

## ouster_ros

```
* Updated LICENSE
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -8315,6 +8315,15 @@ repositories:
       type: git
       url: https://bitbucket.org/osrf/gear.git
       version: master
+  ouster:
+    release:
+      packages:
+      - ouster_ros
+      tags:
+        release: release/kinetic/{package}/{version}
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This is ok, but it's highly recommended to have source and doc entries to allow source checkouts and easier development flows.

tfoote (changes requested):
Actually I was browsing through the package and I noticed that it's not following our naming conventions. http://www.ros.org/reps/rep-0144.html

The package should be called ouster_driver not ouster_ros. This should be resolved before it goes onto the buildfarm to avoid trailing artifacts.

amc-nu:
Replaced by #20553

PR #20491, opened 2019-03-07
[dep] Add imgaug-pip.

130s:
https://pypi.org/project/imgaug/

Not sure where to confirm non-Ubuntu platform (as I'm only on Ubuntu) for `pip` package, but looking at other PRs for pip (e.g. https://github.com/ros/rosdistro/pull/20369/files), is listing pypi page sufficient as a proof?

CC @jonlwowski012

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -122,6 +122,19 @@ gunicorn:
   fedora: [python-gunicorn]
   gentoo: [www-servers/gunicorn]
   ubuntu: [gunicorn]
+imgaug-pip:
+  debian:
+    pip:
+      packages: [imgaug]
+  fedora:
+    pip:
+      packages: [imgaug]
+  osx:
+    pip:
+      packages: [imgaug]
+  ubuntu:
+    pip:
+      packages: [imgaug]
 intelhex-pip:
   debian:
     pip:


rosdep/python.yaml
```diff
@@ -122,6 +122,19 @@ gunicorn:
   fedora: [python-gunicorn]
   gentoo: [www-servers/gunicorn]
   ubuntu: [gunicorn]
+imgaug-pip:
```
nuclearsandwich: @ros/rosdeputies I'm not sure if we have strong feelings about whether the key name should be prefixed with python or not.

nuclearsandwich (approved):
Looks good to me assuming that fellow reviewers don't think the key name should change.

nuclearsandwich:
@130s as you may already know because this is a pip package, no package that depends on it can be released to the ROS buildfarm.

rosdep/python.yaml
130s: Looking at [the current status](https://github.com/ros/rosdistro/blob/c6195496facec04d23a7b98d2bbe29f9ab66482b/rosdep/python.yaml) of pkgs with `*-pip` suffix, some are prefixed by `python-` while others aren't. IMO `-pip` makes it clear enough that a package is Python and taken from pip, unless there's a non-pip package that ends with -pip.

tfoote: No strong feeling here.

130s: I have my thoughts updated in https://github.com/ros/rosdistro/pull/21239#issuecomment-492866362

PR #20492, opened 2019-03-07
Added python-imageio

peci1:
This package has recently been added as a system package to Ubuntu, Debian, Opensuse and Gentoo. On other systems, pip has to be used.

Using it in binary released packages in melodic might be a problem since there's no package manger release for Fedora 28 or Debian Strech. 

https://packages.ubuntu.com/search?keywords=python-imageio
https://packages.ubuntu.com/search?keywords=python3-imageio
https://packages.debian.org/search?keywords=python-imageio
https://software.opensuse.org/package/python2-imageio?search_term=python2-imageio
https://packages.gentoo.org/packages/dev-python/imageio
https://pypi.org/project/imageio/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1798,51 +1798,29 @@ python-hypothesis:
     zesty: [python-hypothesis]
     zesty_python3: [python3-hypothesis]
 python-imageio:
-  alpine:
-    pip:
-      packages: [imageio]
-  arch:
-    pip:
-      packages: [imageio]
-  cygwin:
-    pip:
-      packages: [imageio]
   debian:
-    '*':
+    '*': [python-imageio]
+    jessie:
       pip:
         packages: [imageio]
-    buster: [python-imageio]
+    stretch:
+      pip:
+        packages: [imageio]
+    wheezy:
+      pip:
+        packages: [imageio]
   fedora:
     pip:
       packages: [imageio]
-  freebsd:
-    pip:
-      packages: [imageio]
   gentoo: [dev-python/imageio]
-  macports:
-    pip:
-      packages: [imageio]
-  opensuse:
-    '*':
-      pip:
-        packages: [imageio]
-    '15.0': [python-imageio]
-  osx:
-    pip:
-      packages: [imageio]
-  rhel:
-    pip:
-      packages: [imageio]
-  slackware:
-    pip:
-      packages: [imageio]
   ubuntu:
-    '*':
+    '*': [python-imageio]
+    trusty:
       pip:
         packages: [imageio]
-    bionic: [python-imageio]
-    cosmic: [python-imageio]
-    disco: [python-imageio]
+    xenial:
+      pip:
+        packages: [imageio]
 python-imaging:
   alpine: [py-imaging]
   arch: [python2-pillow]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1841,11 +1841,8 @@ python-imageio:
       pip:
         packages: [imageio]
     bionic: [python-imageio]
-    bionic_python3: [python3-imageio]
     cosmic: [python-imageio]
-    cosmic_python3: [python3-imageio]
     disco: [python-imageio]
-    disco_python3: [python3-imageio]
 python-imaging:
   alpine: [py-imaging]
   arch: [python2-pillow]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1797,6 +1797,55 @@ python-hypothesis:
...
+      packages: [imageio]
+  slackware:
+    pip:
+      packages: [imageio]
+  ubuntu:
+    '*':
+      pip:
+        packages: [imageio]
+    bionic: [python-imageio]
+    bionic_python3: [python3-imageio]
```
nuclearsandwich: I don't believe we're using this system for handling python2 versus python3 keys anymore. There are still a number of instances in the database but instead of adding new ones we should use a separate `python3-LIBRARY` key to resolve the python3 dependency. For packages that wish to support python3 and python2 in the same package, using conditional dependencies which [check the `ROS_PYTHON_VERSION`](http://www.ros.org/reps/rep-0149.html#environment-variables) is the recommended approach.

Same for the cosmic_python3 and disco_python3 entries below.

nuclearsandwich (dismissed):
I appreciate the attempt to satisfy every platform.

I don't think we want to mix pip- and system-satisfied dependencies on the same key.

I think it would be easier to review this PR if it was reduced in scope to cover only the system packages for platforms that they're available on.

rosdep/python.yaml
peci1: removed them

peci1:
> I don't think we want to mix pip- and system-satisfied dependencies on the same key.
> 
> I think it would be easier to review this PR if it was reduced in scope to cover only the system packages for platforms that they're available on.

Well, the idea was to make the package available for everybody during source-install, and for the supported systems during binary install. So everybody could make a binary release of a melodic/bionic package depending on python-imageio, and more, he could also "release" the package for older distros for building from source.

As an alternative, I can imagine creating a separate `python-imageio-pip` key, but that would also bring some chaos in.

rosdep/python.yaml
```diff
@@ -1797,6 +1797,52 @@ python-hypothesis:
...
+    pip:
+      packages: [imageio]
+  rhel:
+    pip:
+      packages: [imageio]
+  slackware:
+    pip:
+      packages: [imageio]
+  ubuntu:
+    '*':
```
tfoote: We definitely don't want a wildcard rule with the pip entry. The logic should be reversed and put the pip entries only on the older platforms.

rosdep/python.yaml
```diff
@@ -1797,6 +1797,52 @@ python-hypothesis:
...
+      packages: [imageio]
+  freebsd:
+    pip:
+      packages: [imageio]
+  gentoo: [dev-python/imageio]
+  macports:
+    pip:
+      packages: [imageio]
+  opensuse:
+    '*':
```
tfoote: Same logic reversal here. We don't want to accidentially have new users install the pip version on newer systems. The wildcard should be used for the expected future releases so as not to need to be updated.

rosdep/python.yaml
```diff
@@ -1797,6 +1797,52 @@ python-hypothesis:
...
+    pip:
+      packages: [imageio]
+  arch:
+    pip:
+      packages: [imageio]
+  cygwin:
+    pip:
+      packages: [imageio]
+  debian:
+    '*':
```
tfoote: See other wildcard mentions.

tfoote (changes requested):
@nuclearsandwich  In general it's ok to mix the pip rules if they are only for supporting older platforms. See my comments about the wildcards.

With respect to all the other platforms. Do you have users on many of those platforms? It's great to fill things in but since pip will theoretically work everywhere it should be obvious if someone needs to fill it in. I'd rather not speculatively add them all, especially if there might become a native package on those platforms before there is a user. It's relatively trivial to add them in later if someone needs it on a specific platform. And hopefully when they are integrated it can also be tested.

nuclearsandwich:
> @nuclearsandwich In general it's ok to mix the pip rules if they are only for supporting older platforms.

:ok: Thanks for clarifying that.

peci1:
@tfoote Okay, I did the changes you requested. 

I left fedora there since it is target platform for melodic.

With opensuse, I'm not sure, so I deleted it completely. I found that 15.0 has the binary package, but 15.1 does not... so let's leave it be until someone on opensuse really needs the rule.

tfoote (approved):
Thanks for the updates. lgtm

PR #20505, opened 2019-03-07
Add python-persist-queue-pip

sgloutnikov:
Dependency description: https://pypi.org/project/persist-queue/
Dependency source: https://github.com/peter-wangxu/persist-queue


tfoote (approved):
We can add this though I note that it's proposed for upstream. But I don't see any recent progress: https://salsa.debian.org/python-team/modules/python-persist-queue

PR #20529, opened 2019-03-10
movie_publisher: 1.3.0-1 in 'melodic/distribution.yaml' [bloom]

peci1:
Increasing version of package(s) in repository `movie_publisher` to `1.3.0-1`:

- upstream repository: https://github.com/peci1/movie_publisher.git
- release repository: https://github.com/peci1/movie_publisher-release.git
- bloom version: `0.7.2`

## movie_publisher

```
* Updated for melodic.
* Contributors: Martin Pecka
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3062,6 +3062,21 @@ repositories:
       url: https://github.com/ros-planning/moveit_visual_tools.git
       version: melodic-devel
     status: developed
+  movie_publisher:
+    doc:
+      type: git
+      url: https://github.com/peci1/movie_publisher.git
+      version: melodic-devel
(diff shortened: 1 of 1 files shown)


sloretz:
@nuclearsandwich How are the dependencies in the debian control file for a ROS 1 package populated? Does it use the `package.xml`? If so, what happens if there is a missing key because a dependency is optional, like `python-moviepy-pip` in this case?

peci1:
@sloretz IMO there's no way to correctly depend on a pip package when releasing debian packages.

But the movie_publisher has an opencv fallback, so the fact moviepy is missing is not a problem.

sloretz (approved):

PR #20555, opened 2019-03-13
ouster: 0.1.6-0 in 'kinetic/distribution.yaml' [bloom]

amc-nu:
Increasing version of package(s) in repository `ouster-release` to `0.1.6-0`:

- upstream repository: https://github.com/CPFL/ouster.git
- release repository: https://github.com/CPFL/ouster-release.git
- bloom version: `0.7.2`

## ouster_driver

```
* Renamed ouster_ros to ouster_driver
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -8339,6 +8339,24 @@ repositories:
       type: git
       url: https://bitbucket.org/osrf/gear.git
       version: master
+  ouster-release:
+    doc:
+      type: git
+      url: https://github.com/CPFL/ouster.git
+      version: autoware_branch
+    release:
+      packages:
+      - ouster_driver
+      tags:
+        release: release/kinetic/{package}/{version}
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -8339,6 +8339,24 @@ repositories:
       type: git
       url: https://bitbucket.org/osrf/gear.git
       version: master
+  ouster-release:
```
clalancette: This is kind of a strange name for the top-level repository, especially since it seems to conflate the source repository with the release one.  It wouldn't actually matter for the package names (since they would end up being `ros-kinetic-ouster-driver`), but for our future reviewing sanity I'm going to request that this just be `ouster`.

clalancette (dismissed):

kinetic/distribution.yaml
amc-nu: Thanks for the review. Updated

tfoote (approved):

PR #20560, opened 2019-03-13
add rule for python3-lxml for mac os

mikaelarguedas:
@mjcarroll based on https://github.com/ros2/ros2/issues/656#issuecomment-472614723 it looks like this definition will be needed for users using rosdep on MacOS. Is that right ?

> With macOS: The dependency should be resolved either with rosdep [...]

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4777,6 +4777,9 @@ python3-lark-parser:
 python3-lxml:
   debian: [python3-lxml]
   fedora: [python3-lxml]
+  osx:
+    pip:
+      packages: [lxml]
   ubuntu: [python3-lxml]
 python3-matplotlib:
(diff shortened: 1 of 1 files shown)


tfoote (commented):
lgtm, though will it be python3 by default?

mikaelarguedas:
> lgtm, though will it be python3 by default?

Yeah I had the same concern but saw that matplotlib below had a pip entry so assumed it worked in some cases.
I don't have any macOS machine available to try it on.
Looking at the [wheels available](https://pypi.org/project/lxml/#files) it looks like there are separate binaries for python2 and 3, so it will likely be python2 :/

mikaelarguedas:
I'll assume that this will actually be the Python 2 version and close this PR.

PR #20562, opened 2019-03-13
Add kobuki and kobuki_desktop repos to melodic index

corot:


corot:
:thinking: URL not valid?  I don't understand travis complain....

melodic/distribution.yaml
```diff
@@ -2382,6 +2382,12 @@ repositories:
       url: https://github.com/ros/kdl_parser.git
       version: melodic-devel
     status: maintained
+  kobuki:
+    source:
+      type: git
+      url: https://github.com/yujinrobot/kobuki.git
+      version: melodic
```
tfoote: It's failing CI because there is no `melodic` branch on the repository to check out.

melodic/distribution.yaml
```diff
@@ -2402,6 +2408,12 @@ repositories:
       url: https://github.com/yujinrobot/kobuki_core.git
       version: melodic
     status: maintained
+  kobuki_desktop:
+    source:
+      type: git
+      url: https://github.com/yujinrobot/kobuki_desktop.git
+      version: melodic
```
tfoote: Same issue with the branch name here. On which branch will melodic content be developed?

corot: :no_mouth:   let's imagine this has never happened... 
melodic branches pushed

tfoote: I retriggered it and it passed. Must have been a network glitch ;-)

PR #20581, opened 2019-03-15
Added Python Supervisor

Vuwij:
https://pypi.org/project/supervisor/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3979,6 +3979,16 @@ python-subprocess32:
   ubuntu:
     '*': [python-subprocess32]
     trusty: null
+python-supervisor-pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
This is a package with native package manager support for a long time. We should use that version not the pip version. 

eg: https://packages.ubuntu.com/trusty/admin/supervisor https://packages.debian.org/jessie/supervisor https://src.fedoraproject.org/rpms/supervisor

tfoote:
Actually we already have that rule so closing this: https://github.com/ros/rosdistro/blob/59cdb7758d0cf8a61ed43775715e5266c90c3036/rosdep/base.yaml#L4754

PR #20586, opened 2019-03-17
add python-backoff

tprasadtp:
PyPi: https://pypi.org/project/backoff/1.8.0/#description

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -500,6 +500,23 @@ python-babel:
   fedora: [python-babel]
   gentoo: [dev-python/Babel]
   ubuntu: [python-babel]
+python-backoff:
+  arch: [python2-backoff]
+  debian:
+    pip:
+      packages: [backoff]
+  fedora:
+    pip:
+      packages: [backoff]
+  opensuse:
+    pip:
+      packages: [backoff]
+  osx:
+    pip:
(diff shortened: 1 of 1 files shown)


cottsay (dismissed):
Two things:
* Per [CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules), pip-only keys should end with `-pip`.
* Fix alphabetization errors that were caught by the Travis build:
  ```
  ERR: list out of alphabetical order line 520.  'python-backoff' should come before 'python-backports.ssl-match-hostname'
  ERR: list out of alphabetical order line 575.  'rhel' should come before 'ubuntu'
  ```

tprasadtp:
Done. It looks like system packages are available for python 2 on arch. So I haven't added pip suffix for python 2.
https://aur.archlinux.org/packages/python2-backoff/

nuclearsandwich:
> Done. It looks like system packages are available for python 2 on arch


Packages in the AUR do not count as available system packages for Archlinux as rosdep currently only uses pacman and binary repositories to satisfy dependencies. You can convert the Archlinux key to pip and update the key name to use the `-pip` suffix.

nuclearsandwich (approved):

PR #20621, opened 2019-03-20
aws_ros1_common: 2.0.0-0 in 'melodic/distribution.yaml' [bloom]

AAlon:
Increasing version of package(s) in repository `aws_ros1_common` to `2.0.0-0`:

- upstream repository: https://github.com/aws-robotics/utils-ros1.git
- release repository: https://github.com/aws-gbp/aws_ros1_common-release.git
- bloom version: `0.7.2`

## aws_ros1_common

```
* Remove legacy portions of the ParameterReader API (#9 <https://github.com/aws-robotics/utils-ros1/issues/9>)
* Update ParameterReader API to support ROS1/ROS2 (#8 <https://github.com/aws-robotics/utils-ros1/issues/8>)
  * Revert "Revert ParameterReader change (#5 <https://github.com/aws-robotics/utils-ros1/issues/5>)"
  * refactor based on new ParameterPath object design
* Revert ParameterReader change (#5 <https://github.com/aws-robotics/utils-ros1/issues/5>)
  * Revert "Parameter Namespacing: Refactoring using the ParameterPath object. (#3 <https://github.com/aws-robotics/utils-ros1/issues/3>)"
  This reverts commit 295f157b32a321e230ef1c7f616ad5abd1bece5e.
  https://github.com/aws-robotics/utils-common/issues/15
* Parameter Namespacing: Refactoring using the ParameterPath object. (#3 <https://github.com/aws-robotics/utils-ros1/issues/3>)
  * Refactoring using the ParameterPath object.
  * Minor test fixes in parameter_reader_test.cpp
  * Adding failure test case for Ros1NodeParameterReader.
  * Bumping major version in package.xml
  * address comments in PR
  * update .travis.yml to build dependencies from latest source
* Contributors: AAlon, M. M
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -285,6 +285,21 @@ repositories:
       url: https://github.com/aws-robotics/utils-common.git
       version: master
     status: maintained
+  aws_ros1_common:
+    doc:
+      type: git
+      url: https://github.com/aws-robotics/utils-ros1.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/aws-gbp/aws_ros1_common-release.git
+      version: 2.0.0-0
+    source:
+      type: git
+      url: https://github.com/aws-robotics/utils-ros1.git
+      version: master
+    status: maintained
   backward_ros:
     release:
       tags:


AAlon:
The [build ](http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__aws_ros1_common__ubuntu_bionic_amd64__binary/1/console)for this seems to fail with:

```
08:57:55   1645 'Unpacking ros-melodic-catkin (0.7.17-0bionic.20190318.174702) ...'
08:57:55   1646 'dpkg: error processing archive /tmp/apt-dpkg-install-VYbYEW/326-ros-melodic-catkin_0.7.17-0bionic.20190318.174702_amd64.deb (--unpack):'
08:57:55   1647 ' trying to overwrite '/opt/ros/melodic/.catkin', which is also in package ros-melodic-aws-common 2.0.0-1bionic.20190320.031845'
08:57:55   1648 'dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)'

08:57:55   1739 'Errors were encountered while processing:'
08:57:55   1740 ' /tmp/apt-dpkg-install-VYbYEW/326-ros-melodic-catkin_0.7.17-0bionic.20190318.174702_amd64.deb'
08:57:55   1741 'E: Sub-process /usr/bin/dpkg returned an error code (1)'
```
@clalancette @tfoote Any clue what might be the problem? 
Thanks!

tfoote:
The short answer is that your package is generating a `/opt/ros/melodic/.catkin' which is colliding with the  one that catkin installs. I'll dig deeper to see if I can see where that's being generated.

tfoote:
The difference is that you appear to not to have the default ROS1 arguments pass into the rules file: 

Compare: https://github.com/aws-gbp/aws_ros1_common-release/blob/3534e704ecd5ff02a466e44375712b67c870da59/debian/rules#L29-L32 which is the standard ROS1 options for dh_auto_configure to https://github.com/aws-gbp/aws_common-release/blob/3fa5902f8899eaf37b4def6a06ccaea60d0f77dc/debian/rules#L29

AAlon:
> The difference is that you appear to not to have the default ROS1 arguments pass into the rules file:
> 
> Compare: https://github.com/aws-gbp/aws_ros1_common-release/blob/3534e704ecd5ff02a466e44375712b67c870da59/debian/rules#L29-L32 which is the standard ROS1 options for dh_auto_configure to https://github.com/aws-gbp/aws_common-release/blob/3fa5902f8899eaf37b4def6a06ccaea60d0f77dc/debian/rules#L29

I've looked and this discrepancy is nothing new, it was this way in previous releases as well. It might be because `aws_common` has `<build_type>cmake</build_type>` i.e. it's not coupled with a ROS distro? Not sure.

tfoote:
Yes, the difference in those flags are what's expected from a [catkin](https://github.com/ros-infrastructure/bloom/blob/53f129595d254e4406432c9ef63bf451d252d751/bloom/generators/debian/templates/catkin/rules.em#L30) versus [cmake](https://github.com/ros-infrastructure/bloom/blob/53f129595d254e4406432c9ef63bf451d252d751/bloom/generators/debian/templates/cmake/rules.em#L29) build type. 

However I see that the package is now using catkin in the CMakeLists.txt 

https://github.com/aws-robotics/utils-common/blob/756b8bf89006500700828bdf88e7da12f12cbedd/aws_common/CMakeLists.txt#L65

Since catkin appears to be in the cmake by default I'd suggest switching the build type over. Otherwise you would need to change the default build to not trigger catkin.

AAlon:
> Yes, the difference in those flags are what's expected from a [catkin](https://github.com/ros-infrastructure/bloom/blob/53f129595d254e4406432c9ef63bf451d252d751/bloom/generators/debian/templates/catkin/rules.em#L30) versus [cmake](https://github.com/ros-infrastructure/bloom/blob/53f129595d254e4406432c9ef63bf451d252d751/bloom/generators/debian/templates/cmake/rules.em#L29) build type.
> 
> However I see that the package is now using catkin in the CMakeLists.txt
> 
> https://github.com/aws-robotics/utils-common/blob/756b8bf89006500700828bdf88e7da12f12cbedd/aws_common/CMakeLists.txt#L65
> 
> Since catkin appears to be in the cmake by default I'd suggest switching the build type over. Otherwise you would need to change the default build to not trigger catkin.

Aha, well we're using it solely for the test macros, it definitely shouldn't create a marker file upon `find_package`.
I'm trying to work around that by manipulating some variables, like setting `_CATKIN_CONFIG_ALL_INCLUDED_`, but that prevents the `catkin_destinations` file from being included (which is a prerequisite of the gtest macros) :\


If we do change the build type to `catkin`, would this affect dependent packages? e.g. `cloudwatch_logs_common` depends on `aws_common`, would we need to change its build type from `cmake` to `catkin` as well? 

Btw if we go with this approach we'll need to use conditionals like the following, which I believe are supported:
```
<build_type condition="$ROS_VERSION == 1">catkin</build_type>
<build_type condition="$ROS_VERSION == 2">ament_cmake</build_type>
```

AAlon:
Actually, I think just `set(CATKIN_INSTALL_INTO_PREFIX_ROOT FALSE)` before `find_package(catkin)` could solve it. Not sure how to test it though seeing as the prerelease builds succeeded.

AAlon:
https://github.com/ros/rosdistro/pull/20654
https://github.com/ros/rosdistro/pull/20653

PR #20639, opened 2019-03-20
e100_base-relase: 0.0.0-0 in 'kinetic/distribution.yaml' [bloom]

brightenlee:
Increasing version of package(s) in repository `e100_base-relase` to `0.0.0-0`:

- upstream repository: https://github.com/gaitech-robotics/e100_base.git
- release repository: https://github.com/gaitech-robotics/e100_base-release.git
- bloom version: `0.7.2`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2464,7 +2464,7 @@ repositories:
       url: https://github.com/tork-a/dynpick_driver.git
       version: master
     status: maintained
-  e100_base-relase:
+  e100_base:
     doc:
       type: git
       url: https://github.com/gaitech-robotics/e100_base-release.git
@@ -2478,7 +2478,7 @@ repositories:
       version: 0.0.0-0
     source:
       type: git
-      url: https://github.com/gaitech-robotics/e100_base-release.git
+      url: https://github.com/gaitech-robotics/e100_base.git
       version: kinetic-devel
     status: developed
   earth_rover_localization:
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2464,6 +2464,23 @@ repositories:
       url: https://github.com/tork-a/dynpick_driver.git
       version: master
     status: maintained
+  e100_base-relase:
+    doc:
+      type: git
+      url: https://github.com/gaitech-robotics/e100_base-release.git
+      version: kinetic-devel
+    release:
+      packages:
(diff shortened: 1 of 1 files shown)


clalancette:
Holding for [sync](https://discourse.ros.org/t/preparing-for-kinetic-sync-2019-03-19/8367) .

kinetic/distribution.yaml
```diff
@@ -2464,6 +2464,23 @@ repositories:
       url: https://github.com/tork-a/dynpick_driver.git
       version: master
     status: maintained
+  e100_base-relase:
```
clalancette: I'm guessing this should be `e100_base-release`, but honestly, we don't need to even put `-release` in here at all.  I'd suggest just `e100_base`.

kinetic/distribution.yaml
```diff
@@ -2464,6 +2464,23 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/gaitech-robotics/e100_base-release.git
+      version: kinetic-devel
```
clalancette: This failed CI.  The likely reason is that the doc and the source entry should point to the upstream sources, not the release repository.  If there is no upstream source repository, then just elide the doc and source entries.

clalancette (changes requested):

kinetic/distribution.yaml
```diff
@@ -2464,6 +2464,23 @@ repositories:
       url: https://github.com/tork-a/dynpick_driver.git
       version: master
     status: maintained
+  e100_base:
+    doc:
+      type: git
+      url: https://github.com/gaitech-robotics/e100_base.git
```
tfoote: This repository does not appear to exist. Maybe it's private?

tfoote (changes requested):

kinetic/distribution.yaml
BrightenLee: It's private now. Is it necessary to change to public to register?

tfoote: It's necessary to be public to be indexed.

If it's private we can't check out the source code and generate the documentation, nor run continuous integration. Note that your release repository contains snapshots of the repository at the tag released so it can be built on the buildfarm so keeping the upstream repository private doesn't make a lot of sense.

BrightenLee: I'll request it again after modification.

Issue #1339, opened 2013-07-12
add verification that ubuntu rosdeps are installable from the ROS repos


tfoote:
This will help avoid typos.

dirk-thomas:
Closing due to age and inactivity.

PR #20643, opened 2019-03-20
Modernized libflann ubuntu rule.

peci1:
There were no rules for the modern distros, whereas there were plenty of (now) uselsess rules for discontinued ones.

https://packages.ubuntu.com/source/bionic/flann

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1713,11 +1713,15 @@ libfftw3:
   ubuntu: [libfftw3-3, libfftw3-dev]
 libflann:
   arch: [flann]
-  debian: [libflann1.7]
+  debian:
+    '*': [libflann1.9]
+    jessie: [libflann1.8]
+    wheezy: [libflann1.7]
   fedora: [flann]
   gentoo: [sci-libs/flann]
   macports: [flann]
   ubuntu:
+    '*': [libflann1.9]
     precise: [libflann1]
     quantal: [libflann1.7]
     raring: [libflann1.7]


tfoote (changes requested):
I'd rather just add the wildcard rule and not remove the old rules as we do have people using this on eol platforms and we don't want to break them if possible. Keeping the few extra lines around isn't much overhead and the wildcard will avoid future changes.

peci1:
Okay, I updated the rule to just add the asterisk for ubuntu, and I also updated debian packages according to https://packages.debian.org/search?suite=default&arch=amd64&searchon=names&keywords=libflann .

Is there actually no automatic mechanism on the ROS build farm that would validate at least if rosdep apt keys exist on a new ubuntu distro supported by rosdistro?

tfoote (approved):
Thanks. lgtm.

There's a long standing ticket to add this #1339 With the availability of docker it's possible for us to do this but we need to scope it appropriately such that it doesn't eat up too many resources.

rosdep/base.yaml
```diff
@@ -1713,11 +1713,16 @@ libfftw3:
   ubuntu: [libfftw3-3, libfftw3-dev]
 libflann:
   arch: [flann]
-  debian: [libflann1.7]
+  debian:
+    '*': [libflann1.9]
+    jessie: [libflann1.8]
+    squeeze: [libflann1.7]
```
sloretz: I don't see `libflann1.7` in squeeze. https://packages.debian.org/squeeze/allpackages

peci1: My bad in detecting when this package started in Debian. I see it in wheezy, though, so I'll just remove the squeeze entry.

sloretz (approved):

PR #20666, opened 2019-03-23
add lark-parser for debian via pip

christianrauch:
Add pip package `lark-parser` (https://pypi.org/project/lark-parser/) for Debian Stretch.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4776,6 +4776,11 @@ python3-gitlab:
       pip:
         packages: [python-gitlab]
 python3-lark-parser:
+  debian:
+    '*': [lark-parser]
+    stretch:
+      pip:
+        packages: [lark-parser]
   fedora:
     '*': [python-lark-parser]
     '28': null


tfoote (changes requested):
This rule is specifically for python3. We don't have a good way of capturing that in the pip rules and it looks like we haven't forced that wrt to other rules. 

But I can't find any debian package lark-parser and it should be made sure that it's a python3 package too which from the name doesn't look likely.

christianrauch:
Yes, it's a python3 package.
I am not sure how I can force the python3 version via the rosdep file. It's up to the version of pip which python version of the package is installed. But since ROS2 defaults to python3 (and ROS1 to python2), the corresponding python version should be used to install pip packages.

tfoote:
I agree that the pip rule is functional. But the debian rule is incorrect as that package is both the wrong python version and doesn't exist for either version.

nuclearsandwich:
With https://github.com/ros/rosdistro/pull/22234 merged I would propose that we close this PR.

christianrauch:
Is `python3-lark-parser` now a Debian package that has been manually added to the ROS repo for stretch and buster?

In that case, the `pip3` installation of the package is not required and this PR can be closed.

nuclearsandwich:
> Is `python3-lark-parser` now a Debian package that has been manually added to the ROS repo for stretch and buster?

Yes. It's both the ROS and ROS 2 repositories and their respective testing repos for Ubuntu Bionic and Debian Stretch and Buster. For future distributions I'm going to try and get the packages submitted upstream.

PR #20673, opened 2019-03-23
Added libxml++2.6 to Bionic

n30phyte:
Add libxml++2.6 to the base rosdep file for Ubuntu 18.04 "Bionic"

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3587,6 +3587,7 @@ libxml++-2.6:
   fedora: [libxml++, libxml++-devel]
   gentoo: ['dev-cpp/libxmlpp:2.6']
   ubuntu:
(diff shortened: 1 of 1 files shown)


tfoote (approved):
looking at this the dev package directly depends on the library package so there's no need to call out the specific library version. 

This holds for all supported versions of ubuntu: 
Trusty: https://packages.ubuntu.com/trusty/libxml++2.6-dev through Disco: https://packages.ubuntu.com/disco/libxml++2.6-dev

The debian rule is actually invalid due to the new library name at the moment. So I've proposed simplifying it too. 
https://packages.debian.org/buster/libxml++2.6-dev

This makes the rule more compact and maintainable going forward.

This rule is also noteably old and doesn't follow our conventions of including the `-dev` in the name, but I've chosen not to update it to avoid breaking any existing users.

PR #20712, opened 2019-03-26
Docker

Vuwij:
https://packages.ubuntu.com/xenial/docker.io
https://packages.ubuntu.com/xenial/docker-compose

https://apps.fedoraproject.org/packages/docker
https://apps.fedoraproject.org/packages/docker-compose

https://packages.debian.org/buster/docker.io
https://packages.debian.org/stretch/docker-compose

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -606,13 +606,13 @@ docker:
   debian: [docker]
   fedora: [docker]
   ubuntu: [docker]
-docker.io:
-  debian: [docker.io]
-  ubuntu: [docker.io]
 docker-compose:
   debian: [docker-compose]
   fedora: [docker-compose]
   ubuntu: [docker-compose]
+docker.io:
+  debian: [docker.io]
+  ubuntu: [docker.io]
 doxygen:
   arch: [doxygen]
   debian: [doxygen]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -607,7 +607,7 @@ docker-compose:
   fedora: [docker-compose]
   ubuntu: [docker-compose]
 docker.io:
-  debian: 
+  debian:
     '*': [docker.io]
     jessie: null
     stretch: null
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -602,16 +602,17 @@ dnsmasq:
   fedora: [dnsmasq]
   gentoo: [net-dns/dnsmasq]
   ubuntu: [dnsmasq]
-docker:
-  debian: [docker]
-  fedora: [docker]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -602,6 +602,17 @@ dnsmasq:
   fedora: [dnsmasq]
   gentoo: [net-dns/dnsmasq]
   ubuntu: [dnsmasq]
+docker:
```
nuclearsandwich: If you don't need the System Tray docklet which is installed via the `docker` package on Debian and Ubuntu this key can be removed.

rosdep/base.yaml
```diff
@@ -602,6 +602,17 @@ dnsmasq:
...
+  debian: [docker]
+  fedora: [docker]
```
nuclearsandwich: The `docker` packages in Debian and Ubuntu refer to something entirely different from the `docker` package in Fedora. Fedora's `docker` package belongs with the `docker.io` keys for Debian and Ubuntu below.

nuclearsandwich (changes requested):

nuclearsandwich:
@Vuwij are you intentionally adding keys for the `docker` package in Stretch and Xenial which is described as a "System tray for KDE3/GNOME2 docklet applications" and is not related to the Moby/Docker project?

rosdep/base.yaml
```diff
@@ -602,6 +602,17 @@ dnsmasq:
...
+  ubuntu: [docker]
+docker-compose:
+  debian: [docker-compose]
+  fedora: [docker-compose]
+  ubuntu: [docker-compose]
+docker.io:
+  debian: [docker.io]
```
tfoote: From the link it appears that docker.io is only available on buster and newer. We should explicitly set None for the other currently supported distros so people don't get surprised when it doesn't work. And use a wildcard for buster and newer.

rosdep/base.yaml
tfoote: This does look like it's unrelated.

tfoote (changes requested):

rosdep/base.yaml
Vuwij: Is this correct now?

Vuwij:
Is it correct now?

Vuwij:
@tfoote @nuclearsandwich should it be ok now?

Vuwij (commented):
Ok

Vuwij (commented):
Ok

nuclearsandwich (approved):

Vuwij:
bump

tfoote (approved):

PR #20719, opened 2019-03-26
Adding doosan-robotics to documentation index for distro

doosan-robotics:
Adding doosan-robotics to documentation index for distro


nuclearsandwich:
@doosan-robotics there appear to be additional commits being pushed to your pull request branches which we cannot merge. If you can, the easiest thing to do is to remove the commits after e38f2e5 from the branch so we can review just that change. Continually closing and re-opening pull requests is beginning to be disruptive for our maintenance team.

PR #20722, opened 2019-03-26
Adding doosan-robotics to documentation index for distro

doosan-robotics:
Adding doosan-robotics to documentation index for distro

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2328,6 +2328,17 @@ repositories:
       url: https://github.com/ct2034/dockeROS.git
       version: master
     status: developed
+  doosan-robot:
+    doc:
+      type: git
+      url: https://github.com/doosan-robotics/doosan-robot.git
+      version: master
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/doosan-robotics/doosan-robot.git
+      version: master
+    status: developed
   dr_base:
     release:
       packages:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -2328,6 +2328,30 @@ repositories:
...
+      - dsr_control
+      - dsr_description
+      - dsr_example_cpp
+      - dsr_example_py
+      - dsr_gazebo
+      - moveit_config_m0609
+      - moveit_config_m0617
+      - moveit_config_m1013
+      - moveit_config_m1509
+      - robotiq_description
```
gavanderhoorn: Is this correct?

kinetic/distribution.yaml
```diff
@@ -2328,6 +2328,30 @@ repositories:
...
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/doosan-robotics/doosan-robot.git
```
tfoote: This is not a release url. The release section should be populated by the bloom release tool after being run. The release stanza should not be filled in manually.

kinetic/distribution.yaml
tfoote: It does look to be a package in the repository, but since it's a fork it should have a name related to the scope of this package such as dsr_robotiq_description

tfoote (changes requested):

doosan-robotics:
Thank you for your comment.
I erased the release stanza. Is there anything else I need to fix?

tfoote:
You're passing the validation, but the comment about the package robotiq_description should be resolved in the source repository to avoid name squatting with your forked version.

PR #20723, opened 2019-03-26
Added libuv-dev to Ubuntu Bionic

AndriyPt:


rosdep/base.yaml
```diff
@@ -3395,6 +3395,7 @@ libuv-dev:
     xenial: [libuv0.10-dev]
     yakkety: [libuv0.10-dev]
     zesty: [libuv0.10-dev]
+    bionic: [libuv0.10-dev]
```
nuclearsandwich: As indicated by the CI failure and specified in the [contribution guide](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules) please move the entry for `bionic` to the top to preserve the alphabetic ordering.

nuclearsandwich (changes requested):

nuclearsandwich:
For other reviewers: The merge commits will be squashed out when this PR is approved and merged. We can safely ignore them for now. I'm not sure if there's a technical reason that we can't use the virtual package `libuv-dev` to condense this rule but the apparent reason for the different package names is that the `libuv-dev` package used to be concrete but is now virtual.

AndriyPt:
@nuclearsandwich thank you for remarks. Changed order.

nuclearsandwich (approved):

PR #20729, opened 2019-03-27
Add python-qt5-bindings-quick

furushchev:
For installing `python-pyqt5.qtquick` from `rosdep install`.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3267,6 +3267,9 @@ python-qt5-bindings-qsci:
   debian: [python-pyqt5.qsci]
   fedora: [python2-qscintilla-qt5]
   ubuntu: [python-pyqt5.qsci]
+python-qt5-bindings-quick:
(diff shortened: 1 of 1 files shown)


tfoote (approved):
https://packages.ubuntu.com/xenial/python-pyqt5.qtquick
https://packages.debian.org/sid/python-pyqt5.qtquick

It's not our standard, which is to match the debian rules, but it follows the same pattern of other bindings so lgtm.

nuclearsandwich:
> It's not our standard to match the debian rules

I want to make sure I have my head on straight. Do you mean here that our standard is to match the Debian package names and this defies it. Or that our standard practice is not to match the Debian names?

tfoote:
I"m saying that this does not match our standard. But is consistent with the existing no-standard pattern. I added punctuation to better express my comment.

furushchev:
@tfoote @nuclearsandwich Thanks for review and merging!

Issue #20591, opened 2019-03-18
libgtkmm install error on Ubuntu


Tiryoh:
There was an error while installing `libgtkmm` on Ubuntu Xenial + ROS Kinetic.
It is defined to install `libgtkmm-dev` on Ubuntu, but it seems like there are only `libgtkmm-2.4-dev` and `libgtkmm-3.0-dev` on Ubuntu repository (and maybe on Debian, too.)

https://github.com/ros/rosdistro/blob/5b3b503bcdef5827853b20f9efc145012c603496/rosdep/base.yaml#L2059-L2064

We think the rules should be defined as `libgtkmm` and `libgtkmm3`, or `libgtkmm2` and `libgtkmm3`.

* libgtkmm2
  * arch: https://www.archlinux.org/packages/extra/x86_64/gtkmm/
  * debian: https://packages.debian.org/stretch/libgtkmm-2.4-dev
  * fedora: https://apps.fedoraproject.org/packages/gtkmm24-devel
  * gentoo: https://packages.gentoo.org/packages/dev-cpp/gtkmm
  * ubuntu: https://packages.ubuntu.com/xenial/libgtkmm-2.4-dev
* libgtkmm3
  * arch: https://www.archlinux.org/packages/extra/x86_64/gtkmm3/
  * debian: https://packages.debian.org/stretch/libgtkmm-3.0-dev
  * fedora: https://apps.fedoraproject.org/packages/gtkmm30-devel
  * ubuntu: https://packages.ubuntu.com/xenial/libgtkmm-3.0-dev

## Console log

```
$ LANG=C rosdep install -r -y --from-paths .
executing command [sudo -H apt-get install -y libgtkmm-dev]
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package libgtkmm-dev
ERROR: the following rosdeps failed to install
apt: command [sudo -H apt-get install -y libgtkmm-dev] failed
apt: Failed to detect successful installation of [libgtkmm-dev]
```
```
$ LANG=C sudo apt search libgtkmm-
Sorting... Done
Full Text Search... Done
libgtkmm-2.4-1v5/xenial 1:2.24.4-2 amd64
  C++ wrappers for GTK+ (shared libraries)

libgtkmm-2.4-dbg/xenial 1:2.24.4-2 amd64
  C++ wrappers for GTK+ (debug symbols)

libgtkmm-2.4-dev/xenial 1:2.24.4-2 amd64
[...]

PR #20738, opened 2019-03-28
Fix libgtkmm install error on Ubuntu/Debian

Tiryoh:
Fix libgtkmm install error on Ubuntu on the current rosdep rule.
This PR is related to the issue #20591.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2070,7 +2070,7 @@ libgstreamer1.0-dev:
   ubuntu: [libgstreamer1.0-dev]
 libgtkmm:
   arch: [gtkmm]
-  debian: [libgtkmm-dev]
+  debian: [libgtkmm-2.4-dev]
   fedora: [gtkmm24]
   gentoo: [dev-cpp/gtkmm]
   ubuntu: [libgtkmm-2.4-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2073,7 +2073,7 @@ libgtkmm:
   debian: [libgtkmm-dev]
   fedora: [gtkmm24]
   gentoo: [dev-cpp/gtkmm]
-  ubuntu: [libgtkmm-dev]
+  ubuntu: [libgtkmm-2.4-dev]
 libgts:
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
The Debian package `libgtkmm-dev` doesn't exist either.
Of the listed packages, only Gentoo's package provides gtkmm2.4 or 3.x. I'd suggest renaming this rosdep key to libgtkmm-2.4-dev which matches our current practice of naming keys for Debian/Ubuntu packages. Aside from Debian, (and Gentoo as a special case) all other platforms resolve this key to the 2.x version of gtkmm. (Archlinux uses a separate gtkmm3 package for the 3.x)

Tiryoh:
> Aside from Debian, (and Gentoo as a special case) all other platforms resolve this key to the 2.x version of gtkmm.

Thanks for the information.
I don't have a Debian environment so I can't test the rosdep rule for Debian.
Is it OK for just editing without running test described [here](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions)?

nuclearsandwich:
> I don't have a Debian environment with ROS installed, so I can't test the rosdep rule for Debian OS.
> Is it OK for just editing without running test described [here](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions)?

As long as the package the key targets is linked for review (https://packages.debian.org/stretch/libgtkmm-2.4-dev for Debian stable) that should be sufficient.

tfoote:
Yeah, it does make sense to update it to use the standard key which matches the debian/ubuntu rule. And we should do a best effort to fill in the other platforms based on online listings.

Tiryoh:
I'll update the Debian rule too.

nuclearsandwich:
> I thought I needed to do an installation test after updating the rules...

It's always a good idea to test what you can easily test. But where it's not easy to test I have a lot of trust in the web-based package databases for Debian and Ubuntu. Thanks for doing your best!

nuclearsandwich (approved):

PR #20739, opened 2019-03-28
Adding dynamixel_control_hw to documentation index for distro

PedroDesRobots:
I would like dynamixel_control_hw to be indexed and documented on ros.org.
Thank you in advance.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2434,6 +2434,11 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/dynamixel-workbench-msgs.git
       version: kinetic-devel
     status: developed
+  dynamixel_control_hw:
+    doc:
+      type: git
+      url: https://github.com/resibots/dynamixel_control_hw.git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -2434,6 +2434,11 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/dynamixel-workbench-msgs.git
       version: kinetic-devel
     status: developed
+  dynamixel_control_hw:
+    doc:
+      type: git
+      url: https://github.com/resibots/dynamixel_control_hw.git
+      version: kinetic-devel
```
tfoote: There does not appear to be a kinetic-devel branch, I'd guess that you want to index master?

```suggestion
      version: master
```

PedroDesRobots: Yes exactly, thank you !

tfoote: Please accept my suggestion or make the same change yourself to this branch and then we can merge this.

PR #20751, opened 2019-03-29
Revert "p2os: 2.2.0-0 in 'melodic/distribution.yaml' [bloom] (#20576)"

clalancette:
This reverts commit e6cdad7c78d7b4f3aed16eec3deae60421705f00.

Fixes the errors on the buildfarm so we can unblock the Melodic sync.  @allenh1 FYI.


allenh1:
@clalancette got it -- I'll re-release after the next sync.

clalancette:
> @clalancette got it -- I'll re-release after the next sync.

Sounds good, thanks.

PR #20759, opened 2019-03-30
Support python-tilestache on Fedora 28 and newer

cberner:
https://apps.fedoraproject.org/packages/python-tilestache


cottsay (approved):
LGTM - Thanks

PR #20779, opened 2019-04-01
Python3-pip

Vuwij:
Not completely sure about this one

https://packages.ubuntu.com/xenial/python3-pip
https://apps.fedoraproject.org/packages/python3-pip
https://packages.debian.org/stretch/python3-pip

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4412,10 +4412,6 @@ psutils:
   fedora: [psutils]
   gentoo: [psutils]
   ubuntu: [psutils]
-python3-pip:
-  debian: [python3-pip]
-  fedora: [python3-pip]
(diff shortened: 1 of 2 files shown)


tfoote:
What are you not sure about?

Vuwij:
> What are you not sure about?

Whether Is it ok to install python3 in rosdep

tfoote (changes requested):
It's fine to install python3 system dependencies. Eventually all of them will be python3. 

Though there is a separate python.yaml where we've been putting python dependencies. Could you move the key into that file for consistency. Otherwise this looks fine.

tfoote (approved):

cottsay (approved):

PR #20832, opened 2019-04-09
Remove source entry for ubiquity_motor

rohbotics:
The buildfarm CI is redundant for us so the emails are a little annoying


tfoote (changes requested):
Instead of removing the entry please just turn off the ci by adding a line `test_commits: false` this will allow people to use the rosinstall generator and other source tools but turn off the testing.

rohbotics:
@tfoote Sorry for the delay. Done

tfoote (approved):

PR #20792, opened 2019-04-04
add gnuplot-x11

yuki-asano:

PR #20851, opened 2019-04-11
add gnuplot x11 (second time)

yuki-asano:
from https://github.com/ros/rosdistro/pull/20792


tfoote (approved):
For easier reference:
ubuntu: https://packages.ubuntu.com/en/trusty/gnuplot-x11
debian: https://packages.debian.org/en/jessie/gnuplot-x11

PR #20869, opened 2019-04-11
osrf_testing_tools_cpp: 1.2.0-1 in 'dashing/distribution.yaml' [bloom]

nuclearsandwich:
Increasing version of package(s) in repository `osrf_testing_tools_cpp` to `1.2.0-1`:

- upstream repository: https://github.com/osrf/osrf_testing_tools_cpp.git
- release repository: https://github.com/ros2-gbp/osrf_testings_tools_cpp-release.git
- bloom version: `0.8.0.dev2`

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -184,6 +184,25 @@ repositories:
       url: https://github.com/osrf/osrf_pycommon.git
       version: master
     status: maintained
+  osrf_testing_tools_cpp:
+    doc:
+      type: git
+      url: https://github.com/osrf/osrf_testing_tools_cpp.git
+      version: master
+    release:
+      packages:
+      - osrf_testing_tools_cpp
+      - test_osrf_testing_tools_cpp
+      tags:
+        release: release/dashing/{package}/{version}
+      url: https://github.com/ros2-gbp/osrf_testings_tools_cpp-release.git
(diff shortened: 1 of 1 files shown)


dashing/distribution.yaml
```diff
@@ -184,6 +184,25 @@ repositories:
       url: https://github.com/osrf/osrf_pycommon.git
       version: master
     status: maintained
+  osrf_testing_tools_cpp:
+    doc:
+      type: git
+      url: https://github.com/osrf/osrf_testing_tools_cpp.git
+      version: master
```
sloretz: will this get a `dashing` branch in the future?

sloretz (approved):
I'm not familiar enough with bootstrapping a ROS distro to know if this is intentionally not being merged, but it LGTM

dashing/distribution.yaml
nuclearsandwich: Yeah, when the actual dashing release is out we can cut all the dashing branches. I am not volunteering to keep dashing and master branches in sync before then.

nuclearsandwich:
> I'm not familiar enough with bootstrapping a ROS distro to know if this is intentionally not being merged, but it LGTM

Only reason this was left around is that I had to go make dinner and walked away right after running bloom :stuck_out_tongue_closed_eyes:

PR #20945, opened 2019-04-16
Added fping to base.yaml

HEllRZA:
We use [fping](https://www.fping.org/) in one of our projects. Nice way to ping multiple hosts very fast.

Package sources (I only tested ubuntu) but i think they should be correct.
[arch](https://www.archlinux.de/packages?search=fping)
[debian](https://packages.debian.org/search?keywords=fping)
[fedora](https://rpms.remirepo.net/rpmphp/zoom.php?rpm=fping)
[gentoo](https://packages.gentoo.org/packages/net-analyzer/fping)
[macports](https://www.macports.org/ports.php?by=library&substr=fping)
[ubuntu](https://packages.ubuntu.com/xenial/net/fping)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -792,6 +792,13 @@ fluid:
   macports: [fltk]
   opensuse: [fltk-devel]
   ubuntu: [fluid]
+fping:
+  arch: [fping]
+  debian: [fping]
+  fedora: [fping]
+  gentoo: [net-analyzer/fping]
+  macports: [fping]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@HEllRZA on your project do you use macports? Otherwise I might leave it off or prefer the homebrew formula for MacOS dependencies.

nuclearsandwich (approved):
Looks good pending the question about macports usage.

HEllRZA:
I don't need macports.

PR #20946, opened 2019-04-16
make python-tk more symmetric on fedora

tfoote:
Following up #20944

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4251,7 +4251,7 @@ python-tinydb-pip:
 python-tk:
   arch: [python2, tk]
   debian: [python-tk]
-  fedora: [blt, tcl, tix, tk]
+  fedora: [python-tkinter]
   ubuntu:
     '*': [python-tk]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -4251,7 +4251,7 @@ python-tinydb-pip:
 python-tk:
   arch: [python2, tk]
   debian: [python-tk]
-  fedora: [blt, tcl, tix, tk]
+  fedora: [python-tkinter]
```
sloretz: Where's a good place to look up fedora packages? I see `python2-tkinter` here:
https://pkgs.org/download/python2-tkinter

tfoote: I usually use https://apps.fedoraproject.org/packages

pkgs.org is a 3rdparty aggregator, it's usually pretty accurate though.

You're right that it's called python2-tkinter

https://apps.fedoraproject.org/packages/python2-tkinter

sloretz (approved):

cottsay (approved):

PR #20998, opened 2019-04-23
Add potrace to rosdistro

mvollrath:
[Debian](https://packages.debian.org/search?keywords=potrace)
[Ubuntu](https://packages.ubuntu.com/source/xenial/potrace)

Using this to vectorize OccupancyGrids for visualization.


cottsay (approved):
Thanks! I added [Fedora/EPEL](https://apps.fedoraproject.org/packages/potrace) rules as well.

PR #20999, opened 2019-04-23
Added UC Berkeley AUTOLAB dependencies to python.yaml

visatish:
We have a few core libraries that we use here at the [UC Berkeley AUTOLAB](http://autolab.berkeley.edu/) for common robotics motion and vision applications. We also have a lot of research collaborators using them, so it would be great it we could include them here!

[autolab-core](https://pypi.org/project/autolab-core/)
[autolab_perception](https://pypi.org/project/autolab_perception/)

Thanks!

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -29,6 +29,14 @@ adafruit-pca9685-pip:
   ubuntu:
     pip:
       packages: [adafruit-pca9685]
+autolab-core-pip:
+  ubuntu:
+    pip: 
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
I was at first weirded out by the mixing of dashes and underscores but that's actually how they're listed on PyPI.org

underdash

nuclearsandwich (approved):

cottsay (approved):

PR #21024, opened 2019-04-26
Add python-pycurl python rosdep

yotabits:
You can find here the package for debian:
https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=python-pycurl

and for Ubuntu:
https://packages.ubuntu.com/search?keywords=python-pycurl&searchon=names&suite=all&section=all

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2861,6 +2861,7 @@ python-pyaudio:
   ubuntu: [python-pyaudio]
 python-pycurl:
   debian: [python-pycurl]
+  fedora: [python2-pycurl]
   ubuntu: [python-pycurl]
 python-pydbus:
   ubuntu:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2859,6 +2859,9 @@ python-pyaudio:
   fedora: [pyaudio]
   gentoo: [dev-python/pyaudio]
   ubuntu: [python-pyaudio]
+python-pycurl:
+  debian: [python-pycurl]
+  ubuntu: [python-pycurl]
 python-pydbus:
   ubuntu:
     artful: [python-pydbus]


rosdep/python.yaml
```diff
@@ -2859,6 +2859,17 @@ python-pyaudio:
   fedora: [pyaudio]
   gentoo: [dev-python/pyaudio]
   ubuntu: [python-pyaudio]
+python-pycurl:
+  debian:
```
nuclearsandwich: Since the package is available on all currently supported Debian packages the debian rule can be expressed more simply as

`debian: [python-pycurl]`

rosdep/python.yaml
```diff
@@ -2859,6 +2859,17 @@ python-pyaudio:
...
+    buster: [python-pycurl]
+    jessie: [python-pycurl]
+    sid: [python-pycurl]
+    stretch: [python-pycurl]
+  ubuntu:
```
nuclearsandwich: Although trusty doesn't have this packaged we could phrase this as:

```
  ubuntu:
    '*': [python-pycurl]
    trusty: null
```

But I don't strongly believe this change is required.

nuclearsandwich (changes requested):
Thanks for providing reviewable links. I've made some suggestions that will make this is easier to maintain.

rosdep/python.yaml
tfoote: Trusty is EOL within a month. I'd suggest saving the extra line.

yotabits:
@tfoote @nuclearsandwich, thank you for the review I made the requested changes.

rosdep/python.yaml
```diff
@@ -2859,6 +2859,9 @@ python-pyaudio:
   fedora: [pyaudio]
   gentoo: [dev-python/pyaudio]
   ubuntu: [python-pyaudio]
+python-pycurl:
+  debian: [python-pycurl]
+  ubuntu: [python-pycurl]
```
clalancette: This is also available in Fedora as https://apps.fedoraproject.org/packages/python2-pycurl

nuclearsandwich (approved):
:+1: with or without adding @clalancettes' Fedora suggestion

rosdep/python.yaml
tfoote: added in 7b03eba

tfoote (approved):

PR #21035, opened 2019-04-26
Test pull requests for perception_pcl

paulbovbel:
I'm not sure why this was previously turned off in a8c6a119ca, but I'll try to find out and dig in.


paulbovbel:
Looks like it was due to warnings from `pcl`, these are resolved in https://github.com/ros-perception/perception_pcl/pull/233

PR #21049, opened 2019-04-29
Add fxload to base.yaml

toliver:


clalancette:
In general, it is easier to review these if you follow the contributing guidelines [here](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md).  We're usually looking to see the keys for Ubuntu, Debian, and Fedora (with extra points for Gentoo and Arch).

Here it looks like this is available:
- Ubuntu - https://packages.ubuntu.com/search?keywords=fxload&searchon=names&suite=all&section=all
- Debian - https://packages.debian.org/search?keywords=fxload&searchon=names&suite=all&section=all
- Fedora - https://apps.fedoraproject.org/packages/fxload

So an update to this key for Debian and Fedora would be appreciated.  Thanks.

tfoote (changes requested):
Please fill in the debian and fedora rules pointed out by @clalancette

toliver:
Thanks for your comments @clalancette. I think I have addressed them.

nuclearsandwich (approved):

clalancette:
> Thanks for your comments @clalancette. I think I have addressed them.

Perfect, thank you!

clalancette (approved):

PR #21072, opened 2019-05-03
alpine: use packaged gmock over source

russkel:
I can see there exists a gmock package in the alpine repository. @at-wat is there any reason to use the source over the packaged version? 

https://pkgs.alpinelinux.org/package/v3.9/main/x86_64/gmock


tfoote (approved):
Using the packaged version sounds a lot more reliable and is generally what we'd prefer.

I believe that it didn't exist at the time of the original submission.
https://github.com/ros/rosdistro/pull/18474

PR #21104, opened 2019-05-07
Add libmodbus5 to base.yaml

yotabits:
Key added for:
- [Debian](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=libmodbus5) 
- [Ubuntu](https://packages.ubuntu.com/search?keywords=libmodbus5&searchon=names&suite=all&section=all)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2356,6 +2356,11 @@ libmodbus-dev:
   debian: [libmodbus-dev]
   fedora: [libmodbus-devel]
   ubuntu: [libmodbus-dev]
+libmodbus5:
+  debian: [libmodbus5]
+  fedora: [libmodbus]
+  gentoo: [dev-libs/libmodbus]
+  ubuntu: [libmodbus5]
 libmongoclient-dev:
   debian: [libmongoclient-dev]
   fedora: [mongo-cxx-driver]


sloretz:
The [guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules) recommend adding Fedora and Gentoo rules if available. I think those would be

```
fedora: [libmodbus]
gentoo: [dev-libs/libmodbus]
```

https://apps.fedoraproject.org/packages/libmodbus/overview/
https://packages.gentoo.org/packages/dev-libs/libmodbus

yotabits:
@sloretz @tfoote 
First of all, thank you for suggestion and the pointer. I will update my PR.
Can't we make Fedora/Gentoo a "must"? On the last PR, I made it has been always added, so, it seems to be more than a "should" but a "must" in the sense of [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)

tfoote (approved):
Thanks for the update. We don't want to make it a "must" as there are definitely packages that don't have equivalents available on the other platforms. And we don't want to block a key being added if it's not available on those other platforms. But we do want to make sure to do our best to keep the index complete.

sloretz (approved):

PR #21105, opened 2019-05-07
add carla Python API to the depends

hakuturu583:
Carla Released Python API as a pip package. (https://github.com/carla-simulator/carla/pull/1608)
So, I want to add this package to the rosdep as carla-pip.


clalancette (approved):

nuclearsandwich:
@hakuturu583 in the future it would be great if you could link to the pypi.org entry for contributed pip packages. https://pypi.org/project/carla/ It will make them easier to review.

PR #21201, opened 2019-05-10
Add ruby ronn

BMarchi:
debian: [https://packages.debian.org/stretch/ruby-ronn](https://packages.debian.org/stretch/ruby-ronn)
fedora: [https://apps.fedoraproject.org/packages/rubygem-ronn](https://apps.fedoraproject.org/packages/rubygem-ronn)
gentoo: [https://packages.gentoo.org/packages/app-text/ronn](https://packages.gentoo.org/packages/app-text/ronn)
ubuntu: [https://packages.ubuntu.com/disco/ruby-ronn](https://packages.ubuntu.com/disco/ruby-ronn)

### DIFF
--- a/rosdep/ruby.yaml
+++ b/rosdep/ruby.yaml
@@ -60,6 +60,9 @@ rdoc:
   fedora: [rubygem-rdoc]
   gentoo: [dev-ruby/rdoc]
   ubuntu: [ruby]
+ronn:
+  debian: [ruby-ronn]
+  ubuntu: [ruby-ronn]
 ruby:
   arch: [ruby]
   debian:
--- a/rosdep/ruby.yaml
+++ b/rosdep/ruby.yaml
@@ -62,6 +62,8 @@ rdoc:
   ubuntu: [ruby]
 ronn:
   debian: [ruby-ronn]
+  fedora: [rubygem-ronn]
+  gentoo: [app-text/ronn]
   ubuntu: [ruby-ronn]
 ruby:
   arch: [ruby]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
Please review the [contribution guidelines for rosdep rules](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions) and update your PR to conform with them.

clalancette:
So it looks like ruby-ronn is only available in Ubuntu since Xenial: https://packages.ubuntu.com/search?keywords=ruby-ronn&searchon=names&suite=all&section=all .  @nuclearsandwich @tfoote I forget what our policy is about new rosdep keys for out-of-support distros.  Do we want to just use what's in this PR, or should we explicitly call out the distros since xenial as being supported?

nuclearsandwich:
> I forget what our policy is about new rosdep keys for out-of-support distros. Do we want to just use what's in this PR, or should we explicitly call out the distros since xenial as being supported?

Since Xenial is the oldest currently-supported Ubuntu release I think it's fine to keep the simplified key.

nuclearsandwich:
Added an archlinux entry: https://www.archlinux.org/packages/community/any/ruby-ronn/

nuclearsandwich:
What are other @ros/rosdeputies opinions on using the rosdep key `ronn` rather than `ruby-ronn` matching the Debian package name as we've been doing lately?

My preference would probably be to use the package name to avoid possible confusion if a future non-ruby package called ronn is introduced.

clalancette:
> My preference would probably be to use the package name to avoid possible confusion if a future non-ruby package called ronn is introduced.

Oh yeah, I missed that.  I agree, a key of `ruby-ronn` is preferable.

nuclearsandwich (approved):

clalancette (approved):

PR #21208, opened 2019-05-10
cnn_bridge: 0.8.4-4 in 'melodic/distribution.yaml' [bloom]

wew84:
Increasing version of package(s) in repository `cnn_bridge` to `0.8.4-4`:

- upstream repository: https://github.com/wew84/cnn_bridge.git
- release repository: https://github.com/wew84/cnn_bridge-release.git
- bloom version: `0.8.0`

## cnn_bridge

```
* Moved source to repo
* Initial commit
* Contributors: Noam C. Golombek
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1317,20 +1317,6 @@ repositories:
       url: https://github.com/ros/cmake_modules.git
       version: 0.4-devel
     status: maintained
-  cnn_bridge:
-    doc:
-      type: git
-      url: https://github.com/wew84/cnn_bridge.git
-      version: 0.8.4
-    release:
-      tags:
-        release: release/kinetic/{package}/{version}
-      url: https://github.com/wew84/cnn_bridge-release.git
-      version: 0.8.4-1
-    source:
-      type: git
-      url: https://github.com/wew84/cnn_bridge.git
-      version: 0.8.4
   cob_android:
     doc:
       type: git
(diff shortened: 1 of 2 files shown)


melodic/distribution.yaml
```diff
@@ -641,6 +641,21 @@ repositories:
       url: https://github.com/ros/cmake_modules.git
       version: 0.4-devel
     status: maintained
+  cnn_bridge:
+    doc:
+      type: git
+      url: https://github.com/wew84/cnn_bridge-doc.git
```
clalancette: While our infrastructure *can* handle a separate repository for documentation, it is much more common to just point the `doc` section here to the source repository (https://github.com/wew84/cnn_bridge.git).  I'd highly suggest that here given that the doc repository above is an empty repository.

clalancette (changes requested):
For the doc section.

melodic/distribution.yaml
wew84: Accepted your comment and changed the doc repo

kinetic/distribution.yaml
```diff
@@ -1317,6 +1317,20 @@ repositories:
       url: https://github.com/ros/cmake_modules.git
       version: 0.4-devel
     status: maintained
+  cnn_bridge:
```
clalancette: Please don't add multiple ROS distributions in the same pull request.  Sometimes we have to revert changes, and since the ROS distributions are on different schedules, we sometimes have to do it asynchronously from each other.  In this case, just remove this change from this PR and open a separate one for Kinetic.

clalancette (changes requested):

kinetic/distribution.yaml
wew84: Done

clalancette (approved):
Looks good now, thanks for iterating.

PR #21224, opened 2019-05-14
Add pynput pip lib

vincentrou:


nuclearsandwich (approved):
@vincentrou in the future it would be helpful if you could specify in your PR that pip packages are being used because the package is not available in the upstream platform repositories and provide links to the package on PyPI.

cottsay (approved):
https://pypi.org/project/pynput/

PR #21239, opened 2019-05-15
[dep] Add pymodbustcp.

130s:
- Since it's listed on PyPi https://pypi.org/project/pyModbusTCP/ I assume it's available on some following distros that seem to be commonly registered on this python.yaml.
  - debian
  - fedora
  - ubuntu
- Manually confirmed:
  - osx

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -257,7 +257,7 @@ pyflakes3:
     bionic: [pyflakes3]
     xenial: [pyflakes3]
     zesty: [pyflakes3]
-pymodbustcp:
+pymodbustcp-pip:
   debian:
     pip:
       packages: [pyModbusTCP]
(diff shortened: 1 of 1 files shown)


130s:
I have a question after reviewing [CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

> - If a package is only available via pip use the -pip extension.

I'm not sure if this is possible, but if a package that used to be only available via `pip` becomes available from non-pip repository, wouldn't -pip suffix be confusing?

I didn't add -pip without knowing this requirement. I'm fine to update but would like to hear the thoughts first.

tfoote (changes requested):
If the package becomes available via non-pip packaging paths then we will create a new key. Usually the alternative packaging approach will be a downgrade from the latest in pip and those expecting the version in pip could get broken by switching the requirement.

130s:
I see. Updated with `-pip` suffix.

tfoote (approved):
Thanks for the update lgtm

clalancette (approved):

PR #21240, opened 2019-05-16
librealsense: 2.16.5-0 in 'dashing/distribution.yaml' [bloom]

RachelRen05:
Increasing version of package(s) in repository `librealsense` to `2.16.5-0`:

- release repository: https://github.com/ros2-gbp/librealsense-release.git
- bloom version: `0.6.9`

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -467,6 +467,23 @@ repositories:
       url: https://github.com/ros2/launch_ros.git
       version: master
     status: developed
+  librealsense:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: ros2debian
+    release:
+      packages:
+      - librealsense2
+      tags:
+        release: release/dashing/{package}/{version}
+      url: https://github.com/ros2-gbp/librealsense-release.git
+      version: 2.16.5-0
+    source:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: ros2debian
+    status: maintained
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Hello @RachelRen05, I realized after merging this release and https://github.com/ros/rosdistro/pull/21230 that your bloom version is older than required for Dashing. I would recommend upgrading to bloom 0.8.0 and re-releasing these two repositories in order to fix some dependency problems that will come from releasing into Dashing with an older version of bloom. I've also just seen that our package release tutorial needs an update for Dashing so I'll work on that next. Thanks!

RachelRen05:
Hi @nuclearsandwich, Thank you.
At first, the version of the bloom is the latest version v0.8.0, but there were a error with the command `bloom-release --rosdistro dashing --track dashing package_name` 

the error log:
**`Bloom cannot cope with distribution_type 'None'`**

I checked the source code, there were no **distribution_type** key in **distributions** structure.so, I changed the version of the bloom to v0.6.9, the same as the **crystal**, it works fine.
Do I must use some else command parameters to fix the problem? 
After resolved the problem, I will upgrade to bloom v0.8.0 and re-releasing these two repositories.

Thank you very much.

nuclearsandwich:
That error is being tracked in https://github.com/ros-infrastructure/bloom/issues/536

You should work around it with the currently released version of bloom by updating rosdistro and catkin_pkg to their latest releases as well as updating bloom.

Sorry for the trouble and please let me know if you run into any other issues.

PR #21252, opened 2019-05-17
Define which sources are valid for rosdep keys

tfoote:


tfoote:
@cottsay @allenh1 @russkel could you take a few moments to fill in the requirements for Fedora, Gentoo, and Alpine as amendments to this PR?

allenh1:
@tfoote I am ready to push but don't have write access here, actually. I can email you a patch to `git am` into your branch or PR it into yours.

Let me know your preference.

tfoote:
A PR into this branch would be great.

allenh1 (approved):
lgtm

clalancette (approved):

nuclearsandwich:
I have added an Archlinux entry. The link to the AUR installer issue in rosdep may be more information than needed.

russkel:
Pedantic: I believe it's referred to as Arch Linux

nuclearsandwich:
> Pedantic: I believe it's referred to as Arch Linux

I swear they change it every few years just to mess with me.

tfoote:
Thanks for all the updates. This looks like a better state than we had before. More followups are welcome too if we need to iterate.

PR #21320, opened 2019-05-21
Updating information for the microstrain_3dm_gx5_45 ROS package

bsb808:
Correcting the source repos. URL and adding package to kinetic and melodic.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -7393,11 +7393,11 @@ repositories:
   microstrain_3dm_gx5_45:
     doc:
       type: git
-      url: https://github.com/ros-drivers/microstrain_mips
+      url: https://github.com/ros-drivers/microstrain_mips.git
       version: master
     source:
       type: git
-      url: https://github.com/ros-drivers/microstrain_mips
+      url: https://github.com/ros-drivers/microstrain_mips.git
       version: master
     status: developed
   microstrain_3dmgx2_imu:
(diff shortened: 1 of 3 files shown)


indigo/distribution.yaml
```diff
@@ -7393,11 +7393,11 @@ repositories:
   microstrain_3dm_gx5_45:
     doc:
       type: git
-      url: https://github.com/bsb808/microstrain_3dm_gx5_45.git
+      url: https://github.com/ros-drivers/microstrain_mips
```
sloretz: It looks like CI is unhappy because this is a URL to a webpage instead of the git repo URL. I think this should be `https://github.com/ros-drivers/microstrain_mips.git` instead, and same everywhere else in this PR.

sloretz (dismissed):

indigo/distribution.yaml
bsb808: My apologies.  I just pushed a new commit which I believe should resolve the mistake.

kinetic/distribution.yaml
```diff
@@ -6722,6 +6722,16 @@ repositories:
       url: https://github.com/xuefengchang/micros_swarm_framework.git
       version: master
     status: developed
+  microstrain_3dm_gx5_45:
```
tfoote: It's not referenced elsewhere, but it would be good to change the name at this level too for easier searching and consistency.

Also in the other 2 distros.

bsb808: Thank you for the catch.  Changed the reference to "microstrain_mips" so that naming is consistent.

PR #21331, opened 2019-05-22
Add Gentoo expected repositories

allenh1:
Signed-off-by: Hunter L. Allen <hunterlallen@protonmail.com>

@tfoote this is for #21252


tfoote (approved):
Thanks!

PR #21332, opened 2019-05-22
Add autograd-pip to Python

wxmerkt:
We'd like to add https://github.com/HIPS/autograd to be pip-installable via a rosdep key. I am not entirely sure whether the below syntax is correct (do we need to specify debian/ubuntu/etc explicitly?) - or whether I even choose the right key - could you please advise?

Thank you very much,
Wolfgang

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -33,6 +33,9 @@ autograd-pip:
   debian:
     pip:
       packages: [autograd]
+  fedora:
+    pip:
+      packages: [autograd]
   ubuntu:
     pip:
       packages: [autograd]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -29,6 +29,10 @@ adafruit-pca9685-pip:
   ubuntu:
     pip:
       packages: [adafruit-pca9685]
+autograd-pip:
+  '*':
+    pip:
+      packages: [autograd]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -29,6 +29,10 @@ adafruit-pca9685-pip:
   ubuntu:
     pip:
       packages: [adafruit-pca9685]
+autograd-pip:
+  '*':
```
tfoote: Unfortunately, wildcard matching is not valid for the os only for the version.

tfoote (changes requested):

rosdep/python.yaml
wxmerkt: Ah, thanks - I was wondering exactly that. I have amended the PR.

tfoote (approved):

rosdep/python.yaml
tfoote: Yeah, that would match the `*` distribution ;-)

Thanks for the update.

clalancette (changes requested):
Since we are adding a pip key, we may as well add it for Fedora as well.

wxmerkt:
@clalancette Done, added key for Fedora

clalancette (approved):

PR #21343, opened 2019-05-23
Add pyosmium

grecim:
Looking to add [pyosmium](https://osmcode.org/pyosmium/), which enables the creation of OpenStreetMap files from GPS coordinates.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -203,6 +203,19 @@ mercurial:
   osx:
     pip:
       packages: [mercurial]
+osmium:
+  debian:
+    pip:
+      [osmium]
+  ubuntu:
+    pip:
+      [osmium]
+  fedora:
+    pip:
+      [osmium]
+  gentoo:
+    pip:
+      [osmium]
 paramiko:
   alpine: [py-paramiko]
   arch: [python2-paramiko]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Following our 

Please link to the packages and take a look for fedora and gentoo packages.

grecim:
Ready for next review. Thanks.

tfoote (changes requested):
Thanks for submitting this rule but a quick search shows that there's binary packages for it that we should prefer to use instead of the pip installations.

https://packages.debian.org/buster/python-pyosmium
https://packages.ubuntu.com/disco/python-pyosmium

On Fedora I only found a python3 version:
https://apps.fedoraproject.org/packages/python3-osmium

But the python3 version also exists for debian and ubuntu:
https://packages.ubuntu.com/disco/python3-pyosmium
https://packages.debian.org/buster/python3-pyosmium

On the other platforms we can fall back to pip. 

As a side note, we do already have the c++ bindings in the database: https://github.com/ros/rosdistro/pull/9056

clalancette:
There's been no response on this for almost a month, so I'm going to close this out for now.  Feel free to reopen when you have time to fix it up like the previous comment.  Thanks.

grecim:
I am ready to re-open

tfoote:
Follow up in #21608

PR #21345, opened 2019-05-24
Adding ros_cvb_camera_driver to documentation index for distro

gleichaufjo:
I'd like ros_cvb_camera_driver to be indexed and documented on ros.org

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -11597,7 +11597,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/gleichaufjo/ros_cvb_camera_driver.git
-      version: kinetic-devel
+      version: master
   ros_emacs_utils:
     doc:
       type: git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -11593,6 +11593,11 @@ repositories:
       url: https://github.com/ros-controls/ros_controllers.git
       version: kinetic-devel
     status: maintained
+  ros_cvb_camera_driver:
+    doc:
+      type: git
+      url: https://github.com/gleichaufjo/ros_cvb_camera_driver.git
+      version: kinetic-devel
   ros_emacs_utils:
     doc:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -11593,6 +11593,11 @@ repositories:
       url: https://github.com/ros-controls/ros_controllers.git
       version: kinetic-devel
     status: maintained
+  ros_cvb_camera_driver:
+    doc:
+      type: git
+      url: https://github.com/gleichaufjo/ros_cvb_camera_driver.git
+      version: kinetic-devel
```
tfoote: There's no `kinetic-devel` branch only master. If we're going to index that it would need to be as follows.

```suggestion
      version: master
```

sloretz (approved):

clalancette (changes requested):
@gleichaufjo Thanks for the PR.  Before merging this, I think we'd really want to know the license of your code, and https://github.com/gleichaufjo/ros_cvb_camera_driver doesn't have either a LICENSE file, the `<license>` tag in the package.xml, or a license header in the source files.  Can you add a license for that (we recommend BSD or Apache 2, but you can choose what you like)?  Once that is done, we can merge this PR.  Thanks.

gleichaufjo:
> 
> 
> @gleichaufjo Thanks for the PR. Before merging this, I think we'd really want to know the license of your code, and https://github.com/gleichaufjo/ros_cvb_camera_driver doesn't have either a LICENSE file, the `<license>` tag in the package.xml, or a license header in the source files. Can you add a license for that (we recommend BSD or Apache 2, but you can choose what you like)? Once that is done, we can merge this PR. Thanks.

I have added the license now.

gleichaufjo:
I have created a ROS wiki page for this package. http://wiki.ros.org/ros_cvb_camera_driver
Will the information on the package as well as the link to the repository automatically be inserted?

PR #21347, opened 2019-05-24
Adding xaxxon_openlidar to documentation index for distro

xaxxontech:


tfoote (changes requested):
This is failing CI as the url's don't end in .git Please add that to the urls.

xaxxontech:
thanks - '.git's added!

indigo/distribution.yaml
```diff
@@ -17769,6 +17769,16 @@ repositories:
       url: https://github.com/ros/xacro.git
       version: indigo-devel
     status: maintained
+  xaxxon_openlidar:
```
sloretz: @tfoote since Indigo and Lunar are EOL, should the new entries in these two distros be removed from this PR?

tfoote: Yes please only update kinetic and melodic due to the others being EOL.

PR #21414, opened 2019-05-30
Adding libompl-dev for MoveIt2

mlautman:


tfoote (changes requested):
Please link to the upstream packages for contributed rosdep rules as well s add at least an ubuntu rule: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

PR #21457, opened 2019-06-04
Added geoalchemy2

agyoungs:
This python module appears to be a debian binary in python 3, but not python 2

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1573,7 +1573,7 @@ python-genshi:
   fedora: [python-genshi]
   gentoo: [dev-python/genshi]
   ubuntu: [python-genshi]
-python-geoalchemy2:
+python-geoalchemy2-pip:
   debian:
     pip:
       packages: [geoalchemy2]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1573,6 +1573,13 @@ python-genshi:
   fedora: [python-genshi]
   gentoo: [dev-python/genshi]
   ubuntu: [python-genshi]
+python-geoalchemy2:
+  debian:
+    pip:
+      packages: [geoalchemy2]
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


cottsay (changes requested):
Hi! Please see the [CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules) and update this change based on those guidelines. Off the top of my head, I can see that the key should end in `-pip`, and a link to the PyPI page in the change description would be appreciated.

Thanks!

agyoungs:
Just added `-pip` to the end of the tag.

[geoalchemy2](https://pypi.org/project/GeoAlchemy2/)

I needed this in a package that I'm working on that uses a SpatiaLite extended SQLite database. This library works with SQLAlchemy which is already included as a rosdep. It looks like it is available in 19.04 as a debian package, but only for python3 at the moment.

Thanks!

agyoungs (commented):
Addressed changes from review (i.e. adding `-pip` to the end)

rosdep/python.yaml
```diff
@@ -1573,6 +1573,13 @@ python-genshi:
   fedora: [python-genshi]
   gentoo: [dev-python/genshi]
   ubuntu: [python-genshi]
+python-geoalchemy2-pip:
+  debian:
+    pip:
+      packages: [geoalchemy2]
+  ubuntu:
```
clalancette: If we're doing a `-pip` package, we may as well add support for fedora and osx as well (like https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L157).  Thanks.

clalancette (changes requested):
Small changes to the key.

clalancette:
No response for almost a month, so I'm closing this.  If you'd still like to get this key in, feel free to reopen this PR and make the changes requested.  Thanks for the contribution.

PR #21467, opened 2019-06-06
std_stamped_msgs: 0.0.2-0 in 'kinetic/distribution.yaml' [bloom]

hakuturu583:
Increasing version of package(s) in repository `std_stamped_msgs` to `0.0.2-0`:

- upstream repository: https://github.com/hakuturu583/std_stamped_msgs.git
- release repository: https://github.com/hakuturu583/std_stamped_msgs-release.git
- bloom version: `0.6.7`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -13840,6 +13840,21 @@ repositories:
       url: https://github.com/ros/std_msgs.git
       version: groovy-devel
     status: maintained
+  std_stamped_msgs:
+    doc:
+      type: git
+      url: https://github.com/hakuturu583/std_stamped_msgs.git
+      version: master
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
These messages are using a confusing name of std_stamped_msgs which is confusingly similar to std_msgs and do not add semantic meaning. We strongly recommend against using these sort of untyped messages. To that end please use a namespace that is specific to your usage and not try to have a semi-colliding name that would imply that these are standard across the ROS ecosystem.

clalancette:
I'm going to close this PR out, since we won't merge this unless it has a different name.  Please resubmit a package with a different name and we'd be happy to merge, thanks.

PR #21484, opened 2019-06-07
add cupy to the rosdistro

hakuturu583:
I want to add cupy to rosdistro.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1131,6 +1131,10 @@ python-cryptography:
   debian: [python-cryptography]
   gentoo: [dev-python/cryptography]
   ubuntu: [python-cryptography]
+python-cupy:
+  debian: [python-cupy]
+  gentoo: [dev-python/cupy]
+  ubuntu: [python-cupy]
 python-cwiid:
   arch: [cwiid]
   debian: [python-cwiid]


tfoote:
Please link to the upstream packages for contributed rosdep rules: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

hakuturu583:
@tfoote Thank you. I forget it.
This is the package description of cupy.
https://pypi.org/project/cupy/

knorth55:
I cannot find `python-cupy` debian package in Ubuntu.
Which package are you trying to install?
If you want to install `CuPy` with `pip`, package name should be `python-cupy-pip`.

tl;dr
`CuPy` is strongly depend on `CUDA`, so it would be harmful to 
install `CuPy` with `rosdep` command
That is because users need to think whether their machine have proper nvidia driver or not.
Also, default `CuPy` runs build during the installation and fails building when users try to install it in PC without CUDA.
In Ubuntu 18.04, you can install CUDA supported by `CuPy` with `nvidia-cuda-toolkit`,
but it would be a problem in `Kinetic`.
https://packages.ubuntu.com/bionic/nvidia-cuda-toolkit

tfoote (changes requested):
If it's a pip only package we could add that. But it would also need the additional dependency onto any required driver packages that should be generally available and installable.

clalancette:
No response in almost a month, so closing this out.  Feel free to reopen if you want to get this key in. Thanks for the contribution.

PR #21509, opened 2019-06-11
add pugixml to base.yaml

mitsudome-r:
## Description
I would like to add pugixml-dev package to base.yaml
Debian: https://packages.debian.org/search?keywords=pugixml-dev&searchon=names&suite=stable&section=all
Fedora: https://apps.fedoraproject.org/packages/s/pugixml
Ubuntu: https://packages.ubuntu.com/search?keywords=pugixml-dev&searchon=names&suite=all&section=all

Added key is "pugixml-dev", not "pugixml" because it was used in distribution.yaml as below. 

## distribution.yaml in Indigo and Jade
I have a concern about distribution.yaml in Indigo and Jade. They seem to be using distributed wrapper for pugixml.
Indigo: https://github.com/ros/rosdistro/blob/master/indigo/distribution.yaml#L10654
Jade: https://github.com/ros/rosdistro/blob/bede3a3127935c5e364322dc04f63a935e35a36d/jade/distribution.yaml#L4762
I wanted to use pugixml in kinetic and melodic, but rather than adding this distribution to kinetic and melodic, I thought it was better to point to official released package in base.yaml.
Please let me know if I have to point to this distribution. 

Signed-off-by: mitsudome-r <ryohsuke.mitsudome@tier4.jp>

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4597,6 +4597,10 @@ psutils:
   fedora: [psutils]
   gentoo: [psutils]
   ubuntu: [psutils]
+pugixml-dev:
+  debian: [libpugixml-dev]
+  fedora: [pugixml]
+  ubuntu: [libpugixml-dev]
 python-dev:
   debian: [python-dev]
   fedora: [python2-devel]


tfoote:
@joselusl Could you comment on what the compatibility of your wrapper is with the upstream package. 

If that's only released for Indigo and Jade which are both EOL, it seems that switching to use the rosdep for kinetic and newer is perfect.

joselusl:
Hi,
I am currently using the package on Kinetic with no problems. I have never tested Lunar or Melodic.
I am quite busy lately, but if I manage to find the time, I will try to test it and push it. Please, feel free to help me with this! ;)
Best regards,

mitsudome-r:
@joselusl 
I 'm afraid that's not what we are concerned here. We would like to know if we can replace your wrapper package with libpugixml-dev package that can be installed using package managers(like `apt-get install`). If you don't have any changes from the [original source code](https://github.com/zeux/pugixml), then it's very likely that we can. 

If that is the case, we can add my rules to base.yaml to retrieve original pugixml package instead of pointing to your repository in distribution.yaml. That way, we don't have to test and update rules in distribution.yaml every time there is a new ROS release.

joselusl:
@mitsudome-r I understand!
I didn't do any change to libpugixml-dev. I guess, although I have not tried it, that my repository is not needed anymore.
I did it a long time ago (Fuerte), catkinized and used since then. Before it was useful. Now I bet it is useless.
Thank you very much
Best regards

mitsudome-r:
@joselusl I'm pretty sure your package helped many people from Fuerte to Jade, and we really admire your work. 
@tfoote It seems we can switch to rosdep for kinetic and newer. Do I just need to wait for reviewer to review the change now?

clalancette (approved):
Looks good to me.

PR #21514, opened 2019-06-11
Remove spinnaker_camera_driver

tfoote:
It's failing to build http://build.ros.org/view/Kbin_uX32/job/Kbin_uX32__spinnaker_camera_driver__ubuntu_xenial_i386__binary/
```
16:27:12 In file included from /tmp/binarydeb/ros-kinetic-spinnaker-camera-driver-0.1.3/include/spinnaker_camera_driver/camera.h:32:0,
16:27:12                  from /tmp/binarydeb/ros-kinetic-spinnaker-camera-driver-0.1.3/src/camera.cpp:25:
16:27:12 /tmp/binarydeb/ros-kinetic-spinnaker-camera-driver-0.1.3/include/spinnaker_camera_driver/set_property.h:29:23: fatal error: Spinnaker.h: No such file or directory
16:27:12 compilation terminated.
```

It looks like a missing dependency or include path.

Partial reversion of #19175 FYI @mhosmar-cpr

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3288,7 +3288,6 @@ repositories:
     release:
       packages:
       - flir_camera_driver
-      - spinnaker_camera_driver
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ros-drivers-gbp/flir_camera_driver-release.git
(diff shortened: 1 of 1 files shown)


tfoote:
I found https://github.com/ros-drivers/flir_camera_driver/issues/24 which suggests that this requires the spinnaker sdk installed which is not a dependency available for restribution so likely this shouldn't be attempted to be released and should be blacklisted.

mhosmar-cpr:
Hi @tfoote ,  It looks to me like they no longer have Spinnaker 1.13 hosted.  I'll have to update to their version.

mhosmar-cpr:
Their hosting has changed significantly. Im ok with blacklisting till im able to make a fix/new release.

PR #21557, opened 2019-06-14
pqdict pip package

fcolas:


rosdep/python.yaml
```diff
@@ -2811,6 +2811,10 @@ python-plyfile-pip:
 python-poster:
   debian: [python-poster]
   ubuntu: [python-poster]
+python-pqdict:
```
clalancette: For a pip package, we require a `-pip` suffix for the key.

rosdep/python.yaml
```diff
@@ -2811,6 +2811,10 @@ python-plyfile-pip:
...
+  ubuntu:
```
clalancette: Also, if we are going to make it a `-pip` key, we may as well enable it for fedora, debian, and osx as well.

clalancette (changes requested):

rosdep/python.yaml
fcolas: Done.

rosdep/python.yaml
fcolas: Done. Thanks for the review.

clalancette (approved):

PR #21570, opened 2019-06-17
Adding boost-python rosdep key.

JWhitleyWork:
Alpine: https://pkgs.alpinelinux.org/packages?name=*boost-python*&branch=edge
Debian: https://packages.debian.org/search?keywords=libboost-python-dev
Fedora: https://apps.fedoraproject.org/packages/boost-python2-devel
Ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libboost-python-dev&searchon=names

This is so the Python bindings for Boost can be installed without installing the whole of Boost (libboost-all-*). Required by a dependency of the Autoware project.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -317,11 +317,6 @@ boost:
     xenial: [libboost-all-dev]
     yakkety: [libboost-all-dev]
     zesty: [libboost-all-dev]
-boost-python:
-  alpine: [boost-python2]
-  debian: [libboost-python-dev]
-  fedora: [boost-python2-devel]
-  ubuntu: [libboost-python-dev]
 box2d-dev:
   arch: [box2d]
   debian: [libbox2d-dev]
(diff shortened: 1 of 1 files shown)


cottsay (approved):

mitsudome-r:
other boost packages uses different naming convention for keys. 
Perhaps it is more appropriate to use "libboost-python",
https://github.com/ros/rosdistro/blob/a53f402d9167cfc05ecfabe6802449836da88942/rosdep/base.yaml#L1568

gbiggs:
I think it would be best to match the naming scheme used by existing similar keys.

JWhitleyAStuff:
Done.

rosdep/base.yaml
```diff
@@ -1579,6 +1579,11 @@ libboost-program-options-dev:
   fedora: [boost-devel]
   openembedded: [boost@openembedded-core]
   ubuntu: [libboost-program-options-dev]
+libboost-python:
```
clalancette: I'm wondering if we should make this key `libboost-python2`; clearly on Alpine and Fedora, there's a separate package for Python 3.  But I don't know what our policy on python2 vs. python3 rosdep keys is, so I'm not sure.

poggenhans: At least on ubuntu, both the python2 and the python3 version are supported when installing `libboost-python-dev`. There is no specialized package for a specific version. 

Afaik, the ROS policy is not to enforce a specific python version but instead work with whatever is set as `PYTHON_VERSION` in CMake. Therefore it makes sense to ensure that boost packages are installed for all possible values of `PYTHON_VERSION`.

JWhitleyAStuff:
@cottsay - Can this be merged? Is there someone else I should ping?

nuclearsandwich (approved):

PR #21573, opened 2019-06-18
added python-namedlist to python.yaml

rfeistenauer:
Added the namedlist module: [pypi namedlist](https://pypi.org/project/namedlist/)
This module provides an fast and easy way to create objects with dynamicly set attributes. I for example use it for items in a database like structure.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2303,7 +2303,7 @@ python-mysqldb:
   ubuntu: [python-mysqldb]
 python-namedlist:
   ubuntu:
-    pip: 
+    pip:
       packages: [namedlist]
 python-netaddr:
   debian: [python-netaddr]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2301,6 +2301,10 @@ python-mysqldb:
   fedora: [python-mysql]
   gentoo: [dev-python/mysqlclient]
   ubuntu: [python-mysqldb]
+python-namedlist:
```
clalancette: Since this is a pip package, we require a `-pip` suffix on the key.

rosdep/python.yaml
```diff
@@ -2301,6 +2301,10 @@ python-mysqldb:
...
+  ubuntu:
```
clalancette: And since it is pip, we can add entries for fedora, debian, and osx as well.

clalancette (changes requested):

clalancette (approved):

PR #21580, opened 2019-06-19
Add omronsentech_camera to documentation index for ROS Melodic.

ose-support-ros:
Hi,

I would like my omronsentech_camera to be indexed and documented for ROS Melodic.

Thanks,

OSE ROS Support.


melodic/distribution.yaml
```diff
@@ -4356,6 +4356,17 @@ repositories:
       url: https://github.com/ros-drivers/omron.git
       version: kinetic-devel
     status: maintained
+  omronsentech_camera:
+    doc:
+      type: git
+      url: https://github.com/ose-support-ros/omronsentech_camera.git
+      version: master
+    source:
+      test_commits: false
```
clalancette: Please remove the `test_commits` line here; false is the default, and we only need to add the line if we want to enable it.

clalancette (changes requested):
One minor change needed.

melodic/distribution.yaml
ose-support-ros: Hi,

The test_commits has been removed.

Thanks,

OSE ROS Support.

clalancette: Thanks!

clalancette (approved):

PR #21612, opened 2019-06-25
Adding pipeline_planner to documentation index for distro

t-ogataseaos:
I'd like my pipeline planner package to be indexed and documented on ros.org.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9216,6 +9216,10 @@ repositories:
       version: master
     status: maintained
   pipeline_planner:
+    doc:
+      type: git
+      url: https://github.com/SeaosRobotics/pipeline_planner_open.git
+      version: master
     source:
       type: git
       url: https://github.com/SeaosRobotics/pipeline_planner_open.git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -9215,6 +9215,12 @@ repositories:
       url: https://github.com/amineHorseman/pioneer_teleop.git
       version: master
     status: maintained
+  pipeline_planner:
+    source:
```
clalancette: We can add this, but this does not add it to the documentation index.  If you want that, you need to add a `doc` section as well; see the `platform_automation_msgs` package just below for an example.

clalancette (changes requested):
Probably needs changes for documentation, and holding for [sync](https://discourse.ros.org/t/preparing-for-kinetic-sync-2019-06-24/9663)

t-ogataseaos:
I've added doc section.
I'd like my pipeline planner package to be indexed and documented on ros.org.

clalancette (approved):

PR #21613, opened 2019-06-25
Add rule for python3-lttng

christophebedard:
Kind of a follow-up to #11991

`python3-lttng` contains the Python bindings to control LTTng tracing sessions

* https://packages.ubuntu.com/bionic/python3-lttng
* https://packages.debian.org/stretch/python3-lttng

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4984,6 +4984,9 @@ python3-lark-parser:
   ubuntu:
     bionic: [python3-lark-parser]
     xenial: [python3-lark-parser]
+python3-lttng:
+  debian: [python3-lttng]
+  ubuntu: [python3-lttng]
 python3-lxml:
   debian: [python3-lxml]
   fedora: [python3-lxml]


rosdep/python.yaml
```diff
@@ -4984,6 +4984,9 @@ python3-lark-parser:
   ubuntu:
     bionic: [python3-lark-parser]
     xenial: [python3-lark-parser]
+python3-lttng:
+  debian: [python3-lttng]
```
clalancette: What's the fedora situation here?

clalancette (changes requested):

rosdep/python.yaml
christophebedard: Doesn't seem to be available (not to be confused with `python3-lttngust`): https://apps.fedoraproject.org/packages/s/python3-lttng

And it does not seem to be available on `pip` either: https://pypi.org/search/?q=lttng

clalancette: OK, thanks for the information.

clalancette (approved):

cottsay (approved):

PR #21623, opened 2019-06-26
Adding xbot_talker to documentation index for distro

jiuyewxy:
I'd like my_repo xbot_talker to be indexed and documented on ros.org

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -16140,6 +16140,16 @@ repositories:
       url: https://github.com/xaxxontech/xaxxon_openlidar.git
       version: master
     status: maintained
+  xbot_talker:
+    doc:
+      type: git
+      url: https://github.com/DroidAITech/xbot_talker.git
+      version: master
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -16140,6 +16140,16 @@ repositories:
       url: https://github.com/xaxxontech/xaxxon_openlidar.git
       version: master
     status: maintained
+  xbot_talker:
+    doc:
+      type: git
+      url: https://github.com/DroidAITech/xbot_talker.git
```
clalancette: The source repository doesn't have any license specified in either the package.xml, LICENSE file, or headers on the sources.  Please choose a license (https://opensource.org/licenses is a good place to start), and add it to the package.xml `<license>` tag and a LICENSE file in the root of the repository.

clalancette (changes requested):

clalancette:
Also, this now needs to be rebased.

clalancette:
No response for 20 days, so I'm going to close this out for now.  Feel free to fix the problems pointed out in review and reopen.  Thanks for the contribution.

PR #21640, opened 2019-06-28
Add nuitka for ubuntu and debian distribuitions in python.yaml

asierfernandez:
I'm not sure about naming convention within rospdep. As it is a python package I included it within python.yaml, and as there is a nuitka package available through `apt` I defined the dependency name with `-pip` suffix. 

If there is a better way to do it, please let me know.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -213,6 +213,9 @@ mercurial:
   osx:
     pip:
       packages: [mercurial]
+nuitka:
+  debian: [nuitka]
+  ubuntu: [nuitka]
 paramiko:
   alpine: [py-paramiko]
   arch: [python2-paramiko]


clalancette (changes requested):
It looks like there are apt packages available, at least for Ubuntu: https://packages.ubuntu.com/search?keywords=nuitka&searchon=names&suite=all&section=all (it's possible that this is true for other distributions as well; I didn't check).  In general, we highly prefer apt packages where available, so please follow the [CONTRIBUTING](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md) guide and find the packages that go along with the distributions.  Thanks.

asierfernandez:
I proposed pip version due to it can be executed as a python module, and apt package can not.
In our developments this makes simpler to call it from the CMakeLists.txt, and there is no need to manually find Nuitka package. This is not a big issue, so I included changes you proposed.

BTW, I checked debian definitions and I included them.

nuclearsandwich:
@asierfernandez as you may already be aware our preference for apt packages comes from the fact that packages released via bloom to a ROS build farm cannot depend on packages from pip. However, if your team needs the pip version of a package for it to be any use to you and you have no use for the apt version of the package, I think it makes sense to add the pip packages so your team can use them without maintaining a separate rosdep definition list. Further, I think it is worth omitting the apt packages if no one will find them useful.

asierfernandez:
@nuclearsandwich definitely having the pip version simplifies our developments.
I've reset to the first commit and push it back again.

tfoote (changes requested):
Since this package has been packaged for apt for a long time I would suggest that we should not be using the pip rule. If it's created others who may want to use this package in the future for making releases will be caused issues if they ever use this key on any of their systems getting the newerversion from pip overlayed on their system. 

If it's just for your local development I would suggest creating your own rosdep sources file and adding it to the general database for all ROS users with the newest version that you want the features from to override the debian version.

This will work for your workflow and not break things for anyone else who's interested in using the newer version. Especially as the newer ubuntu distros appear to have the newer versions already. So the delta to the pip version is relatively small for newer versions.

asierfernandez:
Following your comments, I left the keys for native packages

clalancette (approved):

clalancette:
@tfoote @nuclearsandwich I've approved, but this needs one more approval.  Mind looking at it?  Thanks.

nuclearsandwich (approved):

PR #21649, opened 2019-06-28
Fix libjpeg mapping for OpenEmbedded (#21648)

### LINKED ISSUE #21648: Incorrect OE value for libjpeg in rosdep/base.yaml
Should be `jpeg` not `libjpeq-turbo`.

herb-kuta-lge:
Fixes #21648

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2263,7 +2263,7 @@ libjpeg:
   freebsd: [libjpeg-turbo]
   gentoo: [virtual/jpeg]
   macports: [jpeg]
-  openembedded: [libjpeq-turbo@openembedded-core]
+  openembedded: [jpeg@openembedded-core]
   opensuse: [libjpeg62-devel]
   slackware: [libjpeg-turbo]
   ubuntu:


nuclearsandwich:
Hi @herb-kuta-lge is there any web listing of openembedded packages or other method we can use to check this change? Otherwise it is completely opaque to us as reviewers. Thanks!

herb-kuta-lge:
In general, you can check for OE recipes using the [OE Layer Index for **thud**](http://layers.openembedded.org/layerindex/branch/thud/recipes/). However, in this case, you'll need to look at the recipe itself and see that the one for `libjpeg-turbo` provides `jpeg`: http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-graphics/jpeg/libjpeg-turbo_2.0.0.bb?h=thud#n25, which is the mapping to which this commit changes.

nuclearsandwich:
So the updated key will still resolve to libjpeg-turbo but if ever a different library provided jpeg instead, this would still be valid?

herb-kuta-lge:
If a different recipe ever declared that it provided `jpeg`, then, yes, it would be valid. OpenEmbedded would then warn that multiple recipes provide the same thing, and a PREFERRED_PROVIDER setting would need to be added to `meta-ros` to select the desired one.

tfoote (approved):

clalancette (approved):

PR #21661, opened 2019-07-01
i Would like to add eyantra_drone to ROS docs

simmubhangu:


clalancette (changes requested):
Please open up separate PRs for each distribution.  In this case, both indigo and lunar are out-of-support, so we won't accept new contributions there.  Just open a separate PR for each of kinetic and melodic.  Thanks.

clalancette:
Closing this out since we'll need separate PRs per distribution.

PR #21662, opened 2019-07-01
Adding pipeline_planner to documentation index for melodic

t-ogataseaos:
I'd like pipeline_planner to be indexed and documented on ros.org.


clalancette:
Holding for [sync](https://discourse.ros.org/t/preparing-for-melodic-sync-2019-06-28/9714)

PR #21670, opened 2019-07-02
vmrc project is being renamed to vrx. Remove old references

j-rivero:
We are releasing new packages for the VRX project, previously named VMRC. In kinetic bloom is complaining about the duplicate entries belonging to the old VMRC.

This PR removes the old rosdistro entries for VMRC so we can bloom the new VRX which has the same packages. I'm not sure if there is a better way to handle the renaming transition.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -15741,26 +15741,6 @@ repositories:
       url: https://github.com/JdeRobot/VisualStates.git
       version: master
     status: developed
-  vmrc:
-    doc:
-      type: hg
-      url: https://bitbucket.org/osrf/vmrc
-      version: default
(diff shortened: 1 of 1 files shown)


clalancette:
@j-rivero I'm going to wait on this one until you and @nuclearsandwich figure it out.  Just comment here when you need another review.

j-rivero:
> @j-rivero I'm going to wait on this one until you and @nuclearsandwich figure it out. Just comment here when you need another review.

Thanks Chris. I think that PR  #21675 is a better approach if it's ok to break the CI during some hours until I run the next release tomorrow.

nuclearsandwich:
Closing this in favor of #21675.

PR #21675, opened 2019-07-02
Rename vmrc to vrx.

nuclearsandwich:
Proposed as an alternative to #21670. This PR may fail CI as this release version is invalid. I think we still have to merge this before the new vrx release for Kinetic can be bloomed but if we synchronize the buildfarm should not be upset for too long.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -15754,10 +15754,10 @@ repositories:
       url: https://github.com/JdeRobot/VisualStates.git
       version: master
     status: developed
-  vmrc:
+  vrx:
     doc:
       type: hg
-      url: https://bitbucket.org/osrf/vmrc
+      url: https://bitbucket.org/osrf/vrx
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
cc @j-rivero I think this accomplishes the same thing but with a bit more continuity.

j-rivero:
> cc @j-rivero I think this accomplishes the same thing but with a bit more continuity.

sounds good to me. Probably the best approach.

nuclearsandwich:
Merging this even though the release isn't present in order to make the vrx releas right after. It's not currently possible to release vrx because of the conflicting package names provided by the repository.

nuclearsandwich:
I didn't think this all the way through and forgot to check the lexical sort which changed as a result of this rename.

nuclearsandwich:
The build has now been repaired with https://github.com/ros/rosdistro/pull/21683

PR #21677, opened 2019-07-03
add eyantra_drone

simmubhangu:
I have remove the package from lunar and indigo as suggested.


clalancette:
> I have remove the package from lunar and indigo as suggested.

Can you please split this up into separate PRs for Kinetic and Melodic as requested?  They are on different sync schedules, so in general PRs should not add things to more than one distribution at a time.  Thanks.

PR #21682, opened 2019-07-03
Update sort order for vrx.

nuclearsandwich:
Follow up to #21675. When updating the repository name I forgot to re-sort it.


nuclearsandwich:
With this PR the only failure is the missing package.xml.

cottsay (approved):
As expected, the only failures are for missing `package.xml`: https://api.travis-ci.org/v3/job/553976271/log.txt

PR #21695, opened 2019-07-05
Adding new repository for indexing: rosplan

m312z:


melodic/distribution.yaml
```diff
@@ -6591,6 +6591,15 @@ repositories:
       url: https://github.com/rospilot/rospilot.git
       version: melodic
     status: developed
+  rosplan:
+    doc:
+      type: git
+      url: https://github.com/kcl-planning/rosplan
```
tfoote: The URLs need to end in `.git`

tfoote (changes requested):
These packages do not follow our naming criteria.  Please see: http://wiki.ros.org/ROS/Patterns/Conventions

Package names shouldn't start with `ros`
Plan is not specific enough to understand what's inside the package.

clalancette:
It's been a month with no response, so I'll close this out.  Feel free to reopen once you've addressed the feedback from the review.  Thanks.

PR #21712, opened 2019-07-07
add mosquitto-clients to the rosdep

hakuturu583:
add mosquitto-clients to the rosdep.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4139,9 +4139,7 @@ mosquitto:
   fedora: [mosquitto]
   ubuntu: [mosquitto]
 mosquitto-clients:
-  arch: [mosquitto-clients]
   debian: [mosquitto-clients]
-  fedora: [mosquitto-clients]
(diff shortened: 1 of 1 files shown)


hakuturu583:
these are package description

debian https://packages.debian.org/stretch/mosquitto-clients
ubuntu https://packages.ubuntu.com/disco/mosquitto-clients

tfoote (approved):
This looks fine. It looks like fedora doesn't have a mosquitto-client package. Is that included in the mosquitto main package? If so it should have the main mosquitto listed as a dependency. Otherwise we can skip it.

clalancette (changes requested):
Looking at my Fedora-30 box, it looks like the `mosquitto_{pub,sub}` binaries are provided by the main `mosquitto` package on Fedora, so please add that key as well.

clalancette:
No response for a month, so I'm going to close this.  Feel free to reopen if you still want to get this rosdep key in.

PR #21734, opened 2019-07-11
Adding libzmqpp-dev

ruffsl:
[ubuntu](https://packages.ubuntu.com/bionic/libzmqpp-dev)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3845,6 +3845,8 @@ libzmq3-dev:
   gentoo: [net-libs/cppzmq]
   openembedded: [zeromq@meta-oe]
   ubuntu: [libzmq3-dev]
+libzmqpp-dev:
+  ubuntu: [libzmqpp-dev]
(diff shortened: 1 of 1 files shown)


tfoote:
Presumably this has the debian equivalanet and probably at least fedora too. Could you look up those so we can add them too?

ruffsl:
I couldn't find this c++ package for debian or fedora.

https://pkgs.org/download/libzmqpp-dev
https://www.apt-browse.org/search/?query=libzmqpp-dev

I'm not sure if the project has been backported yet, as I've only seen the normal C binding packages:

https://github.com/zeromq/zmqpp

tfoote (approved):
Thanks for checking. 

Interesting that it went straight to Ubuntu w/o filtering through Debian. From the changelog the maintainer is at Canonical. They must have needed it for an intenal project so pushed it in.

clalancette (approved):

PR #21736, opened 2019-07-11
abseil_cpp: 0.4.2-1 in 'melodic/distribution.yaml' [bloom]

facontidavide:
Increasing version of package(s) in repository `abseil_cpp` to `0.4.2-1`:

- upstream repository: https://github.com/Eurecat/abseil-cpp.git
- release repository: https://github.com/Eurecat/abseil_cpp-release.git
- bloom version: `0.8.0`
- previous version for package: `0.2.3-1`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -19,11 +19,12 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/Eurecat/abseil_cpp-release.git
-      version: 0.2.3-1
+      version: 0.4.2-1
     source:
       type: git
       url: https://github.com/Eurecat/abseil-cpp.git
       version: master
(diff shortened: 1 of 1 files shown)


clalancette (commented):
@facontidavide are there changes in here that will prevent the build problems we had with the upgrade before?  I'm thinking about http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__ros_type_introspection__ubuntu_bionic_amd64__binary/15/

facontidavide:
it is the same version that works on kinetic, the latest. Therefore I think it is "supposed" to work

clalancette:
> it is the same version that works on kinetic, the latest. Therefore I think it is "supposed" to work

Works for me.  Let's see what happens!

PR #21776, opened 2019-07-17
Add python3-mypy

kyrofa:


jacobperron:
Following [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions), here are the web listings:
- [Ubuntu](https://packages.ubuntu.com/bionic/python3-mypy): `python3-mypy`
- [Debian - Buster](https://packages.debian.org/buster/python3-mypy): `python3-mypy`
- [Debian - Stretch](https://packages.debian.org/stretch/mypy): `mypy`
- [Fedora](https://apps.fedoraproject.org/packages/python3-mypy): `python3-mypy`
- [Gentoo](https://packages.gentoo.org/packages/dev-python/mypy): `dev-python/mypy`

I believe the reason for adding this dependency is so that we can add mypy tests to Python packages.

I think we should add an additional rule for Debian Stretch since it a recommended support platform.
I think adding a rule for Gentoo would be good too.

kyrofa:
> I think we should add an additional rule for Debian Stretch since it a recommended support platform.

I can do that, but Stretch is pretty old, where are you getting that info?

nuclearsandwich:
Added an Archlinux key https://www.archlinux.org/packages/community/any/mypy/

nuclearsandwich (approved):

kyrofa:
Thanks @nuclearsandwich!

PR #21777, opened 2019-07-18
create javascript.yaml file

k-okada:
to support npm installer and move nodejs/npm from base.yaml to javascript.yaml

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4263,20 +4263,6 @@ nmap:
   fedora: [nmap]
   gentoo: [net-analyzer/nmap]
   ubuntu: [nmap]
-nodejs:
-  debian: [nodejs]
-  fedora: [nodejs]
-  gentoo: [net-libs/nodejs]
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
Has there been a discussion among the community and @ros/rosdeputies about continuing the use of separate files for language-specific package managers? I seem to remember us wanting to get away from it but perhaps that was the os-specific files and I'm mis-remembering.

tfoote (changes requested):
Yeah, I think in general we'd rather not end up with a proliferation of files. For example if we moved the keys to the new location anyone who didn't reinit rosdep would suddenly stop resolving any of the keys which have been migrated.

tfoote:
I'm going to close this. Please add rosdep rules for the debian to base.yaml The npm support is something we should talk about more generally.

k-okada:
@tfoote  Ok, Please check https://github.com/ros-infrastructure/rosdep/pull/692 for npm support.

PR #21779, opened 2019-07-18
add pip entry for open3d-python, ipython, ipykernel

k-okada:


nuclearsandwich:
@k-okada can you please add pypi.org links for these packages to the pull request description per the contribution guidelines so this is easier to review. Thanks.

tfoote (changes requested):
We already have ipython from apt. We shouldn't have duplicate rules from pip. That will cause people to unknowingly get the overlay and things will potentially break. 

If there's newer features you want this is the sort of time when we need to be patient. 

If it's required for your use case I recommend setting up your own rosdep rules source that can override/extend the default ones and system maintainers can explicitly choose to opt in in which case they will know that they're going to get a newer potentially untested version.

clalancette:
No response for over a month, so I'm going to close this out.  Feel free to reopen if you'd like to continue working on these keys.  Thanks for the contribution.

PR #21791, opened 2019-07-19
ipython for bionic

floweisshardt:
adding ipython for bionic: https://packages.ubuntu.com/bionic/ipython

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -182,22 +182,7 @@ ipython:
   fedora: [python-ipython]
   gentoo: [dev-python/ipython]
   macports: [py27-ipython]
-  ubuntu:
-    bionic: [ipython]
-    lucid: [ipython]
-    maverick: [ipython]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
I think we can elide this to a plain `ubuntu: [ipython]` key and avoid specifying each distro. We wouldn't be able to do so if there were any currently supported distros that ipython is missing from or named differently.

tfoote (approved):
+1 for collapsing it, then lgtm

floweisshardt:
collapsed all ubuntu entries as requested in https://github.com/ros/rosdistro/pull/21791/commits/3c2321fa7dd6fcab60225e29cd2b508471737079

nuclearsandwich (approved):

nuclearsandwich:
Thank you!

floweisshardt:
anything else I should modify before merging?

PR #21806, opened 2019-07-21
Update distribution.yaml

linmianhao:
Adding my_repo_name to documentation index for distro.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -16619,5 +16619,16 @@ repositories:
       url: https://github.com/stonier/zeroconf_msgs.git
       version: indigo
     status: maintained
+   jaka_ros_package:
+    doc:
+      type: git
+      url: https://github.com/linmianhao/jaka_ros_package.git
+      version: kinetic
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/linmianhao/jaka_ros_package.git
+      version: kinetic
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -16619,5 +16619,16 @@ repositories:
       url: https://github.com/stonier/zeroconf_msgs.git
       version: indigo
     status: maintained
+   jaka_ros_package:
```
sloretz: It looks like this has 3 spaces in front, when it should have 2. See the [CI failure here](https://travis-ci.org/ros/rosdistro/jobs/561599239)

```
expected <block end>, but found '<block mapping start>'
  in "<string>", line 16622, column 4:
       jaka_ros_package:
       ^
```

CI will complain about the repositories not being in alphabetical order. Please place the `jaka_ros_package` entry in alphabetical order rather than appending it to the end of the file.

kinetic/distribution.yaml
```diff
@@ -16619,5 +16619,16 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/linmianhao/jaka_ros_package.git
+      version: kinetic
```
sloretz: I don't see a `kinetic` branch on that repo. Did you mean `master`?

sloretz (changes requested):

PR #21814, opened 2019-07-23
Adding lslidar_c16 to documentation index for distro

tongsky723:


indigo/distribution.yaml
```diff
@@ -6893,6 +6893,16 @@ repositories:
       type: git
       url: https://github.com/tum-vision/lsd_slam.git
       version: master
+  lslidar_C16:
```
sloretz: Please remove the changes to `indigo/distribution.yaml` since Indigo reached EOL in April.

sloretz (changes requested):

indigo/distribution.yaml
tongsky723: Hi, I have removed changes in indigo/distribution.yaml.

sloretz (approved):

PR #21622, opened 2019-06-26
noid-ros-pkg: 0.0.5-1 in 'kinetic/distribution.yaml' [bloom]

hi-kondo:
Increasing version of package(s) in repository `noid-ros-pkg` to `0.0.5-1`:

- upstream repository: https://github.com/seed-solutions/noid-ros-pkg.git
- release repository: https://github.com/seed-solutions/noid_ros_pkg-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.8.0`
- previous version for package: `null`

## noid_typef_bringup

- No changes

## noid_typef_description

- No changes

## noid_typef_moveit_config

```
* updated
* Contributors: hi-kondo
```

PR #21816, opened 2019-07-24
Revert "noid-ros-pkg: 0.0.5-1 in 'kinetic/distribution.yaml' [bloom]"

hi-kondo:
Reverts ros/rosdistro#21622

I would like to change repositry name, so once reverted, I'll provide fixed version.


tfoote:
If you just want to change the repository name, you can just change that one line. Or are you changing the package names too?

hi-kondo:
Yes, I would like to change the package name too.

PR #21835, opened 2019-07-25
add rules for pytest-mock

Arnatious:
- Debian https://packages.debian.org/buster/python3-pytest-mock
- Fedora https://apps.fedoraproject.org/packages/python-pytest-mock
- Gentoo https://packages.gentoo.org/packages/dev-python/pytest-mock
- Ubuntu https://packages.ubuntu.com/bionic/python3-pytest-mock

Add rosdep rules for pytest-mock, a thin wrapper around unittest.mock for pytest

Used in
```
./ament/ament_lint/ament_mypy/package.xml
```

Signed-off-by: Ted Kern <ted.kern@canonical.com>

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3276,6 +3276,12 @@ python-pytest-dependency-pip:
   ubuntu:
     pip:
       packages: [pytest-dependency]
+python-pytest-mock:
+  arch: [python2-pytest-mock]
+  debian: [python-pytest-mock]
+  fedora: [python-pytest-mock]
+  gentoo: [dev-python/pytest-mock]
+  ubuntu: [python-pytest-mock]
 python-pytest-qt-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


jacobperron (approved):

rosdep/python.yaml
```diff
@@ -5189,6 +5195,17 @@ python3-pytest:
...
   ubuntu: [python3-pytest]
+python3-pytest-mock:
+  alpine: [py3-pytest-mock]
+  arch: [python-pytest-mock]
+  debian: [python3-pytest-mock]
+  fedora: [python3-pytest-mock]
+  gentoo: [dev-python/pytest-mock]
+  openembedded:
+    pip:
+      packages: [pytest-mock]
```
tfoote: I'm not sure that pip makes sense on openembedded it's not deployable as such there. Can you explain why you added it there with a pip rule?

tfoote (changes requested):
Please link to the packages resources as per https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

tfoote:
Sorry, my review was based on an outdated version browser cache. Though my question about openembedded still stands.

rosdep/python.yaml
Arnatious: @tfoote I was trying to match pytest's footprint. pytest itself is available via rosdep on openembedded, but pytest-mock isn't in the layer index. Is this unnecessary?

Arnatious: @tfoote I've removed it in f0b8bf2

tfoote (approved):

PR #21838, opened 2019-07-26
add packages for KDE Neon to base.yaml

xaedes:
Added packages for KDE Neon which is based on ubuntu:bionic

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3,17 +3,20 @@ ace:
   debian: [libace-dev]
   gentoo: [dev-libs/ace]
   ubuntu: [libace-dev]
+  neon: [libace-dev]
 ack-grep:
   arch: [ack]
   debian: [ack-grep]
   fedora: [ack]
   gentoo: [sys-apps/ack]
   ubuntu: [ack-grep]
+  neon: [ack-grep]
 acpi:
   debian: [acpi]
   fedora: [acpi]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
My understanding is that KDE Neon is based directly on Ubuntu LTS with support for the latest KDE packages on top. Since KDE Neon is _mostly_ Ubuntu this is going to create a lot of additional maintenance. I would suggest instead contributing changes to rospkg and rosdep to alias Ubuntu as KDE Neon. You can see an example adding elementary OS here: https://github.com/ros-infrastructure/rosdep/blob/ab1f06ed84ee1e716eb611457b1168710415ea7a/src/rosdep2/platforms/debian.py#L79-L86

tfoote (changes requested):
Indeed I think that aliasing Neon to the ubuntu packages like all the other variants will be much more sustainable approach. 

If there are Neon specific elements rosdep keys I'd suggest that they be kept and curated in an additional rosdep.yaml that the neon instructions have you add additionally and are specific to the platform. 

Otherwise we will incur the significant cost of maintaining all the parallel keys.

clalancette:
It seems pretty clear that we won't be adding `NEON` as its own distribution type, and also that the fixes will be in other places (ros-infrastructure, for instance).  So I'm going to close this out.  If you disagree, please feel free to reopen this PR and continue the discussion.  Thanks for the contribution.

PR #21839, opened 2019-07-26
rosdep: fix python-ipdb ubuntu value

dandedrick:
'packages' is not a valid ubuntu version. This appears to have been a
typo and would prevent this package from working on any ubuntu version.

---
This change is [Reviewable](https://reviewable.io/reviews/ros/rosdistro/21839)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1980,8 +1980,7 @@ python-inputs-pip:
 python-ipdb:
   debian: [python-ipdb]
   gentoo: [dev-python/ipdb]
-  ubuntu:
-    packages: [python-ipdb]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
The `packages` key is seen under `pip:` in several packages but this does seem to be incorrect and based on the name that looks like it's meant to be an apt-installed package rather than a pip one.

nuclearsandwich (approved):

tfoote:
For future reference packages is correct under an installer. 

```
ubuntu: [foo bar]
```
Is shorthand for 

```
ubuntu: 
  apt:
    packages: [foo bar]
```

Because apt is the default installer and it knows to take packages as the default arguments.

PR #21866, opened 2019-08-01
Add grpcio-testing

DaikiMaekawa:
https://pypi.org/project/grpcio-testing/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1816,7 +1816,7 @@ python-grpcio:
       pip: [grpcio]
     xenial:
       pip: [grpcio]
-python-grpcio-testing:
+python-grpcio-testing-pip:
   debian:
     pip:
       packages: [grpcio-testing]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1816,6 +1816,21 @@ python-grpcio:
       pip: [grpcio]
     xenial:
       pip: [grpcio]
+python-grpcio-testing:
+  debian:
+    '*': [python-grpcio-testing]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@DaikiMaekawa can you please add links to the Debian and Ubuntu package listings for this package https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

@ros/rosdeputies I'm not sure if we're still adding rules that mix pip in with system packages in order to shore up older distributions. If not I think this needs to be split into python-grpcio-testing and python-grpcio-testing-pip

DaikiMaekawa:
@nuclearsandwich Oops, sry. For some reason, I thought that debian and ubuntu has binary packages on them own, but it's not. I modified the Pull Request to just use pip instead.

nuclearsandwich (changes requested):
Please add a `-pip` suffix to the rule name so that if there is a Debian package in the future the pip and apt rules do not share a key name.

Other than that looks good.

nuclearsandwich (changes requested):
Please add a `-pip` suffix to the rule name so that if there is a Debian package in the future the pip and apt rules do not share a key name.

Other than that looks good.

nuclearsandwich:
I do not know why that was doubled :upside_down_face:

DaikiMaekawa:
@nuclearsandwich Addressed your comment.

nuclearsandwich (approved):

nuclearsandwich:
> I do not know why that was doubled upside_down_face

Thanks for iterating.

PR #21906, opened 2019-08-02
Add apparmor

ruffsl:
[debian](https://packages.debian.org/buster/apparmor)
[ubuntu](https://packages.ubuntu.com/bionic/apparmor)

Not found for:
[fedora](https://apps.fedoraproject.org/packages/s/apparmor)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -61,6 +61,7 @@ apache2-mpm-prefork:
   ubuntu: [apache2-mpm-prefork]
 apparmor:
   debian: [apparmor]
+  fedora: null
   ubuntu: [apparmor]
 apr:
   alpine: [apr, apr-util]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -59,6 +59,9 @@ apache2-mpm-prefork:
     wheezy: [apache2-mpm-prefork]
   gentoo: ['www-servers/apache[apache2_mpms_prefork]']
   ubuntu: [apache2-mpm-prefork]
+apparmor:
+  debian: [apparmor]
+  ubuntu: [apparmor]
 apr:
   alpine: [apr, apr-util]
   arch: [apr, apr-util]


tfoote (approved):
It looks like Fedora doesn't support apparmor without a custom kernel: https://serverfault.com/questions/339842/apparmor-on-fedora-rhel-centos

@cottsay any suggestions?

clalancette:
@ruffsl This has been in draft state for almost 2 months.  Any thoughts on whether this is going to move forward?  Thanks.

ruffsl:
Sorry, this fell off my backlog. I don't have any further changes for this, and is ready.

clalancette:
> Sorry, this fell off my backlog. I don't have any further changes for this, and is ready.

Thanks, appreciated.

rosdep/base.yaml
```diff
@@ -59,6 +59,9 @@ apache2-mpm-prefork:
     wheezy: [apache2-mpm-prefork]
   gentoo: ['www-servers/apache[apache2_mpms_prefork]']
   ubuntu: [apache2-mpm-prefork]
+apparmor:
+  debian: [apparmor]
+  ubuntu: [apparmor]
```
clalancette: As we discussed above, this won't realistically ever be available for Fedora.  I'm thinking we should put an explicit `fedora: null` here, to indicate that it can't be available for Fedora (rather than just not being done, or something that could be done later).

clalancette (changes requested):

rosdep/base.yaml
ruffsl: I never realized rosdistro was using `null` to explicitly state unavailability. If this is common practice, then I can update the PR accordingly.

clalancette: It seems to be a relatively new practice, and to be honest, the tools don't support it all that well (rosdep just throws a stacktrace if you try to install a null key).  But I think it should be straightforward enough to enhance the tools to make it nicer.

ruffsl: Ok, I've added 61989c3

clalancette (approved):

PR #21920, opened 2019-08-06
cob_extern: 0.6.13-1 in 'kinetic/distribution.yaml' [bloom]

fmessmer:
Increasing version of package(s) in repository `cob_extern` to `0.6.13-1`:

- upstream repository: https://github.com/ipa320/cob_extern.git
- release repository: https://github.com/ipa320/cob_extern-release.git
- bloom version: `0.8.0`
- previous version for package: `0.6.12-0`

## cob_extern

```
* Merge pull request #99 <https://github.com/ipa320/cob_extern/issues/99> from fmessmer/travis_melodic
  [Melodic] melodify
* remove libqsopt
* remove libconcorde_tsp_solver
* Contributors: Benjamin Maidel, fmessmer
```

## libdlib

- No changes

## libntcan

- No changes

## libpcan

- No changes

## libphidgets

- No changes

## opengm

- No changes

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1562,6 +1562,16 @@ repositories:
       url: https://github.com/ipa320/cob_extern.git
       version: indigo_dev
     status: developed
+  cob_extern_unmaintained:
+    release:
+      packages:
+      - libconcorde_tsp_solver
+      - libqsopt
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/ipa320/cob_extern-release.git
+      version: 0.6.12-0
+    status: unmaintained
   cob_gazebo_plugins:
     doc:
       type: git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1548,17 +1548,15 @@ repositories:
     release:
       packages:
       - cob_extern
-      - libconcorde_tsp_solver
       - libdlib
       - libntcan
       - libpcan
       - libphidgets
-      - libqsopt
       - opengm
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ipa320/cob_extern-release.git
-      version: 0.6.12-0
+      version: 0.6.13-1
     source:
       type: git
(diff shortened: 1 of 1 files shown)


clalancette (commented):
As-is, this PR is removing two packages from Kinetic; `ros-kinetic-libconcorde-tsp-solver` and `ros-kinetic-libqsopt`.  While this is something we do allow, we recommend not to do it; you are effectively abandoning any downstream consumers of it.  If you still want to do it, please let us know explicitly.  Thanks.

fmessmer:
we removed the packages (from the current release) because we are striving for a kinetic/melodic-compatible release...and the removed packages are no longer supported (and broke the melodic compatibility of this repo)

our assumption was that even with removing the package in this release, they still would be available with the previos release version
is this not the case?

clalancette:
> our assumption was that even with removing the package in this release, they still would be available with the previos release version
> is this not the case?

I don't believe that is the case; I think if packages are removed from the distribution.yaml, the packages are removed on the next sync.  But I'm not 100% sure of that.  @tfoote , can you comment here?

fmessmer:
could some buildfarm/release expert help to clarify this, so I can finalize our release set to fix the current regression https://github.com/ipa320/cob_robots/issues/781#issuecomment-518880881 asap

tfoote:
The buildfarm does not actively clear packages. However they will be purged when an upstream dependency is released, and we trigger the removal of all downstream dependencies. If we didn't do that you would end up with potential crashes due to ABI incompatibility. 

As such it won't be gone immediately, but it's not expected to persist for any period of time. If you want the old version to persist, it needs to be in the index at the old version. For example you could make a trimmed down duplicate (with different name) repository entry and only list these packages at the old versions if you want the old versions to persist. 

Also if they're not in the index they're not accessible for any source build toolchains.

fmessmer:
thanks @tfoote 
is https://github.com/ros/rosdistro/pull/21920/commits/5df2ed22ec771ab457dfcb91fcaabd8663bc3d82 what you had in mind?

tfoote:
Exactly thanks.

PR #21958, opened 2019-08-07
merge 'libzmq3-dev' into 'zeromq3'

### LINKED ISSUE #21879: Duplicate keys for ZeroMQ
ZeroMQ can be referenced by two keys:

`zeromq3`:
https://github.com/ros/rosdistro/blob/9e46768ff51391a341f0745ba03b22422c2079ce/rosdep/base.yaml#L5623-L5628

`libzmq3-dev`:
https://github.com/ros/rosdistro/blob/9e46768ff51391a341f0745ba03b22422c2079ce/rosdep/base.yaml#L3852-L3856

This is confusing and I would recommend merging `libzmq3-dev` into `zeromq3`.

christian-rauch:
This PR resolves duplicated keys for ZeroMQ. Fixes #21879 .

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3850,7 +3850,9 @@ libzip-dev:
   gentoo: [dev-libs/libzip]
   ubuntu: [libzip-dev]
 libzmq3-dev:
+  arch: [zeromq]
   debian: [libzmq3-dev]
+  fedora: [zeromq-devel]
   gentoo: [net-libs/cppzmq]
   openembedded: [zeromq@meta-oe]
   ubuntu: [libzmq3-dev]
@@ -5621,11 +5623,11 @@ zbar:
   gentoo: [media-gfx/zbar]
   ubuntu: [libzbar-dev]
 zeromq3:
-  arch: [zeromq3]
-  debian: [libzmq3-dev]
-  fedora: [zeromq3]
+  arch: [zeromq]
+  debian: [libzmq5]
+  fedora: [zeromq]
   gentoo: [zeromq]
-  ubuntu: [libzmq3-dev]
+  ubuntu: [libzmq5]
 zip:
   debian: [zip]
   fedora: [zip]


rosdep/base.yaml
```diff
@@ -3849,11 +3849,6 @@ libzip-dev:
   fedora: [libzip-devel]
   gentoo: [dev-libs/libzip]
   ubuntu: [libzip-dev]
-libzmq3-dev:
```
tfoote: Removing this key will break anyone using it.

rosdep/base.yaml
```diff
@@ -5625,6 +5620,7 @@ zeromq3:
   debian: [libzmq3-dev]
   fedora: [zeromq3]
   gentoo: [zeromq]
+  openembedded: [zeromq@meta-oe]
```
tfoote: This is good to add.

tfoote (changes requested):
libzmq3 and libzmq3-dev are quite different semantically. One is expecting the headers while the other is just the library typically. 

If anything the libzmq3 rule could be reduced in scope to just use the library if the `-dev` package is bringing in a lot of extra dependencies.

Though oddly `libzmq3-dev` depends on `libzmq5` and both are in the major version 4 series... https://packages.ubuntu.com/bionic/libzmq3-dev but switching to `libzmq5` would be reasonable, it appears to be valid across all supported ubuntu distros (xenial to eoan)

But we shouldn't just remove a key that is currently valid.

christian-rauch:
> libzmq3 and libzmq3-dev are quite different semantically. One is expecting the headers while the other is just the library typically.

For Debian/Ubuntu packages, this is indeed true and makes sense. In rosdistro, both keys are pointing to the very same Debian packages that contain header files. Apart from this, is there any reason for a catkin package to just request the libraries without headers when being built? It would make sense to split libraries into rutime and devel (incl. header) if bloom release would make use of this. But at the moment, a bloom release will depend on the devel libraries with header, even if they are not used.

rosdep/base.yaml
christian-rauch: Is it possible to get an overview of which packages make use of this key? Otherwise, could we just forward this key to another, so that at least we do not need to maintain both (i.e. keep the package names up to date)?

tfoote (commented):
>  But at the moment, a bloom release will depend on the devel libraries with header, even if they are not used.

Our packaging toolchains (bloom and superflore) are both aware of the differences between build time and runtime dependencies and use them appropriately when generating packages. Build time dependencies will be installed when building the package. And runtime dependencies will be declared for installing the binaries. 

If the developer doesn't use the appropriate dependency tags clearly we can't do better: http://www.ros.org/reps/rep-0149.html#dependency-tags 

> In rosdistro, both keys are pointing to the very same Debian packages that contain header files.

This is why I proposed above to reduce the scope of the non-dev package resolution to make it semantically accurate. It's currently installing more than is necessary to get the library on those platforms. 

It's also the less used rule: https://index.ros.org/d/zeromq3/

christian-rauch:
I am aware of the different tags in the `package.xml` metafile, but I've not seen in practice that different library and devel packages are used.
As proposed, I reverted my change and have `zeromq3` pointing to the non-devel packages.

tfoote (approved):
Thanks for the update. It's hard sometimes to make sure to keep everything as neat as possible when you know that most developers won't bother to use it right. But as long as we keep the rules clean, at least they have the option to do it cleanly.

christian-rauch:
@tfoote Can this be merged?

nuclearsandwich:
> @tfoote Can this be merged?

Per our [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions) rosdep rule changes require two reviewers.

Also from our contribution guidelines it would greatly aid review if links to packages were provided in the pull request description.

For example, here's the link to the updated Archlinux package: https://www.archlinux.org/packages/community/x86_64/zeromq/

christian-rauch:
Added keys:

| key  | distro | path |
| ------------- | ------------- | ------------- |
| `libzmq3-dev`  | arch  | https://www.archlinux.org/packages/community/x86_64/zeromq/ |
| `libzmq3-dev`  | fedora  | https://apps.fedoraproject.org/packages/zeromq-devel |
| `zeromq3`  | arch  | https://www.archlinux.org/packages/community/x86_64/zeromq/ |
| `zeromq3`  | debian  | https://packages.debian.org/stretch/libzmq5 https://packages.debian.org/buster/libzmq5 |
| `zeromq3`  | fedora  | https://apps.fedoraproject.org/packages/zeromq |
| `zeromq3`  | ubuntu  | https://packages.ubuntu.com/xenial/libzmq5 https://packages.ubuntu.com/bionic/libzmq5 |

You could add a multiple approvals requirement directly on GitHub https://help.github.com/en/articles/enabling-required-reviews-for-pull-requests (point 6) to make it more explicit that the PR can only be merged if N reviewers approve the change.

nuclearsandwich:
> You could add a multiple approvals requirement directly on GitHub [help.github.com/en/articles/enabling-required-reviews-for-pull-requests](https://help.github.com/en/articles/enabling-required-reviews-for-pull-requests) (point 6) to make it more explicit that the PR can only be merged if N reviewers approve the change.

Thanks for the suggestion. As I understand it, it's not possible to apply that requirement to only a subset of the repository. In this case the dual-review requirement is for rosdep DB changes only.

nuclearsandwich (approved):
Thanks for iterating.

PR #21976, opened 2019-08-08
Add vcstool for ubuntu.

iluetkeb:


dirk-thomas (changes requested):
There seem to be multiple problems with this patch as well as with the existing entries:

* `ubunto` is misspelled - should be `ubuntu`
* The new value for Ubuntu maps correctly to `python-vcstool` (without a trailing `s` matching the name of the `rosdep` key. But all the other platforms map to `vcstools` (with a trailing `s` which is not matching the `rosdep` key.

I would suggest to create two separate rosdep keys `python-vcstool` and `python-vcstools` where all values match the correct package (without / with the trailing `s`).

iluetkeb:
@dirk-thomas Thanks for the comments. I'm a bit wary changing the existing entries just now. People didn't have a chance to depend on the correct package so far.

Since I'm mostly interested in doing this for ROS 2, I've added a *new* python3-vcstool entry instead. For some Linux distributions I had to rely on pip to do that. Hope that's okay -- since they offer the package, the only alternative would have been to omit them entirely.

iluetkeb:
Ooops, many things changed. Let me fix that.

dirk-thomas (approved):

dirk-thomas:
@tfoote can you please follow up on the wrong `python-vcstool` rosdep key and decide what to do about it.

tfoote:
We should just make duplicate entries, which map correctly on the different platforms. It's not a commonly used rosdep and there will be no good time to make the switch. Likely if anyone was using it they didn't notice it was installing the other tool as any of them would likely have ticketed it. And we also don't have a rule for our main platform.

dirk-thomas:
Sounds good to me. I created #22035 to fix the existing entries.

This PR can be merged independently of that.

PR #21980, opened 2019-08-09
parameter_assertions: 0.0.0-1 in 'melodic/distribution.yaml'

oswinso:
Increasing version of package(s) in repository `parameter_assertions` to 0.0.0-0:

* upstream repository: https://github.com/RoboJackets/rj-ros-common.git
* release repository: https://github.com/RoboJackets/rj-ros-common-release.git
* distro file: melodic/distribution.yaml
* previous version for package: null

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -5221,17 +5221,6 @@ repositories:
       url: https://github.com/ros-planning/panda_moveit_config.git
       version: melodic-devel
     status: maintained
-  parameter_assertions:
-    release:
-      tags:
-        release: release/melodic/{package}/{version}
-      url: https://github.com/RoboJackets/rj-ros-common-release.git
-      version: 0.0.0-1
-    source:
-      type: git
-      url: https://github.com/RoboJackets/rj-ros-common.git
-      version: master
-    status: developed
   parrot_arsdk:
     release:
       tags:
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -5221,6 +5221,17 @@ repositories:
       url: https://github.com/ros-planning/panda_moveit_config.git
       version: melodic-devel
     status: maintained
+  parameter_assertions:
```
clalancette: It's unusual, though allowed, to have the name of the repository disagree with the name of the package.  However, in this case I think it will cause problems for you further down the road.  If you want to add additional packages to that repository, you won't easily be able to with this structure.  There are a lot of examples in Melodic of how to do this better, one example is here: https://github.com/ros/rosdistro/blob/master/melodic/distribution.yaml#L160 .

But overall this isn't a blocker, so if you don't want to make any changes we can still merge.  I'll wait on merging this until I hear back from what you want to do.

clalancette (commented):
A structural issue that isn't blocking, but would make things nicer.

melodic/distribution.yaml
```diff
@@ -5221,6 +5221,17 @@ repositories:
...
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/RoboJackets/rj-ros-common-release.git
+      version: 0.0.0-1
```
clalancette: Oh, also, a version of 0.0.0 is highly unusual.  I'd suggest at least `0.1.0`.

melodic/distribution.yaml
oswinso: ah ok yeah that makes sense. I'm not sure how the `packages` key works but I'll try using that following https://github.com/ros/rosdistro/blob/cf72116af90c04982328f99abef1aad304fa7e35/melodic/distribution.yaml#L4405-L4437 to see if it works.

melodic/distribution.yaml
oswinso: Do I need to re-release to fix this?

clalancette: Yeah, you'd have to run `catkin_prepare_release --bump minor`, follow the instructions, then do another `bloom-release`.  There's a tutorial on how to do it: http://wiki.ros.org/bloom/Tutorials/ReleaseCatkinPackage

oswinso: Alright I'm closing this one since `bloom-release` made a PR automatically [here](https://github.com/ros/rosdistro/pull/21986)

PR #21982, opened 2019-08-09
add python-zmq for latest ubuntu distros

fmessmer:
https://packages.ubuntu.com/search?keywords=python-zmq

or should I use the default notation?
`    '*': [python-zmq]`


tfoote:
It would be great to switch to use the wildcard notation. When trusty is no longer supported we can then just collapse it to the default rule since the package is unchanging.

fmessmer:
using wildcard now

fmessmer:
this is blocking us quite a bit atm...could this please be reviewed/merged?...thanks

tfoote (approved):
Thanks for iterating. I also removed the now redundant rules. It looks good to me.

floweisshardt (approved):

clalancette (approved):

PR #21987, opened 2019-08-09
Remove jog_arm from rosdistro. It is part of MoveIt now

AndyZe:


sloretz:
@AndyZe it looks like there is a merge conflict. Would you mind resolving it?

PR #21959, opened 2019-08-07
Mark Bouncy as end-of-life.

nuclearsandwich:
Bouncy's support window closed in June. http://www.ros.org/reps/rep-2000.html#bouncy-bolson-june-2018-june-2019

PR #21989, opened 2019-08-09
Bouncy status EOL type ros2

sloretz:
Follow up of https://github.com/ros/rosdistro/pull/21959 . Assuming `ardent` is configured correctly, it looks like the `end-of-life` belongs on `distribution_status` instead of `distribution_type` for `bouncy`


tfoote (approved):
Good catch @sloretz this is definitely the right way.

nuclearsandwich:
Yes. Thanks. I.... can offer no explanation...

PR #21991, opened 2019-08-10
add wildcard for python-rosinstall

fmessmer:
https://packages.ubuntu.com/search?keywords=python-rosinstall


tfoote (approved):
I also collapsed the rules here to avoid the redundancy.

clalancette (approved):

PR #21774, opened 2019-07-17
librealsense2: 2.25.0-3 in 'kinetic/distribution.yaml' [bloom]

doronhi:
Increasing version of package(s) in repository `librealsense2` to `2.25.0-3`:

- upstream repository: https://github.com/doronhi/librealsense.git
- release repository: https://github.com/IntelRealSense/librealsense2-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.8.0`
- previous version for package: `2.23.0-1`

PR #21996, opened 2019-08-12
librealsense2: 2.25.2-1 in 'kinetic/distribution.yaml' [bloom]

doronhi:
Increasing version of package(s) in repository `librealsense2` to `2.25.2-1`:

- upstream repository: https://github.com/doronhi/librealsense.git
- release repository: https://github.com/IntelRealSense/librealsense2-release.git
- bloom version: `0.8.0`
- previous version for package: `2.23.0-1`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6141,7 +6141,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/IntelRealSense/librealsense2-release.git
-      version: 2.23.0-1
+      version: 2.25.2-1
     source:
       type: git
       url: https://github.com/IntelRealSense/librealsense.git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6143,6 +6143,7 @@ repositories:
       url: https://github.com/IntelRealSense/librealsense2-release.git
       version: 2.25.2-1
     source:
+      test_pull_requests: true
       type: git
       url: https://github.com/doronhi/librealsense.git
       version: ros_debian
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6145,7 +6145,7 @@ repositories:
     source:
       type: git
       url: https://github.com/doronhi/librealsense.git
-      version: 2.25.2
+      version: ros_debian
     status: maintained
   libsegwayrmp:
     doc:
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6144,8 +6144,8 @@ repositories:
       version: 2.25.2-1
     source:
       type: git
-      url: https://github.com/IntelRealSense/librealsense.git
-      version: master
+      url: https://github.com/doronhi/librealsense.git
+      version: 2.25.2
     status: maintained
   libsegwayrmp:
     doc:


doronhi:
This PR replace #21774.
I still don't get the results mails but I believe the build log is here:
http://build.ros.org/job/Kdev__librealsense2__ubuntu_xenial_amd64/5/
It states that the revision is : fa13505bb9feb828f575d4d8cd5b0db822c90d01
Which is the last commit in the repository: 
https://github.com/IntelRealSense/librealsense/commits/master

This PR states that:
•	upstream repository: https://github.com/doronhi/librealsense.git
•	release repository: https://github.com/IntelRealSense/librealsense2-release.git
 and in librealsense2-release.git, the file tracks.yaml also states:
-	vcs_uri: https://github.com/doronhi/librealsense.git
-	last_release: v2.25.0 (which I later, wrongly, modified to v2.25.2)

So, Can you please help me understand why the repository : github.com/IntelRealSense/librealsense is taken for the build?

kinetic/distribution.yaml
```diff
@@ -6141,7 +6141,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/IntelRealSense/librealsense2-release.git
-      version: 2.23.0-1
+      version: 2.25.2-1
     source:
       type: git
       url: https://github.com/IntelRealSense/librealsense.git
```
sloretz: > So, Can you please help me understand why the repository : github.com/IntelRealSense/librealsense is taken for the build?

I think devel jobs Like [K**dev**__librealsense2__ubuntu_xenial_amd64](http://build.ros.org/job/Kdev__librealsense2__ubuntu_xenial_amd64/5/) use the repo from the `source` entry. Did you want to change this to `https://github.com/doronhi/librealsense.git`?

kinetic/distribution.yaml
```diff
@@ -6141,11 +6141,12 @@ repositories:
...
         release: release/kinetic/{package}/{version}
       url: https://github.com/IntelRealSense/librealsense2-release.git
-      version: 2.23.0-1
+      version: 2.25.2-1
     source:
+      test_pull_requests: true
       type: git
-      url: https://github.com/IntelRealSense/librealsense.git
-      version: master
+      url: https://github.com/doronhi/librealsense.git
```
clalancette: This is an unusual setup.  You have the doc portion pointing to https://github.com/IntelRealSense/librealsense.git, and the source pointing to https://github.com/doronhi/librealsense.git.  That is not recommended.

Also, why are we using a fork now?  Can you not get the changes needed into upstream https://github.com/IntelRealSense/librealsense.git?  That would be highly preferred.

doronhi: The doc and the source should point to the same location, no argue about it.
I also prefer to have the upstream from the source and not use a fork but I'm not sure it will happen. The default build values are different and it's not possible, as far as I understand it, to set different build options for the ros buildfarm. 
So far I found out that by default librealsense is built with examples requiring X11 and using cache. I disabled both options in my fork and looking forward to see if these are all the modifications required.

nuclearsandwich: @doronhi Bloom has a workflow for applying patches for release. The best tutorial I can find off the top of my head is https://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty#Adding_an_Install_Rule_as_a_Patch rather than adding install rules you could follow that tutorial to customize the default build options.

It's also possible to set build flags for the platform-specific build scripts but it's better to patch at the release source level so that everyone benefits from the patch rather than just the people using the prebuilt binaries.

doronhi: I agree with the policy that it's best to keep all the patches in the source level. I hope it will be possible but the first release must be done and prove to work from the fork before I could move on and try to modify the source.
Any idea why the current PR isn't being built on the ROS buildfarm?
I added the "test_pull_request" flag and set it to true on [distribution.yaml](https://github.com/ros/rosdistro/blob/686b0e3c2ca8e68ef43c35944089ba80b5559758/kinetic/distribution.yaml#L6146)
I added `http://build.ros.org/ghprbhook` to the librealsense webhooks [here](https://github.com/doronhi/librealsense/settings/hooks)
I added `ros-pull-request-builder` as a librealsense collaborator [here](https://github.com/doronhi/librealsense/settings/collaboration)
Any idea what else is missing?

nuclearsandwich: > Any idea why the current PR isn't being built on the ROS buildfarm?

It's not being built because that change has not been merged into the master distribution.yaml

> I hope it will be possible but the first release must be done and prove to work from the fork before I could move on and try to modify the source.

I think it's @tfoote's call whether this change is acceptable since he is the release manager for Kinetic.

doronhi: I tested librealsense from [my fork](https://github.com/doronhi/librealsense/releases/tag/2.25.2) with docker and it worked but obviously I would like to check my code with the build farm.
I can't modify the librealsense source level until I have a working version, fit for the ROS buildfarm. 
It seems that for that to happen I need my PR with the `source` section pointing to my fork merged into the master first. Is that true or am I missing something? @tfoote 

I tried following [the patches tutorial](https://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty#Adding_an_Install_Rule_as_a_Patch) but I couldn't understand how to make it change the build flags of librealsense or if this is the right approach at all.
I must assume that I wont be able to change librealsense's **default** build flags. I may be able to add a patch directory with files in it. Is this an option for modifying CMake's default flags to match the requirements of ros buildfarm?

tfoote: We can try this as is. That sort of patch tutorial is I think the right way to do it. Especially if it's just setting the flags. Please ask a question on answers.ros.org and we can try to help you there get it worked out.

PR #22008, opened 2019-08-14
[crystal][laser_geometry] use crystal branch as the upstream.

seanyen:
A recent [change](https://github.com/ros-perception/laser_geometry/pull/46) from `ros2` branch of `laser_geometry` introduces a new dependency `eigen3_cmake_module` which doesn't exist on crystal. If someone builds the crystal from the upstream repository, it will hit build break. So I suggest to swap the source to `crystal` branch instead.

### DIFF
--- a/crystal/distribution.yaml
+++ b/crystal/distribution.yaml
@@ -743,7 +743,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-perception/laser_geometry.git
-      version: ros2
+      version: crystal
     release:
       tags:
         release: release/crystal/{package}/{version}
@@ -752,7 +752,7 @@ repositories:
     source:
       type: git
       url: https://github.com/ros-perception/laser_geometry.git
-      version: ros2
+      version: crystal
     status: maintained
   launch:
     doc:


seanyen:
@sloretz @jonbinney JFYI, I think the `laser_geometry` might need some updates for its source branch `crystal` distro.

nuclearsandwich:
@sloretz this seems like the right thing to do to me. Even if we are going to backport the eigen3_cmake_module package to Crystal and the two laser_geometry changes along with it that should be done via explicit backport.

nuclearsandwich (approved):
I'd like to get a review from @sloretz as well but this seems like the right change to me.

sloretz (approved):
This looks right to me too. Introducing an `eigen3_cmake_module` dependency in Crystal was unintentional. I don't have any plans to backport it.

jonbinney:
Sounds good, and yes, we can backport any crucial functionality from the ros2 branch to the crystal branch as needed. @seanyen any PR's in particular you think should be backported?

seanyen:
> @seanyen any PR's in particular you think should be backported?

I think the answer would be related to whether the maintainer of `eigen3_cmake_module` would like to make a release and support it in `crystal`. @sloretz any comments?

nuclearsandwich:
I'm going to merge this change since it reconciles what you'd get building from upstream source repositories with what you'd get building our packages.

I don't think changes related to eigen3_cmake_module need to be backported into Crystal to resolve any issues.

PR #22012, opened 2019-08-14
Add python3-rosdistro-modules

claireyywang:
This package is used in https://github.com/ros2/ros2cli/tree/claire_debugtools/ros2doctor to check ROS distribution and status. Package repo can be found at https://github.com/ros-infrastructure/rosdistro.

python3-rosdistro web indexes:
Debian: https://packages.debian.org/sid/python3-rosdistro
Ubuntu: https://packages.ubuntu.com/bionic/python3-rosdistro
Fedora: https://apps.fedoraproject.org/packages/python3-rosdistro
Openembedded: https://layers.openembedded.org/layerindex/recipe/60790/
Rhel: no index found

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5227,6 +5227,13 @@ python3-rosdep:
   fedora: [python3-rosdep]
   gentoo: [dev-util/rosdep]
   ubuntu: [python3-rosdep]
+python3-rosdistro-modules:
+  debian: [python3-rosdistro-modules]
+  fedora: [python3-rosdistro]
+  gentoo: [dev-python/rosdistro]
+  openembedded: [python3-rosdistro@meta-ros]
+  rhel: ['python%{python3_pkgversion}-rosdistro']
+  ubuntu: [python3-rosdistro-modules]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@claireyywang the [contribution docs](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules) for rosdep keys usually ask for links to web-browseable package archives.

This package comes from our own repositories which do not have a browseable web index but it would be nice to at least mention that fact and identify whether there are any places that the package would be expected and not found.

tfoote (approved):

nuclearsandwich:
It's worth noting that for Debian and Ubuntu, this key resolves to the python3-rosdistro-modules packages that we provide via the ROS repositories so that the python libraries can be installed without the python script binaries which conflict between python2 and python3.

For other platforms that are using the upstream package it's likely that this will cause conflicts with the python2 package. This is not a blocker just something to be aware of.

nuclearsandwich (approved):

PR #22031, opened 2019-08-16
[rosdep] Alpine corrections

nuclearsandwich:
This is a followup to #22011 and #21095

I don't think there's any fallout to removing the pycryptodome Alpine definition since we caught it before use.

Before we merge this we should check with @russkel to make sure that they can report any ROS 2 packages which are mistakenly relying on the python2 key `python-qt5-bindings` instead of `python3-qt5-bindings`.

Edit: somehow the PR references were off.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3423,7 +3423,6 @@ python-qt4-gl:
   gentoo: ['dev-python/pyside[opengl]', 'dev-python/PyQt4[opengl]']
   ubuntu: [python-qt4-gl]
 python-qt5-bindings:
-  alpine: [py3-qt5]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
> Before we merge this we should check with @russkel

@tfoote oops?

russkel:
Thanks for fixing this. I will add pycryptodome to the TODO list as well.

nuclearsandwich:
> Thanks for fixing this. I will add pycryptodome to the TODO list as well.

It's only needed for ROS 1 at this time. I'm not sure if that changes whether or not you're interested in adding it.

russkel:
Ah, will throw it onto the ROS1 todo then.

PR #22050, opened 2019-08-19
Removed udp_driver

esteve:
This removes `udp_driver` from the Autoware.Auto repository so that we can release it from the `transport_drivers` repository.

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -28,7 +28,6 @@ repositories:
       - point_cloud_fusion
       - ray_ground_classifier
       - ray_ground_classifier_nodes
-      - udp_driver
       - velodyne_driver
       - velodyne_node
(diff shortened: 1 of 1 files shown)


esteve:
@nuclearsandwich I don't know if this is the right way to do it. I want to remove the `udp_driver` package from the registry so I can release the one from `ros-drivers`. We're in the process of spinning off the drivers from Autoware.Auto into `ros-drivers`.

nuclearsandwich:
@esteve since bloom won't allow you to release a package that's a member of another release I think the two-step method is the only way to go.

nuclearsandwich (approved):

esteve:
@nuclearsandwich thanks, I've opened https://github.com/ros/rosdistro/pull/22051 to add it again.

PR #22064, opened 2019-08-20
Add gstreamer1.0-alsa

mvollrath:
Sub-package for audio I/O.
* https://packages.ubuntu.com/xenial/gstreamer1.0-alsa
* https://packages.debian.org/jessie/gstreamer1.0-alsa
* Does not seem to exist for arch, fedora, or gentoo


nuclearsandwich (approved):
alsa-libs are a dependency of [gst-plugins-base](https://www.archlinux.org/packages/extra/x86_64/gst-plugins-base/) on Archlinux but I'd rather leave the key undefined until someone whose actually going to use this needs to do so.

I would have thought that Gentoo handled alsa support via a use flag but I don't see one on gstreamer.

PR #22071, opened 2019-08-21
Add ifcfg-pip

claireyywang:
This package is used in [ros2doctor network feature](https://github.com/ros2/ros2cli/blob/claire/ros2doctor-network/ros2doctor/ros2doctor/api/network.py) to check network configuration. 
Package repo can be found at https://github.com/ftao/python-ifcfg, and it is compatible with Python2 and Python3.
Web index: https://pypi.org/project/ifcfg/, no platform specific package index found. 
Explicit vendor package can be found at https://github.com/ros2/ifcfg_vendor

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -154,6 +154,19 @@ gunicorn:
   fedora: [python-gunicorn]
   gentoo: [www-servers/gunicorn]
   ubuntu: [gunicorn]
+ifcfg-pip:
+  debian:
+    pip:
+      packages: [ifcfg]
+  fedora:
+    pip:
+      packages: [ifcfg]
+  osx:
+    pip:
+      packages: [ifcfg]
+  ubuntu:
+    pip:
+      packages: [ifcfg]
 imgaug-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


claireyywang:
working on a `deb` vendor package

nuclearsandwich (approved):

rosdep/python.yaml
```diff
@@ -154,6 +154,19 @@ gunicorn:
   fedora: [python-gunicorn]
   gentoo: [www-servers/gunicorn]
   ubuntu: [gunicorn]
+ifcfg-pip:
```
nuclearsandwich: Since we're using this for ROS 2 we'll want the key name for the platform installer to be `python3-ifcfg`. I think this key name is fine for pip. An alternative might be `python3-ifcfg-pip` but since the pip key is not python3 specific I don't feel a preference for it.

nuclearsandwich:
> working on a `deb` vendor package

I like explicit vendor packages rather than packaging things off the farm and importing them as we do for opensplice and lark-parser but let's check in with some other folks tomorrow to decide what to do for ifcfg. I have also added this to my big list of things to try and get upstreamed into Debian for future ROS 2 releases.

tfoote (changes requested):
If we want this to be a core tool dependency I think that using a vendor package is better than using pip, because we can't use pip as a core dependency when we're packaging. 

And if we're going to vendor it then we I think it's better not to have conflicting pip rules for it.

claireyywang:
Closing as a `ifcfg` debian package is made available; `ifcfg_vendor` is archived in ros2.repos. Will open a new PR if needed.

PR #22073, opened 2019-08-22
added yapf for ubuntu bionic rosdep/python.yaml

tylernewnoise:
rosdep rule for [yapf](https://github.com/google/yapf) on ubuntu 18.04 bionic systems

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5620,6 +5620,9 @@ yamllint:
     xenial: [yamllint]
     yakkety: [yamllint]
     zesty: [yamllint]
+yapf:
+  ubuntu:
+    bionic: [python-yapf]
 yasm:
   arch: [yasm]
   debian: [yasm]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5621,8 +5621,12 @@ yamllint:
     yakkety: [yamllint]
     zesty: [yamllint]
 yapf:
+  arch: [yapf]
+  debian:
+    buster: [yapf]
+    stretch: [yapf]
   ubuntu:
-    bionic: [python-yapf]
+    bionic: [yapf]
 yasm:
   arch: [yasm]
   debian: [yasm]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5480,3 +5480,6 @@ yapf:
       pip:
         packages: [yapf]
     zesty: [yapf]
+    bionic:
+      pip:
+        packages: [yapf]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5458,6 +5458,9 @@ yapf:
     stretch: [yapf]
   ubuntu:
     artful: [yapf]
+    bionic:
+      pip:
+        packages: [yapf]
     saucy:
       pip:
         packages: [yapf]
@@ -5480,6 +5483,4 @@ yapf:
       pip:
         packages: [yapf]
     zesty: [yapf]
-    bionic:
-      pip:
-        packages: [yapf]
+
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5458,9 +5458,6 @@ yapf:
     stretch: [yapf]
   ubuntu:
     artful: [yapf]
-    bionic:
-      pip:
-        packages: [yapf]
     saucy:
       pip:
         packages: [yapf]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5480,4 +5480,3 @@ yapf:
       pip:
         packages: [yapf]
     zesty: [yapf]
-


nuclearsandwich (dismissed):
It looks like this package is available in apt repositories starting with zesty. We should prefer the platform package to one installed via pip.

The bionic entry also needs to be placed in alphabetical order rather than at the end of the list.

tylernewnoise:
> It looks like this package is available in apt repositories starting with zesty. We should prefer the platform package to one installed via pip.
Ah, didn't notice, thx! Unfortunatly the platform package is a little bit outdated: 0.20 (apt) vs. 0.28 (pip).

nuclearsandwich:
@ros/rosdeputies should we split the key into platform and pip variants? I'm not sure if there's a way to do so gracefully without breaking the current key for  pip users.

tfoote (changes requested):
It appears that yapf is now packaged as of bionic as a debian package. We should use that key for bionic and newer: https://packages.ubuntu.com/bionic/yapf

tylernewnoise:
> @ros/rosdeputies should we split the key into platform and pip variants? I'm not sure if there's a way to do so gracefully without breaking the current key for pip users.

Actually, this would be great. Because for example if I switch from 16.04 (Xenial) with Kinetic, where yapf is version 0.28, to 18.04 (Bionic) with Melodic I would fall back to yapf 0.20 if I install the debian package. Hence I would recommend the pip-package or the split between platform and pip. 
This could be achieved by adding python-yapf (which is an alias for the debian yapf package) to the base.yaml. Or by adding an extra entry named yapf-pip to the python.yaml for every ubuntu release since bionic. Don't know wich one should be preferred...

tfoote:
I would advocate against creating a parallel pip based key. If one package uses the standard one it will work fine. But if some other package then asks for the pip one it will mask the debian one and potentially break it. In general if we have a key for apt we shouldn't create a duplicate parallel one for pip. That's part of being in the distro is that you stick with what's curated into that distro. We're providing the database of the canonical version that people can rely upon.

clalancette:
@tylernewnoise Any thoughts on making an update here?  I'll likely close this out in a couple more weeks if we don't see any movement.  Thanks in advance.

tylernewnoise:
Sorry for the delay.
Ok, I'm a bit unsure what to do now. Short summary so far:
For Ubuntu versions below bionic yapf is not available as debian system package, hence the key 'yapf' for pip.
For Ubuntu versions from bionic on yapf is available as a debian package, but in a lower version. So using the key for the system package from bionic on would result in a downgrade.
Suggestion was to add an addtional key so the users can choose which version they prefer. Personally I would go this way. But then still it is to decide what the name of the key for the pip/deb package should be.
I would rule against to keep the key for the deb package only, because of the downgrade when coming from 16.04 for example.

clalancette:
I know you've been waiting on us for this.



> Sorry for the delay.
> Ok, I'm a bit unsure what to do now. Short summary so far:
> For Ubuntu versions below bionic yapf is not available as debian system package, hence the key 'yapf' for pip.
> For Ubuntu versions from bionic on yapf is available as a debian package, but in a lower version. So using the key for the system package from bionic on would result in a downgrade.
> Suggestion was to add an addtional key so the users can choose which version they prefer. Personally I would go this way. But then still it is to decide what the name of the key for the pip/deb package should be.
> I would rule against to keep the key for the deb package only, because of the downgrade when coming from 16.04 for example.

Thanks for doing the research here.  I don't think that the fact that the version is higher in pip should prevent us from putting the debian package key only.  For almost all packages available in both the debian/ubuntu repositories and pip, the pip version is newer.

To unstick this, my suggestion is to keep this PR simple and just add in the key for the debian package on the platforms it is supported on.  We can then separately deal with adding a `-pip` key in another PR.  Does that make sense?

tylernewnoise:
Ok, removed the pip-entry and added the key for the apt-repository for ubuntu bionic. :)

rosdep/python.yaml
```diff
@@ -5480,3 +5480,4 @@ yapf:
       pip:
         packages: [yapf]
     zesty: [yapf]
+
```
clalancette: Please get rid of this spurious change.

rosdep/base.yaml
```diff
@@ -5620,6 +5620,9 @@ yamllint:
     xenial: [yamllint]
     yakkety: [yamllint]
     zesty: [yamllint]
+yapf:
+  ubuntu:
+    bionic: [python-yapf]
```
clalancette: So, there's actually an interesting thing here.  According to packages.ubuntu.com, there are actually separate packages: `yapf` provides the Python2 command-line utility `/usr/bin/yapf`, while the package `python-yapf` provides the library.  The situation is similar for python3.  If you want this key to be the command-line utility, then we should use `yapf` as the key name, and `yapf` as the package name.  If you want this key to be the library, then we should use `python-yapf` as the key name, and `python-yapf` as the package.  That way, we can have another set of keys for Python3 as well.

clalancette (changes requested):
Sorry for all of the back-n-forth here, but I just realized the key name doesn't match the package.

Also, while you are at it, please add in Fedora and Debian keys for this package, as it seems to be available on both (see https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions on how to find them).

Thanks.

rosdep/python.yaml
tylernewnoise: Ups, my bad. Removed.

rosdep/base.yaml
tylernewnoise: Oh, I see. I changed it to the command-line utility. Also added the entry for ArchLinux. But couldn't find the package for Fedora 31.

clalancette: Awesome, thanks!

clalancette (approved):

tfoote (approved):

PR #22105, opened 2019-08-28
kinetic: add rxcpp_vendor source and doc entries.

gavanderhoorn:
As per subject.


gavanderhoorn:
@wasowski @henrik7264 fyi.

PR #22119, opened 2019-08-29
Add python3-boto3

AAlon:
This will add the python 3 version of boto3. Needed mainly for ROS2 applications since the Python version is 3 by default so they cannot use the existing `python-boto3` key.

[debian](https://packages.debian.org/search?searchon=names&keywords=boto3)
[fedora](https://apps.fedoraproject.org/packages/python3-boto3)
[opensuse](https://software.opensuse.org/package/python3-boto3)
[ubuntu](https://packages.ubuntu.com/bionic/python3-boto3)

Thanks!

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4909,7 +4909,6 @@ python3-boto3:
   opensuse: [python3-boto3]
   ubuntu:
     '*': [python3-boto3]
-    trusty: null
 python3-cairo:
   arch: [python-cairo]
   debian: [python3-cairo]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4903,6 +4903,13 @@ python3-backoff-pip:
   ubuntu:
     pip:
       packages: [backoff]
+python3-boto3:
+  debian: [python3-boto3]
+  fedora: [python3-boto3]
+  opensuse: [python3-boto3]
+  ubuntu:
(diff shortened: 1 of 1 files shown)


nuclearsandwich (commented):
I think we might even be able to drop the Trusty rule but I don't mind keeping it if it brings value to the contributor.

nuclearsandwich (approved):

AAlon:
Just for consistency sake with the `python-boto3` key :) 
Is it possible to merge this in? 
Thanks!

clalancette:
> @nuclearsandwich Just for consistency sake with the `python-boto3` key :) we don't really need it.

If you don't need it, I'll suggest dropping it completely and just going with the glob rule.  Trusty isn't supported anymore anyways, so there is no reason to call it out specifically.

AAlon:
> > @nuclearsandwich Just for consistency sake with the `python-boto3` key :) we don't really need it.
> 
> If you don't need it, I'll suggest dropping it completely and just going with the glob rule. Trusty isn't supported anymore anyways, so there is no reason to call it out specifically.

Sounds good, removed.

clalancette:
I simplified the rule for ubuntu even more (the '*' is superfluous with a single entry).  Once CI comes back green, I'll merge.

clalancette (approved):

PR #22131, opened 2019-08-31
slam_toolbox: 0.4.0-1 in 'melodic/distribution.yaml' [bloom]

SteveMacenski:
Increasing version of package(s) in repository `slam_toolbox` to `0.4.0-1`:

- upstream repository: https://github.com/SteveMacenski/slam_toolbox.git
- release repository: https://github.com/SteveMacenski/slam_toolbox-release.git
- bloom version: `0.8.0`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -8346,6 +8346,22 @@ repositories:
       url: https://github.com/ros-perception/slam_karto.git
       version: melodic-devel
     status: maintained
+  slam_toolbox:
+    doc:
+      type: git
+      url: https://github.com/SteveMacenski/slam_toolbox.git
+      version: melodic-devel
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/SteveMacenski/slam_toolbox-release.git
+      version: 0.4.0-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/SteveMacenski/slam_toolbox.git
+      version: melodic-devel
+    status: maintained
   soem:
     doc:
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -8346,6 +8346,22 @@ repositories:
...
+  slam_toolbox:
+    doc:
+      type: git
+      url: https://github.com/SteveMacenski/slam_toolbox.git
+      version: melodic-devel
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/SteveMacenski/slam_toolbox-release.git
+      version: 0.4.0-1
```
clalancette: So this is released as version 0.4.0, but I see in the upstream repository that there are tags for much later numbers, i.e. there is a 1.0.0.  While that doesn't really matter for this to work on the buildfarm, it is strange to say the least.  Can I recommend that you remove/deprecate the 0.4.0 tag and use a number > 1.0.0 for the version?  At least that way it will be internally consistent.

clalancette (changes requested):
There are some weird version numbers here that I'd like clarification on before merging.

melodic/distribution.yaml
SteveMacenski: Yeah, so the 1.0.0 was just a release, but there's no bloom or package.xml version along side it. The 1.0.0 is there for another entity to use as a static version for some features they were looking for. 

I can redo this for a > 1.0.0 number if you like, nbd

SteveMacenski:
Closing for https://github.com/ros/rosdistro/pull/22145

PR #22161, opened 2019-09-05
[actionlib] update upstream branch

seanyen:
Recently there are some check-ins on `indigo-devel` not compatible with the `melodic-devel` of `realtime_tools` and it causes build break if everything is built from the upstream. I would recommend to use `melodic-devel` instead as the `actionlib` upstream branch.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -58,7 +58,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros/actionlib.git
-      version: indigo-devel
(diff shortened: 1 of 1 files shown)


seanyen:
cc @sloretz

clalancette (approved):
This seems reasonable to me, but I'll let @mjcarroll review and merge this one since he is the maintainer of actionlib.

sloretz (approved):

mjcarroll (approved):

PR #22209, opened 2019-09-09
Add zivid_camera for kinetic & fix zivid_camera status for melodic

apartridge:


clalancette:
In general, it is best to submit separate PRs for separate ROS distros.  The general reason is that the releases aren't synchronized, so we may be in a situation where one distro is frozen before a release while the other is not.  Right now both Kinetic and Melodic are unfrozen, so I'll merge this as-is, but please keep it in mind for the future.

PR #22229, opened 2019-09-11
Add dbc package

Timple:


clalancette:
It's generally better to open one PR per rosdistro.  That's because the releases for the various ROS distros isn't synchronized, so you can end up with a situation where we can't merge the PR because one of them is in freeze before a sync (that's the case right now with Kinetic).  So I'd suggest closing this PR and opening a separate one for Kinetic and Melodic.  We can merge the latter immediately, and merge the kinetic one when it comes out of freeze.

PR #22238, opened 2019-09-12
Add dbc package for kinetic

Timple:


tfoote (changes requested):
Same as: https://github.com/ros/rosdistro/pull/22239/files#r323872978

Note that there's also a dbc package inside which is simiarly hard to know what it means. A quick search of dbc doesn't get it on the first try. https://www.google.com/search?q=dbc searching for 'dbc file' which is what's mentioned in the package.xml still gets the wrong thing first (something related to Visual FoxPro database development system. I would suggest naming that package can_dbc as that's something that at least will get you to what I believe the right semantic meaning from the name.

Remember that these names are global and not everyone is in say the automotive context.

ROS Naming conventions: http://wiki.ros.org/ROS/Patterns/Conventions

PR #22248, opened 2019-09-16
rxros: 0.1.0-1 in 'melodic/distribution.yaml' [bloom]

gavanderhoorn:
Increasing version of package(s) in repository `rxros` to `0.1.0-1`:

- upstream repository: https://github.com/rosin-project/rxros.git
- release repository: https://github.com/rosin-project/rxros-release.git
- bloom version: `0.8.0`

## rxros

```
* First release of RxROS.
* Contributors: Henrik Larsen, gavanderhoorn
```

## rxros_tf

```
* First release of RxROS TF.
* Contributors: Henrik Larsen, gavanderhoorn
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -8199,6 +8199,16 @@ repositories:
       url: https://github.com/rosin-project/rxcpp_vendor.git
       version: master
     status: maintained
+  rxros:
+    release:
+      packages:
+      - rxros
+      - rxros_tf
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/rosin-project/rxros-release.git
+      version: 0.1.0-1
+    status: developed
   sainsmart_relay_usb:
     doc:
       type: hg


gavanderhoorn:
@henrik7264 @wasowski: FYI.

gavanderhoorn:
Super friendly ping.

clalancette:
I'll be honest; I haven't merged this because the name is somewhat counter to our naming [recommendations](http://wiki.ros.org/ROS/Patterns/Conventions).  It's pretty short, and includes "ros" in the package name, so it is not clear that this is unique enough in a global sense.  On the other hand, "rxros" all together is somewhat unique.

Maybe @tfoote has thoughts here?

gavanderhoorn:
The name RxROS is a combination (portmanteau almost?) on [RxCpp](https://github.com/ReactiveX/RxCpp) and ROS. It's a package that bridges the dataflow and reactive programming architectural patterns made possible by RxCpp into the ROS domain.

I'm not the author, but I believe the idea is that this name also reflects the use of the package: it's essentially the `roscpp` for people looking to apply reactive programming to their ROS application's needs.

Perhaps either @wasowski and/or @henrik7264 can add something here.

wasowski:
@gavanderhoorn got it all, there is not much to add. "Rx" is the short name used for reactive libraries across langauges (rxhaskell, rxpy, rxjava, rxcpp, etc.). I do think that this way rxros (reactive ros) is fairly consistent with what people call these things.

gavanderhoorn:
@tfoote @clalancette: could we discuss whether this (and #22249) could be merged?

We have some other work depending on this and RxROS is also the topic of our upcoming ROSCon presentation, so we'd like to be able to continue our preparations.

tfoote (approved):
Since it's standard to put the rx in front for the rx community I think that merging this as is makes sense. The use of ros in the name is a little redundant, but I don't think it will cause too much confusion.

PR #22252, opened 2019-09-16
catkin_virtualenv: 0.5.1-1 in 'melodic/distribution.yaml' [bloom]

paulbovbel:
Increasing version of package(s) in repository `catkin_virtualenv` to `0.5.1-1`:

- upstream repository: https://github.com/locusrobotics/catkin_virtualenv.git
- release repository: https://github.com/locusrobotics/catkin_virtualenv-release.git
- distro file: `melodic/distribution.yaml`
- bloom version: `0.8.0`
- previous version for package: `0.4.1-1`

## catkin_virtualenv

```
* catkin-pkg-modules has disappeared off pypi (#46 <https://github.com/locusrobotics/catkin_virtualenv/issues/46>)
  * catkin-pkg-modules has disappeared off pypi, but catkin-pkg is still there
  * Version all requirements
* Contributors: Paul Bovbel
```

PR #22262, opened 2019-09-16
Fix branches for catkin_virtualenv

paulbovbel:
Following from the errors in https://github.com/ros/rosdistro/pull/22252/files, these branches needed to be fixed.


clalancette:
Thanks for fixing this up.

PR #22329, opened 2019-09-21
Setup Melodic doc jobs for fanuc and fanuc_experimental.

gavanderhoorn:
As per subject.


tfoote:
I would also encourage you to add the src entries as well which will allow source based tools like rosinstall generator to integrate it. If you don't want the devel jobs yet you can disable them with 

`test_commits: false`

like here: https://github.com/ros/rosdistro/blob/34b0fd01366bf1f25fe9475003d4b1e4d8e4c148/kinetic/distribution.yaml#L941

gavanderhoorn:
See #22338.

PR #22333, opened 2019-09-23
Add rosdep rules for python3-bson

hidmic:
Added rules for:

- [Debian](https://packages.debian.org/search?keywords=python3-bson)
- [Fedora](https://pkgs.org/download/python3-bson)
- [Gentoo](https://packages.gentoo.org/packages/dev-python/pymongo)
- [OSX](https://pypi.org/project/bson/)
- [Ubuntu](https://packages.ubuntu.com/search?keywords=python3-bson)

to match `python-bson` platform coverage. For Gentoo and OSX, the keys are the same as IIUC the Python version being targeted depends on OS configuration and package manager respectively, but I don't have boxes at hand to try this out.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -696,6 +696,14 @@ python-bson:
     pip:
       packages: [bson]
   ubuntu: [python-bson]
+python3-bson:
+  debian: [python3-bson]
+  fedora: [python3-bson]
+  gentoo: [dev-python/pymongo]
+  osx:
+    pip:
+      packages: [bson]
+  ubuntu: [python3-bson]
 python-cairo:
   arch: [python2-cairo]
   debian: [python-cairo]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
The gentoo ones looks odd, but matches the existing rule so I'll assum it's fine. @allenh1 FYI

allenh1 (changes requested):
The Gentoo version is actually a custom ebuild we have in the overlay, `dev-python/bson`.

Added in ros/ros-overlay#574.

hidmic:
@allenh1 I see `dev-python/bson` made it here before and was shortly dropped in favor of `dev-python/pymongo` in https://github.com/ros/rosdistro/pull/17593, days after https://github.com/ros/ros-overlay/pull/574. I presume you mean that `dev-python/bson` is the right version for Python3+ installations. Can you confirm?

allenh1:
@hidmic I'm honestly not sure myself at this point... Let's leave it at `dev-python/pymongo` and just see if things break, I think. I'll definitely hear about it shortly if things aren't working. ;)

dirk-thomas (approved):

PR #22335, opened 2019-09-23
Update the FreeBSD dependencies for Rosdep.

NorwegianRockCat:
Since the pkg backend is  package system that is now used on FreeBSD
some port names have been changed. Since the rosdep and friends have
been added to the FreeBSD ports tree, it is a good thing to keep this
up to date. There may still be holes, but it is complete enough to
build the base comms and desktop releases of melodic.

Also document where FreeBSD packages can be found.

It took a while, but the FreeBSD project added the necessary package to its ports tree, so it is now possible to build ROS again on FreeBSD without lots of extra effort. Now that has happened it's possible to update the things here.


rosdep/base.yaml
```diff
@@ -3304,6 +3312,7 @@ libssl-dev:
   alpine: [libressl-dev]
   arch: [openssl]
   debian: [libssl-dev]
+  freebsd: [openssl]
```
tfoote: This is out of alphabetical order which is failing CI

tfoote (commented):
Thanks for taking the time to update the values. That will be helpful for FreeBSD users. There's one line that's out of order which is failing CI. If you could fix that I think it should be good to merge.

PS if you want to push things further on FreeBSD you might consider extending Superflore to generate the appropriate metadata for the FreeBSD pkg format for each ROS package and possibly building it into a pkg-repository. https://github.com/ros-infrastructure/superflore

rosdep/base.yaml
NorwegianRockCat: Doh! Pardon my messing up on the ordering.  I have corrected that in the next commit.

Also, thank you for the tip about Superflore! That will definitely get a more complete package listing than my ad-hoc method thus far. I'll take a look.

I presented ROS at a BSD conference over the weekend and it has generated some enthusiasm for getting things into the FreeBSD ports tree. So, it should lead to others being able to use FreeBSD and ROS without extensive hacking.

tfoote (approved):

rosdep/python.yaml
```diff
@@ -1628,6 +1629,7 @@ python-gnupg:
   arch: [python2-gnupg]
   debian: [python-gnupg]
   fedora: [python-gnupg]
+  freebsd: [py27-gnupg]
```
sloretz: I didn't find a package with this name. Did you mean [py27-python-gnupg](https://www.freshports.org/security/py-python-gnupg/)?

NorwegianRockCat: Yes, I think I did. It's strange because I know I built melodic-desktop in a jail, which meant that rosdep would have called pkg and pkg would fail if it gets a bad package name. I'll investigate and let you know what I find (with a possible change).

NorwegianRockCat: OK. I did a bit more digging. The origin of the py27-gnupg pkg is [security/py-gnupg](https://www.freshports.org/security/py-gnupg/) on FreeBSD. It seems to be a fork of the python-gnupg with an aim on security. It works, but you need to change the python import statement.

`>>> from pretty_bad_protocol import gnupg`

I've changed python.yaml to use py27-python-gnupg as it doesn't require changing other files in ROS.

Thanks for checking this out.

sloretz (approved):

PR #22337, opened 2019-09-23
fanuc: disable Kinetic experimental repository devel jobs

gavanderhoorn:
As per subject.


gavanderhoorn:
Thanks for the suggestion @tfoote.

PR #22345, opened 2019-09-24
toposens: 1.2.3-2 in 'kinetic/distribution.yaml' [bloom]

sebdengler:
Increasing version of package(s) in repository `toposens` to `1.2.3-2`:

- upstream repository: https://gitlab.com/toposens/public/ros-packages.git
- release repository: https://gitlab.com/toposens/public/toposens-release.git
- bloom version: `0.8.0`
- previous version for package: `1.2.3-1`

## toposens

- No changes

## toposens_description

- No changes

## toposens_driver

- No changes

## toposens_markers

- No changes

## toposens_msgs

- No changes

## toposens_pointcloud

```
* Outsourced saving of cumulative pointcloud into boost::thread
* Contributors: Christopher Lang
```

## toposens_sync

- No changes

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -14980,7 +14980,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://gitlab.com/toposens/public/toposens-release.git
-      version: 1.2.3-1
+      version: 1.2.3-2
     source:
       type: git
       url: https://gitlab.com/toposens/public/ros-packages.git


kinetic/distribution.yaml
```diff
@@ -14980,7 +14980,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://gitlab.com/toposens/public/toposens-release.git
-      version: 1.2.3-1
+      version: 1.2.3-2
```
clalancette: It's unusual to make changes to the source but only bump the debinc.  Did you mean to do a release of the source first (`catkin_generate_changelog`/`catkin_prepare_release`)?

clalancette (changes requested):

kinetic/distribution.yaml
sebdengler: I did not want to do a release of the changes in source for now. I was trying to make the Jenkins jobs run again for the same version of the code. When we did the previous release, the "doc" Jenkins job failed for no obvious reason. I was hoping that this was just a problem with the Jenkins pipeline that could be solved by running it again.

Do you have any advice how I could proceed in this situation?

clalancette: The reason that I flagged this is that the changelog above does show a source change, so basically the built package would no longer match the sources.  In this case, I'm going to close this out, and re-kick http://build.ros.org/job/Kdoc__toposens__ubuntu_xenial_amd64 for you manually.

sebdengler: It worked. Thank you very much!

PR #22347, opened 2019-09-24
Adding niryo_one to documentation index for distro

tinoreycoc:
I'd like to add my repo niryo_one to be indexed and documented on ros.org
niryo_one is a robotic arm made by my company, Niryo.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4766,6 +4766,11 @@ repositories:
       url: https://github.com/astuff/network_interface.git
       version: release
     status: developed
+  niryo_one:
+    doc:
+      type: git
+      url: https://github.com/NiryoRobotics/niryo_one_ros.git
+      version: master
   nmea_comms:
     doc:
       type: git


clalancette (changes requested):
Need to have the license in the source repository clarified.

tinoreycoc:
OK. I have add the license on the top of the main page and there is a license file in the repository. 
Is it OK ? 

Thanks

clalancette:
So the other thing I noticed in your source repository is that you have forked off the `dynamixel_sdk` package and the MCP-CAN libraries.  This is going to be problematic because you will never be able to release this package successfully; we don't allow duplicate entries of packages, and `dynamixel_sdk` is already released into Melodic.  While MCP-CAN is not currently released into Melodic, a similar thinking there applies; we don't want to release the package from here since we may want to release the original sometime in the future.

Thus my suggestion here is to remove both of those forks from your repository.  I'd try to make your code work with the respective upstreams of those packages.  If it turns out that you need to make changes to support your use case, you can submit patches to those upstream repositories and make the fixes you need.

tinoreycoc:
Hello Chris,

Thank you for your answer.
As there is a lot of things to change, we decide to create another github
repository for our robot simulation only.
So I'm going to change github url from the fork and commit again.
Please let me know if it works for you.

Best regards,
Etienne

clalancette:
Sounds good to me.  I'm going to close this out for now, please open a new pull request when you are ready.

PR #22350, opened 2019-09-24
Add in rosdep key for spdlog.

clalancette:

* Debian: https://packages.debian.org/buster/libspdlog-dev
* Fedora: https://apps.fedoraproject.org/packages/spdlog-devel
* Gentoo: https://packages.gentoo.org/packages/dev-libs/spdlog
* macOS: https://formulae.brew.sh/formula/spdlog
* Ubuntu: https://packages.ubuntu.com/bionic/libspdlog-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4965,6 +4965,11 @@ sparsehash:
     xenial: [libsparsehash-dev]
     yakkety: [libsparsehash-dev]
     zesty: [libsparsehash-dev]
+spdlog:
+  debian: [libspdlog-dev]
+  fedora: [spdlog-devel]
+  gentoo: [dev-libs/spdlog]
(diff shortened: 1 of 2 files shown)


mjcarroll (approved):

nuclearsandwich:
Per our own guidelines ought this rosdep key have been `libspdlog-dev`? @tfoote is it worth changing now before it gets established?

Also I just realized that preferring the key name match the Debian/Ubuntu package name isn't actually written in our contribution guidelines

clalancette:
> Per our own guidelines ought this rosdep key have been `libspdlog-dev`? @tfoote is it worth changing now before it gets established?

I actually didn't know that that was a guideline :).  Personally I'll say that there are enough exceptions that it isn't clear to me that one was preferred over the other, so I'd be inclined to stick with it as-is.

PR #22426, opened 2019-09-27
Adding niryo_one_ros_simulation

tinoreycoc:
Hi,

I'd like niryo_one_ros_simulation to be indexed and documented on ros.org. 
Niryo one is an educational robotic arm, made by my startup Niryo. 
The stack works for both Kinematic and Melodic distro. 

By thanking you in advance,
Best regards,

Etienne

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -8120,6 +8120,11 @@ repositories:
       url: https://github.com/astuff/network_interface.git
       version: release
     status: maintained
+  niryo_one_simulation:
+    doc:
+      type: git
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):
The source repository looks generally fine to me.  However, you probably want to split this up into two pull requests.  Kinetic and Melodic are on different sync schedules, so it can happen (like right now) that one of them is in a freeze while the other isn't.  So we can't merge this at present.  I'll close this for now; please open two separate ones, one for Kinetic and one for Melodic.  Thanks.

PR #22462, opened 2019-09-30
add rqt for Noetic

dirk-thomas:


sloretz:
@dirk-thomas it looks like this one depends on `actionlib` (@mjcarroll fyi), which depends on `common_msgs` (@tfoote fyi) https://github.com/ros-visualization/rqt/blob/5549d88d28ae1f0de51823c931aad7265ce1ebe7/rqt_py_common/package.xml#L37 .

@tfoote Is `common_msgs` compatible with Python 3? If so, mind adding a source entry for it?

PR #22480, opened 2019-10-01
jderobot-colortuner: 0.0.3-2 in 'kinetic/distribution.yaml' [bloom]

fqez:
Increasing version of package(s) in repository `jderobot-colortuner` to `0.0.3-2`:

- upstream repository: https://github.com/JdeRobot/ColorTuner.git
- release repository: https://github.com/JdeRobot/ColorTuner-release.git
- bloom version: `0.8.0`

## jderobot_color_tuner

- No changes

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5364,7 +5364,7 @@ repositories:
       url: https://github.com/gstavrinos/jaguar-release.git
       version: 0.1.0-0
     status: developed
-  jderobot-colortuner:
+  jderobot_colortuner:
     release:
       packages:
       - jderobot_color_tuner
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5364,6 +5364,14 @@ repositories:
       url: https://github.com/gstavrinos/jaguar-release.git
       version: 0.1.0-0
     status: developed
+  jderobot-colortuner:
+    release:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -5364,6 +5364,14 @@ repositories:
       url: https://github.com/gstavrinos/jaguar-release.git
       version: 0.1.0-0
     status: developed
+  jderobot-colortuner:
```
clalancette: All packages should use `_`, rather than `-` in the name.

clalancette (changes requested):

clalancette:
This is failing CI because it is not in alphabetical order; this should be after `jderobot_assets`.

While you are in there, I noticed that the name of the package inside here is `jderobot_color_tuner`, while the top-level name you've selected is `jderobot_colortuner`.  While this is allowed, it seems to make more sense to have those be the same.

tfoote:
@fqez I saw your reference from the upstream source. You only need to add an underscore and edit this file to insert it in the right alphabetical order to resolve the issues raised above. 

If you're not going to do that soon we'll close this out due to inactivity soon.

clalancette:
No activity in over a month, so I'll close this out.  If you still want to get this in, please respond to the review comments and re-open.  Thanks.

PR #22524, opened 2019-10-03
Update ainstein_radar for kinetic

nrotella-ainstein:
Manually opening pull request following failure of bloom-release step with the following errors in logfile:

SSLError: ('The read operation timed out',)

[error] Failed to open pull request: SSLError - ('The read operation timed out',)

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -203,7 +203,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/AinsteinAI/ainstein_radar-release.git
-      version: 1.0.2-1
+      version: 1.0.3-1
     source:
       type: git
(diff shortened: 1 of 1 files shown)


tfoote:
For future reference you can ask bloom to retry just opening the PR with the `-p` option. We highly recommend that so that all the meta data gets recorded here, not just "Update distirbution.yaml"

nrotella-ainstein:
> For future reference you can ask bloom to retry just opening the PR with the `-p` option. We highly recommend that so that all the meta data gets recorded here, not just "Update distirbution.yaml"

Thank you for the tip, I was unaware of that option. It would be helpful to add something about this in the tutorial http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease#Notifying_the_Build_Farm

PR #22535, opened 2019-10-03
Add merge-args from pypi

rotu:


tfoote (changes requested):
Please read the CONTRIBUTING guide about rosdep rules. 

Please link to the package and describe how it will be used. 

Also please don't fan out to all OSs that you're not familiar with or expecting to use. This is not appropriate or available for all platforms. 

The name should have a `-pip` extension due to being pip only. 

I found https://pypi.org/project/merge-args/ which is worrysome as the primary description includes "This small library uses Advanced Python Hackery" as it's opening statement and appears to be a proof of concept not an actively supported library. And although we generally don't review the contents of packages, I would not recommend using this as a dependency. It is not available from any native package manger and I don't think that it will ever be packaged that way. Thus it can never be used in a released packages. 

It's a neat trick, but because it's a trick it actually makes it much harder to understand the code when reading it especially when the doc string no longer represents the prototype for the function as it reuses the doc string from the wrapped function which has different arguments.

rotu:
:-( This is too much work for a simple dependency. Closing this PR in frustration and reopening later if I feel like jumping thhrough these hoops.

PR #22566, opened 2019-10-04
Align eloquent with dashing pr testing

mjcarroll:
Here is the audit of the current differences:

```
Repo                           D    , E    
cartographer                   False, True
console_bridge_vendor          False, True
examples                       True, False
laser_geometry                 False, True
navigation_msgs                False, True
osrf_pycommon                  False, True
ros1_bridge                    False, True
ros_environment                False, True
rosidl_typesupport_connext     False, True
rosidl_typesupport_opensplice  False, True
test_interface_files           False, True
tinyxml2_vendor                False, True
tinyxml_vendor                 False, True
uncrustify_vendor              False, True
urdfdom_headers                False, True
```

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -203,7 +203,6 @@ repositories:
       url: https://github.com/ros2-gbp/cartographer-release.git
       version: 1.0.0-1
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros2/cartographer.git
       version: ros2
@@ -279,7 +278,6 @@ repositories:
       url: https://github.com/ros2-gbp/console_bridge_vendor-release.git
       version: 1.2.0-1
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros2/console_bridge_vendor.git
       version: eloquent
@@ -574,6 +572,7 @@ repositories:
       url: https://github.com/ros2-gbp/examples-release.git
       version: 0.8.2-1
     source:
+      test_pull_requests: true
       type: git
       url: https://github.com/ros2/examples.git
       version: eloquent
@@ -853,7 +852,6 @@ repositories:
       url: https://github.com/ros2-gbp/laser_geometry-release.git
       version: 2.1.0-1
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros-perception/laser_geometry.git
       version: eloquent
@@ -1003,7 +1001,6 @@ repositories:
       url: https://github.com/ros2-gbp/navigation_msgs-release.git
       version: 2.0.2-1
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros-planning/navigation_msgs.git
       version: eloquent
@@ -1087,7 +1084,6 @@ repositories:
       url: https://github.com/ros2-gbp/osrf_pycommon-release.git
       version: 0.1.9-2
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/osrf/osrf_pycommon.git
       version: eloquent
@@ -1699,7 +1695,6 @@ repositories:
       url: https://github.com/ros2-gbp/ros1_bridge-release.git
       version: 0.8.1-4
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros2/ros1_bridge.git
       version: eloquent
@@ -1832,7 +1827,6 @@ repositories:
       url: https://github.com/ros2-gbp/ros_environment-release.git
       version: 2.4.1-1
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros/ros_environment.git
       version: eloquent
@@ -2047,7 +2041,6 @@ repositories:
       url: https://github.com/ros2-gbp/rosidl_typesupport_connext-release.git
       version: 0.8.4-1
     source:
-      test_pull_requests: true
       type: git
       url: https://github.com/ros2/rosidl_typesupport_connext.git
       version: eloquent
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
The *_vendor packages were turned off at one point because they were either deemed "trivial" or didn't install correctly enough to work in a devel job. But they're all passing in Eloquent so I would elect to keep them.

urdfdom_headers needs to be disabled because it is a third-party package without a package.xml.

The Connext jobs need to be turned off because our CI has no way to source the custom setup file from the vendor RTI package. We have patches in the debian/rules file which source it for the binaries.

The ros1_bridge can't be tested because we don't have a mechanism for adding the ROS 1 repositories to these jobs.

For everything else, if it is currently building but unstable, I think it's worth examining whether a stable build is possible and get it there rather than disabling them.

dirk-thomas:
What @nuclearsandwich said :+1:

dirk-thomas:
@mjcarroll Friendly ping.

dirk-thomas:
There are also several devel / PR jobs which are constantly failing (and probably can't be changed to work?). If that is the case they should be disabled, too.

dirk-thomas:
Please re-sync the changes based on the current state of the dev jobs.

eloquent/distribution.yaml
```diff
@@ -1307,7 +1307,6 @@ repositories:
       url: https://github.com/ros2-gbp/ros_environment-release.git
       version: 2.4.0-1
     source:
-      test_pull_requests: true
```
dirk-thomas: This seems to pass: http://build.ros2.org/view/Edev/job/Edev__ros_environment__ubuntu_bionic_amd64/6/

clalancette:
@mjcarroll What do you want to do with this PR?

dirk-thomas:
> Please re-sync the changes based on the current state of the dev jobs.

@mjcarroll Another friendly ping.

tfoote:
We're now well into this release. If this should be revived that's good but otherwise this is now out of date and will need to be reproduced if desired due to conflicts. I'm going to close this PR we can open a new one if desired.

dirk-thomas:
I will reopen this ticket since otherwise it gets lost. Also assigning it to @mjcarroll.

dirk-thomas:
@mjcarroll Friendly ping.

dirk-thomas:
@mjcarroll Friendly ping.

mjcarroll:
Apoolgies, this was ready for review.

eloquent/distribution.yaml
```diff
@@ -635,6 +636,8 @@ repositories:
       url: https://github.com/ros2-gbp/foonathan_memory_vendor-release.git
       version: 0.3.0-1
     source:
+      test_commits: false
+      test_pull_requests: true
```
dirk-thomas: Why this combination - testing PRs but not actually the target branch? Also Dashing doesn't have `test_commits: false`.

eloquent/distribution.yaml
```diff
@@ -908,6 +911,7 @@ repositories:
       version: 1.0.0-1
     source:
       test_abi: true
+      test_pull_requests: true
```
dirk-thomas: This isn't set for Dashing. Why should it be for Eloquent?

eloquent/distribution.yaml
```diff
@@ -1594,7 +1598,6 @@ repositories:
       url: https://github.com/ros2-gbp/rmw_connext-release.git
       version: 0.8.1-1
     source:
-      test_pull_requests: true
```
dirk-thomas: Dashing has this set?

eloquent/distribution.yaml
```diff
@@ -1699,7 +1702,6 @@ repositories:
       url: https://github.com/ros2-gbp/ros1_bridge-release.git
       version: 0.8.1-4
     source:
-      test_pull_requests: true
```
dirk-thomas: Dashing has `test_commits: false`?

eloquent/distribution.yaml
```diff
@@ -2697,7 +2698,7 @@ repositories:
       url: https://github.com/ros2-gbp/urdfdom_headers-release.git
       version: 1.0.4-1
     source:
-      test_pull_requests: true
+      test_pull_requests: false
```
dirk-thomas: This is the default so the line can be removed.

eloquent/distribution.yaml
```diff
@@ -2780,6 +2781,7 @@ repositories:
       url: https://github.com/ros2-gbp/yaml_cpp_vendor-release.git
       version: 7.0.0-1
     source:
+      test_pull_requests: true
```
dirk-thomas: Not set in Dashing?

eloquent/distribution.yaml
mjcarroll: This was turned on based on @nuclearsandwich 's comment above:

> The *_vendor packages were turned off at one point because they were either deemed "trivial" or didn't install correctly enough to work in a devel job. But they're all passing in Eloquent so I would elect to keep them.

eloquent/distribution.yaml
mjcarroll: Same reason as above

eloquent/distribution.yaml
mjcarroll: Removed.

eloquent/distribution.yaml
mjcarroll: Added.

eloquent/distribution.yaml
mjcarroll: I set this because it was passing, but it's not actually doing anything due to lack of ros1.

Removed.

eloquent/distribution.yaml
mjcarroll: Removed.

dirk-thomas (approved):

PR #22579, opened 2019-10-06
add pymongo to the depends

hakuturu583:
add Pymongo to the depends.(https://pypi.org/project/pymongo/)


tfoote (changes requested):
I found Ubuntu packages in a quick search. https://packages.ubuntu.com/xenial/python-pymongo

Our policy is to use the native package manager packages if they exist. To that end please check the other platforms for sources and update this with new rules and links in the description.

PR #22606, opened 2019-10-09
Adding new repository for indexing: prophesee_ros_wrapper

nlyubova:


melodic/distribution.yaml
```diff
@@ -6011,6 +6011,24 @@ repositories:
...
+      version: master
+    release:
+      packages:
+      - prophesee_event_msgs
+      - prophesee_ros_driver
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/prophesee-ai/prophesee_ros_wrapper-release.git
+      version: 0.0.1-1
+    source:
```
gavanderhoorn: Hi. Great to see ROS support for this sensor, but I'm curious how this can be released through the ROS buildfarm, as IIUC, building `prophesee_ros_driver` requires the SDK to be installed. The buildfarm doesn't do that, and [the `CMakeLists.txt` also doesn't appear to install it](https://github.com/prophesee-ai/prophesee_ros_wrapper/blob/2bc3c76e8123dbd7fa935f42bfc530ee4d7182ff/prophesee_ros_driver/CMakeLists.txt).

nlyubova: Hi, right, it requires SDK to be installed. In cmakelists, there is  "find_package(prophesee_driver REQUIRED)"

is it still possible to release it in this case? or can I adapt it somehow to have it releasable?

sloretz: > find_package(prophesee_driver REQUIRED)

While the `CMakeLists.txt` requires that `prophesee_dirver` is installed, that doesn't tell the buildfarm how to install it. The buildfarm has to be able to build your package so it can make a binary `deb` that others can install.

> is it still possible to release it in this case? or how can I adapt it to make it releasable?

Yup. Like @gavanderhoorn mentioned, you could make a ROS package that downloads and installs the SDK. Alternatively you could add a rosdep rule for the SDK.

## Rosdep key

Is `prophesee_driver` released on all of this ROS distro's target platforms? If so, this is the easiest. [Add a rosdep key](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions) for `prophesee_driver` and then declare it as a dependency in the `prophesee_ros_driver`'s `package.xml`.

Example:

[Melodic requires support for Ubuntu Artful and Bionic](https://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023), but [`Artful` is EOL](https://fridge.ubuntu.com/2018/07/19/ubuntu-17-10-artful-aardvark-end-of-life-reached-on-july-19-2018/) so that can be ignored. The [`libmodbus-dev` rosdep key](https://github.com/ros/rosdistro/blob/9861e8279e2931aed0459c01cac7c55da0f665ae/rosdep/base.yaml#L2421-L2424) tells the buildfarm to install the debian package `libmodbus-dev` which [is available in Bionic](https://packages.ubuntu.com/bionic/libmodbus-dev). If your package were to [depend on `libmodbus-dev`](https://github.com/PilzDE/pilz_robots/blob/32a380f43b34d3fefbe767fe706e1878a4faf773/prbt_hardware_support/package.xml#L22), then the buildfarm would read the `package.xml` see the dependency, and install `libmodbus-dev` before building your package.

## ROS package that installs the SDK

In this option, you provide another ROS package that just downloads and installs the SDK, usually using [`ExternalProject_Add`](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html). Then make `prophesee_ros_driver` depend on that package in it's `package.xml`.

Example: See [`phidgets_drivers`](https://github.com/ros-drivers/phidgets_drivers) which has a [ROS package `libphidget21`](https://github.com/ros-drivers/phidgets_drivers/tree/melodic/libphidget21) that [downloads and installs the phidgets SDK](https://github.com/ros-drivers/phidgets_drivers/blob/1e97a67bd4cf516be61d770fad6f4286d68655e3/libphidget21/CMakeLists.txt#L9-L40).

clalancette: For what it is worth, having the prophesee SDK behind a wall/webform makes it much less likely that it is possible to package and release this into the ROS buildfarm.  If the code for the prophesee driver is open source, then I'd highly suggest putting it in a publically-accessible location, and then using the second method that @sloretz pointed out (using `ExternalProject_add`).

nlyubova: Thanks a lot for your feedback ! 
I'll check it and get back with my PR

PR #22638, opened 2019-10-11
Add python3-networkmanager

cevans87:
Add python3-networkmanager for all recent ubuntu distributions. Keep python-networkmanager as is.

This is necessary to manipulate the NetworkManager service from ROS2 python nodes.

Package details: https://packages.ubuntu.com/bionic/python3-networkmanager

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2427,7 +2427,14 @@ python-networkmanager:
     buster: [python-networkmanager]
     jessie: [python-networkmanager]
     stretch: [python-networkmanager]
-  ubuntu: [python-networkmanager]
+  ubuntu: 
+    '*': [python-networkmanager]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
The `_python3` rosdep format is a remnant from an earlier attempt to share keys between python2 and python3 dependencies. Please add a new python3-networkmanager rodsep key instead. Then the ros2 branch of your package can use the python3 key and the ROS 1 branch can use a conditional dependency to support either python2 or python3 based on which is used by that distribution.

clalancette:
It's been over a month with no response from the contributor, so I'll close this out for now.  Please feel free to fix the comments from @nuclearsandwich and re-open when you have time.  Thanks for the contribution.

PR #22713, opened 2019-10-19
px4_msgs: 2.0.0-1 in 'crystal/distribution.yaml' [bloom]

TSC21:
Increasing version of package(s) in repository `px4_msgs` to `2.0.0-1`:

- upstream repository: https://github.com/PX4/px4_msgs.git
- release repository: https://github.com/PX4/px4_msgs2-release.git
- distro file: `crystal/distribution.yaml`
- bloom version: `0.8.0`

## px4_msgs

```
* First release of px4_msgs to ROS 2 distros
* Contributors: Nuno Marques, PX4 Build Bot, PX4BuildBot, TSC21
```

### DIFF
--- a/crystal/distribution.yaml
+++ b/crystal/distribution.yaml
@@ -1058,6 +1058,13 @@ repositories:
       url: https://github.com/ros2/poco_vendor.git
       version: crystal
     status: developed
+  px4_msgs:
+    release:
+      tags:
+        release: release/crystal/{package}/{version}
+      url: https://github.com/PX4/px4_msgs2-release.git
+      version: 2.0.0-1
+    status: developed
   py_trees:
     doc:
       type: git


TSC21:
@nuclearsandwich I got some emails stating the builds have failed. Can you help me track the reason? The machines were the build is processed have the ROS_DISTRO env variable set somehow?

nuclearsandwich:
That variable is set by the ros_environment package, if you're going to rely on it in your build you should depend on that package.

TSC21:
> That variable is set by the ros_environment package, if you're going to rely on it in your build you should depend on that package.

So if I added it to package.xml, will the build process then include it as a dependency and automatically set ROS_DISTRO, or are there other things I should be aware of? I want to make sure that I can release this properly.

nuclearsandwich:
> So if I added it to package.xml, will the build process then include it as a dependency and automatically set ROS_DISTRO, or are there other things I should be aware of? I want to make sure that I can release this properly.

I'm not the most familiar with the ros_environment package. If it provided a native CMake variable you'd also need to include it in your cmake logic but since it does appear to just set the environment variables, adding it as a build dependency to your package.xml should be sufficient as the root workspace environment is sourced by the packaging build scripts.

tfoote:
Follow up with that at https://github.com/PX4/px4_msgs/pull/1

TSC21:
> Follow up with that at [PX4/px4_msgs#1](https://github.com/PX4/px4_msgs/pull/1)

It did the trick @tfoote! Thank you very much!

PR #22727, opened 2019-10-22
Debian stretch kernel key is now version 11, so update it here.

clalancette:

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3900,7 +3900,7 @@ linux-headers-generic:
   arch: [linux-headers]
   debian:
     jessie: [linux-headers-3.16.0-4-all]
-    stretch: [linux-headers-4.9.0-3-all]
+    stretch: [linux-headers-4.9.0-11-all]
     wheezy: [linux-headers-3.2.0-4-all]
   fedora: [kernel-headers, kernel-devel]
   gentoo: [sys-kernel/linux-headers]


nuclearsandwich (approved):

nuclearsandwich:
Is there no metapackage that will be evergreen here/

clalancette:
> Is there no metapackage that will be evergreen here?

As far as I can tell, no :(.  I looked briefly, and the options seem to be:

```
linux-headers-4.9.0-11-all - All header files for Linux 4.9 (meta-package)
linux-headers-4.9.0-11-all-amd64 - All header files for Linux 4.9 (meta-package)
linux-headers-4.9.0-11-amd64 - Header files for Linux 4.9.0-11-amd64
linux-headers-4.9.0-11-common - Common header files for Linux 4.9.0-11
linux-headers-4.9.0-11-common-rt - Common header files for Linux 4.9.0-11-rt
linux-headers-4.9.0-11-rt-amd64 - Header files for Linux 4.9.0-11-rt-amd64
linux-headers-amd64 - Header files for Linux amd64 configuration (meta-package)
linux-headers-rt-amd64 - Header files for Linux rt-amd64 configuration (meta-package)
```

Anything that ends with `-amd64` is out, since that won't work across platforms.  That leaves us with `linux-headers-4.9.0-11-all`, `linux-headers-4.9.0-11-common`, and `linux-headers-4.9.0-11-common-rt`, and the latter two are still kernel specific but are a subset of the `-all` one.  If someone else knows of a different metapackage that is available that we could use here, I'd be all ears.

claireyywang (approved):

PR #22736, opened 2019-10-23
Adding mongodb_store to documentation index for distro

hawesie:
The doc tag was missing for older releases, and since this is the first page on the wiki it's a bit annoying!


clalancette:
Indigo is no longer a supported distribution, so we don't make any further changes.  Feel free to do this same update on the supported distributions (kinetic and melodic as of right now).  Thanks for the contribution.  Closing.

PR #22797, opened 2019-10-24
eigenpy: 1.6.2-1 in 'melodic/distribution.yaml' [bloom]

wxmerkt:
### Important notes

This changes the package from a catkin wrapper to a proper third-party released library per the REP. This requires changing dependent projects from `find_package(catkin REQUIRED COMPONENTS eigenpy)` to use `find_package(PkgConfig) ; pkg_check_modules(eigenpy REQUIRED eigenpy)`. 

I have opened such a PR for MoveIt and Exotica; the two projects currently depending on eigenpy. As thus, we need to coordinate a MoveIt release after this eigenpy release in the same sync cycle (cc: @v4hn).

### Regular bloom notes

Increasing version of package(s) in repository `eigenpy` to `1.6.2-1`:

- upstream repository: https://github.com/stack-of-tasks/eigenpy.git
- release repository: https://github.com/ipab-slmc/eigenpy_catkin-release.git
- bloom version: `0.9.0`
- previous version for package: `1.5.1-2`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -1754,7 +1754,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/ipab-slmc/eigenpy_catkin-release.git
-      version: 1.5.1-2
+      version: 1.6.2-1
     source:
       type: git
       url: https://github.com/stack-of-tasks/eigenpy.git


v4hn:
> I have opened such a PR for MoveIt and Exotica; the two projects currently depending on eigenpy. As thus, we need to coordinate a MoveIt release after this eigenpy release in the same sync cycle (cc: @v4hn).

The patch for MoveIt triggers two issues in the infrastructure (catkin_lint and catkin_tools) that have to be resolved (or at least workarounded for us) to merge the request.
@rhaschke is the main person releasing the moveit repositories at the moment, and he mentioned earlier this week that he doesn't have time to work on releases at the moment.

wxmerkt:
@v4hn 

- catkin_lint upstream acknowledged that the issue is with the change from wrapper to third-party library. This issue cannot be easily worked around and only gets resolved by merging this new release.
- The catkin_tools issue is real - the third-party library does not have make targets. You can add ``--continue-on-failure`` to the catkin build and it will work around. Is that an acceptable fix?
- Alternatively, can we deactivate eigenpy inside MoveIt?

rhaschke:
If I understand it correctly, the catkin_lint issue will be resolved automagically if this PR is merged (https://github.com/fkie/catkin_lint/issues/71).

I think the catkin_tools issue is only an issue for MoveIt's Travis config, which (temporarily in https://github.com/ros-planning/moveit/pull/1550) includes the eigenpy repo into the catkin workspace. As we will usually not build eigenpy with catkin_tools in our CI workspace, that is not a real issue for MoveIt as well.

However, as long as https://github.com/ros-planning/moveit/pull/1550 is not merged and released together with eigenpy, the ROS build farm will fail building MoveIt.

To decouple the releases (which I indeed cannot prepare for MoveIt until mid of November after IROS), I suggest for eigenpy to provide both, cmake and pkgconfig files, which allows a downstream package like MoveIt to use either method to find the eigenpy upstream dependency. Thus we would be free to merge https://github.com/ros-planning/moveit/pull/1550 at any time after the eigenpy release.

wxmerkt:
We can provide a CMake config that exports a target. So you would have to link to `eigenpy::eigenpyTargets`. This is not the same as via the catkin config currently provided, I believe?

rhaschke:
No, my idea would be to provide the cmake file of a proper catkin package, such that the current use via `find_package(catkin COMPONENTS eigenpy)` will work.

wxmerkt:
Hi @rhaschke,
Perhaps we can meet in Macau in person to discuss and coordinate/push the releases to unblock this. We'd like to get this merged and unblocked. If you've got examples of catkin-style CMake configs not generated by catkin itself, happy to use such a hotfix as well.

Cheers,
Wolfgang

rhaschke:
I am happy to meet you in Macao. What about meeting around 10:30 (the ROSCon coffee break) close to the Microsoft booth (which was a rather empty spot today).

wxmerkt:
@clancette: I think it does not require further changes - but a coordinated release with MoveIt

PR #22821, opened 2019-10-28
Add python-multiprocess

rickstaa:
Improved version of the multiprocessing library.
- https://pypi.org/project/multiprocess/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2324,6 +2324,22 @@ python-multicast:
     pip: [py-multicast]
   ubuntu:
     pip: [py-multicast]
+python-multiprocess:
+  arch:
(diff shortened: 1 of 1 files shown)


cottsay (changes requested):
> I think this key ought to have a -pip suffix

[Agreed.](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules)

rickstaa:
@nuclearsandwich @cottsay Your right I'm sorry. I just added the -pip suffix.

cottsay (approved):
Looks like this package isn't present in Ubuntu or Fedora, so a pip package seems appropriate.

PR #22871, opened 2019-11-02
Add multiple packages to documentation index for indigo and kinetic

Tobi2001:
Adding new packages to indigo and kinetic documentation index:
- asr_ism
- asr_ism_visualizations
- asr_lib_ism
- asr_lib_pose_prediction_ism
- asr_psm
- asr_psm_visualizations
- asr_relation_graph_generator
- asr_resources_for_psm

PR #22872, opened 2019-11-02
Adding asr packages to documentation index of distro.

FelixMarek:
I would like to add asr packages to be indexed and documented on ros.org

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -480,6 +480,11 @@ repositories:
       type: git
       url: https://github.com/asr-ros/asr_descriptor_surface_based_recognition.git
       version: master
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):
Indigo is no longer a supported platform, so we aren't accepting new changes for it.  Please remove these changes from this PR.  That will probably also fix failing CI.

clalancette:
Also, I'm fairly sure this is going to conflict with #22871 .  Once that lands, you'll need to rebase this.

tfoote:
It has landed and this is also now conflicting and unmergable. Please take a moment to remove the indigo elements and rebase it based off of the latest version.

PR #22916, opened 2019-11-08
Need remove unusable package

harjeb:
Hi, I want to remove this unusable package and release new one, Thanks


tfoote:
Ok, but in general you don't need to remove a package to be able to release a new version.

PR #22938, opened 2019-11-09
Add librealsense and librealsense2

Levi-Armstrong:


melodic/distribution.yaml
```diff
@@ -3806,6 +3806,26 @@ repositories:
...
+      version: legacy
+    source:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: legacy
+    status: maintained
+  librealsense2:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/realsense-ros.git
```
clalancette: This repository isn't the same one as the one released as `librealsense2` in Kinetic: https://github.com/ros/rosdistro/blob/master/kinetic/distribution.yaml#L6354 .  What's going on with this?

clalancette (changes requested):

melodic/distribution.yaml
Levi-Armstrong: In kinetic it was point to  @doronhi personal fork to the realsense repository. Now Intel has its own repository for 2.0 SDK which the same individual has recently contributed to and the one in kinetic has not been touch in a year. Maybe @doronhi can shed some light on  what is the correct github repository to use for librealsense2?

doronhi: I am sorry for the neglected issue of adding librealsense2 as a ROS debian package.
The 2.25.0 version was taken, as you mention, from my own fork, which is a bad practice.
I hit some obstacles on trying to use the official version but mainly it's a matter of available time that I lack. Hope to get around to it as soon as possible.

Levi-Armstrong: @doronhi  Is the repository listed below the correct one to use if building from source for ros development?

Repo: https://github.com/IntelRealSense/realsense-ros.git

doronhi: yes

melodic/distribution.yaml
```diff
@@ -3806,6 +3806,26 @@ repositories:
...
+      version: development
+    source:
+      type: git
+      url: https://github.com/IntelRealSense/realsense-ros.git
+      version: 2.2.9
```
clalancette: This is failing CI because the source entry can't point to a tag, it must point to branch.

melodic/distribution.yaml
clalancette: All right.  In that case, I'm OK to take this.  @doronhi , please update the Kinetic version of this to point to the `IntelRealSense` organization as early as you can so we can remove this discrepancy.

clalancette (changes requested):

melodic/distribution.yaml
clalancette: This was done, thanks @doronhi

melodic/distribution.yaml
clalancette: This is still failing CI because of this problem.

Levi-Armstrong: I just changed it to point to the development branch.

melodic/distribution.yaml
```diff
@@ -3806,6 +3806,26 @@ repositories:
...
+    status: maintained
+  librealsense2:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/realsense-ros.git
+      version: development
+    source:
+      type: git
+      url: https://github.com/IntelRealSense/realsense-ros.git
+      version: development
```
clalancette: CI is upset at the trailing whitespace on this line.  Once that is fixed, I think CI will be happier.

melodic/distribution.yaml
```diff
@@ -3806,6 +3806,26 @@ repositories:
...
+      version: legacy
+    status: maintained
+  librealsense2:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/realsense-ros.git
+      version: development
+    source:
+      type: git
+      url: https://github.com/IntelRealSense/realsense-ros.git
```
tylerjw: This is confusing, I'm not sure you are doing what you intend to here.  What you are calling `librealsense2` here is being linked to realsense-ros which contains two ros packages and one of them depends on librealsense2 which this package doesn't supply.  Instead, what you are calling `librealsense` above is the git repo that supplies the `librealsense2` package.

melodic/distribution.yaml
```diff
@@ -3806,6 +3806,26 @@ repositories:
...
+  librealsense:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: legacy
+    source:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: legacy
+    status: maintained
```
tfoote: Is the legacy branch maintained?

tfoote (changes requested):
@doronhi could you comment, does this make sense?

@Levi-Armstrong could you provide some explanation of what your intent is here I agree that this is a little ambiguous. Your PR has no description.

tfoote:
Oh, I just found the thread in the "resolved" comments. It looks like we did already get feedback that the repositories are right from @doronhi . This just needs the CI to be resolved.

tylerjw:
To try to clear some things up here there are two repos being linked to.

The one that you are referring to as `librealsense` (`IntelRealSense/librealsense`) provides this package: `librealsense2`

The one you are referring to as `librealsense2` (`IntelRealSense/realsense-ros`) is not "newer" version of the first one, instead it is a ros interface to `librealsense2` and depends on a packaged called `librealsense2`

tylerjw (changes requested):
I suggest these changes:

Change the name of the first one from `librealsense` to `librealsense2` as that is the package it provides.
Change the name of the second one from `librealsense2` to `realsense-ros` as it provides the two ros interface packages `realsense2_camera` and `realsense2_description`.

tylerjw:
Here is a PR against Levi's branch for this PR: https://github.com/Levi-Armstrong/rosdistro/pull/1

I'm not sure how else to add these changes to this PR.

tylerjw:
I've also made this pull request to the IntelRealSense/librealsense repo to fix the rosdep issues so it will build for melodic: https://github.com/IntelRealSense/librealsense/pull/5523

tylerjw (approved):

tfoote (approved):
Thanks for iterating @tylerjw and @Levi-Armstrong this looks good to me.

I've requested a rereview from @clalancette to make sure it looks good to him too.

melodic/distribution.yaml
```diff
@@ -6600,6 +6610,16 @@ repositories:
       url: https://gitlab.com/jlack/rdl.git
       version: master
     status: developed
+  realsense-ros:
```
clalancette: In kinetic, this key is called "realsense2_camera" instead of "realsense-ros".  Any reason for the discrepency?  Even if we do want to rename the key, we should use an underscore instead of a dash here.

doronhi:
Sorry I've been inactive in this important conversation for so long.
 
The melodic version of the debian package should be as similar as possible to the kinetic version.
If any name change should be considered, and it could be the case, it should be done for all ROS distributions and the proper forum for such a debate is the [realsesne-ros github repository](https://github.com/IntelRealSense/realsense-ros).

The librealsense2 version for the kinetic distribution builds without the graphical examples and so avoid the need to depend on X11 and make the [IntelRealSense/librealsense#5523](https://github.com/IntelRealSense/librealsense/pull/5523) unneeded.
Could someone add a link to the farms's auto build log? I wonder how is the automatic build process pass now, without the PR#5523 being merged.

tylerjw:
@clalancette @doronhi @Levi-Armstrong here is PR to rename `realsense-ros` to realsense2_camera` in the melodic distribution.yaml file: https://github.com/Levi-Armstrong/rosdistro/pull/2

tylerjw:
> The librealsense2 version for the kinetic distribution builds without the graphical examples and so avoid the need to depend on X11 and make the IntelRealSense/librealsense#5523 unneeded.

@doronhi The reason I created that PR is I couldn't figure out how to build in a ros environment without those dependencies.  I am currently working on a project in Melodic that depends on both realsense2 and realsense-ros repos and my only solution is to build from source and checkout my change to do it.  The reason I've approached it this way is this is the only way I could get my project to pass CI.

To help realsense2 continue to be usable for others who have reason to do a source ROS build I added rules to your .travis.yml file to have your CI test building realsense2 in a Kinetic and Melodic environment.  I would really appreciate your help getting that change merged or made irrelevant in some way (default build changed to not require X11 dependencies).

doronhi:
Hi @tylerjw ,
The default build includes the graphical examples. That is as part of the company policy and was discussed before, only I can't recall where right now.
Although I am aware of the popularity of the package under Ubuntu 18.04, the package is not officially supported for that environment yet. I thought it would be proper to first declare out support of Melodic and then proceed to release a debian package.
This week I encountered a problem with librealsense2 Ubuntu 18.04 kernels (18.04 and higher) which delayed my progress on the topic.
Never the less, I will create the PR for rosdistro melodic, the same was you did, only with the flags for compiling without X11. Basically copying from our kinetic debian package. I hope to get to that in the next couple of days.
Will that help?

clalancette:
> @clalancette @doronhi @Levi-Armstrong here is PR to rename `realsense-ros` to `realsense2_camera` in the melodic distribution.yaml file: [Levi-Armstrong#2](https://github.com/Levi-Armstrong/rosdistro/pull/2)

@tylerjw That PR makes sense to me.  @Levi-Armstrong please merge that PR into this one, and then we can go ahead and get this whole thing committed.  Any follow-up work that @doronhi is discussing can be done in a separate PR.

doronhi:
@tylerjw , I have submitted #23461 . It is basically the same as this PR, but using a release repository I was able to avoid the dependency in X11, despite it's usage in the original repository.
As I feel it is my responsibility as maintainer of the package to maintain it's debian releases as well, I would like to take it from here, if you approve, of course.

tylerjw:
@doronhi :confetti_ball: of course.  I just want this released for melodic because it will really help us as we are developing a product that uses it and right now my solution is to checkout my branch for the PR on the development repo.  I would still like building it from source using ROS to be added to the CI system there.

tfoote:
@doronhi  This was partially replaced by the full release of librealsense2 in #23461 but the realsense-ros components aren't also duplicated there.

tfoote:
Nevermind there's also #23460 I think that makes this PR obsolete so I'm going to close it.

tfoote:
Thanks for your efforts @tylerjw and @Levi-Armstrong to push this along.

PR #22959, opened 2019-11-11
Add objectpath dependency

joseluis12:
Link: https://pypi.org/project/objectpath/

The agile query language for semi-structured data. #JSON

In my particular case, I needed this package to interact between a ROS node and a Cloud API Rest that returned me the information in Json format. Very large and complex messages are sent and I need to filter and sort them. This package has helped me to perform that task.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2597,6 +2597,10 @@ python-oauth2client:
   fedora: [python-oauth2client]
   gentoo: [dev-python/oauth2client]
   ubuntu: [python-oauth2client]
+python-objectpath-pip:
+  ubuntu:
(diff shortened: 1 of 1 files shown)


tfoote:
As in the [contributing guide](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md), can you explain what your use case is and why you need this particular library? Is this something that you expect many others to also need or want to deploy?

nuclearsandwich (changes requested):
Please address @tfoote's comments.

nuclearsandwich (approved):

clalancette (approved):

PR #22980, opened 2019-11-12
move tracetools repository

iluetkeb:
We've moved the organizations root URL. Otherwise things are unchanged.

### DIFF
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -16326,7 +16326,7 @@ repositories:
   tracetools:
     doc:
       type: git
-       url: https://github.com/bosch-robotics-cr/tracetools.git
+      url: https://github.com/bosch-robotics-cr/tracetools.git
       version: devel
     release:
       tags:
--- a/indigo/distribution.yaml
+++ b/indigo/distribution.yaml
@@ -16323,21 +16323,6 @@ repositories:
       url: https://bitbucket.org/traclabs/trac_ik.git
       version: master
     status: developed
-  tracetools:
-    doc:
-      type: git
-      url: https://github.com/boschresearch/ros1_tracetools.git
-      version: devel
-    release:
-      tags:
-        release: release/indigo/{package}/{version}
-      url: https://github.com/boschresearch/ros1-tracetools-release.git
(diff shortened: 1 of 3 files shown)


indigo/distribution.yaml
```diff
@@ -16323,21 +16323,6 @@ repositories:
       url: https://bitbucket.org/traclabs/trac_ik.git
       version: master
     status: developed
-  tracetools:
```
tfoote: Indigo is EOL we don't accept changes to the released content in the distro anymore. It's standing as a snapshot of the state. 

You can update the URL for a moved repository in case someone is trying to build from source.

tfoote (changes requested):

indigo/distribution.yaml
iluetkeb: I tried changing the URL first (see https://github.com/ros/rosdistro/pull/22980/commits/a4e78b6bfb702df3d50cd4b811b3941626cac47e), but since indigo is EOL, that didn't pass the check either...

Both repos are on Github, so we might just leave the old URL in and rely on Github to do the forward, but that's a bit brittle...

tfoote: The CI script doesn't have this exception. But we can manually override it with just the URL change if that's the only error and it's just the url update. I think a4e78b6 + e84dcee12cf360d95d9d1e795c913f1354078c6f should be good.

iluetkeb: Alright, I've reverted eceec4d94c64210339e206e346c32232a4bfc156

clalancette:
CI is still unhappy because we are touching an EOL distribution.  At this point, I'd suggest just not changing indigo at all, and let the GitHub redirects do their stuff.  Then we can move forward here.

clalancette (approved):
This looks OK, but we can't merge because Kinetic is currently in a [hold](https://discourse.ros.org/t/preparing-for-kinetic-sync-2019-12-11/11907).  We generally recommend one PR per ROS distribution to avoid this problem, but here we can just wait until next week and then merge this.

tfoote (approved):

PR #23004, opened 2019-11-13
robot_systemd: 0.1.2-1 in 'melodic/distribution.yaml' [bloom]

LucidOne:
Increasing version of package(s) in repository `robot_systemd` to `0.1.2-1`:

- upstream repository: https://github.com/LucidOne/robot_systemd.git
- release repository: https://github.com/LucidOne-Release/robot_systemd.git
- bloom version: `0.9.0`

## robot_systemd

```
* Removed pre-emptive conflict
* Added TasksMax=infinity
* Added KillSignal=SIGINT
* Added documentation
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6992,6 +6992,21 @@ repositories:
       url: https://github.com/ros/robot_state_publisher.git
       version: melodic-devel
     status: maintained
+  robot_systemd:
+    doc:
+      type: git
+      url: https://github.com/LucidOne/robot_systemd.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/LucidOne-Release/robot_systemd.git
+      version: 0.1.2-1
+    source:
+      type: git
+      url: https://github.com/LucidOne/robot_systemd.git
+      version: master
+    status: developed
   robot_upstart:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -6992,6 +6992,21 @@ repositories:
       url: https://github.com/ros/robot_state_publisher.git
       version: melodic-devel
     status: maintained
+  robot_systemd:
+    doc:
+      type: git
+      url: https://github.com/LucidOne/robot_systemd.git
```
clalancette: Looking at the upstream sources, there are a couple of license things that I think would be best cleared up before we merge this.

1.  The package.xml lists both MIT and Apache.  Is it really under both of them?  Also, "Apache" as a license probably isn't specific enough; see the list https://opensource.org/licenses/alphabetical .
1.  It would be better if there was a top-level LICENSE file in the repository, which would clarify some of these questions.

clalancette (changes requested):

melodic/distribution.yaml
LucidOne: Thanks for the  feedback, I'll get an update out in a bit.

This covers  dual licensing
https://news.ycombinator.com/item?id=15956857

clalancette: > This covers dual licensing
> https://news.ycombinator.com/item?id=15956857

Dual-licensed packages are allowed, but it is unusual to see it with a package that is so young and has commits from only one committer.  If you want to make it maximally applicable, you could just leave it as MIT or BSD, which can be combined with most other things.

But the dual-license thing isn't my major concern; as I said, we'll take it that way.  My concerns are that "Apache" isn't specific enough, and there are no top-level LICENSE files.

clalancette:
No response in over a month.  I'm going to close this out, but feel free to fix the issues pointed out during review and re-open.  Thanks for the contribution.

PR #23014, opened 2019-11-13
added selectors2-pip library

ejalaa12:
Added the selectors2 pip library which is a port of the python3 selectors library to python2.

See the project on [Pypi](https://pypi.org/project/selectors2/)
> Backported, durable, and portable selectors designed to replace the standard library selectors module.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3873,6 +3873,22 @@ python-seaborn:
     trusty:
       pip:
         packages: [seaborn]
+python-selectors2-pip:
+  arch:
+    pip:
+      packages: [selectors2]
+  debian:
+    pip:
+      packages: [selectors2]
+  fedora:
+    pip:
+      packages: [selectors2]
+  osx:
+    pip:
+      packages: [selectors2]
+  ubuntu:
+    pip:
+      packages: [selectors2]
 python-selenium-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5543,6 +5543,22 @@ rpy2:
   debian: [python-rpy2]
   gentoo: [=dev-python/rpy-2*]
   ubuntu: [python-rpy2]
+selectors2-pip:
```
tfoote: Please use the `python-` prefix. We generally want to stick with the debian style name for consistency.

tfoote (changes requested):

nuclearsandwich:
@ejalaa12 friendly ping. Are you able to address tfoote's feedback or may we close the PR?

ejalaa12:
@nuclearsandwich , sorry, it's been a busy month. I'll do the change.
However, @tfoote maybe I might have misunderstood the [Contributing.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md) as it says:

> If a package is only available via pip use the -pip extension.

That's why I didn't add the `python-` prefix. Should I then rename the entry to `python-selectors2` or `python-selectors2-pip` ?

clalancette:
> > 
> 
> That's why I didn't add the `python-` prefix. Should I then rename the entry to `python-selectors2` or `python-selectors2-pip` ?

What we generally aim to do is to have the rosdep keyname match the debian/ubuntu package name (there are *lots* of exceptions to this in the database, but we can't go back and fix them now without breaking everyone).

With things that are only available in pip, it gets a little more unclear.  In this case, the debian package would *probably* be `python-selectors2` (given that it is a Python library), so our preferred key name would be `python-selectors2-pip`.

ejalaa12:
@clalancette I just pushed those modifications. I get it now, it is in anticipation of the debian package being published. But then once it is published we should remove the `-pip` prefix, no ?

clalancette:
> @clalancette I just pushed those modifications. I get it now, it is in anticipation of the debian package being published. But then once it is published we should remove the `-pip` prefix, no ?

Sort-of.  We'll add a second entry without the `-pip` prefix, to avoid breaking packages that are already depending on the pip one.

clalancette (approved):

tfoote (approved):

PR #23090, opened 2019-11-20
Add realtime_tools noetic source entry

sloretz:
Add Noetic source entry for `realtime_tools`. It seems to build fine on Buster using Python 3.

@bmagyar is `melodic-devel` the right branch, or will `realtime_tools` get a new branch for Noetic?


bmagyar (approved):
Thanks for getting in touch, happy for this to go ahead!

PR #23091, opened 2019-11-20
Add control_msgs Noetic source entry

sloretz:
Add Noetic source entry for control_msgs. It seems to build fine on Buster using Python 3.

@bmagyar is kinetic-devel the right branch, or will control_msgs get a new branch for Noetic?


bmagyar (approved):
Thanks for getting in touch, happy for this to go ahead!

PR #23092, opened 2019-11-20
Add four_wheel_steering_msgs Noetic source entry

sloretz:
Add Noetic source entry for four_wheel_steering_msgs. It seems to build fine on Buster using Python 3.

@vincentrou is master the right branch, or will four_wheel_steering_msgs get a new branch for Noetic?


vincentrou:
The branch will not change. Thanks.

PR #23094, opened 2019-11-20
Add pcl_msgs noetic source entry

sloretz:
Add Noetic source entry for `pcl_msgs`. It seems to build fine on Buster using Python 3.

@LucidOne @paulbovbel is `indigo-devel` the right branch, or will `pcl_msgs` get a new branch for Noetic?


paulbovbel:
LGTM!

PR #23122, opened 2019-11-21
Update to eloquent release branches where relevant

mjcarroll:
Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -521,7 +521,7 @@ repositories:
       test_pull_requests: false
       type: git
       url: https://github.com/eProsima/Fast-RTPS.git
-      version: 1.9.x
+      version: 141b778c66c4f1f7aeb5c55551c173e3237beae9
     status: developed
   fmi_adapter_ros2:
     doc:
@@ -2203,7 +2203,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros/ros_tutorials.git
-      version: eloquent
+      version: eloquent-devel
     release:
       tags:
         release: release/eloquent/{package}/{version}
@@ -2213,7 +2213,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/ros/ros_tutorials.git
-      version: eloquent
+      version: eloquent-devel
     status: maintained
   uncrustify_vendor:
     release:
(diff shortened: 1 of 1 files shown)


eloquent/distribution.yaml
```diff
@@ -521,7 +521,7 @@ repositories:
       test_pull_requests: false
       type: git
       url: https://github.com/eProsima/Fast-RTPS.git
-      version: 141b778c66c4f1f7aeb5c55551c173e3237beae9
+      version: 1.9.x
```
nuclearsandwich: For other rosdistros we've specified the precisely released Fast-RTPS tag. That was due in part to not having a stable development branch to track but I could still see merit in keeping an exact release for upstream code. Only CycloneDDS and Fast-RTPS are really in this situation.

eloquent/distribution.yaml
```diff
@@ -2203,7 +2203,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros/ros_tutorials.git
-      version: eloquent-devel
+      version: eloquent
```
nuclearsandwich: I think the `eloquent-devel` branches can remain (and it may make sense to delete the `eloquent` branch. Usually the presence of a -devel branch indicates that the maintainer is sticking with a classic ROS branching strategy of forward-compatible devel branches rather than backward-compatible distro branches which became a thing in ROS 2.

nuclearsandwich (dismissed):
I have a couple observations noted in line comments. Other than that a quick local check verified that at the very least there is agreement bewtween source and doc entries for repositories that have both.

eloquent/distribution.yaml
mjcarroll: I switched this back to the SHA

eloquent/distribution.yaml
mjcarroll: Switched this back to eloquent-devel

eloquent/distribution.yaml
```diff
@@ -1415,13 +1415,13 @@ repositories:
     source:
       type: git
       url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing.git
-      version: master
+      version: 0.2.10
```
nuclearsandwich: Since this is a tag `test_commits` and `test_pull_requests` will need to be made explicitly false.

mjcarroll: Done!

eloquent/distribution.yaml
dirk-thomas: I deleted the created `eloquent` branch in the repo.

dirk-thomas (approved):

nuclearsandwich (approved):

PR #23132, opened 2019-11-23
added mas_cartesian_control for carteisian control of manipulator

abhishekpadalkar:


kinetic/distribution.yaml
```diff
@@ -6958,6 +6958,12 @@ repositories:
         release: release/kinetic/{package}/{version}
       url: https://github.com/MarvelmindRobotics/marvelmind_nav-release.git
       version: 1.0.8-0
+  mas_cartesian_control:
+    source:
+      type: git
+      url: https://github.com/b-it-bots/mas_cartesian_control.git
```
clalancette: The package.xml lists the BSD license, but there is no LICENSE file at the top-level of the repository explicitly saying which BSD license is being used.  Please add a LICENSE file at the top level so we can be sure of the BSD variant you are using; the list of approved open-source licenses is at https://opensource.org/licenses

clalancette (changes requested):
Need clarification on the license.

kinetic/distribution.yaml
abhishek098: A top-level license has been added. Thank you.

clalancette (approved):
Looks good, thanks for iterating.

PR #23141, opened 2019-11-24
[dep] Add rsyslog.

130s:
`rsyslog` is an [alternative to `syslog` etc.](https://serverfault.com/questions/692309/what-is-the-difference-between-syslog-rsyslog-and-syslog-ng) and system monitoring tool can utilize its api.   

debian https://packages.debian.org/buster/rsyslog
fedora https://rpmfind.net/linux/rpm2html/search.php?query=rsyslog&submit=Search+...&system=&arch=
macports [none](https://www.macports.org/ports.php?by=name&substr=rsyslog)
ubuntu https://packages.ubuntu.com/xenial/rsyslog

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4820,6 +4820,10 @@ rsync:
   gentoo: [net-misc/rsync]
   opensuse: [rsync]
   ubuntu: [rsync]
+rsyslog:
+  debian: [rsyslog]
+  fedora: [rsyslog]
+  ubuntu: [rsyslog]
 rti-connext-dds-5.3.1:
   ubuntu:
     bionic: [rti-connext-dds-5.3.1]


clalancette (changes requested):
@130s can you please update the links in the description to point to the packages in the respective distributions, rather than just the wiki pages?  There are links in https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions to those package repositories.  Thanks.

130s:
Sure. Entries in OP updated as following. I removed `gentoo` as I'm way too unfamiliar with it so I have no idea which web page it should list.

debian https://packages.debian.org/buster/rsyslog
fedora https://rpmfind.net/linux/rpm2html/search.php?query=rsyslog&submit=Search+...&system=&arch=
macports [none](https://www.macports.org/ports.php?by=name&substr=rsyslog)
ubuntu https://packages.ubuntu.com/xenial/rsyslog

cottsay (approved):
Here's the official Fedora package link: https://apps.fedoraproject.org/packages/rsyslog

clalancette (approved):

PR #23150, opened 2019-11-25
ubt_msgs: 0.0.1-3 in 'melodic/distribution.yaml' [bloom]

qinlifu-cpp:
Increasing version of package(s) in repository `ubt_msgs` to `0.0.1-3`:

- upstream repository: https://github.com/ubtvisbot/ubt_msgs.git
- release repository: https://github.com/ubtvisbot/ubt_msgs-release.git
- bloom version: `0.9.0`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -9686,6 +9686,21 @@ repositories:
       url: https://github.com/KumarRobotics/ublox.git
       version: master
     status: maintained
+  ubt_msgs:
+    doc:
+      type: git
+      url: https://github.com/ubtvisbot/ubt_msgs.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/ubtvisbot/ubt_msgs-release.git
+      version: 0.0.1-3
+    source:
+      type: git
+      url: https://github.com/ubtvisbot/ubt_msgs.git
+      version: master
+    status: developed
   um6:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -9686,6 +9686,21 @@ repositories:
...
+      url: https://github.com/ubtvisbot/ubt_msgs.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/ubtvisbot/ubt_msgs-release.git
+      version: 0.0.1-3
+    source:
+      type: git
+      url: https://github.com/ubtvisbot/ubt_msgs.git
```
clalancette: There is no license on this repository; the package.xml just says `TODO`, and there is no top-level LICENSE file in the repository.  Please add a license to this repository.

clalancette (changes requested):
Repository needs a license.

melodic/distribution.yaml
ubtvisbot: Hi, is the license of this repository necessary?

clalancette: Yes.  Otherwise, there is no way for a consumer of the package to tell whether they are allowed to use the package in their application.  For instance, some companies can't use proprietary licensed software.  Some can't use GPL-licensed software.  Without this information, they can't make a determination on whether they can use your software.

tfoote:
Yes a license is necessary, if it doesn't have a license we cannot redistribute it either.

clalancette:
No response in over a month, so closing out.  Please feel free to fix the issues pointed out during review and re-open.  Thanks for the contribution.

PR #23160, opened 2019-11-26
py_script-release: 0.0.1-2 in 'kinetic/distribution.yaml' [bloom]

dippingconda:
Increasing version of package(s) in repository `py_script-release` to `0.0.1-2`:

- upstream repository: https://github.com/dippingconda/sim_fall_det.git
- release repository: https://github.com/dippingconda/py_script-release.git
- bloom version: `0.9.0`

## humanoid_shp

```
* first commit
* Contributors: pshyeok
```

## humanoid_shp_moveit

```
* first commit
* Contributors: pshyeok
```

## py_script

```
* first commit
* Contributors: pshyeok
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -10259,6 +10259,16 @@ repositories:
       url: https://github.com/PX4/px4_msgs-release.git
       version: 1.0.0-3
     status: developed
+  py_script-release:
+    release:
+      packages:
+      - humanoid_shp
+      - humanoid_shp_moveit
+      - py_script
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/dippingconda/py_script-release.git
+      version: 0.0.1-2
   py_trees:
     doc:
       type: git


kinetic/distribution.yaml
```diff
@@ -10259,6 +10259,16 @@ repositories:
       url: https://github.com/PX4/px4_msgs-release.git
       version: 1.0.0-3
     status: developed
+  py_script-release:
```
clalancette: While this name isn't used for package names, it should still be descriptive of what the package does.  In this case, it looks like this wants to be something more like `humanoid_shp`, or something like that.

kinetic/distribution.yaml
```diff
@@ -10259,6 +10259,16 @@ repositories:
...
+    release:
+      packages:
+      - humanoid_shp
+      - humanoid_shp_moveit
+      - py_script
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/dippingconda/py_script-release.git
```
clalancette: And then this release repository should also be called something like `humanoid_shp-release.git`.  Also, the source repository name doesn't match up with the release repository name.

kinetic/distribution.yaml
```diff
@@ -10259,6 +10259,16 @@ repositories:
       url: https://github.com/PX4/px4_msgs-release.git
       version: 1.0.0-3
     status: developed
+  py_script-release:
+    release:
+      packages:
+      - humanoid_shp
+      - humanoid_shp_moveit
+      - py_script
```
clalancette: This is much too generic a name.  This will be used as the package name, so just imagine a user trying to install `ros-kinetic-py-script`; what does that do?  The name of the package should reflect what it is scripts for.

clalancette (changes requested):
The naming here is all over the place.  My suggestion to fix all of this is to do the following:

1.  Rename the source repository to something like `humanoid_shp`.
1.  Rename the `py_script` package within the source repository to something like `humanoid_shp_script`.
1.  Rename the release repository to something like `humanoid_shp-release`.
1.  Close this PR.
1.  Use `bloom-release` to create a new PR from the new repository names.

Because of all of this, I'm actually going to close out this PR, since this will need a new PR anyway.  Feel free to keep commenting to ask questions, though.

PR #23162, opened 2019-11-26
humanoid_shp-release: 0.0.1-2 in 'kinetic/distribution.yaml' [bloom]

dippingconda:
Increasing version of package(s) in repository `humanoid_shp-release` to `0.0.1-2`:

- upstream repository: https://github.com/dippingconda/sim_fall_det.git
- release repository: https://github.com/dippingconda/humanoid_shp-release.git
- bloom version: `0.9.0`

## humanoid_shp

```
* first commit
* Contributors: pshyeok
```

## humanoid_shp_moveit

```
* first commit
* Contributors: pshyeok
```

## py_script

```
* first commit
* Contributors: pshyeok
```

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -4898,6 +4898,16 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/humanoid_navigation.git
       version: kinetic-devel
     status: maintained
+  humanoid_shp-release:
+    release:
+      packages:
+      - humanoid_shp
+      - humanoid_shp_moveit
+      - py_script
+      tags:
+        release: release/kinetic/{package}/{version}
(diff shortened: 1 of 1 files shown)


clalancette:
The name of the package shouldn't include `-release`.  Just think about the final debian name, and it would be `ros-kinetic-humanoid-shp-release`, which doesn't make a whole lot of sense to the end-user.

Also, we highly recommend that the release repository be separate from the source repository; doing anything else will lead to lots of problems later.  What we usually suggest is one repository that contains the source code (in this case, `humanoid_shp`), and one release for the release (`humanoid_shp-release`).  There's some more information about doing this [here](http://wiki.ros.org/ROS/ReleasingAPackage).  For these reasons, I'm going to close this out, but feel free to keep commenting here if you have questions or concerns.

PR #23163, opened 2019-11-26
Add benchmark to rosdep

efernandez:
Add benchmark to rosdep
    
Add Google's benchmark library to rosdep/base.yaml, for:
    
* Debian: https://packages.debian.org/buster/libbenchmark-dev
* Fedora: https://src.fedoraproject.org/rpms/google-benchmark
* Ubuntu: https://packages.ubuntu.com/bionic/libbenchmark-dev
    
This library became available in the following stable versions:
    
* Debian: Buster
* Fedora: EPEL 8
* Ubuntu: Bionic

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -226,6 +226,14 @@ beep:
   fedora: [beep]
   gentoo: [app-misc/beep]
   ubuntu: [beep]
+benchmark:
+  debian:
+    '*': [libbenchmark-dev]
+    stretch: null
+  fedora: [google-benchmark]
+  ubuntu:
+    '*': [libbenchmark-dev]
+    xenial: null
 binutils:
   arch: [binutils]
   debian: [binutils-dev]


nuclearsandwich (approved):

rosdep/base.yaml
```diff
@@ -226,6 +226,9 @@ beep:
   fedora: [beep]
   gentoo: [app-misc/beep]
   ubuntu: [beep]
+benchmark:
+  ubuntu:
```
clalancette: Typically I'd like to see this for at least Debian and Fedora as well (assuming the package is available here).  Can you add them?  Thanks.

efernandez: Done.

Also change:
``` bash
ubuntu:
  bionic:
```
to simply
``` bash
ubuntu:
```

Since that seems to be less restrictive and easier to maintain for future version, if I'm understanding things right. :)

efernandez:
@clalancette I addressed your comment and updated the description of this PR with the same comment that is on the commit, that I just updated.

Note that I don't have a system running Debian or Fedora, so I'm just relying on the links provided.

clalancette:
> Note that I don't have a system running Debian or Fedora, so I'm just relying on the links provided.

Yeah, that's totally reasonable.  Thanks.

rosdep/base.yaml
```diff
@@ -226,6 +226,10 @@ beep:
   fedora: [beep]
   gentoo: [app-misc/beep]
   ubuntu: [beep]
+benchmark:
+  debian: [libbenchmark-dev]
```
clalancette: Looking at this further, it looks like this is only supported on buster and later (and stretch-backports, but we don't usually allow those as rosdep keys).  Given that stretch is still supported (and still used by kinetic), we want to make sure we don't give the impression that it is available there.  So this would become:

```
debian:
  stretch: null
  *: [libbenchmark-dev]
```

rosdep/base.yaml
```diff
@@ -226,6 +226,10 @@ beep:
...
+  fedora: [google-benchmark]
+  ubuntu: [libbenchmark-dev]
```
clalancette: The same goes for Ubuntu here, where this package is available in bionic and later, but not xenial (which is still supported and used by Kinetic).  A similar fix here would be:

```
ubuntu:
  xenial: null
  *: [libbenchmark-dev]
```

clalancette (changes requested):
I just noticed that this package isn't available on some still supported and used distributions.  I've given suggestions on how to fix it below.  Sorry for the back-n-forth.

rosdep/base.yaml
efernandez: I didn't know about that syntax. Actually, looking at other places in the file it seems that `*` should be quoted: `'*'`

Today I learnt something! :smile: 

Changed.

rosdep/base.yaml
efernandez: Changed.

efernandez:
> I just noticed that this package isn't available on some still supported and used distributions. I've given suggestions on how to fix it below. Sorry for the back-n-forth.

No problem. Thanks for the suggestions. It's updated now.

rosdep/base.yaml
```diff
@@ -226,6 +226,14 @@ beep:
...
   gentoo: [app-misc/beep]
   ubuntu: [beep]
+benchmark:
+  debian:
+    stretch: null
+    '*': [libbenchmark-dev]
+  fedora: [google-benchmark]
+  ubuntu:
+    xenial: null
+    '*':[libbenchmark-dev]
```
clalancette: CI is failing.  There definitely needs to be a space between `'*'` and the value.  Looking at all of the current rules, it looks like it *may* also be the case that the `*` has to come first in the list.  Can you make those two changes and see if CI is more happy?

efernandez: Oh, that missed blank space was a typo.

I've updated the PR with the changes.

clalancette (approved):
Looks good to me.  Thanks for iterating.

PR #23169, opened 2019-11-27
Add Python 3 entry for imageio

sgvandijk:
Python 3 entry for imageio, the library to for reading and writing image, video, scientific, and volumetric data formats.

* Debian: https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=python3-imageio

    This package is only available since Buster, should it cater for older versions as well? I assume that for those the Python 2 version is more relevant

* Fedora and Gentoo: use same as original Python 2 entry which should install the Python 3 versions

* Ubuntu: https://packages.ubuntu.com/search?keywords=python3-imageio&searchon=names&suite=all&section=all

    Similar question as for Debian, only available since Bionic

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5174,6 +5174,10 @@ python3-ifcfg:
   rhel: ['python%{python3_pkgversion}-ifcfg']
   ubuntu:
     bionic: [python3-ifcfg]
+python3-imageio:
+  debian: [python3-imageio]
+  gentoo: [dev-python/imageio]
+  ubuntu: [python3-imageio]
 python3-lark-parser:
   debian:
     buster: [python3-lark-parser]


nuclearsandwich (changes requested):
>     * This package is only available since Buster, should it cater for older versions as well?

If the package isn't available in the platform repository it can't be released onto the buildfarm. As long as your use case is covered by the support platforms I don't think we need to backfill with pip keys. Leaving the earlier distros undefined makes sense to me.

Let's avoid mixing pip keys with platform keys and remove the Fedora entry since the package isn't available there.

sgvandijk:
Thanks for the feedback :+1: I removed the Fedora entry

nuclearsandwich (approved):

clalancette (approved):

PR #23173, opened 2019-11-27
Updating carla_ros_bridge to documentation index for kinetic distro

vimalrajliya:
I'd like package carla_ros_bridge to be indexed and documented on ros.org.
Earlier it was linked with the master which we want to change and link with branch 0.9.6, so we can maintain it properly.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -1250,7 +1250,7 @@ repositories:
       url: https://github.com/ipa320/care-o-bot-release.git
       version: 0.6.7-0
     status: maintained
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -1250,11 +1250,15 @@ repositories:
       url: https://github.com/ipa320/care-o-bot-release.git
       version: 0.6.7-0
     status: maintained
-  carla_ros_bridge:
+  carla_ros_bridge_0.9.6:
```
fpasch: please keep name as it was before.

PR #23198, opened 2019-12-03
Add GTSAM as source package

jlblancoc:
Added as a follow up of https://github.com/borglab/gtsam/pull/173

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -941,6 +941,16 @@ repositories:
       url: https://github.com/swri-robotics/gps_umd.git
       version: dashing-devel
     status: maintained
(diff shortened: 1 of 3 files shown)


clalancette (approved):
In general, please open a separate PR per ROS distribution.  The general reason is that they are not all on the same release schedule, so it may be the case that we can't merge the PR because one of the ROS distros is in a freeze at the moment.

That being said, you picked an opportune moment to open this one.  In this case, all of the ROS distros in here are currently not frozen, so we can merge this as-is.

jlblancoc:
Ok, that's good to know and totally makes sense.
Thanks

PR #23199, opened 2019-12-03
Push trac ik 1.5.1

BryceStevenWilley:
I noticed that despite the (as far as I can tell) successful [buildfarm](http://build.ros.org/job/Msrc_uB__trac_ik__ubuntu_bionic__source/) that was triggered a few months ago, the package version of `trac_ik` was never updated. The [release repo](https://github.com/traclabs/trac_ik-release/blob/master/README.md) seems to have been updated fine, and I think the only step left is to update this repo. 

Tagging @pbeeson, in case he wants to take over this PR.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -15424,7 +15424,7 @@ repositories:
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/traclabs/trac_ik-release.git
-      version: 1.5.0-0
+      version: 1.5.1-1
     source:
       type: git
       url: https://bitbucket.org/traclabs/trac_ik.git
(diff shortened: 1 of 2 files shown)


clalancette (commented):
In general, please open a separate PR per ROS distro.  The reason for this is that the sync schedules for the separate distributions are different, so it may be the case that we could merge for one distribution, but not another.  In this case, they are both not in freeze, but that may not be the case soon.

Also, I'd prefer not to merge this until the maintainer says they want to.  There may be a particular reason they didn't do this, so let's wait for @pbeeson to comment before moving forward here.

pbeeson:
I did not do this intentionally.  I'm guessing I just missed a step by
either being busy or not following the documentation fully.  Please merge
this, as 1.5.1 was a bug fix over 1.5.0.

BryceStevenWilley:
Thanks for the advice @clalancette, I'll keep separate PRs per distro in the future.

pbeeson:
Thanks for catching this @BryceStevenWilley

PR #23223, opened 2019-12-05
uwb_hardware_package is added under the kinetic.

elcingencturk:
I created a package which is uwb_hardware_driver in my machine - which is in the same directory with indoor_localization package. And also, I've already add this package in ROS index (http://wiki.ros.org/indoor_localization). (Both packages work together.)

When I check the last build in the Travis, I am getting this error: 
the following packages/stacks could not have their rosdep keys resolved to system dependencies:
indoor_localization: Cannot locate rosdep definition for [uwb_hardware_driver]

I thought that this error could be fixed if I add the package uwb_hardware_driver under ROS index. So, I'm openning this pull request for this reason.

### DIFF
--- a/jade/distribution.yaml
+++ b/jade/distribution.yaml
@@ -7638,6 +7638,11 @@ repositories:
       url: https://github.com/bosch-ros-pkg/usb_cam.git
       version: develop
     status: maintained
+  uwb_hardware_driver:
+    doc:
+      type: git
+      url: https://github.com/inomuh/uwb_hardware_driver.git
+      version: master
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):
Jade is a discontinued ROS distributions, so we don't allow any more updates to it.  Please remove the changes to Jade.

Also, please use a separate PR per repository you are adding.  It is just easier to track them that way.  Because of these two things, I'm actually going to close this PR out, but feel free to reopen separate ones for the pieces you are adding to kinetic.

PR #23238, opened 2019-12-05
added vtec_ros to melodic

lukscasanova:
Adding vtec_ros to melodic and updating github repo link.


clalancette (approved):
Looks OK to me, but we can't merge currently since Kinetic is in a freeze.  We generally recommend opening a separate PR per ROS distribution to avoid this problem.

PR #23248, opened 2019-12-07
Add asr_msgs as doc dependency to several packages

Tobi2001:
Adding depends tag to packages which provide interface files depending on asr_msgs to fix build errors on jenkins doc server (as suggested in [this](https://answers.ros.org/question/339311/jenkins-doc-build-fails-because-of-nested-custom-messages-which-are-hosted-in-external-unreleased-package/) ROS-answers post)

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -378,6 +378,8 @@ repositories:
       version: master
   asr_direct_search_manager:
     doc:
+      depends:
+      - asr_msgs
       type: git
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Looks good to me.  We're currently in a Kinetic [sync hold](https://discourse.ros.org/t/preparing-for-kinetic-sync-2019-12-11/11907), so we'll wait for that to clear before merging.

PR #23261, opened 2019-12-09
Add pip install dual quaternions ros

Achllle:
## Description

Python 2 package for dual quaternion operations. This installs the ROS wrapper which has a dependency for the pure Pythonic [dual-quaternions](https://pypi.org/project/dual-quaternions/) package

* Package on PyPi: [dual-quaternions-ros](https://pypi.org/project/dual-quaternions-ros/)
* what are dual quaternions?: https://github.com/Achllle/dual_quaternions_ros/tree/iss42

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -130,10 +130,10 @@ ds4drv-pip:
       packages: [ds4drv]
 dual-quaternions-ros-pip:
   debian:
-    pip: 
+    pip:
       packages: [dual-quaternions-ros]
   fedora:
-    pip: 
+    pip:
       packages: [dual-quaternions-ros]
   ubuntu:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -129,6 +129,12 @@ ds4drv-pip:
     pip:
       packages: [ds4drv]
 dual-quaternions-ros-pip:
+  debian:
+    pip: 
+      packages: [dual-quaternions-ros]
+  fedora:
+    pip: 
+      packages: [dual-quaternions-ros]
   ubuntu:
     pip:
       packages: [dual-quaternions-ros]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -128,6 +128,10 @@ ds4drv-pip:
   ubuntu:
     pip:
       packages: [ds4drv]
+dual-quaternions-ros-pip:
+  ubuntu:
+    pip:
+      packages: [dual-quaternions-ros]
 epydoc:
   arch: [epydoc]
   debian: [python-epydoc]


tfoote (changes requested):
This package looks interesting however I'm not sure that a pip installation is the correct way to deploy it. In particular it's not something that a user can depend on from a package that they'd like to release into a ROS distribution which will make it much less reusable.

And secondly looking inside implementation there are imports from things like [`geometry_msgs`](https://github.com/Achllle/dual_quaternions_ros/blob/1edf6517091dab59581f966f82e1887642b50a90/dual_quaternions_ros/dual_quaternions.py#L11) that are not listed in the `install_requires` [listing](https://github.com/Achllle/dual_quaternions_ros/blob/1edf6517091dab59581f966f82e1887642b50a90/setup.py#L16). As such the package itself it not useable with only it's declared dependencies.

Thus in it's current state I would suggest that this is not a library that's ready to be merged. Instead I would suggest that if it's to be released it should be released as a ROS package with the appropriate dependencies on things like `geometry_msgs` and then it can be leveraged easily by the community.

Achllle:
Thanks for the feedback @tfoote 

I'll fix the lack of dependency on `geometry_msgs`. I'm not sure however if I understand what would be the better way to deploy this. With this package I am trying to encourage the use of dual quaternions in the robotics community, so whatever yields the smoothest installation and adoption. You suggest a ROS package but you seem to suggest there's an even better way. Would that be something like `apt`? I would like for the code to also be used in non-ROS projects.

I appreciate your help, I'm new to deploying open source.

tfoote:
You can first make a ROS package that's designed to work from source in a ROS workspace: http://wiki.ros.org/ROS/Tutorials/CreatingPackage which people will be able to clone and drop into their ROS source workspace to use.

Then you can release it into a distro to have it built on the Open Robotics build farm. When it's on the buildfarm, the package will be built into debs available via apt after adding the ROS sources. This is where you will see the most adoption as it makes things the easiest to use for user. 

There's a tutorial for that here: http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

Achllle:
I'll do that. A follow-up question after trying to figure out how to go about this through looking at documentation and other repos:

How do I keep the [current core functionality/ROS conversion splitup](https://github.com/Achllle/dual_quaternions_ros/pull/45) intact and would you recommend splitting up into two repos for that? Can I still have the pure pythonic part be a pip package? If so, how can I refer to the nosetests if they are not located within the ROS package?

Achllle:
In order to keep the pythonic code separated from the ROS conversion methods, should I either:

1. add a rosdep rule for the dual-quaternions (pure python) package and depend on that in the ROS package dual-quaternions-ros (conversion methods), or
2. have both be catkin packages and release both.

Would you recommend splitting up into two git repos?

Achllle:
Either way, the package depends on numpy-quaternion (#19944) which is pip only. With your comments here and on that PR

> In particular it's not something that a user can depend on from a package that they'd like to release into a ROS distribution which will make it much less reusable
> Note that you will not be able to make releases against a pip entry but you can use it when building from source.

you mean that I cannot use this to release my package, perhaps unless releasing it as a pip package anyways? If that's the case, what would you recommend as the way forward? Release `numpy-quaternion` with bloom too?

gavanderhoorn:
You could perhaps use [catkin_virtualenv](http://wiki.ros.org/catkin_virtualenv).

tfoote:
In general for making releases to any package manager you have to have everything in the same package manager. If you make ROS releases we generate apt packages. 

If you're working from source you can mix dependencies before you build.

You could look at bundling it in like @gavanderhoorn suggests or else you could package it separately.

Achllle:
I'll try @gavanderhoorn 's suggestion after confirming that the license permits it (question [here](https://answers.ros.org/question/340978/can-i-use-catkin_virtualenv-to-help-release-a-non-gplv2-package-with-bloom/)). If that doesn't work, I'll look into packaging `numpy-quaternion` although that seems tough given my lack of experience with releasing.

tfoote:
I'm going to close this as it looks like we don't want to merge these rosdep rules.

@Achllle please ask for help on answers.ros.org if you need more help getting things released.

Achllle:
It was honestly looking to me like releasing as a pip only package was the most viable option here.

PR #23278, opened 2019-12-11
Add python-pyzbar

mortenfyhn:
[This](https://pypi.org/project/pyzbar/) package.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3484,6 +3484,12 @@ python-pyxhook-pip:
     pip:
       packages: [pyxhook]
 python-pyzbar-pip:
+  debian:
+    pip:
+      packages: [pyzbar]
+  fedora:
+    pip:
+      packages: [pyzbar]
   ubuntu:
     pip:
       packages: [pyzbar]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3483,6 +3483,10 @@ python-pyxhook-pip:
   ubuntu:
     pip:
       packages: [pyxhook]
+python-pyzbar:
```
clalancette: Since this is a pip-only package, please add a `-pip` suffix.  Also, while we are at it, we may as well add entries for debian and fedora as well.  Thanks.

clalancette (changes requested):

rosdep/python.yaml
mortenfyhn: Thanks for the feedback. I've fixed it now. With the new suffix i should use the full name `python-pyzbar-pip` in my `package.xml` dependencies, right?

clalancette: > With the new suffix i should use the full name `python-pyzbar-pip` in my `package.xml` dependencies, right?

Correct.

clalancette (approved):

cottsay (approved):

PR #23281, opened 2019-12-11
Enable test_abi in: rclcpp, rcl, rclpy, turtlesim, ament_cmake

j-rivero:
Please note that we probably need to wait until merging:
 * ros_buildfarm support: https://github.com/ros-infrastructure/ros_buildfarm/pull/681

This PR will enable the abi-checking for a reduced set of jobs in order to get a production testing before we enable it widely. All the builds were tested in the testing jenkins instance (attached links on each one). The candidates I want to test are:

 * ament_cmake -- no binary interface, no failure
[Build Status](http://18.206.154.49/job/Ddev__ament_cmake__ubuntu_bionic_amd64/)
 * libyaml_vendor -- 1 header / library
[Build Status](http://18.206.154.49/job/Ddev__libyaml_vendor__ubuntu_bionic_amd64/)
 * rclcpp -- 107 header files / 3 libraries
[Build Status](http://18.206.154.49/job/Ddev__rclpy__ubuntu_bionic_amd64/)
 * rcl -- 43 header files / 4 libraries
[Build Status](http://18.206.154.49/job/Ddev__rcl__ubuntu_bionic_amd64/)
 * rclpy -- no binary interface, no failure
[Build Status](http://18.206.154.49/job/Ddev__rclpy__ubuntu_bionic_amd64/)
 * turtlesim -- 92 headers files / 4 libraries (include idl code, tricky)
[Build Status](http://18.206.154.49/job/Ddev__turtlesim__ubuntu_bionic_amd64/)

Also needs rosdistro 0.8 release [for the implementation of test_abi parsing](https://github.com/ros-infrastructure/rosdistro/pull/147).

//cc @chapulina

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -1804,8 +1804,8 @@ repositories:
       url: https://github.com/ros2-gbp/rcl-release.git
       version: 0.7.8-1
     source:
-      test_pull_requests: true
       test_abi: true
+      test_pull_requests: true
       type: git
       url: https://github.com/ros2/rcl.git
       version: dashing
@@ -1869,8 +1869,8 @@ repositories:
       url: https://github.com/ros2-gbp/rclcpp-release.git
       version: 0.7.12-1
     source:
-      test_pull_requests: true
       test_abi: true
+      test_pull_requests: true
       type: git
       url: https://github.com/ros2/rclcpp.git
       version: dashing
@@ -1886,8 +1886,8 @@ repositories:
       url: https://github.com/ros2-gbp/rclpy-release.git
       version: 0.7.9-1
     source:
-      test_pull_requests: true
       test_abi: true
+      test_pull_requests: true
       type: git
       url: https://github.com/ros2/rclpy.git
       version: dashing
@@ -3119,8 +3119,8 @@ repositories:
       url: https://github.com/ros2-gbp/ros_tutorials-release.git
       version: 1.0.1-1
     source:
-      test_pull_requests: true
       test_abi: true
+      test_pull_requests: true
       type: git
       url: https://github.com/ros/ros_tutorials.git
       version: dashing-devel
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -94,6 +94,7 @@ repositories:
       url: https://github.com/ros2-gbp/ament_cmake-release.git
       version: 0.7.4-1
     source:
+      test_abi: true
       type: git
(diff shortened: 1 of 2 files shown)


dashing/distribution.yaml
```diff
@@ -1261,6 +1262,7 @@ repositories:
       url: https://github.com/ros2-gbp/libyaml_vendor-release.git
       version: 1.0.0-1
     source:
+      test_abi: true
```
chapulina: This one doesn't have `test_pull_requests` or `test_commits` enabled, so does `test_abi` have any effect?

nuclearsandwich: test_commits defaults to true for Dashing builds: https://github.com/ros2/ros_buildfarm_config/blob/b90b4b332131c4e8b66ad1784cf00b9d16a71825/dashing/source-build.yaml#L91-L94

dirk-thomas:
It might make sense to also enable the flag for Eloquent since those are more likely to receive backports.

dashing/distribution.yaml
```diff
@@ -94,6 +94,7 @@ repositories:
       url: https://github.com/ros2-gbp/ament_cmake-release.git
       version: 0.7.4-1
     source:
+      test_abi: true
```
dirk-thomas: What is the ABI test actually doing for this repo which only contains CMake stuff?

j-rivero: the short answer is: nothing. The only point of having it enable for cmake code is to see how the auto-abi tool behaves when it does not find any library or header. It should not fail. So my propose here was to test the non-failure feature of running the tool on non c/c++ code.

j-rivero:
> It might make sense to also enable the flag for Eloquent since those are more likely to receive backports.

I added them although a false positive appeared in rclcpp. I've been investigating it in: https://github.com/osrf/auto-abi-checker/issues/15 . Still unsolved.

eloquent/distribution.yaml
```diff
@@ -1261,6 +1263,7 @@ repositories:
       url: https://github.com/ros2-gbp/rclcpp-release.git
       version: 0.8.3-1
     source:
+      test_abi: false
```
dirk-thomas: This `false` is the default just remove this line.

dirk-thomas (approved):

nuclearsandwich (approved):

ros-discourse:

https://discourse.ros.org/t/enabling-the-abi-checker-for-your-next-ros2-f-releases/12811/1

PR #23297, opened 2019-12-12
Removing serial_driver from Autoware.Auto

JWhitleyWork:
For release under transport_drivers. See https://github.com/ros-drivers/transport_drivers


JWhitleyAStuff:
Approved by @esteve . See https://github.com/ros-drivers/transport_drivers/pull/6

PR #23303, opened 2019-12-13
keras-retinanet: 0.5.1 in 'rosdep/python.yaml'

ricardoet:
Rosdep rule for keras-retinanet 0.5.1. Deep learning architecture used for object detection.

https://pypi.org/project/keras-retinanet/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2132,6 +2132,10 @@ python-keras-pip:
   ubuntu:
     pip:
       packages: [keras]
+python-keras-retinanet:
+  ubuntu:
+    pip:
+      packages: [keras-retinanet]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2132,6 +2132,19 @@ python-keras-pip:
   ubuntu:
     pip:
       packages: [keras]
+python-keras-retinanet:
```
clalancette: Since this is a pip package on all platforms, please append `-pip` to the key.

clalancette (changes requested):

clalancette:
No response in over a month, so closing.  Please feel free to fix the issues pointed out in review and reopen.  Thanks for the contribution.

PR #23307, opened 2019-12-15
Add python-can on Python 3

felixdivo:
This is the same package as `python-can`, just for Python 3 instead of Python 2. To quote the docs:
> The python-can library provides Controller Area Network support for Python, providing common abstractions to different hardware devices, and a suite of utilities for sending and receiving messages on a CAN bus.

It is also used in some robotics environments and thus qualifies for this list.

For [Debian](https://packages.debian.org/search?keywords=python3-can) & [Ubuntu](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-can&searchon=names).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5060,7 +5060,9 @@ python3-cairo:
 python3-can:
   debian: [python3-can]
   fedora: [python3-can]
-  ubuntu: [python3-can]
+  ubuntu:
+    '*': [python3-can]
+    xenial: null
 python3-catkin-pkg:
   debian: [python3-catkin-pkg]
   fedora: [python3-catkin_pkg]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5057,6 +5057,11 @@ python3-cairo:
     slackpkg:
       packages: [py3cairo]
   ubuntu: [python3-cairo]
+python3-can:
+  debian:
+    '*': [python3-can]
+  ubuntu:
+    '*': [python3-can]
 python3-catkin-pkg:
   debian: [python3-catkin-pkg]
   fedora: [python3-catkin_pkg]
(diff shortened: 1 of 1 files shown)


cottsay (changes requested):
The wildcard syntax isn't necessary here - you should use the more condensed direct syntax:
```yaml
python3-can:
  debian: [python3-can]
  ubuntu: [python3-can]
```

Additionally, a [Fedora](https://apps.fedoraproject.org/packages/python3-can) package is available for this key as well:
```yaml
  fedora: [python3-can]
```

felixdivo:
Okay, sure. Thanks for the review. I addressed it in the last commit.

cottsay (changes requested):
Thanks for incorporating my feedback.

Looks like Travis is upset because the entries need to be alphabetized, then this LGTM.

felixdivo:
Done. :smile:

rosdep/python.yaml
```diff
@@ -5057,6 +5057,10 @@ python3-cairo:
     slackpkg:
       packages: [py3cairo]
   ubuntu: [python3-cairo]
+python3-can:
+  debian: [python3-can]
+  fedora: [python3-can]
+  ubuntu: [python3-can]
```
clalancette: Unfortunately, while this package is available on Ubuntu Bionic forwards, we also want to maintain backwards compatibility with Xenial.  So I think we'll need change back to the wildcard, but nulling out xenial, something like:

```
python3-can:
  debian: [python3-can]
  fedora: [python3-can]
  ubuntu:
    '*': [python3-can]
    xenial: null
```

clalancette (dismissed):

rosdep/python.yaml
felixdivo: Okay, did that.

cottsay (approved):
Thanks for iterating.

tfoote (approved):

PR #23316, opened 2019-12-16

mbaat:
I'd like pattern_manager to be indexed and documentet on ros.org

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9402,6 +9402,11 @@ repositories:
       type: git
       url: https://github.com/teknologisk-institut/pattern_manager.git
       version: master
+    source:
+      type: git
+      url: https://github.com/teknologisk-institut/pattern_manager.git
+      version: master
+    status: developed
   pcdfilter_pa:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
In reading your readme it mentions that you're missing dependencies to be able to run. Will they cause an issue with generating the documentation if they cannot be imported? If so we'll need to make sure to add that as a system dependency.  Looking at your usage, is the pluginlib actually helpful beyond the ability of entrypoints to declare find and load plugins? They're much more common. Pluginlib looks like a very young project and it's also a name conflict with a c++ library in ROS. 

Also I noticed that you don't have a top level LICENSE file that's standard for the Apache-2.0 license.

mbaat:
I have added a LICENSE and NOTICE file in accordance with Apache licensing standards. The pluginlib dependency has been removed from the README and plugins will be reimplemented using standard python libraries instead.

mbaat:
Sorry for pushing for a re-review but the pull request is sort of an urgent matter. Thanks in advance

PR #23346, opened 2019-12-20
Add fmt to base.yaml

kenji-miyake:
## Description
I would like to add [fmtlib/fmt](https://github.com/fmtlib/fmt) package to base.yaml.
It is an efficient formatting library for C++, and will be C++20 std::format.

arch: https://www.archlinux.org/packages/extra/x86_64/fmt/
debian: https://packages.debian.org/buster/libfmt-dev
fedora: https://fedora.pkgs.org/31/fedora-x86_64/fmt-devel-5.3.0-2.fc31.x86_64.rpm.html
gentoo: https://packages.gentoo.org/packages/dev-libs/libfmt
ubuntu: https://packages.ubuntu.com/bionic/libfmt-dev

We can implicitly use fmt if we depend on [rosfmt](https://github.com/xqms/rosfmt) which is a rosconsole extension using fmt.
However, I would like to directly depend on fmt when we don't use rosfmt features.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -806,6 +806,12 @@ fluid:
   macports: [fltk]
   opensuse: [fltk-devel]
   ubuntu: [fluid]
+fmt:
+  arch: [fmt]
+  debian: [libfmt-dev]
+  fedora: [fmt-devel]
+  gentoo: [dev-libs/libfmt]
+  ubuntu: [libfmt-dev]
 fping:
   arch: [fping]
   debian: [fping]


nuclearsandwich (approved):

tfoote:
@xqms Moving forward can we switch rosfmt to not be a wrapper but depend on the system installation? And will this collide with rosfmt which seems to be targeting older platforms.

xqms:
@tfoote Using the system libfmt would be nice, yes. However, fmt is not available on Ubuntu 16.04 and is outdated on Ubuntu 18.04 (version 4.0.0 vs. 6.0.0 as integrated in rosfmt). Between 4.0 and 6.0 many interesting features were added:

 * `std::chrono` formatting
 * colored output
 * locale-independent float formatting
 * and a large number of critical fixes.

I would definitely prefer to stay with a fixed internal version of fmt 6.0.0 for rosfmt.

So far, I haven't seen any conflicts with a system installation of fmt, so this PR should not cause problems with rosfmt as it is now.

tfoote:
Thanks for the update @xqms I'll merge this then if there won't be conflicts.

PR #23347, opened 2019-12-20
[popf] Change the name of the package

fmrico:
Signed-off-by: Francisco Martin Rico <fmrico@gmail.com>

I have just noticed that I uploaded my package as `popf-release` instead of `popf`. This PR fixes it.

I have just released a new version of this package to fix dependencies that made fail the built. Now it's version 0.011. Bloom failed in the last step to make the PR. It said that there was already a popf package. Could you help me with this? 🥺

I apologize for these errors. It's the first time that I release a package.

Best

PR #23354, opened 2019-12-21
New Version of popf 0.0.11-1

fmrico:
Signed-off-by: Francisco Martin Rico <fmrico@gmail.com>

Hi

This new release tries to fix the building of popf. The last version did not include the dependencies to build the source.

This version is 0.0.11-1 and the last one is 0.0.3-1. I am still not an expert with bloom, and the automated script increased the minor every time I tried to fix it. I hope that 0.0.11-1 will be ok. 

There is also a PR #23347 to fix the name of the package.

Best

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -1093,7 +1093,7 @@ repositories:
       url: https://github.com/ros2/poco_vendor.git
       version: eloquent
     status: maintained
-  popf-release:
+  popf:
     doc:
       type: git
       url: https://github.com/fmrico/popf.git
@@ -1104,7 +1104,7 @@ repositories:
       tags:
         release: release/eloquent/{package}/{version}
       url: https://github.com/fmrico/popf-release.git
-      version: 0.0.3-1
+      version: 0.0.11-1
     source:
       type: git
       url: https://github.com/fmrico/popf.git
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -1115,8 +1115,6 @@ repositories:
       url: https://github.com/fmrico/popf.git
       version: eloquent-devel
     release:
-      packages:
-      - popf
       tags:
         release: release/eloquent/{package}/{version}
       url: https://github.com/fmrico/popf-release.git


tfoote (changes requested):
With the repository rename lines 1102 and 1103 can be removed because if there's no package listed, it's assumed to be the same name as the repository. This is failing CI as in needs to be the canonical representation which is as simple as possible.

tfoote:
Higher version numbers are fine for us. If you get into semantic versioning you may want to control things. I don't know what automated script you're using. But catkin_prepare_release by default bumps the patch number. 

Though I'm surprised to see 0.0.5 in the source: https://github.com/fmrico/popf/blob/eloquent-devel/package.xml but 0.0.11 in the release repository. 

What process are you following? It's unexpected to have versions in the release repository higher than the source repository.  Oh, now I find it in your ros2 branch: https://github.com/fmrico/popf/blob/ros2/package.xml however your rosdistro content references the eloquent-devel branch.

You also appear not to have the 0.3 releases in your release repository anymore. Was that intentional?

fmrico:
Dear @tfoote 

I have removed the unnecessary lines.

I have also updated the brach `eloquent-devel` with the commits in the `ros2` branch at [https://github.com/fmrico/popf](https://github.com/fmrico/popf). Sorry for the inconsistency. Now both are in 0.0.11, consistent with changelog and package.xml. It is also consistent with https://github.com/fmrico/popf-release. The `master` branch will maintain the original work of the author, independent from ROS2.

About the 0.3 releases in your release repository, it was not intentional. As I commented, this is my first release, and trying to fix the errors, I made a mistake and I removed https://github.com/fmrico/popf-release.git, and I lost that release. I am not an expert in bloom yet, and some parts of the procedure were still dark for me.

I will maintain this package as a dependency for an imminent first release of [PlanSys2](https://github.com/IntelligentRoboticsLabs/ros2_planning_system).

Thanks. Best
Francisco.

clalancette (approved):
This looks OK to me now, but I'll let @tfoote re-review and merge it.

tfoote (approved):

tfoote:
Ok, yeah there's often a learning curve for making releases and unfortunately there's not easy way to practice offline. We try to ask a bunch of questions to make sure that we can be as helpful as possible in the reviews. Sometimes we have to ferret out invalid assumptions. It looks like you've gotten most of the way through now though.

PR #23363, opened 2019-12-25
Add python3-packaging rosdep

max-krichenbauer:
Packaging is part of the core utilities for Python packages and includes version handling, specifiers, markers, requirements, tags, utilities.

Homepage: https://github.com/pypa/packaging

Package references:

* Debian: https://packages.debian.org/sid/python3-packaging
* Fedora: https://src.fedoraproject.org/rpms/python-packaging
* Ubuntu: https://packages.ubuntu.com/bionic/python3-packaging

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2720,7 +2720,15 @@ python-oyaml-pip:
 python-packaging:
   debian: [python-packaging]
   fedora: [python-packaging]
-  ubuntu: [python-packaging]
+  ubuntu:
+    '*': [python-packaging]
+    artful_python3: [python3-packaging]
+    bionic_python3: [python3-packaging]
+    trusty_python3: [python3-packaging]
+    wily_python3: [python3-packaging]
+    xenial_python3: [python3-packaging]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2720,7 +2720,15 @@ python-oyaml-pip:
 python-packaging:
   debian: [python-packaging]
   fedora: [python-packaging]
-  ubuntu: [python-packaging]
+  ubuntu:
```
tfoote: These changes are unnecessary. This approach was tested and determined not to scale well. We've left the old rules in to not break anyone using the that pattern but do not create new ones. The old ones will eventually age out when the platforms go EOL.

tfoote (changes requested):

rosdep/python.yaml
max-krichenbauer: Thank you! I undid those changes and updated the PR.

tfoote (approved):
Thanks for the update. Looks good to me.

clalancette (approved):

PR #23370, opened 2019-12-26
add librealsense2 to melodic/distribution.yaml

tylerjw:
This adds librealsense2 to melodic.  If this fails to build due to cmake errors for missing packages I have a PR to fix that (https://github.com/IntelRealSense/librealsense/pull/5523).  Hopefully that can get merged soon.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4069,6 +4069,22 @@ repositories:
       url: https://github.com/ros-gbp/libg2o-release.git
       version: 2018.3.25-0
     status: maintained
+  librealsense2:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -4069,6 +4069,22 @@ repositories:
...
     status: maintained
+  librealsense2:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/IntelRealSense/librealsense2-release.git
```
tfoote: There's no melodic releases in the repository you cannot add a release section here without running bloom for melodic.

tfoote (changes requested):
This looks to be an attempt to supercede #22938 however it's failing CI due to a missing package.xml

tfoote:
Could you please propose an update to #22938 instead of just trying to replace it so we can keep the discussion in one place where we already have the maintainer's input.

PR #23381, opened 2019-12-29
rosdep: add python3-future

vooon:
On the next release of mavlink i'm going to do release for ROS2.
But in that case we need to use python 3, so I'll update package.xml to third format and add conditional depends on py3 version of `future`.

That PR introduces that new dependency.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5166,6 +5166,10 @@ python3-flake8:
     '*': ['python%{python3_pkgversion}-flake8']
     '7': null
   ubuntu: [python3-flake8]
+python3-future:
+  debian: [python3-future]
+  fedora: [python3-future]
+  ubuntu: [python3-future]
 python3-gitlab:
   debian:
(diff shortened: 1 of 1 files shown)


vooon:
Could you please check that new package.xml: https://github.com/mavlink/mavlink-gbp-release/blob/master/patch/package.xml

Works for Eloquent, but not sure about Kinetic and Melodic.

tfoote (approved):

tfoote:
I suspect that package.xml will work for ROS 1 too. Python's a core dependency, I suspect it's needed for both but will work with the conditional as it's already expected. Unless that dependency pull something that's not in the default developer tools.

vooon:
Regarding `<depend>python</depend>` i think that it's legacy from time when package also ships pymavlink. Now it's only required for build, then it's just a bunch of C and C++11 headers.

I checked new packages on Melodic - works fine.

mjcarroll (approved):

PR #23385, opened 2019-12-30
Adding roboclaw_ros to documentation index for melodic

alexmanning9904:


tfoote (changes requested):
The package.xml states the license is BSD, but neither of the source files contain a BSD header. And they also state:

```
 * This code is a modified version of the Ion Motion Control Arduino library.
 * To view the code in its original form, download it at
 *     http://downloads.ionmc.com/code/arduino.zip 
```

Which I checked does not appear to have a license embedded in it. Please provide clear documentation of the licensing of this code.

clalancette:
No response in over a month, so closing.  Please feel free to fix the issues pointed out in review and reopen.  Thanks for the contribution.

PR #23393, opened 2020-01-02
adding proaut-packages also to melodic

peterweissig:


tfoote:
It looks like you might have accidentally added those packages to the same branch as the radar ones in: https://github.com/ros/rosdistro/pull/23348 I think I already see that content having landed.

peterweissig:
thank you for the hint - I think your are right - sorry for the double request ...

PR #23400, opened 2020-01-03
use noetic branch of ros_environment

dirk-thomas:
See ros/ros_environment#22.


sloretz (approved):
LGTM pending ros/ros_environment#22

PR #23405, opened 2020-01-04
add opencl-headers package to rosdep

spuetz:
opencl headers are needed for OpenCl with CL2 used within c++

- https://packages.debian.org/de/sid/opencl-headers
- https://packages.ubuntu.com/search?keywords=opencl-headers
- https://src.fedoraproject.org/rpms/opencl-headers
- https://www.archlinux.org/packages/extra/any/opencl-headers/


tfoote (approved):
Thanks for all the links. It looks good to me.

clalancette (approved):

PR #23408, opened 2020-01-05
Add python-can-pip

felixdivo:
This adds the `python-can` packages as an explicit pip dependency. It already exists as an dependency for Python 2 (called `python-can`) and Python 3 (called `python3-can`). But we still need it from pip in our project, as we the version available in Ubuntu Bionic (18.04) is to old for our use case. We only require Python 3 btw. Is it valid to add such a pip-only *additional* dependency? What would be another way if it is not allowed?

link to the package: https://pypi.org/project/python-can/

The rationale for `python-can` is the same as in #23307.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -802,6 +802,25 @@ python-can:
     yakkety:
       pip:
         packages: [python-can]
+python-can-pip:
+  alpine:
+    pip:
+      packages: [python-can]
+  arch:
+    pip:
+      packages: [python-can]
+  debian:
+    pip:
+      packages: [python-can]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
We don't want to add a pip rule for something that's packaged in the native package manger. This will cause version mismatches with people who unintentially install the pip version which will overlay the system version. 

If you need it for your project you can setup your own rosdep rules to add them for your system. But we don't want to push this onto the community at large. We've seen it bite people badly.

felixdivo:
That's sad, but understandable.

PR #23414, opened 2020-01-06
Deduplicate yapf entry (fix for issue #23341)

### LINKED ISSUE #23341: Can't find dependency although existing in yaml-file.
On Ubuntu 16.04 I tried to install yapf via rosdep but rosdep can't find the package although it exists in [python.yaml](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml) (line 5754). Although it works if I change the order of the yaml files in `/etc/ros/rosdep/sources.list.d/20-default.list`. 
Please note that yapf also exits in [base.yaml](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml), but not for Ubuntu Xenial. I guess the search-algorithm for the package stops too early.

tylernewnoise:
This PR is intended to fix #23341.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5705,13 +5705,6 @@ yamllint:
     xenial: [yamllint]
     yakkety: [yamllint]
     zesty: [yamllint]
-yapf:
-  arch: [yapf]
-  debian:
-    buster: [yapf]
-    stretch: [yapf]
-  ubuntu:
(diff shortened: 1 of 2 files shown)


tfoote (approved):
Thanks for the followups. I also swapped it to use the wildcard syntax for better future compatibility.

clalancette (approved):

PR #23422, opened 2020-01-06
Add asr packages to melodic distro docu index

Tobi2001:
* Adding all ASR packages (documented here: http://wiki.ros.org/asr) which are currently in the docu index for kinetic to melodic
* Removing asr_approx_mvbb from kinetic docu index since it is no longer in a public repo


tfoote (changes requested):
Please verify all your documentation entry elements. The CI checking has found some that it appears not to find. 

Also please add a description to this PR summarizing the additions.

Tobi2001:
Updated PR due to requested changes

tfoote (approved):

PR #23423, opened 2020-01-07

doosan-robotics:
delete dsr_example_cpp package


tfoote (changes requested):
The title does not match the content. Please update this so we know what your intention is. 

Also please differentiate this from #23425 which looks to make this obsolete. If this is obsolete please close it so we don't spend time reviewing it.

clalancette:
I'm going to close this based on the last comment.  Please feel free to address those comments and resubmit if you'd like to make changes here.

PR #23429, opened 2020-01-07
add python-schedule and python3-schedule

knorth55:
add `python-schedule` and `python3-schedule` package in rosdep.

pypi: https://pypi.org/project/schedule/
github: https://github.com/dbader/schedule/
ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python-schedule&searchon=names
ubuntu (python3): https://packages.ubuntu.com/search?keywords=python3-schedule&searchon=names&suite=all&section=all
debian: https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=python-schedule
debian (python3): https://packages.debian.org/search?keywords=python3-schedule&searchon=names&suite=all&section=all

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3907,6 +3907,9 @@ python-scapy:
   fedora: [scapy]
   gentoo: [dev-python/scapy-python3]
   ubuntu: [python-scapy]
+python-schedule:
+  debian: [python-schedule]
+  ubuntu: [python-schedule]
 python-scipy:
   arch: [python2-scipy]
   debian: [python-scipy]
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
We typically prefer to use system packages if at all possible.  In this case, it looks like this is available as a system package on both [Ubuntu](https://packages.ubuntu.com/search?keywords=python3-schedule&searchon=names&suite=all&section=all) and [Debian](https://packages.debian.org/search?keywords=python3-schedule&searchon=names&suite=all&section=all) (though not Fedora).  Since that is the case, please switch this to be a key in [base.yaml](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml) instead.

knorth55:
@clalancette thank you for your comment. I didn't know that it is released in debia and ubuntu.
I will change this PR to add `python-schedule` in `base.yaml`

knorth55:
I update to add `python-schedule` and `python3-schedule` in `python.yaml`.

clalancette (approved):
Looks good to me!  Thanks for iterating.

tfoote (approved):

PR #23437, opened 2020-01-08
Feature/libgrpc

hakuturu583:
add libgrpc package to the rosdistro

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2083,6 +2083,11 @@ libgps:
   gentoo: [sci-geosciences/gpsd]
   opensuse: [gpsd-devel]
   ubuntu: [libgps-dev]
+libgrpc-dev:
+  arch: [grpc]
+  debian: [libgrpc-dev]
+  fedora: [grpc-devel]
+  ubuntu: [libgrpc-dev]
 libgsl:
   arch: [gsl]
   debian:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -189,13 +189,6 @@ jupyter-notebook:
 libgv-python:
   debian: [libgv-python]
   ubuntu: [libgv-python]
-mdrec:
-  debian:
-    pip:
-      packages: [mdrec]
-  ubuntu:
-    pip:
-      packages: [mdrec]
 mercurial:
   osx:
     pip:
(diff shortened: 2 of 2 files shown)


hakuturu583:
I want to make GRPC-ROS bridge.
I want to add these packages below.

debian package is here : https://packages.debian.org/sid/libgrpc-dev
ubuntu package is here : https://packages.ubuntu.com/ja/bionic/arm64/libgrpc-dev
arch package is here : https://www.archlinux.org/packages/community/x86_64/grpc/
fedora package is here : https://apps.fedoraproject.org/packages/grpc-devel

hakuturu583:
Oh ! I found it includes another commit, so I remove this change.(https://github.com/ros/rosdistro/pull/23437/commits/63e86b75f61428c47ec54b396e81583bc5a3b012)

hakuturu583:
I removed mdrec
Please check it from file changed. (https://github.com/ros/rosdistro/pull/23437/files)

tfoote (approved):
Thanks for the PR and adding the links for reference. 

A more generic ubuntu link for review is here: https://packages.ubuntu.com/bionic/libgrpc-dev

clalancette (changes requested):
The one slight problem here is that the `libgrpc-dev` package is only available on Debian for Buster, Sid, and Stretch-backports.  We don't usually allow keys to depend on the debian "backports" repository.  Since stretch is still supported, I think we need to split the debian key up so that it has a 'null' for stretch, and a '*' for everything else.

nuclearsandwich:
@hakuturu583 are you able to make the changes requested by @clalancette?

rosdep/base.yaml
```diff
@@ -2083,6 +2083,11 @@ libgps:
   gentoo: [sci-geosciences/gpsd]
   opensuse: [gpsd-devel]
   ubuntu: [libgps-dev]
+libgrpc-dev:
+  arch: [grpc]
+  debian: [libgrpc-dev]
```
tfoote: Filling in Chris' suggestion more concretely.
```suggestion
  debian:
    '*': [libgrpc-dev]
    stretch: null
```

clalancette:
No response from the original submitter for over a month, so I'm closing this.  @hakuturu583 , if you'd like to see this merged, please fix things according to @tfoote 's suggestion and re-open.

PR #23461, opened 2020-01-09
add librealsense2 package

doronhi:
This PR should replace #22938 
It incorporates flags for building without X11 dependency and maintains the naming convention of the Kinetic distribution.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4095,8 +4095,7 @@ repositories:
       type: git
       url: https://github.com/IntelRealSense/librealsense.git
       version: master
-    status: maintained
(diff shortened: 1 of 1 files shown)


doronhi:
@clalancette , if, by any chance, you could point me to the build log of "http://build.ros.org/" it would help me a lot. I could fix build issues and speed things up. I just can't seem to find it.

tfoote:
Because you're adding the package here in this PR you won't find any builds on build.ros.org until after this PR has been merged and the buildfarm has been reconfigured. You can find them by searching like this now: http://build.ros.org/search/?q=realsense

PR #23470, opened 2020-01-10
Add openai gym to rosdep/python.yaml

gerd2s:


tfoote (changes requested):
```
For convenience in reviewing, please comment in the PR with links to the web listings of packages such as on http://packages.ubuntu.com, http://packages.debian.org, and https://apps.fedoraproject.org/packages or if a pip package pypi.python.org.

Please also briefly describe the package being added and what use case you want to use it for. It's valuable to have a record of the packages as submitted and their intended purpose for clarity in the future so that if there's a conflict there's information to fall back on instead of speculation about the original use cases.
```

gerd2s:
[openai pip package](https://pypi.org/project/gym/) 

From the [openaigym website](https://gym.openai.com/): A toolkit for developing and comparing reinforcement learning algorithms. It supports teaching agents everything from walking to playing games like Pong or Pinball.

Use case is training of software agents for robotic applications.

clalancette (approved):

tfoote (approved):
Thanks for the background. Since there doesn't appear to be any debian packages we can add the pip ones for now.

PR #23231, opened 2019-12-05
Revert octomap back to 1.9.0

clalancette:
Some downstream packages are failing to find `liboctomap.so` since we upgraded beyond 1.9.0.  One example is http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__mrpt1__ubuntu_bionic_amd64__binary/5/console.  This PR is an experiment to go back to an older octomap to see if that fixes the problem.  @wxmerkt FYI.

PR #23475, opened 2020-01-10
Revert "octomap: 1.9.3-1 in 'melodic/distribution.yaml' [bloom] (#233…

clalancette:
…66)"

This reverts commit 0773b60f73f90012758a0098327028effab48226.

@wxmerkt @jlblancoc This is still causing issues between octomap and mrpt.  I'd like to do a sync, so we can resolve this in three ways:

1.  Going back again to octomap 1.9.0 (this PR).
2.  Dropping mrpt1 from Melodic.
3.  Fixing whatever the underlying problem is (though I won't have time to look into it, and I want to do a Melodic sync next week).

In the absence of any additional information, I'm going to go with option 1 (this PR), but feel free to comment here about what you want to do.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -5472,7 +5472,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/ros-gbp/octomap-release.git
-      version: 1.9.3-1
+      version: 1.9.0-1
     source:
       type: git
       url: https://github.com/OctoMap/octomap.git


wxmerkt:
Hi @clalancette,
I can try locally to compile mrpt1 with the new octomap to gain an understanding of what's going on. Do you have a link to an output of the problem?

Thanks,
Wolfgang

clalancette:
Sure.  It's the same problem as before, where mrpt1 can't find liboctomap and liboctomath to link to: http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__mrpt1__ubuntu_bionic_amd64__binary/30/consoleFull

clalancette:
Thanks for taking a look, its appreciated.

For now, I'm going to merge this revert.  Assuming it fixes the problem (it did last time), we can hopefully try again to fix it for real for the next sync.

tfoote:
The first failure was clearly triggered by the octomap release: http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__mrpt1__ubuntu_bionic_amd64__binary/20/

```
Started by upstream project Mbin_uB64__octomap__ubuntu_bionic_amd64__binary build number 7
originally caused by:

Started by upstream project Msrc_uB__octomap__ubuntu_bionic__source build number 6
originally caused by:
Started by upstream project Mrel_trigger-jobs build number 19834
originally caused by:
Started by timer
```

Looking at the error of the downstream builds it appears that they can't link aginst `liboctomath.so.1.9`

```
dpkg-shlibdeps: error: cannot find library liboctomap.so.1.9 needed by debian/ros-melodic-mrpt1/opt/ros/melodic/lib/libmrpt-maps.so.1.5.9 (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '')
dpkg-shlibdeps: error: cannot find library liboctomath.so.1.9 needed by debian/ros-melodic-mrpt1/opt/ros/melodic/lib/libmrpt-maps.so.1.5.9 (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '')
dpkg-shlibdeps: error: cannot continue due to the errors listed above
Note: libraries are not searched in other binary packages that do not have any shlibs or symbols file.
To help dpkg-shlibdeps find private libraries, you might need to use -l.
```

Looking into the changes that you highlighted I see one related changes: 

In particular it looks like the install rule was changed for liboctomath: https://github.com/OctoMap/octomap/compare/v1.9.0...v1.9.3#diff-8ee8aeb2c137d7e24ee66da018702527L13 



From: https://github.com/OctoMap/octomap/pull/262

The [mrpt1 cmake script](https://github.com/MRPT/mrpt/blob/develop/cmakemodules/script_octomap.cmake) appears to have found the built in version. 
`--  Octomap                              : Yes (System) [Version: 1.9.3]`

jlblancoc:
Hi guys, 

I'm struggling to understand *why* it fails as it does... cmake, build and link all work fine. 
It's only at `dpkg-shlibdeps` time when it cannot find the octomap libraries... 
Do we have other packages that depend on octomap? Just for checking if they do something different... Supposedly, ROS's `LD_LIBRARY_PATH` should be set by the time `dpkg-shlibdeps` is invoked so if the octomap package is installed (and it must be because everything builds and links!) I cannot understand why `dpkg-shlibdeps` fails... :-S

wxmerkt:
@tfoote Thank you for analysing. The commit you reference (https://github.com/OctoMap/octomap/commit/31af07e5bacc210cbf3c32929acf50338c108123) was merged on **December 8th** and is part of v1.9.3 which is presently failing on the buildfarm. However, it was not part of v1.9.2. which was reverted **December 5th** (https://github.com/ros/rosdistro/pull/23231) - this version already had the issues with MRPT1. So it's likely not the culprit here :-(.

PR #23497, opened 2020-01-13
add key for python-fcl-pip

furushchev:
Added key for Python Interface for the Flexible Collision Library

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1438,6 +1438,23 @@ python-fastdtw-pip:
   ubuntu:
     pip:
       packages: [python-fastdtw]
+python-fcl-pip:
+  debian:
+    pip:
+      depends: [libfcl-dev]
+      packages: [python-fcl]
+  fedora:
+    pip:
+      depends: [libfcl-dev]
+      packages: [python-fcl]
+  osx:
+    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1438,6 +1438,23 @@ python-fastdtw-pip:
...
+    pip:
+      depends: [libfcl-dev]
+      packages: [python-fcl]
+  fedora:
+    pip:
+      depends: [fcl]
+      packages: [python-fcl]
+  osx:
+    pip:
+      depends: [dartsim/dart/fcl]
```
sloretz: I'm not familiar with `depends`, but the source code of `rosdep2` seems to [suggest it should reference other rosdep keys](https://github.com/ros-infrastructure/rosdep/blob/edb89c4de2912e9895f26273573ecf9267b68cc1/src/rosdep2/installers.py#L396-L404). If so, it seems like this [should be `libfcl-dev` for all platforms](https://github.com/ros/rosdistro/blob/31d8f07d20179f5b7a14e2bedaec23163af80bd4/rosdep/base.yaml#L1832)

furushchev: @sloretz Thank you for feedback! True. I updated the change.

clalancette (approved):

PR #23506, opened 2020-01-13
Adding kvh_geo_fog_3d to documentation index for distro

zlacelle:
Requesting merge of MITRE kvh_geo_fog_3d set of packages to rosdistro for kinetic.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6221,16 +6221,6 @@ repositories:
       type: git
       url: https://github.com/MITRE/kvh_geo_fog_3d.git
       version: master
-    release:
-      packages:
-      - kvh_geo_fog_3d
-      - kvh_geo_fog_3d_driver
-      - kvh_geo_fog_3d_msgs
-      - kvh_geo_fog_3d_rviz
-      tags:
-        release: release/kinetic/{package}/{version}
-      url: https://github.com/MITRE/kvh_geo_fog_3d-release.git
-      version: 1.3.0-1
     source:
       type: git
       url: https://github.com/MITRE/kvh_geo_fog_3d.git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6224,7 +6224,7 @@ repositories:
     source:
       type: git
       url: https://github.com/MITRE/kvh_geo_fog_3d.git
-      version: master
+      version: kinetic-devel
     status: maintained
   laser_assembler:
     doc:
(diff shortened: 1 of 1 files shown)


zlacelle:
The travis-ci is failing when trying to find package.xml for each of the packages within this repository. I'm not sure why this is happening, as the package.xml are present.

kinetic/distribution.yaml
```diff
@@ -6216,6 +6216,26 @@ repositories:
       url: https://github.com/ros-drivers/kvh_drivers.git
       version: master
     status: maintained
+  kvh_geo_fog_3d:
+    doc:
+      type: git
+      url: https://github.com/MITRE/kvh_geo_fog_3d.git
+      version: master
+    release:
```
sloretz: I get a 404 error when I try to look at the release repo https://github.com/MITRE/kvh_geo_fog_3d-release.git (is it a private repo?). If all that's desired is to generate documentation (and not generate debian packages) then the `release` section should be deleted.

tfoote: @zlacelle This is why it can't pull the package.xml

If you have not run bloom there should be no release stanza and you'll just have the `source` and `doc` stanzas for building from source.

zlacelle:
@tfoote @sloretz Yes, you're right. My mistake. I'll modify the pull request to remove the -release repo until I get it moved over from our internal environment.

kinetic/distribution.yaml
```diff
@@ -6216,6 +6216,16 @@ repositories:
       url: https://github.com/ros-drivers/kvh_drivers.git
       version: master
     status: maintained
+  kvh_geo_fog_3d:
+    doc:
+      type: git
+      url: https://github.com/MITRE/kvh_geo_fog_3d.git
+      version: master
```
sloretz: Just checking, the repo has a `kinetic-devel` branch. Is `master` correct, or should this be `kinetic-devel`?

zlacelle: I envisioned that the documentation will live in master, while kinetic-devel, melodic-devel, etc would track code releases. Is this a generally acceptable way of tracking ROS repos? The documentation will remain consistent across ROS distros.

sloretz: Sounds find to me. You may want the `source` branch to point to `kinetic-devel` though, since that's what users will get if they do `rosinstall_generator ... --upstream-devel`

zlacelle: Indeed. Corrected.

sloretz (approved):

tfoote (approved):
Thanks for iterating.

PR #23513, opened 2020-01-14
mavlink: 2020.1.14-1 in 'eloquent/distribution.yaml' [bloom]

vooon:
Increasing version of package(s) in repository `mavlink` to `2020.1.14-1`:

- bloom version: `0.9.0`
- previous version for package: `2019.12.30-1`

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -920,7 +920,7 @@ repositories:
       tags:
         release: release/eloquent/{package}/{version}
       url: https://github.com/mavlink/mavlink-gbp-release.git
-      version: 2019.12.30-1
+      version: 2020.1.14-1
     source:
       type: git
       url: https://github.com/mavlink/mavlink-gbp-release.git


vooon:
There is a problem building binary deb:
http://build.ros2.org/view/Ebin_uB64/job/Ebin_uB64__mavlink__ubuntu_bionic_amd64__binary/12/consoleFull
```
23:03:36 dh_auto_configure -- -DCMAKE_INSTALL_PREFIX="/opt/ros/eloquent"
23:03:36 	install -d obj-x86_64-linux-gnu
23:03:36 	cd obj-x86_64-linux-gnu && cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_INSTALL_PREFIX=/opt/ros/eloquent
23:03:36 CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
23:03:36   Could NOT find PythonInterp (missing: PYTHON_EXECUTABLE) (Required is at
23:03:36   least version "2")
23:03:36 Call Stack (most recent call first):
23:03:36   /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
23:03:36   /usr/share/cmake-3.10/Modules/FindPythonInterp.cmake:152 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
23:03:36   CMakeLists.txt:30 (find_package)
23:03:36 
23:03:36 
23:03:36 -- Configuring incomplete, errors occurred!
```

Seems that `ROS_PYTHON_VERSION` is unset, but [CmakeLists.txt L20][cmake] rely on it to select proper python version for ROS1 and ROS2.

Package.xml changed to include `python3-dev` as build depend when `ROS_PYTHON_VERSION == 3`.

[cmake]: https://github.com/mavlink/mavlink-gbp-release/blob/master/patch/CMakeLists.txt#L20

mjcarroll:
Two notes:

1. All ROS2 distros are Python3, so it would probably be safe to update your patch to expect that.
2. You should be able to surface that variable by depending on https://github.com/ros2/ros_workspace

nuclearsandwich:
>     2\. You should be able to surface that variable by depending on https://github.com/ros2/ros_workspace

I think this should be ros_environment, ros_workspace is implicitly added to every bloomed ROS 2 release.

mjcarroll:
> I think this should be ros_environment, ros_workspace is implicitly added to every bloomed ROS 2 release.

Transcription error, thanks for catching that.

vooon:
> All ROS2 distros are Python3, so it would probably be safe to update your patch to expect that.

I'd like to keep same patch for both ROS versions because i still will release that package monthly.
If i set py3 for default that likely break Kinetic and Melodic builds.

Is there any method to detect ROS2 from cmake? Present code with ROS_PYTHON_VERSION works ok for local builds in docker.
Also please note that this package is a plain cmake.

nuclearsandwich:
> Is there any method to detect ROS2 from cmake? Present code with ROS_PYTHON_VERSION works ok for local builds in docker.

The ros_environment package provides hooks for a number of ROS-related environment variables ([example source for ROS 2 Foxy](https://github.com/ros/ros_environment/tree/foxy/env-hooks)). CMake should be able to use the environment variables provided by that package to check ROS_PYTHON_VERSION and ROS_VERSION. The former should be used for deciding whether to use Python 2 or 3 since ROS Noetic will use Python 3 by default but for other cases where you are checking for ROS 1 or ROS 2 the ROS_VERSION variable will tell you.

As long as you don't mind depending on an ament package and sourcing the appropriate workspace files (which the colcon build tool can manage for you) even plain CMake packages will be able to use the environment variables set by ros_environment.

vooon:
@nuclearsandwich i added buildtool_depend on ament_cmake.
New PR will be soon.

nuclearsandwich:
> @nuclearsandwich i added buildtool_depend on ament_cmake.
> New PR will be soon.

I'm not sure that the buildtool_depend on ament_cmake will help you. What I think you want is a buildtool_depend on ros_environment. ament_cmake is orthogonal to ROS and doesn't rely on ros_environment itself.

vooon:
@nuclearsandwich it's a bit unclear what to do for 3-rd party libs in ros2, for ros1 as i remember i just done step-by-step with the tutorial. For ROS1 as i remember you just need to add depend in catkin but define build type as cmake.

So i done similar steps for #23527.
Also my updated ntpd-driver just depends on ament_cmake, and not on ros_environment, and all ok (i used `ros2 pkg create` to get fresh xml and cmake).

PR #23525, opened 2020-01-16
Add python3-kitchen rosdep key

sloretz:
* https://www.archlinux.org/packages/community/any/python-kitchen/ 
* https://packages.debian.org/buster/python3-kitchen
* https://apps.fedoraproject.org/packages/python3-kitchen
* https://packages.gentoo.org/packages/dev-python/kitchen
* https://pypi.org/project/kitchen/
* https://packages.ubuntu.com/bionic/python3-kitchen

Used by [rosdoc_lite](https://github.com/ros-infrastructure/rosdoc_lite/blob/ec884e5f17f962698c653453ece65ffed03fb21d/package.xml#L25)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5248,7 +5248,9 @@ python3-kitchen:
   osx:
     pip:
       packages: [kitchen]
-  ubuntu: [python3-kitchen]
+  ubuntu:
+    '*': [python3-kitchen]
+    xenial: null
 python3-lark-parser:
   debian:
     buster: [python3-lark-parser]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5240,6 +5240,15 @@ python3-imageio:
   debian: [python3-imageio]
   gentoo: [dev-python/imageio]
   ubuntu: [python3-imageio]
+python3-kitchen:
+  arch: [python-kitchen]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5240,6 +5240,15 @@ python3-imageio:
...
   ubuntu: [python3-imageio]
+python3-kitchen:
+  arch: [python-kitchen]
+  debian: [python3-kitchen]
+  fedora: [python3-kitchen]
+  gentoo: [dev-python/kitchen]
+  osx:
+    pip:
+      packages: [kitchen]
+  ubuntu: [python3-kitchen]
```
clalancette: The one problem with this is that `python3-kitchen` doesn't seem to be available on xenial, which is still a supported platform.  I'd suggest using the "null" syntax for xenial, like https://github.com/ros/rosdistro/blob/06a8165d68496e3d7c21da497cab4174d5c20fc1/rosdep/python.yaml#L4699

clalancette (changes requested):

rosdep/python.yaml
sloretz: Null syntax for xenial in 0e5a2f63b556510a86fdd81230a93562b8f31856

clalancette (approved):

PR #23589, opened 2020-01-24
Revert "sbg_driver: 2.0.0-1 in 'melodic/distribution.yaml' [bloom]"

tfoote:
Reverts ros/rosdistro#23413

This also fails to build on melodic, kinetic revert at: #23585 it appears to have the same issue in melodic.

Upstream issue: https://github.com/SBG-Systems/sbg_ros_driver/issues/28


clalancette:
Closing since I merged #23596 instead.

PR #23606, opened 2020-01-25
noetic branch for roscpp_core

dirk-thomas:


noetic/distribution.yaml
```diff
@@ -372,7 +372,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/ros/roscpp_core.git
-      version: kinetic-devel
+      version: noetic-devel
```
sloretz: It looks like `kinetic-devel` and `noetic-devel` are on the same commit. Is that intentional?

dirk-thomas: Yes, because ros/roscpp_core#104 will be merged into Noetic only.

PR #23622, opened 2020-01-26
add libboost-dev key

mikaelarguedas:
I'm not sure about the fedora rule.
I didnt find a package with just the boost headers so I fell back to the `boost-devel` package that contains all the headers and all the libraries


gavanderhoorn:
- https://packages.ubuntu.com/xenial/libboost-dev
 - https://packages.debian.org/buster/libboost-dev
 - https://apps.fedoraproject.org/packages/boost-devel

tfoote (approved):

mikaelarguedas:
Please consider merging without squashing to not conflict with https://github.com/ros/rosdistro/pull/23624

nuclearsandwich (approved):

PR #23627, opened 2020-01-27
Fixing branch associated with astuff_sensor_msgs on dashing.

JWhitleyWork:
Somehow `bloom` knows which branch it should be on but the PR that was produced for the latest release didn't update the branch. This fixes that omission.

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -272,7 +272,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/astuff/astuff_sensor_msgs.git
-      version: dashing-devel
+      version: master
     release:
       packages:
       - astuff_sensor_msgs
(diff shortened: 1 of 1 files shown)


JWhitleyAStuff:
Related PR is #23591.

nuclearsandwich:
> Somehow `bloom` knows which branch it should be

Bloom stores the development branch for each release track in tracks.yaml and uses it in some situations to fetch the latest tag on that branch for release. https://github.com/astuff/astuff_sensor_msgs-release/blob/master/tracks.yaml#L15

>  but the PR that was produced for the latest release didn't update the branch.

Bloom might use the devel branch as the default version when first populating a source entry but if a repository in the rosdistro already has a source or doc entry bloom won't touch those when updating releases. I think having bloom check for consistency between the source/doc branch and the track devel branch would be a reasonable enhancement if you'd like to contribute a PR.

JWhitleyAStuff:
Thanks for the detailed explanation! I will put that on my list. ;)

PR #23635, opened 2020-01-28
remove cpswarm_msgs and dependencies

tfoote:
It's failing to build since the removal of swarmros in #23586


tfoote:
https://github.com/amilankovich-slab/swarmros/issues/1

PR #23639, opened 2020-01-28
Add rosdep entry for Carla

vimalrajliya:
[carla_ros_bridge](https://github.com/carla-simulator/ros-bridge) ROS package aims at providing a simple ROS bridge for the [CARLA simulator](http://carla.org/l).
https://github.com/ros/rosdistro/pull/23173
To resolve the dependency on Carla, I would like to add Carla in rosdep.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -339,6 +339,10 @@ can-utils:
   debian: [can-utils]
   fedora: [can-utils]
   ubuntu: [can-utils]
+carla:
+  ubuntu:
+    source:
+      uri: https://raw.githubusercontent.com/carla-simulator/ros-bridge/0.9.6/carla_ros_bridge.rdmanifest
 cccc:
   debian: [cccc]
   gentoo: [dev-util/cccc]


tfoote (changes requested):
Unfortunately we don't accept rdmanifests into the rosdep database. They can be used locally by developers, but they are too fragile and insecure to be recommended for sharing.

https://answers.ros.org/question/255238/what-is-the-status-of-rdmanifests/

It looks like there's a pypi project for it already: https://pypi.org/project/carla/
And some related discussions here: https://pypi.org/project/carla/

A ROS wrapper package for carla might make sense if there's not plans for packaging carla into existing pipelines like the debian/ubuntu packages. Though I did see comments about 2GB of files which might be excessive.

vimalrajliya:
Carla doesn't have current plans to release pip installable for the latest release of Carla.

We do have a Debian package for Carla's latest release but because of dependency on Unreal Engine, it is not possible to release it from Ubuntu PPA.

Carla is planning to have personal apt-repo for Debian package but to install from it, we need to import the gpg key and I am not sure how to add that in Rosdep.

So in my limited understanding, we have to options to resolve the dependency on Carla :
1. Install Carla adding apt-repo but that is not supported according to this answer. 
https://answers.ros.org/question/340470/how-to-depend-on-apt-install-that-requires-a-source-list-file/
2. The only option we have is to install Carla from the source which is adding a 'rdmanifest' file.

If you can suggest some alternative or merge this request as the last resort, it would be a great help and addition to the community.

tfoote:
The rosdep database that we maintain is designed to represent the available packages for users from our repositories and the default upstream repositories. If Carla won't be in those then we should not add it to our database.

We've setup the tools to be extended exactly for this use case. If Carla will have it's own apt repo. My recommendation would be to create your own set of rosdep rules and host them in/near the apt repository. And then the instructions for your users to add the apt repository source with GPG key as well as add the a new rosdep source for your set of rosdep rules which cover the contents of the custom apt repository. This way when the apt source is added the rosdep keys will resolve for those users. For other users the keys will not resolve because they're not available.

As an aside, if you cannot host the binary packages upstream due to license encumbrances, automating building from source just hides that from the unsuspecting user which we don't want to do as well as causing them a lot of effort to compile it themselves when it would be better that they should actively accept the license and use the binaries.

clalancette:
I'm going to close this out since I don't currently see a way forward with this.  But feel free to keep commenting or re-open if you disagree.

PR #23647, opened 2020-01-29
SwarmROS updated for building

amilankovich-slab:
[Previously ](https://github.com/ros/rosdistro/pull/23586)SwarmROS did not build successfully, as some dependencies were mission. this is an attempt to fix that issue.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -17295,8 +17295,8 @@ repositories:
     status: maintained
   swarmros:
     doc:
-      type: svn
-      url: https://github.com/cpswarm/swarmio/trunk/master/swarmros
-      version: master
+      type: git
+      url: https://github.com/amilankovich-slab/swarmros
+      version: 0.3.1
 type: distribution
 version: 2
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -15025,6 +15025,10 @@ repositories:
       version: kinetic-devel
     status: developed
   swarmros:
+    doc:
+      type: git
+      url: https://github.com/amilankovich-slab/swarmros.git
+      version: master
     release:
       tags:
         release: release/kinetic/{package}/{version}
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -17572,5 +17573,10 @@ repositories:
       url: https://github.com/zivid/zivid-ros.git
       version: master
     status: maintained
+  swarmros:
```
tfoote: This section should be a subsection of the already existing swarmros entry.

kinetic/distribution.yaml
```diff
@@ -17572,5 +17573,10 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/amilankovich-slab/swarmros-release.git
```
tfoote: The doc url should be a branch (preferred, but tag is ok) on the upstream repository. Usually it's the same as the source entry.

tfoote (changes requested):
I noticed that the package.xml [shows 0.9.3 as the version](https://github.com/amilankovich-slab/swarmros/blob/86179063f45df659f1b5559adb80c17cbad46823/package.xml#L6)  but this PR is for 0.3.1

kinetic/distribution.yaml
```diff
@@ -15029,13 +15029,18 @@ repositories:
...
       test_pull_requests: true
       type: git
       url: https://github.com/amilankovich-slab/swarmros.git
       version: master
     status: developed
     status_description: developed
+    doc:
+      type: git
+      url: https://github.com/amilankovich-slab/swarmros.git
+      version: 0.3.1-4
```
tfoote: This version needs to be a branch (or tag) in the upstream repo. Usually it's set to the same branch as the source entry above so `master`.
```suggestion
      version: master
```

tfoote (approved):
Thanks for the updates. this looks good to me now.

tfoote:
Sorry one last fixup is that the doc block needs to be at the top due to alphabetical ordering constraints.

clalancette:
Looks good now, but holding for [sync](https://discourse.ros.org/t/preparing-for-kinetic-sync-2020-02-07/12677)

PR #23656, opened 2020-01-30
add libogre package

simonschmeisser:
it is copied from libogre-dev for most distro except current Ubuntu and Debian releases. 
The plan is to reduce installation size by splitting dependencies in build and exec_depend


tfoote (changes requested):
It looks like libogre-1.9.0v5 is valid for everything newer than xenial. I'd suggest we use a wildcard for that and leave the older rules for backwards compatibility. 

1.9 is likely to exist for a long time in parallel with an upcoming 2.0 series.

https://packages.ubuntu.com/xenial/libs/libogre-1.9.0v5

simonschmeisser:
1.12.4 should hit debian ftp this weekend and subsequently get into ubuntu 20.04 if I didn't have any major issues remaining

tfoote (approved):
Ok, then lets go with this for now then.

clalancette (approved):

PR #23682, opened 2020-02-04
Add libkml and python-kml rosdep keys

pjreed:
libkml is Google's reference implementation of OGC KML 2.2.

Home page: https://github.com/libkml/libkml

- Debian: https://packages.debian.org/source/buster/libkml
- Ubuntu: https://packages.ubuntu.com/source/bionic/libkml

I was unable to find Gentoo or Fedora packages, unfortunately.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2386,6 +2386,10 @@ libkdtree++-dev:
   debian: [libkdtree++-dev]
   fedora: [libkdtree++-devel]
   ubuntu: [libkdtree++-dev]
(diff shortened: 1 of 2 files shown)


nuclearsandwich (approved):
The only thing I'm unsure about is adding the Archlinux key for a package known to be in in AUR instead of the official repositories.

Unless you're using the Archlinux key I'd prefer if we omitted it until explicit AUR handling is part of rosdep.

pjreed:
I'm not using it, so that's fine by me.

PR #23692, opened 2020-02-06
Add Connext 5.3.1 rosdep rule for Ubuntu Focal

jacobperron:
Soon to be available via the ROS repositories

cc/ @nuclearsandwich


wjwwood (approved):
lgtm once we confirm (test) it is there.

nuclearsandwich (approved):

PR #23706, opened 2020-02-07
Add pcg-gazebo to rosdep list

musamarcusso:
`pcg-gazebo` is a Python package for procedural generation for Gazebo that has been published as a `pip` package recently. 

PyPi: https://pypi.org/project/pcg-gazebo/

It includes three dependencies, two of which can already be found in `base.yaml` (`geos` and `pybind11-dev`). It was necessary to add the third one (`libspatialindex-dev`) to `base.yaml`.

- Debian: https://packages.debian.org/buster/libspatialindex-dev
- Ubuntu: https://packages.ubuntu.com/bionic/libspatialindex-dev

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -233,15 +233,6 @@ paramiko:
       packages: [paramiko]
   rhel: [python-paramiko]
   ubuntu: [python-paramiko]
-pcg-gazebo-pip:
-  debian:
-    pip:
-      depends: [libspatialindex-dev, geos, pybind11-dev]
-      packages: [pcg-gazebo]
-  ubuntu:
-    pip:
-      depends: [libspatialindex-dev, geos, pybind11-dev]
-      packages: [pcg-gazebo]
 pi-ina219-pip:
   debian:
     pip:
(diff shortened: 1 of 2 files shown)


clalancette (approved):

rosdep/python.yaml
```diff
@@ -233,6 +233,15 @@ paramiko:
       packages: [paramiko]
   rhel: [python-paramiko]
   ubuntu: [python-paramiko]
+pcg-gazebo-pip:
```
nuclearsandwich: Could you please rename (and therefore relocate) this package to `python3-pcg-gazebo-pip` to conform to the upcoming contribution guidelines in #23836.

nuclearsandwich (changes requested):
@musamarcusso we're likely updating our contribution guidelines for pip packages (#23836) and I'd be grateful if you would update your PR to follow the upcoming change even though it is not merged yet.

rosdep/python.yaml
musamarcusso: Hello @nuclearsandwich, thanks for the tip, I had not seen that pull request. I added the tags for both 2.x and 3.x versions since it is available for both. Is the duplication OK? Python 2.7 is EOL but I don't know how to set these tags any other way for systems that still are using 2.7.

nuclearsandwich: I was about to speak out against duplicate keys but @tfoote is in favor based on https://github.com/ros/rosdistro/pull/23836#discussion_r382871821

nuclearsandwich (approved):
Thanks for iterating.

PR #23761, opened 2020-02-13
Add information about rewriting _python3 suffix

kyrofa:
These should be rewritten as top-level rules prefixed with `python3-`.


dirk-thomas:
Thank you!

PR #23771, opened 2020-02-14
ColorTuner: 0.0.3-2 in 'melodic/distribution.yaml' [bloom]

pankhurivanjani:
Increasing version of package(s) in repository `ColorTuner` to `0.0.3-2`:

- upstream repository: https://github.com/JdeRobot/ColorTuner.git
- release repository: https://github.com/JdeRobot/ColorTuner-release.git
- bloom version: `0.9.1`

## jderobot_color_tuner

- No changes

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -8,6 +8,19 @@ release_platforms:
   ubuntu:
   - bionic
 repositories:
+  ColorTuner:
+    release:
+      packages:
+      - jderobot_color_tuner
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/JdeRobot/ColorTuner-release.git
+      version: 0.0.3-2
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/JdeRobot/ColorTuner.git
+      version: master
   abb:
     release:
       packages:


melodic/distribution.yaml
```diff
@@ -8,6 +8,19 @@ release_platforms:
   ubuntu:
   - bionic
 repositories:
+  ColorTuner:
```
clalancette: Can you please rename this to `jderobot_color_tuner`?  That will make it alphabetize with the rest of the jderobot stuff and be easier to see in the future.

clalancette (changes requested):

clalancette:
I also notice that there is no LICENSE file in the source repository.  It should have a LICENSE to make clear what the license of the code is.

You also have no dependencies listed in your package.xml, so this is unlikely to work for users that install a binary package.

Finally, I also noticed that you also have very large binary files checked into the repository that seem to be pre-compiled python artifacts.  That's not very nice to people consuming this; they'll need to fetch megabytes of stuff every time they clone the repository.  I'll suggest that you remove these artifacts and then rewrite your git history to remove them from ever existing.

There are enough problems here that this is almost certainly going to require another PR.  So I'm actually going to close this one out; feel free to address my points above (and/or ask questions) before opening a new PR.

PR #23776, opened 2020-02-14
ifm3d-release: 0.17.0-9 in 'kinetic/distribution.yaml' [bloom]

theseankelly:
Increasing version of package(s) in repository `ifm3d-release` to `0.17.0-9`:

- upstream repository: https://github.com/ifm/ifm3d
- release repository: https://github.com/ifm/ifm3d-release.git
- bloom version: `0.9.1`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5039,6 +5039,7 @@ repositories:
       url: https://github.com/ifm/ifm3d-release.git
       version: 0.17.0-9
     source:
+      test_commits: false
       type: git
       url: https://github.com/ifm/ifm3d.git
       version: master
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5034,8 +5034,6 @@ repositories:
       url: https://github.com/ifm/ifm3d.git
       version: master
     release:
-      packages:
-      - ifm3d_core
       tags:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ifm/ifm3d-release.git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5029,6 +5029,10 @@ repositories:
       version: release
     status: developed
   ifm3d-release:
+    doc:
+      type: git
+      url: https://github.com/ifm/ifm3d.git
+      version: master
     release:
       packages:
       - ifm3d-core
@@ -5036,6 +5040,11 @@ repositories:
         release: release/kinetic/{package}/{version}
       url: https://github.com/ifm/ifm3d-release.git
       version: 0.17.0-6
+    source:
+      type: git
+      url: https://github.com/ifm/ifm3d.git
+      version: master
+    status: maintained
   ifm_o3mxxx:
     doc:
       type: git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -5029,20 +5029,11 @@ repositories:
       version: release
     status: developed
   ifm3d_core:
-    doc:
-      type: git
-      url: https://github.com/ifm/ifm3d.git
-      version: master
     release:
       tags:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -5028,6 +5028,23 @@ repositories:
       url: https://github.com/astuff/ibeo_lux.git
       version: release
     status: developed
+  ifm3d-release:
```
clalancette: This is an odd name for a key.  Usually the key is either the name of the repository (ifm3d), or the name of the package (ifm3d_core).

kinetic/distribution.yaml
```diff
@@ -5028,6 +5028,23 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/ifm/ifm3d.git
+      version: master
+    release:
+      packages:
+      - ifm3d-core
```
clalancette: Also, I think every other package in all of the distribution.yaml files uses `ifm3d_core`, so I'd suggest changing your package name to that.

clalancette (dismissed):

kinetic/distribution.yaml
theseankelly: hmm yes I agree, sorry for the oversight. This PR was auto-generated by `bloom-release`. This is my first release so pardon my learning curve -- shall I simply modify the PR manually or is there some metadata in my [tracks file](https://github.com/ifm/ifm3d-release/blob/master/tracks.yaml) that's missing which ought to specify this key name?

kinetic/distribution.yaml
theseankelly: Yup you're right -- REP144 mandates I should use an _ instead of -. I'll correct.

kinetic/distribution.yaml
theseankelly: Or is it the `repository` parameter in the `bloom-release` command that I should have set to ifm3d_core?

```
usage: bloom-release [-h] [--list-tracks] [--track TRACK] [--non-interactive]
                     --ros-distro ROS_DISTRO [--new-track] [--pretend]
                     [--no-pull-request] [--no-web] [--pull-request-only]
                     [--override-release-repository-url OVERRIDE_RELEASE_REPOSITORY_URL]
                     [--override-release-repository-push-url OVERRIDE_RELEASE_REPOSITORY_PUSH_URL]
                     [-d] [--version] [--unsafe]
                     repository
```

kinetic/distribution.yaml
```diff
@@ -5028,6 +5028,21 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/ifm/ifm3d.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/ifm/ifm3d-release.git
+      version: 0.17.0-9
+    source:
```
tfoote:
```suggestion
    source:
      test_commits: false
```

Since this repository doesn't have a package.xml it will fail our CI testing.

tfoote (changes requested):
The issues raised by @clalancette look to have been resolved, but I noticed one more thing to update so we don't try to build and test this non-ros package. The package.xml is being injected by the bloom process.

On second thought since this is a non-ROS package it won't generate documentation correctly either so we should not have a doc stanza either. And the source stanza won't work in a source checkout due to the lack of package.xml so we should probably remove the whole source stanza too.

kinetic/distribution.yaml
theseankelly: Sounds good.

Also, just FYI this non-ROS package is in support of the real ROS package (https://github.com/ifm/ifm3d-ros) and that'll be next on my list once the core driver/library is in the distribution.

Thanks for both your time and help.

theseankelly:
I've removed the two stanzas from both PRs

tfoote (approved):

PR #23793, opened 2020-02-17
Add rafcon-pip

nickswalker:
[RAFCON](https://dlr-rm.github.io/RAFCON/) is a state machine library with a graphical authoring interface.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5567,6 +5567,16 @@ python3-qt5-bindings-webkit:
   fedora: [python3-qt5-webkit]
   gentoo: ['dev-python/PyQt5[webkit]']
   ubuntu: [python3-pyqt5.qtwebkit]
+python3-rafcon-pip:
+  debian:
+    pip:
+      packages: [rafcon]
+  fedora:
+    pip:
+      packages: [rafcon]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5749,6 +5749,16 @@ quadprog-pip:
     pip: [quadprog]
   ubuntu:
     pip: [quadprog]
+rafcon-pip:
```
nuclearsandwich: Can you please rename (and therefore relocate) this key to `python3-rafcon-pip`. This isn't currently specified as a requirement but it's being added to the contribution guidelines in #23836.

nuclearsandwich (changes requested):
I'd like to see the key name changed to conform to the soon-to-be-updated contribution guidelines. As per the existing contribution guidelines, can you please add a link in the PR description to the package on PyPI when adding future pip packages, for example: https://pypi.org/project/rafcon/

rosdep/python.yaml
nickswalker: Done!

nuclearsandwich (approved):

clalancette (approved):

PR #23806, opened 2020-02-18
disable PR tests for rosbag2_bag_v2_plugins

Karsten1987:
Due to not sourcing `melodic`, the bridge is not available which makes the source PR jobs fail.
@zmichaels11 fyi

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -2003,7 +2003,7 @@ repositories:
       url: https://github.com/ros2-gbp/rosbag2_bag_v2-release.git
       version: 0.0.7-4
     source:
-      test_pull_requests: true
+      test_pull_requests: false
       type: git
       url: https://github.com/ros2/rosbag2_bag_v2.git
       version: master


emersonknapp:
Is there a way to fix the build environment instead of disabling the tests?

Karsten1987:
I've patched the workflow for the actual debian releases in here: https://github.com/ros2-gbp/rosbag2_bag_v2-release/pull/3

I don't know how to do that for source builds to be honest.

emersonknapp:
Perhaps @nuclearsandwich may have an idea?

mjcarroll (approved):

nuclearsandwich:
> Is there a way to fix the build environment instead of disabling the tests?

https://github.com/ros-infrastructure/ros_buildfarm does not currently have a method for supporting cross-rosdistro testing. Nor does rosdep itself have first-class support for dependencies from another rosdistro. Both those issues prevent the ROS buildfarm from running these tests or tests for the bridge or any other package which depends on ROS 1.

PR #23819, opened 2020-02-20
[melodic] Add the upstream source for cartographer_ros and cartographer

seanyen:
Add the upstream repository for completeness.

This is motivated by that https://aka.ms/ros project currently is not able to use `rosinstall_generator` to pull down the upstream repository information because of the missing information and has to maintain a separate `.rosinstall` file for `cartographer` related builds.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -945,6 +945,10 @@ repositories:
         release: release/melodic/{package}/{version}
       url: https://github.com/ros-gbp/cartographer-release.git
       version: 1.0.0-0
+    source:
+      type: git
+      url: https://github.com/googlecartographer/cartographer.git
+      version: master
     status: developed
   cartographer_ros:
     doc:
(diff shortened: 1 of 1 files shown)


seanyen:
@clalancette This is ready for review and merge. Thanks!

mikaelarguedas:
The code released in melodic does not match the `master` branch and does not compile as is: http://build.ros.org/job/Mdev__cartographer_ros__ubuntu_bionic_amd64/1
http://build.ros.org/job/Mdev__cartographer__ubuntu_bionic_amd64/1

IIRC the code on the upstream repository never compiled as is on the farm, that's the reason we have patches on the [release repository](https://github.com/ros-gbp/cartographer-release) and never added a source entry here.
If I understood correctly, the source repositories added in this PR are currently unmaintained so I would expect the code from the master branch to not build successfully in the foreseeable future. Therefore I suggest reverting this PR

@clalancette FYI

Issue #23767, opened 2020-02-14
PIP only package naming?


MatthijsBurgh:
Should PIP only packages be named `XXX-pip` or `python-XXX-pip`? Both variants are in the rosdistro, so it is unclear to me what the standard is.

tfoote:
I would suggest that we standardize to use `python-XXX-pip`. This will make it much clearer when a similar debian rule appears side by side for sorting. 

To that end it should probably be `python3-XXX-pip` since all new platforms will be on python3 now.

MatthijsBurgh:
Fixed by #23836

PR #23836, opened 2020-02-21
Add guidelines about pip packages

MatthijsBurgh:
Solves #23767

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,6 @@ Guidelines for rosdep rules
    Similarly don't call out indirect dependencies or default options for packages.
  * Python packages should go in the `python.yaml`.
  * Homebrew into `osx-homebrew.yaml`.
- * If a package is only available via pip use the `-pip` extension.
  * Supported Platforms
    * Rules can be contributed for any platform.
      However to be released they must be at least cover the supported platforms in REP 3: http://www.ros.org/reps/rep-0003.html So:
@@ -118,9 +117,11 @@ Work has been proposed to add a separate installer for AUR packages [ros-infrast
 
 #### pip
 
-For pip installers they are expected to be in the main pypi index.
+For pip installers they are expected to be in the main PyPI index https://pypi.org/.
 
-### Python 3 rules
+### Python rules
+
+#### Python 3
 
 When adding rules for python 3 packages, create a separate entry prefixed with `python3-` rather than `python`
 For example:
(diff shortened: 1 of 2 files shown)


tfoote (approved):

CONTRIBUTING.md
```diff
@@ -145,6 +143,23 @@ These were trialed as a possible style of Python 3 rules and should not be used.
 The guidance above should be followed for new rules.
 Additionally, if you rely on a dependency that uses `_python3`-suffixed codenames, add a new rule for it that follows the guidance above.
 
+#### pip
+
+Python packages, which are only available on pip, should use the `-pip` extension. Also the `python-`/`python3-` prefix needs to be used.
```
nuclearsandwich: Did we solve the python2 vs python3 pip issue by using the `python_version` key in the rosdistro index when installing rosdeps via pip?

nuclearsandwich (approved):
Thanks for contributing to the contributing guide. This will improve the process for future contributors!

As a general suggestion, it would be helpful if the formatting changes were made separately as they clutter the diff when reviewing the topic change.

CONTRIBUTING.md
MatthijsBurgh: @nuclearsandwich is this a question to me or to @tfoote? Because I don't understand your question at all.

tfoote: We do invoke the appropriate pip instance based on the `ROS_PYTHON_VERSION` https://github.com/ros-infrastructure/rosdep/pull/709

Which I guess does mean that we could omit the python version from the keys since it will translate to the appropriate one for your rosdistro. For the other package managers however they won't auto switch so any hybrid rules would require both versions in the package.xml with the conditional. 

So for consistency I think it makes sense to use the prefix even with the pip only packages. Technically it will cause duplication. But we're also approaching the EOL of python2 so the duplication will be short lived and then we will just have consistent rules and the pip ones will appear side by side with the non-pip ones.

nuclearsandwich: > @nuclearsandwich is this a question to me...

Sorry for the confusion @MatthijsBurgh it was directed back at the team. I ought to have said so.


> So for consistency I think it makes sense to use the prefix even with the pip only packages. Technically it will cause duplication.

I would have opted for just keeping a python3-foobar-pip key which would look odd in a python2 context but wouldn't be wrong. However if we're okay with the duplication (or doing something ugly but clever like using YAML anchors) then that's going to be better for downstream folks.

PR #23841, opened 2020-02-21
Add rosdep rules for orocos kdl and orocos bfl

tfoote:
Re: https://github.com/ros-gbp/bfl-release/issues/12

bfl rosdep keys:
ubuntu: https://packages.ubuntu.com/source/focal/orocos-bfl
debian: https://packages.debian.org/source/buster/orocos-bfl
fedora: https://debian.pkgs.org/10/debian-main-amd64/liborocos-bfl-dev_0.8.0-5_amd64.deb.html

python3-pykdl:
ubuntu: https://packages.ubuntu.com/focal/python3-pykdl
debian: https://packages.debian.org/buster/python3-pykdl

kdl keys: 
Fedora: https://fedora.pkgs.org/30/fedora-x86_64/orocos-kdl-devel-1.4.0-3.fc30.x86_64.rpm.html
Ubuntu: https://packages.ubuntu.com/xenial/liborocos-kdl-dev
Debian: https://packages.debian.org/source/buster/misc/orocos-kdl
Homebrew: https://formulae.brew.sh/formula/orocos-kdl
Gentoo: https://packages.gentoo.org/packages/sci-libs/orocos_kdl

@sloretz @jacobperron We should look at not releasing bfl and kdl into ROS packages for Noetic and Foxy with these being available upstream now. @meyerj @smits Does that make sense?

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2854,20 +2854,20 @@ liborocos-bfl:
   debian:
     '*': [liborocos-bfl0.8]
     'stretch': null
-  fedora: [liborocos-bfl]
+  fedora: [orocos-bfl]
   ubuntu:
     '*': [liborocos-bfl0.8]
     'bionic': null
     'xenial': null
 liborocos-bfl-dev:
   debian: [liborocos-bfl-dev]
-  fedora: [liborocos-bfl-devel]
+  fedora: [orocos-bfl-devel]
   ubuntu: [liborocos-bfl-dev]
 liborocos-kdl:
   debian:
     '*': [liborocos-kdl1.4]
     'stretch': [liborocos-kdl1.3]
-  fedora: [liborocos-kdl]
+  fedora: [orocos-kdl]
   gentoo: [sci-libs/orocos_kdl]
   ubuntu:
     '*': [liborocos-kdl1.4]
@@ -2875,7 +2875,7 @@ liborocos-kdl:
     'xenial': [liborocos-kdl1.3]
 liborocos-kdl-dev:
   debian: [liborocos-kdl-dev]
-  fedora: [liborocos-kdl-devel]
+  fedora: [orocos-kdl-devel]
   gentoo: [sci-libs/orocos_kdl]
   ubuntu: [liborocos-kdl-dev]
 libosmesa6-dev:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2854,7 +2854,7 @@ liborocos-bfl:
   debian:
     '*': [liborocos-bfl0.8]
     'stretch': null
-  fedora: [liborocos-kdl]
+  fedora: [liborocos-bfl]
   ubuntu:
     '*': [liborocos-bfl0.8]
     'bionic': null
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2850,6 +2850,19 @@ libopenvdb-dev:
   fedora: [openvdb-devel]
   gentoo: [media-gfx/openvdb]
   ubuntu: [libopenvdb-dev, libilmbase-dev]
+liborocos-bfl:
+  debian:
+    '*': [liborocos-bfl0.8]
+    'stretch': null
+  fedora: [liborocos-kdl]
+  ubuntu:
+    '*': [liborocos-bfl0.8]
+    'bionic': null
+    'xenial': null
+liborocos-bfl-dev:
+  debian: [liborocos-bfl-dev]
+  fedora: [liborocos-bfl-devel]
+  ubuntu: [liborocos-bfl-dev]
 liborocos-kdl:
   debian:
     '*': [liborocos-kdl1.4]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2850,6 +2850,21 @@ libopenvdb-dev:
   fedora: [openvdb-devel]
   gentoo: [media-gfx/openvdb]
   ubuntu: [libopenvdb-dev, libilmbase-dev]
+liborocos-kdl:
+  debian:
+    '*': [liborocos-kdl1.4]
+    'stretch': [liborocos-kdl1.3]
+  fedora: [liborocos-kdl]
+  gentoo: [sci-libs/orocos_kdl]
+  ubuntu:
+    '*': [liborocos-kdl1.4]
+    'bionic': [liborocos-kdl1.3]
+    'xenial': [liborocos-kdl1.3]
+liborocos-kdl-dev:
+  debian: [liborocos-kdl-dev]
+  fedora: [liborocos-kdl-devel]
+  gentoo: [sci-libs/orocos_kdl]
+  ubuntu: [liborocos-kdl-dev]
 libosmesa6-dev:
   arch: [mesa]
   debian: [libosmesa6-dev]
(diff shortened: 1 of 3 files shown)


meyerj:
I was not aware that KDL was packages for all those distributions. It seems like the right way to go, but on the other hand I would not know the process or how long it takes to get a bugfix release in all distributions. Taking into account that KDL is quite stable and there is not a lot of activity (at least not from the maintainer's side), it is probably okay to transition to upstream releases.

What does that mean for packages that currently depend on`orocos_kdl`? Would all the manifests require an update then? Or could the rosdep key `orocos_kdl` not point to the respective upstream packages starting from Noetic? I don't like the fact that sometimes different source branches need to be maintained per ROS distro because of those small differences, although the actual code could be exactly the same (e.g. for Gazebo before https://github.com/ros-simulation/gazebo_ros_pkgs/pull/473).

I will check in the coming days how the different versions compare, how much they lack behind the current master branch and which pending pull requests could be merged before another 1.4 release. But if the KDL release is blocking you and the upstream versions work, feel free to go on. I assume that if something would break along the way, the decision could still be revised until the Noetic release?

tfoote:
> how long it takes to get a bugfix release in all distributions.

To set the expectation for this, don't expect bugfixes to get propagated. In general once things are in the distro after release they're effectively static.

Since KDL is quite stable now I hope that we're ready to do this transition.

The challenge with overlaying our packagest if there's upstream packages, it's very hard to coexist. Think of very nasty linking problems because you accidentally end up linking against the system version which is just slightly different. For many of our core packages we have to end up setting `conflicts` and `replaces` rules as well as make sure that our version numbers are strictly higher than the upstream ones. On the flip side projects and tools upstream that depend on the system version may have problems working in the ROS workspace if we have a slightly newer version.

In general as long as there's not any critical issues with the version upstream I'd suggest we should work with that version. We'll be a little bit behind the latest, but we'll be that way anyway. If there's a super critical bugfix we could roll our own patched version of the upstream package that's ABI compatible and add it to our repositories.

Also if you identify anything missing from the focal version, it's still in the prerelease and we could work with the debian/ubuntu maintainers to get patches into that before the release. @j-rivero FYI.

We can make the decision up until the noetic release but it really needs to be noteably sooner as all downstream packages will need to change their dependencies to point to the rosdep key instead of the pure package dependency or vice versa. The new package.xml has support for conditional dependencies so you don't have to branch development to support the change. That's of course assuming the code is the same and it's just the packaging.

nuclearsandwich (approved):

j-rivero:
> Also if you identify anything missing from the focal version, it's still in the prerelease and we could work with the debian/ubuntu maintainers to get patches into that before the release. @j-rivero FYI.

We are only 3 days away of closing the Focal autosync from Debian and after that, getting changes into Ubuntu will be harder. It is probably too late for something touching the API/ABI but small we can give a try for small patches.

jacobperron (approved):

rosdep/base.yaml
```diff
@@ -2850,6 +2850,34 @@ libopenvdb-dev:
   fedora: [openvdb-devel]
   gentoo: [media-gfx/openvdb]
   ubuntu: [libopenvdb-dev, libilmbase-dev]
+liborocos-bfl:
+  debian:
+    '*': [liborocos-bfl0.8]
+    'stretch': null
+  fedora: [liborocos-kdl]
```
jacobperron: Should this point to a `*-bfl` package instead of `*-kdl`?

tfoote: Good catch

jacobperron (approved):

rosdep/base.yaml
```diff
@@ -2850,6 +2850,34 @@ libopenvdb-dev:
...
+    '*': [liborocos-bfl0.8]
+    'stretch': null
+  fedora: [liborocos-bfl]
+  ubuntu:
+    '*': [liborocos-bfl0.8]
+    'bionic': null
+    'xenial': null
+liborocos-bfl-dev:
+  debian: [liborocos-bfl-dev]
+  fedora: [liborocos-bfl-devel]
```
sloretz: Should this be [orocos-bfl-devel](https://fedora.pkgs.org/30/fedora-x86_64/orocos-bfl-devel-0.8.99-17.20180529git3d0d149.fc30.x86_64.rpm.html)?

rosdep/base.yaml
```diff
@@ -2850,6 +2850,34 @@ libopenvdb-dev:
   fedora: [openvdb-devel]
   gentoo: [media-gfx/openvdb]
   ubuntu: [libopenvdb-dev, libilmbase-dev]
+liborocos-bfl:
+  debian:
+    '*': [liborocos-bfl0.8]
+    'stretch': null
+  fedora: [liborocos-bfl]
```
sloretz: Is this supposed to be [orocos-bfl](https://fedora.pkgs.org/30/fedora-x86_64/orocos-bfl-0.8.99-17.20180529git3d0d149.fc30.x86_64.rpm.html)?

rosdep/base.yaml
```diff
@@ -2850,6 +2850,34 @@ libopenvdb-dev:
...
+    'stretch': [liborocos-kdl1.3]
+  fedora: [liborocos-kdl]
+  gentoo: [sci-libs/orocos_kdl]
+  ubuntu:
+    '*': [liborocos-kdl1.4]
+    'bionic': [liborocos-kdl1.3]
+    'xenial': [liborocos-kdl1.3]
+liborocos-kdl-dev:
+  debian: [liborocos-kdl-dev]
+  fedora: [liborocos-kdl-devel]
```
sloretz: [orocos-kdl-devel](https://fedora.pkgs.org/30/fedora-x86_64/orocos-kdl-devel-1.4.0-3.fc30.x86_64.rpm.html)?

rosdep/base.yaml
```diff
@@ -2850,6 +2850,34 @@ libopenvdb-dev:
...
+    'xenial': null
+liborocos-bfl-dev:
+  debian: [liborocos-bfl-dev]
+  fedora: [liborocos-bfl-devel]
+  ubuntu: [liborocos-bfl-dev]
+liborocos-kdl:
+  debian:
+    '*': [liborocos-kdl1.4]
+    'stretch': [liborocos-kdl1.3]
+  fedora: [liborocos-kdl]
```
sloretz: [orocos-kdl](https://fedora.pkgs.org/30/fedora-x86_64/orocos-kdl-1.4.0-3.fc30.x86_64.rpm.html)?

rosdep/base.yaml
nuclearsandwich: Good catch. I think this correction (and the similar comment below, are correct)

tfoote:
Fixed fedora links:

* https://fedora.pkgs.org/30/fedora-x86_64/orocos-bfl-devel-0.8.99-17.20180529git3d0d149.fc30.x86_64.rpm.html
* https://fedora.pkgs.org/30/fedora-x86_64/orocos-bfl-0.8.99-17.20180529git3d0d149.fc30.x86_64.rpm.html
* https://fedora.pkgs.org/30/fedora-x86_64/orocos-kdl-1.4.0-3.fc30.x86_64.rpm.html
* https://fedora.pkgs.org/30/fedora-x86_64/orocos-kdl-devel-1.4.0-3.fc30.x86_64.rpm.html

sloretz (approved):

ros-discourse:

https://discourse.ros.org/t/updates-on-orocos-ros-1-and-2-integration/15146/13

PR #23844, opened 2020-02-21
Update Boost python rules

mikaelarguedas:
Firs commit creates a `libboost-python-dev` key and copies the content of `libboost-python` into it

The second updates `libboost-python` to not include all the development headers. Note that this is changing the rules of the key existing before this PR

Note: on debian/ubuntu: libboost-python is a hybrid package providing libraries for both python2 and python3. It's unclear to me what rule to use for system that have different packages for python2 and python3 e.g. fedora. So I opted to both to be the closest to ubuntu/debian behavior

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1683,7 +1683,7 @@ libboost-python:
     bionic: [libboost-python1.65.1]
     disco: [libboost-python1.67.0]
     eoan: [libboost-python1.67.0]
-    focal: [libboost-python1.67.0]
+    focal: [libboost-python1.71.0]
     trusty: [libboost-python-1.54.0]
     xenial: [libboost-python-1.58.0]
 libboost-python-dev:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1674,9 +1674,18 @@ libboost-program-options-dev:
   ubuntu: [libboost-program-options-dev]
 libboost-python:
   alpine: [boost-python2]
-  debian: [libboost-python-dev]
+  debian:
+    buster: [libboost-python1.67.0]
+    jessie: [libboost-python1.55.0]
+    stretch: [libboost-python1.62.0]
   fedora: [boost-python2-devel]
-  ubuntu: [libboost-python-dev]
+  ubuntu:
+    bionic: [libboost-python1.65.1]
+    disco: [libboost-python1.67.0]
+    eoan: [libboost-python1.67.0]
+    focal: [libboost-python1.67.0]
+    trusty: [libboost-python-1.54.0]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> Note that this is changing the rules of the key existing before this PR

It would be good to check how packages use this key and how likely this change will result in regressions.

mikaelarguedas:
> It would be good to check how packages use this key and how likely this change will result in regressions.

As far as I can tell it's not used by any released package on any active ROS distribution.

It was added fairly recently: https://github.com/ros/rosdistro/pull/21570
Maybe @JWhitleyWork can provide more context on their use of this key

JWhitleyWork:
@mikaelarguedas / @dirk-thomas It's currently used by https://github.com/fzi-forschungszentrum-informatik/Lanelet2. This is a pretty major part of the Autoware project and it would be a breaking change. Would one of you mind coordinating with the maintainers of that repo to implement the change at the same time and keep me in the loop? We will need to update our reference to that repo to keep from breaking our builds once the change is in place.

mikaelarguedas:
Sounds good. Do you happen to know of any other repo than that one using it ?

To be fair, it's unlikely to break as the majority of ROS packages bring in the boost headers ([all the way down to ros_comm](https://github.com/ros/ros_comm/pull/1871)). For example `lanelet_python` depends directly on `lanelet_core` that `<depend>boost</depend>` so it pulls `libboost-all-dev`.
I opened https://github.com/fzi-forschungszentrum-informatik/Lanelet2/pull/88 for correctness

tfoote (approved):
This looks like a good cleanup.

JWhitleyWork (approved):
Based on the above explanation, this should be fine.

rosdep/base.yaml
```diff
@@ -1673,6 +1673,20 @@ libboost-program-options-dev:
...
+  debian:
+    buster: [libboost-python1.67.0]
+    jessie: [libboost-python1.55.0]
+    stretch: [libboost-python1.62.0]
+  fedora: [boost-python2-devel, boost-python3-devel]
+  ubuntu:
+    bionic: [libboost-python1.65.1]
+    disco: [libboost-python1.67.0]
+    eoan: [libboost-python1.67.0]
+    focal: [libboost-python1.67.0]
```
dirk-thomas: This looks outdated: https://packages.ubuntu.com/focal/libboost-python1.71.0

mikaelarguedas: It was still in "proposed' when I opened the PR. Updated in https://github.com/ros/rosdistro/pull/23844/commits/d6effc59f2ba003d5aef6fb4f14598dd007c2964

PR #23859, opened 2020-02-24
Adding carla_msgs to documentation index for kinetic and melodic

vimalrajliya:
I'd like package carla_ros_bridge to be indexed and documented on ros.org.


germanros1987:
@tfoote @clalancette is there any way we can help expedite this PR? Thanks!

germanros1987:
Thank you very much! :D

PR #23867, opened 2020-02-25
ros_numpy: 0.0.2-0 in 'melodic/distribution.yaml' [bloom]

Minipada:
Add ros_numpy to melodic. It is exactly the same as in jade and kinetic as there were no releases since.

Repo link [https://github.com/eric-wieser/ros_numpy](https://github.com/eric-wieser/ros_numpy)

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -8277,7 +8277,7 @@ repositories:
       version: master
     release:
       tags:
-        release: release/jade/{package}/{version}
+        release: release/melodic/{package}/{version}
       url: https://github.com/eric-wieser/ros_numpy-release.git
       version: 0.0.2-0
     source:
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -8270,6 +8270,21 @@ repositories:
...
       version: master
     status: maintained
+  ros_numpy:
+    doc:
+      type: git
+      url: https://github.com/eric-wieser/ros_numpy.git
+      version: master
+    release:
+      tags:
+        release: release/jade/{package}/{version}
```
clalancette: This isn't going to work.  Even though there have been no changes, you still need to re-bloom this for Melodic, which will create the melodic release branch.

clalancette (changes requested):

melodic/distribution.yaml
Minipada: Alright, I changed it to melodic

tfoote:
That tag won't exist until you rurun bloom

I'm going to close this as rerunning bloom will create the new PR with the correct content.

PR #23950, opened 2020-03-04
Added software-properties-common to rosdep

pablogs9:
Added software-properties-common key for micro-ROS project

- Ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=software-properties-common&searchon=names
- Debian: https://packages.debian.org/search?searchon=names&keywords=software-properties-common

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5184,6 +5184,9 @@ socat:
   fedora: [socat]
   gentoo: [net-misc/socat]
   ubuntu: [socat]
+software-properties-common:
+  debian: [software-properties-common]
+  ubuntu: [software-properties-common]
 sound-theme-freedesktop:
   debian: [sound-theme-freedesktop]
   fedora: [sound-theme-freedesktop]


clalancette (commented):
I have to say, this is a weird key to put into the rosdep database.  It's very specific to Ubuntu/Debian, and it is not something that downstream packages need to build or run, typically.  I guess you could depend on it in order to bring in `apt-add-repository`, but it still feels weird to me.  @pablogs9 can you explain more about your use-case for this key?

pablogs9:
In the context of micro-ROS we need `software-properties-common` in order to install the last CMake version (3.13.1 or higher), which is not available by default on Debian/Ubuntu and it is needed by the Zephyr RTOS toolchain:

https://docs.zephyrproject.org/latest/getting_started/index.html#install-dependencies

Here you have the Zephyr + micro-ROS build system: 

https://github.com/micro-ROS/micro-ros-build/tree/feature/zephyr/micro_ros_setup

tfoote:
I agree with @clalancette that this isn't really a great "dependency" for a software package. This is really an installation prerequisite. And although we don't want to be judging the usefulness of packages that we add to the rosdep rules we can add the rule. It would be reasonable to make a package that has the scripts to help setup the environment. So I don't think that we should block this.

Circling back though, I would strongly suggest that the use of this tool is less generic than just piping the contents to the file like most installation instructions do. Ala: http://wiki.ros.org/kinetic/Installation/Ubuntu 

The command has a lot of dependencies for doing something that's one line of bash and is the same level of complexity as the previous line of the installation above in the installation tutorials for the keys. As such it is only adding complexity not reducing it. Thus I'd ask do you really need this for this use case when 
`sudo sh -c 'echo "deb https://apt.kitware.com/ubuntu/ bionic main" > /etc/apt/sources.list.d/kitware.list'`

pablogs9:
Ok undestood, thanks!

PR #23966, opened 2020-03-05
Added ed key

pablogs9:
Added ed key for micro-ROS project:
- Ubuntu: https://packages.ubuntu.com/search?keywords=ed&searchon=names
- Debian: https://packages.debian.org/search?searchon=names&keywords=ed
- Fedora: https://src.fedoraproject.org/rpms/ed


clalancette (approved):

nuclearsandwich (approved):
ed is the standard Unix text editor....

PR #23967, opened 2020-03-05
Add rosdep rule for range-v3 library

miguelprada:
Rosdep key for [range-v3](https://ericniebler.github.io/range-v3/) library, which: _is the basis of the range support in C++20_.

I believe the ubuntu, debian and gentoo packages are correct. Package listings in the respective repositories can be found in:

- https://packages.ubuntu.com/search?keywords=range-v3&searchon=names&suite=eoan&section=all
- https://packages.debian.org/buster/librange-v3-dev
- https://packages.gentoo.org/packages/dev-cpp/range-v3

I'm not so sure about fedora. I could not find any related result in [fedora packages](https://apps.fedoraproject.org/packages/s/range-v3), but there's some result [here](https://src.fedoraproject.org/rpms/range-v3). I don't know what that means. Any help here would be appreciated.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4994,6 +4994,11 @@ r-base-dev:
   fedora: [R-devel]
   gentoo: [dev-lang/R]
   ubuntu: [r-base-dev]
+range-v3:
+  debian: [librange-v3-dev]
+  fedora: [range-v3]
+  gentoo: [dev-cpp/range-v3]
+  ubuntu: [librange-v3-dev]
 rapidjson-dev:
   debian: [rapidjson-dev]
   fedora: [rapidjson]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4997,14 +4997,11 @@ r-base-dev:
 range-v3:
   debian:
     '*': [librange-v3-dev]
-    jessie: null
     stretch: null
   fedora: [range-v3]
   gentoo: [dev-cpp/range-v3]
   ubuntu:
     '*': [librange-v3-dev]
-    artful: null
-    wily: null
     xenial: null
 rapidjson-dev:
   debian: [rapidjson-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4995,10 +4995,17 @@ r-base-dev:
   gentoo: [dev-lang/R]
   ubuntu: [r-base-dev]
 range-v3:
-  debian: [librange-v3-dev]
+  debian:
+    '*': [librange-v3-dev]
+    jessie: null
+    stretch: null
   fedora: [range-v3]
   gentoo: [dev-cpp/range-v3]
-  ubuntu: [librange-v3-dev]
+  ubuntu:
+    '*': [librange-v3-dev]
+    artful: null
+    wily: null
+    xenial: null
 rapidjson-dev:
   debian: [rapidjson-dev]
   fedora: [rapidjson]


rosdep/base.yaml
```diff
@@ -4994,6 +4994,11 @@ r-base-dev:
   fedora: [R-devel]
   gentoo: [dev-lang/R]
   ubuntu: [r-base-dev]
+range-v3:
+  debian: [librange-v3-dev]
```
clalancette: This package is only available for debian buster and later.  We still support debian stretch for Melodic, so this key will have to be more specific.  See https://github.com/ros/rosdistro/blob/1cf2a0266f6abaaf86293772cb3f3570a2b8bf82/rosdep/base.yaml#L229 for an example on how to do this.

rosdep/base.yaml
```diff
@@ -4994,6 +4994,11 @@ r-base-dev:
...
+  fedora: [range-v3]
+  gentoo: [dev-cpp/range-v3]
+  ubuntu: [librange-v3-dev]
```
clalancette: Same comment for ubuntu; this package is only available on eoan and later, but we still support xenial and bionic for Kinetic and Melodic, respectively.  This key will need to be more specific.

clalancette (changes requested):
Need changes to the Debian and Ubuntu keys.  Also, this package is available on all current fedora distributions, so please add a fedora key as well.

rosdep/base.yaml
miguelprada: Marked as unavailable for stretch. The package is available in stretch-backports repository, but I guess that doesn't count, right?

rosdep/base.yaml
miguelprada: It's available for [bionic](https://packages.ubuntu.com/bionic/librange-v3-dev) as well.

miguelprada:
Thanks for the review, @clalancette. I think I addressed your comments about the debian/ubuntu distribution specificity.

Re. the fedora key: there's one there already, but I'm not sure that's correct. Should it be specified differnently?

rosdep/base.yaml
clalancette: Correct, we can only depend on things in the main, security, universe, and multiverse repositories.

rosdep/base.yaml
```diff
@@ -4994,6 +4994,18 @@ r-base-dev:
   fedora: [R-devel]
   gentoo: [dev-lang/R]
   ubuntu: [r-base-dev]
+range-v3:
+  debian:
+    '*': [librange-v3-dev]
+    jessie: null
```
clalancette: We don't need to add jessie here, as it is out-of-support.

rosdep/base.yaml
```diff
@@ -4994,6 +4994,18 @@ r-base-dev:
...
+    stretch: null
+  fedora: [range-v3]
+  gentoo: [dev-cpp/range-v3]
+  ubuntu:
+    '*': [librange-v3-dev]
+    artful: null
+    wily: null
```
clalancette: Same for artful and wily; since they are out-of-support, we don't need to add exceptions for them.

clalancette (changes requested):
Don't add exceptions for out-of-support distros.  Once that is fixed, looks good to me.

rosdep/base.yaml
miguelprada: I added it because it's one of the _recommended_ supported platforms for Kinetic and I guessed it would be nice to include. It's gone now.

rosdep/base.yaml
miguelprada: Same here. Added those because they were in the list of supported platforms for Kinetic (wily) and Melodic (artful). Also removed these.

rosdep/base.yaml
clalancette: Yeah, we're a bit inconsistent there.  We support these other Debian/Ubuntu versions when we first release the ROS distributions.  Once they go out of support, we stop making changes to rosdep keys for them, but we don't remove them from the REP.  Anyway, looks good to me now, thanks for iterating.

clalancette (approved):

rosdep/base.yaml
miguelprada: Got it. Thanks for the explanation :+1:

nuclearsandwich (approved):
Thanks for iterating and getting that Fedora key in.

PR #23973, opened 2020-03-05
Add triangle key

JulianKu:
pip Triangle package (https://pypi.org/project/triangle/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4707,6 +4707,19 @@ python-trep:
   ubuntu:
     pip:
       packages: [trep]
+python-triangle-pip:
+  debian:
+    pip:
+      packages: [triangle]
+  fedora:
+    pip:
+      packages: [triangle]
+  osx:
+    pip:
+      packages: [triangle]
+  ubuntu:
+    pip:
+      packages: [triangle]
 python-twisted-bin:
   arch: [python2-twisted]
   debian: [python-twisted-bin]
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
Please follow the pip rules in https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1 .  In particular, make sure the key is prefixed with `python-` and/or `python3-`.

clalancette (approved):
Thanks for iterating.

tfoote (changes requested):
This is described as "Python binding to the triangle library". Do we need a dependency on the triangle library?

I see a triangle package already in ubuntu? https://packages.ubuntu.com/xenial/triangle-bin how will this play with that? It looks like this has a copy of that implementation actually inside that, which might lead to masking or otherwise interfering with the native package if installed as well?

JulianKu:
> This is described as "Python binding to the triangle library". Do we need a dependency on the triangle library?

The triangle package offers an easy way to triangulate in Python between a set of vertices that could, for instance, represent a map. Triangulated maps can then be used by Path Planning algorithms, for example. This is why I definitely see a good use for this dependency.

> I see a triangle package already in ubuntu? https://packages.ubuntu.com/xenial/triangle-bin how will this play with that? It looks like this has a copy of that implementation actually inside that, which might lead to masking or otherwise interfering with the native package if installed as well?

From what I saw, the Python package is self-contained (when installed via pip, the package becomes just some Python wrappers + the compiled Cython file).
Having the pip package installed alongside the ubuntu package does not cause any issues on my computer for either of them.

JulianKu:
Are there any updates? Is there something else I can do?

sloretz:
> I see a triangle package already in ubuntu? https://packages.ubuntu.com/xenial/triangle-bin how will this play with that? It looks like this has a copy of that implementation actually inside that, which might lead to masking or otherwise interfering with the native package if installed as well?

It looks like it won't interfere with the debian package so long as it doesn't install a `showme`, `triangle` or `tricall` executable. https://packages.ubuntu.com/xenial/amd64/triangle-bin/filelist . It doesn't look like the pip package installs any executables at all: https://github.com/drufat/triangle/blob/master/setup.py

sloretz (approved):

tfoote (approved):
If this won't interfere than we can add it.

PR #23992, opened 2020-03-06
pyquaternion: 0.9.6-1 in 'kinetic/distribution.yaml' [bloom]

Achllle:
Increasing version of package(s) in repository `pyquaternion` to `0.9.6-1`:

- upstream repository: https://github.com/Achllle/pyquaternion
- release repository: https://github.com/Achllle/pyquaternion.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.9.3`
- previous version for package: `null`

## pyquaternion

```
Post ROS package conversion
* Change exec depend naming from numpy to python-numpy
* Add numpy dependency to package.xml
* Create catkin package, rename and move some files
* Fix casting error in trace_method
* Add setter for vector
```

This is a second attempt at #23794 which was merged and undone

PR #23993, opened 2020-03-07
pyquaternion: 0.9.6-4 in 'kinetic/distribution.yaml' [bloom] replaces rosdep python-pyquaterion-pip

Achllle:
Increasing version of package(s) in repository `pyquaternion` to `0.9.6-4`:

- upstream repository: https://github.com/Achllle/pyquaternion
- release repository: https://github.com/Achllle/pyquaternion-release.git
- bloom version: `0.9.3`

## pyquaternion

```
Post ROS package conversion
* Change exec depend naming from numpy to python-numpy
* Add numpy dependency to package.xml
* Create catkin package, rename and move some files
* Fix casting error in trace_method
* Add setter for vector
```

See also #23992 and #23794

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -10520,6 +10520,21 @@ repositories:
       url: https://github.com/ipab-slmc/pybind11_catkin.git
       version: master
     status: developed
+  pyquaternion:
+    doc:
+      type: git
+      url: https://github.com/Achllle/pyquaternion.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/Achllle/pyquaternion-release.git
+      version: 0.9.6-4
+    source:
+      type: git
+      url: https://github.com/Achllle/pyquaternion.git
+      version: master
+    status: maintained
   pyros:
     doc:
       type: git
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3377,16 +3377,6 @@ python-pyqrcode:
     xenial:
       pip:
         packages: [PyQRCode]
-python-pyquaternion-pip:
-  debian:
-    pip:
-      packages: [pyquaternion]
-  fedora:
-    pip:
-      packages: [pyquaternion]
-  ubuntu:
-    pip:
-      packages: [pyquaternion]
 python-pyquery:
   debian: [python-pyquery]
   fedora: [python-pyquery]


Achllle:
I will release for Melodic as well but currently running into [this issue on ros_buildfarm](https://github.com/ros-infrastructure/ros_buildfarm/issues/695) for Melodic. For Noetic, I'm running into a
```
rosdep2.lookup.ResolutionError: No definition of [python-numpy] for OS version [focal]
```

nuclearsandwich:
@tfoote do you prefer the rosdep changes be moved to a separate PR for visibility? I'm also not sure if we should remove the rosdep key since the "cat is out of the bag" on that. Those wishing to target the packaged dependency can update their dependency from `python-pyquaternion-pip` to `pyquaternion`.

tfoote (approved):
We could update the title to be searchable for the rosdep removal. 

pip rules in particular we cannot maintain perfect continuity. See my comments here: https://github.com/ros/rosdistro/pull/23794#issuecomment-590691124

We are changing the version of the package that we recommend and to that end we shouldn't effectively keep recommending the other version. In particular the pip rules tend to override and persist over any other installation and thus will often silently break installations and cause insidious bugs due to slightly different versions. 

Removing the rule so that rosdep fails will also give people clear errors. If someone needs the pip version it's moderately easy for them to add their own rosdep source with the pip key. But the more people who use the pip version the more problems we are going to have down the line.

It was released by the maintainer as an experiment and he's chosen to release another way so it can be better integrated into our systems. 


 > rosdep2.lookup.ResolutionError: No definition of [python-numpy] for OS version [focal]

This will need a new rosdep rule contribution for python-numpy on focal. You can go ahead and submit a PR for it if it's blocking you.

tfoote:
@PymZoR  you're going to either need to update your dependency to use the newly released packages or otherwise create your own rosdep sources.list with this entry that we're removing. Unfortunately we can't persist all pip rules when there are preferred native packages as they will cause conflicts generally.

sloretz:
> > rosdep2.lookup.ResolutionError: No definition of [python-numpy] for OS version [focal]
> 
> This will need a new rosdep rule contribution for python-numpy on focal. You can go ahead and submit a PR for it if it's blocking you.

A package released to Noetic should use the Python 3 rosdep key `python3-numpy` instead.

nuclearsandwich (approved):

PR #23996, opened 2020-03-08
Add all libboost-* keys to gentoo

awesomebytes:
Note that in Gentoo the libboost-* and libboost-*-dev are the same as everything is built from source.

Previously I only added a couple of keys as they were stopping builds of ROS on Gentoo. A couple more were neede now, so I figured out I could just fill them all.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1588,6 +1588,7 @@ libboost-atomic:
     buster: [libboost-atomic1.67.0]
     stretch: [libboost-atomic1.62.0]
   fedora: [boost-atomic]
+  gentoo: [dev-libs/boost]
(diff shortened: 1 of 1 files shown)


allenh1 (approved):
:+1: from me

tfoote (approved):
Thanks for filling these in.

clalancette (approved):

clalancette:
@awesomebytes This is all approved now, but I think we've merged some other boost stuff in the meantime.  Can you please rebase this?  Sorry for the trouble.

awesomebytes:
@clalancette Done (I just tried this rebase feature from GitHub itself, and it's great!).

PR #24007, opened 2020-03-09
WIP: add moveit_jog_arm to melodic distribution

tylerjw:
`moveit_jog_arm` is a package provided by moveit.  This would allow us to depend on it and install it as it gets released in debians.  @AndyZe is the primary developer/maintainer of it.

I marked this WIP so it is ignored until we get jog_arm in melodic-devel, which will require the latest changes in master of moveit_msgs to get merged into it's own melodic-devel.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -5190,6 +5190,7 @@ repositories:
       - moveit_core
       - moveit_experimental
       - moveit_fake_controller_manager
+      - moveit_jog_arm
       - moveit_kinematics
       - moveit_planners
       - moveit_planners_chomp


AndyZe:
looks good to me

AndyZe (approved):

tylerjw:
@JafarAbdi pointed out that this will fail OSRF build farm because jog_arm doesn't exist in `melodic-devel` branch of moveit, only in `master`.

tylerjw:
I marked this WIP so it is ignored until we get jog_arm in melodic-devel.

tfoote:
I'm going to close this. If jog_arm is enabled the next time that bloom runs it will automatically add it to the list. There's no need to preadd it to the list, it will just be wrong until the next bloom release.

PR #24009, opened 2020-03-10
Remove orocos_kinematics_dynamics Noetic source entry

sloretz:
Signed-off-by: Shane Loretz <sloretz@osrfoundation.org>


sloretz:
Now that https://github.com/ros/rosdistro/pull/23841 is merged, ROS Noetic will use orocos kdl from the apt repos. This replaces https://github.com/orocos/orocos_kinematics_dynamics/issues/218 . See https://github.com/ros/geometry2/pull/447 for an example change to the `package.xml` of dependent packages.

tfoote (approved):

PR #24019, opened 2020-03-10
Add python-pathlib in rosdep/python.yaml

Rezenders:
https://packages.ubuntu.com/bionic/python-pathlib
https://pypi.org/project/pathlib/

pathlib offers a set of classes to handle filesystem paths. It offers the following advantages over using string objects:

No more cumbersome use of os and os.path functions. Everything can be done easily through operators, attribute accesses, and method calls.
Embodies the semantics of different path types. For example, comparing Windows paths ignores casing.
Well-defined semantics, eliminating any warts or ambiguities (forward vs. backward slashes, etc.).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2855,6 +2855,8 @@ python-path.py:
   ubuntu:
     pip:
       packages: [path.py]
+python-pathlib:
+  ubuntu: [python-pathlib]
 python-pathos-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2857,6 +2857,10 @@ python-path.py:
       packages: [path.py]
 python-pathlib:
   ubuntu: [python-pathlib]
+  debian: [python-pathlib]
+  fedora: [python-pathlib]
+  gentoo: [dev-python/pathlib]
+  arch: [python2-pathlib]
 python-pathos-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2856,11 +2856,11 @@ python-path.py:
     pip:
       packages: [path.py]
 python-pathlib:
-  ubuntu: [python-pathlib]
+  arch: [python2-pathlib]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
pathlib is packaged via apt. If you'd like to add this as a dependency we should use the native packages not the pip installation technique. See the [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions). 

For example: https://packages.ubuntu.com/search?keywords=python-pathlib

Rezenders:
I am sorry but I didn't understand. Isn't  `python-pathlib` already the name of the native package? Should I add it to `base.yaml` instead of `python.yaml`?

tfoote:
Sorry I saw your pypi link and only one entry. That one is valid. Please fill in at least our default cross platform targets called out in the Contributing guide.

Rezenders:
Done at 9c292aa

tfoote (approved):
Great thanks for the quick update. LGTM just one more signoff needed.

rosdep/python.yaml
```diff
@@ -2855,6 +2855,12 @@ python-path.py:
   ubuntu:
     pip:
       packages: [path.py]
+python-pathlib:
+  arch: [python2-pathlib]
+  debian: [python-pathlib]
+  fedora: [python-pathlib]
```
clalancette: As of Fedora 31, python-pathlib is no longer available.  This is because of the larger effort to remove python2 from the Fedora default install.  Given Fedora's typical [support schedule](https://en.wikipedia.org/wiki/Fedora_version_history), I'd expect that Fedora 30 is going out of support by the end of May.  Because of that, I'd suggest that we just remove the fedora entry here, as it won't be valid in 2 more months.

clalancette (changes requested):

rosdep/python.yaml
Rezenders: Done 112e671

clalancette (approved):

PR #24035, opened 2020-03-11
Add noetic in list of ubuntu distros for python-numpy

Achllle:
Currently, depending on `python-numpy` in Noetic fails with:
```
rosdep2.lookup.ResolutionError: No definition of [python-numpy] for OS version [focal]
```

As [suggested here](https://github.com/ros/rosdistro/pull/23993#pullrequestreview-372323089), I'm adding it in here for Noetic. I left the previous version starting with 'n' (natty) in but can remove that.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2628,6 +2628,7 @@ python-numpy:
     lucid: [python-numpy]
     maverick: [python-numpy]
     natty: [python-numpy]
+    noetic: [python-numpy]
     oneiric: [python-numpy]
     precise: [python-numpy]
     quantal: [python-numpy]


rosdep/python.yaml
```diff
@@ -2628,6 +2628,7 @@ python-numpy:
     lucid: [python-numpy]
     maverick: [python-numpy]
     natty: [python-numpy]
+    noetic: [python-numpy]
```
sloretz: Noetic is the name of the ROS distro. This would need to be the Ubuntu distro `focal`.

sloretz:
`python-numpy` is a Python 2 package, and ROS Noetic is only targeting Python 3. Please use the rosdep key `python3-numpy` instead ([it already exists, no need to open a PR](https://github.com/ros/rosdistro/blob/4bfca115e18c733686b3b39524fbf1e227177574/rosdep/python.yaml#L5409-L5415)). Here are some pages about transitioning ROS packages to Python 3: http://wiki.ros.org/UsingPython3/ .

Achllle:
Makes sense. Clearly I had no idea what I was doing :)

PR #24054, opened 2020-03-11
Add python3-pykdl for Fedora 32+

sloretz:
This adds an entry for `python3-pykdl` for Fedora 32 and beyond. I put `null` entries for 31 and 30 only since [29 and below seem to be EOL](https://fedoraproject.org/wiki/End_of_life#Unsupported_Fedora_Releases).

I think this package should exist in Fedora 32 as of @cottsay's PR https://src.fedoraproject.org/rpms/orocos-kdl/pull-request/1 , but I don't actually know where to get a link proving a Fedora package exists. @cottsay is this rosdistro PR premature?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5567,6 +5567,10 @@ python3-pykdl:
     '*': [python3-pykdl]
     'bionic': null
     'xenial': null
+  fedora:
+    '*': [python3-pykdl]
+    '31': null
+    '30': null
   ubuntu:
     '*': [python3-pykdl]
     'bionic': null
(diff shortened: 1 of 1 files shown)


cottsay (approved):
The maintainer accepted the PR and built the package, and the update is currently on its way to the Fedora 32 testing repository. It will remain there for 3 days unless it reaches sufficient "karma" through manual community testing before then.

https://bodhi.fedoraproject.org/updates/FEDORA-2020-a5c9af23c1

Since Fedora 32 hasn't even reached beta status, I think it's OK to merge this now, especially if it is holding up Blooming.

Good call on the EOL and null'd entries.

tfoote (approved):

PR #24096, opened 2020-03-16
[Gentoo] Remove dev-python/rosdep

allenh1:
@tfoote Currently, there's a conflict that happens because there's two different rosdep ebuilds. I'm just removing the one the overlay supplies, and going with the upstream (since they tend to be much better about staying up to date than I).

In its current state, it blocks `melodic` from installing. See ros/ros-overlay#966

Signed-off-by: Hunter L. Allen <hunter.allen@ghostrobotics.io>

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3712,7 +3712,7 @@ python-rosdep:
   freebsd:
     pip:
       packages: [rosdep]
-  gentoo: [dev-python/rosdep]
+  gentoo: [dev-util/rosdep]
   openembedded: [python-rosdep@meta-ros]
   opensuse: [python-rosdep]
   osx:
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

flabrosse:
> @tfoote Currently, there's a conflict that happens because there's two different rosdep ebuilds. I'm just removing the one the overlay supplies, and going with the upstream (since they tend to be much better about staying up to date than I).

The one left is the one from the overlay.  Was that intended? It is very old and suffers from the unsafe load().

tfoote:
@allenh1 Can you look at the above?

allenh1:
@tfoote that was the intended fix here, just to remove the overlay supplied key.

I'm going to remove the ebuild from the overlay entirely in the near future.

allenh1:
Oh -- I misread that. I'll look into it. Thanks.

PR #24097, opened 2020-03-16
Add libqglviewer-qt5, libqglviewer-qt5-dev

wxmerkt:
- Arch: Not found - https://aur.archlinux.org/packages.php?K=libqglviewer-qt5
- Fedora: Not found - https://apps.fedoraproject.org/packages/s/libQGLViewer
- Gentoo: Not found - https://packages.gentoo.org/packages/x11-libs/libQGLViewer
- Debian: https://packages.debian.org/search?keywords=libqglviewer
- Ubuntu: https://packages.ubuntu.com/search?keywords=libqglviewer-dev-qt5

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3221,6 +3221,10 @@ libqglviewer-qt4:
     xenial: [libqglviewer2-qt4]
     yakkety: [libqglviewer2-qt4]
     zesty: [libqglviewer2-qt4]
+libqglviewer-qt5:
+  arch: [libqglviewer-qt5]
+  debian: [libqglviewer2-qt5]
+  ubuntu: [libqglviewer2-qt5]
 libqglviewer-qt4-dev:
   arch: [libqglviewer-qt4]
   debian:
@@ -3248,6 +3252,9 @@ libqglviewer-qt4-dev:
     xenial: [libqglviewer-dev-qt4]
     yakkety: [libqglviewer-dev-qt4]
     zesty: [libqglviewer-dev-qt4]
+libqglviewer-qt5-dev:
+  debian: [libqglviewer-dev-qt5]
+  ubuntu: [libqglviewer-dev-qt5]
 libqhull:
   arch: [qhull]
   debian: [libqhull-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3222,7 +3222,6 @@ libqglviewer-qt4:
     yakkety: [libqglviewer2-qt4]
     zesty: [libqglviewer2-qt4]
 libqglviewer-qt5:
-  arch: [libqglviewer-qt5]
   debian: [libqglviewer2-qt5]
   ubuntu: [libqglviewer2-qt5]
 libqglviewer-qt4-dev:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3221,9 +3221,6 @@ libqglviewer-qt4:
     xenial: [libqglviewer2-qt4]
     yakkety: [libqglviewer2-qt4]
     zesty: [libqglviewer2-qt4]
-libqglviewer-qt5:
-  debian: [libqglviewer2-qt5]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -3221,6 +3221,10 @@ libqglviewer-qt4:
     xenial: [libqglviewer2-qt4]
     yakkety: [libqglviewer2-qt4]
     zesty: [libqglviewer2-qt4]
+libqglviewer-qt5:
+  arch: [libqglviewer-qt5]
```
tfoote: In your notes you say this isn't found and I can only find a qt4 version by updating the search.

tfoote (changes requested):

rosdep/base.yaml
wxmerkt: Thanks for spotting. This line was by accident and I've now removed it.

tfoote (approved):
Thanks for the fixup. Looks good to me.

rosdep/base.yaml
```diff
@@ -3221,6 +3221,9 @@ libqglviewer-qt4:
     xenial: [libqglviewer2-qt4]
     yakkety: [libqglviewer2-qt4]
     zesty: [libqglviewer2-qt4]
+libqglviewer-qt5:
```
tfoote: The CI is complaining because this should be after the qt4 rule.

wxmerkt: I'll fix that. I tried the `rosdep_format` tool but it only could get it to work for the distribution files, not the rosdep/base.

rosdep/base.yaml
```diff
@@ -3248,6 +3248,12 @@ libqglviewer-qt4-dev:
     xenial: [libqglviewer-dev-qt4]
     yakkety: [libqglviewer-dev-qt4]
     zesty: [libqglviewer-dev-qt4]
+libqglviewer-qt5:
```
sloretz: I think the rosdep key name should be libqglviewer**2**-qt5 to match the Debian package name. @tfoote thoughts?

tfoote: Yeah, that would be good to match too.

wxmerkt: Done :-)

rosdep/base.yaml
```diff
@@ -3248,6 +3248,12 @@ libqglviewer-qt4-dev:
     xenial: [libqglviewer-dev-qt4]
     yakkety: [libqglviewer-dev-qt4]
     zesty: [libqglviewer-dev-qt4]
+libqglviewer2-qt5:
+  debian: [libqglviewer2-qt5]
+  ubuntu: [libqglviewer2-qt5]
+libqglviewer-qt5-dev:
```
sloretz: oops, one more, I think this should match the debian package name `libqglviewer-dev-qt5` (although the debian package name is very very strange)

wxmerkt: I mirrored what was done for qt4, where the same inconsistency exists. I am happy to fix it for qt5 though to mirror the debian name if desired?

sloretz: I'd mirror the debian name even if the previous key didn't.

Also, it looks like CI is unhappy with the alphabetical ordering after de707bd :(

Thanks for iterating!

wxmerkt: Done and thank you for the detailed feedback :-)

sloretz:
One more fix for CI needed:

```
If this fails you can run 'scripts/clean_rosdep_yaml.py' to help cleanup.

Checking rosdep file: rosdep/base.yaml

checking for trailing spaces...

checking for incorrect indentation...

checking for non-bracket package lists...

checking for item order...

  ERR: list out of alphabetical order line 3251.  'libqglviewer-dev-qt5' should come before 'libqglviewer-qt4-dev'

building yaml dict...

there were errors, please correct the file
```

sloretz (approved):

tfoote (approved):

PR #24167, opened 2020-03-18
Set optris_drivers to maintained, add versions for Melodic and Dashing

w-haag:
The package optris_drivers is to be maintained for both Kinetic and Melodic. Furthermore, the package optris_drivers2 is now available for Dashing.

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -1579,6 +1579,12 @@ repositories:
         release: release/dashing/{package}/{version}
       url: https://github.com/ros-gbp/ompl-release.git
       version: 1.4.2-2
+  optris_drivers:
+    doc:
+      type: git
+      url: https://github.com/evocortex/optris_drivers2.git
+      version: master
+    status: maintained
   orocos_kinematics_dynamics:
     doc:
       type: git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9319,6 +9319,7 @@ repositories:
       type: git
       url: https://github.com/evocortex/optris_drivers.git
       version: kinetic-devel
+    status: maintained
   orb_slam2_ros:
     doc:
       type: git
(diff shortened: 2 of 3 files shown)


dashing/distribution.yaml
```diff
@@ -1579,6 +1579,12 @@ repositories:
         release: release/dashing/{package}/{version}
       url: https://github.com/ros-gbp/ompl-release.git
       version: 1.4.2-2
+  optris_drivers:
+    doc:
+      type: git
+      url: https://github.com/evocortex/optris_drivers2.git
+      version: master
+    status: maintained
```
nuclearsandwich: Just a head's up @w-haag there are not currently doc jobs for packages in the Dashing distribution. Would you like to add a source entry as well?

w-haag: Thank you for the heads-up. As the optris_drivers need an additional library to compile, I'd like to take care of rosdep first and add the source entry once the prerelease tests finish successfully.

nuclearsandwich (approved):

clalancette (changes requested):
@w-haag A few overall things:

1.  In general, it's a bad idea to modify multiple ROS distributions in the same pull request.  The reason is that the distros are all on different sync schedules, so it can be difficult to find a time when they are all out of freeze.  I won't insist on it right now, but please split them up in the future.
1.  https://github.com/evocortex/optris_drivers.git lists BSD as the license in the package.xml, but is missing a LICENSE file.
1.  https://github.com/evocortex/optris_drivers2.git doesn't list any license in the package.xml, and is missing a LICENSE file.

Please correct the latter two, at least, before we move forward with this.  Thanks.

w-haag:
@clalancette 

1. Acknowledged. I'm closing this pull request and redoing it the proper way.
2. Fixed.
3. Fixed.

PR #24171, opened 2020-03-19
Add gstreamer1.0-rtsp and libgstrtspserver-1.0 packages

mortenfyhn:
gstreamer1.0-rtsp:
https://packages.ubuntu.com/bionic/gstreamer1.0-rtsp
https://packages.debian.org/buster/gstreamer1.0-rtsp
(Can't find on fedora, ditto for all below.)

libgstrtspserver-1.0-0:
https://packages.ubuntu.com/bionic/libgstrtspserver-1.0-0
https://packages.debian.org/buster/libgstrtspserver-1.0-0

libgstrtspserver-1.0-0-dbg:
https://packages.ubuntu.com/bionic/libgstrtspserver-1.0-0-dbg
https://packages.debian.org/buster/libgstrtspserver-1.0-0-dbg

libgstrtspserver-1.0-dev:
https://packages.ubuntu.com/bionic/libgstrtspserver-1.0-dev
https://packages.debian.org/buster/libgstrtspserver-1.0-dev

libgstrtspserver-1.0-doc:
https://packages.ubuntu.com/bionic/libgstrtspserver-1.0-doc
https://packages.debian.org/buster/libgstrtspserver-1.0-doc

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2414,15 +2414,9 @@ libgstreamer1.0-dev:
 libgstrtspserver-1.0-0:
   debian: [libgstrtspserver-1.0-0]
   ubuntu: [libgstrtspserver-1.0-0]
-libgstrtspserver-1.0-0-dbg:
-  debian: [libgstrtspserver-1.0-0-dbg]
-  ubuntu: [libgstrtspserver-1.0-0-dbg]
 libgstrtspserver-1.0-dev:
   debian: [libgstrtspserver-1.0-dev]
   ubuntu: [libgstrtspserver-1.0-dev]
-libgstrtspserver-1.0-doc:
-  debian: [libgstrtspserver-1.0-doc]
-  ubuntu: [libgstrtspserver-1.0-doc]
 libgtkmm:
   arch: [gtkmm]
   debian: [libgtkmm-2.4-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2411,6 +2411,18 @@ libgstreamer1.0-dev:
   fedora: [gstreamer1-devel]
   gentoo: ['media-libs/gstreamer:1.0']
   ubuntu: [libgstreamer1.0-dev]
+libgstrtspserver-1.0-0:
+  debian: [libgstrtspserver-1.0-0]
+  ubuntu: [libgstrtspserver-1.0-0]
+libgstrtspserver-1.0-0-dbg:
+  debian: [libgstrtspserver-1.0-0-dbg]
+  ubuntu: [libgstrtspserver-1.0-0-dbg]
+libgstrtspserver-1.0-dev:
+  debian: [libgstrtspserver-1.0-dev]
+  ubuntu: [libgstrtspserver-1.0-dev]
+libgstrtspserver-1.0-doc:
+  debian: [libgstrtspserver-1.0-doc]
+  ubuntu: [libgstrtspserver-1.0-doc]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -2408,6 +2411,18 @@ libgstreamer1.0-dev:
   fedora: [gstreamer1-devel]
   gentoo: ['media-libs/gstreamer:1.0']
   ubuntu: [libgstreamer1.0-dev]
+libgstrtspserver-1.0-0:
+  debian: [libgstrtspserver-1.0-0]
+  ubuntu: [libgstrtspserver-1.0-0]
+libgstrtspserver-1.0-0-dbg:
+  debian: [libgstrtspserver-1.0-0-dbg]
+  ubuntu: [libgstrtspserver-1.0-0-dbg]
```
clalancette: We don't typically add the `-dbg` packages to rosdep keys; it's just not something that downstream consumers are going to depend on.  So please remove this.

rosdep/base.yaml
```diff
@@ -2408,6 +2411,18 @@ libgstreamer1.0-dev:
...
+libgstrtspserver-1.0-dev:
+  debian: [libgstrtspserver-1.0-dev]
+  ubuntu: [libgstrtspserver-1.0-dev]
+libgstrtspserver-1.0-doc:
+  debian: [libgstrtspserver-1.0-doc]
+  ubuntu: [libgstrtspserver-1.0-doc]
```
clalancette: Same goes for documentation packages; it's just not something that we'll need to build or on the buildfarm.  So please remove this.

clalancette (changes requested):
Two things to fix inline.

Also, it looks like Fedora has gstreamer1-rtsp-server: https://apps.fedoraproject.org/packages/gstreamer1-rtsp-server .  So please add that key as well (though it is unclear to me whether it goes under `gstreamer1.0-rtsp` or `libgstrtspserver-1.0-0`).

rosdep/base.yaml
mortenfyhn: Makes sense, will fix.

rosdep/base.yaml
mortenfyhn: Also makes sense.

mortenfyhn:
I think Fedora's `gstreamer1-rtsp-server-devel` is equivalent to Ubuntu's `libgstrtspserver-1.0-dev`, based on similar contents / file lists, so I'll add it under the `libgstrtspserver-1.0-dev` key.

rosdep/base.yaml
```diff
@@ -2408,6 +2411,13 @@ libgstreamer1.0-dev:
   fedora: [gstreamer1-devel]
   gentoo: ['media-libs/gstreamer:1.0']
   ubuntu: [libgstreamer1.0-dev]
+libgstrtspserver-1.0-0:
+  debian: [libgstrtspserver-1.0-0]
+  ubuntu: [libgstrtspserver-1.0-0]
```
clalancette: Add:

```
fedora: [gstreamer1-rtsp-server]
```

clalancette (changes requested):
One more fix, then I think this is good.  Thanks for iterating.

rosdep/base.yaml
mortenfyhn: Oh yeah, that seems right.

clalancette (approved):
Looks good to me, thanks!

nuclearsandwich (approved):

PR #24193, opened 2020-03-22
add focal version for some base packages

flixr:
Add ubuntu distro `focal` for some base packages:
* libfcl-dev
* libgsl
* libpng-dev
* protobuf

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2011,12 +2011,9 @@ libfcl-dev:
   gentoo: [sci-libs/fcl]
   openembedded: [fcl@meta-ros]
   ubuntu:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
These all look like they have been stable across the last 2-3 LTS releases. To that end lets switch to use a wildcard notation for everything that's the same as the current release and only call out the older ones on the expectation that they will continue to remain static and will save us future maintenance.

flixr:
good idea, done...

tfoote (approved):
Thanks for the update. lgtm

clalancette (approved):

PR #24196, opened 2020-03-22
kobuki_firmware: 1.2.0-1 in 'eloquent/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `kobuki_firmware` to `1.2.0-1`:

- upstream repository: https://github.com/kobuki-base/kobuki_firmware.git
- release repository: https://github.com/stonier/kobuki_firmware-release.git
- bloom version: `0.9.0`

## kobuki_firmware

```
* Custom PID gain setting of wheel velocity controlled added
```

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -967,6 +967,17 @@ repositories:
       url: https://github.com/kobuki-base/kobuki_core.git
       version: devel
     status: maintained
+  kobuki_firmware:
+    doc:
+      type: git
+      url: https://github.com/kobuki-base/kobuki_firmware.git
+      version: release/1.2.x
+    release:
+      tags:
+        release: release/eloquent/{package}/{version}
+      url: https://github.com/stonier/kobuki_firmware-release.git
+      version: 1.2.0-1
+    status: maintained
   kobuki_ftdi:
     doc:
       type: git


eloquent/distribution.yaml
```diff
@@ -967,6 +967,17 @@ repositories:
       url: https://github.com/kobuki-base/kobuki_core.git
       version: devel
     status: maintained
+  kobuki_firmware:
+    doc:
+      type: git
+      url: https://github.com/kobuki-base/kobuki_firmware.git
```
clalancette: The package.xml lists the license as BSD for this repository, but it is distributing stm32flash (which is GPL, judging from https://sourceforge.net/p/stm32flash/code/ci/master/tree/main.c), and the kobuki firmware, which is under some other license (I'm not sure which one).  Questions:

1.  I believe we want to change the license field to GPL (at least).
1.  What's the license on the kobuki firmware?  Do we have permission to redistribute it?

clalancette (changes requested):
Some open questions to answer here.

eloquent/distribution.yaml
stonier: * There is no license on the kobuki firmware (we were naive engineers :)). In spirit also, we had no intention of blocking distribution of the binaries.
* Is it possible for the package to be of one license and internal parts thereof be of another license, so long as that license is installed in `share`?

clalancette:
@stonier and I discussed this offline.  He's going to go back and make sure those binaries are redistributable, and then report back here.  For now, I'm going to close this PR out.

PR #24210, opened 2020-03-24
Renaming pacifica_dbw_ros to raptor_dbw_ros

effi-engineering:
This package is being updated and renamed. Thank you!


tfoote (changes requested):
With the rename it needs to be moved lower in the list to keep alphabetical ordering.

tfoote (approved):
Thanks for fixing the ordering

PR #24275, opened 2020-03-28
Add a rosdep rule for libopenal-soft

sktometometo:
add libopenal-soft rule for arch, debian, fedora, gentoo and ubuntu.
- arch: https://www.archlinux.org/packages/extra/x86_64/openal/
- debian: https://packages.debian.org/ja/sid/libopenal-dev
- fedora: https://apps.fedoraproject.org/packages/openal-soft
- gentoo: https://packages.gentoo.org/packages/media-libs/openal
- ubuntu: https://packages.ubuntu.com/ja/xenial/libopenal-dev

This library is for playing spatial audio, and I am developing a ROS interface to it.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2826,6 +2826,12 @@ libopal-dev:
   fedora: [opal-devel]
   gentoo: [net-libs/opal]
   ubuntu: [libopal-dev]
+libopenal-dev:
+  arch: [openal]
+  debian: [libopenal-dev]
+  fedora: [openal-soft]
+  gentoo: [media-libs/openal]
+  ubuntu: [libopenal-dev]
 libopenblas-dev:
   debian: [libopenblas-dev]
   fedora: [openblas-devel]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -2826,6 +2826,12 @@ libopal-dev:
   fedora: [opal-devel]
   gentoo: [net-libs/opal]
   ubuntu: [libopal-dev]
+libopenal-dev:
+  arch: [openal]
+  debian: [libopenal-dev]
+  fedora: [openal-soft]
```
clalancette: For Fedora, I think you want `openal-soft-devel` to get the development packages.

clalancette (changes requested):
One change, then I'm happy to approve.

tfoote (changes requested):
Same comment as @clalancette

sktometometo:
changed fedora entry to openal-soft-devel. Thank you for reviewing.

clalancette (approved):
You may want to consider adding a `libopenal` key that just has the runtime library.  That way your package can do a `<build_depend>` on `libopenal-dev`, and an `<exec_depend>` on `libopenal`.  But that doesn't block this change from going in, so approving.  Thanks.

tfoote (approved):

PR #24292, opened 2020-03-31
Added python3-requests to bionic

klintan:
For this PR https://github.com/ros-perception/image_pipeline/pull/447 python3 `requests` library is needed to download the test data. Unfortunately it is not available for py3 in bionic (which the testing image is using). This PR adds that.

Let me know if there is anything that needs to change.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3709,7 +3709,6 @@ python-requests:
       packages: [requests]
   ubuntu:
     '*': [python-requests]
-    bionic_python3: [python3-requests]
     trusty_python3: [python3-requests]
     utopic_python3: [python3-requests]
     vivid_python3: [python3-requests]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5706,6 +5706,9 @@ python3-rafcon-pip:
   ubuntu:
     pip:
       packages: [rafcon]
+python3-requests:
+  ubuntu:
+    bionic: [python3-requests]
 python3-retrying:
   arch: [python-retrying]
   debian: [python3-retrying]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3709,14 +3709,6 @@ python-requests:
       packages: [requests]
   ubuntu:
     '*': [python-requests]
-    bionic_python3: [python3-requests]
-    trusty_python3: [python3-requests]
-    utopic_python3: [python3-requests]
-    vivid_python3: [python3-requests]
-    wily_python3: [python3-requests]
-    xenial_python3: [python3-requests]
-    yakkety_python3: [python3-requests]
-    zesty_python3: [python3-requests]
 python-requests-oauthlib:
   debian: [python-requests-oauthlib]
   fedora: [python-requests-oauthlib]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3709,6 +3709,7 @@ python-requests:
       packages: [requests]
   ubuntu:
     '*': [python-requests]
+    bionic_python3: [python3-requests]
```
clalancette: We've deprecated this method of doing python3 keys; we found it did not work very well.  Instead, we are now using separate keys for python2 and python3; see https://github.com/ros/rosdistro/blob/4a422638d8506f2930bbd2d00993a0522425fe72/rosdep/python.yaml#L5325 for one example on how to do it.  The `python3-requests` is indeed missing, so please make it look like the other python3 keys.  Thanks.

clalancette (changes requested):

rosdep/python.yaml
klintan: Thanks, see the new changes. I moved the other `python3-requests` entries to conform to this new way as well. Let me know if that is ok or if you want me to change those back.

rosdep/python.yaml
```diff
@@ -3709,13 +3709,6 @@ python-requests:
       packages: [requests]
   ubuntu:
     '*': [python-requests]
-    trusty_python3: [python3-requests]
-    utopic_python3: [python3-requests]
-    vivid_python3: [python3-requests]
-    wily_python3: [python3-requests]
-    xenial_python3: [python3-requests]
-    yakkety_python3: [python3-requests]
-    zesty_python3: [python3-requests]
```
clalancette, line 3718: I don't think we should remove these; it is possible that someone is depending on them, even though this method is deprecated.

tfoote: Yeah, we'd prefer to keep these there for compatibility in case someone is using them. We'll clean them out in the future when we do a systematic pruning of old distros.

klintan: Of course, changed it back.

Just the new changes in there now.

flixr:
How about using a wildcard and also adding it for debian?

rosdep/python.yaml
```diff
@@ -5705,6 +5705,9 @@ python3-rafcon-pip:
   ubuntu:
     pip:
       packages: [rafcon]
+python3-requests:
+  ubuntu:
+    bionic: [python3-requests]
```
clalancette, line 5710: @flixr Has a good point here; python3-requests is [available on Ubuntu](https://packages.ubuntu.com/search?keywords=python3-requests&searchon=names&suite=all&section=all) going all the way back to Xenial, so a wildcard would be best here.

Same goes for Debian, and Fedora.

clalancette (changes requested):

rosdep/python.yaml
tfoote: Xenial is the oldest supported platform so we don't need to do the wildcard syntax, just the straight 'ubuntu' rule without calling out the version.

tfoote (changes requested):
We should definitely have debian rules for this and fedora too. The other platforms should be easy to add too since we have the equivalent python2 versions of all the packages. If they don't have a python3 version then we can omit it.

klintan:
> We should definitely have debian rules for this and fedora too. The other platforms should be easy to add too since we have the equivalent python2 versions of all the packages. If they don't have a python3 version then we can omit it.

I check latest debian and fedora and they both have `python3-requests` also fixed the ubuntu "wildcard" aka enabling it for all supported releases. Let me know if there is anything else of is misunderstood something.

clalancette (approved):

tfoote (approved):
Thanks for iterating! It looks good to me.

PR #24293, opened 2020-03-31
Adding new repository for indexing: robosense_simulator

tomlogan501:
Hi,
I create a new package to simulate the robosense sensors (only RS-Lidar-16) for now.
 I used as reference the package velodyne_simulator

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -8024,7 +8024,7 @@ repositories:
   robosense_simulator:
     doc:
       type: git
-      url: https://github.com/tomlogan501/robosense_simulator
+      url: https://github.com/tomlogan501/robosense_simulator.git
       version: master
     release:
       packages:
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -8021,6 +8021,26 @@ repositories:
...
+      url: https://github.com/tomlogan501/robosense_simulator
+      version: master
+    release:
+      packages:
+      - robosense_description
+      - robosense_gazebo_plugins
+      - robosense_simulator
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/tomlogan501/robosense_simulator
```
clalancette: The release repository should never be the same repository as the sources; it makes it very difficult to work with.  Also, it looks like there are no release branches on that repository anyway, so this can't work.  Did you follow the guide at http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease for releasing your package?

clalancette (changes requested):

melodic/distribution.yaml
tomlogan501: Thanks for your reply, I didn't follow this tutorial. I just fork rosdistro using the tutorial
I will push the changes.

tfoote: Note to just be indexed and not make releases there's a different tutorial: http://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation

tomlogan501: Hi, I made the changes for both the release repo and this file.
Should be good. Sorry it is my first ROS package release.
Should I open a discussion about it in ros discourse for a follow up ?

tomlogan501:
Closed in favor of #24325

PR #24296, opened 2020-03-31
Adding omt_coscan to documentation index for distro

siyandong:
I'd like omt_coscan to be indexed and documented on ros.org.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6251,6 +6251,16 @@ repositories:
       url: https://github.com/ose-support-ros/omronsentech_camera.git
       version: master
     status: developed
+  omt_coscan:
+    doc:
+      type: git
+      url: https://github.com/siyandong/CoScan.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/siyandong/CoScan.git
+      version: master
+    status: maintained
   open_karto:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -6251,6 +6251,16 @@ repositories:
...
       version: master
     status: developed
+  omt_coscan:
+    doc:
+      type: git
+      url: https://github.com/siyandong/CoScan.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/siyandong/CoScan.git
```
clalancette: The source repository here needs a LICENSE file, and to have the license declared in the package.xml files.

Also, the directory layout is pretty non-standard for a ROS package.  That doesn't mean it won't work, but it will look odd to people accustomed to ROS.  To make it more like ROS, I'd suggest moving all of the package directories (src/co_scan, src/srbot_description, src/virtual_scan) into the top-level of the source repository, and then removing the dummy src/CMakeLists.txt .  That will make it look a lot more standard.

clalancette (changes requested):

melodic/distribution.yaml
siyandong: I'm sorry for the non-standard layout. I have reorganized the directories and added a LICENSE file.

tfoote (approved):
Thanks for the updates. Rerequesting review from @clalancette

melodic/distribution.yaml
clalancette: The repository layout looks a lot more standard, thanks for iterating.  Also thanks for adding a LICENSE file.

The last thing missing is declaring the `<license>` tag in each of the package.xml files.  Since your top-level LICENSE file is MIT, that would look like:

```
<license>MIT</license>
```

in each of the three package.xml files.  Once that is done, I'm happy to approve and merge.

clalancette (changes requested):

melodic/distribution.yaml
siyandong: I add the license declaration in three .xml files. Thanks for your approval!

siyandong:
Thanks! I added the license declaration in the three package.xml files.

Best,
Siyan

Chris Lalancette <notifications@github.com> 于2020年4月1日周三 下午8:20写道：

clalancette (approved):

PR #24311, opened 2020-04-01
Added python for focal

ahcorde:
according with this [failure](http://build.ros.org/job/Npr__vision_opencv__ubuntu_focal_amd64/3/console): added python for focal

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -373,6 +373,7 @@ python:
   ubuntu:
     artful: [python-dev]
     bionic: [python-dev]
+    focal: [python3-dev]
     lucid: [python-dev]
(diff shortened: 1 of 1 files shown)


sloretz (dismissed):
LGTM https://packages.ubuntu.com/focal/python3-dev

sloretz (changes requested):
Please use the `python3` key instead. In general there are separate keys for Python 2 and Python 3, mostly matching upstream Debian packages.

https://github.com/ros/rosdistro/blob/8c633d40f3b5abc0b2b5e8f0f04c7c78f4c1e243/rosdep/python.yaml#L5148-L5152

PR #24312, opened 2020-04-01
Added python-psutil for focal

ahcorde:
According with this [failured](http://build.ros.org/job/Npr__diagnostics__ubuntu_focal_amd64/1/console). There is no entrance for focal


sloretz (changes requested):
Please use the key `python3-psutil` instead

https://github.com/ros/rosdistro/blob/8c633d40f3b5abc0b2b5e8f0f04c7c78f4c1e243/rosdep/python.yaml#L5545-L5559

PR #24333, opened 2020-04-02
Switch to use GitHub Actions from Travis

tfoote:
We've been having issues with Travis reporting. I think that this should be a drop in replacement.
To merge this we'll also have to update the required status checks on the protected branch master.


clalancette:
Yes, yes, yes!  Thanks @tfoote!  Will review.

.github/workflows/build_test.yaml
```diff
@@ -0,0 +1,32 @@
+name: Build Test
+on: [push]
+
+jobs:
+  build:
+    name: Validate Changes
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [2.7, 3.8]
```
clalancette, line 10: Do we still need to bother with Python 2.7?  I know we historically had it there, but I'm not sure there is much value in it anymore.  Thoughts?

.github/workflows/build_test.yaml
```diff
@@ -0,0 +1,32 @@
...
+        pip install PyYAML argparse
+        pip install catkin-pkg ros-buildfarm rosdistro nose coverage
+        pip install yamllint
+        pip install unidiff
+        pip install rosdep
+        pip install PyGithub
+    - name: Run yamllint
+      run: yamllint */
+    - name: Run Nose Tests
+      run: nosetests -s
```
clalancette, line 32: This is probably just my unfamiliarity with Github actions/rosdistro tests, but how does just fetching the master data work?  Do the tests themselves checkout different branches and do the diffs?

Pretty minor, but in the travis code, we had nosetests first, followed by yamllint.  I'm not sure if we trip up one or the other more often.

Also, nit: newline at the end of the file.

clalancette (commented):
Looking at the output from the checks, one thing that was nicer about travis is the collapsible blocks for the output.  Is something like that possible in Github actions?  It's not a huge deal either way, but it would just be easier to review failures.

.github/workflows/build_test.yaml
tfoote: I'm happy to remove it. I don't think we really need it long term. But I tried to change as little as possible to make sure we're still getting the same results.

.github/workflows/build_test.yaml
tfoote: RE: Fetch

We can't afford to check all the urls for validity so the unit tests computes the diff to `origin/master` and then only validates changed and added urls.  This is also how we detect modifications of EOL distros etc. The checkout in in the actions is stripped down more, on travis the whole of origin was available without explicitly requesting it. 

Re: Ordering

I moved the yamllint first because it takes much less time to run. So I figured to run it first to be able to give quicker feedback if it failed. Maybe I can run it in parallel with a separate check.

I'll add the new line.

.github/workflows/build_test.yaml
clalancette: All right, that's fine.  We can leave it for now, and remove it in a separate PR.

clalancette (approved):
This looks good to me.  It's probably a good idea to get another person to take a quick perusal before merging.  Also, I think we'll need to disable the travis "hook" somehow, since it still seems to be running (even with .travis.yml removed here).

nuclearsandwich:
Great minds [think alike](https://github.com/ros-infrastructure/rosdep/pull/751). I like splitting linting and other tests. One thing I noticed when experimenting with Windows builds in the matrix is that the default runners seem to abort all builds when the first one fails. I haven't looked for how to change that behavior yet so I don't know if we can.

.github/workflows/build_test.yaml
```diff
@@ -0,0 +1,29 @@
+name: Nosetests
+on: [push]
+
+jobs:
+  build:
+    name: Validate Changes
+    runs-on: ubuntu-latest
```
nuclearsandwich, line 7: Is it worth pinning this to ubuntu-18.04? I don't like surprises.

.github/workflows/build_test.yaml
```diff
@@ -0,0 +1,29 @@
...
+        python-version: [2.7, 3.8]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Fetch master (to enable diff)
+      run: |
+        git fetch origin master
```
nuclearsandwich, line 19: actions/checkout has a fetch-depth option that we could use here. A value of '0' would give us a full checkout but '2-N' for some small N might do exactly what we need depending on the behavior of PR setups.

nuclearsandwich (approved):
One question and one suggestion/alternative. Neither is blocking and this is way cool.

.github/workflows/build_test.yaml
tfoote: I looked through the docs and if you want to fetch other branches they point to calling fetch directly. but switching to the shallow fetch with `--depth` cuts it down to a zero second fetch since it's typically just a ref just behind the existing PR.

.github/workflows/build_test.yaml
tfoote: Yeah, we might as well be explicit.

tfoote:
When I went to update the protected branches I noticed the naming was all off so I cleaned that up too. We can give it a shot. it looks like with the required tests switched that outstanding PRs will need a manual review to make sure travis passed. But that count should drop away quickly.

PR #24338, opened 2020-04-03
Added source for executive_smach

ahcorde:
Added source for executive_smach

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -265,6 +265,13 @@ repositories:
       url: https://github.com/ros/eigen_stl_containers.git
       version: master
     status: maintained
+  executive_smach:
+    source:
+      test_pull_requests: true
+      type: git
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -265,6 +265,13 @@ repositories:
       url: https://github.com/ros/eigen_stl_containers.git
       version: master
     status: maintained
+  executive_smach:
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/ros/executive_smach.git
+      version: noetic-devel
```
clalancette, line 273: It looks like the branch is called "noetic_devel" on the source repository (which is why this failed CI).  Personally, I'd suggest renaming the source branch to "noetic-devel", as that is a lot more idiomatic.

ahcorde: updated

PR #24346, opened 2020-04-03
Add pyrealsense2 and open3d pip packages for ubuntu/fedora/debian

luisrayas3:
Package documentations:
https://pypi.org/project/pyrealsense2/
https://pypi.org/project/open3d-python/

Use case:
We are using realsense cameras in our ros project and open3d for processing the data.

Test on Ubuntu with:
```
$ rosdep update && rosdep resolve open3d-pip pyrealsense2-pip
#ROSDEP[open3d-pip]
#pip
open3d-python
#ROSDEP[pyrealsense2-pip]
#pip
pyrealsense2
```

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3400,6 +3400,16 @@ python-pyramid:
   fedora: [python-pyramid]
   gentoo: [dev-python/pyramid]
   ubuntu: [python-pyramid]
+python-pyrealsense2-pip:
+  debian:
+    pip:
+      packages: [pyrealsense2]
+  fedora:
+    pip:
+      packages: [pyrealsense2]
+  ubuntu:
+    pip:
+      packages: [pyrealsense2]
 python-pyside:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -219,6 +219,10 @@ opcua-pip:
   ubuntu:
     pip:
       packages: [opcua]
+open3d-pip:
```
clalancette, line 222: For both of these, our style is `python-open3d-pip` (so we can also add a python3 key as needed).  Also, since this is pip, we may as well add it for Fedora and debian as well.

clalancette (changes requested):

clalancette:
Closing and reopening to kick Github actions.

rosdep/python.yaml
luisrayas3: done

clalancette (approved):

tfoote (approved):

PR #24370, opened 2020-04-06
rc_dynamics_api: 0.10.2-1 in 'dashing/distribution.yaml' [bloom]

flixr:
Increasing version of package(s) in repository `rc_dynamics_api` to `0.10.2-1`:

- upstream repository: https://github.com/roboception/rc_dynamics_api.git
- release repository: https://github.com/roboception-gbp/rc_dynamics_api-release.git
- bloom version: `0.9.3`

## rc_dynamics_api

```
* remove build_export_depend on catkin from package.xml
```

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -1917,6 +1917,22 @@ repositories:
       url: https://github.com/ros-visualization/qt_gui_core.git
       version: crystal-devel
     status: maintained
+  rc_dynamics_api:
+    doc:
+      type: git
+      url: https://github.com/roboception/rc_dynamics_api.git
+      version: master
+    release:
+      tags:
+        release: release/dashing/{package}/{version}
+      url: https://github.com/roboception-gbp/rc_dynamics_api-release.git
+      version: 0.10.2-1
+    source:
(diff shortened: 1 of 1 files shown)


dashing/distribution.yaml
```diff
@@ -1917,6 +1917,22 @@ repositories:
       url: https://github.com/ros-visualization/qt_gui_core.git
       version: crystal-devel
     status: maintained
+  rc_dynamics_api:
+    doc:
+      type: git
+      url: https://github.com/roboception/rc_dynamics_api.git
+      version: master
```
clalancette, line 1924: I see that both the ROS 1 and ROS 2 versions of this package point to the master branch of the repository.  Does it work for both of them simultaneously?

flixr: yes, as that is a non-ROS third party package that is then used in separate ROS1 and ROS2 packages as system dependency...

clalancette: Sounds good, just checking.

flixr: Thanks for staying vigilant :-)

PR #24407, opened 2020-04-07
Update eigenpy to 2.3.0-3 to include patch for installation to share/cmake

wxmerkt:
Implements suggestion by @rhaschke in #24341 - thanks for the detailed analysis!

This PR uses patches to change the `INSTALL_DESTINATION` of `eigenpyConfig.cmake` to `share/cmake/eigenpy`.

It also explicitly sets the `CMAKE_BUILD_TYPE` to `Release`. It appears that the buildfarm by default sets `-DCMAKE_BUILD_TYPE=None`.


clalancette:
@wxmerkt @rhaschke Thanks for looking into this.  Let's give it a try...

PR #24410, opened 2020-04-07
Update eigenpy to 2.3.0-4 to fix-up patch

wxmerkt:
I didn't realise the variable for `INSTALL_DESTINATION` gets set in two different places - this PR addresses this. I also accidentally installed to `share/cmake/{package}` instead of `share/{package}/cmake`. This should hopefully resolve these final issues.

Fix-up for #24407


wxmerkt:
This worked - cf. http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__moveit_ros_planning_interface__ubuntu_bionic_amd64__binary/51/

Thank you for all your patience and input in getting this resolved @rhaschke , @clalancette.

PR #24416, opened 2020-04-07
Add new dependencies for MacOS X

jpardeiro:


rosdep/sources.list.d/20-default.list
```diff
@@ -1,10 +1,10 @@
 # os-specific listings first
-yaml https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/osx-homebrew.yaml osx
+yaml https://raw.githubusercontent.com/jpardeiro/rosdistro/add_deps/rosdep/osx-homebrew.yaml osx
```
nuclearsandwich, line 2: The changes to this file likely shouldn't be included in this PR

nuclearsandwich (changes requested):

rosdep/sources.list.d/20-default.list
jpardeiro: You are absolutely right, sorry for the mistake. I just reverted the lines and rebased the branch,

nuclearsandwich:
Question for @ros/rosdeputies, since Homebrew on Mac has a policy of not re-packaging packages provided by language-specific package managers, there won't ever be official homebrew packages for the python dependencies listed here. On other platforms we've been pushing pip-installed packages into separate keys. Do we want to do that here as well or should we take the python contributions as they are integrated into the system keys?

tfoote (approved):
@nuclearsandwich Since the pip installation is the only expected source there's not a worry about masking the system installation accidentally. And there's also not an issue with being able to build "native" packages with the dependencies so we can let the osx rules use pip to resolve the regular keys.

clalancette (approved):

nuclearsandwich:
Thanks. Appreciate the clarification.

PR #24337, opened 2020-04-03
euslime: 1.0.1-1 in 'melodic/distribution.yaml' [bloom]

Affonso-Gui:
Increasing version of package(s) in repository `euslime` to `1.0.1-1`:

- upstream repository: https://github.com/jsk-ros-pkg/euslime.git
- release repository: https://github.com/jsk-ros-pkg/euslime-release.git
- distro file: `melodic/distribution.yaml`
- bloom version: `0.9.3`
- previous version for package: `null`

## euslime

```
* First public release for melodic
* Contributors: Yuki Furuta, Guilherme Affonso
```

PR #24421, opened 2020-04-08
Add python-sexpdata to rosdep

Affonso-Gui:
Adding [sexpdata](https://pypi.org/project/sexpdata/), a python S-expression parser needed in the recently added [euslime](https://github.com/jsk-ros-pkg/euslime) package. 
https://github.com/ros/rosdistro/pull/24337

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4137,16 +4137,21 @@ python-setuptools:
     zesty: [python-setuptools]
 python-sexpdata:
   debian:
-    '*':
+    '*': [python-sexpdata]
+    jessie:
       pip:
         packages: [sexpdata]
-    buster: [python-sexpdata]
+    stretch:
+      pip:
+        packages: [sexpdata]
   ubuntu:
-    '*':
+    '*': [python-sexpdata]
+    bionic:
       pip:
         packages: [sexpdata]
-    disco: [python-sexpdata]
-    eoan: [python-sexpdata]
+    xenial:
+      pip:
+        packages: [sexpdata]
 python-sh:
   debian: [python-sh]
   fedora: [python-sh]
@@ -5838,9 +5843,6 @@ python3-sexpdata:
   ubuntu:
     '*': [python3-sexpdata]
     bionic:
-      pip:
-        packages: [sexpdata]
-    trusty:
       pip:
         packages: [sexpdata]
     xenial:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4136,9 +4136,17 @@ python-setuptools:
     yakkety: [python-setuptools]
     zesty: [python-setuptools]
 python-sexpdata:
+  debian:
+    '*':
+      pip:
+        packages: [sexpdata]
+    buster: [python-sexpdata]
   ubuntu:
-    pip:
-      packages: [sexpdata]
+    '*':
+      pip:
+        packages: [sexpdata]
+    disco: [python-sexpdata]
+    eoan: [python-sexpdata]
 python-sh:
   debian: [python-sh]
   fedora: [python-sh]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
A python-sexpdata package is available for [buster](https://packages.debian.org/buster/python-sexpdata) and for disco and eoan but it's not present in Bionic although the python3 variant is available in upcoming distros like bullseye and focal. I'm not sure whether we should switch this to a `-pip` key or keep the key name and backfill with pip on platforms where the dependency is unavailable.

clalancette:
> keep the key name and backfill with pip on platforms where the dependency is unavailable.

I think we should keep the key name here and use the system package where possible, then add a separate key for the pip stuff.

Affonso-Gui:
If I understand correctly something like this?
```
python-sexpdata:
  debian:
    buster: [python-sexpdata]
  ubuntu:
    focal: [python3-sexpdata]

sexpdata-pip:
  debian:
    pip:
      packages: [sexpdata]
  ubuntu:
    pip:
      packages: [sexpdata]
```

nuclearsandwich:
> I think we should keep the key name here and use the system package where possible, then add a separate key for the pip stuff.

This would require a package to know whether the dependency is available on its target platforms in order to know which key to use. Whereas a one-key approach would let the system package be used where it becomes available but fall back to the pip key without having to change source. I don't have a strong opinion on which way we go but if your intent is to use the system package "where possible" that isn't easy if the pip dependency comes from a separate rosdep key.

> If I understand correctly something like this?

That's close but we still need separate keys for python2 vs python3 and for pip keys we still maintain the python- prefix.

```yaml
python-sexpdata:
  debian:
    buster: [python-sexpdata]
# ...
python3-sexpdata
  debian:
    '*': python3-sexpdata
    jessie: null
    stretch: null
  ubuntu:
    focal: [python3-sexpdata]
# ...
python-sexpdata-pip:
  debian:
    pip:
      packages: [sexpdata]
  ubuntu:
    pip:
      packages: [sexpdata]
```

Affonso-Gui:
In my opinion sticking to the python-sexpdata key while back-filling with pip does seem more natural.

I tried updating the PR following this approach, please let me know of any corrections or in the case we conclude that adding the pip key is for the best.

tfoote:
Note that for ubuntu we should use the wildcard with explicit nulls for earlier versions back to bionic. 

And I would actually advocate against having the alternative pip rule. We should fill in the nulls with the pip rule on platforms that don't have binary installations. However we should not create a parallel rule that people could accidentally install the pip version on one of the supported platforms. This can lead to peoples systems getting the pip overlay when some package they try to use declares the pip dependency, and will potentially interfere with a properly declared dependency on the non-pip version.

Affonso-Gui:
> Note that for ubuntu we should use the wildcard with explicit nulls for earlier versions back to bionic.

In this case shouldn't we fill in with pip rules rather than with explicit nulls?

Affonso-Gui:
Any more comments here? I need this dependency to fix the build for my package...

rosdep/python.yaml
```diff
@@ -4135,6 +4135,18 @@ python-setuptools:
     xenial_python3: [python3-setuptools]
     yakkety: [python-setuptools]
     zesty: [python-setuptools]
+python-sexpdata:
+  debian:
+    '*':
+      pip:
+        packages: [sexpdata]
+    buster: [python-sexpdata]
```
clalancette, line 4143: I'm sorry to keep iterating here, but I actually think we should reverse this.  That way it will be more future proof.  Something like this instead:

```
    debian:
        '*': [python-sexpdata]
        jessie:
            pip:
                packages: [sexpdata]
       stretch:
           pip:
               packages: [sexpdata]
```

rosdep/python.yaml
```diff
@@ -4135,6 +4135,18 @@ python-setuptools:
...
+  ubuntu:
+    '*':
+      pip:
+        packages: [sexpdata]
+    disco: [python-sexpdata]
+    eoan: [python-sexpdata]
```
clalancette, line 4149: Same goes here; we should make the package the '*' (so we will get it automatically for future Ubuntu releases), and just call out the still supported ones with pip (in this case, it's just 'xenial' and 'bionic').

rosdep/python.yaml
```diff
@@ -5814,6 +5826,26 @@ python3-setuptools:
...
+      pip:
+        packages: [sexpdata]
+  ubuntu:
+    '*': [python3-sexpdata]
+    bionic:
+      pip:
+        packages: [sexpdata]
+    trusty:
+      pip:
+        packages: [sexpdata]
```
clalancette, line 5845: Trusty is out of support, so we don't need to add support for it.  Just remove this.

clalancette (commented):
A couple more things to fix, then I'll approve.

Affonso-Gui:
@clalancette Thanks for the comments! Just fixed the code, please let me know if I forgot something.

tfoote (approved):

clalancette (approved):

PR #24423, opened 2020-04-08
[dep] Add some exfat* tools.

130s:
arch
- expat-fuse equivalent not found.
- https://www.archlinux.org/packages/community/x86_64/exfat-utils

Debian jessie onward
- https://packages.debian.org/jessie/exfat-fuse
- https://packages.debian.org/jessie/exfat-utils

gentoo
- https://packages.gentoo.org/packages/sys-fs/fuse-exfat
- https://packages.gentoo.org/packages/sys-fs/exfat-utils

ubuntu
- https://packages.ubuntu.com/xenial/exfat-fuse
- https://packages.ubuntu.com/xenial/exfat-utils

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -739,6 +739,15 @@ espeak:
   fedora: [espeak]
   gentoo: [app-accessibility/espeak]
   ubuntu: [espeak]
+exfat-fuse:
+  debian: [exfat-fuse]
+  gentoo: [sys-fs/fuse-exfat]
+  ubuntu: [exfat-fuse]
+exfat-utils:
+  arch: [exfat-utils]
+  debian: [exfat-utils]
+  gentoo: [sys-fs/exfat-utils]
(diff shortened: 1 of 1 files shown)


clalancette (approved):

rosdep/base.yaml
```diff
@@ -727,6 +727,15 @@ espeak:
   fedora: [espeak]
   gentoo: [app-accessibility/espeak]
   ubuntu: [espeak]
+exfat-fuse:
+  debian: [exfat-fuse]
+  gentoo: [fuse-exfat]
```
tfoote, line 732: @allenh1 should the gentoo be `sys-fs/` prefixed here and below?

allenh1: @tfoote yes, that would be my preference.

130s: Got it, addressed.

PR #24426, opened 2020-04-09
kinetic: Update eigenpy to 2.3.1_2

wxmerkt:
This updates EigenPy on Kinetic to match the 2.3.x release on Melodic. It will fix the current source build failures for Pinocchio on Kinetic.

The release contains the same patches as on Melodic - plus a patch that used to be on the 1.6.9 release to fix lib installation prefix.


wxmerkt:
I noticed this is causing similar issues to the Melodic release (despite the same patches), and `moveit_ros_planning_interface` is failing. I'll get that fixed (fyi: @rhaschke)

PR #24432, opened 2020-04-10
add nlohmann-json to rosdistro

hakuturu583:
I want to add nlohmann-json to rosdistro.

Fedra package is here : https://rpms.remirepo.net/rpmphp/zoom.php?rpm=json
Gentoo package is here : https://packages.gentoo.org/packages/dev-cpp/nlohmann_json
Ubuntu package is here : https://packages.ubuntu.com/search?keywords=nlohmann-json-dev
Debian package is here : https://packages.debian.org/stretch/nlohmann-json-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4621,7 +4621,7 @@ nlohmann-json-dev:
   debian: [nlohmann-json-dev]
   gentoo: [dev-cpp/nlohmann_json]
   ubuntu:
-    '*': [nlohmann-json-dev ]
+    '*': [nlohmann-json-dev]
     xenial: null
 nmap:
   archlinux: [nmap]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4617,6 +4617,11 @@ nkf:
   fedora: [nkf]
   gentoo: [app-i18n/nkf]
   ubuntu: [nkf]
+nlohmann-json:
+  debian: [nlohmann-json-dev]
+  fedra: [json]
+  gentoo: [nlohmann_json]
+  ubuntu: [nlohmann-json-dev ]
 nmap:
   archlinux: [nmap]
   debian: [nmap]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4621,7 +4621,9 @@ nlohmann-json:
   debian: [nlohmann-json-dev]
   fedra: [json]
   gentoo: [nlohmann_json]
-  ubuntu: [nlohmann-json-dev ]
+  ubuntu: 
+    '*': [nlohmann-json-dev ]
+    xenial: null
 nmap:
   archlinux: [nmap]
   debian: [nmap]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4617,7 +4617,7 @@ nkf:
   fedora: [nkf]
   gentoo: [app-i18n/nkf]
   ubuntu: [nkf]
-nlohmann-json:
+nlohmann-json-dev:
   debian: [nlohmann-json-dev]
   gentoo: [dev-cpp/nlohmann_json]
   ubuntu: 
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4619,7 +4619,7 @@ nkf:
   ubuntu: [nkf]
 nlohmann-json:
   debian: [nlohmann-json-dev]
-  gentoo: [nlohmann_json]
+  gentoo: [dev-cpp/nlohmann_json]
   ubuntu: 
     '*': [nlohmann-json-dev ]
     xenial: null
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4619,7 +4619,6 @@ nkf:
   ubuntu: [nkf]
 nlohmann-json:
   debian: [nlohmann-json-dev]
-  fedra: [json]
   gentoo: [nlohmann_json]
   ubuntu: 
     '*': [nlohmann-json-dev ]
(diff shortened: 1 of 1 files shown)


tfoote (approved):

rosdep/base.yaml
```diff
@@ -4617,6 +4617,11 @@ nkf:
   fedora: [nkf]
   gentoo: [app-i18n/nkf]
   ubuntu: [nkf]
+nlohmann-json:
+  debian: [nlohmann-json-dev]
+  fedra: [json]
+  gentoo: [nlohmann_json]
+  ubuntu: [nlohmann-json-dev ]
```
clalancette, line 4624: This looks to only be available in Bionic and later.  Since we still support Xenial, you'll need to break this rule out into a "null" for xenial, and the package name here for the wildcard.  See elsewhere in this file on how to do that.  Also, there is a trailing whitespace in the brackets (that I'm surprised our CI didn't catch).

rosdep/base.yaml
```diff
@@ -4617,6 +4617,11 @@ nkf:
   fedora: [nkf]
   gentoo: [app-i18n/nkf]
   ubuntu: [nkf]
+nlohmann-json:
+  debian: [nlohmann-json-dev]
+  fedra: [json]
```
clalancette, line 4622: This should be `fedora`.  Also, while the link you pasted *does* show this as the nlohmann version of json, I actually don't think that it is the same one.  If you look at https://apps.fedoraproject.org/packages/json , it seems to be a json-rpc package from glib.  Since I don't see another equivalent package for Fedora, let's just remove it here.

rosdep/base.yaml
```diff
@@ -4617,6 +4617,11 @@ nkf:
...
+  gentoo: [nlohmann_json]
```
clalancette, line 4623: I don't think this is the right format for Gentoo, but I'm not entirely sure.  Most of the other Gentoo keys seem to be of the sort `dev-cpp/nlohmann_json`.

rosdep/base.yaml
```diff
@@ -4617,6 +4617,11 @@ nkf:
   fedora: [nkf]
   gentoo: [app-i18n/nkf]
   ubuntu: [nkf]
+nlohmann-json:
```
clalancette, line 4620: Since you are going for the development packages, this key should be `nlohmann-json-dev`.  That way, we can add (either in this PR, or a later one) a separate key of `nlohmann-json` which is just the runtime portions.

clalancette (changes requested):

rosdep/base.yaml
tfoote: @clalancette historically we've let this slide. It's known not to break anyone who's using the older platform because they are guaranteed not to be using the rule currently and saves us complexity in our management. That was particularly important before we had the wildcard support. Now we do have the ability to do this without incurring the cost of repeatedly adding new rules so I guess it's worth doing now.

rosdep/base.yaml
tfoote: good catch

rosdep/base.yaml
tfoote: @allenh1

rosdep/base.yaml
tfoote: +1

rosdep/base.yaml
allenh1: > Most of the other Gentoo keys seem to be of the sort `dev-cpp/nlohmann_json`

@clalancette This is true. The reason behind this is that some packages have different categories but the same package name. For example, the ROS overlay uses `ros-[distro]/[package]` for the version of the package per ROS distribution (just to keep things organized).

*TL;DR*: this _would_ work, but I strongly prefer it be `dev-cpp/nlohmann_json` instead of just the bare `nlohmann_json`.

@hakuturu583 thanks for including the Gentoo keys in this patch! I very much appreciate it.

rosdep/base.yaml
hakuturu583: I fixed this issue in https://github.com/ros/rosdistro/pull/24432/commits/000e1763d563544e2eec3f54a6bba064e2aac610

rosdep/base.yaml
hakuturu583: Thanks! I fixed in https://github.com/ros/rosdistro/pull/24432/commits/600a51b70b054f97ce633e5a1839a427e504a094

rosdep/base.yaml
hakuturu583: Thanks! I fixed this issue in https://github.com/ros/rosdistro/pull/24432/commits/925e15528003f84a2c05d9cb8cebf59c1a5be4df

rosdep/base.yaml
hakuturu583: Thanks! I fixed in https://github.com/ros/rosdistro/pull/24432/commits/8178b1406ac6142cf17d622437727cdb1a75fe6f

hakuturu583:
Hi, I found my PR was failed in running Nosetest.
And I found the error is below.

```

hakuturu583:
What is the error and what is the nosetests

clalancette:
> Hi, I found my PR was failed in running Nosetest.

So nosetest is a standard way of running python tests: https://nose.readthedocs.io/en/latest/ .  It's what we use to run our unittests.

Usually there is a line in the test that points out exactly which line failed, but for some reason it is not showing it here.  Running it locally, I see that the problem is that there is a trailing whitespace after the `ubuntu: ` line.  There is also still trailing whitespace after nlohmann-json-dev .  Both of those should be fixed, then I think CI will be happier and I'll be happy to approve.

hakuturu583:
@clalancette Thanks! I removed whitespace.

clalancette (approved):

PR #24435, opened 2020-04-10
Update python-hypothesis rule for Ubuntu.

airballking:
The rule python-hypothesis on Ubuntu seems to have not been touched in a while. This refactors the existing entries into a generic rule such that python-hypothesis resolves for newer Ubuntu versions, too.


clalancette (approved):

tfoote (approved):
thanks for the cleanup

PR #24447, opened 2020-04-11
add libedit support

ahoarau:


clalancette:
Following https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions , can you please add the packages location for each of these packages?  Also, it looks like there is a failure in CI somewhere that needs tracking down.

ahoarau:
Ubuntu : https://packages.ubuntu.com/search?keywords=libedit-dev
Debian: https://packages.debian.org/search?keywords=libedit-dev
arch: https://www.archlinux.org/packages/core/x86_64/libedit/
Fedora: https://fedora.pkgs.org/31/fedora-x86_64/libedit-devel-3.1-28.20190324cvs.fc31.i686.rpm.html
gentoo: https://packages.gentoo.org/packages/dev-libs/libedit
macports: https://www.macports.org/ports.php?by=variant&substr=libedit
rhel: https://pkgs.org/download/libedit-devel(x86-64)

Not really sure what the CI issue is. Any hints ?

ahoarau:
Everything is alright now. Ready to be merged

rosdep/base.yaml
```diff
@@ -1977,6 +1977,22 @@ libdxflib-dev:
   debian: [libdxflib-dev]
   fedora: [libdxflib-devel]
   ubuntu: [libdxflib-dev]
+libedit:
+  arch: [libedit]
+  debian: [libedit-dev]
```
clalancette, line 1982: For debian, ubuntu, and rhel, the `libedit` key should be the name of the library, not the devel version of the library.  That is, for debian and ubuntu it looks like it should be `libedit2`, and for rhel it should just be `libedit`.

clalancette (changes requested):

rosdep/base.yaml
ahoarau: fixed !

ahoarau: But it show that there is an issue with libreadline (ubuntu and debian says libreadline-dev for the non-dev version, it should be libreadline5)

clalancette: > fixed !

Great, thanks.

> But it show that there is an issue with libreadline (ubuntu and debian says libreadline-dev for the non-dev version, it should be libreadline5)

Excellent point, that seems like a bug.

The problem with that now is that there are people depending on that key, and getting more than they asked for.  If we fix it, then some stuff is probably going to start failing.  @tfoote What's our backwards-compatibility policy here?

clalancette (approved):

ahoarau:
Everything is clear and approved !

clalancette:
We're just waiting on one more approval, then this can be merged.

tfoote (approved):

PR #24455, opened 2020-04-13
Update cyclone branches

rotu:
All the CycloneDDS stuff was out of date. This makes it current.

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -589,7 +589,7 @@ repositories:
       test_pull_requests: false
       type: git
       url: https://github.com/eclipse-cyclonedds/cyclonedds.git
-      version: 76fa68808682a15dd8aff26da20622ac574fa1a1
+      version: releases/0.5.x
     status: developed
   demos:
     doc:
@@ -2205,7 +2205,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/ros2/rmw_cyclonedds.git
-      version: master
+      version: dashing-eloquent
     status: developed
   rmw_fastrtps:
     doc:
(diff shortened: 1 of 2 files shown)


rotu:
FYI @ThijsSassen, @eboasson, @joespeed. Looks like this was a significant missed step in multiple releases of Cyclone.

rotu:
As per https://github.com/ros2/ros2/pull/891#issuecomment-612444236
@nuclearsandwich, why is this separate from https://github.com/ros2/ros2/blob/master/ros2.repos? Is there a reason to have their source branches different between this repo and ros2, or is the duplication just a historical accident?

nuclearsandwich:
> why is this separate from [ros2/ros2:ros2.repos@`master`](https://github.com/ros2/ros2/blob/master/ros2.repos) [](/ros2/ros2/blob/HEAD@{2020-04-13T01:04:07Z}/ros2.repos)

The duplication is due to the different tools used to produce ROS 2 releases on different platforms. I can't think of a reason right now that they'd disagree intentionally.

During the early phases of ROS 2 development its releases were not part of any rosdistro and the repos file which was used for development also became the source of truth for releases. For current ROS 2 releases it's still the primary input for ci.ros2.org and the repos file in ros2/ros2 is still the thing we tag and upload binaries to. In general the version for the source entry in the rosdistro distribution.yaml should match the version in the development repos file and the version in the release repos file should match the release tag in the release repo.

nuclearsandwich (approved):

rotu:
@mjcarroll, if you intend to review, please do so expeditiously

mjcarroll (approved):

PR #24468, opened 2020-04-14
Adding omnibase to documentation index for distro

hardesh:


hardesh:
I would like omnibase to be indexed and documented on ros.org.

melodic/distribution.yaml
```diff
@@ -6269,6 +6269,16 @@ repositories:
...
       version: indigo-devel
     status: unmaintained
+  omnibase:
+    doc:
+      type: git
+      url: https://github.com/ERC-BPGC/omnibase.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/ERC-BPGC/omnibase.git
```
clalancette, line 6279: The package.xml for all of the packages in that source repository say that the license is "ERC", but there is no text of exactly what that license is.

If this is going to be open source, we highly recommend using one of the standard licenses (BSD or Apache 2).  We are not lawyers, so we aren't qualified to review new licenses and determine whether they give us the ability to redistribute the software.

hardesh: I will change the license to BSD then, the packages are going to be open source. This is my first time in this matter. 
Thanks for the suggestions!

hardesh: I have changed the licenses for all the packages.

clalancette, line 6279: One last thing, then I think we're good; can you put a LICENSE file with the BSD license in the source repository?  I'm guessing you want to use https://opensource.org/licenses/BSD-3-Clause .  Thanks.

clalancette (changes requested):

melodic/distribution.yaml
hardesh: I've added the BSD-3 license file in the source repository.
Thanks.

clalancette:
Thanks for iterating.

PR #24474, opened 2020-04-15
Add python-serial entry for focal

garyservin:
Needed for rosserial release in Noetic.


rosdep/python.yaml
```diff
@@ -4101,6 +4101,7 @@ python-serial:
     artful_python3: [python3-serial]
     bionic: [python-serial]
     bionic_python3: [python3-serial]
+    focal: [python3-serial]
```
tfoote, line 4104: We should not be using python3-serial for a python-serial key. It should be in a new python3-serial key.

Also instead of adding just focal we should use `'*': [python-serial]` and just leave the legacy `_python3` version rules.

tfoote (changes requested):
Thanks for the update but I have two requests for this change.

rosdep/python.yaml
garyservin: Thanks for the comment, I didn't check that there was already a key for python3-serial, my bad.

I've added the `focal_python3` entry instead and created `'*': [python-serial]` as requested

rosdep/python.yaml
```diff
@@ -4097,31 +4097,16 @@ python-serial:
   gentoo: [dev-python/pyserial]
   openembedded: [python-pyserial@meta-python]
   ubuntu:
-    artful: [python-serial]
+    '*': [python-serial]
```
clalancette, line 4100: This is good.

rosdep/python.yaml
```diff
@@ -4097,31 +4097,16 @@ python-serial:
...
-    lucid: [python-serial]
-    maverick: [python-serial]
-    natty: [python-serial]
-    oneiric: [python-serial]
-    precise: [python-serial]
-    quantal: [python-serial]
-    raring: [python-serial]
-    saucy: [python-serial]
-    trusty: [python-serial]
+    focal_python3: [python3-serial]
```
clalancette, line 4103: But we should not add this.  The `_python3` sub-keys were a failed experiment, and they should not be used or added to anymore.  Instead, you should use the `python3-serial` key, which already exists.

clalancette (changes requested):

clalancette:
No response for a month, so closing.  Do feel free to fix up the issues pointed out in review and reopen.  Thanks for the contribution.

PR #24486, opened 2020-04-16
Adding gios to ros melodic index

manucalop:
Pull request to add gios package to melodic index

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3160,6 +3160,17 @@ repositories:
       url: https://github.com/ros/geometry_tutorials.git
       version: indigo-devel
     status: maintained
+  gios:
+    doc:
+      type: git
+      url: https://github.com/manucalop/gios.git
+      version: master
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/manucalop/gios.git
+      version: master
+    status: maintained
   gl_dependency:
     doc:
       type: git


tfoote (changes requested):
Thanks for submitting this. The first think that I looked at is your package.xml and it has a important data missing with the license and maintainer listed as todo.

Also can you confirm that you've given the buildfarm the necessary access to enable pull request tests? http://wiki.ros.org/buildfarm/Pull%20request%20testing

manucalop:
> Thanks for submitting this. The first think that I looked at is your package.xml and it has a important data missing with the license and maintainer listed as todo.
> 
> Also can you confirm that you've given the buildfarm the necessary access to enable pull request tests? http://wiki.ros.org/buildfarm/Pull%20request%20testing

Done. I solved the issues with the package.xml and granted access to the buildfarm

clalancette (commented):
The reason I've been hesitant to merge this is that it copies a whole bunch of code from other projects like [Acado](https://acado.github.io/) and [qpoases](https://github.com/coin-or/qpOASES).  There are two problems with that:

1.  I don't know the licenses of those projects, and I don't really have time to go tracking them down.
1.  Embedding them here might cause problems if we ever want to release those packages separate from this.

@manucalop Can you lay out exactly which projects you've copied code from, and what their licenses are?  Also, are any of the header files/methods from those projects installed as part of the package and exported for downstream consumers?

tfoote:
You should add dependencies on those packages instead of embedding them. 

Acado is already released and packaged in melodic: https://index.ros.org/p/acado/#melodic It also has LGPL as the license. 

And so is qpoases https://index.ros.org/p/qpoases_vendor/gitlab-autowarefoundation-autoware-ai-qpoases_vendor/#melodic under the Apache 2.0 license.

manucalop:
@clalancette @tfoote, great thoughts, thanks!. However, my package is independent from acado and qpoases. They're included just to run tests on acado-gios interface. If I include acado as a testing dependency that would solve the issue right?

clalancette:
> @clalancette @tfoote, great thoughts, thanks!. However, my package is independent from acado and qpoases. They're included just to run tests on acado-gios interface. If I include acado as a testing dependency that would solve the issue right?

If you did that and removed the copied code from your repository, that would be most ideal.  That way we keep things modular and able to be combined in different ways.

clalancette:
I'm going to close this PR out for now, assuming that there will be a follow-up PR doing what we discussed above.  If you think that is in error, please feel free to reopen.

PR #24493, opened 2020-04-17
noetic: Update octomap/octovis to 1.9.5-3

wxmerkt:
Fixes the build failures for projects dependent on octomap, e.g., octovis by using a patch to set the `CMAKE_INSTALL_LIBDIR`.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -808,7 +808,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/ros-gbp/octomap-release.git
-      version: 1.9.5-2
+      version: 1.9.5-3
     source:
       type: git
       url: https://github.com/OctoMap/octomap.git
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -808,7 +808,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/ros-gbp/octomap-release.git
-      version: 1.9.5-2
+      version: 1.9.5-3
```
sloretz, line 811: I only see [`1.9.5-2` in the release repo](https://github.com/ros-gbp/octomap-release#octomap-noetic---195-2). Did an error occur during the release?

wxmerkt: I used `git-bloom-release` instead of `bloom-release`. The release tags are there: https://github.com/ros-gbp/octomap-release/releases. I reran it and will update the release number. I believe `git-bloom-release` does not update the README, only the release number in `tracks.yaml`.

PR #24501, opened 2020-04-19
rqt_robot_monitor: 0.5.10-1 in 'noetic/distribution.yaml' [bloom]

arne48:
Increasing version of package(s) in repository `rqt_robot_monitor` to `0.5.10-1`:

- upstream repository: https://github.com/ros-visualization/rqt_robot_monitor.git
- release repository: https://github.com/ros-gbp/rqt_robot_monitor-release.git
- bloom version: `0.9.3`

## rqt_robot_monitor

```
* updated package to format 3
* added mutex and checks to fix (#13 <https://github.com/ros-visualization/rqt_robot_monitor/issues/13>)
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1565,6 +1565,21 @@ repositories:
       url: https://github.com/ros-visualization/rqt_py_console.git
       version: master
     status: maintained
+  rqt_robot_monitor:
+    doc:
+      type: git
+      url: https://github.com/ros-visualization/rqt_robot_monitor.git
+      version: master
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/ros-gbp/rqt_robot_monitor-release.git
+      version: 0.5.10-1
+    source:
+      type: git
+      url: https://github.com/ros-visualization/rqt_robot_monitor.git
+      version: master
+    status: maintained
   rqt_robot_steering:
     doc:
       type: git


ahcorde (approved):

sloretz:
@arne48 it looks like this package depends on a Python 2 dependency. ROS Noetic is targeting Python 3.

https://github.com/ros-visualization/rqt_robot_monitor/blob/407c9467f5deb2f351640e7ffbde0987a925ca1f/package.xml#L45

```xml
  <exec_depend>python-rospkg</exec_depend>
```

Does the package really use the CLI tool `rospkg`, or does it only `import rospkg` python modules? If it's the latter than it should use `python-rospkg-modules` for Python 2 and `python3-rospkg-modules` for Python 3.

arne48:
@sloretz this is a good point the condition attribute should be used to specify python2 and python3 dependencies.
About the import of rospkg or rospkg-modules:
Several rqt plugins use this dependency to locate the .ui file within their package folders.
Others also don't specify this dependency at all.
So I considered this the best practice.

I now would create a patched version of this package (using conditions) and open another PR, but stay with importing rospkg.
Would that be OK with you?

sloretz:
> I now would create a patched version of this package (using conditions) and open another PR, but would like to stay with importing rospkg.

I'm not suggesting the import be changed, just the `package.xml`. That said, for Noetic I'm happy to merge with just the Python 2/3 conditional dependencies.

If the package only does `import rospkg` in a python module ([which it looks like that's all this package does](https://github.com/ros-visualization/rqt_robot_monitor/blob/10c5196ff625c3d90b8308369fade50505f13796/src/rqt_robot_monitor/robot_monitor.py#L36)) then it depends on either `python-rospkg-modules` or `python3-rospkg-modules`. The `package.xml` incorrectly lists `python-rospkg`. This will work for some cases because `python-rospkg` depends on `python-rospkg-modules`, but `python-rospkg` installs a system-wide command line tool called `rospkg`. Try `which rospkg` in your shell. This is a problem for packages released to both ROS 1 and ROS 2 on Ubuntu Bionic because `python-rospkg` and `python3-rospkg` conflict: they both provide a system wide command line tool called `rospkg` on `$PATH`, and so they can't be installed at the same time. `python-rospkg-modules` and `python3-rospkg-modules` just install python modules so they can be installed at the same time. This doesn't matter for ROS Noetic because both ROS Noetic and ROS Foxy will use Python 3 on Ubuntu Focal.

arne48:
@sloretz thank you very much for this clarifying this.
I guess some packages just expect some rospkg being installed and not adding it as a dependency, because python-rospkg and as its dependency python-rospkg-modules would be installed even in base installations of ROS as dependencies of ros-melodic-ros-base.
So your opinion about choosing python(3)-rospkg-modules over rospkg as a dependency makes sense.
I will change the package accordingly and open another PR.
Thank you very much again for discussing this topic.

PR #24522, opened 2020-04-21
ros2_intel_realsense: 2.0.7-1 in 'dashing/distribution.yaml' [bloom]

sharronliu:
Increasing version of package(s) in repository `ros2_intel_realsense` to `2.0.7-1`:

- upstream repository: https://github.com/intel/ros2_intel_realsense.git
- release repository: https://github.com/ros2-gbp/ros2_intel_realsense-release.git
- bloom version: `0.9.3`
- previous version for package: `2.0.4-2`

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -2364,12 +2364,14 @@ repositories:
       version: master
     release:
       packages:
-      - realsense_camera_msgs
-      - realsense_ros2_camera
+      - realsense_examples
+      - realsense_msgs
+      - realsense_node
+      - realsense_ros
       tags:
         release: release/dashing/{package}/{version}
       url: https://github.com/ros2-gbp/ros2_intel_realsense-release.git
-      version: 2.0.4-2
+      version: 2.0.7-1
     source:
       type: git
       url: https://github.com/intel/ros2_intel_realsense.git


nuclearsandwich:
Hi @sharronliu I'm concerned that this PR seems to contain a completely different set of packages from those released into Dashing previously. We have the intention that released rosdistros are stable and the changes here are definitely breaking as whole packages have been removed and replaced. Can you elaborate on the changes made and what the migration path looks like?

sharronliu:
> 
> 
> Hi @sharronliu I'm concerned that this PR seems to contain a completely different set of packages from those released into Dashing previously. We have the intention that released rosdistros are stable and the changes here are definitely breaking as whole packages have been removed and replaced. Can you elaborate on the changes made and what the migration path looks like?

@nuclearsandwich Your observation is correct. The branch has been switched to `refactor` since after the Dashing release.
The reasons we switching to `refactor` branch are:
- Keep upgrading to the latest version of librealsense2 (the native driver).
- Keep porting from [ROS Realsense](https://github.com/IntelRealSense/realsense-ros) to support more Realsense models, like `d435i`, `t265`, as well as the Lidar model coming soon. You may noticed the `master` branch is synced with 2.0.1 (of [ROS version](https://github.com/IntelRealSense/realsense-ros/tags)) which is more than two years ago.
- Keep improving performance, e.g., FPS and latency of Point Cloud2 frame, by leveraging ROS2 advantages (like, rclcpp_components etc.) to support intro-process comms and dynamic parameters configure.

Briefly the development is switched to the `refactor` branch. You've raised a good question, shall we provide the above advantages to the user of ROS2 Dashing LTS, or keep the existing and stable features? To my understanding, Dashing LTS will be `Supported until May 2021`, and shall we keep stable till then?

sharronliu:
@nuclearsandwich When looking at the question on the other side, if switching to the new branch, people who are using Dashing may feel the way of launching changed, e.g.
- old: `ros2 run realsense_ros2_camera realsense_ros2_camera`
- new: `ros2 run realsense_node realsense_node`

Maybe we can initialize a discourse discussion and get to know if the Dashing users would like to take such a slight change, or giving up to get more models supported :)

nuclearsandwich:
Released ROS distributions are intended to be as stable and reliable as we can make them. Updates to packages in these repositories should be limited to compatible bug fixes and features.

You are welcome to start a discussion on discourse regarding this question. I'm going to close this pull request for now.

PR #24525, opened 2020-04-21
Added clang-tidy on Ubuntu Bionic

pablogs9:
Updated Clang-tidy on Ubuntu bionic

https://packages.ubuntu.com/bionic/devel/clang-tidy

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -404,6 +404,7 @@ clang-format:
   ubuntu: [clang-format]
 clang-tidy:
   debian:
+    bionic: [clang-tidy]
     buster: [clang-tidy]
     stretch: [clang-tidy]
   fedora: [clang-tools-extra]


rosdep/base.yaml
```diff
@@ -404,6 +404,7 @@ clang-format:
   ubuntu: [clang-format]
 clang-tidy:
   debian:
+    bionic: [clang-tidy]
```
clalancette, line 407: Bionic is an Ubuntu distribution, not a Debian one.  Also, bionic is already covered by the wildcard on the ubuntu keys.  So I'm going to close this out.

pablogs9: In Ubuntu 18.04.4 for Raspberry Pi 4, it says that there is no package candidate for clang-tidy.

nuclearsandwich: [clang-tidy](https://packages.ubuntu.com/bionic/clang-tidy) is listed as being present for arm64 and armhf. It sounds like your Pi may not be configured with the upstream Ubuntu repositories.

PR #24526, opened 2020-04-21
[rosdep] add python-rpi.gpio-pip key

mikaelarguedas:
deb package python-rpi.gpio is only avialable on arm archs
the pip package RPi.gpio is available on all platforms

We could also change the existing `python-rpi.gpio` key to only map to pip packages (https://github.com/ros/rosdistro/pull/16545#issuecomment-616664611) but the name would be misleading so it looks like adding a `-pip` suffixed key still makes sense

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3964,6 +3964,16 @@ python-rpi.gpio:
       pip:
         packages: [RPi.GPIO]
     zesty: [python-rpi.gpio]
+python-rpi.gpio-pip:
+  debian:
+    pip:
+      packages: [RPi.GPIO]
+  fedora:
+    pip:
+      packages: [RPi.GPIO]
+  ubuntu:
+    pip:
+      packages: [RPi.GPIO]
 python-rrdtool:
   debian: [python-rrdtool]
   fedora: [rrdtool-python]


tfoote (approved):
That's an awkward level of architecture support. But I guess it makes sense as they're no non-arm raspberry pi's. Being explicit about the pip dependency makes sense. Though are you actually using this on non-arm targets? If not I'd probably say just stick with the arm specific key.

mikaelarguedas:
I faced it when auditing missing rosdep keys on a past project and I was surprised that rosdep failed in an unusual way when I tried to install all the project deps on an amd platform.

One would have expected rosdep to either say "I don't know this key" or "this package is not available on this platform", but seeing it attempt and the apt invocation fail was unexpected.

Granted on amd I should be skipping that key :+1:, but it still seems like a useful thing to have a key/rules that work on all arches.

> If not I'd probably say just stick with the arm specific key.

That makes sense. The only thing holding me off is (laziness) that pip keys just work according to the `ROS_PYTHON_VERSION` used while to use the apt ones I need to migrate all my package.xml to format 3 and add conditions for each python key which is pretty cumbersome.

tfoote:
Yeah, the one thing that will prevent you from doing is releasing the package into debian packages, but in this case it seems that supporting either could be valuable. Though having the duplicate rule can/will cause issues for some people with accidental masking. But we don't really have the ability to do different arch resolution.

clalancette (approved):

PR #24552, opened 2020-04-24
add python3-gi key

knorth55:
add python3-gi key

arch: https://www.archlinux.org/packages/extra/x86_64/python-gobject/
debian: https://pkgs.org/download/python3-gi
fedora: https://apps.fedoraproject.org/packages/pygobject3
gentoo: https://packages.gentoo.org/packages/dev-python/pygobject
ubuntu: https://pkgs.org/download/python3-gi


sloretz (approved):
Used by ros-drivers/audio_common#133

clalancette (approved):

PR #24623, opened 2020-04-28
Splitting scipy into python- and python3-.

JWhitleyWork:
Splits the `python-scipy` entry into `python-` and `python3-` variants.

- Arch: https://www.archlinux.org/packages/community/x86_64/python-scipy/
- Debian:
  - Stretch: https://packages.debian.org/stretch/python3-scipy
  - Buster: https://packages.debian.org/buster/python3-scipy
- Fedora: https://apps.fedoraproject.org/packages/scipy#
- FreeBSD: https://www.freebsd.org/cgi/ports.cgi?query=scipy&stype=all
- Gentoo: https://packages.gentoo.org/packages/sci-libs/scipy
- MacPorts: https://www.macports.org/ports.php?by=name&substr=scipy
- OpenSUSE: https://software.opensuse.org/package/python3-scipy
- Ubuntu:
  - Bionic: https://packages.ubuntu.com/bionic/python3-scipy
  - Focal: https://packages.ubuntu.com/focal/python3-scipy


tfoote (changes requested):
Thanks for the update and adding the new target. Our policy for the deprecated format of the <distro>_python3 rules has been to keep them for backwards compatibility and then we can let them age out. 

This can be done relatively compactly by using a wildcard for the standard name and then just leaving the specific versions with the python3 suffix enumerated. 

```
'*': [python-scipy]
<distroA>_python3: [python3-scipy]
...
```

It's very conservative but it's easy enough to keep the legacy rules in case someone is using it.

JWhitleyWork:
@tfoote Thanks for the review! Please let me know if the changes look the way you described.

rosdep/python.yaml
```diff
@@ -5867,6 +5851,19 @@ python3-ruamel.yaml:
 python3-schedule:
   debian: [python3-schedule]
   ubuntu: [python3-schedule]
+python3-scipy:
+  arch: [python-scipy]
+  debian: [python3-scipy]
+  fedora: [scipy]
```
clalancette, line 5857: This key isn't correct on Fedora; that will actually get you the python2 version of it.  This key should be `python3-scipy`.  While we are at it, we may as well change the `python-scipy` key for Fedora to be `python2-scipy`, as it is more descriptive.

clalancette (changes requested):

rosdep/python.yaml
JWhitleyWork: Thanks for the clarification, @clalancette. I believe I saw this when looking for the packages but I was worried about changing the `python-scipy` entry and disrupting older versions of Fedora where python == python2. I've implemented these changes, though.

tfoote (approved):
Thanks for the iteration. It looks good to me.

clalancette (approved):

PR #24627, opened 2020-04-28
[rosdep] Add python-influxdb

garaemon:
* debian: https://packages.debian.org/search?keywords=python-influxdb
* ubuntu: https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python-influxdb&searchon=names

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2091,8 +2091,13 @@ python-impacket:
   fedora: [python-impacket]
   ubuntu: [python-impacket]
 python-influxdb:
-  debian: [python-influxdb]
-  ubuntu: [python-influxdb]
+  debian:
+    buster: [python-influxdb]
+    jessie: [python-influxdb]
+  ubuntu:
+    bionic: [python-influxdb]
+    disco: [python-influxdb]
+    xenial: [python-influxdb]
 python-inject-pip:
   ubuntu:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2090,6 +2090,9 @@ python-impacket:
   debian: [python-impacket]
   fedora: [python-impacket]
   ubuntu: [python-impacket]
+python-influxdb:
+  debian: [python-influxdb]
+  ubuntu: [python-influxdb]
 python-inject-pip:
   ubuntu:
     pip:


tfoote (approved):
lgtm, it looks like there's not a fedora package upstream: https://computingforgeeks.com/how-to-install-influxdb-on-fedora/

rosdep/python.yaml
```diff
@@ -2090,6 +2090,9 @@ python-impacket:
   debian: [python-impacket]
   fedora: [python-impacket]
   ubuntu: [python-impacket]
+python-influxdb:
+  debian: [python-influxdb]
```
clalancette, line 2094: Oddly, this is available in jessie and buster, but not in the release between them (stretch).  Since we still support stretch, I think we need to call it out separately with a 'null' entry.

rosdep/python.yaml
```diff
@@ -2090,6 +2090,9 @@ python-impacket:
...
+  ubuntu: [python-influxdb]
```
clalancette, line 2095: This is only available on Ubuntu up until [disco](https://packages.ubuntu.com/search?keywords=python-influxdb&searchon=names&suite=all&section=all); for eoan and focal, I assume it got purged as the general purging of Python 2 packages happened.  So I think we should make specific rule for it for the currently supported releases (xenial and bionic)

clalancette (changes requested):
It looks like we need changes since the packaging situation is a bit all over the place in the Debian/Ubuntu ecosystem.

clalancette:
> lgtm, it looks like there's not a fedora package upstream: https://computingforgeeks.com/how-to-install-influxdb-on-fedora/

For what it is worth, Fedora 31 does have `python3-influxdb`, but they aren't allowing any new Python 2 packages.  For the `python-influxdb` (implicit Python 2) this is correct, but if we add a python3 variant of the key we can add Fedora.

rosdep/python.yaml
tfoote: It's not missing stretch has been delisted from the website I guess they're pulling down stretch as it's been replaced by buster . The package is actually available in stretch.


<details>

```
tfoote@snowman4:~ Last: [0] (0s Seconds)
$ rocker debian:stretch 
Plugins found: ['dev_helpers', 'env', 'git', 'home', 'nvidia', 'pulse', 'ssh', 'user', 'x11']
Active extensions []
Writing dockerfile to /tmp/tmpbbwhdvie/Dockerfile
vvvvvv

FROM debian:stretch

^^^^^^
Building docker file with arguments:  {'path': '/tmp/tmpbbwhdvie', 'rm': True, 'nocache': False}
building > Step 1/1 : FROM debian:stretch
building >  ---> bd753a4a335e
building > Successfully built bd753a4a335e
Executing command: 
docker run -it   --rm      bd753a4a335e 
root@44284de1d705:/# apt-get update
Ign:1 http://deb.debian.org/debian stretch InRelease
Get:2 http://security.debian.org/debian-security stretch/updates InRelease [94.3 kB]
Get:3 http://deb.debian.org/debian stretch-updates InRelease [91.0 kB]
Get:4 http://deb.debian.org/debian stretch Release [118 kB]         
Get:5 http://deb.debian.org/debian stretch Release.gpg [2410 B]
Get:6 http://security.debian.org/debian-security stretch/updates/main amd64 Packages [525 kB]
Get:7 http://deb.debian.org/debian stretch-updates/main amd64 Packages [27.9 kB]
Get:8 http://deb.debian.org/debian stretch/main amd64 Packages [7083 kB]       
Fetched 7942 kB in 1s (4846 kB/s)  
Reading package lists... Done
root@44284de1d705:/# apt-cache search influxdb
golang-github-influxdb-enterprise-client-dev - Golang client for speaking to the InfluxDB Enterprise application
golang-github-influxdb-usage-client-dev - library for speaking to the InfluxDB Anonymous Usage Reporting API
golang-github-influxdb-influxdb-dev - Scalable datastore for metrics, events, and real-time analytics. Dev package
influxdb - Scalable datastore for metrics, events, and real-time analytics
influxdb-client - command line interface for InfluxDB
influxdb-dev - Transitional package for golang-github-influxdb-influxdb-dev
ruby-influxdb - library for InfluxDB
```
</details>

tfoote (commented):
Yeah, the python3 rule is a whole different story with the new replacement. 

For the python2 things that are being slowly purged I'm not sure that we need to explicitly call them out like this. We have a lot of python2 rules that are untied to versions because they've been the same for many releases that are going to go away. The error that people will get is that the packages cannot install and it will be relatively straight forward to discover that the package no longer exists on their target platform. The marginal benefit of requiring that new rules for python2 content explicitly call out the availaibility of that package when the majority of the rules will become invalid unless we do a significant audit. It would be great to be able to check installability, but until we have automation along those lines the simpler generic rules for python2 packages stay seems to make sense. In the not too distant future they will all be removable as non of our platforms will support python2.

rosdep/python.yaml
garaemon:
```
root@44284de1d705:/# apt-cache search influxdb
golang-github-influxdb-enterprise-client-dev - Golang client for speaking to the InfluxDB Enterprise application
golang-github-influxdb-usage-client-dev - library for speaking to the InfluxDB Anonymous Usage Reporting API
golang-github-influxdb-influxdb-dev - Scalable datastore for metrics, events, and real-time analytics. Dev package
influxdb - Scalable datastore for metrics, events, and real-time analytics
influxdb-client - command line interface for InfluxDB
influxdb-dev - Transitional package for golang-github-influxdb-influxdb-dev
ruby-influxdb - library for InfluxDB
```

It looks that the output of `apt-cache search influxdb` does not include python-influxdb. I will remote the package for stretch distribution.

rosdep/python.yaml
garaemon: I've just updated to list xenial, bionic and disco as python-influxdb targets.

tfoote (approved):
Sorry yeah I was too quick it really is missing on stretch. I guess we might as well call them all out then.

clalancette:
> The marginal benefit of requiring that new rules for python2 content explicitly call out the availaibility of that package when the majority of the rules will become invalid unless we do a significant audit.

That's a fair point.  Sounds good to me then.

clalancette (approved):

PR #24675, opened 2020-04-30
Add migration tools.

nuclearsandwich:
Based on the script developed in http://github.com/nuclearsandwich/rosnroll this tool should be suitable for the following cases

* Creating a new ROS distribution from an existing one by re-triggering bloom releases on all repositories.
* Updating a previously migrated distribution by re-triggering bloom releases on missing repositories.
* Updating the release platform of an existing ROS distribution by re-triggering bloom releases for the new target platform.


tfoote:
Close and reopen to try to retrigger the actions

tfoote (approved):

PR #24741, opened 2020-05-06
rosdep: added python3-ruamel.yaml

atiderko:
Added rosdep for Ubuntu focal:
https://packages.ubuntu.com/focal/python3-ruamel.yaml

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5938,6 +5938,7 @@ python3-ruamel.yaml:
   ubuntu:
     artful: [python3-ruamel.yaml]
     bionic: [python3-ruamel.yaml]
+    focal: [python3-ruamel.yaml]
     xenial: [python3-ruamel.yaml]
     yakkety: [python3-ruamel.yaml]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5938,6 +5938,7 @@ python3-ruamel.yaml:
   ubuntu:
     artful: [python3-ruamel.yaml]
     bionic: [python3-ruamel.yaml]
+    focal: [python3-ruamel.yaml]
     xenial: [python3-ruamel.yaml]
     yakkety: [python3-ruamel.yaml]
     zesty: [python3-ruamel.yaml]
```
clalancette, line 5944: Given that this is supported on all of our supported platforms with the same name, I'll suggest changing this to:

```
ubuntu: [python3-ruamel.yaml]
```

clalancette (changes requested):

atiderko:
Applied suggestion from Chris

clalancette (approved):

PR #24773, opened 2020-05-08
Fix libvlc rosdep key in Ubuntu/Debian

pjreed:
In Ubuntu Zesty and later and Debian Stretch and later, the `vlc-nox`
package has been deprecated and turned into a transitional dummy package
that cannot be installed.  It has been replaced by the `vlc-bin` package.
For reference:
- https://packages.ubuntu.com/zesty/allpackages
- https://packages.debian.org/stretch/vlc-nox

This updates the `libvlc` rosdep key to install the correct packages
in the appropriate versions of Ubuntu and Debian.

Signed-off-by: P. J. Reed <preed@swri.org>

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4001,7 +4001,6 @@ libv4l-dev:
 libvlc:
   debian:
     '*': [libvlc5, vlc-bin]
-    buster: [libvlc5, vlc-nox]
     jessie: [libvlc5, vlc-nox]
   fedora: [vlc-core]
   gentoo: [media-video/vlc]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3999,10 +3999,22 @@ libv4l-dev:
       packages: [v4l-utils]
   ubuntu: [libv4l-dev]
 libvlc:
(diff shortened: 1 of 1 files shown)


tfoote (approved):

rosdep/base.yaml
```diff
@@ -3999,10 +3999,22 @@ libv4l-dev:
       packages: [v4l-utils]
   ubuntu: [libv4l-dev]
 libvlc:
-  debian: [libvlc5, vlc-nox]
+  debian:
+    '*': [libvlc5, vlc-bin]
+    buster: [libvlc5, vlc-nox]
```
clalancette, line 4004: Debian Buster came after Debian Stretch, so shouldn't this line be removed?

clalancette (changes requested):

rosdep/base.yaml
pjreed: You are right, I got my Debian version names mixed up.

pjreed, line 4004:
```suggestion
```

clalancette (approved):

PR #24916, opened 2020-05-14
Added libglew-dev for ubuntu:focal to the rosdep base rules

suddrey-qut:
This is needed to build tools such as the MoveIt Perception pipeline.

Signed-off-by: Gavin Suddrey <g.suddrey@qut.edu.au>


rosdep/base.yaml
```diff
@@ -2310,6 +2310,7 @@ libglew-dev:
   ubuntu:
     artful: [libglew-dev]
     bionic: [libglew-dev]
+    focal: [libglew-dev]
     oneiric: [libglew1.5-dev]
     precise: [libglew1.6-dev]
     quantal: [libglew-dev]
```
clalancette, line 2316: I know you are just adding a new key here, but I think we can simplify this given the current supported Ubuntu releases.  That is, I think we can remove this whole block and replace it with:

```
ubuntu: [libglew-dev]
```

Which should work on Xenial and newer, and also make us future-proof.

clalancette (changes requested):

suddrey-qut:
I thought the same thing, but did not want to assume on whether precise and earlier would no longer be supported.

clalancette (approved):

PR #24919, opened 2020-05-15
rqt_common_plugins indigo-EOL branch

sloretz:
https://github.com/ros-visualization/rqt_common_plugins/pull/459#issuecomment-604547280

https://github.com/ros-visualization/rqt_common_plugins/tree/indigo-EOL

AFAIK this avoids breaking users building indigo from source using the --upstream-development flag. There won't be any upstream development in an EOL distro, so I've used the branch name `indigo-EOL` instead of the usual `indigo-devel`.

PR #24921, opened 2020-05-15
rqt_robot_plugins indigo-EOL

sloretz:
https://github.com/ros-visualization/rqt_robot_plugins/pull/124#issuecomment-604547473

https://github.com/ros-visualization/rqt_robot_plugins/tree/indigo-EOL

Same as https://github.com/ros/rosdistro/pull/24919 for rqt_robot_plugins


tfoote (approved):
I'll merge this with an admin override since it's modifying old distros but if that branch has moved forward these should be changed for access.

PR #24685, opened 2020-05-01
add 8 new packages

incebellipipo:
add

-  libmongoc-1.0-0
-  libmongoc-dev
-  libcap2
-  libcap-ng0
-  libcap-ng-dev
-  libprocps6
-  libprocps-dev
-  libdpkg-dev

PR #24936, opened 2020-05-16
add blank line check

incebellipipo:
In PR https://github.com/ros/rosdistro/pull/24685 I made mistake and add a blank line at the bottom of the yaml file. @clalancette replied with https://github.com/ros/rosdistro/pull/24685#discussion_r425428017 and pointed out that blank line check is missing in linter script. So I add this functionality to `check_rosdep.py` script and update contributing guide accordingly.

Proposal: https://github.com/ros/rosdistro/pull/24685#issuecomment-629109065

> Adding this will do.
> ```python
> def no_empty_lines(buf):
>    clean = True 
>    for i, l in enumerate(buf.split('\n')[:-1]):
>        if re.match(r'^\s*$', l):
>            print_err("empty line %u" % (i+1))
>            clean = False
>    return clean
>```
>Results:
>```bash
>checking for trailing spaces...
>checking for empty lines...
>  ERR: empty line 6225
>checking for incorrect indentation...
>checking for non-bracket package lists...
>checking for item order...
>building yaml dict...
>there were errors, please correct the file
>```
>I also checked for a case where an empty line also has trailing whitespaces which is kinda rare. If that happens script returns this:
>```bash
>$ ./check_rosdep.py ../rosdep/base.yaml
>checking for trailing spaces...
>  ERR: trailing space line 6225
>checking for empty lines...
>  ERR: empty line 6225
>checking for incorrect indentation...
>  ERR: line 6224:   
>Traceback (most recent call last):
>  File "./check_rosdep.py", line 223, in <module>
>    if not main(args.infile):
>  File "./check_rosdep.py", line 181, in main
>    my_assert(correct_indent(buf))
>  File "./check_rosdep.py", line 116, in correct_indent
>    return generic_parser(buf, fun)
>  File "./check_rosdep.py", line 82, in generic_parser
>    s = re.search(r'(?!' + indent_atom + ')[^\s]', l).start()
>AttributeError: 'NoneType' object has no attribute 'start'
>$ echo $?
>1
>```
>I didn't want to commit my local change, since it will automatically be a part of this PR which labelled as rosdep. If its ok, I can commit it, if its not I'll wait for this PR to be merged and create a new PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,6 +63,7 @@ Guidelines for rosdep rules
        However please don't target 'sid' as it's a rolling target and when the keys change our database gets out of date.
   * Keep everything in alphabetical order for better merging.
   * No trailing whitespace.
+  * No blank line.
 
 ### Expected Rosdep Sources
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ Guidelines for rosdep rules
        However please don't target 'sid' as it's a rolling target and when the keys change our database gets out of date.
   * Keep everything in alphabetical order for better merging.
   * No trailing whitespace.
-  * No blank line.
+  * No blank lines.
 
 ### Expected Rosdep Sources
 
--- a/scripts/check_rosdep.py
+++ b/scripts/check_rosdep.py
@@ -57,6 +57,15 @@ def no_trailing_spaces(buf):
     return clean
 
 
+def no_blank_lines(buf):
+    clean = True
+    for i, l in enumerate(buf.split('\n')[:-1]):
+        if re.match(r'^\s*$', l):
+            print_err("empty line %u" % (i+1))
+            clean = False
+    return clean
+
+
 def generic_parser(buf, cb):
     ilen = len(indent_atom)
     stringblock = False
@@ -166,6 +175,8 @@ def main(fname):
     if ydict != {}:
         print_test("checking for trailing spaces...")
         my_assert(no_trailing_spaces(buf))
+        print_test("checking for blank lines...")
+        my_assert(no_blank_lines(buf))
         print_test("checking for incorrect indentation...")
         my_assert(correct_indent(buf))
         print_test("checking for non-bracket package lists...")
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -63,6 +63,7 @@ Guidelines for rosdep rules
        However please don't target 'sid' as it's a rolling target and when the keys change our database gets out of date.
   * Keep everything in alphabetical order for better merging.
   * No trailing whitespace.
+  * No blank line.
```
clalancette, line 66:
```suggestion
  * No blank lines.
```

scripts/check_rosdep.py
```diff
@@ -57,6 +57,15 @@ def no_trailing_spaces(buf):
     return clean
 
 
+def no_blank_lines(buf):
+    clean = True
+    for i, l in enumerate(buf.split('\n')[:-1]):
```
clalancette, line 62:
```suggestion
    for i, line in enumerate(buf.split('\n')[:-1]):
```

clalancette (commented):
Two small things to fix, but otherwise looks good to me.  Thanks for the change.

CONTRIBUTING.md
incebellipipo: done it in 89a360c

scripts/check_rosdep.py
incebellipipo: I just followed convention as it is used in one function above and below.
- https://github.com/incebellipipo/rosdistro/blob/89a360c6d17078f1603b70119d16b5ba72dcb99d/scripts/check_rosdep.py#L53
- https://github.com/incebellipipo/rosdistro/blob/89a360c6d17078f1603b70119d16b5ba72dcb99d/scripts/check_rosdep.py#L76

If you think it should change it `l`->`line`, than I probably should change every single variable that indicates a line. It seems fine to me though, also fits with previous convention.

clalancette: Good call, that's a fair argument.  I don't like single-letter variable names in general (and neither does newer versions of flake8), but I'd rather fit into convention here.

clalancette (approved):

PR #24942, opened 2020-05-18
nmc_nlp_lite-release: 0.0.5-1 in 'melodic/distribution.yaml' [bloom]

nmcbins:
Increasing version of package(s) in repository `nmc_nlp_lite-release` to `0.0.5-1`:

- upstream repository: file:///mnt/c/inetpub_nlp/wwwroot/LN_home/SDK/catkin_ws/src/nmc_nlp_lite
- release repository: https://github.com/nmcbins/nmc_nlp_lite-release.git
- bloom version: `0.9.7`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6078,8 +6078,8 @@ repositories:
   nmc_nlp_lite-release:
     doc:
       type: git
-      url: https://github.com/nmcbins/nmc_nlp_lite-release.git
-      version: 0.0.5
+      url: https://github.com/nmcbins/nmc_nlp_lite_ros.git
+      version: master
     release:
       packages:
       - nmc_nlp_lite
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6075,6 +6075,18 @@ repositories:
       type: git
       url: https://github.com/NiryoRobotics/niryo_one_ros_simulation.git
       version: master
+  nmc_nlp_lite-release:
+    doc:
+      type: git
+      url: https://github.com/nmcbins/nmc_nlp_lite-release.git
+      version: 0.0.5
+    release:
+      packages:
+      - nmc_nlp_lite
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/nmcbins/nmc_nlp_lite-release.git
+      version: 0.0.5-1
   nmea_comms:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -6075,6 +6075,18 @@ repositories:
       type: git
       url: https://github.com/NiryoRobotics/niryo_one_ros_simulation.git
       version: master
+  nmc_nlp_lite-release:
+    doc:
+      type: git
+      url: https://github.com/nmcbins/nmc_nlp_lite-release.git
+      version: 0.0.5
```
clalancette, line 6082: This should point to the release repository, and must point to a branch (not a tag).

clalancette (changes requested):

melodic/distribution.yaml
nmcbins: Thank you so much! I will fix it and retry!

nmcbins:
I'd appreciate it very much if anyone can tell me what actions I need to take from here ... 
I am learning about the system and cannot seem to get a clear idea about the process (I got the general idea, like someone will merge this pull request ... but don't know what triggers that)
Thanks again!

clalancette:
> I'd appreciate it very much if anyone can tell me what actions I need to take from here ...
> I am learning about the system and cannot seem to get a clear idea about the process (I got the general idea, like someone will merge this pull request ... but don't know what triggers that)

Yes, generally when we are satisfied that the package meets most of our standards, we'll go ahead and merge it.  At that point it will be built on the build farm, and (assuming it succeeds) will be available in the next sync.

For this PR, there are 2 minor and 2 major things that has been preventing me from merging it:
1.  (Minor) There is no need for this [LICENSE](https://github.com/nmcbins/nmc_nlp_lite-release/blob/master/LICENSE) file in the release repository.  It doesn't hurt anything, but there isn't really "code" there to LICENSE (it is all auto-generated).
1.  (Minor) The [vcs_uri](https://github.com/nmcbins/nmc_nlp_lite-release/blob/master/tracks.yaml#L24) should point essentially back to the same repository, so it should be `https://github.com/nmcbins/nmc_nlp_lite-release/`.
1.  (Major) There is no LICENSE file on the upstream source repository: https://github.com/nmcbins/nmc_nlp_lite_ros
1.  (Major) This really looks like an example package, not something we should release in the buildfarm.
    1.  https://github.com/nmcbins/nmc_nlp_lite_ros/blob/master/srv/nmcNLP.srv uses the name of "str" for both of the service fields.  We highly recommend that you use descriptive, semantic names for the names of fields in topics and services.  Otherwise nobody knows what it means.
     1.  The [main program](https://github.com/nmcbins/nmc_nlp_lite_ros/blob/master/nmc_nlp_lite_server.py) here uses a dependency that doesn't exist (nmcbins).  So I don't think it will build on the buildfarm, and a user that installs it will have no hope of it actually working.

I think we should backup here and figure out what it is you are trying to accomplish with the package; we then may be able to find a way forward.

nmcbins:
This is SOOOOO helpful!  Thanks! 
In short, we have built a service to convert natural language instructions to robot understandable instructions -- it's cloud based, I am trying to wrap it into a ROS package.  Hence I started learning about writing/publishign this empty ROS package as a stub -- you are right that this is an example package, I will close the PR, and resubmit, after I fix all the problems, and paste in real code & functionality.

All your 4 points are precisely very helpful, as those are questions I searched and searched and couldn't find answer for. Especially point 4-ii,  I was wondering what about dependency to a python module (nmcbins, hosted by pypi.org but not yet included in ROS) ... so for now I'll just include the nmcbins source code inside  nmc_nlp_lite to avoid the dependency.

Thanks again.

PR #24944, opened 2020-05-18
Add release of mrpt2

jlblancoc:
(manual edit after bloom failed to create automatic PR)

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -5578,6 +5578,15 @@ repositories:
       version: mrpt-1.5
     status: maintained
   mrpt2:
+    doc:
+      type: git
+      url: https://github.com/MRPT/mrpt.git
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -5578,6 +5578,15 @@ repositories:
       version: mrpt-1.5
     status: maintained
   mrpt2:
+    doc:
+      type: git
+      url: https://github.com/MRPT/mrpt.git
```
clalancette, line 5583:
```suggestion
      url: https://github.com/mrpt/mrpt.git
```

(this is a really small nit, as github is case-insensitive, but I think it would be better for it to be consistent)

clalancette (changes requested):

jlblancoc:
Done! :+1:

PR #24954, opened 2020-05-19
rosdep: MQTT C++ development package

sevenbill:
Adding the C++ dev package for MQTT. Though, it looks like Fedora has the C++ development files in `mosquitto-devel` so alternatively the existing key can be updated to include the C++ packages.
```
mosquitto-dev:
  debian: [libmosquitto-dev, libmosquittopp-dev]
  fedora: [mosquitto-devel]
  ubuntu: [libmosquitto-dev, libmosquittopp-dev]
```

https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=libmosquittopp-dev
https://packages.debian.org/search?keywords=libmosquittopp-dev&searchon=names&suite=stable&section=all
https://apps.fedoraproject.org/packages/mosquitto-devel/contents

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4875,6 +4875,7 @@ mosquitto-dev:
   ubuntu: [libmosquitto-dev]
 mosquittopp-dev:
   debian: [libmosquittopp-dev]
+  fedora: [mosquitto-devel]
   ubuntu: [libmosquittopp-dev]
 mpg123:
   arch: [mpg123]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4873,6 +4873,9 @@ mosquitto-dev:
   debian: [libmosquitto-dev]
   fedora: [mosquitto-devel]
   ubuntu: [libmosquitto-dev]
+mosquittopp-dev:
+  debian: [libmosquittopp-dev]
+  ubuntu: [libmosquittopp-dev]
 mpg123:
   arch: [mpg123]
   debian: [mpg123]


rosdep/base.yaml
```diff
@@ -4670,6 +4670,9 @@ mosquitto-dev:
   debian: [libmosquitto-dev]
   fedora: [mosquitto-devel]
   ubuntu: [libmosquitto-dev]
+mosquittopp-dev:
+  debian: [libmosquittopp-dev]
+  ubuntu: [libmosquittopp-dev]
```
clalancette, line 4675: For development packages in particular, I think that we should be aiming for ease-of-use, rather than as-small-as-possible.  Thus, I think that it would be better to fold these into the `mosquitto-dev` key above.  @ros/rosdeputies , thoughts?

tfoote (commented):
I think if they package it separately it's not really for us to judge if they're the same. They appear to present non-overlapping content. In particular this would inject the c++ version onto what someone might want to depend on the c implementation alone which is a case that might be valuable for something like cross compiling. 

@sevenbill I think there's at least parallel implementations on fedora that we should add for parity.

sevenbill:
Fedora groups the C and C++ implementations in the same package.
https://apps.fedoraproject.org/packages/mosquitto-devel/contents

clalancette:
> I think if they package it separately it's not really for us to judge if they're the same.

To be fair, we make this judgement in other cases (the [apr key](https://github.com/ros/rosdistro/blob/700271941ec3327cd8b35e3e2a7f609891513b3d/rosdep/base.yaml#L66) is one example).

> They appear to present non-overlapping content. In particular this would inject the c++ version onto what someone might want to depend on the c implementation alone which is a case that might be valuable for something like cross compiling.

Right, and for non-dev keys, I agree with you.  But for the `-dev` keys in particular, I think adding additional stuff is generally fine; the worst case is that your development machine gets additional dependencies installed that it doesn't need.

sloretz:
> Right, and for non-dev keys, I agree with you. But for the `-dev` keys in particular, I think adding additional stuff is generally fine; the worst case is that your development machine gets additional dependencies installed that it doesn't need.

From a reviewer perspective, it seems easier to mimic the upstream packaging rather than combining for convenience. Also it seems like `-dev` keys can affect more than space used on a development machine. Say `mosquitto-dev` was available for armhf, but `mosquitopp-dev` was not. IIUC I think that means a combined `mosquito-dev` rosdep key would be unresolvable when building for `armhf`.

In this specific case mosquitopp seems to be deprecated, so I think it would make sense to not include it by default. https://github.com/eclipse/mosquitto/issues/1386

> Fedora groups the C and C++ implementations in the same package.
https://apps.fedoraproject.org/packages/mosquitto-devel/contents

If the C++ interface is in the same package then I think this key still needs a `fedora` entry to make sure that package gets installed when a `package.xml` only specifies `mosquitopp-dev`.

nuclearsandwich:
:+1: to separate keys tracking debian packages. I think we've gotten more feedback anecdotally saying  keys should be decomposed for better precision and smaller images than combined for ease of use.

Before I can approve I'd like to see the fedora key added.

clalancette:
All right, looks like consensus is to keep the packages separated.  I'll re-review with that in mind.

rosdep/base.yaml
clalancette, line 4675: We've agreed that this should be a separate key like this.  We also know that Fedora bundles mosquittopp inside of mosquitto-devel.

With those two pieces of information, I think we should make the Fedora key for `mosquittopp-dev` the following:

```
  fedora: [mosquitto-devel]
```

That way it will still resolve (even though it would be a duplicate of mosquitto-dev).

clalancette (changes requested):
One thing to fix, then I think this can go in.

claireyywang:
@sevenbill friendly ping

rosdep/base.yaml
clalancette: I've gone ahead and done this now.

clalancette (approved):

claireyywang (approved):

PR #24964, opened 2020-05-19
Replace the pip python-rtree dependency for the apt package

musamarcusso:
The current `pip` package for `rtree` is currently leading to problems regarding it's non-Python dependencies. This PR sets the `apt` packages for `rtree`  as dependencies for the `pcg-gazebo` installations.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2969,11 +2969,11 @@ python-pcapy:
 python-pcg-gazebo-pip:
   debian:
     pip:
-      depends: [libspatialindex-dev, geos, pybind11-dev]
+      depends: [libspatialindex-dev, geos, pybind11-dev, python-rtree]
       packages: [pcg-gazebo]
   ubuntu:
     pip:
-      depends: [libspatialindex-dev, geos, pybind11-dev]
+      depends: [libspatialindex-dev, geos, pybind11-dev, python-rtree]
       packages: [pcg-gazebo]
 python-pep8:
   arch: [python2-pep8]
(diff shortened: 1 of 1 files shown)


sloretz:
To be honest, I'm not 100% sure how to review this one. 

One thing that would help is [adding web links](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions) to the `python-rtree` and `python3-rtree` packages on non-EOL Debian and Ubuntu platforms. Mind doing so?

tfoote:
The package appears to be documented here: https://pypi.org/project/pcg-gazebo/ But in only lists extra dependencies `libspatialindex-dev pybind11-dev libgeos-dev` which look like the existing dependency list. Is this something that should be documented upstream first? Or has something changed?

@sloretz Note that those are references to rosdep keys that already exist. ([py2](https://github.com/ros/rosdistro/blob/9cbc9496247a677be3dabfe727d340fd48dd60a7/rosdep/python.yaml#L4055), [py3](https://github.com/ros/rosdistro/blob/9cbc9496247a677be3dabfe727d340fd48dd60a7/rosdep/python.yaml#L6034)) This is just about adding a dependency to make sure it's installed first.

musamarcusso:
Hi @sloretz, @tfoote thanks for the feedback. I am updating the documentation. The issue is mainly that the `rtree` package for some reason has been showing issues when installed with `pip` and makes then the installation of `pcg-gazebo` fail. I will post it here as soon as the docs are updated.

clalancette:
@musamarcusso I'm going to close this out for lack of activity.  Please feel free to reopen if that is in error, or if you finish the documentation and want to update it.  Thanks.

PR #24982, opened 2020-05-20
Added bullet extras for debian and ubuntu

BryceStevenWilley:
The Bullet Collision library contains some useful extra features like Inverse Dynamics, file serialization, and some packaged versions of HACD/Convex decomposition. These extras aren't installed by default, so I've added them here.
 
* Debian: https://packages.debian.org/unstable/libbullet-extras-dev
* Ubuntu: https://packages.ubuntu.com/bionic/libs/libbullet-extras-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -321,9 +321,11 @@ bullet:
     '*': [libbullet-dev]
     trusty_python3: [libbullet-dev]
 bullet-extras:
+  arch: [bullet]
   debian: [libbullet-extras-dev]
-  ubuntu:
-    '*': [libbullet-extras-dev]
+  fedora: [bullet-extras]
+  gentoo: [sci-physics/bullet]
+  ubuntu: [libbullet-extras-dev]
 bzip2:
   alpine: [bzip2-dev]
   arch: [bzip2]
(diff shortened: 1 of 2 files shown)


tfoote (changes requested):
The wildcard ( `'*'` ) syntax without a second rule can be collapsed to the one line style without any version callout. 

And while we're reviewing this are these extras available on Fedora Gentoo Arch, or homebrew?

BryceStevenWilley:
I didn't add those OSs at first since I couldn't test on them, but I added the other packages that I could find. If the extra libs were installed by default in the bullet package, I just added the same bullet target to bullet-extras.

* Fedora: https://pkgs.org/download/bullet-extras
* Gentoo: extra libs are installed [by default](https://gitweb.gentoo.org/repo/gentoo.git/tree/sci-physics/bullet/bullet-2.86.ebuild#n49)
* Arch: extra libs are installed [by default](https://www.archlinux.org/packages/extra/x86_64/bullet/)
* Homebrew: extras libs are installed [by default](https://github.com/Homebrew/homebrew-core/blob/287e14d47d8c0c0fdf3f8c4d7ee34f52cbfe56de/Formula/bullet.rb#L22)

tfoote (approved):
Thanks for filling it in. There's a balancing act between keeping things uniform and being fully tested. If there's good documentation we will generally take the more uniform approach.

clalancette (approved):

PR #25016, opened 2020-05-22
rosdep: Add libgeographic for ubuntu eouan and focal (noetic)

vooon:
Links to Ubuntu packages:

- https://packages.ubuntu.com/en/eoan/libgeographic-dev
- https://packages.ubuntu.com/en/focal/libgeographic-dev

Required for Mavros.


tfoote (changes requested):
These all look to be the same so there's no reason for us to call out the specific version. Could you just collapse it into a generic ubuntu rule and then we won't have to worry about this in the future.

Otherwise we'd want fill in cosmic and disco as well

vooon:
Before vivid it called libgeographic**lib**-dev, but since all of ros released targeted to 14.04 EOLed, it think it's ok to collapse to libgeographic-dev.

tfoote (approved):
Yeah, we don't worry about tracking differences into significantly EOL'd platforms.

Thanks for the update. LGTM.

sloretz (approved):

PR #25075, opened 2020-05-26
add wxpython for bullseye/focal

k-okada:
c.f. 
~~https://packages.debian.org/sid/python-wxgtk3.0~~
https://packages.ubuntu.com/focal/python-wxgtk3.0

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6306,10 +6306,7 @@ wxpython:
   opensuse: [python-wxGTK]
   rhel: [wxPython-devel]
   ubuntu:
-    artful: [python-wxgtk3.0]
-    bionic: [python-wxgtk3.0]
-    cosmic: [python-wxgtk3.0]
-    focal: [python-wxgtk3.0]
+    '*': [python-wxgtk3.0]
     lucid: [python-wxgtk2.8]
     maverick: [python-wxgtk2.8]
     natty: [python-wxgtk2.8]
@@ -6323,9 +6320,6 @@ wxpython:
     utopic: [python-wxgtk2.8]
     vivid: [python-wxgtk2.8]
     wily: [python-wxgtk2.8]
-    xenial: [python-wxgtk3.0]
-    yakkety: [python-wxgtk3.0]
-    zesty: [python-wxgtk3.0]
 yapf:
   arch: [yapf]
   debian:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6309,6 +6309,7 @@ wxpython:
     artful: [python-wxgtk3.0]
     bionic: [python-wxgtk3.0]
     cosmic: [python-wxgtk3.0]
+    focal: [python-wxgtk3.0]
     lucid: [python-wxgtk2.8]
     maverick: [python-wxgtk2.8]
     natty: [python-wxgtk2.8]


rosdep/python.yaml
```diff
@@ -6309,6 +6309,7 @@ wxpython:
     artful: [python-wxgtk3.0]
     bionic: [python-wxgtk3.0]
     cosmic: [python-wxgtk3.0]
+    focal: [python-wxgtk3.0]
```
clalancette, line 6312: For all of our currently-supported Ubuntu releases, the key is `python-wxgtk3.0`.  So I think we should just replace this whole block with:

```
ubuntu: [python-wxgtk3.0]
```

Which has the added benefit of making us forwards-compatible.

clalancette (changes requested):

rosdep/python.yaml
k-okada: @clalancette thanks! chaned to keep both backward/forward compatible

rosdep/python.yaml
```diff
@@ -6306,9 +6306,7 @@ wxpython:
   opensuse: [python-wxGTK]
   rhel: [wxPython-devel]
   ubuntu:
-    artful: [python-wxgtk3.0]
-    bionic: [python-wxgtk3.0]
-    cosmic: [python-wxgtk3.0]
+    '*': [python-wxgtk3.0]
```
clalancette, line 6309: I think we should change it to exactly:

```
ubuntu: [python-wxgtk3.0]
```

And remove everything else.  We don't need to keep the keys for unsupported distributions anymore.

clalancette (changes requested):

rosdep/python.yaml
k-okada: humm, for me, `rosdep` is so general that we can use non-ROS environment, and also we still have a robot with outdated Ubuntu distros (mostly because of old proprietary  device drivers for sensor/actuator I/Os. So we prefer backward compatibility.

sloretz:
> python-wxgtk3.0

Note that this Debian package depends on Python 2. It shouldn't be used in a package released into Noetic and Foxy because they are both targeting Python 3.

rosdep/python.yaml
Timple: Using the `ubuntu` keyword as suggested will still allow you to use old ubuntu distro's I assume.

k-okada:
@sloretz  not sure if I understand you comment correctly, `python-wxgtk3.` means `gtk3.0` on `python2`. For python3 they have `python3-wxgtk4.0`

18.04
```
Package: python-wxgtk3.0
Version: 3.0.2.0+dfsg-7
Priority: optional
Section: universe/python
Source: wxpython3.0
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: wxWidgets Maintainers <freewx-maint@lists.alioth.debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 29.0 MB
Provides: python2.7-wxgtk3.0
Depends: python-wxversion, python (<< 2.8), python (>= 2.7~), python:any (<< 2.8), python:any (>= 2.7.5-5~), libc6 (>= 2.14), libgcc1 (>= 1:3.0), libstdc++6 (>= 5.2), libwxbase3.0-0v5 (>= 3.0.4+dfsg), libwxgtk3.0-gtk3-0v5 (>= 3.0.4+dfsg)
```
20.04
```
Package: python-wxgtk3.0
Version: 3.0.2.0+dfsg-8build1
Priority: optional
Section: universe/python
Source: wxpython3.0
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: wxWidgets Maintainers <team+wx@tracker.debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 30.5 MB
Provides: python2.7-wxgtk3.0
Depends: python-wxversion, python2 (<< 2.8), python2 (>= 2.7~), python2:any (<< 2.8), python2:any (>= 2.7~), libc6 (>= 2.14), libgcc1 (>= 1:3.0), libstdc++6 (>= 5.2), libwxbase3.0-0v5 (>= 3.0.4+dfsg), libwxgtk3.0-gtk3-0v5 (>= 3.0.4+dfsg-10~)

```

sloretz:
> @sloretz not sure if I understand you comment correctly, python-wxgtk3. means gtk3.0 on python2

Yup. Since it depends on Python 2, this rosdep key shouldn't be used by any ROS packages in ROS Noetic or ROS Foxy. If the plan was to release a ROS package into Noetic or Foxy that currently depends on the rosdep key `wxpython` in Melodic or earlier, it should be updated to not use the `wxpython` rosdep key, but instead use a (to be created) `python3-wxgtk4.0` rosdep key.

rosdep/python.yaml
nuclearsandwich: > Using the `ubuntu` keyword as suggested will still allow you to use old ubuntu distro's I assume.

Only if the package name is the same for all distributions. In this case distros prior to xenial resolve to `python-wxgtk2.8` not `python-wxgtk3.0`.

nuclearsandwich (approved):
@clalancette has requested a simplification which is only possible by dropping distributions which are no longer supported. I'm in general in favor of that but also fine honoring specific requests from *contributors* to *maintain* current levels of support in our rosdep db.

Which is to say, I don't think we should add new keys, or restore dropped keys for older distributions on request since we can instead push individuals who need those to custom rosdep sources. But I don't see a problem with delaying the reduction of support based on contributor feedback.

clalancette:
> Which is to say, I don't think we should add new keys, or restore dropped keys for older distributions on request since we can instead push individuals who need those to custom rosdep sources. But I don't see a problem with delaying the reduction of support based on contributor feedback.

That logic works for me.  I'll re-review here.

clalancette (approved):

PR #25113, opened 2020-05-27
Add python-opencv-pip

MatthijsBurgh:
See https://pypi.org/project/opencv-python/


clalancette:
Two observations here:

1.  Do you really want to add a Python2 key?
1.  Do you really need the `-pip` key?  There is the debian package key right above, which is more useful for the buildfarm.

Neither question above prevents this from going in, but I'm curious on what the use-case here is.

clalancette:
No response in over a month.  Closing, but feel free to reopen and answer the questions above.

PR #25120, opened 2020-05-27
Simplify python-sip

MatthijsBurgh:
Removes outdated `_python3` suffixes. Therefore all distros of ubuntu can be combined. python3-sip entry is already in rosdistro, so this shouldn't cause any problems.

Fixes one part of https://github.com/orocos/orocos_kinematics_dynamics/issues/249#issuecomment-634758907


clalancette (approved):

rosdep/python.yaml
```diff
@@ -4294,31 +4294,7 @@ python-sip:
   slackware:
     slackpkg:
       packages: [sip]
-  ubuntu:
-    artful: [python-sip-dev]
-    artful_python3: [python3-sip-dev]
-    bionic: [python-sip-dev]
-    bionic_python3: [python3-sip-dev]
```
tfoote, line 4301: Please leave the existing `_python3` rules so that we don't break people who might be using them. Our general policy is to let them age out when the platform goes EOL. 

We can get the same forwards compatibilty by using a `'*': [python-sip-dev]` rule next to just the legacy python3 rules while not breaking those using the deprecated style.

tfoote (changes requested):

clalancette (approved):

MatthijsBurgh:
ping @tfoote

rosdep/python.yaml
```diff
@@ -4291,29 +4291,13 @@ python-sip:
     slackpkg:
       packages: [sip]
   ubuntu:
-    artful: [python-sip-dev]
+    '*': [python-sip-dev]
     artful_python3: [python3-sip-dev]
```
sloretz, line 4295: If I understand @tfoote's answer, the `_python3` keys on EOL platforms may be removed. Looks like the EOL ones are

* `artful_python3`
* `trust_python3`
* `wily_python3`
* `yakkety_python3`
* `zesty_python3`

MatthijsBurgh: Thank, I removed these.

sloretz (approved):

PR #25140, opened 2020-05-28
add focal / libcollada-dom

k-okada:
c.f. https://packages.ubuntu.com/focal/libcollada-dom2.4-dp-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -482,6 +482,7 @@ collada-dom:
   ubuntu:
     artful: [libcollada-dom2.4-dp-dev]
     bionic: [libcollada-dom2.4-dp-dev]
+    focal: [libcollada-dom2.4-dp-dev]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -482,6 +482,7 @@ collada-dom:
   ubuntu:
     artful: [libcollada-dom2.4-dp-dev]
     bionic: [libcollada-dom2.4-dp-dev]
+    focal: [libcollada-dom2.4-dp-dev]
```
clalancette, line 485: At this point, the name of this key for all supported Ubuntu distributions is `libcollada-dom2.4-dp-dev`.  For future compatibility, I think we can collapse this whole key into:

```
ubuntu: [libcollada-dom2.4-dp-dev]
```

clalancette (changes requested):

clalancette (approved):

PR #25143, opened 2020-05-28
Update distribution.yaml

smoker771:


clalancette (changes requested):
A few questions and things to change:

1.  Is there a particular reason you didn't use `bloom-release` to do this release?  It does these modifications for you, and adds metadata that helps us.
1.  Do you want to add a source entry to point to your sources here?
1.  Your source repository lists `BSD` as the license in the package.xml, but doesn't have a corresponding `LICENSE` file.  Please add that and do a new release.

PR #25148, opened 2020-05-28
Update distribution.yaml

smoker771:


smoker771:
Unfortunately 'bloom-release' failed to make pull request, so I make them manually. Sorry

melodic/distribution.yaml
```diff
@@ -5048,6 +5048,17 @@ repositories:
...
       version: master
     status: developed
+  marvelmind_nav:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/MarvelmindRobotics/marvelmind_nav-release.git
+      version: 1.0.10-1
+    source:
+      test_pull_requests: false
```
tfoote, line 5058: We can only support pull request builds on github due to using https://plugins.jenkins.io/ghprb/

tfoote (changes requested):
One small fixup to remove the PR builds. 

I also noticed browsing your repository that you don't have the BSD headers in each of your source files. As it's a file based license all source files should have it at the top of the file.

tfoote:
And I'd highly recommend debugging bloom opening PRs automatically. If you're using 2 factor you need to do a little bit extra to support it: https://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization 

The automated PRs are much less prone to small typos that cause extra iterations over time. They also have better titles etc which makes them easier to review.

melodic/distribution.yaml
smoker771: I just added license headers to all source files in the source repository. Do I need to create new release and/or pull request? Thank you

clalancette: Yes, that's the easiest thing to do.  Bloom will automatically reopen the PR for you.

PR #25171, opened 2020-05-28
fadecandy_ros: 0.1.0-1 in 'noetic/distribution.yaml' [bloom]

jonbinney:
Increasing version of package(s) in repository `fadecandy_ros` to `0.1.0-1`:

- upstream repository: https://github.com/iron-ox/fadecandy_ros.git
- release repository: https://github.com/iron-ox/fadecandy_ros-release.git
- bloom version: `0.9.7`

## fadecandy_driver

```
* Initial release.
```

## fadecandy_msgs

```
* Initial release
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -487,6 +487,20 @@ repositories:
       url: https://github.com/wxmerkt/exotica_val_description-release.git
       version: 1.0.0-1
     status: maintained
+  fadecandy_ros:
+    release:
+      packages:
+      - fadecandy_driver
+      - fadecandy_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/iron-ox/fadecandy_ros-release.git
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/iron-ox/fadecandy_ros.git
+      version: master
+    status: developed
   fcl_catkin:
     release:
       tags:


sloretz (changes requested):
It looks like this [depends on](https://github.com/iron-ox/fadecandy_ros/blob/d1daba45829d9ca100e34cf9719c43547ef9e7bb/fadecandy_driver/package.xml#L15) the rosdep key [python-usb](https://github.com/ros/rosdistro/blob/9f35018ae8fa34b97ed7de4a54bf644ff83efbc2/rosdep/python.yaml#L4969-L4973) which is [a Python 2 package](https://packages.ubuntu.com/focal/python-usb), but Noetic targets Python3.

I think this could be resolved by adding a rosdep key [python3-usb](https://packages.ubuntu.com/focal/python3-usb) and [using conditional dependencies on `ROS_PYTHON_VERSION`](http://wiki.ros.org/UsingPython3/SourceCodeChanges#Conditional_Dependencies).

One other thing I noticed is the [install of `fadecandy_node`](https://github.com/iron-ox/fadecandy_ros/blob/d1daba45829d9ca100e34cf9719c43547ef9e7bb/fadecandy_driver/CMakeLists.txt#L10-L13) should use [catkin_install_python() so the shebang in the installed script is correct](http://wiki.ros.org/UsingPython3/SourceCodeChanges#Changing_shebangs), otherwise it likely won't be runable.

jonbinney:
Good catch, thanks @sloretz ! I'll fix that and re-release.

jonbinney:
Looks like python3-usb doesn't exist as a rosdep key yet. I'll make a PR to fix that first.

PR #25192, opened 2020-05-29
Update pyhton.yaml. Add python-pypcd-pip.

ilozano80:
Add python-pypcd-pip to python.yaml file.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3412,6 +3412,10 @@ python-pynmea2:
   ubuntu:
     pip:
       packages: [pynmea2]
+python-pypcd-pip:
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3412,6 +3412,10 @@ python-pynmea2:
   ubuntu:
     pip:
       packages: [pynmea2]
+python-pypcd-pip:
```
clalancette, line 3415: A key named `python-pypcd-pip` implies a Python 2 version of the key.  Are you really looking for the Python 2 version, or the Python 3 one?

ilozano80: Python 2 version

clalancette (approved):

PR #25271, opened 2020-06-02
[foxy] updated the dev branch of behaviortree_cpp_v3

seanyen:
The `ros2` is removed from the upstream so I figured it needs a update.


seanyen:
@facontidavide fyi.

jacobperron:
@facontidavide Can you confirm if `master` is the correct branch to point to for Foxy?

facontidavide:
yes, it is

jacobperron:
Thanks!

PR #25273, opened 2020-06-02
rosparam_shortcuts: 0.4.0-1 in 'noetic/distribution.yaml' [bloom]

JafarAbdi:
Increasing version of package(s) in repository `rosparam_shortcuts` to `0.4.0-1`:

- upstream repository: https://github.com/PickNikRobotics/rosparam_shortcuts.git
- release repository: https://github.com/PickNikRobotics/rosparam_shortcuts-release.git
- bloom version: `0.9.7`

## rosparam_shortcuts

```
* bump cmake version (#13 <https://github.com/PickNikRobotics/rosparam_shortcuts/issues/13>)
  get rid of CMP0048 warning.
  https://github.com/ros-planning/geometric_shapes/pull/129
* Contributors: Michael Görner
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2399,6 +2399,21 @@ repositories:
       url: https://github.com/ros/rospack.git
       version: noetic-devel
     status: maintained
+  rosparam_shortcuts:
+    doc:
+      type: git
+      url: https://github.com/davetcoleman/rosparam_shortcuts.git
+      version: noetic-devel
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/PickNikRobotics/rosparam_shortcuts-release.git
+      version: 0.4.0-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/PickNikRobotics/rosparam_shortcuts.git
+      version: noetic-devel
   rospy_message_converter:
     doc:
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -2399,6 +2399,21 @@ repositories:
...
+      version: noetic-devel
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/PickNikRobotics/rosparam_shortcuts-release.git
+      version: 0.4.0-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/PickNikRobotics/rosparam_shortcuts.git
```
sloretz, line 2415: I'm surprised the `doc` and `source` sections have different URLs. Is that intended? If not, which is correct: `davetcoleman` or `PickNicRobotics`?

JafarAbdi: Thanks @sloretz for catching it, I didn't intend to have them different but seem like it's ok since melodic does have the same URLs for both https://github.com/ros/rosdistro/blob/master/melodic/distribution.yaml#L9343

clalancette: It's certainly allowed, but its not recommended.  I'd suggest making it consistent here, and then updating Melodic to match.

PR #25328, opened 2020-06-04
urg_node: 0.1.14-1 in 'noetic/distribution.yaml' [bloom]

tonybaltovski:
Increasing version of package(s) in repository `urg_node` to `0.1.14-1`:

- bloom version: `0.9.7`

## urg_node

```
* Bumped CMake version.
* Removed trailing whitespace.
* Contributors: Tony Baltovski
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -3389,6 +3389,21 @@ repositories:
       url: https://github.com/ros-drivers/urg_c.git
       version: master
     status: maintained
+  urg_node:
+    doc:
+      type: git
+      url: https://github.com/ros-drivers/urg_node.git
+      version: kinetic-devel
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/ros-gbp/urg_node-release.git
+      version: 0.1.14-1
+    source:
+      type: git
+      url: https://github.com/ros-drivers/urg_node.git
+      version: kinetic-devel
+    status: maintained
   urg_stamped:
     doc:
       type: git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -16559,7 +16559,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-drivers/urg_node.git
-      version: indigo-devel
+      version: kinetic-devel
     release:
       tags:
         release: release/kinetic/{package}/{version}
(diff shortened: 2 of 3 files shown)


sloretz:
holding for the [upcoming Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2020-06-05/14451)

noetic/distribution.yaml
```diff
@@ -3389,6 +3389,21 @@ repositories:
       url: https://github.com/ros-drivers/urg_c.git
       version: master
     status: maintained
+  urg_node:
+    doc:
+      type: git
+      url: https://github.com/ros-drivers/urg_node.git
+      version: kinetic-devel
```
clalancette, line 3396: The Kinetic and Melodic versions of this package are both on the "indigo-devel" branch.  The noetic one here is being setup for "kinetic-devel", which is fine.  But it seems like if "kinetic-devel" supports everything back to Kinetic, then all of Kinetic, Melodic, and Noetic should use that same branch.  Thoughts?

clalancette (changes requested):

noetic/distribution.yaml
tonybaltovski: I agree, the only reason for the kinetic-devel was to bump the CMakeLists version.

tonybaltovski: Do you want me to update the tracks now?

clalancette: Yeah, you can go ahead and do that.  Typically it would be separate PRs, but right now all of Kinetic, Melodic, and Noetic are *not* in freeze.  If we get it in today, we can do all of them at once.

tonybaltovski: Whenever I run `bloom-release --edit-track -r kinetic -t kinetic urg_node` the information is pre-filled with kinetic-devel.  However, the PR that is generated doesn't change indigo-devel to kinetic-devel.  Any idea why not?

clalancette: Hm, sorry, no idea here.  In theory bloom can do this, but I typically do it manually.

sloretz (approved):
@clalancette @tfoote I'm inclined to merge this. While Kinetic and Melodic are preparing for a sync, I think it would be ok to merge this PR because the kinetic/melodic changes are only the `source` and `doc` entries. Thoughts?

clalancette:
> @clalancette @tfoote I'm inclined to merge this. While Kinetic and Melodic are preparing for a sync, I think it would be ok to merge this PR because the kinetic/melodic changes are only the `source` and `doc` entries. Thoughts?

I just did the Melodic sync, so it is open anyway, so :+1: from me.

clalancette:
All of the distros are now open, so I'm going to go ahead and merge this.

clalancette (approved):

PR #25358, opened 2020-06-06
Add roswasm_suite to index

nilsbore:
Libraries for compiling C++ ROS nodes to Webassembly using Emscripten. Allows you to write C++ ROS nodes similar to how you would otherwise, and to then have them run in a webbrowser and communicate with ROS through rosbridge_websocket.


melodic/distribution.yaml
```diff
@@ -9439,6 +9439,11 @@ repositories:
       url: https://github.com/UTNuclearRoboticsPublic/rosthrottle-release.git
       version: 1.2.0-3
     status: maintained
+  roswasm_suite:
+    doc:
+      type: git
+      url: https://github.com/nilsbore/roswasm_suite.git
```
clalancette, line 9445: Looking at the source repository, it seems like https://github.com/nilsbore/roswasm_suite/blob/master/roscpp_json_serialize/package.xml (at least) still has a TODO for the license.  Can you update that please?

clalancette (changes requested):

clalancette:
No response in almost a month, so closing.  Please feel free to respond to the review and reopen.  Thanks for the contribution.

PR #25365, opened 2020-06-07
add leo_description and leo_viz packages to kinetic and melodic distributions

bjsowa:
packages for the Leo Rover platform

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6446,6 +6446,18 @@ repositories:
       url: https://github.com/ros-drivers/leap_motion.git
       version: hydro
     status: maintained
+  leo_description:
+    source:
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):
A few things about this pull request:
1.  The package.xml files in both repositories list the license as MIT (good), but neither has a LICENSE file.  Please add a LICENSE file to both repositories.
1.  In general, it's not a good idea to open a PR touching multiple distributions at once.  That's because the distributions are on different sync schedules, so it can be hard to find a time when they are both "unfrozen" so we can merge this.  I'll suggest that you close this PR and open two separate ones, one for each of Kinetic and Melodic.  If you don't want to do that, then we'll just have to wait on a time when both are available.

PR #25372, opened 2020-06-08
add rosdep focal key for python-websocket

tdenewiler:
Adding rule for new Ubuntu distribution.

* Ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python3-websocket&searchon=names

Local testing with forked repository showed rosdep key resolves to expected package. Uses Python2 version for Ubuntu Bionic and Debian Sid, and uses Python3 version for Ubuntu Focal.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5108,6 +5108,7 @@ python-websocket:
     artful_python3: [python3-websocket]
     bionic: [python-websocket]
     bionic_python3: [python3-websocket]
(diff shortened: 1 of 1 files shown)


clalancette:
This particular key is used for the (implicit) Python 2 version of the package.  There is already a `python3-websocket` key that includes this package for Python 3.  Given that, I'm going to close this.  Feel free to reopen if you feel this was closed in error.

PR #25376, opened 2020-06-08
add rosdep focal key for libjsoncpp

tdenewiler:
Used to read JSON data. An example is the Ouster lidar driver.
https://github.com/ouster-lidar/ouster_example/blob/master/ouster_client/package.xml#L8

* Ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libjsoncpp&searchon=names

Local testing with forked repository showed rosdep key resolves to expected package. Uses libjsoncpp0 version for Ubuntu Precise and uses libjsoncpp1 version for Ubuntu Bionic and Focal.


rosdep/base.yaml
```diff
@@ -2627,6 +2627,7 @@ libjsoncpp:
   ubuntu:
     artful: [libjsoncpp1]
     bionic: [libjsoncpp1]
+    focal: [libjsoncpp1]
```
clalancette, line 2630: It looks like all of our currently supported distributions have the package name as `libjsoncpp1`.  In order to automatically get support future Ubuntu distributions, I'll suggest that we change this whole key to:

```
ubuntu: [libjsoncpp1]
```

clalancette (changes requested):

rosdep/base.yaml
tdenewiler: Good idea. Done in 3a6c777.

clalancette:
Superseded by #25418, so closing.

PR #25394, opened 2020-06-10
Added yapf3 for ubuntu focal to python.yaml

belorenz:
`yapf` does not exist under Ubuntu focal but is available under `yapf3`.
I changed the python.yaml accordingly.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6399,6 +6399,7 @@ yapf:
     stretch: [yapf]
   ubuntu:
     '*': [yapf]
+    focal: [yapf3]
     saucy:
       pip:
         packages: [yapf]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6398,8 +6398,7 @@ yapf:
     buster: [yapf]
     stretch: [yapf]
   ubuntu:
-    '*': [yapf]
-    focal: [yapf3]
+    '*': [yapf]]
     saucy:
       pip:
         packages: [yapf]
@@ -6421,3 +6420,6 @@ yapf:
     yakkety:
       pip:
         packages: [yapf]
+yapf3: 
+  ubuntu: 
+    focal: [yapf3]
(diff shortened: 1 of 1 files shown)


clalancette:
Closing and reopening to kick github actions.

rosdep/python.yaml
```diff
@@ -6399,6 +6399,7 @@ yapf:
     stretch: [yapf]
   ubuntu:
     '*': [yapf]
+    focal: [yapf3]
```
clalancette, line 6402: So, I feel like this should be a separate key.  That's because in https://github.com/google/yapf/blob/master/README.rst , it says that you should run `yapf` for Python 2 code and `yapf3` for Python 3 code.  So it is clearly a different beast.

We may want to go even further and clean this `yapf` key up.  Since this is only supported up until eoan, we could just remove all of the entries for Ubuntu except for the `xenial` (pip) and `bionic` (debian) entries.  But I won't insist on that, you can do that if you feel like it.

clalancette (changes requested):

rosdep/python.yaml
belorenz: Could no follow your last paragraph and deciced to leave `yapf`as is. Added `yapf3` key as you suggested though. Thanks for the hint!

rosdep/python.yaml
```diff
@@ -6420,3 +6420,6 @@ yapf:
     yakkety:
       pip:
         packages: [yapf]
+yapf3: 
+  ubuntu: 
+    focal: [yapf3]
```
clalancette, line 6425: Since this is available on other distributions than focal, we should add support for them.  The most forwards-compatible way here would be this:

```
yapf3:
  debian: [yapf3]
  fedora: [python3-yapf]
  ubuntu:
    '*': [yapf3]
    xenial: null
```

clalancette (changes requested):

clalancette:
No response in over a month, so closing this out.  Please feel free to respond to the review comments and fix CI, then reopen.

PR #25395, opened 2020-06-10
add rosdep key for pika-pip

tdenewiler:
Pika is a RabbitMQ client library for Python, useful for interacting with devices and systems that use the AMQP protocol. The version available from the existing `pika` key on Ubuntu Bionic is `0.11.0-1`. The version from PyPI is `1.1.0`. There are many systems using the newer version.

* PyPI: https://pypi.org/project/pika/
* Ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python-pika&searchon=names

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -249,6 +249,22 @@ pika:
     pip:
       packages: [pika]
   ubuntu: [python-pika]
+pika-pip:
+  arch:
+    pip:
+      packages: [pika]
+  debian:
+    pip:
+      packages: [pika]
+  fedora:
+    pip:
+      packages: [pika]
+  gentoo:
+    pip:
+      packages: [pika]
+  ubuntu:
+    pip:
+      packages: [pika]
 pydocstyle:
   debian:
     buster: [pydocstyle]


clalancette (approved):

reinzor:
Related PR https://github.com/ros/rosdistro/pull/18840

nuclearsandwich (changes requested):
When packages are available via distribution package manager those packages should be used rather than adding additional pip keys.

Although it doesn't appear that the contribution guidelines were updated, the rationale in https://github.com/ros/rosdistro/pull/18840#issuecomment-426483773 still applies. Keys from pip cannot be used in ROS packages released via Bloom to the official ROS distributions or alternative ones.

I believe our recommendation for those who wish to use pip keys rather than system keys knowing the trade-offs is to add a custom rosdep source on systems where you want to use the pip keys.

tdenewiler:
I agree with the approach @tfoote is advocating for. I'll close this PR and add pika via pip in our Dockerfile or with a custom rosdep source. Thanks for the reviews.

PR #25424, opened 2020-06-11
carViz: 0.1.5-1 in 'melodic/distribution.yaml' [bloom]

pankhurivanjani:
Increasing version of package(s) in repository `carViz` to `0.1.5-1`:

- upstream repository: https://github.com/JdeRobot/carViz.git
- release repository: https://github.com/JdeRobot/carViz-release.git
- bloom version: `0.9.1`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -946,6 +946,19 @@ repositories:
       url: https://github.com/osrf/capabilities.git
       version: master
     status: maintained
+  carViz:
+    release:
+      packages:
+      - jderobot_carviz
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/JdeRobot/carViz-release.git
+      version: 0.1.5-1
+    source:
+      test_pull_requests: true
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -946,6 +946,19 @@ repositories:
       url: https://github.com/osrf/capabilities.git
       version: master
     status: maintained
+  carViz:
```
clalancette, line 949: This is an unusual name for the key.  I'd suggest making this the same as the package, namely "jderobot_carviz".

clalancette (changes requested):

melodic/distribution.yaml
pankhurivanjani: Solved in a new PR [#25551 ](https://github.com/ros/rosdistro/pull/25551)

clalancette: Please, if you open a new PR, close the old one.  That way we don't waste time looking at outdated stuff.

PR #25174, opened 2020-05-29
dual_quaternions: 0.3.1-1 in 'kinetic/distribution.yaml' [bloom]

Achllle:
Increasing version of package(s) in repository `dual_quaternions` to `0.3.1-1`:

- upstream repository: https://github.com/Achllle/dual_quaternions.git
- release repository: https://github.com/Achllle/dual_quaternions-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.9.7`
- previous version for package: `null`

## dual_quaternions

```
* Fix rst reference
* Initial commit: copy dual_quaternions over from dual_quaternions_ros
* Contributors: Achille
```

PR #25433, opened 2020-06-12
dual_quaternions: switch to kinetic-devel branch 

Achllle:
I must've missed this while running the prerelease test. Got an error that setuptools isn't supported in kinetic (doesn't happen in Noetic) so switched back to `distutils.core` on branch `kinetic-devel`.

I'm not sure if this is the way to do it or if I have to use the bloom-release tool.

Previous PR: #25174

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2722,7 +2722,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/Achllle/dual_quaternions.git
-      version: master
+      version: kinetic-devel
     release:
       tags:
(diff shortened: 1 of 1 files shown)


clalancette:
Holding for Kinetic [sync](https://discourse.ros.org/t/preparing-for-kinetic-sync-2020-06-10/14614).

tfoote:
This will only change the devel and doc jobs that build from source. The debian packages only use the artifacts generated by bloom-release. You'll have to make a new release and rerun bloom to get the debians updated to use that release.

PR #25474, opened 2020-06-15
add rosdep key for geomag

tdenewiler:
Used to find magnetic declination for any given latitude/longitude/altitude. Allows easier calculation of global heading from IMU or compass data in robots distributed around the world. This package is only available via PyPI, not OS package managers.

* PyPI: https://pypi.org/project/geomag/
* Arch: https://www.archlinux.org/packages/?sort=&q=geomag&maintainer=&flagged=
* Debian: https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=geomag
* Fedora: https://apps.fedoraproject.org/packages/s/geomag
* Gentoo: https://packages.gentoo.org/packages/search?q=geomag
* Ubuntu: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=geomag&searchon=names

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -149,7 +149,7 @@ exhale-pip:
   ubuntu:
     pip:
       packages: [exhale]
-geomag:
+geomag-pip:
   debian:
     pip:
       packages: [geomag]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -149,19 +149,6 @@ exhale-pip:
   ubuntu:
     pip:
       packages: [exhale]
-geomag-pip:
-  debian:
-    pip:
-      packages: [geomag]
-  fedora:
-    pip:
-      packages: [geomag]
-  osx:
-    pip:
-      packages: [geomag]
-  ubuntu:
-    pip:
-      packages: [geomag]
 gunicorn:
   debian: [gunicorn]
   fedora: [python-gunicorn]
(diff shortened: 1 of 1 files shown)


clalancette (approved):

rosdep/python.yaml
```diff
@@ -149,6 +149,19 @@ exhale-pip:
   ubuntu:
     pip:
       packages: [exhale]
+geomag:
```
sloretz, line 152: Mind using the key name `geomag-pip`?

sloretz (changes requested):

rosdep/python.yaml
sloretz: Also I see the contributing guide asks for `python-` and `python3-` suffixes. I guess that means this should be `python-geomag-pip` and `python3-geomag-pip`.

 @ros/rosdeputies Why is that the rule? Whether it's Python 2 or Python 3 depends on the version of python set in `ROS_PYTHON_VERSION` or the default rosdistro, what's the value of separate `python-` and `python3-` prefixes for keys only on PyPI?

tdenewiler: c1858da is where I add `-pip` suffix. Let me know what you want for a prefix and I'll add it. I have the same questions you do about the prefix.

sloretz: Opening PR to change recommendation: https://github.com/ros/rosdistro/pull/25508

clalancette: I'm sorry for the back and forth on this, but in #25508 we eventually landed on putting the `python3-` prefix on the front and the `-pip` suffix on the back.  @tdenewiler Would you mind updating one more time to conform to that?  Then I think we're good to go here.

tdenewiler: That was a good discussion in #25508, and this approach makes sense to me. Package management is not easy, I appreciate the thought going into this. I made the requested updates.

clalancette: > That was a good discussion in #25508, and this approach makes sense to me. Package management is not easy, I appreciate the thought going into this. I made the requested updates.

Perfect, thanks for your patience.

clalancette (approved):

claireyywang (approved):

PR #25476, opened 2020-06-15
[migration-tools] Handle repositories with no release entry.

nuclearsandwich:
Earlier versions of foxy which this script was tested with did not have any repositories without release entries. This covers that case by:
* Copying all repositories entries, even those that lack release entries.
* Only attempting to bloom repositories which have release entries.

### DIFF
--- a/migration-tools/migrate-rosdistro.py
+++ b/migration-tools/migrate-rosdistro.py
@@ -80,13 +80,15 @@ new_repositories = []
 repositories_to_retry = []
 for repo_name, repo_data in sorted(source_distribution.repositories.items()):
     if repo_name not in dest_distribution.repositories:
-        new_repositories.append(repo_name)
         dest_repo_data = copy.deepcopy(repo_data)
-        release_tag = dest_repo_data.release_repository.tags['release']
-        release_tag = release_tag.replace(args.source,args.dest)
-        dest_repo_data.release_repository.tags['release'] = release_tag
+        if dest_repo_data.release_repository:
+            new_repositories.append(repo_name)
+            release_tag = dest_repo_data.release_repository.tags['release']
+            release_tag = release_tag.replace(args.source,args.dest)
+            dest_repo_data.release_repository.tags['release'] = release_tag
         dest_distribution.repositories[repo_name] = dest_repo_data
-    elif dest_distribution.repositories[repo_name].release_repository.version is None:
+    elif dest_distribution.repositories[repo_name].release_repository is not None and \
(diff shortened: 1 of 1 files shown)


migration-tools/migrate-rosdistro.py
```diff
@@ -80,13 +80,15 @@ def write_tracks_file(tracks, commit_msg=None):
...
 for repo_name, repo_data in sorted(source_distribution.repositories.items()):
     if repo_name not in dest_distribution.repositories:
-        new_repositories.append(repo_name)
         dest_repo_data = copy.deepcopy(repo_data)
-        release_tag = dest_repo_data.release_repository.tags['release']
-        release_tag = release_tag.replace(args.source,args.dest)
-        dest_repo_data.release_repository.tags['release'] = release_tag
+        if dest_repo_data.release_repository:
+            new_repositories.append(repo_name)
+            release_tag = dest_repo_data.release_repository.tags['release']
```
tfoote, line 86: These tags should be checked for presence before being modified. They might not exist too.

migration-tools/migrate-rosdistro.py
```diff
@@ -80,13 +80,15 @@ def write_tracks_file(tracks, commit_msg=None):
...
+            release_tag = release_tag.replace(args.source,args.dest)
+            dest_repo_data.release_repository.tags['release'] = release_tag
         dest_distribution.repositories[repo_name] = dest_repo_data
-    elif dest_distribution.repositories[repo_name].release_repository.version is None:
+    elif dest_distribution.repositories[repo_name].release_repository is not None and \
```
tfoote, line 90: Why `is not None` here versus just pythonic check on line 84?

tfoote (commented):
two thoughts on this change.

migration-tools/migrate-rosdistro.py
nuclearsandwich: The `release` tag is required according to REP-141 https://www.ros.org/reps/rep-0141.html#distribution-file so if we've got a release repo this should be there. With that being said, if you think this merits a defensive check, I'll add one.

migration-tools/migrate-rosdistro.py
nuclearsandwich: I prefer explicit `is not None` in compound conditionals but I'm easily swayed.

tfoote (approved):
I'd forgotten it's required, I think that's a little overkill as if the version is not there the tag with the version embedded in it can't be evaluated. But that's for another day.

I'm not strongly feeling either way, but just noticed the choice and wanted to make sure it was deliberate.

PR #25482, opened 2020-06-16
[dep] Update python*-rosinstall-generator.

130s:
# Changes
- Asterisk `python-rosinstall-generator` for all Ubuntu.
- Add python3 for specific LTS Ubuntu distros.

# Background of the changes
On Ubuntu 16.04 I got the following upon `rosdep install`:
```
pkg_NOODLE: No definition of [python-rosinstall-generator] for OS version [xenial]
```
Noticed rosdep keys for the pkg in question isn't defined after Ubuntu 14.04. The pkg is available in `apt` world.

```
$ docker run -it ros:kinetic-ros-base bash

:
The following NEW packages will be installed:
  python-rosinstall-generator
:
  Unpacking python-rosinstall-generator (0.1.19-1) ...
  Setting up python-rosinstall-generator (0.1.19-1) ...
  root@6960e681f1c6:/#
```
Disclaimer: I haven't confirmed asterisking all Ubuntu works.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3966,8 +3966,7 @@ python-rosinstall-generator:
   macports: [py27-rosinstall-generator]
   rhel:
     '7': [python2-rosinstall_generator]
-  ubuntu:
-    '*': [python-rosinstall-generator]
+  ubuntu: [python-rosinstall-generator]
 python-rospkg:
   alpine:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6070,7 +6070,7 @@ python3-rosdistro-modules:
   rhel: ['python%{python3_pkgversion}-rosdistro']
   ubuntu: [python3-rosdistro-modules]
 python3-rosinstall-generator:
-  ubuntu: [python3-rospkg]
+  ubuntu: [python3-rosinstall-generator]
 python3-rospkg:
   alpine:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3967,16 +3967,7 @@ python-rosinstall-generator:
   rhel:
     '7': [python2-rosinstall_generator]
   ubuntu:
-    lucid: [python-rosinstall-generator]
-    maverick: [python-rosinstall-generator]
-    natty: [python-rosinstall-generator]
-    oneiric: [python-rosinstall-generator]
-    precise: [python-rosinstall-generator]
-    quantal: [python-rosinstall-generator]
-    raring: [python-rosinstall-generator]
-    saucy: [python-rosinstall-generator]
-    trusty: [python-rosinstall-generator]
-    trusty_python3: [python3-rosinstall-generator]
+    '*': [python-rosinstall-generator]
 python-rospkg:
   alpine:
     pip:
@@ -6079,6 +6070,8 @@ python3-rosdistro-modules:
   openembedded: [python3-rosdistro@meta-ros]
   rhel: ['python%{python3_pkgversion}-rosdistro']
   ubuntu: [python3-rosdistro-modules]
+python3-rosinstall-generator:
+  ubuntu: [python3-rospkg]
 python3-rospkg:
   alpine:
     pip:


rosdep/python.yaml
```diff
@@ -3957,16 +3957,11 @@ python-rosinstall-generator:
...
-    precise: [python-rosinstall-generator]
-    quantal: [python-rosinstall-generator]
-    raring: [python-rosinstall-generator]
-    saucy: [python-rosinstall-generator]
-    trusty: [python-rosinstall-generator]
+    '*': [python-rosinstall-generator]
+    bionic_python3: [python3-rosinstall-generator]
+    focal_python3: [python3-rosinstall-generator]
     trusty_python3: [python3-rosinstall-generator]
+    xenial_python3: [python3-rosinstall-generator]
```
clalancette, line 3964: I'm fine with switching this to use a `*`, but let's not add new `_python3` keys.  They were a failed experiment and we're trying to get rid of them.  If you'd like to add a Python 3 compatible key, please add a new `python3-rosinstall-generator` key instead.

clalancette (changes requested):

rosdep/python.yaml
130s: I see. My bad I also found [CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-rules) is updated for not mixing py3 with py2. Commit updated.

I'm not yet super familiar with Python3 situation esp. over other platforms than Ubuntu, so I've only filled in Ubuntu for Py3. Had I a bit more time I could dig into it, but there could also be a general guidance I assume (I can't find it e.g. in the CONTRIBUTING.md). Looking at a few other py3 entries the package names diverse...some with "3" but some without. If there's already an info that'd be helpful.

rosdep/python.yaml
```diff
@@ -6051,6 +6042,8 @@ python3-rosdistro-modules:
   openembedded: [python3-rosdistro@meta-ros]
   rhel: ['python%{python3_pkgversion}-rosdistro']
   ubuntu: [python3-rosdistro-modules]
+python3-rosinstall-generator:
+  ubuntu: [python3-rospkg]
```
clalancette, line 6046:
```suggestion
python3-rosinstall-generator:
  debian:
    '*': [python3-rosinstall-generator]
    stretch: null
  ubuntu:
    '*': [python3-rosinstall-generator]
    xenial: null
```

clalancette (changes requested):

clalancette (approved):

rosdep/python.yaml
```diff
@@ -3957,16 +3957,7 @@ python-rosinstall-generator:
...
-    lucid: [python-rosinstall-generator]
-    maverick: [python-rosinstall-generator]
-    natty: [python-rosinstall-generator]
-    oneiric: [python-rosinstall-generator]
-    precise: [python-rosinstall-generator]
-    quantal: [python-rosinstall-generator]
-    raring: [python-rosinstall-generator]
-    saucy: [python-rosinstall-generator]
-    trusty: [python-rosinstall-generator]
-    trusty_python3: [python3-rosinstall-generator]
```
nuclearsandwich, line 3969: I think @tfoote wanted to leave these `_python3` keys to avoid breaking people using them. Then again this is a trusty key so maybe we can retire it.

130s: I see. I'm fine in either way, will wait for direction.

Indeed EoL doesn't mean there are no users.

clalancette: Let's just put the `trusty_python3` key back as it was.  All of the others will be subsumed by the wildcard matching.  While I would like to get rid of the old keys, this is the safest thing to do.

sloretz: I think we can remove the EOL `trusty_python3` key: https://github.com/ros/rosdistro/pull/25120#discussion_r432116597

130s: Thanks...So, this PR is now finalized?

> Indeed EoL doesn't mean there are no users.

I might have been confused. I guess my statement above is always true but that's what EoL deals with.

130s: > So, this PR is now finalized?

From an author's perspective it is finalized for now but I guess we're waiting for the discussion in https://github.com/ros/rosdistro/pull/25507 to finalize?

rosdep/python.yaml
```diff
@@ -3957,16 +3957,7 @@ python-rosinstall-generator:
...
+    '*': [python-rosinstall-generator]
```
nuclearsandwich, line 3960: Since the older distributions have all been cleaned up we can simplify this key expression to
```
ubuntu: [python-rosinstall-generator]
```

rosdep/python.yaml
```diff
@@ -6051,6 +6042,13 @@ python3-rosdistro-modules:
   openembedded: [python3-rosdistro@meta-ros]
   rhel: ['python%{python3_pkgversion}-rosdistro']
   ubuntu: [python3-rosdistro-modules]
+python3-rosinstall-generator:
+  debian:
+    '*': [python3-rosinstall-generator]
+    stretch: null
+  ubuntu:
+    '*': [python3-rosinstall-generator]
+    xenial: null
```
nuclearsandwich, line 6051: Is there a reason this key is nulled for Xenial and Stretch? I'm able to install the python3 package for rosinstall_generator on both platforms?

nuclearsandwich (changes requested):
I have one comment around simplification and one question around the stretch and xenial keys.

rosdep/python.yaml
130s: Turned out the nulled entries were the suggestion (that I accepted without reviewing. My bad)  in https://github.com/ros/rosdistro/pull/25482#discussion_r441018846.  @clalancette care for the rationale?

clalancette: As far as I can tell, it isn't available in Xenial: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python3-rosinstall-generator&searchon=names

Nor in Stretch: https://packages.debian.org/search?keywords=python3-rosinstall-generator&searchon=names&suite=all&section=all

tfoote: Those packages are not in the upstream, but we provide instances of them in our repositories. For example: http://packages.ros.org/ros/ubuntu/dists/xenial/main/binary-amd64/Packages

Thus the keys are considered valid for rosdeps.

rosdep/python.yaml
```diff
@@ -6079,6 +6069,8 @@ python3-rosdistro-modules:
   openembedded: [python3-rosdistro@meta-ros]
   rhel: ['python%{python3_pkgversion}-rosdistro']
   ubuntu: [python3-rosdistro-modules]
+python3-rosinstall-generator:
+  ubuntu: [python3-rospkg]
```
tfoote, line 6073: Without the explicit null entries this is still pointing to the wrong packages. `python3-rospkg` is not the same as `python3-rosinstall-generator`

```suggestion
  ubuntu: [python3-rosinstall-generator]
```

tfoote (changes requested):

rosdep/python.yaml
130s: What a terrible mistake. Thank you for catching.

130s (commented):
Suggestion applied, all threads resolved hopefully.

nuclearsandwich (approved):
Thanks for iterating @130s.

tfoote (approved):

PR #25487, opened 2020-06-16
migration-tools: fix minor typos

gavanderhoorn:
Not entirely sure whether these are really typos (but I don't believe one can "make assertions").

### DIFF
--- a/migration-tools/migrate-rosdistro.py
+++ b/migration-tools/migrate-rosdistro.py
@@ -18,7 +18,7 @@ from rosdistro.writer import yaml_from_distribution_file
 
 # These functions are adapted from Bloom's internal 'get_tracks_dict_raw' and
 # 'write_tracks_dict_raw' functions.  We cannot use them directly since they
-# make assertions about the release repository that are not true during the
+# make assumptions about the release repository that are not true during the
 # manipulation of the release repository for this script.
 def read_tracks_file():
     return yaml.safe_load(show('master', 'tracks.yaml'))
@@ -61,7 +61,7 @@ if len(index_yaml['distributions'][args.source]['distribution']) != 1 or \
         len(index_yaml['distributions'][args.dest]['distribution']) != 1:
             raise RuntimeError('Both source and destination distributions must have a single distribution file.')
 
-# There is a possibility that the source_ref has different a distribution file
+# There is a possibility that the source_ref has a different distribution file
 # layout. Check that they match.
 source_ref_index_yaml = yaml.safe_load(show(args.source_ref, 'index-v4.yaml'))
 if source_ref_index_yaml['distributions'][args.source]['distribution'] != \


gavanderhoorn:
@nuclearsandwich

migration-tools/migrate-rosdistro.py
```diff
@@ -18,7 +18,7 @@
 
 # These functions are adapted from Bloom's internal 'get_tracks_dict_raw' and
 # 'write_tracks_dict_raw' functions.  We cannot use them directly since they
-# make assertions about the release repository that are not true during the
+# make assumptions about the release repository that are not true during the
```
clalancette, line 21: For what its worth, I do believe you can "make assertions"; that is, you are asserting that something is true.

That being said, I also believe that the author of this comment probably meant "make assumptions", so I'm good with this change.

clalancette (approved):
Looks good to me, but I'll let @nuclearsandwich review and merge.

migration-tools/migrate-rosdistro.py
gavanderhoorn: > I do believe you can "make assertions"; that is, you are asserting that something is true.

true. People can "make assertions", but this is code, so it sounded weird to me.

clalancette:
I'll just go ahead and merge this since it is just typos in comments.

migration-tools/migrate-rosdistro.py
nuclearsandwich: In this case bloom does indeed *assert* that there is only one remote with [check_for_multiple_remotes](https://github.com/ros-infrastructure/bloom/blob/9e81373b0e9be8d0e154dcc57ec8530b4ddc8ea5/bloom/config.py#L283-L294) and issues a fatal error.

Since the purpose of the comment is to explain and the explanation was insufficient, we should not revert to the previous comment but the merged state of this PR is also incomplete. The code does not make an assumption, it enforces a constraint which produces an unrecoverable error if violated.

In the interest not of advocating for the previous state but of describing how I chose those words: I have in the past described any code that uses an `assert`-style function or generating an unrecoverable error based on a checked condition as code that "makes an assertion" or code that "asserts a condition".

@gavanderhoorn does the following text make the original intent clearer?
```
We cannot use them directly since they require the release repository to have exactly one remote`
```

gavanderhoorn: That's fine by me.

> In this case bloom does indeed *assert* [..]

yes, the code asserts. It does not *make assertions* ;)

At least to me, the first sounds ok (as in: the code makes sure that ..), the second does not sound ok.

But I'm not a native speaker, and we're almost bike-shedding here.

nuclearsandwich: I interpret both phrases as synonymous and over a cold beverage would love to discuss personification of code in comments some time. However my overriding concern is clarity having been shown that one phrasing causes confusion I have no problem adapting.

PR #25488, opened 2020-06-16
Add dynamic-graph to melodic

Rascof:
Release dynamic-graph a third-party package from SOT on melodic

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2095,7 +2095,7 @@ repositories:
       url: https://github.com/Achllle/dual_quaternions_ros-release.git
       version: 0.1.3-1
     status: maintained
-  dynamic_graph:
+  dynamic-graph:
     doc:
       type: git
       url: https://github.com/stack-of-tasks/dynamic-graph.git
@@ -2106,7 +2106,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/stack-of-tasks/dynamic-graph-ros-release.git
-      version: 4.2.1-2
+      version: 4.2.1-4
     source:
       type: git
       url: https://github.com/stack-of-tasks/dynamic-graph.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2095,6 +2095,23 @@ repositories:
       url: https://github.com/Achllle/dual_quaternions_ros-release.git
       version: 0.1.3-1
     status: maintained
+  dynamic-graph-ros-release:
+    doc:
+      type: git
+      url: https://github.com/stack-of-tasks/dynamic-graph.git
+      version: devel
+    release:
+      packages:
+      - dynamic-graph
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/stack-of-tasks/dynamic-graph-ros-release.git
+      version: 4.2.1-2
+    source:
+      type: git
+      url: https://github.com/stack-of-tasks/dynamic-graph.git
+      version: devel
+    status: maintained
   dynamic_reconfigure:
     doc:
       type: git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2101,8 +2101,6 @@ repositories:
       url: https://github.com/stack-of-tasks/dynamic-graph.git
       version: devel
     release:
-      packages:
-      - dynamic-graph
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/stack-of-tasks/dynamic-graph-ros-release.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2095,7 +2095,7 @@ repositories:
       url: https://github.com/Achllle/dual_quaternions_ros-release.git
       version: 0.1.3-1
     status: maintained
-  dynamic-graph-ros-release:
+  dynamic_graph:
     doc:
       type: git
       url: https://github.com/stack-of-tasks/dynamic-graph.git


melodic/distribution.yaml
```diff
@@ -2095,6 +2095,23 @@ repositories:
       url: https://github.com/Achllle/dual_quaternions_ros-release.git
       version: 0.1.3-1
     status: maintained
+  dynamic-graph-ros-release:
```
clalancette, line 2098: Almost all of the other keys use `_` instead `-`; please do the same here.  Also, this should just be `dynamic_graph`; no need for the `ros-release` suffix.

clalancette (changes requested):

melodic/distribution.yaml
Rascof: Thanks for the comment. It is my first release so I am still new to this.
Should i modify the name in packages too ?
`    release:
      packages:
      - dynamic-graph
      tags:`

clalancette: Yes, the package name should also be `dynamic_graph`.  Note that this will require a change to the source package.xml and a re-release to work properly.

Rascof: Ok thank you for your time I will patch this

nim65s: @clalancette : Hi. Do you have any reference for this naming convention, or is this only a remark ?

nim65s: To provide a bit more details and context: I'm one of the maintainers of the dynamic-graph project. This project is more than 10 years old, and is used as a dependency of many other projects, in various research labs, and on multiple robots.

If we have to rename it, I would really like to understand why. 
I will also need some time to think at all the implications and ensure that this will not break anything for anybody.

clalancette: There are a couple things to mention here.

The first is that it is probably not a *hard* requirement to change the name to use underscores.  However, there are currently no other examples of packages in the ROS ecosystem that use dashes, so I have to wonder why this one is different.

Second, while the project may be old enough, it looks like your [package.xml](https://github.com/stack-of-tasks/dynamic-graph/blob/master/package.xml) is fairly new, having been introduced at the end of 2019.  It may be good enough to just change the name in the package.xml.  That would cause the name of your package here to use the underscores, but the rest of the project would still remain the same name.  That being said, I've never tested this configuration, so I don't know if there are any pitfalls you may run into.

nim65s: Thanks for your answer.

It is hard to change the name of the package, as this name is used in many hardcoded places, like [C++ include paths](https://github.com/stack-of-tasks/sot-core/blob/master/include/sot/core/sot.hh#L25) and [package managers configuration](https://github.com/Gepetto/robotpkg-wip/blob/master/dynamic-graph-v3/depend.mk#L18).

I can provide an example of a package in the ROS ecosystem that use a dash: [hpp-fcl](https://github.com/ros/rosdistro/blob/f300c5e590c9bd5ce1634cac1897ab2bc0a76ed1/melodic/distribution.yaml#L3620]). I'm also one of the maintainers of the that upstream project , so if ROS forbids this or if there is any practical issue with that, I'd like to know. However, I just discovered that it is not actually included in its [pinocchio dependency](https://github.com/stack-of-tasks/pinocchio/blob/a179f5619a24000d71d748243b5068605cce4453/package.xml#L31) on ROS side.


You are right, the package.xml file is fairly new, and if you require only a change in that file, that wouldn't change too many things for me. This would still be confusing, so I think it would be better to document that. I would also need to check in which places this name will be used before actually changing that.

For example, after this package, @Rascof will probably want to include [its python bindings](https://github.com/stack-of-tasks/dynamic-graph-python) into the ROS ecosystem. Does that mean that catkin will fail to find cmake config files in paths containing `dynamic_graph`, while they are installed in `dynamic-graph` ?

clalancette: > I can provide an example of a package in the ROS ecosystem that use a dash: [hpp-fcl](https://github.com/ros/rosdistro/blob/f300c5e590c9bd5ce1634cac1897ab2bc0a76ed1/melodic/distribution.yaml#L3620%5D).

Ah, I had missed that in my searches; thanks.  Part of my concern was that our tooling wouldn't work with dashes, but it seems like it will because of that.

In that case, I guess I am OK with this.  I'll go ahead and approve, but I would like someone else from @ros/rosdeputies to take a look before I merge.

clalancette (approved):

melodic/distribution.yaml
Rascof: Hello,
A kindly reminder @clalancette @ros/rosdeputies
I need this package in order to continue working properly on other packages since they depend on this one
Thank you in advance

clalancette (approved):

melodic/distribution.yaml
jacobperron: Perhaps our tooling can handle it, but it is documented as a mandatory rule in [REP 144](https://www.ros.org/reps/rep-0144.html#mandatory-rules) that packages names should only consist of alphanumeric characters and underscores.

At the very least, I'd suggest avoiding dashes (`-`) in ROS package names in the future.

nim65s: Hi @jacobperron,
Thanks for this pointer, I have been looking for such kind of documentation, but this REP refers to ROS package names. dynamic-graph is not a ROS package, but a third party one.

jacobperron: > this REP refers to ROS package names. dynamic-graph is not a ROS package, but a third party one.

I think that's right. The reason for the naming convention is so that code generation tools (e.g. when defining ROS messages) can safely generate valid code in different languages. Since dynamic-graph is not generating any ROS-specific code, I think it should be okay to leave the name as-is.

PR #25489, opened 2020-06-16
Migrate Foxy to Rolling using migrate-rosdistro.py

nuclearsandwich:
Rolling Ridley releases were bloomed using the script invocation:

```
python3 migration-tools/migrate-rosdistro.py --source foxy --source-ref 5dc357a5d2351032b62938d4ab2e10027e61570f --dest rolling --release-org ros2-gbp
```

During this initial run, two repositories failed to bloom and one was left in an incomplete state and so so have no current version in Rolling.

* marti_common due to an Internal Server Error when pushing changes to the ros2-gbp repository on GitHub, this has happened on several retries and requires further investigation. It may just be too large of a push to handle with https.
* fmi_adapter_ros2 due to release_tag being set to :{ask} with no last_release set for Foxy.
* cyclonedds I made a bloom release of cyclonedds 0.6.0 for Foxy which hadn't yet merged and the disparity between the foxy/distribution.yaml and the bloom tracks caused the wrong release to be made. This is easily resolved by a bloom release of cyclonedds directly into rolling with the corrected version and I'll also be adding a check to the script to catch this situation for future migrations.

An additional commit is on this branch updating the source branches for core packages.

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -10,7 +10,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ament/ament_cmake.git
-      version: foxy
+      version: master
     release:
       packages:
       - ament_cmake
@@ -41,13 +41,13 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/ament/ament_cmake.git
-      version: foxy
+      version: master
     status: developed
   ament_cmake_ros:
     doc:
       type: git
       url: https://github.com/ros2/ament_cmake_ros.git
-      version: foxy
+      version: master
     release:
       packages:
       - ament_cmake_ros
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
> Is the fact that all releases have a deb inc of `2` related to the process or was it just that for the first time it took two attempts?

It's an artifact of the process. Although it's more accurate to say that all releases have a debinc of the source distribution debinc + 1. The rolling track was copied from foxy and the same version was released so the debinc in incremented. While our process normally involves resetting the debinc for new rosdistros I think that maintaining the debinc lineage indicates the continuity between the source and destination distributions that wasn't previously there. For example, the debinc 3 of fastrtps released different sources from the debinc 2 of the same version since we were targeting a pre-release commit.

> How did you determine the upstream branch for repos which are not part of the ros2.repos file?

Other repositories retain the source branch from Foxy.

dirk-thomas:
> Other repositories retain the source branch from Foxy.

Sounds reasonable. It would be good to add that to the announcement so that maintainer know that they might have to change the branch name.

dirk-thomas (approved):

tfoote (approved):

PR #25503, opened 2020-06-17
Added rosdep rule to for python package pycpd under Ubuntu.

airballking:
This adds a rosdep rule for the Python package [pycpd](https://pypi.org/project/pycpd/).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3247,6 +3247,10 @@ python-pycodestyle:
   gentoo: [dev-python/pycodestyle]
   rhel: [python2-pycodestyle]
   ubuntu: [pycodestyle]
+python-pycpd:
+  ubuntu:
+    pip:
+      packages: [pycpd]
 python-pycryptodome:
   arch: [python2-pycryptodome]
   debian: [python-pycryptodome]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3247,6 +3247,10 @@ python-pycodestyle:
   gentoo: [dev-python/pycodestyle]
   rhel: [python2-pycodestyle]
   ubuntu: [pycodestyle]
+python-pycpd:
```
clalancette, line 3250: Since this is going to be pip-only key, please suffix it with `-pip`.  Also, while we are here, we may as well add in entries for debian and fedora as well.

clalancette (changes requested):

rosdep/python.yaml
airballking: Thanks for the fast review. Good ideas!

airballking: Addressed improvements suggested by @clalancette

nuclearsandwich:
Just a head's up there is an ongoing discussion about key names for pip-only keys in https://github.com/ros/rosdistro/pull/25508

airballking:
@nuclearsandwich Thanks for the info. I read this as a request to remove the prefix "python-". Correct? Just to avoid misunderstandings. Will gladly do it. :)

rosdep/python.yaml
airballking: @clalancette Are the changes fine? I'd like to use this rosdep rule in one of my projects. ;)

airballking:
Ping.

clalancette (approved):

nuclearsandwich (approved):

airballking:
Thanks!

PR #25505, opened 2020-06-17
disable dev jobs for rosbag2_bag_v2

Karsten1987:
given that dev jobs can't source a ROS1 distro beforehand, it makes sense to disable the dev jobs for the `rosbag2_bag_v2` packages as they rely on `roscpp` to be found.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1735,7 +1735,7 @@ repositories:
       url: https://github.com/ros2-gbp/rosbag2_bag_v2-release.git
       version: 0.0.10-1
     source:
-      test_pull_requests: false
+      test_commits: false
       type: git
       url: https://github.com/ros2/rosbag2_bag_v2.git
       version: foxy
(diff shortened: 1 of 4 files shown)


foxy/distribution.yaml
```diff
@@ -1735,7 +1735,7 @@ repositories:
       url: https://github.com/ros2-gbp/rosbag2_bag_v2-release.git
       version: 0.0.10-1
     source:
-      test_pull_requests: true
+      test_pull_requests: false
```
dirk-thomas, line 1738: This line should be removed since `false` is the default. Additionally it needs a line `test_commits: false` to get rid of the `dev` job.

dirk-thomas (commented):
The same fix is needed for all active distros - including Dashing and Eloquent.

foxy/distribution.yaml
Karsten1987: done in https://github.com/ros/rosdistro/pull/25505/commits/5988f98a416fb833e598401a4019eb79f687955f

jacobperron (approved):

nuclearsandwich (approved):

dirk-thomas (approved):

Karsten1987:
thank you all for your help. 
I don't have permissions to merge.

PR #25507, opened 2020-06-17
Allow removal of *_python3 keys when EOL

sloretz:
Discussion:

https://github.com/ros/rosdistro/pull/25120#discussion_r432116597
https://github.com/ros/rosdistro/pull/25120#discussion_r441699931
https://github.com/ros/rosdistro/pull/25482#discussion_r441103028

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,6 +142,7 @@ python3-foobar:
 
 You may see existing rules that use `_python3`-suffixed distribution codenames.
 These were trialed as a possible style of Python 3 rules and should not be used.
+`_python3`-suffixed keys may be removed when the platform they're targeting reaches end of life.
 The guidance above should be followed for new rules.
 Additionally, if you rely on a dependency that uses `_python3`-suffixed codenames, add a new rule for it that follows the guidance above.
 


130s:
+1 this helps contributors like me, and it aligns with @tfoote's comment https://github.com/ros/rosdistro/pull/25120#discussion_r432116597

nuclearsandwich (approved):

clalancette:
I'll say that I am all for this, but I have gotten complaints from some users when I've attempted to do this.  Their point being that while the distribution may be EOL, the robot they have may still be running the EOL distribution.

This can clearly get crazy (I don't think we should keep `precise_python3` keys around, for instance), but if we take this stance we'll definitely get some disappointed users.

nuclearsandwich:
>  Their point being that while the distribution may be EOL, the robot they have may still be running the EOL distribution.

There may be situations I'm unfamiliar with, but my first recommendation to anyone having rosdep diffficulties with a platform that is no longer supported would be to use rosdep sources which is frozen in a state which meets their needs.

tfoote (approved):
Yeah, if we can provide best practices such as leveraging a snapshot/hashed value or tag then we don't need to worry about keeping far backwards compatibility. It might make sense to have an official <distro>-eol tag which demarcates the end of support, and then we'd prune after that.

PR #25515, opened 2020-06-18
Set doc+source branches for ros2_tracing in Eloquent

christophebedard:
Makes it a bit cleaner/more uniform.

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -2226,7 +2226,7 @@ repositories:
     doc:
       type: git
       url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing.git
-      version: 0.2.10
+      version: eloquent
     release:
       packages:
       - ros2trace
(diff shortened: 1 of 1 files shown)


eloquent/distribution.yaml
```diff
@@ -2226,7 +2226,7 @@ repositories:
     doc:
       type: git
       url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing.git
-      version: 0.2.10
+      version: eloquent
```
clalancette, line 2229: This is fine, but I'll also point out that you should update your release tracks.yaml [here](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing-release/-/blob/master/tracks.yaml#L39) to point to `eloquent` as well.

clalancette (approved):

eloquent/distribution.yaml
christophebedard: oh, right, thanks for pointing that out!

PR #25533, opened 2020-06-19
Add python3-collada

DavidMansolino:
My package (webots_ros2_importe: https://github.com/cyberbotics/webots_ros2/blob/master/webots_ros2_importer/package.xml#L16) currently depends on `python3-collada-pip` but I would prefer to avoid adding pip dependencies.

debian: https://packages.debian.org/sid/main/python3-collada
fedora: https://fedora.pkgs.org/31/fedora-aarch64/python3-collada-0.6-3.fc31.noarch.rpm.html
ubuntu: https://packages.ubuntu.com/focal/python3-collada

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5410,9 +5410,15 @@ python3-click:
   rhel: ['python%{python3_pkgversion}-click']
   ubuntu: [python3-click]
 python3-collada:
-  debian: [python3-collada]
+  debian:
+    '*': [python3-collada]
+    buster: null
+    stretch: null
   fedora: [python3-collada]
-  ubuntu: [python3-collada]
+  ubuntu:
+    '*': [python3-collada]
+    bionic: null
+    xenial: null
 python3-collada-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5409,6 +5409,10 @@ python3-click:
   gentoo: [dev-python/click]
   rhel: ['python%{python3_pkgversion}-click']
   ubuntu: [python3-click]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5409,6 +5409,10 @@ python3-click:
   gentoo: [dev-python/click]
   rhel: ['python%{python3_pkgversion}-click']
   ubuntu: [python3-click]
+python3-collada:
+  debian: [python3-collada]
```
clalancette, line 5413: It looks like this is only available on bullseye and later.  Thus, to maintain compatibility with our older distributions, I suggest the following:

```
debian:
  '*': [python3-collada]
  buster: null
  stretch: null
```

rosdep/python.yaml
```diff
@@ -5409,6 +5409,10 @@ python3-click:
...
+  fedora: [python3-collada]
+  ubuntu: [python3-collada]
```
clalancette, line 5415: It looks like this is only available on focal and later.  Thus, to maintain compatibility with our older distributions, I suggest the following:

```
ubuntu:
  '*': [python3-collada]
  bionic: null
  xenial: null
```

clalancette (changes requested):

DavidMansolino:
Thank you @clalancette for your pertinent comments, it should be fixed now.

clalancette (approved):

DavidMansolino:
Any news about the merge of this PR?

omichel:
@clalancette: when can we expect to see this PR merged?

nuclearsandwich (approved):

PR #25544, opened 2020-06-20
kinetic: migrate abb_driver to new repository

gavanderhoorn:
As per subject.

This PR combines a Bloom release of `ros-industrial/abb_driver` and the removal of the same packages from the `ros-industrial/abb` repository in a single commit.

The `release` entry was created manually, as Bloom doesn't let me release a package which is already part of a releases repository (and rightfully so).

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -14,7 +14,6 @@ repositories:
     release:
       packages:
       - abb
-      - abb_driver
       - abb_irb2400_moveit_config
       - abb_irb2400_moveit_plugins
       - abb_irb2400_support
@@ -38,6 +37,11 @@ repositories:
       type: git
       url: https://github.com/ros-industrial/abb_driver.git
       version: kinetic-devel
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/ros-industrial-release/abb_driver-release.git
+      version: 1.4.0-1
     source:
       type: git
       url: https://github.com/ros-industrial/abb_driver.git


gavanderhoorn:
Related: #25545.

I'm not entirely sure whether I have to release `abb_driver` from the new repository before submitting this PR (I expect builds of `ros-industrial/abb` on the buildfarm to fail without `abb_driver` being part of the release)?

tfoote:
Removing this looks like it will break a lot of other things. Is there more context to this? Another thread or something you can link to? It sounds like you're about to replace it with something. Will that something be backwards compatible? Then you could release the new version and then extend that PR to just remove this package and make it an atomic swap of one simultaneous merge of the add and removal. 

There may be some corner cases that bloom doesn't like but I don't think it's doing duplication checks at the moment. 

@clalancette FYI re parallel melodic PR #25545

gavanderhoorn:
> Removing this looks like it will break a lot of other things.

It's not being removed, it's being migrated.

I'm removing it here as it's no longer part of the release of this specific repository.

And the breakage is pretty minimal I believe. There aren't that many *released* packages which depend on this.

> Is there more context to this? Another thread or something you can link to? 

If you're interested, see ros-industrial/abb#179.

> It sounds like you're about to replace it with something. 

No, that's not what is happening here.

> Will that something be backwards compatible?

It's the exact same package, but in a different location.

> Then you could release the new version and then extend that PR to just remove this package and make it an atomic swap of one simultaneous merge of the add and removal.

I thought of that, but wanted some input on whether that would be a good idea.

gavanderhoorn:
Just tried to release `abb_driver` from the new repository without first having this PR merged.

Bloom doesn't like that and gives me this error:

```
Failed to open pull request: AssertionError - Duplicate package name 'abb_driver' exists in repository 'abb_driver' as well as in repository 'abb'
```

which is of course true.

gavanderhoorn:
@tfoote: I've updated this PR.

I believe this does what you suggested.

I'll update #25545 in the same way.

clalancette (approved):

tfoote (approved):
Thanks for the additional context this should work. (Note that there's at least half a dozen downstream binaries which is why I asked http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__abb_driver__ubuntu_xenial_amd64__binary/)

gavanderhoorn:
Seems the build [fails](http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__abb_driver__ubuntu_xenial_amd64__binary/13/console#console-section-3), as it can't build the Docker image.

Seeing the errors I'm not entirely sure why this happens btw:

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 ros-kinetic-industrial-robot-client : Depends: ros-kinetic-actionlib but it is not going to be installed
                                       Depends: ros-kinetic-industrial-utils but it is not going to be installed
                                       Depends: ros-kinetic-robot-state-publisher but it is not going to be installed
                                       Depends: ros-kinetic-urdf but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

Is this something transient?

gavanderhoorn:
Seeing similar errors in other builds, mostly those using the `ros-testing` repository.

ros-discourse:

https://discourse.ros.org/t/migration-of-abb-driver-to-a-separate-repository/13522/3

PR #25545, opened 2020-06-20
melodic: migrate abb_driver to new repository

gavanderhoorn:
As per subject.

This PR combines a Bloom release of `ros-industrial/abb_driver` and the removal of the same packages from the `ros-industrial/abb` repository in a single commit.

The `release` entry was created manually, as Bloom doesn't let me release a package which is already part of a releases repository (and rightfully so).

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -12,7 +12,6 @@ repositories:
     release:
       packages:
       - abb
-      - abb_driver
       - abb_irb2400_moveit_config
       - abb_irb2400_moveit_plugins
       - abb_irb2400_support
@@ -32,6 +31,11 @@ repositories:
       type: git
       url: https://github.com/ros-industrial/abb_driver.git
       version: kinetic-devel
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/ros-industrial-release/abb_driver-release.git
+      version: 1.4.0-1
     source:
       type: git
       url: https://github.com/ros-industrial/abb_driver.git


gavanderhoorn:
Related: #25544.

melodic/distribution.yaml
```diff
@@ -32,6 +31,11 @@ repositories:
       type: git
       url: https://github.com/ros-industrial/abb_driver.git
       version: kinetic-devel
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
```
gavanderhoorn, line 36: Releasing from the Kinetic release branch as there is no difference between the Kinetic and Melodic versions.

clalancette: Hm, I don't know if that actually works.  The reason being that the "overlays" that we use in the release repository also contain the name of the Ubuntu/Debian distribution that they are being released on, and that is different between Kinetic and Melodic.  Certainly I haven't seen this use before; @nuclearsandwich , will this work?

clalancette (changes requested):

melodic/distribution.yaml
gavanderhoorn: I can create a `melodic` release branch. It'll just be a copy really.

gavanderhoorn: Done.

nuclearsandwich:
> I don't know if that actually works. The reason being that the "overlays" that we use in the release repository also contain the name of the Ubuntu/Debian distribution that they are being released on, and that is different between Kinetic and Melodic. Certainly I haven't seen this use before; @nuclearsandwich , will this work?

Short answer: I think it will work for rosinstall generator but cause us problems when building binarydeb packages. It looks like @gavanderhoorn has already updated the PR but for completeness here's why:

The platform- and distribution-specific information is usually found in the debian/* and rpm/* branches. The release/ branches are meant to be platform independent but still give a chance for release patches against upstream source. It's possible to add other entries to tags here: for example one specific to `debian` packages that points to another location, but I'm not current on whether those possible tags are actually used by current infrastructure or if they're determined by heuristic from the release tag no matter what. Bloom resolves package dependencies from the template at bloom-time so trying to use the kinetic debian metadata in melodic would result in trying (and failing) to depend on kinetic packages (e.g. ros-kinetic-catkin) rather than melodic ones. Similarly the xenial rosdep resolutions would be used for system packages.

tfoote (approved):
lgtm with the tags updated now.

@nuclearsandwich is right that it would break the debian builds having the wrong tags. It's the most common error when people manually create PRs instead of using the bloom ones.

gavanderhoorn:
> @nuclearsandwich is right that it would break the debian builds having the wrong tags. It's the most common error when people manually create PRs instead of using the bloom ones.

I'm (and was) fully aware of this. It was simply a case of getting the release branch in the repository mixed up with the release branch in the release repository.

clalancette (approved):

ros-discourse:

https://discourse.ros.org/t/migration-of-abb-driver-to-a-separate-repository/13522/3

PR #25554, opened 2020-06-22

DanielWang123321:
I'd like my repo to be indexed and documented on ros.org.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -3569,6 +3569,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/andreasgustavsson/find_moving_objects.git
+      version: kinetic-devel
     source:
       type: git
       url: https://github.com/andreasgustavsson/find_moving_objects.git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -3565,7 +3565,6 @@ repositories:
     doc:
       type: git
       url: https://github.com/andreasgustavsson/find_moving_objects.git
-      version: kinetic-devel
```
clalancette, line 3568: Please don't remove entries from other packages.

kinetic/distribution.yaml
```diff
@@ -17513,6 +17512,16 @@ repositories:
...
       version: kinetic-devel
     status: developed
+  xarm:
+    doc:
+      type: git
+      url: https://github.com/xArm-Developer/xarm_ros.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/xArm-Developer/xarm_ros.git
```
clalancette, line 17522: At least some of the packages in the source repository here have TODO listed as the license in the package.xml.  Please fix those up so that all of the packages declare their licenses.

clalancette (changes requested):

kinetic/distribution.yaml
DanielWang123321: noted, thanks

kinetic/distribution.yaml
DanielWang123321: Noted, license upgraded. Thanks.

clalancette (approved):

PR #25556, opened 2020-06-22
Update python.yaml add pyswarms

shonigmann:
added rosdep python.yaml entry for pyswarms package 
https://pypi.org/project/pyswarms/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5932,6 +5932,12 @@ python3-pyqt5.qtwebengine:
   fedora: [python3-qt5-webengine]
   ubuntu: [python3-pyqt5.qtwebengine]
 python3-pyswarms-pip:
+  debian:
+    pip:
+      packages: [pyswarms]
+  fedora:
+    pip:
+      packages: [pyswarms]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5931,6 +5931,10 @@ python3-pyqt5.qtwebengine:
   debian: [python3-pyqt5.qtwebengine]
   fedora: [python3-qt5-webengine]
   ubuntu: [python3-pyqt5.qtwebengine]
+python3-pyswarms-pip:
+  ubuntu:
```
clalancette, line 5935: Can you please add equivalent pip rules for debian and fedora?  Thanks.

clalancette (changes requested):

rosdep/python.yaml
shonigmann: I'm quite new to this, so hopefully I'm right in assuming the rules are exactly the same debian and fedora... if so, done!

clalancette (approved):

nuclearsandwich (approved):

PR #25596, opened 2020-06-25
Add python rosdep for grpcio, grpcio-reflection and parameterized

f-fl0:
- add `python-grpcio` entry for `ubuntu:focal` (https://pypi.org/project/grpcio/)
- add `python-grpcio-reflection-pip` for `ubuntu` and `debian` (https://pypi.org/project/grpcio-reflection/)
- add `python-parameterized` entry for `ubuntu:focal` (https://pypi.org/project/parameterized/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1966,8 +1966,6 @@ python-grpcio:
     '*': [python-grpcio]
     bionic:
       pip: [grpcio]
-    focal:
-      pip: [grpcio]
     trusty:
       pip: [grpcio]
     xenial:
@@ -2897,9 +2895,6 @@ python-parameterized:
   ubuntu:
     '*': [python-parameterized]
     artful:
-      pip:
-        packages: [parameterized]
-    focal:
       pip:
         packages: [parameterized]
     trusty:
@@ -5776,6 +5771,11 @@ python3-pandas:
   fedora: [python3-pandas]
   gentoo: [dev-python/pandas]
   ubuntu: [python3-pandas]
+python3-parameterized:
+  alpine: [py3-parameterized]
+  debian: [python3-parameterized]
+  fedora: [python3-parameterized]
+  ubuntu: [python3-parameterized]
 python3-paramiko:
   debian: [python3-paramiko]
   fedora: [python3-paramiko]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1966,10 +1966,19 @@ python-grpcio:
     '*': [python-grpcio]
     bionic:
       pip: [grpcio]
+    focal:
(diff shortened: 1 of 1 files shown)


clalancette (approved):

rosdep/python.yaml
```diff
@@ -2890,6 +2899,9 @@ python-parameterized:
     artful:
       pip:
         packages: [parameterized]
+    focal:
+      pip:
+        packages: [parameterized]
```
nuclearsandwich, line 2904: According to https://packages.ubuntu.com/focal/python3-parameterized version 0.7.0 is packaged in Ubuntu Focal. Would now be a good time to add a `python3-parameterized` key which uses the system dependency instead of relying on pip keys?

rosdep/python.yaml
```diff
@@ -1966,10 +1966,19 @@ python-grpcio:
     '*': [python-grpcio]
     bionic:
       pip: [grpcio]
+    focal:
+      pip: [grpcio]
```
nuclearsandwich, line 1970: Since the default ROS installation on Ubuntu Focal targets python3. I'm wondering if we should create a new key, `python3-grpcio-pip`. It also appears that this package is now available directly from the Ubuntu repositories for Focal https://packages.ubuntu.com/focal/python3-grpcio. Would it make sense to switch away from the pip key entirely?

rosdep/python.yaml
f-fl0: Do you suggest adding an entry like this instead?
```
python3-parameterized:
  ubuntu:
    focal: [python3-parameterized]
```

rosdep/python.yaml
f-fl0: I just realized there is already an entry for `python3-grpcio` https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L5574. I can add `focal` there if it is better.

clalancette: > I just realized there is already an entry for `python3-grpcio` https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L5574. I can add `focal` there if it is better.

That shouldn't be necessary; the wildcard syntax there already covers focal.  So in point of fact, this part of the change shouldn't be necessary at all.

rosdep/python.yaml
tfoote: yes, we should use the debian packaged version if it exists. pip can be used to backfill necessary platforms where the debian package doesn't exist.

tfoote (changes requested):

nuclearsandwich (approved):

clalancette (approved):

PR #25615, opened 2020-06-27
Pip installed scipy for  rosdep

bjin2364:
PyPI version of package are far ahead of the ubuntu/debian apt repository version of `scipy`. Adding a pip installed version for people who need the latest version. Also fixed the existing `scipy` tags to follow the naming convention outlined in https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

- `PyPI`: https://pypi.org/project/scipy/
- `ubuntu`: https://packages.ubuntu.com/search?keywords=python3-scipy&searchon=names

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6108,11 +6108,15 @@ python3-scipy:
   gentoo: [sci-libs/scipy]
   macports: [py37-scipy]
   opensuse: [python3-scipy]
+  ubuntu: [python3-scipy]
+python3-scipy-pip:
   osx:
     pip:
       depends: [gfortran]
       packages: [scipy]
-  ubuntu: [python3-scipy]
+  ubuntu:
+    pip:
+      packages: [scipy]
 python3-scp:
   debian: [python3-scp]
   fedora: [python3-scp]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6108,12 +6108,12 @@ python3-scipy:
   gentoo: [sci-libs/scipy]
   macports: [py37-scipy]
   opensuse: [python3-scipy]
-  ubuntu: [python3-scipy]
-python3-scipy-pip:
   osx:
     pip:
       depends: [gfortran]
       packages: [scipy]
+  ubuntu: [python3-scipy]
+python3-scipy-pip:
   ubuntu:
     pip:
       packages: [scipy]


clalancette (approved):

rosdep/python.yaml
```diff
@@ -6108,11 +6108,15 @@ python3-scipy:
   gentoo: [sci-libs/scipy]
   macports: [py37-scipy]
   opensuse: [python3-scipy]
+  ubuntu: [python3-scipy]
+python3-scipy-pip:
```
nuclearsandwich, line 6112: >  Also fixed the existing `scipy` tags to follow the naming convention outlined in [/CONTRIBUTING.md@`master`](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md?rgh-link-date=2020-06-27T16%3A05%3A26Z)

I appreciate the effort to bring the rosdep database into conformity with our current contributing guidelines. To avoid breaking keys that may be actively in use we tend not to correct keys which don't conform until they are used only by platforms that are no longer supported.

In short, I think that adding the updated key is reasonable but I think we ought to preserve the existing osx via pip definition under the `python3-scipy` key since it could still be in use.

nuclearsandwich (changes requested):
We have been starting to push back against adding / propagating pip keys for packages which are available from system package managers. Officially released ROS packages cannot rely on pip keys. Our recommendation if your own packages require newer dependencies sourced from pip is to add custom rosdep sources covering what you need.

@clalancette since you've already given a :heavy_check_mark: here do you think there's a reasonable override to what I've said above?

rosdep/python.yaml
bjin2364: no problem. I can revert that part

tfoote (changes requested):
We should not have an overriding pip rule that will conflict with the system managed version. There are cases where we might consider exceptions. However this core of a package is sure to significantly side effect other components that are using or relying on scipy. 

Unfortunately being part of a distro is about working with the versions of things that have been vetted and QA'd. It may work for your use case, but you don't know that it will necessarily work for every other use case that's been verified by the maintainers against the distro version and not against the latest pip version. rosdep rules change the entire system and will likely break anyone relying on 0.17 API (on xenial) by installing 1.5 in front of that on you python path.

bjin2364:
> rosdep rules change the entire system and will likely break anyone relying on 0.17 API (on xenial) by installing 1.5 in front of that on you python path.

Maybe there's something I'm missing here, but how would this change have this effect? Existing `package.xml` that depend on `python3-scipy` will still get the 0.17 API (if on xenial) or whatever the appropriate version is from the apt repository, etc. Only a `package.xml` with a tag `python3-scipy-pip` (which doesn't even exist) would get the latest `scipy`

tfoote:
> Only a package.xml with a tag python3-scipy-pip (which doesn't even exist) would get the latest scipy 

That's not correct, system dependencies are not scoped.

If I have a package that depends on python3-scipy it will work fine on it's own. But if I or anyone else installs a package that depends on python3-scipy-pip the pip version will override scipy for all users and all packages on the system. The package that depends on python3-scipy will assert that it's dependencies are met appropriately as it will make sure that the debian package is available, but it will get the pip version.

nuclearsandwich:
> I feel it would be more powerful to give developers the choice to create packages that use a newer version of `scipy`/other python packages (but as a consequence, let them figure out which packages are compatible with each other).

This is still doable with custom local rosdep sources. Developers who are willing to take on the challenges of mixing-and-matching dependencies from different sources are not prevented from doing so. I think our goal is for the default rosdep sources to produce sound systems in as many cases as possible. We are not removing any options here but instead providing stable defaults.

bjin2364:
Got it, thanks for the explanation. I'll close this request then.

PR #25619, opened 2020-06-29
Set package name for python-shapely on Focal to python3-shapely

gbiggs:


rosdep/python.yaml
```diff
@@ -4282,6 +4282,7 @@ python-shapely:
       packages: [shapely]
   ubuntu:
     '*': [python-shapely]
+    focal: [python3-shapely]
```
clalancette, line 4285: I don't think we actually need this change.  We already have a `python3-shapely` key, and that is what should be used for newer Python 3 packages.  This key is implicitly Python 2.

clalancette (changes requested):

rosdep/python.yaml
gbiggs: I completely missed that. In that case, the correct solution would be to depend on `python3-shapely` in our package, correct?

clalancette: > I completely missed that. In that case, the correct solution would be to depend on `python3-shapely` in our package, correct?

Yeah, exactly.  If you need to support both Python 2 and Python 3 (either because you want to support Melodic and Noetic on the same branch, or Melodic and ROS 2), you can use the `condition` attribute like in https://github.com/ros/kdl_parser/blob/d3d4064ebed784f1c9a3a8d4f30e1ebebf6a6400/kdl_parser_py/package.xml#L27

gbiggs: In our case it's all ROS 2 so no problem there, but I didn't know about the condition attribute so thanks for the heads up!

gbiggs:
Closing as not relevant. The correct solution is to depend on `python3-shapely`.

PR #25620, opened 2020-06-29
Adding some libitnition deps for bionic

marcoag:
Dependencies needed for the release of RMF [traffic_edtor](https://github.com/osrf/traffic_editor) in ubuntu bionic


nuclearsandwich:
Since these packages aren't available via the Ubuntu repositories and are not imported into the ROS repositories this change should be made in a local rosdep database if needed.

marcoag:
Sounds reasonable. Thanks for reviewing.

PR #25636, opened 2020-06-30
Move ros2/kdl_parser -> ros/kdl_parser

sloretz:
`ros2/kdl_parser` is being recombined with `ros/kdl_parser`.

### DIFF
--- a/ardent/distribution.yaml
+++ b/ardent/distribution.yaml
@@ -344,7 +344,7 @@ repositories:
       version: 2.0.0-0
     source:
       type: git
-      url: https://github.com/ros2/kdl_parser.git
+      url: https://github.com/ros/kdl_parser.git
       version: ardent
     status: maintained
   launch:
--- a/bouncy/distribution.yaml
+++ b/bouncy/distribution.yaml
@@ -542,7 +542,7 @@ repositories:
     source:
       test_pull_requests: true
       type: git
-      url: https://github.com/ros2/kdl_parser.git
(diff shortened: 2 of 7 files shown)


sloretz:
The single test failure is expected because this changes the repo URL in EOL distros

```
AssertionError: ["In file '''/home/runner/work/rosdistro/rosdistro/crystal/distribution.yaml''': There is a change to a doc section of an EOLed distribution. Lines: [735]"] is not false
```

clalancette (approved):
> The single test failure is expected because this changes the repo URL in EOL distros

I think we've typically left the EOL distros alone, but I don't feel strongly about it one way or the other.  I'm going to approve, but I'll suggest getting one other approval before merging.

jacobperron (commented):
I'm not sure about modifying the EOL distros, maybe someone else can weigh in.

Can you also open PRs to update the repos file in ros2/ros2 for the active distros?

tfoote (approved):
It's fine to update references to moved resources in EOL as long as we expect the content to be the same. The EOL distro can still be useful for people potentially building from source etc. And we would break them if we refused to update the urls for the moved resources.

We don't want to have "new" releases or other changes. but fixing external references is fine. Since we don't necessarily host everything we can't guarantee the urls will resolve forever. But hopefully there will remain copies of the open source projects.

nuclearsandwich (approved):

mjcarroll (approved):

PR #25641, opened 2020-07-01
Add zmq for Ubuntu 20.04

5i0770:
Add zmq entry for Ubuntu 20.04 (focal)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5272,6 +5272,7 @@ python-zmq:
   gentoo: [dev-python/pyzmq]
   ubuntu:
     '*': [python-zmq]
+    focal: [python3-zmq]
     trusty_python3: [python3-zmq]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5272,6 +5272,7 @@ python-zmq:
   gentoo: [dev-python/pyzmq]
   ubuntu:
     '*': [python-zmq]
+    focal: [python3-zmq]
```
clalancette, line 5275: This is implicitly the Python 2 version of the package, so we shouldn't add focal here.  Instead, please add a new key called `python3-zmq`, and add the entry there.

clalancette (changes requested):

tfoote:
I'm going to close this and the new entry can be submitted to add the python3 entry.

PR #25689, opened 2020-07-07
Add connexion to python.yaml

mbeutelspacher:
I'd like to add a key for a pip package named connexion (https://connexion.readthedocs.io/en/latest/index.html).

Connexion is a framework on top of Flask that automagically handles HTTP requests defined using OpenAPI (formerly known as Swagger), supporting both v2.0 and v3.0 of the specification.

It's nice to write REST Interfaces for ROS applications.

``nostests`` runs successfully.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5437,6 +5437,16 @@ python3-colorama:
   gentoo: [dev-python/colorama]
   rhel: ['python%{python3_pkgversion}-colorama']
   ubuntu: [python3-colorama]
+python3-connexion:
+  debian:
+    pip:
+      packages: [connexion]
+  fedora:
+    pip:
+      packages: [connexion]
+  ubuntu:
+    pip:
+      packages: [connexion]
 python3-coverage:
   debian: [python3-coverage]
   fedora: [python3-coverage]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5437,6 +5437,16 @@ python3-colorama:
   gentoo: [dev-python/colorama]
   rhel: ['python%{python3_pkgversion}-colorama']
   ubuntu: [python3-colorama]
+python3-connexion:
```
clalancette, line 5440: Since this is a pip-only key, please add the `-pip` suffix to the key, as per [CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1).

clalancette (changes requested):

rosdep/python.yaml
ipa-mtb: Sorry to have missed that

clalancette:
For reference, https://pypi.org/project/connexion/ is the link to PyPI.

clalancette (approved):

tfoote:
One note is that there's an intent to package ticket here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958042 but it's not in debian yet.

PR #25718, opened 2020-07-08
Add libpqxx rosdeps for focal

nickswalker:
Key listing on Ubuntu packages: https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=libpqxx


rosdep/base.yaml
```diff
@@ -3401,6 +3402,7 @@ libpqxx-dev:
   ubuntu:
     artful: [libpqxx-dev]
     bionic: [libpqxx-dev]
+    focal: [libpqxx-dev]
```
clalancette, line 3405: It looks like this has been `libpqxx-dev` all the way back to trusty.  With that in mind, I'll suggest that for future compatibility we change this entire block to:

```
ubuntu: [libpqxx-dev]
```

clalancette (changes requested):

tfoote (approved):

clalancette (approved):

PR #25736, opened 2020-07-09
Add lcm to rosdep rules

kmhallen:
Available in Ubuntu Bionic and newer
- https://packages.debian.org/buster/liblcm-dev
- https://packages.debian.org/buster/liblcm1
- https://fedora.pkgs.org/31/fedora-x86_64/lcm-devel-1.3.1-13.fc31.i686.rpm.html
- https://fedora.pkgs.org/31/fedora-x86_64/lcm-1.3.1-13.fc31.x86_64.rpm.html
- https://packages.ubuntu.com/bionic/libs/liblcm-dev
- https://packages.ubuntu.com/bionic/libs/liblcm1
- https://packages.ubuntu.com/focal/libs/liblcm-dev
- https://packages.ubuntu.com/focal/libs/liblcm1

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2806,6 +2806,18 @@ liblapack-dev:
 liblapacke-dev:
   debian: [liblapacke-dev]
   ubuntu: [liblapacke-dev]
+liblcm:
+  debian: [liblcm1]
+  fedora: [liblcm]
+  ubuntu:
+    '*': [liblcm1]
+    'xenial': null
+liblcm-dev:
+  debian: [liblcm-dev]
+  fedora: [liblcm-devel]
+  ubuntu:
+    '*': [liblcm-dev]
+    'xenial': null
 libleptonica-dev:
   arch: [leptonica]
   debian: [libleptonica-dev]


rosdep/base.yaml
```diff
@@ -2806,6 +2806,14 @@ liblapack-dev:
 liblapacke-dev:
   debian: [liblapacke-dev]
   ubuntu: [liblapacke-dev]
+liblcm-dev:
+  debian: [liblcm-dev]
+  fedora: [liblcm-devel]
+  ubuntu: [liblcm-dev]
```
clalancette, line 2812: Since we still support Kinetic on Xenial, and it looks like these packages aren't available there, we need to change this to:

```
ubuntu:
  '*': [liblcm-dev]
  xenial: null
```

The same goes for the key below.

rosdep/base.yaml
```diff
@@ -2806,6 +2806,14 @@ liblapack-dev:
...
+liblcm1:
```
clalancette, line 2813: I'd be more inclined to make this key: `liblcm`.  If it ever gets a newer version in the future, it will be more forwards-compatible that way.

clalancette (changes requested):

rosdep/base.yaml
nuclearsandwich: I think it's somewhat dependent on whether the expectation is that future releases contain liblcm2 and not liblcm1. Since the Fedora package names don't include the version I think this is fine as is although my preference would be to follow the Debian package naems for consistency with other rosdep keys.

kmhallen: I'm fine with it either way

tfoote (approved):

clalancette (approved):

PR #25738, opened 2020-07-09
fix Eloquent/Foxy/Rolling branch of rqt_publisher

dirk-thomas:
Closes ros2-gbp/rqt_publisher-release#1.

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -2709,7 +2709,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-visualization/rqt_publisher.git
-      version: crystal-devel
+      version: dashing-devel
     release:
       tags:
         release: release/eloquent/{package}/{version}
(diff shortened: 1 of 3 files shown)


eloquent/distribution.yaml
```diff
@@ -2709,7 +2709,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-visualization/rqt_publisher.git
-      version: crystal-devel
+      version: dashing-devel
```
clalancette, line 2712: So this seems to me that this would be right, but I'll point out that all branches (including master) of https://github.com/ros2/ros2/blob/0ee1ca223bcc7185743da1ecccc754b774cd0958/ros2.repos still point to `crystal-devel`.  So I'm honestly not sure which branch we released from.

dirk-thomas: Only the `dashing-devel` branch contains the version 1.1.0.

Same fix for Foxy and Rolling: d93152d3b05bccda7e59e6b7e8fd927215a64bc1.

The `ros2.repos` files will need to be patched too.

clalancette (approved):
Ah, and now I see that Dashing was the only distro to point to the new branch properly.  This looks good to me (I'll review the changes to ros2.repos next).

PR #25757, opened 2020-07-10
Added boston dynamic sdks

dniewinski:
https://pypi.org/project/bosdyn-api/
https://pypi.org/project/bosdyn-mission/
https://pypi.org/project/bosdyn-core/
https://pypi.org/project/bosdyn-client/

These are needed for a ROS driver for the Spot robot from Boston Dynamics

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5369,6 +5369,18 @@ python3-bokeh-pip:
   ubuntu:
     pip:
       packages: [bokeh]
+python3-bosdyn-api:
+  debian: [python3-bosdyn-api]
+  ubuntu: [python3-bosdyn-api]
+python3-bosdyn-client:
+  debian: [python3-bosdyn-client]
+  ubuntu: [python3-bosdyn-client]
+python3-bosdyn-core:
+  debian: [python3-bosdyn-core]
+  ubuntu: [python3-bosdyn-core]
+python3-bosdyn-mission:
+  debian: [python3-bosdyn-mission]
+  ubuntu: [python3-bosdyn-mission]
 python3-boto3:
   debian: [python3-boto3]
   fedora: [python3-boto3]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -71,6 +71,34 @@ azure-iothub-device-client-pip:
   ubuntu:
     pip:
       packages: [azure-iothub-device-client]
+bosdyn-api:
+  debian:
+    pip:
+      packages: [bosdyn-api]
+  ubuntu:
+    pip:
+      packages: [bosdyn-api]
+bosdyn-client:
+  debian:
+    pip:
+      packages: [bosdyn-client]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@dniewinski can you please add links to these packages on pypi to your PR description as requested in our [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions).

tfoote (changes requested):
Please update the PR to include a description. Since these are pip rules they need to be prefixed with `python3-` and postfixed with `-pip` Please review the contributing guidelines: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions

rosdep/python.yaml
```diff
@@ -5341,6 +5369,18 @@ python3-bokeh-pip:
   ubuntu:
     pip:
       packages: [bokeh]
+python3-bosdyn-api:
+  debian: [python3-bosdyn-api]
+  ubuntu: [python3-bosdyn-api]
```
sloretz, line 5374: This key says there is a debian package named `python3-bosdyn-mission`, but I wasn't able to find one. If these are only available on PyPI then I think what @tfoote is asking for is an end result like:

```yaml
python3-bosdyn-api-pip:
  debian:
    pip:
      packages: [bosdyn-api]
  ubuntu:
    pip:
      packages: [bosdyn-api]
```

That is, please replace the `python3-bosdyn-*` keys and `bosdyn-*` keys with `python3-bosdyn-*-pip` keys using the content from the `bosdyn-*` keys.

sloretz (changes requested):

clalancette (approved):

tfoote (approved):

sloretz (approved):

dniewinski:
Hi @sloretz .  Thanks for accepting the PR.  I tried it today and it installed with pip2.  I need it to install with pip3.  Isn't what why it is prefixed with "python3"?

sloretz:
@dniewinski Which ROS distro? Are you trying to use Python 3 in a ROS distro that only supports Python 2?

> I tried it today and it installed with pip2. I need it to install with pip3. Isn't what why it is prefixed with "python3"

Pip keys can be installed with either Python 3 or Python 2. Which one is used depends on the environment variable `ROS_PYTHON_VERSION`, which is set for each distro. See [this section in REP 151](https://www.ros.org/reps/rep-0151.html#ros-python-version).

The `python3` prefix is just in case upstream Debian or Ubuntu adds these as apt packages. If they do, they'll likely name them `python3-bosdyn` etc. It's easier for rosdep key reviewers to catch that pip keys need to be replaced with upstream ones because they'll be next to each other alphabetically.

dniewinski:
I'm working with melodic.  The package works with python3.  I guess i assumed the prefix just took care of it

PR #25780, opened 2020-07-12
disable PR testing

mikeferguson:
supporting rolling and dashing isn't gonna
happen on same branch - not planning additional
releases into dashing

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -3618,7 +3618,6 @@ repositories:
       url: https://github.com/ros2-gbp/urg_c-release.git
       version: 1.0.4000-1
     source:
-      test_pull_requests: true
       type: git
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

nuclearsandwich:
> supporting rolling and dashing isn't gonna
> happen on same branch - not planning additional
> releases into dashing

I've approved this since I don't want to block changes on your ros2-devel branch however I think it also makes sense to create a branch for Dashing and update the branch entries here so that people who use rosinstall_generator with the --upstream-development flag don't get a branch intended for Rolling Ridley. Whether you want PR testing on that branch is entirely up to you.

dirk-thomas:
@mikeferguson You might want to add `test_commits: false` then to also disable the `dev` jobs which otherwise will send email to the maintainer for unstable / failing builds.

PR #24934, opened 2020-05-16
rosdep: add Ignition Citadel

chapulina:
Some of these keys are already being used by the [ros_ign](https://github.com/osrf/ros_ign/) packages to bloom it into https://packages.osrfoundation.org:

https://github.com/osrf/osrf-rosdep/blob/master/ignition/ignition.yaml

---

This is a follow up to #24646

PR #25789, opened 2020-07-13
Update Ignition rosdep keys

chapulina:
Originally on #24934 we added rosdep keys that matched the debian package names as recommended by the reviewers: https://github.com/ros/rosdistro/pull/24934#discussion_r427444961.

However, this makes usage of these keys complicated when compiling Ignition libraries (pure CMake packages) and ROS packages together in the same colcon workspace (our team often does this). 

[Here](https://github.com/ignitionrobotics/ros_ign/blob/e2feecf0cbba07a2f2b630c3dd42a674d251bc61/ros_ign_bridge/package.xml#L15-L16)'s an example that can be in the same colcon workspace as Ignition:

```.xml
  <depend>ignition-msgs5</depend>
  <depend>ignition-transport8</depend>
```

However, the above doesn't work with `rosdep install`. Ideally, we'd declare Ignition dependencies once, and they would work with either `rosdep` or `colcon`.

Therefore, I propose here that we change the `-dev` keys to match the CMake project names.

Let me know if this is a big taboo and we can discuss alternatives.

CC @j-rivero @sloretz @dirk-thomas

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2549,7 +2549,7 @@ libicu-dev:
   fedora: [libicu-devel]
   gentoo: [dev-libs/icu]
   ubuntu: [libicu-dev]
-libignition-cmake2-dev:
+ignition-cmake2:
   debian:
     buster: [libignition-cmake2-dev]
   ubuntu:
@@ -2559,7 +2559,7 @@ libignition-common3:
     buster: [libignition-common3]
   ubuntu:
     focal: [libignition-common3]
-libignition-common3-dev:
+ignition-common3:
   debian:
     buster: [libignition-common3-dev]
   ubuntu:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Sounds reasonable to me. (To pass CI the renamed entries must be moved into alphabetical order.)

jacobperron (approved):
LGTM

nuclearsandwich:
@tfoote has been the driving editorial force behind rosdep keys. I'd like to give him the opportunity to review or explicitly abstain from doing so.

tfoote (approved):
Yeah, in general we pick the debian names because it's the most generic. However if this can work with the upstream packages from source too then an exception to that makes sense.

The biggest thing though that I'd point out is that if you do this you do not have the ability to explicitly call out the runtime libraries versus the development libraries which is valuable for packing. But I guess in the future we can consider explicitly adding `lib*` and `*-dev` keys in parallel. But for now if this works for both source space and rosdep resolutions that sounds like a win.

chapulina:
> you do not have the ability to explicitly call out the runtime libraries

So... I left the `lib*` keys in, so people can still use those for runtime dependencies. And I turned the `lib*-dev` keys into `ignition-*`. Not sure if it was a good idea, the assumption is that when compiling from source, you often want the `-dev` packages.

tfoote:
That's a good compromise.

PR #25806, opened 2020-07-14
Enable pull request testing for system_tests.

nuclearsandwich:
From the discussion in https://github.com/ros2/system_tests/pull/441#issuecomment-657741298 enables pull request testing for the system_tests repository on Rolling.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2476,6 +2476,7 @@ repositories:
     status: developed
   system_tests:
     source:
+      test_pull_requests: true
       type: git
       url: https://github.com/ros2/system_tests.git
       version: foxy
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Should PR jobs also be enabled for other distros?

nuclearsandwich:
> Should PR jobs also be enabled for other distros?

@jacobperron do you want system_tests PR testing enabled in Foxy? It is the only other distro with a system_tests entry.

jacobperron:
>  do you want system_tests PR testing enabled in Foxy? It is the only other distro with a system_tests entry.

@nuclearsandwich Yes please, if you'd be so kind :)

jacobperron (approved):

PR #25841, opened 2020-07-16
udp_com: 0.0.6-1 in 'melodic/distribution.yaml' [bloom]

flynneva:
Increasing version of package(s) in repository `udp_com` to `0.0.6-1`:

- upstream repository: https://github.com/continental/udp_com.git
- release repository: https://github.com/flynneva/udp_com-release.git
- bloom version: `0.9.7`

## udp_com

```
* switched branch to ros1-devel
* change branch name to ros1-devel
* switched destination branch to melodic-devel
* Contributors: Evan Flynn
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -11661,6 +11661,12 @@ repositories:
       url: https://github.com/peci1/ubnt_airos_tools.git
       version: master
     status: developed
+  udp_com:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/flynneva/udp_com-release.git
+      version: 0.0.6-1
   um6:
     doc:
       type: git


flynneva:
so this was my first time ever using bloom or releasing a package to ROS. I am going to guess I did something wrong so that was one reason why I started with a version <1.0.0.

I think for this PR I didnt add anything for documentation or maintenance status. I should be able to change that on the next release right?

tfoote:
Yeah, the first attempts are often learning experiences. 

You can change that meta data in the next release or even just a straight pull request here. We strongly recommend adding the source entries so that people can check it out from source using the automated tools. And if you add the documentation urls we'll run the documentation generation too.

flynneva:
for the documentation url's, ours is just doxygen in the same repo so we should just set it to the same repo as the source then? how does your tool know which documentation generator to use or do we specify doxygen somewhere?

tfoote:
By default we look for doxygen using the [rosdoc_lite](http://wiki.ros.org/rosdoc_lite) tool. [Doxygen in ROS guide](http://wiki.ros.org/Doxygen) and [indexing overview](http://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation) are also good references.

PR #25857, opened 2020-07-19
py_trees_msgs: 0.3.6-1 in 'noetic/distribution.yaml' [bloom]

stonier:
Increasing version of package(s) in repository `py_trees_msgs` to `0.3.6-1`:

- upstream repository: https://github.com/splintered-reality/py_trees_msgs.git
- release repository: https://github.com/stonier/py_trees_msgs-release.git
- bloom version: `0.9.7`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2524,6 +2524,21 @@ repositories:
       type: git
       url: https://github.com/fetchrobotics/power_msgs.git
       version: ros1
+  py_trees_msgs:
+    doc:
+      type: git
+      url: https://github.com/splintered-reality/py_trees_msgs.git
+      version: release/0.3.x
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -2524,6 +2524,21 @@ repositories:
       type: git
       url: https://github.com/fetchrobotics/power_msgs.git
       version: ros1
+  py_trees_msgs:
+    doc:
+      type: git
+      url: https://github.com/splintered-reality/py_trees_msgs.git
```
clalancette, line 2530: @stonier It looks like previous releases of this package were done from the `stonier` github organization.  This one looks to be released from `splintered-reality`.  That's totally fine, but should we update the older versions (in Melodic) to point to the splintered-reality organization instead?

stonier: Thought about that, but should be fine with github redirects right?

sloretz (approved):
Seems fine for Noetic, merging without resolving what to do about melodic distribution.yaml

PR #25812, opened 2020-07-14
Enable test_abi for Foxy: class_loader, console_bridge_vendor, libyaml_vendor, spdlog_vendor, unique_identifier_msgs

j-rivero:
Following pull 25638, more ABI checker for Foxy packages under quality level
//cc @chapulina

PR #25880, opened 2020-07-21
Enable test_abi for Foxy repositories: rmw_cyclonedds, ros2_tracing, rosidl_defaults, rmw_implementation, rmw_dds_common

j-rivero:
Following pull #25812, more ABI checker for Foxy packages under quality level
//cc @chapulina

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1744,6 +1744,7 @@ repositories:
       url: https://github.com/ros2-gbp/rmw_cyclonedds-release.git
       version: 0.7.2-1
     source:
+      test_abi: true
       test_pull_requests: true
       type: git
       url: https://github.com/ros2/rmw_cyclonedds.git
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -1899,6 +1902,7 @@ repositories:
       url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing-release.git
       version: 1.0.1-2
     source:
+      test_abi: true
```
jacobperron, line 1905: Does it matter that this repository is hosted on GitLab instead of GitHub?

j-rivero: The abi checking uses `auto-abi` tool and in the [ROS buildfarm it does work with](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/scripts/devel/build_and_install.py#L49-L50):
 * released packages
 * install directory in the workspace

gitlab/github should not impact in how it works, as far as I can say.

jacobperron (approved):

PR #25892, opened 2020-07-22
PyPy sense-hat sense-emu x Raspberry Pi Sense HAT

allxone:
Pip-only Python packages:
- add `python-sense-hat-pip` entry for `ubuntu` and `debian` to support Raspberry Pi Sense HAT (https://pypi.org/project/sense-hat/)
- add `python-sense-emu-pip` entry for `ubuntu` and `debian` to support tests and emulation (https://pypi.org/project/sense-emu/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4186,6 +4186,20 @@ python-semver:
   debian: [python-semver]
   fedora: [python-semver]
   ubuntu: [python-semver]
+python-sense-emu-pip:
+  debian:
+    pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Could you update this to use `python3-` as the prefix. Almost all platforms python3 will be the default going forward so we might as well match that for new pip rules so they sort together.

clalancette (approved):

tfoote (approved):

PR #25908, opened 2020-07-23
Add in a script to sync source entries with release repo devel_branch.

clalancette:
When doing a release of a ROS package through bloom, it is possible
to set the branch in the rosdistro 'source' entry different from the
"devel_branch" that is used in the release repository.  This is
not recommended, as doing further releases will now cause bloom
to look at the "wrong" branch for tags.  The situation gets even
worse since both the rosdistro 'source' entry and the devel_branch
in the release repository can be changed by hand.

This commit adds a script to sync the "devel_branch" in the release
repository with the rosdistro 'source' branch.  This ensures that
future releases will always find the version from the correct
branch.

See the script itself for more documentation on how to run it,
and the options it has.

### DIFF
--- a/scripts/sync-ros2-gbp-devel-branch.py
+++ b/scripts/sync-ros2-gbp-devel-branch.py
@@ -1,3 +1,30 @@
+# Copyright (c) 2020, Open Source Robotics Foundation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of the Willow Garage, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived from
+#       this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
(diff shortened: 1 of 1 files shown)


scripts/sync-ros2-gbp-devel-branch.py
```diff
@@ -0,0 +1,219 @@
+# This is a program to sync rosdistro source entries with the release repository devel_branch.
```
nuclearsandwich, line 1: Based on the licensing discussion in #26055 this should probably get a copyright header.

scripts/sync-ros2-gbp-devel-branch.py
```diff
@@ -0,0 +1,219 @@
...
+
+        if tracks_yaml_distro['devel_branch'] != repo['source']['version']:
+            print("Package '{reponame}' rosdistro source branch ({source_branch}) does not match release branch ({release_branch})".format(reponame=tracks_yaml_distro['name'], source_branch=repo['source']['version'], release_branch=tracks_yaml_distro['devel_branch']))
+
+            if args.dry_run:
+                continue
+
+            gh_body = """This PR from an automated script updates the devel_branch for {ros_distro} to match the source branch as specified in https://github.com/ros/rosdistro/{ros_distro}/distribution.yaml .
+
+Links to https://github.com/ros2/ros2/issues/963 .
```
nuclearsandwich, line 185: Does it make sense to keep this issue link if the issue is closed?

scripts/sync-ros2-gbp-devel-branch.py
clalancette: Good call, will add.

scripts/sync-ros2-gbp-devel-branch.py
clalancette: Yeah, that's a good question.  I had a vague idea that if we were ever to use this script again, we could just reopen that issue.

Maybe a better idea is to have a command-line argument that allows the user to pass the link they want to use in.  What do you think?

nuclearsandwich: I'm generally on team don't add options you don't already need. I think it'd make more sense to first PR the script to use a different issue body.

clalancette: Should be fixed in 2db8383

jacobperron (approved):
Is the idea that we run this during the release process for a new distro? I wonder if it's worth running semi-regularly (e.g. in a Github workflow), then we can catch inconsistencies more often.

scripts/sync-ros2-gbp-devel-branch.py
```diff
@@ -0,0 +1,244 @@
...
+
+            if args.dry_run:
+                continue
+
+            gh_body = """This PR from an automated script updates the devel_branch for {ros_distro} to match the source branch as specified in https://github.com/ros/rosdistro/{ros_distro}/distribution.yaml .
+""".format(ros_distro=ros_distro)
+            commit_message = """Change the devel_branch for {ros_distro}.
+
+This makes it match the source entry in https://github.com/ros/rosdistro/{ros_distro}/distribution.yaml
+""".format(ros_distro=ros_distro)
```
hidmic, line 215: @clalancette nit^2: you may split strings and use `textwrap.dedent` to not break indentation and keep line lengths reasonable.

hidmic (approved):
LGTM too!

clalancette:
> Is the idea that we run this during the release process for a new distro?

That was the original idea when I wrote it.  However, given that we are going to likely fork off of rolling for new distributions going forward, this script is less important.  That is, as long as rolling is correct, the distro we fork off of it should also be correct.

> I wonder if it's worth running semi-regularly (e.g. in a Github workflow), then we can catch inconsistencies more often.

Nevertheless, this would still probably be worthwhile to run once a month or something, just to make sure we don't miss anything.

Thanks for the reviews.  I'm going to go ahead and merge this as is, and we can do some of the other stuff as follow-up.

PR #25963, opened 2020-07-28
Add dynamic-graph-python to melodic

Rascof:
Release dynamic-graph-python a third-party package from SOT on melodic
It is dependent on dynamic-graph that is already released on melodic : #25488 for reference

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2130,6 +2130,22 @@ repositories:
       url: https://github.com/stack-of-tasks/dynamic-graph.git
       version: devel
     status: maintained
+  dynamic-graph-python:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


sloretz (approved):

melodic/distribution.yaml
```diff
@@ -2130,6 +2130,22 @@ repositories:
...
       version: devel
     status: maintained
+  dynamic-graph-python:
+    doc:
+      type: git
+      url: https://github.com/stack-of-tasks/dynamic-graph-python.git
+      version: devel
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
```
sloretz, line 2140: whoops, missed this; I see it's fixed in https://github.com/ros/rosdistro/pull/25965/files

PR #25986, opened 2020-07-29
add keypress_monitor

smihael:
Added keypress_monitor to index for melodic according to http://wiki.ros.org/ROS/ReleasingAPackage#Adding_your_package_to_the_index.

It's a simple tool to monitor whether a key has been pressed or not. Useful to map keys on bluetooth media keyboards to robot actions. Similar but more general than not actively maintained [keyboard package](http://wiki.ros.org/keyboard).

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4551,6 +4551,16 @@ repositories:
       url: https://github.com/ros/kdl_parser.git
       version: melodic-devel
     status: maintained
+  keypress_monitor:
+    doc:
+      type: git
+      url: https://repo.ijs.si/msimonic/keypress_monitor.git
+      version: master
+    source:
+      type: git
+      url: https://repo.ijs.si/msimonic/keypress_monitor.git
+      version: master
+    status: maintained
   kinesis_manager:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -4551,6 +4551,16 @@ repositories:
       url: https://github.com/ros/kdl_parser.git
       version: melodic-devel
     status: maintained
+  keypress_monitor:
+    doc:
+      type: git
+      url: https://repo.ijs.si/msimonic/keypress_monitor.git
```
clalancette, line 4557: It looks like the license listed in the package.xml is GPLv2.  Can you please also add a COPYING file to the top-level as specified in the GPL license?

clalancette (changes requested):

melodic/distribution.yaml
smihael: @clalancette Thanks for pointing that out. I added the copying file in https://repo.ijs.si/msimonic/keypress_monitor/-/commit/4b1616e147cfc8573ac3c8f8bfd0731d64e1cf98

clalancette (approved):

PR #26005, opened 2020-07-30
Update rqt_graph development branch for Dashing and later

jacobperron:
The latest development has changed to `dashing-devel` since https://github.com/ros-visualization/rqt_graph/pull/49


dirk-thomas:
The branch needs to be updated for all distros starting with Dashing.

dirk-thomas (approved):

PR #26049, opened 2020-08-04
added udp_com to kinetic

flynneva:


flynneva:
tried to use bloom to automatically do this but it failed and recommend manual PR

PR #26056, opened 2020-08-05
[rosdep] add python3 keys for pymodbus and urllib3

mikaelarguedas:
pymodbus:
debian: https://packages.debian.org/buster/python3-pymodbus
fedora: https://fedora.pkgs.org/32/fedora-x86_64/python3-pymodbus-2.3.0-2.fc32.noarch.rpm.html
ubuntu: https://packages.ubuntu.com/focal/python3-pymodbus

urllib3:
debian: https://packages.debian.org/buster/python3-urllib3
fedora: https://fedora.pkgs.org/32/fedora-x86_64/python3-urllib3-1.25.7-3.fc32.noarch.rpm.html
gentoo: https://packages.gentoo.org/packages/dev-python/urllib3
ubuntu: https://packages.ubuntu.com/focal/python3-urllib3

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6408,6 +6408,11 @@ python3-twisted:
     '*': ['python%{python3_pkgversion}-twisted']
     '7': null
   ubuntu: [python3-twisted]
+python3-urllib3:
+  debian: [python3-urllib3]
+  fedora: [python3-urllib3]
+  gentoo: [dev-python/urllib3]
+  ubuntu: [python3-urllib3]
 python3-usb:
   debian: [python3-usb]
   gentoo: [dev-python/pyusb]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5998,6 +5998,10 @@ python3-pykdl:
     '*': [python3-pykdl]
     'bionic': null
     'xenial': null
+python3-pymodbus:
+  debian: [python3-pymodbus]
+  fedora: [python3-pymodbus]
+  ubuntu: [python3-pymodbus]
 python3-pyparsing:
   arch: [python-pyparsing]
   debian: [python3-pyparsing]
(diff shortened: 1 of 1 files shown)


sloretz (approved):

rosdep/python.yaml
```diff
@@ -5998,6 +5998,10 @@ python3-pykdl:
     '*': [python3-pykdl]
     'bionic': null
     'xenial': null
+python3-pymodbus:
+  debian: [python3-pymodbus]
```
clalancette, line 6002: This key isn't available in `stretch`.  It *is* available in `stretch-backports`, but we don't usually allow keys in the backports repository, as I understand it.  So I think this will have to get split out to a `null` key for `stretch` and `python3-pymodbus` for everything else.

rosdep/python.yaml
```diff
@@ -5998,6 +5998,10 @@ python3-pykdl:
...
+  fedora: [python3-pymodbus]
+  ubuntu: [python3-pymodbus]
```
clalancette, line 6004: There's a similar situation for Ubuntu; the key is not available in Xenial, so a null key there and `python3-pymodbus` everywhere else.

clalancette (changes requested):

rosdep/python.yaml
mikaelarguedas: My bad, addressed in https://github.com/ros/rosdistro/pull/26056/commits/85aee79a8a6de248b97f0596cfc9a2324b4d1025

rosdep/python.yaml
mikaelarguedas: I didnt know stretch was still a required platform for new keys as it went EOL last month.
Addressed in https://github.com/ros/rosdistro/pull/26056/commits/85aee79a8a6de248b97f0596cfc9a2324b4d1025

clalancette: > I didnt know stretch was still a required platform for new keys as it went EOL last month.

I didn't realize that.  Thanks for the heads up; for now, I think we can leave the key since we've added it.

clalancette (approved):

PR #26150, opened 2020-08-10
costmap_tf_layer: 1.0.1-1 in 'melodic/distribution.yaml' [bloom]

igorbanfi:
Increasing version of package(s) in repository `costmap_tf_layer` to `1.0.1-1`:

- upstream repository: https://github.com/igorbanfi/costmap_tf_layer.git
- release repository: https://github.com/igorbanfi/costmap_tf_layer-release.git
- bloom version: `0.9.7`

## costmap_tf_layer

```
* Initial commit
* Contributors: Igor Banfi
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -1762,6 +1762,17 @@ repositories:
       url: https://github.com/rst-tu-dortmund/costmap_converter.git
       version: master
     status: developed
+  costmap_tf_layer:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/igorbanfi/costmap_tf_layer-release.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/igorbanfi/costmap_tf_layer.git
+      version: melodic-devel
+    status: developed
   cpr_multimaster_tools:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -11283,6 +11283,17 @@ repositories:
       url: https://github.com/RobotWebTools/tf2_web_republisher.git
       version: master
     status: maintained
+  tf_layer:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/igorbanfi/tf_layer-release.git
```
clalancette, line 11290: Since this is such a new package, I'm going to suggest that you rename it.  `tf_layer` is very ambiguous, and actually doesn't totally describe what it does; from the description, it looks like it does costmaps from tfs.

The name used here matters, as this is the name that gets used to build the debian packages.  Imagine seeing `ros-melodic-tf-layer` in your list of packages; it doesn't tell you much about what it does, and it could be used for a lot of things.

I'll suggest renaming repository to something like `costmap_tf_layer` (GitHub will automatically redirect from the old URL), and also renaming the package to `costmap_tf_layer`.  You can choose another name, but please make it less ambiguous.

clalancette (changes requested):

melodic/distribution.yaml
igorbanfi: Have renamed the package to `costmap_tf_layer`.

clalancette: Thanks for agreeing to that!  It makes it much easier to fit into the ecosystem.

clalancette (approved):

PR #26211, opened 2020-08-17
Adding new repository for indexing: gpio_control

cst0:
This package provides control of GPIO pins from ROS. Support is offered both through hardware-specific API's of popular platforms (most notably the Raspberry Pi and Nvidia Jetson) which come bundled with the official operating system for the device, as well as through the Linux filesystem hierarchy standard, which allows for control of GPIO pins through manipulating the contents of the /sys/class/gpio directory. This in turn allows for support of a wide variety of Linux devices with GPIO pins. Additionally, pieces of the implementation of the node are made available in a utilities class, which provides a platform-agnostic implementation of GPIO pin control to aid with software stacks that attempt to avoid locking themselves into one platform.

This code has been tested on Melodic and Kinetic on a handful of platforms in a handful of different configurations. This is my first time contributing code to the ROS index, so please let me know if there are any additional steps I need to take or things I need to change.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -4260,6 +4260,11 @@ repositories:
         release: release/kinetic/{package}/{version}
       url: https://github.com/robustify/gmplot_ros-release.git
       version: 1.0.1-0
+  gpio_control:
+    doc:
+      type: git
+      url: https://github.com/cst0/gpio_control-release
+      version: master
   gps_goal:
     doc:
       type: git
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -4265,6 +4265,7 @@ repositories:
       type: git
       url: https://github.com/cst0/gpio_control-release
       version: master
+    status: maintained
   gps_goal:
     doc:
       type: git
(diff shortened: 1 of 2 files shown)


cst0:
I believe I see the issue with the automatic checks: the URL in my package.xml was incorrect. I've pushed a fix to that issue.

cst0:
Well, I'm stumped. Still getting an issue in regards to a link being incorrect somewhere, but all the links I've put into the yaml's and into my package.xml resolve fine. What am I missing here?

clalancette (changes requested):
There are a few different problems here.

First, please split up adding your package to different ROS distributions into different PRs.  The main reason for this is that the ROS distros are on different sync schedules, so it can sometimes be hard to find a window when they are both not in freeze.

Second, you really should not name your repository `gpio_control-release`.  While we allow any arbitrary repository names, by convention the `-release` repository is the repository that contains the metadata for building the binary packages.  The source repositories generally don't have that `-release` suffix, so I'll propose you rename your repository to `gpio_control`.

Third, there should be no "version" tag at the same level as `doc`; that's what the linter is complaining about.

I think with all of those fixed, this will be in much better shape.

cst0:
OK, thanks for the help, I'll make those changes. I'm going to go ahead and close this PR since you've asked to split the addition into two different PR's.

PR #26214, opened 2020-08-17
neo_local_planner: 1.0.0-1 in 'melodic/distribution.yaml'

padhupradheep:


melodic/distribution.yaml
```diff
@@ -6268,6 +6268,17 @@ repositories:
       url: https://github.com/ros-planning/navigation_tutorials.git
       version: indigo-devel
     status: maintained
+  neo_local_planner:
+    doc:
+      type: git
+      url: https://github.com/neobotix/neo_local_planner.git
+      version: 1.0.0
```
clalancette, line 6275: First, this tag doesn't exist in the repository.  Second, we actually don't support tags here; it must be a branch.  If you switch this to the `melodic` branch, the linters will probably pass.

Note that your source repository is also missing a LICENSE file, so please add that as well.

clalancette (changes requested):

melodic/distribution.yaml
padhupradheep: Changed the tags and added the LICENSE file! 

Pradheep

clalancette: Thanks, appreciated.

PR #26215, opened 2020-08-17
Add jupyter-nbconvert

azazdeaz:
Package Name: `jupyter-nbconvert`
 - arch: https://www.archlinux.org/packages/community/any/jupyter-nbconvert/
 - debian: https://packages.debian.org/buster/jupyter-nbconvert
 - ubuntu: https://packages.ubuntu.com/eoan/jupyter-nbconvert

It makes it simple to implement ROS executables as notebooks. ([example](https://github.com/azazdeaz/ros-grpc-api-generator/blob/ros/scripts/generate#L3))

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -204,6 +204,7 @@ jupyter-nbconvert:
   arch: [jupyter-nbconvert]
   debian: [jupyter-nbconvert]
   fedora: [jupyter-nbconvert]
+  gentoo: [dev-python/nbconvert]
   ubuntu: 
     '*': [jupyter-nbconvert]
     xenial: null
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -205,7 +205,7 @@ jupyter-nbconvert:
   debian: [jupyter-nbconvert]
   fedora: [jupyter-nbconvert]
   gentoo: [dev-python/nbconvert]
-  ubuntu: 
+  ubuntu:
     '*': [jupyter-nbconvert]
     xenial: null
 jupyter-notebook:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -200,6 +200,10 @@ ipython3:
   debian: [ipython3]
   fedora: [ipython3]
   ubuntu: [ipython3]
+jupyter-nbconvert:
+  arch: [jupyter-nbconvert]
+  debian: [jupyter-nbconvert]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -200,6 +200,10 @@ ipython3:
   debian: [ipython3]
   fedora: [ipython3]
   ubuntu: [ipython3]
+jupyter-nbconvert:
+  arch: [jupyter-nbconvert]
+  debian: [jupyter-nbconvert]
+  ubuntu: [jupyter-nbconvert]
```
clalancette, line 206: Unfortunately, `jupyter-nbconvert` isn't available on Ubuntu Xenial, which we are still supporting until next year.  We'll have to make a concession to that:

```
ubuntu:
  '*': [jupyter-nbconvert]
  xenial: null
```

clalancette (dismissed):
One thing inline.

Also, it looks like this is available in Fedora as well: https://fedora.pkgs.org/31/fedora-x86_64/python3-nbconvert-5.6.0-2.fc31.noarch.rpm.html .  Can you please add that key?

rosdep/python.yaml
azazdeaz: Ah, thank you for catching it! :)

claireyywang (changes requested):
there's a gentoo one as well https://packages.gentoo.org/packages/dev-python/nbconvert for our packaging job, would be nice to add that.

Not sure why the CI failed though. It seems to be about the formatting of this PR. Maybe check if there's any trailing white space?

azazdeaz:
> Not sure why the CI failed though. It seems to be about the formatting of this PR. Maybe check if there's any trailing white space?

That was the issue, thanks! :)

rosdep/python.yaml
```diff
@@ -200,6 +200,14 @@ ipython3:
   debian: [ipython3]
   fedora: [ipython3]
   ubuntu: [ipython3]
+jupyter-nbconvert:
+  arch: [jupyter-nbconvert]
+  debian: [jupyter-nbconvert]
+  fedora: [jupyter-nbconvert]
```
clalancette, line 206: From the link I posted earlier, the Fedora key should be `python3-nbconvert`

azazdeaz: Thank you!

claireyywang (approved):
LGTM

PR #26226, opened 2020-08-18
adding graphql client library as pip to rosdep

bbferka:
new entry for the graphql python client library
- GIT:  https://github.com/graphql-python/gql
- pypi: https://pypi.org/project/gql/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1908,13 +1908,6 @@ python-gpxpy:
     zesty:
       pip:
         packages: [gpxpy]
-python-graphitesend-pip:
-  debian:
-    pip:
-      packages: [graphitesend]
-  ubuntu:
-    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1908,6 +1908,16 @@ python-gpxpy:
     zesty:
       pip:
         packages: [gpxpy]
+python-gql-pip:
```
tfoote, line 1911: Our new policy is to make new rules prefixed with python3 not python as that's the new default.

tfoote (changes requested):
please update the prefix. Note that you'll have to move it to maintain alphabetization.

tfoote (approved):
No problem. Thanks for iterating. lgtm

jacobperron (approved):

PR #26235, opened 2020-08-19
noetic: Add doc info for sensehat_ros

allxone:
I'd like sensehat_ros to be indexed and documented on ros.org.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4303,6 +4303,11 @@ repositories:
         release: release/noetic/{package}/{version}
       url: https://github.com/ros-gbp/sbpl-release.git
       version: 1.3.1-3
+  sensehat_ros:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


mabelzhang (approved):

mabelzhang (commented):
Looks fine. I noticed this isn't created by Bloom. Is there a reason for that? I only ask because that is the typical preferred way.

allxone:
@mabelzhang I understand ROS packages with Python pip dependencies in the package.xml can't be built by the ROS buildfarm ([#764](https://github.com/ros-infrastructure/ros_buildfarm/issues/764)).
sensehat_ros has two Python pip dependencies installed via rosdep (python3-sense-emu-pip and python3-sense-hat-pip). These Python packages are not available via official Debian packages.
I can't see any viable option here, please let me know if I am wrong.

mabelzhang:
@allxone Thank you for the explanation!

PR #26250, opened 2020-08-20
diagnostics: 1.9.6 in melodic/distribution.yaml

g-gemignani:


g-gemignani:
I found an issue in the release that slipped through the tests. I am closing it and reopening it once I clarified it.

PR #26325, opened 2020-08-26
Update python-dev for focal

paulbovbel:
I apologize if I'm barking up the wrong tree, so bear with me please.

I'm trying to fix the following build: http://build.ros.org/view/Ndev/job/Ndev__catkin_virtualenv__ubuntu_focal_amd64/4/console

catkin_virtuaelnv allows downstream packages to bundle a python2 or python3 virtualenv, so it needs to add a `python-dev` build_export_depends.

This breaks in focal, even though AFAIK `python-dev` exists as a virtual package for `python-dev-is-python2`. Is there a reason the builder is unable to pull that package in?

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5381,7 +5381,9 @@ pybind11-dev:
 python-dev:
   debian: [python-dev]
   fedora: [python2-devel]
-  ubuntu: [python-dev]
+  ubuntu:
+    '*': [python-dev]
+    'focal': [python-dev-is-python2]
 qhull-bin:
   arch: [qhull]
   debian: [qhull-bin]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5383,7 +5383,7 @@ python-dev:
   fedora: [python2-devel]
   ubuntu:
     '*': [python-dev]
-    'focal': [python-dev-is-python2]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -5381,7 +5381,9 @@ pybind11-dev:
 python-dev:
   debian: [python-dev]
   fedora: [python2-devel]
-  ubuntu: [python-dev]
+  ubuntu:
+    '*': [python-dev]
+    'focal': [python-dev-is-python2]
```
jacobperron, line 5386: I think the issue you're seeing is because `python-dev` does not exist in Focal. Maybe the correct thing to do here is install [`python2-dev`](https://packages.ubuntu.com/focal/python2-dev) instead?

paulbovbel: python-dev is 'pure virtual', though I'm not super familiar with what that means. I think you're right though that I should just use python2-dev.

Should I create a new rosdep key for that, or should I replace the focal entry for `python-dev` so as to keep the package.xml consistent?

jacobperron: Replacing the focal entry as you've done seems right to me. Let's get a third opinion in any case.

jacobperron (approved):

ahcorde (approved):

PR #26336, opened 2020-08-27
Adding heifu to documentation index for distro melodic

BV-OpenSource:


melodic/distribution.yaml
```diff
@@ -3766,6 +3766,31 @@ repositories:
       url: https://github.com/tu-darmstadt-ros-pkg/hector_slam.git
       version: melodic-devel
     status: maintained
+  heifu:
+    doc:
+      type: git
+      url: https://heifu:G2SzWMcydduaaXwttrxA@gitlab.pdmfc.com/drones/ros1/heifu.git
```
tfoote, line 3772: Please update this to be a publicly accessible repository which does not require credentials to access. Not all tools can handle embedded tokens.

melodic/distribution.yaml
```diff
@@ -3766,6 +3766,31 @@ repositories:
...
+      version: debian
+    release:
+      packages:
+      - heifu_bringup
+      - heifu_description
+      - heifu_diagnostic
+      - heifu_interface
```
tfoote, line 3779: Not all of these appear to be packages.

melodic/distribution.yaml
```diff
@@ -3766,6 +3766,31 @@ repositories:
...
+      - heifu_mavros
+      - heifu_msgs
+      - heifu_safety
+      - heifu_scripts_firmware
+      - heifu_simple_waypoint
+      - heifu_tools
+      - heifu_usb
+      - scripts
+      url: https://heifu:G2SzWMcydduaaXwttrxA@gitlab.pdmfc.com/drones/ros1/heifu.git
```
tfoote, line 3788: This url should be a release repository not the package repository. To make your first release please see this tutorial: http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

melodic/distribution.yaml
```diff
@@ -3766,6 +3766,31 @@ repositories:
...
+    source:
+      type: git
+      url: https://heifu:G2SzWMcydduaaXwttrxA@gitlab.pdmfc.com/drones/ros1/heifu.git
```
tfoote, line 3791: Likewise this repository should be accessible anonymously without credentials over https.

melodic/distribution.yaml
```diff
@@ -3766,6 +3766,31 @@ repositories:
...
+      - heifu_diagnostic
+      - heifu_interface
+      - heifu_mavros
+      - heifu_msgs
+      - heifu_safety
+      - heifu_scripts_firmware
+      - heifu_simple_waypoint
+      - heifu_tools
+      - heifu_usb
+      - scripts
```
tfoote, line 3787: I don't know if this is a ROS package or not, but if it is it doesn't meet our naming guidelines.

tfoote (dismissed):
Thank you for submitting your package(s). To be indexed please open up your repository for people to be able to browse it. It's also valuable to enable people to open tickets and interact using the standard open source processes.

It also looks like you've manually recreated the full stanzas without running the release processes required to generate the necessary artifacts. I've linked to the tutorial for making your first release. Please consider going through that and the other tutorials you can find here: http://wiki.ros.org/bloom/Tutorials

melodic/distribution.yaml
amferreiraBEV: Where can I find those naming guidelines?

melodic/distribution.yaml
```diff
@@ -3766,6 +3766,29 @@ repositories:
...
+      - heifu_description
+      - heifu_diagnostic
+      - heifu_mavros
+      - heifu_msgs
+      - heifu_safety
+      - heifu_simple_waypoint
+      - heifu_tools
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/amferreiraBEV/heifu-release.git
```
dirk-thomas, line 3786: The `release` section lacks a `version` (see other entries in this file for an example).

amferreiraBEV: Should the version be the same as the tag name? Because it's "upstream/0.7.1".
Can it be 0.7.1?

dirk-thomas: Looking at the latest Melodic release in this gbp repo I think it should be `0.7.1-2`.

amferreiraBEV: You're totally right! I'm still new at this things.

dirk-thomas: Looking good now. No worry, next release is going to be much easier. :+1:

shr-project:
How are rosdistro commits being tested? I've just noticed in meta-ros builds, that heifu-{diagnostics,mavros,safety,simple-waypoint} have a dependency on ros_utils, which isn't defined anywhere in rosdistro for melodic nor other distributions and based on
https://gitlab.pdmfc.com/drones/ros1/heifu/-/blob/master/README.md
it's some heifu related component which doesn't seem to be public at all (the link in in README.md https://gitlab.pdmfc.com/drones/ros1/drone-generic/tree/master/src/ros_utils just asks me to login).

How should be build these components?

mjcarroll:
@shr-project The `ros/rosdistro` pull requests are tested via Github actions, which are defined here: https://github.com/ros/rosdistro/blob/master/.github/workflows/build_test.yaml   These are mainly checking for syntax/linting.

It looks like these packages have never successfully built in `melodic` according to the current status page: http://repositories.ros.org/status_page/ros_melodic_default.html?q=heifu

Since that is the case, the correct course of action will be to revert these for now, and @amferreiraBEV can open a new pull request with a corrected release.

amferreiraBEV:
@mjcarroll and @shr-project Thank you for your attention.

The path for `ros_utils` was update to the public one. Do I have to change anything else?

PR #26372, opened 2020-08-29
Add python3-protobuf and python3-jinja2

f-fl0:
Add rosdep key for: 
- `python3-protobuf` in `alpine`, `debian`, `fedora`, `gentoo` and `ubuntu`.
- `python3-jinja2` in `alpine`, `debian`, `fedora`, `gentoo` and `ubuntu`.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5781,6 +5781,12 @@ python3-interpreter:
   rhel: [python3]
   slackware: [python3]
   ubuntu: [python3-minimal]
+python3-jinja2:
+  alpine: [py3-jinja2]
+  debian: [python3-jinja2]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6028,6 +6034,11 @@ python3-pkg-resources:
 python3-prometheus-client:
   debian: [python3-prometheus-client]
   ubuntu: [python3-prometheus-client]
+python3-protobuf:
```
dirk-thomas, line 6037: Is there a reason this doesn't have a Fedora?

f-fl0: @dirk-thomas I added the fedora entry.

hidmic (approved):

PR #26412, opened 2020-09-02
Update package names in rb1_base_common and rb1_base_sim repositories

David-rn:
The packages name that appear in the `distribution.yaml` file are outdated. This PR is intended to update those packages so it is up-to-date.


hidmic (commented):
Hmm, this won't work. This patch is attempting to change the list of packages of existing repository releases. Tests are failing to find those packages in upstream repositories at the last released tags.

@David-rn you have to `bloom` these repositories again.

tfoote:
This listing is of the packages as available in the currently released version 1.1.0  I see you have run bloom for version 2.0.0 with new packages. You will have to complete that process before updating the packages here. Note that since you're moving packages between repositories you should do simultaneous releases of both repositories into the rosdistro to make sure there's only ever one version of the package in the index.

I'm going to close this as we can't merge the changed packages without a release. Bloom should be able to open the appropriate pull request for you or at least create the necessary diff if you don't want to authorize it to open pull requests and since you should batch up the two releases to not create overlapping packages in the distro.

David-rn:
Oh, okay. I didn't know that. Thank you for your response! I'll try your suggestions.

David-rn:
Should I open a PR manually? Seems that [rb1_base_common-release](https://github.com/RobotnikAutomation/rb1_base_common-release) and [rb1_base_sim-release](https://github.com/RobotnikAutomation/rb1_base_sim-release) repositories are updated and do not have overlapping packages now. However, when I do the `bloom` command it fails opening a pull request.

PR #26446, opened 2020-09-03
adds libnlopt-cxx-dev

mgruhler:
Starting with `eoan`, `libnlopt-dev` has been split into `libnlopt-dev` (containing the C library) and `libnlopt-cxx-dev` (containing the C++ library) of the `nlopt` package.
The split effects Ubuntu and Debian.

This PR now adds the C++ library (for focal+ / buster+).

Ubuntu: https://packages.ubuntu.com/search?keywords=libnlopt-cxx-dev&searchon=names
Debian: https://packages.debian.org/search?searchon=names&keywords=libnlopt-cxx-dev

For convenience, I added the (unaffected) packages for Fedora and Gentoo, which are the same as for the `libnlopt-dev` package.

I excluded `bionic`, `xenial` and `stretch`, as `libnlopt-cxx-dev` does not exist there. Onwards, I assume it will be released.

This package split is actually a breaking change for current users of the `libnlopt-dev` package,  as the `nlopt.hpp` header which is usually used is not distributed with that package anymore, but with `libnlopt-cxx-dev`.
For reference, here are the Ubuntu file lists:
- [Focal / libnlopt-cxx-dev](https://packages.ubuntu.com/focal/amd64/libnlopt-cxx-dev/filelist)
- [Focal / libnlopt-dev](https://packages.ubuntu.com/focal/amd64/libnlopt-dev/filelist)
- [Bionic / libnlopt-dev](https://packages.ubuntu.com/bionic/amd64/libnlopt-dev/filelist)

Thus, one could also change the `libnlop-dev` rosdep entry to switch use `libnlopt-cxx-dev` starting from `focal` / `buster`. This is under the assumption that as a ROS user you'll usually want the C++ library, and not the C one...

What do you prefer?

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3051,6 +3051,16 @@ libnl-dev:
   fedora: [libnl-devel]
   gentoo: [dev-libs/libnl]
   ubuntu: [libnl-dev]
+libnlopt-cxx-dev:
+  debian:
+    '*': [libnlopt-cxx-dev]
+    stretch: null
+  fedora: [NLopt-devel]
+  gentoo: [sci-libs/nlopt]
+  ubuntu:
+    '*': [libnlopt-cxx-dev]
+    bionic: null
+    xenial: null
 libnlopt-dev:
   debian: [libnlopt-dev]
   fedora: [NLopt-devel]


hidmic:
> This package split is actually a breaking change for current users of the libnlopt-dev package, as the nlopt.hpp header which is usually used is not distributed with that package anymore, but with libnlopt-cxx-dev.

That'a good point.

> Thus, one could also change the libnlop-dev rosdep entry to switch use libnlopt-cxx-dev starting from focal / buster. 

That's not a bad idea.

However, even if we merge pull in both libraries with the `libnlop-dev` key, packages authors will still have to adapt their `CMakeLists.txt`, won't they? If that's true, then the added benefit is small, while making it impossible to pull the core C library only.

rosdep/base.yaml
```diff
@@ -3051,6 +3051,16 @@ libnl-dev:
   fedora: [libnl-devel]
   gentoo: [dev-libs/libnl]
   ubuntu: [libnl-dev]
+libnlopt-cxx-dev:
+  debian:
+    '*': [libnlopt-cxx-dev]
```
hidmic, line 3056: @mgruhler is the package available in all distros but ` stretch`?

rosdep/base.yaml
```diff
@@ -3051,6 +3051,16 @@ libnl-dev:
...
+    stretch: null
+  fedora: [NLopt-devel]
+  gentoo: [sci-libs/nlopt]
+  ubuntu:
+    '*': [libnlopt-cxx-dev]
```
hidmic, line 3061: @mgruhler isn't the package available starting with `eoan`?

mgruhler: It is. My understanding is that this rule will work for all distros except the `null` ones (i.e. bionic and focal).
`eoan` is EOL'd anyways...

rosdep/base.yaml
mgruhler: it is in `buster (stable)`, `bullseye (testing)` and `sid (unstable)`. It's also available in `stretch-backports`. But obviously, it is not available in the older ones, i.e. the ones before stretch.

mgruhler:
> However, even if we pull in both libraries with the `libnlop-dev` key ... while making it impossible to pull the core C library only

`libnlopt-cxx-dev` depends on `libnlopt-dev`, so both libraries would be installed. But true, we couldn't pull in only the core C library via rosdep.

> packages authors will still have to adapt their `CMakeLists.txt`, won't they?

I'm not sure. I haven't used the plain C library with `CMake`, and prior to focal (well, I only tested Ubuntu LTS Releases), `nlopt` wasn't shipping any proper cmake files anyways, so you'd have to go the [`PkgConfig` way](https://bitbucket.org/traclabs/trac_ik/src/8396bb67157179ae704c816b0b56af8ac4eef62c/trac_ik_lib/CMakeLists.txt#lines-18:19). Now that there are proper cmake files, I think a change to the `CMakeLists.txt` would make sense whatever way we go.

> If that's true, then the added benefit is small

That's totally true. This is why I added it as a separate key. However, I wanted to get your opinions and follow whatever you deem more appropriate. I have no favorite :-)


As a last note, according to https://index.ros.org/d/libnlopt-dev/, this is only used by one package anyways. So I think the effects are small...

rosdep/base.yaml
hidmic: Exactly. Perhaps we should be specific if not too verbose.

rosdep/base.yaml
hidmic: > My understanding is that this rule will work for all distros except the `null` ones 

Correct, even for older ones where it does not apply. Perhaps we should be specific if not too verbose.

rosdep/base.yaml
mgruhler: so what do you suggest? should I rather add it specifically for the distros it DOES exist in? i.e.:

```yaml
debian:
  buster: [liblnopt-cxx-dev]
  bullseye: [liblnopt-cxx-dev]
```

rosdep/base.yaml
mgruhler: as above:

```yaml
ubuntu:
  eoan: [libnlopt-cxx-dev]  # well, this is EOL. Should we still add it?
  focal: [libnlopt-cxx-dev]
```

rosdep/base.yaml
mgruhler: Do we need to exclude the other ones? i.e. `'*': null`?
drawback is also that we'd have to add them manually for each new distro to come that does contain it.

In the end, I'm fine with whatever... :-)

dirk-thomas: > Do we need to exclude the other ones? i.e. `'*': null`?

Not existing mapping are equivalent to `null`. You only need to set explicit distros to `null` when using the wildcard `*` mapping to some value.

> drawback is also that we'd have to add them manually for each new distro to come that does contain it.

We try to avoid this wherever possible since it is a continuous maintenance burden and likely doesn't happen all the time.

> there are other keys where it is done similar to what I did here, basically exclude the current LTS versions that don't have this package and add for all others to come...

:+1:

rosdep/base.yaml
dirk-thomas: > `eoan: [libnlopt-cxx-dev]`  # well, this is EOL. Should we still add it?

Yes.

rosdep/base.yaml
hidmic: Hmm, yeah, that's a good point. Alright, disregard my comment.

rosdep/base.yaml
mgruhler: @dirk-thomas thanks for the feedback. I'm not quite sure though, what you want me to do. Should I 

1. keep it the way it is (wildcard set and thus available for `eoan`, but excluding ROS-supported LTS Distros `bionic` and `xenial`, where it does not exist, and effectively ignoring all non-LTS older then eoan) or 
1. change it to be specific (i.e. set explicitely for `eoan` and `focal` and ignore all others, effectively adding the "constant maintanence burden" you mention above)?

dirk-thomas: Option 1 is sufficient. Non-EOL distros (like option 2) isn't usually something we require for new keys. So it can stay as is imo.

dirk-thomas (approved):

rosdep/base.yaml
mgruhler: Great, thx.

rosdep/base.yaml
```diff
@@ -3051,6 +3051,16 @@ libnl-dev:
   fedora: [libnl-devel]
   gentoo: [dev-libs/libnl]
   ubuntu: [libnl-dev]
+libnlopt-cxx-dev:
+  debian:
+    '*': [libnlopt-cxx-dev]
+    stretch: null
```
ivanpauno, line 3057: It seems that `libnlopt-cxx-dev` will be available in the future in debian Stretch:
https://packages.debian.org/stretch-backports/libnlopt-cxx-dev

when that happens, this line can be removed.

ivanpauno (approved):

PR #26525, opened 2020-09-11
Adding orb_slam2_with_maps-odom to documentation index for distro

SuperAruba:
Adding orb_slam2_with_maps-odom to documentation index for distro.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9498,6 +9498,16 @@ repositories:
       url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
       version: master
     status: maintained
+  orb_slam2_with_maps-odom:
+    doc:
+      type: git
+      url: https://github.com/Inspur-ROS/ORB_SLAM2_WITH_MAPS-ODOM.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/Inspur-ROS/ORB_SLAM2_WITH_MAPS-ODOM.git
+      version: master
+    status: maintained
   orocos_kinematics_dynamics:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -9514,6 +9514,16 @@ repositories:
       url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
       version: master
     status: maintained
+  orb_slam2_with_maps-odom:
```
ivanpauno, line 9517: From the [reviewing guidelines](https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md):

> The source repository should generally contain one or more ROS packages (ROS packages are ones that have a package.xml and use either catkin or ament build helpers). Packages that are not ROS packages can be accepted, but they should be rare and require special handling in the release repository.

I don't think that the repository satisfies that requirement, as there aren't any ros packages in it.

ivanpauno (changes requested):

kinetic/distribution.yaml
tfoote: Yeah, without ROS packages in it we cannot index it or generate documentation. I'm going to close this.

tfoote (changes requested):

kinetic/distribution.yaml
SuperAruba: I have changed repository.

ivanpauno:
@SuperAruba the repository seems to be a ros package now, so it can be added.
The repository is still missing a `LICENSE.md` file.

Reopening for more visibility.

ivanpauno:
Oh, there's a new PR so I cannot reopen this one: https://github.com/ros/rosdistro/pull/26594/.

PR #26570, opened 2020-09-16
Add python3-distro

ruffsl:
* https://pkgs.org/download/python3-distro
* https://pypi.org/project/distro/

> It is the recommended replacement for Python's original platform.linux_distribution function (which will be removed in Python 3.8). It also provides much more functionality which isn't necessarily Python bound, like a command-line interface.


wjwwood (approved):
lgtm

mjcarroll (approved):

PR #26572, opened 2020-09-16
librealsense2: 2.38.1-1 in 'eloquent/distribution.yaml' [bloom]

doronhi:
Increasing version of package(s) in repository `librealsense2` to `2.38.1-1`:

- release repository: https://github.com/IntelRealSense/librealsense2-release.git
- bloom version: `0.9.8`

### DIFF
--- a/eloquent/distribution.yaml
+++ b/eloquent/distribution.yaml
@@ -1238,6 +1238,22 @@ repositories:
       url: https://github.com/ros2-gbp/libg2o-release.git
       version: 2019.11.23-4
     status: maintained
+  librealsense2:
+    doc:
+      type: git
+      url: https://github.com/IntelRealSense/librealsense.git
+      version: master
+    release:
+      tags:
+        release: release/eloquent/{package}/{version}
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

mjcarroll:
@doronhi This seems to be having trouble turning over in the farm,  can you take a look and address any issues?  Otherwise I will need to revert this until it builds.

doronhi:
> @doronhi This seems to be having trouble turning over in the farm, can you take a look and address any issues? Otherwise I will need to revert this until it builds.
> 
> Example: http://build.ros2.org/job/Edev__librealsense2__ubuntu_bionic_amd64/2/console

Sorry for failing to pay attention. This same package builds well for kinetic and melodic and I see now that some changes will have to take place. 
I see that the current error is the catkin dependency that does not fulfilled under eloquent but I can't see the command that is being called that creates the error.
Can you recommend a method for testing it before uploading a corrected version?

doronhi:
Fixed in #26805

PR #26574, opened 2020-09-16
added rosdep rule for fastnumbers pip package

niomate:
Added rosdep rule "python-fastnumbers-pip" for Arch, Debian, Fedora, OSX and Ubuntu. Passes all unit tests on my machine. Additionally ran "scripts/clean_rosdep_yaml.py" for as suggested by the tests.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5768,7 +5768,7 @@ python3-gql-pip:
 python3-grpc-tools:
   debian:
     '*': [python3-grpc-tools]
-    stretch:
+    stretch: null
   ubuntu:
     '*': [python3-grpc-tools]
     bionic: null
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -207,7 +207,7 @@ jupyter-nbconvert:
   gentoo: [dev-python/nbconvert]
   ubuntu:
     '*': [jupyter-nbconvert]
-    xenial: null
+    xenial:
 jupyter-notebook:
   debian:
     '*': [jupyter-notebook]
@@ -731,7 +731,7 @@ python-boto3:
   opensuse: [python-boto3]
   ubuntu:
     '*': [python-boto3]
-    trusty: null
+    trusty:
 python-bottle:
   debian: [python-bottle]
   fedora: [python-bottle]
@@ -1513,6 +1513,22 @@ python-fastdtw-pip:
   ubuntu:
     pip:
       packages: [python-fastdtw]
+python-fastnumbers-pip:
+  arch:
+    pip:
+      packages: [fastnumbers]
+  debian:
+    pip:
+      packages: [fastnumbers]
+  fedora:
+    pip:
+      packages: [fastnumbers]
+  osx:
+    pip:
+      packages: [fastnumbers]
+  ubuntu:
+    pip:
+      packages: [fastnumbers]
 python-fcl-pip:
   debian:
     pip:
@@ -2098,7 +2114,7 @@ python-imaging:
   arch: [python2-pillow]
   debian:
     '*': [python-pil]
-    jessie:  [python-imaging]
+    jessie: [python-imaging]
     stretch: [python-imaging]
   fedora: [python-pillow, python-pillow-qt]
   freebsd: [py27-pillow]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -207,7 +207,7 @@ jupyter-nbconvert:
   gentoo: [dev-python/nbconvert]
   ubuntu:
     '*': [jupyter-nbconvert]
-    xenial: null
+    xenial:
```
wjwwood, line 210: This change looks unrelated to your pull request. Please remove this part of the diff.

rosdep/python.yaml
```diff
@@ -731,7 +731,7 @@ python-boto3:
   opensuse: [python-boto3]
   ubuntu:
     '*': [python-boto3]
-    trusty: null
```
wjwwood, line 734: Same here.

wjwwood (changes requested):
There are a lot of changes that are unrelated to your new rosdep rule please remove those changes.

rosdep/python.yaml
Ramdambo: Oh yes, did not see that, must have been removed by clean_rosdep_yaml.py. I'll fix it!

rosdep/python.yaml
```diff
@@ -5676,7 +5692,7 @@ python3-flake8:
       packages: [flake8]
   rhel:
     '*': ['python%{python3_pkgversion}-flake8']
-    '7': null
+    '7':
```
wjwwood, line 5695: There are still some unwanted changes here. Even if I were to agree with some of them, they should be done as part of another pull request. Please get the diff down to just your new stanza for the new rosdep rule, thanks :)

tfoote: We specifically use null to indicate that it's unmet. 

From: https://yaml.org/type/null.html
> Note that a null is different from an empty string and that a mapping entry with some key and a null value is valid and different from not having that key in the mapping.

tfoote (changes requested):

rosdep/python.yaml
Ramdambo: Ah ok. Maybe the clean_rosdep_yaml script should be updated then ? Because when I ran it, it removed all the null values.

Ramdambo:
Should be all reverted now again.

rosdep/python.yaml
clalancette: > Ah ok. Maybe the clean_rosdep_yaml script should be updated then ? Because when I ran it, it removed all the null values. Or are you actually not supposed to run this script ? Because an error message in the tests suggested to run it.

Ah, that script may be out-of-date.  I don't know that its been updated to handle the `null` values; those are somewhat new.  If you'd like to contribute a PR to update it, we'd be happy to review!

rosdep/python.yaml
```diff
@@ -2098,7 +2114,7 @@ python-imaging:
   arch: [python2-pillow]
   debian:
     '*': [python-pil]
-    jessie:  [python-imaging]
+    jessie: [python-imaging]
```
clalancette, line 2117: This change (and the ones below) still don't belong in this PR.  I actually think they are valid fixes, but please submit them separately.  This PR should just be focused on the `fastnumbers` package.

clalancette (changes requested):

rosdep/python.yaml
Ramdambo: Ok, I'll have a look once I get the chance!

rosdep/python.yaml
```diff
@@ -1513,6 +1513,22 @@ python-fastdtw-pip:
   ubuntu:
     pip:
       packages: [python-fastdtw]
+python-fastnumbers-pip:
```
clalancette, line 1516: Just to be clear; this is implicitly a package for Python 2.  Are you using it for Python 2?  If so, this is OK as-is.  If not, I'd prefer to make this:

```
python3-fastnumbers-pip
```

Ramdambo: Good point, I oversaw that completely. I was using it with Python 2.7 and it seems to work just fine, but I double checked the documentation on PyPi because I was not completely sure anymore whether 2.7 is actually supported and turns out it's for versions >= 3.5.
I'll update it ASAP !

rosdep/python.yaml
```diff
@@ -1513,6 +1513,22 @@ python-fastdtw-pip:
   ubuntu:
     pip:
       packages: [python-fastdtw]
+python3-fastnumbers-pip:
```
clalancette, line 1516: This is failing CI because the whole section has to move down into the `python3-` section.

Ramdambo: Oooooh, I was wondering why it failed. Sorry, I overread the section on `python-` vs `python3-` placements.

jacobperron (approved):

ivanpauno (approved):

PR #26579, opened 2020-09-16
Add moving_average

VictorLamoine:
Compute and publish moving average values of ROS topic values.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6027,6 +6027,12 @@ repositories:
       url: https://github.com/peci1/movie_publisher.git
       version: melodic-devel
     status: developed
+  moving_average:
+    doc:
+      type: git
+      url: https://gitlab.com/InstitutMaupertuis/moving_average.git
+      version: melodic
+    status: maintained
   mpc_local_planner:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -6027,6 +6027,12 @@ repositories:
       url: https://github.com/peci1/movie_publisher.git
       version: melodic-devel
     status: developed
+  moving_average:
```
mjcarroll, line 6030: `moving_average` is a generic name, is there something more specific that it can be named?

melodic/distribution.yaml
```diff
@@ -6027,6 +6027,12 @@ repositories:
...
+    doc:
+      type: git
+      url: https://gitlab.com/InstitutMaupertuis/moving_average.git
+      version: melodic
+    status: maintained
```
mjcarroll, line 6035: If you intended to release the package, this is not sufficient, this will only add it to the documentation index.

If you need help getting a package released, follow the [bloom first time release guide](http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease)

mjcarroll (changes requested):

melodic/distribution.yaml
VictorLamoine: I don't intend to release the package, my goal is just to make it indexed to the documentation

melodic/distribution.yaml
VictorLamoine: Any suggestion? For a package that computes moving averages I don't see what could be a better name.

PR #26594, opened 2020-09-18
Adding orb_slam2_with_maps-odom to documentation index for distro

SuperAruba:
Adding orb_slam2_with_maps-odom to documentation index for distro

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -9532,7 +9532,7 @@ repositories:
       url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
       version: master
     status: maintained
-  orb_slam2_with_maps-odom:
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -9514,6 +9514,16 @@ repositories:
       url: https://github.com/appliedAI-Initiative/orb_slam_2_ros.git
       version: master
     status: maintained
+  orb_slam2_with_maps-odom:
```
wjwwood, line 9517: Using `-` in a repository name is discouraged, would you consider renaming it?

SuperAruba: OK,I will change it.

SuperAruba:
I have changed what "-" to "_".

wjwwood (approved):

PR #26619, opened 2020-09-21
python3-catkin-lint is valid for debian buster

erelson:
Looks like the package exists in debian buster:
https://packages.debian.org/buster/python3-catkin-lint

Didn't see anything specific about why it was entered as null initially in June: https://github.com/ros/rosdistro/pull/25055; probably we just didn't look further back than bullseye?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5483,7 +5483,6 @@ python3-can:
 python3-catkin-lint:
   debian:
     '*': [python3-catkin-lint]
-    buster: null
     jessie: null
     stretch: null
   fedora: [python3-catkin_lint]


wjwwood:
My guess if that it is an "upstream" version and we have a newer version in packages.ros.org, see: https://wiki.ros.org/UpstreamPackages

I'll close this under that assumption, but please let me know if you think this is not the case.

erelson:
Hmmm, this is evidently outside stuff I've messed with before, so I remain a bit confused.  Are you saying there might be a (newer than debian-buster) verison of python3-catkin-lint hosted at packages.ros.org? I don't think that's the case? http://packages.ros.org/ros/ubuntu/pool/main/p/

To clarify, the thing that led me to proposing this was that `bloom` complained about this missing dependency for the debian-buster build when I was attempting to release a package for ROS noetic.  Ultimately, since I don't forsee the package I'm releasing being used with non-Ubuntu installs, I certainly can just tell bloom to skip the debian-buster build. ... does that make sense?

erelson:
@wjwwood Follow-up question: As a result of this dep not being listed for debian-buster, bloom did not push a branch for debian-buster to the fetch_tools GBP repo.  None-the-less, the build farm is now trying to build a debian-buster version of the package, and failing.  What's the right thing to do here? I haven't come across any indication of a way to blacklist building for some distros.
http://build.ros.org/job/Nsrc_dB__fetch_tools__debian_buster__source/9/console

wjwwood:
Sorry, I my assumption was wrong I guess, since we don't have it in packages.ros.org, but that's an uncommon way to distribute python packages for us. Maybe the maintainers of catkin lint will know for sure.

It is in debian buster, so maybe we should just reopen this:

https://packages.debian.org/buster/python3-catkin-lint

The current rosdistro folks can have a look at this a second time.

wjwwood (approved):

clalancette (approved):
Since this does seem to be available in buster, I think it makes sense to make this change.

One other note: it *is* possible to prvent packages from building on certain platforms.  To do that, you have to open a PR to https://github.com/ros-infrastructure/ros_buildfarm_config/blob/3214a8e896e8dc571f9b652bfd0e6e65c624516c/noetic/release-buster-build.yaml#L16 , for instance.  But I kind of view that as a last result, and if we can get things building on all platforms, we should strive to do that.

PR #26620, opened 2020-09-21
adding jlb_pid: 0.1.2 to 'foxy/distribution.yaml'

Larsbl00:
Added a PID control node to `foxy/distribution.yaml`
- repository: <https://gitlab.com/Juulbl/ros2-pid>
- distro file: foxy/distribution.yaml

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1120,7 +1120,7 @@ repositories:
     doc:
       type: git
       url: https://gitlab.com/Juulbl/ros2-pid
-      version: 0.1.2
+      version: ros2
     status: maintained
   joint_state_publisher:
     doc:
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1733,7 +1733,7 @@ repositories:
       type: git
       url: https://gitlab.com/Juulbl/ros2-pid
       version: 0.1.0
-      status: developed
+      status: maintained
   plotjuggler:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -1728,6 +1728,12 @@ repositories:
       url: https://github.com/ros-drivers/phidgets_drivers.git
       version: dashing
     status: maintained
+  pid:
```
wjwwood, line 1731: @Larsbl00 this is a very generic name for a repository or package. Are you sure it's universally applicable to the community?

Also there is already a `pid` package in the ROS ecosystem, so you might consider changing the name to avoid collisions:

https://index.ros.org/p/pid/bitbucket-AndyZe-pid/

Larsbl00: The package is universally applicable as it simplifies    the implementation of a PID controller for ROS 2.

Unlike the `pid` package present in the ecosystem, this one is built in ROS 2, which means you do not need to setup the ROS 1 bridge.

As for the name, I was thinking about using either `ros2-pid` or `pid2`, I do not know which one would be better. Any preference?

wjwwood: Ok, well unless you're really set on the `pid` name, leaving it for the other package to at some point migrate to ROS 2 under the same name as it has in ROS 1 would be nice.

As for the name of your package, it's nice if you can include your name, or your organization/universities name in the package name. That gives it something slightly more identifiable than pid2 vs pid3 etc...

That being said, `ros2_pid` would be ok, but we recommend avoiding `ros` or `ros2` in the name, see:

https://www.ros.org/reps/rep-0144.html#global-rules

Also, if I had been aware of the `pid` package from ROS 1, I would have recommended the same name change for them, but sometimes they slip by. I'm not just giving you a hard time for no reason. <3

wjwwood (approved):

PR #26655, opened 2020-09-22
computer_status_msgs: 2.0.0 in {kinetic, melodic, noetic}/distribution.yaml [non-bloom]

130s:
Initial release of package(s) in repository `computer_status_msgs` to `2.0.0-2`:
- upstream repository: https://github.com/plusone-robotics/computer_status_msgs.git
- release repository: https://github.com/130s/computer_status_msgs-release.git
- distro file: `{kinetic, melodic, noetic}/distribution.yaml`
- bloom version: `0.9.8`  (bloom failed to open a pull request to rosdistro)
- previous version for package: None

Related discussion https://github.com/PR2/pr2_common/issues/286#issuecomment-695975839

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2078,6 +2078,22 @@ repositories:
       url: https://github.com/ros/common_tutorials.git
       version: indigo-devel
     status: maintained
+  computer_status_msgs:
+    doc:
+      type: git
+      url: https://github.com/plusone-robotics/computer_status_msgs.git
+      version: master
+    release:
+      tags:
+        release: release/kinetic/{package}/{version}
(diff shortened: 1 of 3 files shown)


tfoote (commented):
In the future please submit these as separate pull requests for each distro, just like bloom would have made. Spanning PRs across multiple distros means that you have to fit into the sync schedule of the union of all distros. You have luckily hit a window right now just before the kinetic sync preparation window. And if there's an issue that needs reverting for a specific distro we can no longer can simply revert the whole PR but need to get much more granular and communicate between the different release managers much more carefully and look at partially reverting etc which gets harder to track.

130s:
That makes sense, I should've thought about that :facepalm:

PR #26656, opened 2020-09-23
fetch_tools: 0.3.1-1 in 'noetic/distribution.yaml' [bloom]

erelson:
Increasing version of package(s) in repository `fetch_tools` to `0.3.1-1`:

- upstream repository: https://github.com/fetchrobotics/fetch_tools.git
- release repository: https://github.com/fetchrobotics-gbp/fetch_tools-release.git
- bloom version: `0.9.8`

## fetch_tools

```
* Update dependencies from py2 to py3
* Contributors: Eric Relson
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1023,6 +1023,17 @@ repositories:
         release: release/noetic/{package}/{version}
       url: https://github.com/wxmerkt/fcl_catkin-release.git
       version: 0.6.1-1
+  fetch_tools:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/fetchrobotics-gbp/fetch_tools-release.git
+      version: 0.3.1-1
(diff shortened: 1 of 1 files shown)


clalancette (commented):
Two questions.  Neither is blocking.

clalancette:
> @clalancette thanks for checking in! I'm fairly new at blooming packages, and didn't think to check what the previous releases did for the various questions asked by bloom. In this case, fetch_tools doesn't have any tests or documentation (since it's mostly just some bash and python workflow tools written years ago). Does it generally make sense to create the doc/test jobs anyways?

So if you don't have tests, then there is really no reason to have `test_pull_requests: true` on.  So I'll suggest skipping that one.

If you don't have any documentation (API or otherwise), then it is also fine to skip the doc job.

With both of those said, then, I'll go ahead and merge this.  Thanks for responding.

PR #26671, opened 2020-09-23
Remove cross-compile

shr-project:


clalancette:
Can you please explain why you want to remove this one?

shr-project:
> Can you please explain why you want to remove this one?

The restored snapshot of cross_compile-release seems to have only master branch and no tags. rosdistro_build_cache is failing because of that, see
https://github.com/ros-tooling/cross_compile/issues/248#issuecomment-678384795

@emersonknapp said about this:
> Thanks for bringing this up - this was deleted in error as part of an attempt to clean up unused repositories (as cross_compile is not a ros package and does not release to rosdistro any more). We've contacted GitHub support to recover the repository, I'll keep this ticket updated.

emersonknapp:
I'm fine with removing this - the tool is perfectly usable for dashing and eloquent without relying on the rosdistro ecosystem, it's a pure python ros-external package like colcon.

clalancette:
> I'm fine with removing this - the tool is perfectly usable for dashing and eloquent without relying on the rosdistro ecosystem, it's a pure python ros-external package like colcon.

Just to be clear; if we remove this, then the package will no longer be available to install via `apt-get install ros-<distro>-cross-compile`.  That means that new users won't be able to install it.  People who already have it installed will still have it, but will get no more updates this way.

Assuming you are OK with all of those constraints, its fine by me to remove it.  I just wanted to make sure you are aware.

emersonknapp:
Yes, I think that is better actually. Users should get the latest version of the tool via pip, they'll get new improvements that way whereas these apt versions for dashing and eloquent are way less capable old versions of the tool that will not be updated.

PR #26683, opened 2020-09-24
add key for python-twilio-pip

fmessmer:
adding `rosdep` key for `twilio` 
:point_right: https://pypi.org/project/twilio/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4915,6 +4915,16 @@ python-trimesh-pip:
   ubuntu:
     pip:
       packages: [trimesh]
+python-twilio-pip:
+  debian:
+    pip:
+      packages: [twilio]
+  fedora:
+    pip:
+      packages: [twilio]
+  ubuntu:
+    pip:
+      packages: [twilio]
 python-twisted-bin:
   arch: [python2-twisted]
   debian: [python-twisted-bin]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4915,16 +4915,24 @@ python-trimesh-pip:
   ubuntu:
     pip:
       packages: [trimesh]
-python-twilio-pip:
+python-twilio:
   debian:
-    pip:
-      packages: [twilio]
-  fedora:
-    pip:
-      packages: [twilio]
+    buster: [python-twilio]
+    jessie:
+      pip:
+        packages: [twilio]
+    stretch:
+      pip:
+        packages: [twilio]
+  fedora: [python3-twilio]
   ubuntu:
-    pip:
-      packages: [twilio]
+    '*': [python3-twilio]
+    trusty:
+      pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -4915,6 +4915,16 @@ python-trimesh-pip:
   ubuntu:
     pip:
       packages: [trimesh]
+python-twilio-pip:
```
clalancette, line 4918: Looking around, it looks like there is a system package available for all of [Debian](https://packages.debian.org/search?keywords=twilio&searchon=names&suite=all&section=all), [Ubuntu](https://packages.ubuntu.com/search?keywords=twilio&searchon=names&suite=all&section=all), and [Fedora](https://fedora.pkgs.org/31/fedora-x86_64/python3-twilio-6.29.1-1.fc31.noarch.rpm.html).  Since we prefer system packages over pip packages, can you change the key to use the system packages instead?  Thanks.

clalancette (changes requested):

fmessmer:
@clalancette 
hope I added the system dependencies to the versions where they are available in the correct way
I kept the `pip` version for the other distros

rosdep/python.yaml
```diff
@@ -4915,6 +4915,24 @@ python-trimesh-pip:
   ubuntu:
     pip:
       packages: [trimesh]
+python-twilio:
```
clalancette, line 4918: The way this key is currently arranged, it is a mixture between Python 2 and Python 3.  Let's instead just make the key for python3, so make this line:

```
python3-twilio:
```

rosdep/python.yaml
```diff
@@ -4915,6 +4915,24 @@ python-trimesh-pip:
...
+        packages: [twilio]
+    stretch:
+      pip:
+        packages: [twilio]
+  fedora: [python3-twilio]
+  ubuntu:
+    '*': [python3-twilio]
+    trusty:
+      pip:
+        packages: [twilio]
```
clalancette, line 4932: Let's not bother with trusty; it isn't supported anymore.

rosdep/python.yaml
```diff
@@ -4915,6 +4915,24 @@ python-trimesh-pip:
...
       packages: [trimesh]
+python-twilio:
+  debian:
+    buster: [python-twilio]
+    jessie:
+      pip:
+        packages: [twilio]
+    stretch:
+      pip:
+        packages: [twilio]
```
clalancette, line 4926: Let's not bother with jessie or stretch here, as they are both unsupported: https://wiki.debian.org/LTS

clalancette (changes requested):

fmessmer:
how about this then? @clalancette

jacobperron (approved):

clalancette (approved):

PR #26692, opened 2020-09-25
Adding rosaic to documentation index for melodic

septentrio-users:
Would be great if rosaic could be indexed and documented on ros.org.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -9726,6 +9726,12 @@ repositories:
       url: https://github.com/facontidavide/ros_type_introspection.git
       version: master
     status: developed
+  rosaic:
+    doc:
+      type: git
+      url: https://github.com/septentrio-gnss/rosaic
+      version: master
+    status: maintained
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -9726,6 +9726,12 @@ repositories:
       url: https://github.com/facontidavide/ros_type_introspection.git
       version: master
     status: developed
+  rosaic:
+    doc:
+      type: git
+      url: https://github.com/septentrio-gnss/rosaic
```
clalancette, line 9732: This is failing our CI because the URL needs to end with `.git`.

clalancette (changes requested):

melodic/distribution.yaml
septentrio-users: Many thanks for pointing this out. I will append the .git momentarily.

mabelzhang:
Closing in favor of #26701

PR #26694, opened 2020-09-25
[dep] Add AWS CLI

130s:
Rationale
- arch https://www.archlinux.org/packages/community/any/aws-cli/
- debian https://packages.debian.org/jessie/awscli
- fedora https://src.fedoraproject.org/rpms/awscli
- gentoo https://packages.gentoo.org/packages/dev-python/awscli
- Ubuntu xenial onward https://packages.ubuntu.com/xenial/admin/awscli

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -225,6 +225,12 @@ avrdude:
   fedora: [avrdude]
   gentoo: [dev-embedded/avrdude]
   ubuntu: [avrdude]
+awscli:
+  arch: [aws-cli]
+  debian: [awscli]
+  fedora: [awscli]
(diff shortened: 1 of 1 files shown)


clalancette (approved):

mabelzhang (approved):

130s:
Btw, I just ran into an issue with a PyPi package called `aws`, NOT `awscli` that I suggested to add a rosdep key in this PR. Looks like it's a very common issue https://github.com/pypa/pypi-support/issues/614. If I were relying on `rosdep` to install pkgs, I would have not run into an issue in installing a wrong pkg. Another reason for me to recommend `rosdep` for defining software dependency.

PR #26699, opened 2020-09-25
[dep] Add wireshark

130s:
wireshark 
- arch https://wiki.archlinux.org/index.php/Wireshark says qt and cli are separate package, while those in other distros likely bundle both gui and cli. So I'm not sure what to do with arch, hence not added.
- debian jessie https://packages.debian.org/jessie/wireshark
- fedora https://src.fedoraproject.org/rpms/wireshark
- gentoo https://packages.gentoo.org/packages/net-analyzer/wireshark
- Ubuntu xenial https://packages.ubuntu.com/search?keywords=wireshark

wireshark common module
- fedora https://fedora.pkgs.org/30/fedora-x86_64/wireshark-cli-3.0.1-1.fc30.x86_64.rpm.html  (Couldn't find any on fedoraproject.org)
- debian.org/jessie https://packages.debian.org/jessie/wireshark-common
- Ubuntu https://packages.ubuntu.com/xenial/net/wireshark-common

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6301,6 +6301,10 @@ wireshark:
   fedora: [wireshark]
   gentoo: [net-analyzer/wireshark]
   ubuntu: [wireshark]
+wireshark-common:
+  debian: [wireshark-common]
+  fedora: [wireshark-cli]
+  ubuntu: [wireshark-common]
 wkhtmltopdf:
   arch: [wkhtmltopdf]
   debian: [wkhtmltopdf]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6296,6 +6296,11 @@ wireless-tools:
   fedora: [wireless-tools]
   gentoo: [net-wireless/wireless-tools]
   ubuntu: [wireless-tools]
+wireshark:
+  debian: [wireshark]
+  fedora: [wireshark]
+  gentoo: [net-analyzer/wireshark]
+  ubuntu: [wireshark]
 wkhtmltopdf:
   arch: [wkhtmltopdf]
   debian: [wkhtmltopdf]


rosdep/base.yaml
```diff
@@ -6290,6 +6290,11 @@ wireless-tools:
   fedora: [wireless-tools]
   gentoo: [net-wireless/wireless-tools]
   ubuntu: [wireless-tools]
+wireshark:
+  debian: [wireless-tools]
+  fedora: [wireless-tools]
```
clalancette, line 6295: These both look like a typo; I think they should probably be `wireshark` instead.

clalancette (changes requested):
There are some typos inline.

Besides that, @130s, are you interested in the packages for CLI access, or for GUI access?  It looks like at least on Debian, Ubuntu, and Fedora, there are separate packages for CLI and GUI.  On Debian and Ubuntu, it is `wireshark-common`.  On Fedora, it is `wireshark-cli`.

130s:
> Besides that, @130s, are you interested in the packages for CLI access, or for GUI access? It looks like at least on Debian, Ubuntu, and Fedora, there are separate packages for CLI and GUI. On Debian and Ubuntu, it is `wireshark-common`. On Fedora, it is `wireshark-cli`.

Thanks for pointing that out. In my case we likely need GUI + whatever backend, but it makes sense to separately provide dependency. So I added a commit for the common module.

wireshark common module
- fedora https://fedora.pkgs.org/30/fedora-x86_64/wireshark-cli-3.0.1-1.fc30.x86_64.rpm.html  (Couldn't find any on fedoraproject.org)
- debian.org/jessie https://packages.debian.org/jessie/wireshark-common
- Ubuntu https://packages.ubuntu.com/xenial/net/wireshark-common

clalancette (approved):

PR #26706, opened 2020-09-26
fcl: 0.6.1-1 in 'noetic/distribution.yaml' [bloom]

rhaschke:
Increasing version of package(s) in repository `fcl` to `0.6.1-1`:

- upstream repository: https://github.com/flexible-collision-library/fcl.git
- release repository: https://github.com/ros-gbp/fcl-release.git
- bloom version: `0.9.8`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1033,6 +1033,17 @@ repositories:
       url: https://github.com/iron-ox/fadecandy_ros.git
       version: master
     status: developed
+  fcl:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/ros-gbp/fcl-release.git
+      version: 0.6.1-1
+    source:
+      type: git
+      url: https://github.com/flexible-collision-library/fcl.git
+      version: master
+    status: maintained
   fcl_catkin:
     release:
       tags:


rhaschke:
This addresses https://github.com/ros/rosdistro/issues/26527 by releasing `fcl` as a ROS package as suggested in https://github.com/ros/rosdistro/issues/26527#issuecomment-698994675.

mabelzhang:
I read the discussion there but couldn't decide whether the discussion is finished and final.
I'll let @sloretz or someone more familiar with the matter merge this PR.

ivanpauno (approved):
LGTM!
I would like @sloretz to confirm if this PR resolves https://github.com/ros/rosdistro/issues/26527 correctly.

rhaschke:
@sloretz, friendly ping: Is there anything blocking this release? Do you have objections?

sloretz:
> @sloretz, friendly ping: Is there anything blocking this release? Do you have objections?

Nothing blocking, just on my TODO list to look closely at before the next sync.

tfoote:
Since the sync just came out. I'm going to merge this so we can get testing of it in the majority of this cycle.

sloretz:
It looks like there were some build failures on armhf and aarch64, [which appears to have caused MoveIt to not build](http://build.ros.org/view/Nbin_ufhf_uFhf/job/Nbin_ufhf_uFhf__moveit_core__ubuntu_focal_armhf__binary/). I opened a PR upstream: https://github.com/flexible-collision-library/fcl/pull/506

peci1:
@rhaschke I understand the need for newer (and the same) versions of fcl on noetic. However, the situation it created is pretty non-straightforward for people not seeing into the reasons why there are these two different versions.

See e.g. https://github.com/flexible-collision-library/fcl/issues/533 where I got pretty confused :)

So, to prevent this confusion, I suggest adding a few lines to http://wiki.ros.org/noetic/Migration . I think this definitely is a thing people migrating from Melodic need to take special care of. Ideally, with examples what CMake commands should be used to always get the system version, and what commands should be used to always get the ROS version. Because if you do it "somehow", you can pretty easily get a mixed version that doesn't work and is hard to debug. I'm sorry, but I don't have the knowledge to write these lines myself...

rhaschke:
@peci1, I'm not sure how you managed to get a mix of two different versions. There are no special cmake commands to get one or the other version. By default, overlaying `/opt/ros/noetic`, you should get the ROS version by default, shouldn't you?
The (only?) way to switch between the two versions is to change your `package.xml`:
```xml
  <depend condition="$ROS_DISTRO != noetic">libfcl-dev</depend>
  <depend condition="$ROS_DISTRO == noetic">fcl</depend>
```
I don't understand what exactly [failed on the buildfarm](https://build.ros.org/job/Ndev__robot_body_filter__ubuntu_focal_amd64/1/consoleFull#console-section-13). I think, you should first try to fully understand this, before we can provide some more hints.

peci1:
It's actually pretty easy to get a mix of versions.

This line in a package that found FCL 0.6 via CMake:

```
#include <fcl/BV/OBB.h>
```

will pull the 0.5 header, because the 0.6 header would be:

```
#include <fcl/math/bv/OBB.h>
```

If you assume the default system install location, the 0.5 headers will be in `/usr/include`, which is practically always on the include path.

That's why I say it is important to mention this in migration.

If a developer just plainly does the migration in the way "try to build it and hammer it until it builds", there might be surprises. Without the notice in Migration guide, the developer would leave the `libfcl-dev` dependency in package.xml. And the system will find FCL 0.5. UNLESS - some other (even unrelated) program installs ros-noetic-fcl. Then the program would find FCL 0.6 without the developer knowing why. And `apt show libfcl-dev` is still showing 0.5...

peci1:
I see two options for a developer:

1. keep depending on `libfcl-dev`... then the code has to be properly `ifdef`ed to work correctly with both 0.5 and 0.6
2. explicitly depend on `fcl` ROS package for Noetic... then 0.6 should be sure in all places

rhaschke:
Ok, the key issue here is that `fcl` changed its includes, such that you find wrong includes when both, system fcl 0.5 and ROS fcl 0.6 are installed. I suggest to always depend on fcl instead of libfcl-dev. Otherwise, combining any two libraries, that were linked against different fcl versions, in a downstream application will cause obscure issues. Hence, I suggest the following addition to [noetic/Migration](http://wiki.ros.org/action/login/noetic/Migration):

## Use `fcl` rosdep key instead of `libfcl-dev`

ROS Noetic provides FCL 0.6, which is ABI and API-incompatible to the system-installed FCL 0.5. To avoid any conflicts in downstream packages, when linking package libraries that are in turn linked against different FCL versions, we suggest to upgrade to FCL 0.6 and use the new rosdep key `fcl` instead of `libfcl-dev`.

If you agree, @peci1, please add this (or any augmented version) to the migration notes. Thanks.

peci1:
Thanks for the inspiration, added: http://wiki.ros.org/action/info/noetic/Migration?action=diff&rev2=9&rev1=8 .

PR #26711, opened 2020-09-26
update clean_rosdep_yaml script to represent null values properly

niomate:
Tried my best to update the clean script such that null values are not removed when the script is run. If the change in formatting is an issue, I can revert it easily.

### DIFF
--- a/scripts/clean_rosdep_yaml.py
+++ b/scripts/clean_rosdep_yaml.py
@@ -7,6 +7,7 @@ import io
 
 dont_bracket = ['uri', 'md5sum']
 
+
 def paddify(s, l):
     a = s.split('\n')
     buf = ''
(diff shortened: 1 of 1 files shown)


clalancette (approved):
This is definitely an improvement to the script.  As such, I'm going to approve and merge it.  Thanks for the contribution.

However, I will point out that this isn't sufficient to make this script work properly.  In particular, the script blindly removes quotes, which are actually needed in several cases in the rosdep base.yaml.  So this will still need follow-up.

clalancette:
> Do you know when quotes should be applied ?

I know that they are needed e.g. on this line: https://github.com/ros/rosdistro/blob/839827608314b7d2f9bb19a14ac508334dd053c4/rosdep/base.yaml#L60 .  At least one rule is that if they are already in a list, they should not be touched.  But I don't know a general rule on when they should and should not be applied.

Ramdambo:
Do you know where I could get more information ? Or should I just read through the yaml files and try to find a pattern if I want to contribute this ?

PR #26754, opened 2020-09-29
Added base rule for gphoto2 CLI under Ubuntu.

airballking:


jacobperron (commented):
Following our [contributing guide](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions), please add rules for:

- Debian: https://packages.debian.org/jessie/gphoto2
- Fedora: https://fedora.pkgs.org/30/fedora-x86_64/gphoto2-2.5.20-3.fc30.x86_64.rpm.html
- Gentoo: https://packages.gentoo.org/packages/media-gfx/gphoto2

It looks like we can add Arch too: https://www.archlinux.org/packages/extra/x86_64/gphoto2/

airballking:
@jacobperron Thank you for the review and suggested improvement! I added the rules for the other operating systems.

rosdep/base.yaml
```diff
@@ -1199,6 +1199,12 @@ gperftools:
   fedora: [gperftools, gperftools-devel]
   gentoo: [dev-util/gperf]
   ubuntu: [google-perftools, libgoogle-perftools-dev]
+gphoto2:
+  arch: [gphoto2]
+  debian: [gphoto2]
+  fedora: [gphoto2]
+  gentoo: [gphoto2]
```
jacobperron, line 1206: I think we need to include the category for Gentoo:
```suggestion
  gentoo: [media-gfx/gphoto2]
```

jacobperron (commented):
Thanks for adding the other platforms! One more comment to address.

jacobperron (approved):

ivanpauno (approved):

PR #26768, opened 2020-09-30
add python-progressbar for focal

floweisshardt:
- debian: https://packages.debian.org/en/stretch/python3-progressbar
- ubuntu: https://packages.ubuntu.com/focal/python3-progressbar

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6095,6 +6095,9 @@ python3-pkg-resources:
   openembedded: [python3-setuptools@openembedded-core]
   rhel: ['python%{python3_pkgversion}-setuptools']
   ubuntu: [python3-pkg-resources]
+python3-progressbar:
+  debian: [python3-progressbar]
+  ubuntu: [python3-progressbar]
 python3-prometheus-client:
   debian: [python3-prometheus-client]
   ubuntu: [python3-prometheus-client]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6096,7 +6096,10 @@ python3-pkg-resources:
   rhel: ['python%{python3_pkgversion}-setuptools']
   ubuntu: [python3-pkg-resources]
 python3-progressbar:
+  arch: [python-progressbar]
   debian: [python3-progressbar]
+  fedora: [python3-progressbar2]
+  gentoo: [dev-python/progressbar]
   ubuntu: [python3-progressbar]
 python3-prometheus-client:
   debian: [python3-prometheus-client]


rosdep/python.yaml
```diff
@@ -3160,6 +3160,7 @@ python-progressbar:
     artful: [python-progressbar]
     bionic: [python-progressbar]
     cosmic: [python-progressbar]
+    focal: [python3-progressbar]
```
jacobperron, line 3163: I think it would be better to add a new key for the python3 version of the package, [per our guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-3).

tfoote (changes requested):
Looking forward the python3 rule should be used. Please add that instead of putting python3 into the python2 definiitons.

floweisshardt:
@tfoote @jacobperron I added a new key `python3-progressbar` as you requested.

However that forces us to use different branches for our ROS package for kinetic/melodic (`python-progressbar`) and noetic (`python3-progressbar`) and not use the same source code version for all ROS distros anymore. Is this the recommended way or is there other best practices to avoid that?

tfoote:
You can use the new new conditional dependencies in package.xml format 3 to make your code dual compatible: https://www.ros.org/reps/rep-0149.html and ROS_PYTHON_VERSION

REP 151 lays out how to migrate to python 3: https://www.ros.org/reps/rep-0151.html#dependencies-and-package-xml

rosdep/python.yaml
```diff
@@ -6095,6 +6095,9 @@ python3-pkg-resources:
   openembedded: [python3-setuptools@openembedded-core]
   rhel: ['python%{python3_pkgversion}-setuptools']
   ubuntu: [python3-pkg-resources]
+python3-progressbar:
+  ubuntu:
+    focal: [python3-progressbar]
```
tfoote, line 6100: This package appears to be available from xenial and newer. We shouldn't specifically target focal. 

This also doesn't cover all our main platforms as outlined in the contributing guide: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

floweisshardt:
changed to generic key for ubuntu and added debian key.

rosdep/python.yaml
```diff
@@ -6095,6 +6095,9 @@ python3-pkg-resources:
   openembedded: [python3-setuptools@openembedded-core]
   rhel: ['python%{python3_pkgversion}-setuptools']
   ubuntu: [python3-pkg-resources]
+python3-progressbar:
+  debian: [python3-progressbar]
+  ubuntu: [python3-progressbar]
```
ivanpauno, line 6100: Can you add
- https://packages.gentoo.org/packages/dev-python/progressbar
- https://www.archlinux.org/packages/community/any/python-progressbar/
- https://fedora.pkgs.org/31/fedora-x86_64/python3-progressbar2-3.39.3-3.fc31.noarch.rpm.html
?

floweisshardt: gentoo, arch and fedora is added in https://github.com/ros/rosdistro/pull/26768/commits/b99dbfb2340b43f608b043b4eb260e3241701073

ivanpauno (approved):

tfoote (approved):

PR #26775, opened 2020-10-01
sot-core: 4.10.1-1 in 'melodic/distribution.yaml' [bloom]

Rascof:
Increasing version of package(s) in repository `sot-core` to `4.10.1-1`:

- upstream repository: https://github.com/stack-of-tasks/sot-core.git
- release repository: https://github.com/stack-of-tasks/sot-core-ros-release.git
- bloom version: `0.10.0`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -11403,6 +11403,22 @@ repositories:
       url: https://github.com/yujinrobot-release/sophus-release.git
       version: 1.0.1-1
     status: maintained
+  sot-core:
+    doc:
+      type: git
+      url: https://github.com/stack-of-tasks/sot-core.git
+      version: devel
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/stack-of-tasks/sot-core-ros-release.git
+      version: 4.10.1-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/stack-of-tasks/sot-core.git
+      version: devel
+    status: maintained
   sparse_bundle_adjustment:
     doc:
       type: git


Rascof:
The build failed due to a plugin from dynamic-graph that cannot be copied inside /opt/ros/melodic/lib/dynamic-graph-plugins
```
CMake Error at src/cmake_install.cmake:60 (file):
file INSTALL cannot copy file
"/tmp/ws/build_isolated/sot-core/install/src/libsot.so.4.10.1" to
"/opt/ros/melodic/lib/dynamic-graph-plugins/libsot.so.4.10.1".
Call Stack (most recent call first):
cmake_install.cmake:461 (include)
Makefile:88: recipe for target 'install' failed
make: *** [install] Error 1
Failed to process package 'sot-core': 
Command '['make', 'install']' returned non-zero exit status 2
Reproduce this error by running:
cd /tmp/ws/build_isolated/sot-core && make install
```
@wxmerkt Did you already have a similar error with your previous releases?
@jacobperron @clalancette Do you know a way to fix this error?

Running : `chmod 777 /opt/ros/melodic/lib/dynamic-graph-plugins` during the prerelease didn't fix the issue.

wxmerkt:
Hi @Rascof,
I haven't seen this kind of error with jrl-cmakemodules packages before (mostly focused on requiring patches for lib architecture prefixes). However from the output I note it's trying to copy a library from _src_ which seems odd?

nim65s:
Hello,

For this library, we basically have, in the `src` directory of the source tree, a `add_library(sot SHARED […])` and a `install(TARGETS sot DESTINATION ${DYNAMIC_GRAPH_PLUGINDIR})`.

So I guess the `install` directory is actually a build one… Isn't it ? With this setup, we usually have `build/src/*.so`, but never ever a `src` dir in the install prefix.

Anyway, the correct install path for this library is `/opt/ros/melodic/lib/dynamic-graph-plugins`, which seems good here, but maybe there is a system that prevent anything from being installed outside of the install prefix ? We need this for the plugin system.

jacobperron:
I don't know the details, but I think install destinations are meant to be relative to the package being built and infrastructure will relocate artifacts as necessary.

Instead, I suggest trying to install plugins to a directory, relative to the catkin install prefix, e.g.

```
install(TARGETS sot DESTINATION "${CATKIN_PACKAGE_LIB_DESTINATION}/dynamic-graph-plugins"
```

nim65s:
Thanks @jacobperron, I think we can try that. We just need to know that unit tests won't work until the relocation.

PR #26778, opened 2020-10-01
add python3-docopt

jschleicher:
argument parser available as system package at least on debian and ubuntu

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5636,6 +5636,9 @@ python3-docker:
   rhel:
     '7': ['python%{python3_pkgversion}-docker']
   ubuntu: [python3-docker]
+python3-docopt:
+  debian: [python3-docopt]
+  ubuntu: [python3-docopt]
 python3-docutils:
   arch: [python-docutils]
   debian: [python3-docutils]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5636,6 +5636,9 @@ python3-docker:
   rhel:
     '7': ['python%{python3_pkgversion}-docker']
   ubuntu: [python3-docker]
+python3-docopt:
+  debian: [python3-docopt]
```
jacobperron, line 5640: Please add rules for [Fedora](https://fedora.pkgs.org/32/fedora-x86_64/python3-docopt-0.6.2-16.fc32.noarch.rpm.html), [Gentoo](https://packages.gentoo.org/packages/dev-python/docopt) and [Arch](https://www.archlinux.org/packages/community/any/python-docopt/).

tfoote (changes requested):
Please add the rules that Jacob did the additional research to find.

rosdep/python.yaml
jschleicher: Thank you for the links!

ivanpauno (approved):

PR #26795, opened 2020-10-02
python-matplotlib for focal

fmessmer:
collapsing it like this will make the key available for focal


jacobperron (commented):
I don't see a version of `python-matplotlib` for Focal: https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python-matplotlib&searchon=names

There is a Python 3 version of the package (https://packages.ubuntu.com/focal/python3-matplotlib), which is [already indexed](https://github.com/ros/rosdistro/blob/cad58b7f8fb36cfed02a3b6858db3790c2c73541/rosdep/python.yaml#L5917).

fmessmer:
right, sorry about that...will be using conditional `depend` then in my `package.xml`...closing

PR #26801, opened 2020-10-03
Update repo URLs for ros2_tracing & tracetools_analysis

christophebedard:
See https://github.com/ros2/ros2/pull/1045

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -2622,7 +2622,7 @@ repositories:
   ros2_tracing:
     doc:
       type: git
-      url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing.git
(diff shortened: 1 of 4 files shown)


jacobperron (commented):
Since the same release repo is being used for `tracetools_analysis` in rolling, we should also update the URLs in the [tracks.yaml](https://github.com/ros2-gbp/tracetools_analysis-release/blob/master/tracks.yaml).

christophebedard:
oops, I forgot about that one. I created a new release and changed the release repo back to the one on gitlab.com (with the current gitlab link for now): https://github.com/ros/rosdistro/pull/26927

I'll then rebase this branch on that one and update the repo URL.

christophebedard:
The repos have been moved, so this should be ready.

clalancette (approved):
All right, this looks like everything got moved, so I'm merging.

PR #26814, opened 2020-10-06
base: add libtiff-dev for focal

russkel:
It seems the focal release was missing for libtiff-dev.

Relevant package can be found here:
https://ubuntu.pkgs.org/20.04/ubuntu-main-amd64/libtiff-dev_4.1.0+git191117-2build1_amd64.deb.html

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4089,6 +4089,7 @@ libtiff-dev:
   ubuntu:
     artful: [libtiff5-dev]
     bionic: [libtiff5-dev]
+    focal: [libtiff-dev]
     lucid: [libtiff4-dev]
     maverick: [libtiff4-dev]
     natty: [libtiff4-dev]


ahcorde (approved):

rosdep/base.yaml
```diff
@@ -4090,6 +4090,7 @@ libtiff-dev:
     artful: [libtiff5-dev]
     bionic: [libtiff5-dev]
     lucid: [libtiff4-dev]
+    focal: [libtiff-dev]
```
ivanpauno, line 4093: libtiff5-dev?

That was the one using in bionic, so I don't think we should pick an older version in focal.

ivanpauno (changes requested):

rosdep/base.yaml
jacobperron: I think libtiff5-dev is a transitional package; i.e. it's a dummy package that points to libtiff-dev: https://packages.ubuntu.com/focal/libtiff5-dev

ivanpauno: good point

ivanpauno (approved):

rosdep/base.yaml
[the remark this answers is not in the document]
russkel: Sigh. Appears I don't know my alphabet.

PR #26886, opened 2020-10-09
Adding shadow robot repos to documentation index for distro kinetic and melodic

georgiablanco:
I'd like shadow robot repos to be indexed and documented on ros.org.


kinetic/distribution.yaml
```diff
@@ -15150,6 +15150,151 @@ repositories:
...
+      url: https://github.com/shadow-robot/sr_common.git
+      version: kinetic-devel
+    release:
+      packages:
+      - sr_common
+      - sr_description
+      - sr_robot_msgs
+      tags:
+        release: release/kinetic/{package}/{version}
+      url: https://github.com/shadow-robot/sr_common.git
```
tfoote, line 15165: The release repository should be a release repository not the upstream source repository.

kinetic/distribution.yaml
```diff
@@ -15150,6 +15150,151 @@ repositories:
...
+      version: kinetic-devel
```
tfoote, line 15166: This should be the released version number.

tfoote (changes requested):
If you're just doing indexing and doc please skip the release stanzas in this PR. They will be generated by bloom at the time of release with all the correct information. 

I gave specific info on one of them as to things that are technically incorrect. But you'll want to go through this tutorial: http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease carefully to make the releases.

tfoote:
One other suggestion is to separate PRs for different distros. They have different release cycles and it's easier to process them if they're separate.

tfoote:
Replaced by #26918 and #26919 

In the future please close, superceded requests like this and reference them either forwards or backwards so we don't need to keep track of them.

PR #26908, opened 2020-10-11
Adding rosaic to documentation index for noetic

septentrio-users:


tfoote (changes requested):
Reviewing the package quickly it doesn't appear to meet our guidelines of not using ros in the name: https://www.ros.org/reps/rep-0144.html And I can't even understand what aic is after skimming through the overview docs which suggests it's less useful in the descriptive department. To that end I'd like to request that you rename this before indexing it and spreading it wider. After that happens it's much harder to rename things.

hidmic:
@septentrio-users friendly ping

septentrio-users:
Thank you for pointing out that "ros" cannot be included in the repo name. After renaming to "septentrio_gnss_driver" I have made a new pull request. Cheers

clalancette:
With #26976 merged, I'm going to close this out.  Thanks for iterating.

tfoote:
Thanks for updating. That's a much clearer name!

PR #26913, opened 2020-10-12
add xmlschema to rosdistro

hakuturu583:
I want to add xmlschema to rosdistro.

Pypy package: https://pypi.org/project/xmlschema/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5270,6 +5270,16 @@ python-xmlplain-pip:
   ubuntu:
     pip:
       packages: [xmlplain]
+python-xmlschema:
+  debian:
+    pip:
+      packages: [xmlschema]
+  fedora:
+    pip:
+      packages: [xmlschema]
+  ubuntu:
+    pip:
+      packages: [xmlschema]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5270,6 +5270,16 @@ python-xmlplain-pip:
   ubuntu:
     pip:
       packages: [xmlplain]
+python-xmlschema:
```
tfoote, line 5273: If this is a pip only package it should have the `-pip` suffix. 

And while we're at it these days most platforms are on python3 now so we should make this rule more future consistent using python3 in the key as pip will choose the local environment anyway.

```suggestion
python3-xmlschema-pip:
```

tfoote (changes requested):
Looking at this we generally prefer to use mature natively packaged libraries so that they can be released and used in packages. Pip entries are only useful for from source builds. Is there any more formally packaged alternative?

hidmic:
@hakuturu583 friendly ping

dirk-thomas:
Closing to due no response from the author @hakuturu583.

PR #26919, opened 2020-10-12
Adding shadow robot repos to documentation index for distro melodic

georgiablanco:
I'd like shadow robot repos to be indexed and documented on ros.org.


georgiablanco:
when do you think this will merged?

PR #26998, opened 2020-10-16
node_interface_test: 0.1.0-1 in 'melodic/distribution.yaml' [bloom]

aremazeilles:
Increasing version of package(s) in repository `node_interface_test` to `0.1.0-1`:

- upstream repository: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test.git
- release repository: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test-release.git
- bloom version: `0.10.0`

## node_interface_test

```
* change package name
* Contributors: Anthony Remazeilles
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6923,7 +6923,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test.git
-      version: 0.1.0
+      version: melodic-devel
     release:
       tags:
         release: release/melodic/{package}/{version}
@@ -6932,7 +6932,7 @@ repositories:
     source:
       type: git
       url: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test.git
-      version: 0.1.0
+      version: melodic-devel
     status: maintained
   nodelet_core:
     doc:
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6919,6 +6919,21 @@ repositories:
       url: https://github.com/OUXT-Polaris/nmea_to_geopose.git
       version: master
     status: developed
+  node_interface_test:
+    doc:
+      type: git
+      url: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test.git
+      version: 0.1.0
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test-release.git
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test.git
+      version: 0.1.0
+    status: maintained
   nodelet_core:
     doc:
       type: git


hidmic (changes requested):
@aremazeilles check the failing job

aremazeilles:
I will check Monday with colleagues, the error is not clear to me. 
Could it be because I stated `0.1.0` for the source->version tag?

```yaml
 source:
      type: git
      url: https://github.com/tecnalia-advancedmanufacturing-robotics/node_interface_test.git
      version: 0.1.0
```
comparing with the other packages defined in `distribution.yaml`, it seems the others place there what I believe to be the branch name, and not a particular tag.
Same happens actually for the doc part.

clalancette:
I also have to say that the package name is very generic.  There's no real indication of what it does and what it is for.  Think about trying to install `ros-melodic-node-interface-test`; I think it would be hard for the user to have any idea what that is for.

I'll suggest a different name that more clearly indicates what it is for.

clalancette:
> comparing with the other packages defined in `distribution.yaml`, it seems the others place there what I believe to be the branch name, and not a particular tag.
> Same happens actually for the doc part.

Yes, you cannot use tags for the `source` and `doc` entries, only branches.

aremazeilles:
Thanks @clalancette for your comments.
This package is about providing some additional testing functionalities with respect to `rostest`. In particular it permits writing ros test files for verifying that a service server provides the expected reply given a specific request. And also to verify that a filter-like node generates the expected output message after receiving a specific message.

According to the [documentation on Automatic Testing with ROS](http://wiki.ros.org/Quality/Tutorials/UnitTesting), we are developing tools enhancing the ROS node unit testing : 

> Level 2. ROS node unit test (rostest + unittest/gtest): node unit tests start up your node and test its external API, i.e. published topics, subscribed topics, and services.

I thought that `interface` could be better that API as package name, but I was probably wrong. Would `node_api_test` sound better to you? I will give it a thought with my colleagues here, but if you think that would be still generic, I am happy to receive some suggestions.

hidmic (commented):
LGTM once @clalancette concerns have been addressed

aremazeilles:
So to replace the current package name, `node_interface_test`, we thought of the following options:

* `node_api_test`
* `node_api_unit_testing`
* `node_api_test_generation`

If none sounds good, I would really appreciate some suggestions.

Thank you a lot for your time.

clalancette:
> This package is about providing some additional testing functionalities with respect to `rostest`. In particular it permits writing ros test files for verifying that a service server provides the expected reply given a specific request. And also to verify that a filter-like node generates the expected output message after receiving a specific message.

Thank you for the detailed explanation of what you are trying to accomplish here.  That helps a lot.

Since this is an extension to `rostest`, I'm considering whether we should include that in here.  That would normally break our rules about using `ros` in the package name, but if it is extending that package name, maybe that is OK.  Maybe something like `rostest-node-interfaces`?  @tfoote what do you think?

dirk-thomas:
@clalancette @tfoote What is the status of this?

nuclearsandwich:
> If none sounds good, I would really appreciate some suggestions.

Since this package is enhancing rostest does it make sense to include rostest in the package name? For example rostest_node_interface_test?

Another option is to give it an arbitrary name like flaxseed and let the description explain what you can do with the package.

aremazeilles:
> Another option is to give it an arbitrary name like flaxseed and let the description explain what you can do with the package.

If we relax the constraint on the name self-explanation, then the current name `node_interface_test` could be sufficient?

nuclearsandwich:
> If we relax the constraint on the name self-explanation, then the current name `node_interface_test` could be sufficient?

I'm making an informed guess, but a guess nonetheless so I could be mistaken, but I think the concerns being raised about the name `node_interface_test` is twofold:
1. There is insufficient description in the package name to describe the package within a reasonable scope. ("reasonable" here being subjective)
2. The name is also broad enough that it occupies a large amount of "space" and could either cause confusion or collision with any package that performs any kind of test across any kind of interface for any kind of node.

I'm not sure what other ROS distribution maintainers think of my suggestion to make the name arbitrary / nonsensical but if that's not what you prefer then I can reiterate my suggestion to scope the name. In ROS 2 we have several packages that provide testing utilities which we call testing tool so I'll put forward the suggestion `rostest_node_interface_testing_tools or rostest_node_interfaces` as @clalancette suggested above.

Thanks for taking time to hear the feedback @aremazeilles.

aremazeilles:
Thanks @nuclearsandwich, I appreciate your clarification. Your second point is indeed particularly right, the name may be too broad. Regardings your proposition, I would be happy with a in-between naming, like `rostest_node_interface_tools`, but if another name is preferred by the other ROS distribution maintainers, no problems.

clalancette:
I'm fine with any of:

* `rostest_node_interface_testing_tools`
* `rostest_node_interface_tools`
* `rostest_node_interfaces`

It sort of violates our guidelines to not have "ros" in the name, but since this is using "rostest", I think that is reasonable here.

@aremazeilles Feel free to pick which one you think is best.

tfoote:
The name needs to keep some level of scope. Test as a postfix is usually representing tests of another package with the smaller name. `node_interface` is very broad. Thus since this is rostest specific the scoping to rostest is important. And what this package does is validate the node interfaces. So I'd suggest `rostest_node_interface_validation`

This makes it clear on inspection of what it's scope is for where it might be extended. And if you're reading a unit tests and see this referenced you can know exactly what the elements provide.

nuclearsandwich:
@aremazeilles since updating the name will require an updated bloom release do you mind if I close this now that the naming discussion has settled?

aremazeilles:
@nuclearsandwich , please go ahead and close it. I will make another PR along this week.

Thanks you all for the time you spent looking at this.

nuclearsandwich:
> Thanks you all for the time you spent looking at this.

Thanks for letting us help you decide what color to paint your [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality)!

PR #27017, opened 2020-10-17
add python3-mechanize

fmessmer:
`python-mechanize` is not available for `focal`, thus adding a new key for `python3-mechanize`:
- ubuntu: https://packages.ubuntu.com/focal/python3-mechanize
- fedora: https://pkgs.org/search/?q=python3-mechanize
- debian: https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=python3-mechanize

no `python3-mechanize` available for:
- arch: https://www.archlinux.org/packages/?sort=&q=python3-mechanize&maintainer=&flagged=
- gentoo: https://packages.gentoo.org/packages/search?q=python3-mechanize

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5969,6 +5969,10 @@ python3-matplotlib:
     '7': null
   slackware: [python3-matplotlib]
   ubuntu: [python3-matplotlib]
+python3-mechanize:
+  debian: [python3-mechanize]
+  fedora: [python3-mechanize]
+  ubuntu: [python3-mechanize]
 python3-mock:
   alpine: [py3-mock]
   debian: [python3-mock]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5970,9 +5970,14 @@ python3-matplotlib:
   slackware: [python3-matplotlib]
   ubuntu: [python3-matplotlib]
 python3-mechanize:
-  debian: [python3-mechanize]
+  debian:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5969,6 +5969,10 @@ python3-matplotlib:
     '7': null
   slackware: [python3-matplotlib]
   ubuntu: [python3-matplotlib]
+python3-mechanize:
+  debian: [python3-mechanize]
```
clalancette, line 5973: It looks like `python3-mechanize` is only available in bullseye and later, but we still support `buster`.  So I'll suggest:

```
debian:
  '*': [python3-mechanize]
  buster: null
```

rosdep/python.yaml
```diff
@@ -5969,6 +5969,10 @@ python3-matplotlib:
...
+  fedora: [python3-mechanize]
+  ubuntu: [python3-mechanize]
```
clalancette, line 5975: Same here; it looks like `python3-mechanize` is only available since focal, but we still support xenial and bionic.  So I'll suggest:

```
ubuntu:
  '*': [python3-mechanize]
  bionic: null
  xenial: null
```

clalancette (changes requested):

fmessmer:
@clalancette done

clalancette (approved):
Looks good, thanks for iterating.

PR #26837, opened 2020-10-06
Add leo_desktop to melodic

bjsowa:
I had to manually make changes and open Pull Request, because the `leo_viz` package has been moved to `leo_desktop` repository.

PR #27032, opened 2020-10-19
Add leo_common and leo_desktop repositories to kinetic

bjsowa:
This is a backport of #26837 to kinetic distribution.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6573,7 +6573,7 @@ repositories:
       - leo_description
       - leo_teleop
       tags:
-        release: release/melodic/{package}/{version}
+        release: release/kinetic/{package}/{version}
       url: https://github.com/fictionlab-gbp/leo_common-release.git
       version: 1.1.0-1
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -6562,34 +6562,41 @@ repositories:
...
     release:
+      packages:
+      - leo
+      - leo_description
+      - leo_teleop
       tags:
-        release: release/kinetic/{package}/{version}
-      url: https://github.com/fictionlab-gbp/leo_description-release.git
-      version: 0.3.0-1
+        release: release/melodic/{package}/{version}
```
clalancette, line 6576: This doesn't seem right; this should be kinetic.  The same goes below.

clalancette (changes requested):

kinetic/distribution.yaml
bjsowa: Sorry, I forgot to change the tags. Please check the newest commit.

PR #27037, opened 2020-10-19
add python3-influxdb-client-pip

jstiefel:
I would like to add the InfluxDB 2.0 Python Client pip package. This is the recent version of influxdb-python, which is already in the deps. This package only exists in pip and requires at least python 3.6.

https://pypi.org/project/influxdb-client/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2151,6 +2151,16 @@ python-influxdb:
     bionic: [python-influxdb]
     disco: [python-influxdb]
     xenial: [python-influxdb]
+python3-influxdb-client-pip:
+  debian:
+    pip:
+      packages: [influxdb-client]
+  fedora:
+    pip:
+      packages: [influxdb-client]
+  ubuntu:
+    pip:
+      packages: [influxdb-client]
 python-inject-pip:
   ubuntu:
     pip:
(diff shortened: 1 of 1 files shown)


hidmic (changes requested):
@jstiefel please check failing jobs. It appears your patch has formatting issues.

Also, it seems that for other platforms like `gentoo`, [`influxdb-client` has been packaged](https://packages.gentoo.org/packages/dev-python/influxdb). Would you consider providing keys for `gentoo`, `arch`, and `osx` as well? Links to package managers lists would help the review.

jstiefel:
> @jstiefel please check failing jobs. It appears your patch has formatting issues.

My bad, wrong alphabetic order. 

> Also, it seems that for other platforms like `gentoo`, [`influxdb-client` has been packaged](https://packages.gentoo.org/packages/dev-python/influxdb). Would you consider providing keys for `gentoo`, `arch`, and `osx` as well? Links to package managers lists would help the review.

This is the previous client I was referring to: https://github.com/influxdata/influxdb-python (vs. https://github.com/influxdata/influxdb-client-python). The apt package of this one is already in the deps.

hidmic (approved):
I see. Thanks for the clarification. 

Still, I'd like to see `arch`, `gentoo`, and `osx` rules in there, but it's not mandatory.

clalancette (approved):

PR #27079, opened 2020-10-21
grbl_ros: 0.0.14-1 in 'foxy/distribution.yaml' [bloom]

flynneva:
Increasing version of package(s) in repository `grbl_ros` to `0.0.14-1`:

- upstream repository: https://github.com/flynneva/grbl_ros.git
- release repository: https://github.com/flynneva/grbl_ros-release.git
- bloom version: `0.10.0`
- previous version for package: `0.2.1-1`

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1079,7 +1079,7 @@ repositories:
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/flynneva/grbl_ros-release.git
-      version: 0.2.1-1
+      version: 0.0.14-1
     source:
       type: git
       url: https://github.com/flynneva/grbl_ros.git
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1074,7 +1074,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/flynneva/grbl_ros.git
-      version: foxy
+      version: main
     release:
       tags:
         release: release/foxy/{package}/{version}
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -1079,7 +1079,7 @@ repositories:
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/flynneva/grbl_ros-release.git
-      version: 0.2.1-1
+      version: 0.0.14-1
```
jacobperron, line 1082: Typically version numbers go up (not down). Is this change correct?

jacobperron (commented):
The version for source and doc entries (`foxy`) does not exist.

foxy/distribution.yaml
flynneva: im switching from managing distro-specific branches to a single-source style. the old version reflects the distro-specific approach that was used before.

flynneva:
turns out when you run bloom-release --edit it does not give you the option to change the doc or source branches.

changing it manually here fixes the CI error.

foxy/distribution.yaml
dirk-thomas: If users have the previous version 0.2.1 installed they won't be able to update to 0.0.14 (since the version number is lower). They will need to uninstall and reinstall instead. Is that still desired / acceptable?

flynneva: no one is using this package since the previous released version does not even function really. so yeah thats fine.

flynneva: i would be shocked if a single person other than myself has tried out this grbl_ros package to be completely honest :smile:

PR #27054, opened 2020-10-21
rmw_implementation: 1.0.1-1 in 'foxy/distribution.yaml' [bloom]

ahcorde:
Increasing version of package(s) in repository `rmw_implementation` to `1.0.1-1`:

- upstream repository: https://github.com/ros2/rmw_implementation.git
- release repository: https://github.com/ros2-gbp/rmw_implementation-release.git
- distro file: `foxy/distribution.yaml`
- bloom version: `0.10.0`
- previous version for package: `1.0.0-1`

## rmw_implementation

```
* Foxy updated maintainers (#158 <https://github.com/ros2/rmw_implementation/issues/158>)
* Contributors: Alejandro Hernández Cordero
```

## test_rmw_implementation

```
* Fixed rmw_create_node arguments
* Add fault injection tests to construction/destroy APIs.  (#144 <https://github.com/ros2/rmw_implementation/issues/144>)
* Fixed function rmw_create_node in tests
* Add tests bad type_support implementation (#152 <https://github.com/ros2/rmw_implementation/issues/152>)
* Add tests for localhost-only node creation (#150 <https://github.com/ros2/rmw_implementation/issues/150>)
* Use 10x the intraprocess delay to wait for sent requests. (#148 <https://github.com/ros2/rmw_implementation/issues/148>)
* Added rmw_wait, rmw_create_wait_set, and rmw_destroy_wait_set tests (#139 <https://github.com/ros2/rmw_implementation/issues/139>)
* Added rmw_service_server_is_available tests (#140 <https://github.com/ros2/rmw_implementation/issues/140>)
* Add tests service/client request/response with bad arguments (#141 <https://github.com/ros2/rmw_implementation/issues/141>)
* Fixed function rmw_create_node in tests
* Added rmw_wait, rmw_create_wait_set, and rmw_destroy_wait_set tests (#139 <https://github.com/ros2/rmw_implementation/issues/139>)
* Added test for rmw_get_serialized_message_size (#142 <https://github.com/ros2/rmw_implementation/issues/142>)
* Add service/client construction/destruction API test coverage. (#138 <https://github.com/ros2/rmw_implementation/issues/138>)
* Added rmw_publisher_allocation and rmw_subscription_allocation related tests (#137 <https://github.com/ros2/rmw_implementation/issues/137>)
* Add tests take serialized with info bad arguments (#130 <https://github.com/ros2/rmw_implementation/issues/130>)
[...]

PR #27104, opened 2020-10-22
Remove test_rmw_implementation from rolling

ivanpauno:
See https://github.com/ros/rosdistro/pull/27054#discussion_r510289851.


dirk-thomas:
@ivanpauno Please see the failing CI for what needs to be changed.

PR #27106, opened 2020-10-22
any_node: 0.5.0-1 in 'melodic/distribution.yaml' [bloom]

mikekaram:
Increasing version of package(s) in repository `any_node` to `0.5.0-1`:

- upstream repository: https://github.com/ANYbotics/any_node.git
- release repository: https://github.com/mikekaram/any_node-release.git
- bloom version: `0.10.0`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -166,6 +166,27 @@ repositories:
       url: https://github.com/ros/angles.git
       version: master
     status: maintained
+  any_node:
+    doc:
+      type: git
+      url: https://github.com/ANYbotics/any_node.git
+      version: master
+    release:
+      packages:
+      - any_node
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
While I appreciate you are ANYbotics, this is just way too generic a name for a general ROS distribution.  This also goes for the package names below.  If you saw a package called `ros-melodic-any-node`, what do you think it would do?  I'll suggest adding the name of the robot(s), and changing this to be `anybotics`.

mikekaram:
@clalancette thanks for the input. Although I'm not ANYbotics, I needed this package for releasing mine, so I took the liberty to do so. I will change the name as you suggested.

Thanks,
Mike

PR #27164, opened 2020-10-27
[dep] Add pip version of boto3.

130s:
Rationale for the `pip` key: https://pypi.org/project/boto3/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5486,6 +5486,16 @@ python3-boto3:
     '*': ['python%{python3_pkgversion}-boto3']
     '7': null
   ubuntu: [python3-boto3]
+python3-boto3-pip:
+  debian:
+    pip:
+      packages: [boto3]
+  fedora:
+    pip:
+      packages: [boto3]
+  ubuntu:
+    pip:
+      packages: [boto3]
 python3-bson:
   debian: [python3-bson]
   fedora: [python3-bson]


nuclearsandwich:
@130s our current review policy on pip keys is that they should only be added when the system package is unavailable (https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1) since packages with pip keys as dependencies can't be released into a ROS distribution and built on the build farm.

If you have a need for this key in a package that's not meant to be released on into an official rosdistro you could add an additional custom rosdep source to include it.

130s:
Thanks for the headsup.

I'm unsure for the 2nd passage so appreciate clarification.

> If you have a need for this key in a package that's not meant to be released on into an official rosdistro

My immediate usecase is yes, the package I have in mind won't be released into rosdistro.

> you could add an additional custom rosdep source to include it.

Do you mean this PR can still get merged in, or do you mean a custom, non-official rosdep key?

With a quick look at the current rosdep key list, I've found a few pkgs that seem both deb and pip are registered. If those are early exceptions that were merged before the policy you mentioned above, I understand.

- [python-attrs](https://github.com/ros/rosdistro/blob/d6101638e721d1615d62435055bf5ba9ad7ac67a/rosdep/python.yaml#L532), [python-attrs-pip](https://github.com/ros/rosdistro/blob/d6101638e721d1615d62435055bf5ba9ad7ac67a/rosdep/python.yaml#L542)
- [python-bitstring](https://github.com/ros/rosdistro/blob/d6101638e721d1615d62435055bf5ba9ad7ac67a/rosdep/python.yaml#L651), [python-bitstring-pip](https://github.com/ros/rosdistro/blob/d6101638e721d1615d62435055bf5ba9ad7ac67a/rosdep/python.yaml#L662)

nuclearsandwich:
> Do you mean this PR can still get merged in, or do you mean a custom, non-official rosdep key?

I mean a separate set of rosdep sources that you can use for your projects: http://docs.ros.org/en/independent/api/rosdep/html/contributing_rules.html instead of replacing the default rules you could add a source that just adds the pip keys you want for your project.

>  If those are early exceptions that were merged before the policy you mentioned above, I understand.

To avoid disruption we've made the decision not to remove keys which have already been added but we're doing our best not to add new pip keys unless there is a very compelling use case for the whole community.

130s:
Got it, thank you for clarification.

Being able to add (instead of replacing) rosdep key (files) is nice. Great design!

PR #27188, opened 2020-10-29
Shadow robot melodic distro

georgiablanco:
I'd like to add Shadow Robot repos to be indexed and documented on ros.org

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -11524,6 +11524,46 @@ repositories:
       url: https://github.com/SteveMacenski/spatio_temporal_voxel_layer.git
       version: melodic-devel
     status: maintained
+  sr_common:
+    doc:
+      type: git
+      url: https://github.com/shadow-robot/sr_common.git
+      version: melodic-devel
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Hi @georgiablanco it looks like these repositories were added to Kinetic in #26919 and the commits later applied to this PR are removing the source entries. Is that intended or can this PR be closed?

georgiablanco:
Yes I have removed the source on purpose.

nuclearsandwich:
Why remove the source entries and not the doc entries?

georgiablanco:
because I was meant to do [this](http://wiki.ros.org/rosdistro/Tutorials/Indexing Your ROS Repository for Documentation Generation) and so realised source doesn't need to be there.

nuclearsandwich:
Thanks for clarifying.

georgiablanco:
Thank you. May you please do the Kinetic branch also? https://github.com/ros/rosdistro/pull/27190

nuclearsandwich:
Please see https://github.com/ros/rosdistro/pull/27190#issuecomment-718898597 for why that one is still pending.

PR #27195, opened 2020-10-30
Add libgmock-dev to Buster & Focal.

MichaelGrupp:
`google-mock` is a transitional package and shouldn't be used anymore.
See https://packages.debian.org/buster/google-mock

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2582,6 +2582,19 @@ libglw1-mesa:
   fedora: [mesa-libGlw]
   gentoo: [x11-libs/libGLw]
   ubuntu: [libglw1-mesa]
+libgmock-dev:
+  alpine: [gmock]
+  arch: [gmock]
+  debian:
+    buster: [libgmock-dev]
+  fedora: [gmock-devel]
+  freebsd: [googlemock]
+  gentoo: [dev-cpp/gtest]
+  openembedded: [gtest@meta-oe]
+  opensuse: [gmock-devel]
+  rhel: [gmock-devel]
+  ubuntu:
+    focal: [libgmock-dev]
 libgmp:
   debian: [libgmp-dev]
   fedora: [gmp-devel]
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
Would it also be supported to use a rule like `libgmock-dev (>= 1.8.1) | google-mock`, as it can be done in Debian?

nuclearsandwich:
> Would it also be supported to use a rule like `libgmock-dev (>= 1.8.1) | google-mock`, as it can be done in Debian?

It might work if it was tried but as far as I know only by accident and I do not think we want to encourage the use of complex dependency logic within rosdep itself as not every target system has the same mechanisms for expressing that logic.

My initial feeling is that it makes more sense to add a new key for libgmock-dev so that projects can switch to rather than trying to affect our own transition within the old package name.

nuclearsandwich (changes requested):
I think creating a new key for the new package name and making sure that we provide a null definition for this transitional package key when it is no longer available.

claireyywang:
@MichaelGrupp  Friendly ping

MichaelGrupp:
Sorry for the delay. I addressed your comments, I'm not completely sure if I did it correct tbh. Please have a look.

sloretz:
If the plan is to recommend people move away from the `google-mock` key, I think the `libgmock-dev` key should include the other platforms too

https://github.com/ros/rosdistro/blob/85888e1d3c0f663ecf3e446fc7905dab317482c8/rosdep/base.yaml#L1185-L1193

* alpine: gmock https://pkgs.alpinelinux.org/package/edge/main/x86/gmock
* arch: gmock https://www.archlinux.org/packages/community/x86_64/gmock/
* fedora: gmock-devel https://src.fedoraproject.org/rpms/gtest/blob/master/f/gtest.spec
* freebsd: currently `googlemock`, but should this be `devel/googlemock` - not sure how freebsd packaging works? https://www.freshports.org/devel/googlemock
* gentoo: dev-cpp/gtest https://packages.gentoo.org/packages/dev-cpp/gtest
* Openembedded Currently gtest@meta-oe, but maybe should be gmock@meta-oe? https://git.congatec.com/yocto/meta-openembedded/blob/146c382fadb5312c77adda9e7a94e02188e364ea/meta-oe/recipes-test/gmock/gmock_1.7.0.bb
* opensuse: currently gmock-devel, but maybe should be `gmock`? https://software.opensuse.org/download.html?project=devel%3Alibraries%3Ac_c%2B%2B&package=gmock
* rhel: gmock-dvel (Assuming package in CentOS is also in  Redhat enterprise) https://centos.pkgs.org/8/centos-powertools-x86_64/gmock-devel-1.8.0-5.el8.x86_64.rpm.html

@ros/rosdeputies Thoughts on what keys are correct for freebsd, openembedded, opensuse, and rhel?

tfoote:
Yes if libgmock-dev is the new recommended one we should make sure to fill it in on all platforms so people can transition.

jacobperron:
@MichaelGrupp Friendly ping. Are you able to fill in the requested platforms described above (https://github.com/ros/rosdistro/pull/27195#issuecomment-728605062)? Thanks!

MichaelGrupp:
@jacobperron Done.

Sorry for the late reply, just came back from vacation.

clalancette (approved):

ivanpauno (approved):

PR #27174, opened 2020-10-28
ros_babel_fish: 0.9.0-2 in 'kinetic/distribution.yaml' [bloom]

StefanFabian:
Increasing version of package(s) in repository `ros_babel_fish` to `0.9.0-2`:

- upstream repository: https://github.com/StefanFabian/ros_babel_fish.git
- release repository: https://github.com/StefanFabian/ros_babel_fish-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.10.0`
- previous version for package: `null`

## ros_babel_fish

```
* Added install target for service_info. Cleaned up package.xml
* Fixed service spec loading wrong response if service spec doesn't contain '---'.
* Updated integrated description provider for service descriptions to reflect changes in genmsg that caused some tests to fail.
  Both implementations are interoperable and the change should have no impact on service calls using ros_babel_fish but new implementation is now again consistent with message_traits::definition<...>().
  The new implementation is also less complex and faster.
* Added macros for template calls.
  This replaces the boiler plate code for situations where you would do switch(msg.type()) { ... } and call a template function with the C++ type of the message as a template parameter.
* Contributors: Stefan Fabian
```

## ros_babel_fish_test_msgs

```
* Added install target for service_info. Cleaned up package.xml
* Contributors: Stefan Fabian
```

PR #27198, opened 2020-10-30
Revert "ros_babel_fish: 0.9.0-2 in 'kinetic/distribution.yaml' [bloom]"

StefanFabian:
Reverts ros/rosdistro#27174

Sorry for the unnecessary work.
While the code works on kinetic, unfortunately, the gtest version of kinetic seems to be incompatible with the test code.
I don't have the time right now to fix this for a four year old distro, so, removing it seems like the best option for now.


Blast545 (approved):
Reverting automatically because never reached `main`

PR #27236, opened 2020-11-03
Update expected Debian sources.

nuclearsandwich:
This looks like it was copy-pasted from the above entry for Ubuntu and
used the Ubuntu repository components rather than the Debian ones.

The inclusion of contrib and non-free comes from https://wiki.ros.org/Installation/Debian#melodic.2FInstallation.2FDebianSources.Configure_your_Debian_repositories

There is no Debian specific article for setting up ROS sources but the
Ubuntu article applies if you update the distribution name.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,8 +94,8 @@ Keys in the rosdep database are required to come from packages contained in the 
 
 #### Debian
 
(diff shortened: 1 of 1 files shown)


wjwwood (approved):
Makes sense to me, might want a second review? If not I'll let you merge @nuclearsandwich.

mjcarroll (approved):

PR #27243, opened 2020-11-04
Add 'conan' rosdep key to be resolved using pip

miguelprada:
This enables using [conan](https://conan.io/) ([pypi package](https://pypi.org/project/conan/)) package manager to resolve dependencies (e.g. using [conan-cmake](https://github.com/conan-io/cmake-conan)) which may not be available via the usual rosdep sources, while respecting the usual workflow to build packages from source.

I reckon this may not be desired for packages to be released as binaries using the buildfarm, but it would be very helpful for packages which are never intended to go through the binary release process.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -108,7 +108,7 @@ cmakelint-pip:
   ubuntu:
     pip:
       packages: [cmakelint]
-conan:
+python3-conan-pip:
   debian:
     pip:
       packages: [conan]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -108,16 +108,6 @@ cmakelint-pip:
   ubuntu:
     pip:
       packages: [cmakelint]
-python3-conan-pip:
-  debian:
-    pip:
-      packages: [conan]
-  fedora:
-    pip:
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
Other discussions notwithstanding the rosdep key should conform to our guidelines for pip packages https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1

miguelprada:
Apologies for missing that one. I believe this now conforms to the rules.

Isn't it a bit confusing to prepend the key with `python3-` even though the same dependency is available for Python2 as well? This should even resolve to the Python2 version in systems where `ROS_PYTHON_VERSION` is set to 2, shouldn't it?

wjwwood (approved):
> I reckon this may not be desired for packages to be released as binaries using the buildfarm, but it would be very helpful for packages which are never intended to go through the binary release process.

This is fine, and an already understood limitation of pip based rosdep rules, you cannot use them with released packages since we cannot include those in our debian packages.

nuclearsandwich (approved):

miguelprada:
Thanks all!

PR #27335, opened 2020-11-14
add python3-selenium-pip key

knorth55:
I add `selenium` pip package for `python3`.

https://pypi.org/project/selenium/


knorth55:
I switch to use system packge #27358

PR #27336, opened 2020-11-14
Add rosdep key ruby-backports

meyerj:
Adds [ruby-backports](https://github.com/marcandre/backports) to the rosdep database. The package is a dependency of Orocos [typelib](https://github.com/orocos-toolchain/typelib) since https://github.com/orocos-toolchain/typelib/pull/107.

Links to package databases:
- https://aur.archlinux.org/packages/ruby-backports/
- https://packages.debian.org/search?keywords=ruby-backports
- https://packages.gentoo.org/packages/dev-ruby/backports
- https://packages.ubuntu.com/search?keywords=ruby-backports

I did not find the package in any database of Fedora packages. https://apps.fedoraproject.org/packages/ seems to be offline at the moment.

### DIFF
--- a/rosdep/ruby.yaml
+++ b/rosdep/ruby.yaml
@@ -87,7 +87,6 @@ ruby:
     yakkety: [ruby, ruby-dev]
     zesty: [ruby, ruby-dev]
 ruby-backports:
-  arch: [ruby-backports]
   debian: [ruby-backports]
   gentoo: [dev-ruby/backports]
   ubuntu: [ruby-backports]
--- a/rosdep/ruby.yaml
+++ b/rosdep/ruby.yaml
@@ -86,6 +86,11 @@ ruby:
     xenial: [ruby, ruby-dev]
     yakkety: [ruby, ruby-dev]
     zesty: [ruby, ruby-dev]
+ruby-backports:
+  arch: [ruby-backports]
+  debian: [ruby-backports]
+  gentoo: [dev-ruby/backports]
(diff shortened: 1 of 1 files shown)


rosdep/ruby.yaml
```diff
@@ -86,6 +86,11 @@ ruby:
     xenial: [ruby, ruby-dev]
     yakkety: [ruby, ruby-dev]
     zesty: [ruby, ruby-dev]
+ruby-backports:
+  arch: [ruby-backports]
```
nuclearsandwich, line 90:
```suggestion
```

AUR packages should not be added to Arch Linux directly: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#arch-linux

nuclearsandwich (changes requested):

rosdep/ruby.yaml
meyerj: Done. I noticed that `ruby-nokogiri` and `ruby-facets` are not core packages either:
- https://www.archlinux.org/packages/?q=ruby-nokogiri
- https://www.archlinux.org/packages/?q=ruby-facets

Do you want me to remove those entries, too?

nuclearsandwich: If you do please do so in a separate PR for easy review. One of the ongoing challenges with maintaining the rosdep definitions for Arch is that packages can move between the AUR and community repositories based on Trusted User adoption or orphaning.

nuclearsandwich (approved):

PR #27356, opened 2020-11-17
Add python3-pymongo key

gurbain:
python-pymongo cannot be found in Ubuntu focal packages and is named _python3-pymongo_ instead: https://packages.ubuntu.com/focal/python3-pymongo

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3442,14 +3442,7 @@ python-pymongo:
   osx:
     pip:
       packages: [pymongo]
-  ubuntu:
-    '*': [python-pymongo]
-    focal: [python3-pymongo]
-    trusty_python3: [python3-pymongo]
-    utopic_python3: [python3-pymongo]
-    vivid_python3: [python3-pymongo]
-    wily_python3: [python3-pymongo]
-    xenial_python3: [python3-pymongo]
+  ubuntu: [python-pymongo]
 python-pymouse:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


tfoote (dismissed):
Instead of aliasing this rule to be partially python2 and partially python3 please make a new one for `python3-mongodb`

gurbain:
Hi @tfoote,

Thank you for your input! This is my first PR in this repo, so I was not aware of this rule, I should have read the documentation further.

My concern was that the _python-pymongo_ entry was already partly aliased for several Ubuntu distros back to trusty_python3. I took the liberty to replace this with a single package name and create the __python3-pymongo__ entry as you suggested (see 9259bfc00aaa4f1cdc4265df81c7442b884029b8). I also checked the corresponding package names for other Linux distros.

Please, tell me if anything is missing!

tfoote:
Thanks for cleaning that up. But we keep those old rudundant rules in place to make sure not to break people who might still be using them. Please just add the new key so that it's available and not potentially breaking people already using the existing keys. The old keys won't effect people who choose to use the newer version.

gurbain:
Ok, it makes sense!

clalancette (approved):

sloretz (approved):
* arch https://www.archlinux.org/packages/community/x86_64/python-pymongo/
* debian https://packages.debian.org/stretch/python3-pymongo
* Fedora https://src.fedoraproject.org/rpms/python-pymongo/blob/master/f/python-pymongo.spec#_47 ( I think )
* gentoo https://packages.gentoo.org/packages/dev-python/pymongo

PR #27364, opened 2020-11-17
Added new package 'stag_ros' to noetic

Brenn10:
Added an entry in noetic/distributions.yaml for a new ros package 'stag_ros'. Bloom was used to create the release and this is a manual pull request.


mjcarroll (approved):
You may also consider adding a documentation entry.

PR #27377, opened 2020-11-19
Change python-rpi.gpio install's on Ubuntu Bionic

remi2257:
What ?
I changed **python-rpi.gpio** package installation from _apt_ to _pip_ on Bionic.

Why ?
_apt_ only work for arm architecture for this package (see the following link) : https://packages.ubuntu.com/bionic/python-rpi.gpio

Cheers

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4069,7 +4069,9 @@ python-rpi.gpio:
         packages: [RPi.GPIO]
   ubuntu:
     artful: [python-rpi.gpio]
-    bionic: [python-rpi.gpio]
+    bionic:
+      pip:
+        packages: [RPi.GPIO]
     trusty:
       pip:
         packages: [RPi.GPIO]


tfoote (changes requested):
Our standard policy is to use the apt based version if it exists. That is the only way we can build binary packages.

In this case how do you intend to use GPIO of a raspberry pi without using an ARM platform?

remi2257:
Good question, I should have answer at this question directly in my request 

_Context_
Currently working at [Niryo](https://niryo.com/niryo-one/), we are working on our next open source Robot and we want to make it as ROS friendly as possible. Our ROS stack allows to use the "physical" robot and also a fully simulated robot on Gazebo (Gripper & Camera). As there are a lot of dependencies, and we want to avoid the user to install by hand each one, we use rosdep.

_Why this PR_
The simulation is intend to be run on every Laptop, but with the current rosdep package list, the installation leads to a failure : cannot find python-rpi.gpio3. My PR aims to solve this issue and also provide a solution that works for every user on Bionic : whereas they are on RPI or just testing/simulating on their laptop.
Moreover, we cannot just skip the hardware package as it is also running in simulation with a fake interface.

Cheers

mjcarroll:
> Moreover, we cannot just skip the hardware package as it is also running in simulation with a fake interface.

Is there a way to isolate the hardware-specific `gpio` interactions?

tfoote:
This PR goes against our policy of preferring natively packaged content for the more generic content. This will remove the ability for people to leverage and use the most convenient and reproducible way to install this package. As you mature you likely should also be considering releasing your packages so that others can download and install them via debian package. And if we do this you will be unable to release your packages into debian packages because of the dependency on a pip package. 

As Michael pointed out since this is a hardware IO module it's highly likely that you can isolate these interactions such that the user does not interact with them when not using the hardware.

remi2257:
No worries, I just didn't understand why the Ubuntu package was only used with Bionic & Artful.
We are going to change our level of abstraction in order to meet your policy !
I'm closing the request, cheers :robot:

PR #27390, opened 2020-11-19
adding qt4-linguist-tools rosdep key

zflat:
Provides binaries `lrealease` and `lupdate`

* Debian: https://packages.debian.org/buster/qt4-linguist-tools
* Ubuntu: https://packages.ubuntu.com/bionic/qt4-linguist-tools
* Gentoo: https://packages.gentoo.org/packages/dev-qt/linguist-tools
* Fedora: https://src.fedoraproject.org/rpms/qt/blob/master/f/qt.spec

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5541,8 +5541,8 @@ qt4-linguist-tools:
   gentoo: ['dev-qt/linguist-tools']
   ubuntu:
     '*': null
-    xenial: [qt4-linguist-tools]
     bionic: [qt4-linguist-tools]
+    xenial: [qt4-linguist-tools]
 qt4-qmake:
   arch: [qt4]
   debian: [qt4-qmake]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5541,8 +5541,8 @@ qt4-linguist-tools:
   gentoo: ['dev-qt/linguist-tools']
   ubuntu:
     '*': null
-    xenial: qt4-linguist-tools
-    bionic: qt4-linguist-tools
+    xenial: [qt4-linguist-tools]
+    bionic: [qt4-linguist-tools]
 qt4-qmake:
   arch: [qt4]
   debian: [qt4-qmake]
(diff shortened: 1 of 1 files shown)


tfoote:
This is in draft status. Is there more you plan to add?

zflat:
> This is in draft status. Is there more you plan to add?

I just left it in draft state while checks were running and did not come back to change it. Thanks for checking!

tfoote (approved):

rosdep/base.yaml
```diff
@@ -5535,6 +5535,11 @@ qt4-dev-tools:
   gentoo: ['dev-qt/assistant:4', 'dev-qt/linguist:4', 'dev-qt/pixeltool:4']
   opensuse: [libqt4-devel-doc]
   ubuntu: [qt4-dev-tools]
+qt4-linguist-tools:
+  debian: [qt4-linguist-tools]
+  fedora: [qt]
+  gentoo: ['dev-qt/linguist-tools']
+  ubuntu: [qt4-linguist-tools]
```
clalancette, line 5542: Looking at https://packages.ubuntu.com/bionic/qt4-linguist-tools , I'm guessing that as of Focal there is no more support for this package.  Thus, I'll suggest that we change this to:

```
  ubuntu:
    '*': null
    xenial: qt4-linguist-tools
    bionic: qt4-linguist-tools
```

rosdep/base.yaml
```diff
@@ -5535,6 +5535,11 @@ qt4-dev-tools:
   gentoo: ['dev-qt/assistant:4', 'dev-qt/linguist:4', 'dev-qt/pixeltool:4']
   opensuse: [libqt4-devel-doc]
   ubuntu: [qt4-dev-tools]
+qt4-linguist-tools:
+  debian: [qt4-linguist-tools]
+  fedora: [qt]
```
clalancette, line 5540: From what I can tell, the `qt` package on fedora does not install the `lrelease` and `lupdate` binaries.  To get that, you actually need `qt-devel`.  So I'll suggest we switch to that for Fedora.

clalancette (dismissed):

zflat:
@clalancette I agree with your suggestions and applied them. Thanks for the feedback!

nuclearsandwich (approved):

PR #27418, opened 2020-11-23
Add rosdep key for python3 owlready2

marioney:
Adds `python3-owlready2-pip` to `rosdep/python.yaml` file

Owlready2 is used for ontology-oriented programming in Python. It can load OWL 2.0 ontologies as Python objects, modify them, save them, and perform reasoning.

pip: https://pypi.org/project/Owlready2/
source: https://bitbucket.org/jibalamy/owlready2/commits/


mjcarroll (approved):
I have verified that this is currently only available via `pip` on Fedora and Ubuntu.

PR #27431, opened 2020-11-23
Add cnpy package

PeterMitrano:
Forgive me if I'm mistaken, but I think this is the first step towards releasing. This package has a dependency on zlib, so I feel like that should be listed somewhere (rosdep?)

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -483,6 +483,21 @@ repositories:
       url: https://github.com/ros/cmake_modules.git
       version: 0.4-devel
     status: maintained
+  cnpy:
+    doc:
+      type: git
+      url: https://github.com/PeterMitrano/cnpy.git
+      version: master
+    release:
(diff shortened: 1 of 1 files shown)


PeterMitrano:
Ok, I see the error 
```
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/rosdistro/rosdistro/test/test_build_caches.py", line 57, in test_build_caches
    raise RuntimeError('\n'.join(errors))
RuntimeError: noetic: missing package.xml file for package "cnpy"
```

But I'm not sure what that means. Currently running the test locally to try to debug.

PeterMitrano:
ok I figured out that the bloom release scripts does much of what I need and also creates a PR for me, so closing this one.

PR #27432, opened 2020-11-23
Updating llvm in Debian Stretch to match Ubuntu 18.04 version.

JWhitleyWork:
The default version of LLVM tied to the `llvm` and `llvm-dev` packages in Debian Stretch is _very_ old and does not match the version provided in Ubuntu 18.04. This updates the version for Debian Stretch to match the version available in the equivalent Ubuntu version.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4694,12 +4694,16 @@ linux-kernel-headers:
   ubuntu: [linux-libc-dev]
 llvm:
   arch: [llvm]
-  debian: [llvm]
+  debian:
+    '*': [llvm]
+    stretch: [llvm-6.0]
   gentoo: [sys-devel/llvm]
   ubuntu: [llvm]
 llvm-dev:
   arch: [llvm]
-  debian: [llvm-dev]
+  debian:
+    '*': [llvm-dev]
+    stretch: [llvm-6.0-dev]
   gentoo: [sys-devel/llvm]
   ubuntu: [llvm-dev]
 lm-sensors:


tfoote (changes requested):
This is only in stretch backports not in stretch. https://packages.debian.org/stretch-backports/llvm-6.0-dev we can't change it over. 

This is potentially pretty disruptive which is why they presumably haven't pushed it back as the default and I don't think that we should swap this out from under people in the same way either. 

We don't have backports in our default sources so if we wanted to do this we'd have to import this and potentially any of the recursive dependency tree into our repositories and monitor for any updates.

Part of the reason for using a stable platform is that it's stable. Unless there's some blocker types of problems with the existing version of llvm I would suggest that we keep using the default version in the distribution.

JWhitleyWork:
@tfoote Fair enough. So, here's my problem: I'm trying to release the package `tvm_vendor` for Melodic. However, specifically with Debian Stretch, it needs a newere version of LLVM to build. Newer versions are available in the default repos (see https://packages.debian.org/search?keywords=llvm&searchon=names&suite=all&section=all) like 4, 5, and 7 (just not 6 for some reason, though it is the default for Ubuntu 18.04's `llvm` and `llvm-dev`) but they aren't linked to the metapackage `llvm` or `llvm-dev` in Stretch. What would you suggest? Making a new key (e.g. something like `llvm-7` or `llvm-latest`)?

tfoote:
Yes a separate key for the alternative versions would be best, assuming it won't collide/conflict with the default one. Most users can keep using the generic ones and you can use this one temporarily as a workaround on the older platform.

JWhitleyWork:
Thanks, @tfoote! Closing in favor of #27460 .

PR #27456, opened 2020-11-24
sot-tools: 2.3.2-1 in 'melodic/distribution.yaml' [bloom]

Rascof:
Increasing version of package(s) in repository `sot-tools` to `2.3.1-1`:

- upstream repository: https://github.com/stack-of-tasks/sot-tools.git
- release repository: https://github.com/stack-of-tasks/sot-tools-ros-release.git
- bloom version: `0.10.0`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -11794,7 +11794,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/stack-of-tasks/sot-tools-ros-release.git
-      version: 2.3.1-1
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Thanks for submitting this. A quick review I saw that the license was stated as BSD in the package.xml however there's no top level LICENSE file with the license nor do there appear to be any license headers in the individual source or header files which is where the BSD license usually resides. 

Could you make sure to fill those in so that we can confidently redistribute them.

tfoote (approved):

PR #27488, opened 2020-11-25
Add rosdep key for python3 thriftpy

rluqueb:
Adds python3-thriftpy-pip to rosdep/python.yaml file

ThriftPy is a pure python implementation of Apache Thrift in a pythonic way.

pip: https://pypi.org/project/thriftpy/
doc: https://thriftpy.readthedocs.io/en/latest/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6730,10 +6730,9 @@ python3-texttable:
   gentoo: [dev-python/texttable]
   rhel: ['python%{python3_pkgversion}-texttable']
   ubuntu: [python3-texttable]
-python3-thriftpy-pip:
-  ubuntu:
-    pip:
-      packages: [thriftpy]
+python3-thriftpy:
+  fedora: [python3-thriftpy]
+  ubuntu: [python3-thriftpy]
 python3-tilestache-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6730,6 +6730,10 @@ python3-texttable:
   gentoo: [dev-python/texttable]
   rhel: ['python%{python3_pkgversion}-texttable']
   ubuntu: [python3-texttable]
+python3-thriftpy-pip:
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
This appears to be long standing packaged in native packages. 

- ubuntu: https://packages.ubuntu.com/search?keywords=python3-thriftpy
- debian: https://packages.debian.org/search?keywords=python3-thriftpy

If native packages are available they should take precedence following our guide: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

Please take a look at where this is packaged and update this rule to include at least ubuntu and debian and preferably other target platforms listed in the guide.

rluqueb:
I added thriftpy from Pypip repository because it is not available in Ubuntu 16.04 in order to install ros dependency for a kinetic package. If it not possible, i will proceed to change the rule.

rosdep/python.yaml
```diff
@@ -6730,6 +6730,9 @@ python3-texttable:
   gentoo: [dev-python/texttable]
   rhel: ['python%{python3_pkgversion}-texttable']
   ubuntu: [python3-texttable]
+python3-thriftpy:
+  fedora: [python3-thriftpy]
+  ubuntu: [python3-thriftpy]
```
clalancette, line 6735: Since this is not available on Xenial, I'll suggest:

```
ubuntu:
  '*': [python3-thiftpy]
  xenial: null
```

clalancette (changes requested):
Besides my comment inline, it would also be good to add keys for at least Debian.

rosdep/python.yaml
tfoote: I'd actually suggest using the pip rule specifically for xenial. Since it's not available there via debian package it won't conflict so we can use the pip rule as a fallback on that platform.

clalancette: That works too :+1:

clalancette (approved):

tfoote (approved):

PR #27496, opened 2020-11-26
Xmlschema

hakuturu583:
Fix problems described in this PR.
Sorry for missing ping.
https://github.com/ros/rosdistro/pull/26913

I want to add xmlschema to rosdistro.

Pypy package: https://pypi.org/project/xmlschema/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5270,6 +5270,16 @@ python-xmlplain-pip:
   ubuntu:
     pip:
       packages: [xmlplain]
+python-xmlschema:
+  debian:
+    pip:
+      packages: [xmlschema]
+  fedora:
+    pip:
+      packages: [xmlschema]
+  ubuntu:
+    pip:
+      packages: [xmlschema]
 python-xmltodict:
   debian:
     buster: [python-xmltodict]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5260,16 +5260,6 @@ python-xlrd:
   debian: [python-xlrd]
   fedora: [python-xlrd]
   ubuntu: [python-xlrd]
-python-xmlplain-pip:
-  debian:
-    pip:
-      packages: [xmlplain]
-  fedora:
-    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5281,16 +5281,16 @@ python-xlrd:
   debian: [python-xlrd]
   fedora: [python-xlrd]
   ubuntu: [python-xlrd]
-python-xmlplain-pip:
+python-xmlschema:
```
tfoote, line 5284: you seem to be removing an existing key here and overwriting it.

rosdep/python.yaml
```diff
@@ -6874,6 +6874,16 @@ python3-wxgtk4.0:
   fedora: [python3-wxpython4]
   gentoo: [dev-python/wxpython]
   ubuntu: [python3-wxgtk4.0]
+python3-xmlschema-pip:
+  debian:
+    pip:
+      packages: [xmlplain]
```
tfoote, line 6880: This looks like the new set of rules that you're adding, but they appear to be referencing  a different packge.

tfoote (changes requested):
This PR needs a little bit of cleanup to just be inserting the new rule mentioned in the description.

rosdep/python.yaml
tfoote, line 5284: You're still removing the python-xmlplain-pip. Please maintain that key and add the new keys.

hakuturu583:
I add python-xmlplain-pip package

rosdep/python.yaml
```diff
@@ -5291,6 +5291,16 @@ python-xmlplain-pip:
   ubuntu:
     pip:
       packages: [xmlplain]
+python-xmlschema:
```
clalancette, line 5294: Since this is a pip-only package, it should be: `python3-xmlschema-pip`

hakuturu583:
Sorry, I took a lot of mistakes in this branch, so I re-create PR about xmlschema package.

PR #27498, opened 2020-11-26
Feature/python3 bs4

hakuturu583:
This package is based on Python2. (https://github.com/ros/rosdistro/blob/ddaa4b4b257cac3915a0a85f57a14d9fb82f43bf/rosdep/python.yaml#L754)
https://packages.ubuntu.com/ja/bionic/python-beautifulsoup
We want to use beautifulsoup4 in Python3.
So we want to add this package into rosdistro.
https://pypi.org/project/beautifulsoup4/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -751,6 +751,11 @@ python-breathe:
   debian: [python-breathe]
   fedora: [python-breathe]
   ubuntu: [python-breathe]
+python-bs4:
+  debian: [python-bs4]
+  fedora: [python-beautifulsoup4]
+  gentoo: [=dev-python/beautifulsoup-4*]
+  ubuntu: [python-bs4]
 python-bson:
   debian: [python-bson]
   fedora: [python-bson]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
This appears to be packaged for Ubuntu already and we should use that version:

https://packages.ubuntu.com/bionic/python3-bs4

Also Debian https://packages.debian.org/buster/python3-bs4

Please search for other platforms and switch this to use the native packages not pip.

hakuturu583:
Change to the native package.

hakuturu583:
Ubuntu : https://packages.ubuntu.com/bionic/python3-bs4
Debian : https://packages.debian.org/buster/python3-bs4
Arch : https://www.archlinux.org/packages/community/any/python-beautifulsoup4/
Fedora：https://src.fedoraproject.org/rpms/python-beautifulsoup4

mjcarroll (approved):

tfoote (approved):
Thanks for the iteration lgtm

PR #27542, opened 2020-12-01
tvm_vendor: 0.7.1-1 in 'melodic/distribution.yaml' [bloom]

JWhitleyWork:
Increasing version of package(s) in repository `tvm_vendor` to `0.7.1-1`:

- upstream repository: https://github.com/autowarefoundation/tvm_vendor.git
- release repository: https://github.com/autowarefoundation/tvm_vendor-release.git
- distro file: `melodic/distribution.yaml`
- bloom version: `0.10.0`
- previous version for package: `0.7.0-1`

## tvm_vendor

```
* Updating LLVM dependency.
* Moving catkin_package() back to after add_library().
* Contributors: Joshua Whitley
```

PR #27544, opened 2020-12-01
tvm_vendor: 0.7.1-1 in 'dashing/distribution.yaml' [bloom]

JWhitleyWork:
Increasing version of package(s) in repository `tvm_vendor` to `0.7.1-1`:

- upstream repository: https://github.com/autowarefoundation/tvm_vendor.git
- release repository: https://github.com/autowarefoundation/tvm_vendor-release.git
- bloom version: `0.10.0`
- previous version for package: `0.7.0-1`

## tvm_vendor

```
* Updating LLVM dependency.
* Moving catkin_package() back to after add_library().
* Contributors: Joshua Whitley
```

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -3735,7 +3735,7 @@ repositories:
       tags:
         release: release/dashing/{package}/{version}
       url: https://github.com/autowarefoundation/tvm_vendor-release.git
-      version: 0.7.0-1
+      version: 0.7.1-1
     source:
       type: git
       url: https://github.com/autowarefoundation/tvm_vendor.git


nuclearsandwich:
Holding for the [upcoming Dashing sync](https://discourse.ros.org/t/preparing-for-dashing-sync-and-patch-release-2020-12-03/17391).

JWhitleyWork:
@nuclearsandwich This package is currently failing to build on the build farm and I believe this will fix it. Does that mean that this simply won't get sync'd until the next package sync? I'm OK with that if it is the case - the [melodic one](#27542) is the one I really care about short-term.

nuclearsandwich:
> @nuclearsandwich This package is currently failing to build on the build farm and I believe this will fix it. Does that mean that this simply won't get sync'd until the next package sync?

I generally consider packages that have never built low priority to fix during a pre-sync hold. In general if you'd like something to be released during the hold I'd appreciate a post in the sync discussion thread on Discourse in addition to comments on the rosdistro PR since not everyone is following the ros/rosdistro repo and so won't know to comment if they'll be affected by a late release.

I'd normally merge this one since there's not much possible harm in potentially fixing a broken package but since the farm is backed up with the recent Eloquent and Foxy builds I do not want to risk delaying the sync waiting for this to come in behind them, so yes, this will come in the next Dashing sync after this week's.

JWhitleyWork:
Thanks for the explanation! Not a problem at all.

PR #27545, opened 2020-12-01
Bumped version for stag_ros

Brenn10:
This new release should fix issues in the build farm. I had run the prerelease before the original release, but the build farm sent additional errors while building. This will hopefully increase compatibility so that it will not fail.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -6110,7 +6110,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/usrl-uofsc/stag_ros-release.git
-      version: 0.3.4-2
+      version: 0.3.6-1
     source:
       type: git
       url: https://github.com/usrl-uofsc/stag_ros.git


clalancette:
> This new release should fix issues in the build farm. I had run the prerelease before the original release, but the build farm sent additional errors while building. This will hopefully increase compatibility so that it will not fail.

That's fine.  Is there a reason you are not using bloom to do the update?  Using bloom gives us a lot more metadata about what the change is, for what distribution, etc.

Brenn10:
I am having an issue when bloom is asking about oauth. I try inputting my user name and password, but it fails. When I say not to use oauth, bloom exits and recommends a manual pull request. I use bloom to create the release, but it fails to create the pull request.

clalancette:
Do you have two-factor authentication enabled in Github?  If so, you need to follow the instructions at http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization to set it up properly.

Brenn10:
I do not.

Brenn10:
I am closing this request and will attempt again to use bloom.

PR #27571, opened 2020-12-02
Cyclone Dashing upstream development branch 0.7 version

sloretz:
It looks like the upstream branch should be the 0.7.x version instead of the 0.5.x version.

https://github.com/eclipse-cyclonedds/cyclonedds/tree/releases/0.7.x


nuclearsandwich (approved):

nuclearsandwich:
Good catch! Thanks.

PR #27579, opened 2020-12-03
Adding new repository for indexing: tb3_tools

xuyuan-ict:
This repo includes some extended packages for monitoring and controlling turtlebot3 robot and we will enrich this repository in the near future.

Thanks

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -18289,5 +18289,15 @@ repositories:
       url: https://github.com/zivid/zivid-ros.git
       version: master
     status: maintained
+  tb3_tools:
+    doc:
+      type: git
+      url: https://github.com/xuyuan-ict/tb3_tools.git
+      version: main
+    source:
+      type: git
+      url: https://github.com/xuyuan-ict/tb3_tools.git
+      version: main
+    status: maintained
 type: distribution
 version: 2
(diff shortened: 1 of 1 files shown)


kinetic/distribution.yaml
```diff
@@ -15723,6 +15723,16 @@ repositories:
...
       version: master
     status: developed
+  tb3_tools:
+    doc:
+      type: git
+      url: https://github.com/xuyuan-ict/tb3_tools.git
+      version: main
+    source:
+      type: git
+      url: https://github.com/xuyuan-ict/tb3_tools.git
```
clalancette, line 15733: I've noticed a couple of problems in your repository:

1.  The license is declared in the package.xml (which is good), but there is no LICENSE file in the repository.  Can you please add one?
1.  You have checked in the `.pyc` files, which is not a good idea.  That will only work for the specific version of python that you used.  I'll suggest removing those files from the repository.
1.  The `tb3_safe_teleop` package depends on `message_generation`, but doesn't actually generate any messages.  I'll suggest you remove that dependency.

clalancette (changes requested):

kinetic/distribution.yaml
xuyuan-ict: Thank you for your valuable advice.
I have fixed these three problems in my repository : )

clalancette (approved):
Looks good, thanks for fixing things up.

PR #27580, opened 2020-12-03
add python-xmlschema-pip package

hakuturu583:
Signed-off-by: Masaya Kataoka <ms.kataoka@gmail.com>

I want to add xmlschema pip package to the rosdistro.
The URL of the pypi package is here.
https://pypi.org/project/xmlschema/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6915,7 +6915,7 @@ python3-wxgtk4.0:
 python3-xmlschema:
   debian:
     '*': [python3-xmlschema]
-    'buster':
+    buster:
       pip:
         packages: [xmlschema]
   fedora: [python3-xmlschema]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5296,6 +5296,16 @@ python-xmlplain-pip:
   ubuntu:
     pip:
       packages: [xmlplain]
+python-xmlschema-pip:
+  debian:
+    pip:
+      packages: [xmlschema]
+  fedora:
+    pip:
+      packages: [xmlschema]
+  ubuntu:
+    pip:
+      packages: [xmlschema]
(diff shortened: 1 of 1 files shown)


hakuturu583:
This PR related to #27496

rosdep/python.yaml
```diff
@@ -5296,6 +5296,16 @@ python-xmlplain-pip:
...
+python-xmlschema-pip:
+  debian:
+    pip:
+      packages: [xmlschema]
+  fedora:
+    pip:
+      packages: [xmlschema]
+  ubuntu:
+    pip:
+      packages: [xmlschema]
```
clalancette, line 5308: This is implicitly a Python 2 key.  Additionally, it looks like this package is available in Fedora, and also Debian as a system package.  That probably means it will eventually become a package in Ubuntu (though that is not guaranteed).

So with all of that said, my suggestion here is that we make this look more like the following:

```
python3-xmlschema:
  debian:
    '*': [python3-xmlschema]
    'buster':
      pip:
        packages: [xmlschema]
  fedora: [python3-xmlschema]
  ubuntu:
    pip:
      packages: [xmlschema]
```

clalancette (changes requested):

rosdep/python.yaml
hakuturu583: Thank you, I fixed this problem.

hakuturu583:
I fixed package names.
debian : https://debian.pkgs.org/sid/debian-main-amd64/python3-xmlschema_1.3.1-1_all.deb.html
fedora : https://fedora.pkgs.org/33/fedora-x86_64/python3-xmlschema-1.0.18-4.fc33.noarch.rpm.html

clalancette (approved):

sloretz (approved):

PR #27583, opened 2020-12-03
add python3-pypng

fmessmer:
ubuntu: https://packages.ubuntu.com/focal/python3-png
debian: https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=python3-png
arch: https://www.archlinux.org/packages/community/any/python-pypng/
fedora: https://fedora.pkgs.org/31/fedora-x86_64/python3-pypng-0.0.18-15.fc31.noarch.rpm.html
gentoo: https://packages.gentoo.org/packages/dev-python/pypng

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6387,7 +6387,9 @@ python3-pypng:
   debian: [python3-png]
   fedora: [python3-pypng]
   gentoo: [dev-python/pypng]
-  ubuntu: [python3-png]
+  ubuntu:
+    '*': [python3-png]
+    xenial: null
 python3-pyproj:
   arch: [python-pyproj]
   debian: [python3-pyproj]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6369,6 +6369,12 @@ python3-pyparsing:
   gentoo: [dev-python/pyparsing]
   openembedded: [python3-pyparsing@meta-python]
   ubuntu: [python3-pyparsing]
+python3-pypng:
+  arch: [python-pypng]
+  debian: [python3-png]
+  fedora: [python3-pypng]
+  gentoo: [dev-python/pypng]
+  ubuntu: [python3-png]
```
clalancette, line 6377: Unfortunately, it looks like this package isn't available in Ubuntu Xenial.  Since that is still one of our supported platforms, I think we need to change this to:

```
  ubuntu:
    '*': [python3-png]
    xenial: null
```

clalancette (changes requested):

rosdep/python.yaml
fmessmer: done

clalancette (approved):

tfoote (approved):

PR #27595, opened 2020-12-04
Added pyquaternion pip package to python.yaml for ubuntu and debian

fbottarel:


rosdep/python.yaml
```diff
@@ -354,6 +354,13 @@ pyqt5-dev-tools:
   fedora: [python-qt5-devel]
   gentoo: [dev-python/PyQt5]
   ubuntu: [pyqt5-dev-tools]
+pyquaternion-pip:
```
sloretz, line 357: Please use the key name `python3-pyquaternion-pip`

sloretz (changes requested):

tfoote (changes requested):
We already had this temporarily and reverted it to replace it with fully released packages. 

Initial duplicate of this: https://github.com/ros/rosdistro/pull/17065
Roll back: https://github.com/ros/rosdistro/pull/23861
First release with discussion: https://github.com/ros/rosdistro/pull/23794

We already have pyquaternion in the distributions https://github.com/ros/rosdistro/blob/47a96867e39139de012185b81e4f819292e1bd69/noetic/distribution.yaml#L4317

https://github.com/ros/rosdistro/blob/de886291d9f3e1500705c27b9e02c251642d748c/melodic/distribution.yaml#L8682

Unless there's a significant change in the current state I would consider this a duplicate of the initial request and recommend closing in favor of using the fully packaged versions.

clalancette:
Because of @tfoote 's last comment, I'm going to go ahead and close this.  If you still think that this should be in the distribution, feel free to reopen and address Tully's comments.

fbottarel:
Thanks for the swift response @tfoote. I am quite a ros noob, please forgive if I got this wrong. The reason for me asking to include it is that I have a ros package with a couple of python scripts depending on `pyquaternion` and I was looking for a way to cleanly include it in the `package.xml`. Following [this guide](https://docs.ros.org/en/melodic/api/catkin/html/user_guide/setup_dot_py.html) I tried to look it up with `rosdep` and couldn't find any correspondences, that is why I opened the PR. However, the links you provided both feature `pyquaternion`. 

There is clearly something I am missing here. Would you be able to help? I have a full desktop ROS Melodic installation.

tfoote:
It's not in desktop there are many other packages in the distro that aren't installed by default. You'll need to install it with `sudo apt-get install ros-melodic-pyquaternion` and/or add it as a rosdep on `pyquaternion` which is a package already in melodic and use the usual dependency resolution techniques.

PR #27597, opened 2020-12-04
[dep] gpg-agent

130s:
- debian https://packages.debian.org/jessie/gnupg-agent
- Ubuntu https://packages.ubuntu.com/search?keywords=gnupg-agent

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1219,9 +1219,6 @@ glut:
   openembedded: [freeglut@meta-oe]
   rhel: [freeglut-devel]
   ubuntu: [freeglut3-dev]
-gnupg-agent:
-  debian: [gnupg-agent]
-  ubuntu: [gnupg-agent]
 gnuplot:
   arch: [gnuplot]
   debian: [gnuplot]
@@ -1269,6 +1266,13 @@ gperftools:
   fedora: [gperftools, gperftools-devel]
   gentoo: [dev-util/gperf]
   ubuntu: [google-perftools, libgoogle-perftools-dev]
+gpg-agent:
+  debian:
+    '*': [gpg-agent]
+    jessie: [gnupg-agent]
+  ubuntu:
+    '*': [gpg-agent]
+    xenial: [gnupg-agent]
 gphoto2:
   arch: [gphoto2]
   debian: [gphoto2]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1219,6 +1219,9 @@ glut:
   openembedded: [freeglut@meta-oe]
   rhel: [freeglut-devel]
   ubuntu: [freeglut3-dev]
+gnupg-agent:
+  debian: [gnupg-agent]
+  ubuntu: [gnupg-agent]
 gnuplot:
   arch: [gnuplot]
   debian: [gnuplot]


rosdep/base.yaml
```diff
@@ -1168,6 +1168,9 @@ glut:
   openembedded: [freeglut@meta-oe]
   rhel: [freeglut-devel]
   ubuntu: [freeglut3-dev]
+gnupg-agent:
+  debian: [gnupg-agent]
```
clalancette, line 1172: If I `apt-get install gnupg-agent` on Ubuntu 20.04, I get a package that only has documentation (no binaries, libraries, or headers).  So it looks like just a metapackage that depends on other things.  Given that, I'm inclined not to add this to the rosdep keys.  Maybe we can add something more specific, like gpg-agent, which *does* have binaries in it.

clalancette (changes requested):

rosdep/base.yaml
130s: I see. I guess I'm seeing the same on 18.04.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.5 LTS

$ apt-file list gnupg-agent
gnupg-agent: /usr/share/doc/gnupg-agent/NEWS.Debian.gz
gnupg-agent: /usr/share/doc/gnupg-agent/changelog.Debian.gz
gnupg-agent: /usr/share/doc/gnupg-agent/copyright

$ apt-file list gpg-agent
gpg-agent: /etc/X11/Xsession.d/90gpg-agent
gpg-agent: /etc/logcheck/ignore.d.server/gpg-agent
gpg-agent: /usr/bin/gpg-agent
gpg-agent: /usr/lib/gnupg/gpg-check-pattern
gpg-agent: /usr/lib/gnupg/gpg-preset-passphrase
:
```

Thing is that `gnupg-agent` has been listed in Docker installation https://docs.docker.com/engine/install/ubuntu/. As of today I still see it:
```
Set up the repository

    Update the apt package index and install packages to allow apt to use a repository over HTTPS:

    $ sudo apt-get update

    $ sudo apt-get install \
        apt-transport-https \
        ca-certificates \
        curl \
        gnupg-agent \
        software-properties-common
```
That said I'm not sure why `gnupg-agent` is needed. I actually haven't found Docker resource that talks about this pkg, but I do find discussion about `gpg-agent`. I can double-check in Docker community.

130s: Asked in [forums.docker.com#101592](https://forums.docker.com/t/linux-debian-ubuntu-dependency-on-gnupg-agent/101592).

tfoote (changes requested):
After some digging I found that docker is using gnupg-agent because it's available on all platforms. It's technically only a transitional dummy package after xenial. 

https://packages.ubuntu.com/source/focal/gnupg2

To that end I think that we should leverage the new name and key and since we can do platform dependent keys just override the old key for xenial to use the old package name.

So `'*': [gpg-agent]` with `xenial: [gnupg-agent]`
with key `gpg-agent`

Debian has `gpg-agent` available back to stretch https://packages.debian.org/sid/gpg-agent so the jessie would also get the old key.

[Similar follow up on the docker forum](https://forums.docker.com#101592/2)

130s:
Thanks Tully for taking a deeper look. Please re-review as I think I've updated accordingly based on the finding and the suggestion.

tfoote (approved):

clalancette (approved):

PR #27640, opened 2020-12-07
Remove stretch from the distribution as it is no longer supported.

clalancette:


clalancette:
This needs to go along with https://github.com/ros-infrastructure/ros_buildfarm_config/pull/188

sloretz (approved):
Looks like the same process as https://github.com/ros/rosdistro/pull/18151

PR #27641, opened 2020-12-07
[dep] python3-git.

130s:
- arch https://www.archlinux.org/packages/extra/x86_64/python-gobject/ Note I do not see explicitly in this link that this pkgs is py3 compatible. Looking at other rosdep key for py3 pkgs for arch, this seems to be the convention?
- Debian https://packages.debian.org/search?keywords=python3-git
- Fedora https://www.rpmfind.net/linux/rpm2html/search.php?query=python3-GitPython
- gentoo https://packages.gentoo.org/packages/dev-python/GitPython. Note that the pkg name is the same one as `python-git` key. I actually haven't found explicit info about the package in this link is python3 not 2, but looking at other `gentoo` rosdep keys for py3 pkgs this seems to still work?
- Ubuntu https://packages.ubuntu.com/search?keywords=python3-git

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5852,6 +5852,12 @@ python3-gi:
   gentoo: [dev-python/pygobject]
   openembedded: [python3-pygobject@openembedded-core]
   ubuntu: [python3-gi]
+python3-git:
+  arch: [python-gitpython]
+  debian: [python3-git]
+  fedora: [python3-GitPython]
+  gentoo: [dev-python/git-python]
+  ubuntu: [python3-git]
 python3-github:
   debian: [python3-github]
   fedora: [python3-github]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1786,7 +1786,7 @@ python-gi-cairo:
   gentoo: [dev-python/pygobject]
   ubuntu: [python-gi-cairo]
 python-git:
-  arch: [python2-gobject]
+  arch: [python2-gitpython]
   debian: [python-git]
   fedora: [GitPython]
   gentoo: [dev-python/git-python]


rosdep/python.yaml
```diff
@@ -5799,6 +5799,12 @@ python3-gi:
   fedora: [python3-gobject]
   gentoo: [dev-python/pygobject]
   ubuntu: [python3-gi]
+python3-git:
+  arch: [python-gobject]
```
clalancette, line 5803: Is this really the right package for arch?  I would guess that this is `python-gitpython` or something like that, but I'm not sure.

clalancette (dismissed):

rosdep/python.yaml
130s: I feel you. That looks a bit strange indeed. For Py2 version that was added in https://github.com/ros/rosdistro/pull/16946 it's the same package. As in my comment in this PR, I couldn't find an info about Py2 and 3 differentiation for Arch. I'm not entirely sure.

nuclearsandwich: I somehow missed that when I reviewed #16946 or I would have asked there. I'm not sure why the GObject bindings for python would include the gitpython module so that seems either like a mistake or a very unusual circumstance. In any case, the Arch Linux package https://www.archlinux.org/packages/community/any/python-gitpython/ is the one that matches the library being provided on other platforms so I think this should change to use that.

130s: Makes sense. Then I pushed 2 commits, one for updating Py3 Arch, and the other for fixing Py2 Arch.

jacobperron (approved):

hidmic:
@clalancette ping.

tfoote (approved):

PR #27702, opened 2020-12-09
add flask-restplus from pip

Minipada:
**Added package**: flask-restplus
**Pypi Link**: https://pypi.org/project/flask-restplus/
**Description**: Fully featured framework for fast, easy and documented API development with Flask
**Usage**: Making an API interacting with ROS. Flask is quite famous but for APIs, restplus gives some nice features such as integrated modelling, marshalling, swaggers and more. It is very easy to use if you already know some flask.

I added it as pip since there is no deb for it.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1626,10 +1626,6 @@ python-flask-restful:
         packages: [flask-restful]
     xenial: [python-flask-restful]
     yakkety: [python-flask-restful]
-python3-flask-restplus-pip:
-  ubuntu:
-    pip:
-      packages: [flask-restplus]
 python-flatdict-pip:
   ubuntu:
     pip:
@@ -5774,6 +5770,10 @@ python3-flake8:
     '*': ['python%{python3_pkgversion}-flake8']
     '7': null
   ubuntu: [python3-flake8]
+python3-flask-restplus-pip:
+  ubuntu:
+    pip:
+      packages: [flask-restplus]
 python3-future:
   debian: [python3-future]
   fedora: [python3-future]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1626,6 +1626,10 @@ python-flask-restful:
         packages: [flask-restful]
     xenial: [python-flask-restful]
     yakkety: [python-flask-restful]
+python-flask-restplus-pip:
```
nuclearsandwich, line 1629:
```suggestion
python3-flask-restplus-pip:
```

Per our [contribution guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1) this should be changed to `python3-flask-restplus-pip`.

nuclearsandwich (changes requested):
A small change is recommended to conform to contribution guidelines.

Minipada:
@nuclearsandwich I get your change, but what about old version. If I want to make it work on melodic for example. Will it install it properly too?

nuclearsandwich:
> I get your change, but what about old version. If I want to make it work on melodic for example, which use python2. Will it install it properly too?

So the short answer is yes. The pip installer and pip itself does not make any distinction between python2 or python3 when fetching packages from pypi.

nuclearsandwich (dismissed):

nuclearsandwich:
@Minipada it's my fault, I always forget that when I suggest inline it will break the lexical sorting. The definition needs to be moved down with the rest of the python3 definitions and then this will be good to go.

Minipada:
Thanks for the precision.
It's mine, I just saw the suggestion and went with it. I did PR in the past, so I should have checked too. It should be right now.

nuclearsandwich (approved):
Thanks for iterating!

jacobperron (approved):

PR #27739, opened 2020-12-10
fix inconistent naming of the package behaviortree_cpp_v3

facontidavide:
In the past I used inconsistent names for the library (my bad) and I think it is finally time to try fixing that.

Is there anything else that I need to change to make this work?

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -315,7 +315,7 @@ repositories:
       url: https://gitlab.com/autowarefoundation/autoware.auto/autoware_auto_msgs.git
       version: master
     status: developed
-  behaviortree_cpp:
+  behaviortree_cpp_v3:
     doc:
       type: git
       url: https://github.com/BehaviorTree/BehaviorTree.CPP.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -839,7 +839,7 @@ repositories:
       url: https://github.com/BehaviorTree/BehaviorTree.CPP.git
       version: master
     status: developed
-  behaviotree_cpp_v3:
+  behaviortree_cpp_v3:
     doc:
       type: git
       url: https://github.com/BehaviorTree/BehaviorTree.CPP.git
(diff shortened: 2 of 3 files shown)


mabelzhang:
The failing check is indicating you have some formatting issues.
```
FAIL: test.rosdistro_verify_test.test_verify_files_identical
```
which comes from [this test](https://github.com/ros/rosdistro/blob/46d4535f7da0588af144c20f6652d9f6949fcaee/test/rosdistro_verify_test.py#L37).

If I run `rosdistro_reformat index.yaml` locally, it auto-corrects the three files to:
```
--- file:///data/ws/ros/rosdistro/foxy/distribution.yaml
+++ loaded-and-saved
@@ -323,4 +323,2 @@
     release:
-      packages:
-      - behaviortree_cpp_v3
       tags:

--- file:///data/ws/ros/rosdistro/melodic/distribution.yaml
+++ loaded-and-saved
@@ -847,4 +847,2 @@
     release:
-      packages:
-      - behaviortree_cpp_v3
       tags:

--- file:///data/ws/ros/rosdistro/noetic/distribution.yaml
+++ loaded-and-saved
@@ -297,4 +297,2 @@
     release:
-      packages:
-      - behaviortree_cpp_v3
       tags:
```

So I think you'll need to run that command and double-check the result looks right to you.

I don't know what the right answer is regarding whether there's something else you need to do.
Looks to me the package names are already correct, so maybe it's okay? Can someone else chime in?
http://repositories.ros.org/status_page/ros_melodic_default.html
http://repositories.ros.org/status_page/ros_noetic_default.html
http://repo.ros2.org/status_page/ros_foxy_default.html

clalancette:
> So I think you'll need to run that command and double-check the result looks right to you.
> I don't know what the right answer is regarding whether there's something else you need to do.
> Looks to me the package names are already correct, so maybe it's okay? Can someone else chime in?

Yep, that seems about the right thing to do.  Once that is done, we should be able to merge it.

One thing that I'll say in general is that making a PR across multiple ROS distributions is hard to get in.  The reason is that all of the distributions need to *not* be in a sync freeze, and they are all on different schedules.  So it can be hard to find a window in which to merge it.  @facontidavide It is up to you whether you want to split it up, but splitting it up into 2 or 3 pull requests will make it easier to get it in.

facontidavide:
@clalancette . ok, I will split this!

PR #27780, opened 2020-12-16
[dep] Add python3-sh

130s:
- Debian https://packages.debian.org/stretch/python3-sh
- Fedora: Not sure.
   - Can't find a generic webpage on https://src.fedoraproject.org/
   - Other py3 pkg, e.g. python3-shapely, I do see py2 page https://src.fedoraproject.org/rpms/python-shapely but no py3 page found. So I assume python-sh follows the same custom? I'm not sure.
   - There's python-sh in the rosddep key already.
- Gentoo https://packages.gentoo.org/packages/dev-python/sh
- Ubuntu: https://packages.ubuntu.com/bionic/python3-sh

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6765,6 +6765,11 @@ python3-sexpdata:
     xenial:
       pip:
         packages: [sexpdata]
+python3-sh:
+  debian: [python3-sh]
+  fedora: [python3-sh]
+  gentoo: [dev-python/sh]
+  ubuntu: [python3-sh]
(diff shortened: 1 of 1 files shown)


mabelzhang (approved):
I think you're right about Fedora. I see python3-sh for Fedora 32 here.
https://fedora.pkgs.org/32/fedora-x86_64/python3-sh-1.12.14-15.fc32.noarch.rpm.html

jacobperron (approved):

PR #27782, opened 2020-12-17
Add pathlib2 to python yaml

tsaoyu:
Add pathlib2 on the following distro

Arch https://www.archlinux.org/packages/community/any/python2-pathlib2/
Gentoo https://packages.gentoo.org/packages/dev-python/pathlib2
Ubuntu https://packages.ubuntu.com/bionic/python-pathlib2
Debian https://packages.debian.org/stretch/python-pathlib2

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3018,6 +3018,11 @@ python-pathlib:
   debian: [python-pathlib]
   gentoo: [dev-python/pathlib]
   ubuntu: [python-pathlib]
+python-pathlib2:
+  arch: [python2-pathlib2]
+  debian: [python-pathlib2]
+  gentoo: [dev-python/pathlib2]
+  ubuntu: [python-pathlib2]
 python-pathos-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3022,7 +3022,10 @@ python-pathlib2:
   arch: [python2-pathlib2]
   debian: [python-pathlib2]
   gentoo: [dev-python/pathlib2]
-  ubuntu: [python-pathlib2]
+  ubuntu: 
+    '*': [python-pathlib2]
+    xenial: null
+  fedora: [python-pathlib2]
 python-pathos-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3021,11 +3021,11 @@ python-pathlib:
 python-pathlib2:
   arch: [python2-pathlib2]
   debian: [python-pathlib2]
+  fedora: [python-pathlib2]
   gentoo: [dev-python/pathlib2]
   ubuntu: 
     '*': [python-pathlib2]
     xenial: null
-  fedora: [python-pathlib2]
 python-pathos-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3023,7 +3023,7 @@ python-pathlib2:
   debian: [python-pathlib2]
   fedora: [python-pathlib2]
   gentoo: [dev-python/pathlib2]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3018,6 +3018,11 @@ python-pathlib:
   debian: [python-pathlib]
   gentoo: [dev-python/pathlib]
   ubuntu: [python-pathlib]
+python-pathlib2:
+  arch: [python2-pathlib2]
+  debian: [python-pathlib2]
+  gentoo: [dev-python/pathlib2]
+  ubuntu: [python-pathlib2]
```
mabelzhang, line 3025: I think xenial needs to be marked out with `xenial: null` since that package doesn't exist for xenial

mabelzhang (commented):
I think a package exists for Fedora too? https://src.fedoraproject.org/rpms/python-pathlib2

rosdep/python.yaml
tsaoyu: Yes, `python-pathlib2` wasn't released on xenial. Would it better to install from `pip` rather than the package manager? e.g.
```yaml
ubuntu:
      pip:
           pathlib2
```

jacobperron: No, we prefer to use the OS package manager. We should either mark Xenial with `null` or add a separate top-level key for installing from pip (e.g. `python-pathlib2-pip`). I think marking xenial is enough, unless you really want an option for Xenial.

hidmic:
@tsaoyu friendly ping

tsaoyu:
Sorry for the delay. Please comment/review my changes.

rosdep/python.yaml
```diff
@@ -3018,6 +3018,14 @@ python-pathlib:
...
   gentoo: [dev-python/pathlib]
   ubuntu: [python-pathlib]
+python-pathlib2:
+  arch: [python2-pathlib2]
+  debian: [python-pathlib2]
+  gentoo: [dev-python/pathlib2]
+  ubuntu: 
+    '*': [python-pathlib2]
+    xenial: null
+  fedora: [python-pathlib2]
```
jacobperron, line 3028: Put this in alphabetical order. It should fix CI.

ivanpauno:
I tried to understand the failing check output, but it's not clear to me why it failed.
Maybe @jacobperron can check too

rosdep/python.yaml
```diff
@@ -3018,6 +3018,14 @@ python-pathlib:
...
   gentoo: [dev-python/pathlib]
   ubuntu: [python-pathlib]
+python-pathlib2:
+  arch: [python2-pathlib2]
+  debian: [python-pathlib2]
+  fedora: [python-pathlib2]
+  gentoo: [dev-python/pathlib2]
+  ubuntu: 
+    '*': [python-pathlib2]
+    xenial: null
```
nuclearsandwich, line 3028: Fix lexical sorting of Ubuntu definitions

```suggestion
  ubuntu: 
    xenial: null
    '*': [python-pathlib2]
```

nuclearsandwich (approved):
The CI failure is caused by a lexical sort issue.

The test output is voluminous but the failure indicated is in"rosdep_formatting_test"

rosdep/python.yaml
ivanpauno: I'm a bit confused because the current order (first '*', then xenial) is already being used in other places https://github.com/ros/rosdistro/blob/074cafd4261f88aaab8b11c4592b2b017732e5c3/rosdep/python.yaml#L216-L217.

clalancette: I don't think this is the issue.  I think the problem is trailing whitespace after `ubuntu:`.  I'll push a fix.

clalancette (approved):

PR #27819, opened 2020-12-20
Add rosdep rules for python3-websockets package

Goldob:
Added rules for `python3-websockets` in Debian and Ubuntu. This package provides a [library](https://websockets.readthedocs.io/en/stable/) for creating WebSocket servers and clients. Existing package `python3-websocket` is [client-only](https://packages.ubuntu.com/xenial/python3-websocket).

My particular use case involves building a web interface for controlling the robot. I need a websocket server to receive commands from user's browser.

Package listings:
- Debian: https://packages.debian.org/search?keywords=python3-websockets
- Ubuntu: https://packages.ubuntu.com/search?keywords=python3-websockets

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5225,7 +5225,7 @@ python-websocket:
   debian: [python-websocket]
   fedora: [python-websocket-client]
   gentoo: [dev-python/websocket-client]
-  openembedded: [python3-websockets@meta-python]
+  openembedded: [python3-websocket-client@meta-python]
   ubuntu:
     artful: [python-websocket]
     artful_python3: [python3-websocket]
@@ -7015,11 +7015,12 @@ python3-websocket:
   debian: [python3-websocket]
   fedora: [python3-websocket-client]
   gentoo: [dev-python/websocket-client]
-  openembedded: ['python3-websockets@meta-python']
+  openembedded: [python3-websocket-client@meta-python]
   rhel: ['python%{python3_pkgversion}-websocket-client']
   ubuntu: [python3-websocket]
 python3-websockets:
   debian: [python3-websockets]
+  openembedded: [python3-websockets@meta-python]
   ubuntu: [python3-websockets]
 python3-werkzeug:
   arch: [python-werkzeug]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5225,7 +5225,7 @@ python-websocket:
   debian: [python-websocket]
   fedora: [python-websocket-client]
   gentoo: [dev-python/websocket-client]
-  openembedded: [python3-websocket-client@meta-python]
+  openembedded: [python3-websockets@meta-python]
   ubuntu:
     artful: [python-websocket]
     artful_python3: [python3-websocket]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -7018,6 +7018,9 @@ python3-websocket:
   openembedded: ['python3-websockets@meta-python']
   rhel: ['python%{python3_pkgversion}-websocket-client']
   ubuntu: [python3-websocket]
+python3-websockets:
+  debian: [python3-websockets]
+  ubuntu: [python3-websockets]
```
jacobperron, line 7023: Please add entries for 

- arch: https://archlinux.org/packages/community/x86_64/python-websockets/
- fedora: https://src.fedoraproject.org/rpms/python-websockets
- gentoo: https://packages.gentoo.org/packages/dev-python/websockets

jacobperron (changes requested):

rosdep/python.yaml
```diff
@@ -7018,6 +7018,9 @@ python3-websocket:
   openembedded: ['python3-websockets@meta-python']
```
shr-project, line 7018: Can you please update this to ['python3-websocket-client@meta-python']
https://layers.openembedded.org/layerindex/recipe/97740/
I didn't realize that websocket and websockets are 2 different packages - or I can fix it my next PR as it's used also in python-websocket mapping above.

rosdep/python.yaml
shr-project: Please add openembedded ['python3-websockets@meta-python'] as well https://layers.openembedded.org/layerindex/recipe/82184/

rosdep/python.yaml
Goldob: Done in 3102f3c.

rosdep/python.yaml
Goldob: Both done.

rosdep/python.yaml
```diff
@@ -5225,7 +5225,7 @@ python-websocket:
   debian: [python-websocket]
   fedora: [python-websocket-client]
   gentoo: [dev-python/websocket-client]
-  openembedded: [python3-websockets@meta-python]
+  openembedded: [python3-websocket-client@meta-python]
```
shr-project, line 5228: This one should actually be:
'${PYTHON_PN}-websocket-client@meta-python'

shr-project (approved):
Thanks for quick update on openembedded, one more change inline (but it can wait for more changes with https://github.com/ros/rosdistro/pull/27826)

rosdep/python.yaml
Goldob: Will do. This is only for `python-websocket`? `python3-websocket` is OK?

shr-project: yes only here (and the apostrophes are important)

PYTHON_PN is either python- or python3- based on selected python version, so when both versions are available then we use this variable (python3-websockets was an exception because there is no recipe for python2 version), but now there are both for websocket-client.

rosdep/python.yaml
hidmic: @shr-project @Goldob I would not make changes to unrelated rules. 

To modify `python-websocket` and `python3-websocket`, please open a different PR.

shr-project: I'm sorry @Goldob

separate PR for OE websocket is already created https://github.com/ros/rosdistro/pull/27826 just a draft now, waiting to collect more changes before submitting it.

Goldob: No worries @shr-project. I reverted the changes.

hidmic (approved):
LGTM, any other thought @jacobperron?

jacobperron (approved):
LGTM :+1:

PR #27824, opened 2020-12-22
release realsense2_camera version 3.1.2 for eloquent

doronhi:
Requires librealsense2 version 2.40.0 as in #27823


clalancette:
Eloquent is now end-of-life, so there will be no more updates for it. See https://discourse.ros.org/t/new-packages-for-eloquent-2020-12-14/17907 and the timeline at https://index.ros.org/doc/ros2/Releases/#releases . So I'll close this out.

PR #27827, opened 2020-12-22
novatel_oem7_driver: 2.0.0-1 in noetic/distribution.yaml

novatel-applications-engineering:
Creating pull request manually, due to bloom auth token failure.


hidmic (commented):
@novatel-applications-engineering we do prefer releases via `bloom-release`. I won't block on it, but it'd be best if at least we had a description that conveys the same amount of information. For documentation purposes.

novatel-applications-engineering:
Understood - thank you.

ivanpauno (approved):

PR #27848, opened 2020-12-28
release librealsense2 version 2.40.0 for eloquent

doronhi:


doronhi:
Will fix and resubmit.

PR #27849, opened 2020-12-28
Release realsense2 camera 3.1.2 eloquent

doronhi:
depends on #27848


doronhi:
will fix and resubmit

PR #27850, opened 2020-12-28
Added 20.04 Focal version of python-setuptools.

AzaelCicero:
Added Ubuntu 20.04 Focal support for **python-setuptools** package.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4298,6 +4298,7 @@ python-setuptools:
   ubuntu:
     artful: [python-setuptools]
     bionic: [python-setuptools]
+    focal: [python3-setuptools]
     lucid: [python-setuptools]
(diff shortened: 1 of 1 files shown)


hidmic (commented):
@AzaelCicero there's [`python3-setuptools`](https://github.com/ros/rosdistro/blob/8f7b681348ba6dee4282a56ece93c9bce99cca34/rosdep/python.yaml#L6750) already.

AzaelCicero:
@hidmic Yes there is such package, yet it lacks entry for Ubuntu **20.04 focal**. I added one line to existing package which allows install via apt package **python3-setuptools** on **focal** OS version (on new Ubuntu there is not default Python2 support).

AzaelCicero:
@hidmic Thanks for pointing it out. I think this is sufficient solution.

PR #27852, opened 2020-12-28
Add pystemd

Timple:


tfoote (changes requested):
This appears to be available as system packages on at least ubuntu: https://packages.ubuntu.com/focal/python3-pystemd and debian: https://packages.debian.org/unstable/python3-pystemd

Following the contributing guide, those should be used and not the pip rules.

Are you actually using both python2 and python3 versions? If not I'd recommend not adding the one that you're not using either. We'd rather not take effort to maintain unused rules. If someone else needs it in the future they can contribute it as needed.

Timple:
Actually we're using python2 currently but are in planning on migrating to python3 around spring.

You are right that there are python3 apt binaries available, I didn't look that far ahead yet but am pleased to hear since we also prefer apt over pip.

Shall I extend the PR with the python3 version?

clalancette:
> Shall I extend the PR with the python3 version?

Yes please.  We always prefer to use the system packaged versions where available, so in this case please point the `python3` key to the system packages.

rosdep/python.yaml
```diff
@@ -3633,6 +3633,13 @@ python-pysnmp:
   fedora: [pysnmp]
   gentoo: [dev-python/pysnmp]
   ubuntu: [python-pysnmp4]
+python-pystemd:
```
sloretz, line 3636: Confusingly, while I understand this is meant to be used with Python 2, the contributing guide says this key should be named [`python3-systemd-pip`](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1).

Maybe this should be named `python-systemd-pip` instead? @clalancette thoughts?

rosdep/python.yaml
```diff
@@ -6496,6 +6503,19 @@ python3-pyqtgraph:
   fedora: [python3-pyqtgraph]
   gentoo: [dev-python/pyqtgraph]
   ubuntu: [python3-pyqtgraph]
+python3-pystemd:
```
Timple, line 6506: I am actually happy to leave all the pip stuff out of this target as we won't be using it. 
Not sure how to declare this target then.

rosdep/python.yaml
Timple: That is confusing indeed!

Happy to convert to `python-systemd-pip`

tfoote: Please convert it to `python3-systemd-pip` The name will not effect which version of python it uses. But as we look towards the future all default platforms will be using python3* and so we want it to be sorted into the right place.

rosdep/python.yaml
tfoote: So after my comment above about the naming, The naming is precisely designed to prevent us from overlooking a conflicting rule like this one. 

In thinking about it I believe that this key as written is actually what you want to use. The pip rules will install to your default python version for the rosdistro. Which for the older platforms Melodic and older is Python 2. If you're using Python 2 with Noetic or newer the default rosdep resolutions will not work.

tfoote (changes requested):
So a little confusingly I think that the `python3-rosdep` key as written is all that you need. You don't include any description of where and how you're using this so I'm having to guess about what you're looking to resolve on what platforms.  The python3 in the name isn't going to change how it actually resolves for pip environments.

rosdep/python.yaml
Timple: I get the whole conflicting part with future updates now I guess.

Rewrote the PR to comply!

rosdep/python.yaml
```diff
@@ -6496,6 +6496,16 @@ python3-pyqtgraph:
...
+  debian: [python3-pystemd]
+  ubuntu: [python3-pystemd]
```
nuclearsandwich, line 6501: The `python3-pystemd` package is only available in Ubuntu Focal and newer. https://packages.ubuntu.com/focal/python3-pystemd

```suggestion
  ubuntu: 
    '*': [python3-pystemd]
    bionic: null
    xenial: null
```

rosdep/python.yaml
```diff
@@ -6496,6 +6496,16 @@ python3-pyqtgraph:
   fedora: [python3-pyqtgraph]
   gentoo: [dev-python/pyqtgraph]
   ubuntu: [python3-pyqtgraph]
+python3-pystemd:
+  debian: [python3-pystemd]
```
nuclearsandwich, line 6500: The `python3-pystemd` package is only available in Debian Bullseye, which is the current testing / upcoming stable distribution.

```suggestion
  debian:
    bullseye: [python3-pystemd]
```

nuclearsandwich (dismissed):
@Timple thanks for iterating as we dial in this rosdep change.

I added a couple of suggested based on the availability of the package in various distributions.

Timple:
Thanks for helping out!

jacobperron (approved):
LGTM

wjwwood (approved):

wjwwood:
I think all of the issues pointed out by @tfoote and @nuclearsandwich were resolved? But I'll wait for a 👍 from them to merge.

tfoote (approved):
Sorry I was unclear in my guidance for the specific rules for the older platforms. The pip rules can specifically be used for the older platforms under the same key because they won't collide with an existing system package. And then that key can be used from source on the older platforms.

I merged and pushed a rebase to simplify it to one key.

Timple:
Alrighty. So now in all cases we can simply depend on `python3-pystemd` regardless of python version used by ROS.

I'm good :+1:

PR #27854, opened 2020-12-28
librealsense2 v2.41.0

doronhi:
rename package librealsense to librealsense2.
upgrade to version 2.41.0

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1493,23 +1493,22 @@ repositories:
       url: https://github.com/ros2-gbp/libg2o-release.git
       version: 2020.5.29-1
     status: maintained
-  librealsense:
+  librealsense2:
     doc:
       type: git
       url: https://github.com/IntelRealSense/librealsense.git
-      version: ros2debian
+      version: master
     release:
-      packages:
-      - librealsense2
       tags:
         release: release/foxy/{package}/{version}
-      url: https://github.com/ros2-gbp/librealsense-release.git
-      version: 2.34.0-3
+      url: https://github.com/IntelRealSense/librealsense2-release.git
+      version: 2.41.0-1
     source:
+      test_pull_requests: true
       type: git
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -1493,23 +1493,22 @@ repositories:
       url: https://github.com/ros2-gbp/libg2o-release.git
       version: 2020.5.29-1
     status: maintained
-  librealsense:
+  librealsense2:
```
ivanpauno, line 1496: Renaming a key doesn't look like a good idea, as all packages using the previous key will start failing, maybe you want to create a new key?

ivanpauno (dismissed):

foxy/distribution.yaml
clalancette: I'm not 100% sure of this, but I *think* that this will be OK.  Previously there was a `packages` section that contained the `librealsense2` package; that happens when a repository contains packages in subdirectories.  In that case, the "top-level" key here is mostly ignored, and only the package names matter.

In the case where a repository just contains a single top-level package, the key name here is just the name of the package (and there is no `packages` section).  Since this new name matches the old name in the `packages` section, I think we will be OK.

But pinging @nuclearsandwich and @tfoote for confirmation before I merge this.

jacobperron: I think this is okay too.

jacobperron (approved):

foxy/distribution.yaml
tfoote: Confirmed this key is only for the repository and doesn't effect the package. However since this is a single package repository it simplifies the list as it assumes a single length package list of the name of the repository so the results will be the same.

tfoote (approved):

doronhi:
Thanks. When is the next release planned, please?

jacobperron:
I'll likely do a Foxy sync next week.

PR #27855, opened 2020-12-28
realsense2_camera version 3.1.3 for foxy

doronhi:
Add realsense2_camera for foxy, version 3.1.3
Remove old name: ros2_intel_realsense
This PR depends on #27854

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2556,6 +2556,26 @@ repositories:
       url: https://github.com/ros2/rcutils.git
       version: foxy
     status: maintained
+  realsense2_camera:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


ivanpauno (changes requested):
Removing the old key doesn't seem acceptable, as packages using it will start failing.
The new key looks ok.

jacobperron:
Holding for [sync](https://discourse.ros.org/t/preparing-for-foxy-sync-2020-01-07/18268)

doronhi:
> Removing the old key doesn't seem acceptable, as packages using it will start failing.
> The new key looks ok.

Restored the previously removed package ros2_intel_realsense.

ivanpauno (approved):

PR #27894, opened 2021-01-05
add dgps_ros package to kinetic version

qinyunchuan:
dgps_ros is a ROS package used for differential global position system, with no external dependency

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2628,15 +2628,15 @@ repositories:
       type: git
       url: https://gitee.com/rivercloud/dgps_ros
       version: master
-    source:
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This looks good. Would you consider adding a source stanza so that it can work with the source build toolchains? And a doc entry to generate documentation for the website?

And a maintenance status would also be recommended.

qinyunchuan:
source and doc entries  added

tfoote (approved):
Thanks this looks good. We'll land it immediately after the [sync](https://discourse.ros.org/t/preparing-for-kinetic-sync-2021-01-08/18347) tomorrow.

PR #27906, opened 2021-01-06
Add cmake_common_scripts repo

Levi-Armstrong:


ahcorde (approved):

Levi-Armstrong:
I had to manually create the PR because bloom-release kept giving me the same error when trying to create the oauth token. I have since manually created the token and added it the `~/.config/bloom` file so hopefully on the next release it will not have an issue. 

``` bash
HTTP Error 404: Not Found (https://api.github.com/authorizations): None

This sometimes fails when the username or password are incorrect, try again?
```

Levi-Armstrong:
Is this ready to be merged?

clalancette (changes requested):
I have to say that the name of this package seems too generic to me.  I'm sure that they are common for you, but in the broader context of a ROS distribution they aren't really common.  I think some kind of prefix (`ros_industrial_cmake_scripts`), or something along those lines, would make it a lot more clear.

Levi-Armstrong:
> I have to say that the name of this package seems too generic to me. I'm sure that they are common for you, but in the broader context of a ROS distribution they aren't really common. I think some kind of prefix (`ros_industrial_cmake_scripts`), or something along those lines, would make it a lot more clear.

I want to check if this is absolutely necessary because a rename will cause us to have to update 15-20 package in the opensource along with our internal projects? I want to avoid causing issues on active projects if possible.

Levi-Armstrong:
@clalancette What are your thought? If the changes needs to be made I will try and get everything updated over the weekend to limit the disruption to projects.

Levi-Armstrong:
@ahcorde, @clalancette and @sloretz Is it possible to keep the package name `cmake_common_scripts`? This will require some work to go through and update all depending packages because this is the lowest level package that all of our packages depend on.

clalancette:
My opinion is still that the name of the package needs to be changed.  It's just too generic a name.  However, I'm willing to listen to other arguments about why we should keep the name as-is.

Levi-Armstrong:
I find this very disruptive to find out that a package name needs to change at this point. If this is the case that this repository will be controlling the names of packages then there needs to be a process for requesting a name to avoid this complication.  I am not aware of any other package manager which controls the package name other than they must not conflict with an existing package name and as it stands now this does not conflict with any package name within rosdistro.

gavanderhoorn:
@clalancette wrote:
> I'm willing to listen to other arguments about why we should keep the name as-is.

what sort of other arguments are you thinking of @clalancette?

Levi-Armstrong:
> I have to say that the name of this package seems too generic to me. I'm sure that they are common for you, but in the broader context of a ROS distribution they aren't really common. I think some kind of prefix (`ros_industrial_cmake_scripts`), or something along those lines, would make it a lot more clear.

I don't think prefixing it with ros_industrial is a relevant prefix because they are not specific to ros_industrial they are specific to the software package cmake. How does adding a prefix make it more clear?

This package is a collection of cmake scripts no different than ament_cmake to make things more convenient. This package contains scripts for cmake commonly found when writing modern cmake reducing the amount cmake code written. I would argue that if you are writing modern cmake then the cmake code behind the macros within this package are common things you would do as everything should be attached to the target opposed to setting things globally. These macros keep your cmake file cleaner  by hiding the logic less prone to user error very similar to ament_cmake.

Levi-Armstrong:
Would cmake_helper_scripts be better ok?

clalancette:
> what sort of other arguments are you thinking of @clalancette?

Maybe "arguments" was the wrong term; other *opinions* is what I am looking for.

> Would cmake_helper_scripts be ok?

Given my original argument, not really.  It is still really generic.  [REP-144](https://www.ros.org/reps/rep-0144.html) lays out some guidelines.  Mostly I think this package name violates the first two "Global Rules".  But again, I am willing to listen to dissenting opinions and be overridden.

gavanderhoorn:
Could you perhaps suggest a name, based on what @Levi-Armstrong described is the purpose of the package?

There's indeed nothing really `ros_industrial` about it, so that would seem to rule out that prefix.

Some suggestions may help get this moving.

Levi-Armstrong:
@clalancette Do you mind providing more information on how it violates the "Global Rules"? As @gavanderhoorn mentioned, suggestions would help move things along.  I have provided the Global Rules for others. 

Global Rules:  
- Package names should be specific enough to identify what the package does. For example, a motion planner should not be called planner. If it implements the wavefront propagation algorithm, it might be called wavefront_planner. There's obviously tension between making a name specific and keeping it from becoming overly verbose
- Using catchall names such as utils should be avoided. They do not scope what goes into the package or what should be outside the package
- A package name should not contain ros as it is redundant. Exceptions include core packages and ROS bindings of an upstream library (e.g. moveit_ros)
- The package name should describe what the package does, not where it came from. One of ROS's goals is to develop a canonical set of tools for making robots do interesting things. Then again, as stated in the rules below, if a package is specialized by an entity (lab, company, ...), prepend the name of the entity. But once the package is commonly used, owned and maintained, that name can be dropped as the package becomes the reference
- Do not use a name that's already been taken. To check whether a name is taken, consult [2]. If you'd like your repository included in that list, see the tutorial at [3]

tfoote:
The name as it stands is way too generic. It fails both describing it's contents as well as using catchall naming elements. From the package name it defines the programming language and says "common" which doesn't say anything about what's inside of it. 

Reading the headline of the README: "This contains a collection of useful CMake macros." which still doesn't even define a scope. 

Looking inside I see content related to linting, cppcheck, coverage. test registration and running tests. There doesn't appear to be cross correlation between these except that they're in the same language and mostly relate to testing. The name should reflect what's expected to be found inside. So that as a potential user I can guess at what's inside. And as a potential contributor I can know if something I wrote would be good to add.

We have good examples of this in ROS 2 where there's [ament_cmake_cppcheck](https://index.ros.org/p/ament_cmake_cppcheck/) which makes it clear what to look for inside the package, these could be similarly broken out instead of just lumped together. This level of granularity is valuable to the community to make things highly reusable and not cause cross coupling. The filter for naming is a proxy for being able to understand the scope of the package. If you can't name it easily that suggests that you don't have a clearly defined scope.

In slightly different circumstances, aka not inside the last release, with this large collection of miscellaneous macros, are there existing packages that it would make sense to contribute to instead of extending. 

We don't require this level of granularity of developers and offer the escape valve of not having to use good names if the package is prefixed with the organization or other entity so as not to claim generality. If these packages are designed to be reused by many different organizations projects the scoping should be addressed.

Right now the only coherent scoping is by who the maintainer is and I think that for the sake of a minimal effort to deploy it prefixing with your choice of ros-i and swri would make sense. Since there's hopefully equivalents in ROS 2 already or places to contribute these functions there I hope that this package use is obviated in ROS 2. Otherwise if this was going to be something that could be used more generally it would be broken up and scoped better and or directly contributed to existing packages with related cmake functionality.

Levi-Armstrong:
@clalancette and @tfoote Before I go through and change things, would the following name be acceptable `swri_cmake_scripts`?

tfoote:
`swri_cmake_scripts` sounds good to me.

Levi-Armstrong:
@clalancette and @tfoote Thank you for the support. 

> We don't require this level of granularity of developers and offer the escape valve of not having to use good names if the package is prefixed with the organization or other entity so as not to claim generality. If these packages are designed to be reused by many different organizations projects the scoping should be addressed.

Makes sense.

----------------------

Since we are on the subject, I have an educational questions below.

>We have good examples of this in ROS 2 where there's ament_cmake_cppcheck which makes it clear what to look for inside the package, these could be similarly broken out instead of just lumped together. This level of granularity is valuable to the community to make things highly reusable and not cause cross coupling. The filter for naming is a proxy for being able to understand the scope of the package. If you can't name it easily that suggests that you don't have a clearly defined scope.

Some recent discussion have come up about leveraging Project Modularization technique for some of our repositories other than the approach you provided having everything in separate packages. It sounds like this would not be an acceptable approach to use for a package to be released for ROS use?

tfoote:
> Some recent discussion have come up about leveraging Project Modularization technique for some of our repositories other than the approach you provided having everything in separate packages. It sounds like this would not be an acceptable approach to use for a package to be released for ROS use?

There's more than one approach to modularity. I'm not familiar with a canonical 'Project Modularization' technique do you have a link to the approach so we can look at the tradeoffs? The tradeoffs of any given approach will be dependent on the use cases.

PR #27944, opened 2021-01-11
qpoases: 3.2.1-4 in 'noetic/distribution.yaml' [bloom]

Levi-Armstrong:
Increasing version of package(s) in repository `qpoases` to `3.2.1-4`:

- upstream repository: https://github.com/Levi-Armstrong/qpOASES.git
- release repository: https://github.com/ros-industrial-release/qpoases-release.git
- bloom version: `0.10.0`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4634,6 +4634,17 @@ repositories:
       url: https://github.com/ros-visualization/python_qt_binding.git
       version: melodic-devel
     status: maintained
+  qpoases:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/ros-industrial-release/qpoases-release.git
+      version: 3.2.1-4
+    source:
+      type: git
+      url: https://github.com/coin-or/qpOASES.git
+      version: master
+    status: maintained
   qt_gui_core:
     doc:
       type: git


clalancette (changes requested):
There are a few minor oddities with this package:

* The source repository doesn't agree with the (forked) one listed by bloom.  This isn't a problem in the long run, since the one listed in the distribution.yaml will always be used, but it is odd.
* The upstream source repository isn't a catkin package; it is a standard library.  I see you are overlaying the package.xml in the release repository, which is fine.
* In melodic, we have a [qpoases_vendor](https://github.com/ros/rosdistro/blob/64179eeb26834630f4b84d230dd60ec424b96644/melodic/distribution.yaml#L8970) package that was added in #22226 .  I don't really care which one we have, but I don't think we should have both.  @Levi-Armstrong can you work with @JWhitleyWork to figure out which one we should have in Noetic?

Levi-Armstrong:
>The source repository doesn't agree with the (forked) one listed by bloom. This isn't a problem in the long run, since the one listed in the distribution.yaml will always be used, but it is odd.

The reason for the fork was they tag the things as `release/v#.#.#` and this cause error when using this tag during the bloom release process. So I forked it and created a new tag that works with bloom. Is there away around this using the bloom?

> The upstream source repository isn't a catkin package; it is a standard library. I see you are overlaying the package.xml in the release repository, which is fine.

I followed the third party release tutorial, should I have done something different?

> In melodic, we have a qpoases_vendor package that was added in #22226 . I don't really care which one we have, but I don't think we should have both. @Levi-Armstrong can you work with @JWhitleyWork to figure out which one we should have in Noetic?

@JWhitleyWork Do you plan to do a noetic release of qpOASIS? If not, @clalancette should I just rename this to match what is in melodic?

JWhitleyWork:
@Levi-Armstrong: https://github.com/ros/rosdistro/pull/27998

I'm happy to continue maintaining this. Please let me know if you have an issue with it.

Levi-Armstrong:
> @Levi-Armstrong: #27998
> 
> I'm happy to continue maintaining this. Please let me know if you have an issue with it.

@JWhitleyWork  Would you be willing to do a noetic and foxy release?

Levi-Armstrong:
This has been replaced by #27998

PR #27969, opened 2021-01-13
dgps_ros: 1.0.0-4 in 'kinetic/distribution.yaml'

qinyunchuan:
increasing version of dgps_ros  to 1.0.0-4, to fix the building problem

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2626,7 +2626,7 @@ repositories:
   dgps_ros:
     doc:
       type: git
-      url: https://github.com/qinyunchuan/dgps_ros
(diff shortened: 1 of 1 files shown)


clalancette (commented):
This doesn't only update the package, it also moves the URL to a new place.  Why is that?

qinyunchuan:
Because there are many timeouts in the Nosetests report  when connecting to the original repository,  I made a clone on github.

clalancette:
> Because there are many timeouts in the Nosetests report when connecting to the original repository, I made a clone on github.

OK, thanks.  Since you are the original contributor for this, it is OK to do this.

PR #27973, opened 2021-01-13
add gpp-package

dorezyuk:
Hey everyone,

I would like to add my gpp-package to the noetic distro

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2161,6 +2161,10 @@ repositories:
       version: kinetic-devel
     status: maintained
   gpp:
+    doc:
+      type: git
+      url: https://github.com/dorezyuk/gpp.git
+      version: master
     source:
       type: git
       url: https://github.com/dorezyuk/gpp.git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2160,6 +2160,12 @@ repositories:
       url: https://github.com/ros-visualization/gl_dependency.git
       version: kinetic-devel
     status: maintained
+  gpp:
+    source:
+      type: git
+      url: https://github.com/dorezyuk/gpp.git
+      version: master
+    status: maintained
   gps_umd:
     doc:
       type: git


noetic/distribution.yaml
```diff
@@ -2160,6 +2160,12 @@ repositories:
       url: https://github.com/ros-visualization/gl_dependency.git
       version: kinetic-devel
     status: maintained
+  gpp:
```
clalancette, line 2163: I have to say that the name of this package is a little too short for my liking.  In the wide world of robotics, it would be hard to distinguish exactly what this package did.  I'd much prefer "global_planner_pipieline", but just renaming the key here wouldn't be enough.  The name of the package in the source repository would also have to be updated.

dorezyuk: agreed - the name is a little too ambiguous. however, it's not associated with any package: the repository has no meta-package and the package-names are gpp_plugin and gpp_interface - which i hope is less easy to mixup. I could nonetheless rename the repository to global_planner_pipeline. I fear that doing so might be more confusing for users looking for sources of their gpp_<bla> packages. What do you think?

dorezyuk: any verdict?

ivanpauno: > and the package-names are gpp_plugin and gpp_interface

I agree that the repo name is "too short", but considering that the package names are longer it seems to be okey to me.

ivanpauno: @clalancette friendly ping

clalancette: Yeah, good point.  It's OK with me, feel free to proceed with this PR.

noetic/distribution.yaml
```diff
@@ -2160,6 +2160,12 @@ repositories:
...
+    source:
```
ivanpauno, line 2164: @dorezyuk could you add a `doc` entry?

ivanpauno (approved):

noetic/distribution.yaml
dorezyuk: sure

ivanpauno:
Thanks @dorezyuk

PR #27976, opened 2021-01-14
rosdep: add ansible-runner as python-ansible-runner-pip

hocjordan:
This PR adds a rosdep key for the pip package `ansible-runner`. I am building a ros package that performs internal testing and benchmarking, and using ansible-runner to run an ansible playbook as part of the setup process.

The pip package page is here: https://pypi.org/project/ansible-runner/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -473,6 +473,19 @@ python-annoy-pip:
   ubuntu:
     pip:
       packages: [annoy]
+python-ansible-runner-pip:
+  debian:
+    pip:
+      packages: [ansible-runner]
+  fedora:
+    pip:
+      packages: [ansible-runner]
+  gentoo:
+    pip:
+      packages: [ansible-runner]
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -473,6 +473,19 @@ python-annoy-pip:
   ubuntu:
     pip:
       packages: [annoy]
+python-ansible-runner-pip:
```
clalancette, line 476: Can you please change this to `python3-ansible-runner-pip`, according to our [contribution guidelies](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1).

clalancette (changes requested):

rosdep/python.yaml
hocjordan: Thanks for the review, and apologies for misunderstanding. Should be fixed now!

clalancette (approved):

ivanpauno (approved):

PR #27983, opened 2021-01-14
rx_service_tools: 1.0.1-1 in 'melodic/distribution.yaml' [bloom]

Timple:
Increasing version of package(s) in repository `rx_service_tools` to `1.0.1-1`:

- upstream repository: https://github.com/nobleo/rx_service_tools.git
- release repository: https://github.com/nobleo/rx_service_tools-release.git
- bloom version: `0.10.0`

## rx_service_tools

```
* Roslaunch compatible by allowing unkown arguments (#1 <https://github.com/nobleo/rx_service_tools/issues/1>)
  Only parse known arguments as roslaunch tends to add arguments like __name:= and __log:=
* feat: Interactive service server GUI
  Based on the actionlib/axserver.py. Works exactly the same but for
  services instead of actions.
* Contributors: Felix Widmaier, Tim Clephas
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -11590,6 +11590,21 @@ repositories:
       url: https://github.com/PickNikRobotics/rviz_visual_tools.git
       version: melodic-devel
     status: developed
+  rx_service_tools:
+    doc:
+      type: git
+      url: https://github.com/nobleo/rx_service_tools.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/nobleo/rx_service_tools-release.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/nobleo/rx_service_tools.git
+      version: master
+    status: maintained
   rxcpp_vendor:
     doc:
       type: git


clalancette (commented):
This package is a fork of the existing `rx_service_tools` from @Timple, so before going forward with this I'd want to make sure that is OK with the original author.

Further, could you explain briefly how this is different/better than the ability to call services from the rqt tools?

clalancette:
> This package is a fork of the existing `rx_service_tools` from @Timple, so before going forward with this I'd want to make sure that is OK with the original author.

Oh, ha.  Sorry, @Timple you *are* the author.  Why the fork?

> Further, could you explain briefly how this is different/better than the ability to call services from the rqt tools?

I'm still interested in this.

Timple:
> Why the fork?

I wanted to transfer maintenance to the company. Probably didn't realise at the time a github transfer would have been nicer.

> how this is different/better than the ability to call services from the rqt tools?

I didn't know rqt tools had these abilities. A quick search shows https://github.com/ros-visualization/rqt_service_caller but that does the exact opposite. Is there an rqt tool to answer service calls?

Timple:
I've looked into it. But rqt only provides methods of calling services. This is about answering services. Identical to `axserver.py` from actionlib.

clalancette:
> I didn't know rqt tools had these abilities. A quick search shows https://github.com/ros-visualization/rqt_service_caller but that does the exact opposite. Is there an rqt tool to answer service calls?

I see.  I misunderstood what this package did.  No problem then.

> I wanted to transfer maintenance to the company. Probably didn't realise at the time a github transfer would have been nicer.

Yeah, that would be nicer.  In particular, when we perform due diligence on incoming packages, we vastly prefer a non-forked repository.  One of the reasons is to ensure that this isn't a "hostile" takeover of a package without the original owners knowledge.  So in this case, if you don't mind transferring it over, it would just make our lives easier.

Timple:
Done

nuclearsandwich:
It seems like the issues with this release are resolving but we're still going to hold it temporarily for the Melodic [sync](https://discourse.ros.org/t/preparing-for-melodic-sync-2021-01-22/18607).

clalancette (approved):

PR #27986, opened 2021-01-14
[rosdep] python3-numpy-stl

reinzor:
Debian: https://packages.debian.org/bullseye/numpy-stl
Ubuntu: https://packages.ubuntu.com/bionic/python3-numpy-stl
Fedora: https://fedora.pkgs.org/33/fedora-x86_64/python3-numpy-stl-2.11.2-2.fc33.x86_64.rpm.html

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6177,8 +6177,8 @@ python3-numpy-stl:
   debian: [python3-numpy-stl]
   fedora: [python3-numpy-stl]
   ubuntu: 
-    xenial: [null]
     '*': [python3-numpy-stl]
+    xenial: [null]
 python3-opencv:
   debian:
     buster: [python3-opencv]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6176,7 +6176,7 @@ python3-numpy:
 python3-numpy-stl:
   debian: [python3-numpy-stl]
   fedora: [python3-numpy-stl]
-  ubuntu: 
+  ubuntu:
     '*': [python3-numpy-stl]
     xenial: null
 python3-opencv:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6173,6 +6173,10 @@ python3-numpy:
   openembedded: [python3-numpy@openembedded-core]
   rhel: ['python%{python3_pkgversion}-numpy']
   ubuntu: [python3-numpy]
+python3-numpy-stl:
+  debian: [numpy-stl]
+  fedora: [python3-numpy-stl]
+  ubuntu: [python3-numpy-stl]
 python3-opencv:
   debian:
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):

rosdep/python.yaml
```diff
@@ -6173,6 +6173,10 @@ python3-numpy:
   openembedded: [python3-numpy@openembedded-core]
   rhel: ['python%{python3_pkgversion}-numpy']
   ubuntu: [python3-numpy]
+python3-numpy-stl:
+  debian: [numpy-stl]
+  fedora: [python3-numpy-stl]
+  ubuntu: [python3-numpy-stl]
```
nuclearsandwich, line 6179: This package does not appear to be available on Ubuntu Xenial which is still a few months away from end-of-support. I could bend on this if another rosdep maintainer decides that It's Fine:tm:.
```suggestion
  ubuntu: 
    xenial: [null]
    '*': [python3-numpy-stl]
```

nuclearsandwich (changes requested):
Two questions for @reinzor 

1. Both Debian and Ubuntu appear to provide python3-numpy-stl virtual packages as well as numpy-stl packages. I think it makes sense to use the same definition on both platforms. Based on the key name I'd lean toward making the debian definition `python3-numpy-stl` as well.

2. Ubuntu Xenial does not have this package so we can't use the elided definition form. I've made a suggestion which I think is accurate and compact but I've got a weird hangup with rosdep rules and am occasionally wrong.

Once those questions are addressed this looks good to me.

reinzor:
@nuclearsandwich thanks for the review. Didn't see the `python3-numpy-stl` package on debian before. Just checked it and updated the PR. Thanks for the Ubuntu suggestion for Xenial, accepted the change :).

rosdep/python.yaml
```diff
@@ -6173,6 +6173,12 @@ python3-numpy:
   openembedded: [python3-numpy@openembedded-core]
   rhel: ['python%{python3_pkgversion}-numpy']
   ubuntu: [python3-numpy]
+python3-numpy-stl:
+  debian: [python3-numpy-stl]
+  fedora: [python3-numpy-stl]
+  ubuntu: 
+    '*': [python3-numpy-stl]
+    xenial: [null]
```
ivanpauno, line 6181: should be

```suggestion
    xenial: null
```

?

ivanpauno: I think checkers are failing because of that

reinzor: @ivanpauno thanks for that!

mabelzhang (commented):
Could you add this to Debian,
```
stretch: null
```
since the package is only in Buster and Bullseye.

mabelzhang (approved):
Never mind! We don't need stretch anymore.

PR #27989, opened 2021-01-14

JudithInfineon:
I'd like ros_factorial_isprime to be indexed and documented on ros.org.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -10138,6 +10138,11 @@ repositories:
       type: git
       url: https://github.com/shadow-robot/ros_ethercat_eml.git
       version: melodic-devel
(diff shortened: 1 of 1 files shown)


clalancette:
This package has numerous problems, starting with the name that doesn't follow our conventions.  The repository is also setup strangely.  There is no LICENSE file.  The license in some of the files seems to be a non open-source license.  However, the biggest problem that I see is that this package doesn't really add much useful functionality.  It only adds a `Factorial` and an `IsPrime` function.  While those may be useful, I don't think a whole new package for them is warranted.  It would instead make much more sense for those to be in an existing library.

For all of those reasons, I'm closing this out.  Feel free to keep commenting if you disagree with this decision.

PR #27993, opened 2021-01-14
graft: 0.2.3-1 in 'noetic/distribution.yaml' [bloom]

mikeferguson:
Increasing version of package(s) in repository `graft` to `0.2.3-1`:

- upstream repository: https://github.com/ros-perception/graft.git
- release repository: https://github.com/ros-gbp/graft-release.git
- bloom version: `0.10.0`

## graft

```
* add no_delay parameter
* Contributors: Chad Rockey, Michael Ferguson
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2213,6 +2213,21 @@ repositories:
       url: https://github.com/mikeferguson/graceful_controller.git
       version: ros1
     status: developed
+  graft:
+    doc:
+      type: git
+      url: https://github.com/ros-perception/graft.git
+      version: hydro-devel
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/ros-gbp/graft-release.git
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -2213,6 +2213,21 @@ repositories:
       url: https://github.com/mikeferguson/graceful_controller.git
       version: ros1
     status: developed
+  graft:
+    doc:
+      type: git
+      url: https://github.com/ros-perception/graft.git
+      version: hydro-devel
```
clalancette, line 2220: Is it really the case that the `hydro-devel` still builds/works on Noetic?  I guess it is not impossible, it would just be surprising.

mikeferguson: It compiles and works - there are basically no outside dependencies in this package

clalancette: Heh, OK.  Thanks.

PR #28013, opened 2021-01-15
map_transformer: 1.0.0-1 in 'foxy/distribution.yaml' [bloom]

gbiggs:
Increasing version of package(s) in repository `map_transformer` to `1.0.0-1`:

- upstream repository: https://github.com/osrf/map_transformer.git
- release repository: https://github.com/ros-gbp/map_transformer-release.git
- bloom version: `0.10.0`

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1604,7 +1604,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/osrf/map_transformer.git
-      version: 1.0.0
+      version: foxy
     release:
       tags:
         release: release/foxy/{package}/{version}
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -1600,6 +1600,21 @@ repositories:
       url: https://github.com/ros2/libyaml_vendor.git
       version: foxy
     status: maintained
+  map_transformer:
+    doc:
+      type: git
+      url: https://github.com/osrf/map_transformer.git
+      version: 1.0.0
```
jacobperron, line 1607: As I understand, Git tags aren't valid for for the version in `source` and `doc` entries. Prefer to point to branch where the release was made from (or where docs are living). I think should fix CI.

gbiggs: Thanks for the tip. I've changed it to use the branch name.

PR #28041, opened 2021-01-19
Added vl53l1x_ros for melodic and noetic

deadln:
I would like to add package vl53l1x for melodic and noetic

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13642,6 +13642,7 @@ repositories:
       type: git
       url: https://github.com/okalachev/vl53l1x_ros.git
       version: master
+    status: maintained
   volksbot_driver:
     doc:
       type: git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13633,7 +13633,7 @@ repositories:
       version: master
     release:
       packages:
-        - vl53l1x
+      - vl53l1x
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/okalachev/vl53l1x_ros-release.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13683,7 +13683,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/okalachev/vl53l1x_ros-release.git
-      version: 0.4.1-1
+      version: 1.0.0-1
     source:
       type: git
       url: https://github.com/okalachev/vl53l1x_ros.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13626,6 +13626,22 @@ repositories:
       type: git
       url: https://github.com/JdeRobot/VisualStates.git
       version: master
+  vl53l1x_ros:
+    doc:
+      type: git
+      url: https://github.com/okalachev/vl53l1x_ros.git
+      version: master
+    release:
+      packages:
+        - vl53l1x
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/okalachev/vl53l1x_ros-release.git
+      version: 0.4.1-1
+    source:
+      type: git
+      url: https://github.com/okalachev/vl53l1x_ros.git
+      version: master
   volksbot_driver:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -7467,7 +7467,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/okalachev/vl53l1x_ros-release.git
-      version: 0.4.1-1
+      version: 1.0.0-1
     source:
       type: git
       url: https://github.com/okalachev/vl53l1x_ros.git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -7370,6 +7370,23 @@ repositories:
       url: https://github.com/ros-visualization/visualization_tutorials.git
       version: noetic-devel
     status: maintained
+  vl53l1x_ros:
+    doc:
+      type: git
+      url: https://github.com/okalachev/vl53l1x_ros.git
+      version: master
+    release:
+      packages:
+        - vl53l1x
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/okalachev/vl53l1x_ros-release.git
+      version: 0.4.1-1
+    source:
+      type: git
+      url: https://github.com/okalachev/vl53l1x_ros.git
+      version: master
+    status: maintained
   vrpn:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -7377,7 +7377,7 @@ repositories:
       version: master
     release:
       packages:
-        - vl53l1x
+      - vl53l1x
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/okalachev/vl53l1x_ros-release.git


melodic/distribution.yaml
```diff
@@ -13626,6 +13626,23 @@ repositories:
       type: git
       url: https://github.com/JdeRobot/VisualStates.git
       version: master
+  vl53l1x_ros:
+    doc:
+      type: git
+      url: https://github.com/okalachev/vl53l1x_ros.git
```
clalancette, line 13632: The source licensing here seems...murky.  The top-level [LICENSE](https://github.com/okalachev/vl53l1x_ros/blob/master/LICENSE) file says MIT, the [package.xml](https://github.com/okalachev/vl53l1x_ros/blob/master/vl53l1x/package.xml) matches that, and the files in [src](https://github.com/okalachev/vl53l1x_ros/blob/master/vl53l1x/src/vl53l1x.cpp) also have that header.  That's all fine.

However, the files under [lib](https://github.com/okalachev/vl53l1x_ros/tree/master/vl53l1x/lib) are a different story.  Those seem to have a dual ST proprietary license and a BSD license.  I briefly read through the ST proprietary license, and while I am not a lawyer, it does seem to say that you cannot relicense this, nor combine it with other open-source software for redistribution.  So it's not at all clear to me that dual-licensing it with BSD is allowed.

@deadln Can you explain more about where the dual licensing of the files under lib/ comes from?

clalancette (changes requested):

melodic/distribution.yaml
okalachev: Hi! This library is downloaded from the official ST website.

If it says one of the licenses available to that is BSD, then it's a free software, isn't it? So I can bundle it to my free software, why not?

If you say, that I must change the license of the package itself to BSD, I can do that.

clalancette: > If it says one of the licenses available to that is BSD, then it's a free software, isn't it? So I can bundle it to my free software, why not?

That's true, you definitely can if that is the case.  Can you provide the link to the ST site where you downloaded the code originally?

okalachev: I was a long time ago. I believe it could be this download: https://www.st.com/en/embedded-software/stsw-img007.html. But they have changed the library significantly since then.

clalancette: The reason I'm concerned is that in my previous dealings with ST, none of the code was open source.  Further, even just trying to download the code from the link above requires acceptance of a non-open-source license.  That doesn't mean that some of this code isn't available under the BSD license, but it does mean that it is hard for us to verify whether that is the case.

mikeferguson: @clalancette I'm also not a lawyer, but I think the header on each file, for instance https://github.com/okalachev/vl53l1x_ros/blob/master/vl53l1x/lib/core/vl53l1_api.c, is pretty clear - they dual license under either their old, terrible proprietary license, or under a 3-clause BSD, at YOUR option. ST has slowly been updating their licensing terms over the past few years, and so I think it's pretty clear the library is BSD.

Claiming the package is then MIT is a bit of a challenge - it'd be much easier if the package itself was 3-clause BSD.

okalachev: @clalancette, header in every source file in this library clearly states, that this library may be licensed with BSD license. And I didn't change them, of course.

sloretz:
holding for the [upcoming Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2021-01-22/18570)

melodic/distribution.yaml
deadln: @clalancette the recent Noetic sync is out but our package does not appear in rosdistro. Should we change the overall licence in our source repo, or not?

deadln:
@sloretz i've seen that Noetic sync is out our package does not appear in rosdistro. Is it because troubles with licence? Or we should do something else?

ivanpauno:
> @sloretz i've seen that Noetic sync is out our package does not appear in rosdistro. Is it because troubles with licence? Or we should do something else?

The package will appear when this PR is merged, and IIUC the only blocker was to adapt the license in the `package.xml`.

ivanpauno:
> The package will appear when this PR is merged

To be more clear, in the next sync after this PR is merged.

okalachev:
@clalancette and @mikeferguson didn't confirmed they're sure I must change the overall license to BSD. BSD and MIT are pretty compatible, so I'm not sure BSD code can't be included to MIT licensed library.

But if you confirm that this is necessary, I'll change it.

okalachev:
Or should list both licenses? Is it even possieble?

okalachev:
Or should list both licenses? Is it possible?

ivanpauno:
> Or should list both licenses? Is it possible?

I'm not sure, listing both sound fine to me but I would wait a confirmation from the others (I'm not sure if that's possible in a package.xml though).

You could also have a top level license file explaining which parts of the library are licensed under which license, I think that would help. The complete license notices can be in the same file after the explanation, or in different files called e.g. `Licence.mit` (and linked from the main one). A lot of repositories use approaches similar to that when they have multilicensed content.

deadln:
@clalancette @mikeferguson @sloretz could somebody answer already, should we change licencing or not? We have a bunch of other packages with MIT licence to add in rosdistro and so we want avoid troubles like this in future.

mikeferguson:
I'm not sure how more concrete of an answer I could provide:

 * The ST-provided code is clearly licensed BSD - so the package.xml can't just say MIT (while they are similar, they are not the same, and by redistributing the ST-provided code under an MIT license which is has different text from the BSD text that ST put at the top of the files you would be in violation of the ST license).
 * I don't know what individual/organization wrote the additional code that has been added - and whether that individual/organization can put their code under a BSD License to match the ST-provided code.

(Personally, I think the easiest path, if possible, is simply to license the whole package under BSD, assuming you have the right to license the wrapper code you've added under BSD). While BSD and MIT are pretty compatible, they aren't exactly the same (and so you can't claim that the ST-provided BSD-licensed code is MIT).

I've also never had a dual licensed ROS-package (I'm not actually sure if the package.xml and the downstream tools actually support that). In cases where I've had multiple licenses in one repo, I've had them in different packages (for instance, https://github.com/mikeferguson/graceful_controller has two packages: graceful_controller which is LGPL and graceful_controller_ros which is BSD).

melodic/distribution.yaml
deadln: @clalancette We've managed to change overall licencing to BSD. Please confirm changes

okalachev:
Hi, I changed the license to BSD-3-Clause.

But, practically all the ROS packages, written in C++, include the code from the `roscpp` package, which is also [licensed under BSD license](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/package.xml#L18). Does this mean it’s impossible to publish a C++ ROS package under a license other than BSD?

What if an MIT licensed library is included to a C++ ROS package, too? Would such a package be impossible to publish at all?

mikeferguson:
> But, practically all the ROS packages, written in C++, include the code from the `roscpp` package, which is also [licensed under BSD license](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/package.xml#L18). Does this mean it’s impossible to publish a C++ ROS package under a license other than BSD?

In this scenario, you are LINKING against roscpp - so the license of your package can be whatever and you are still in compliance with BSD license (that theory works for MIT/LGPL as well, but is of course not true for GPL - if you link against GPL, you are GPL). You also also aren't redistributing the code of roscpp within your package.

okalachev:
OK, can this package be merged now, please?

jacobperron (approved):
LGTM, thanks for iterating and working out the licensing issue!

@clalancette Can you take another look?

clalancette (approved):

PR #28063, opened 2021-01-23
Adding twist_stamper to src/doc index for foxy & rolling

joshnewans:
Hi, I would like to add my package `twist_stamper` to the index for `foxy` and `rolling`.
No binary release yet.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -4313,6 +4313,16 @@ repositories:
       url: https://github.com/autowarefoundation/tvm_vendor.git
       version: main
     status: maintained
+  twist_stamper:
+    doc:
+      type: git
+      url: https://github.com/joshnewans/twist_stamper.git
+      version: main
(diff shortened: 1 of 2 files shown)


ivanpauno (changes requested):
The package xml says that the license is [BSD](https://github.com/joshnewans/twist_stamper/blob/924efdf8be5d4e61445c399e47099d5469c2bcad/package.xml#L9), but the only source file has a [Apache 2.0 license notice](https://github.com/joshnewans/twist_stamper/blob/924efdf8be5d4e61445c399e47099d5469c2bcad/twist_stamper/twist_stamper.py#L1-L13).

It would also be good to add a top level license file to the repo.
Thanks!

joshnewans:
Whoops! Thanks Ivan, I've fixed that up, please let me know if there is any other issue.

ivanpauno (approved):

PR #28065, opened 2021-01-23
add embree to rosdep

hakuturu583:
Signed-off-by: Masaya Kataoka <ms.kataoka@gmail.com>

I want to add embree ray tracing kernels to the rosdep.

```
Arch : https://archlinux.org/packages/community/x86_64/embree/
Debian : https://packages.debian.org/bullseye/libembree-dev
Fedora : https://src.fedoraproject.org/rpms/embree
Ubuntu : https://launchpad.net/ubuntu/+source/embree
```

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -795,6 +795,11 @@ emacs:
   gentoo: [virtual/emacs]
   openembedded: [emacs@meta-oe]
   ubuntu: [emacs]
+embree:
+  arch: [embree]
+  debian: [libembree-dev]
+  fedora: [embree]
+  ubuntu: [libembree-dev]
 enblend:
   debian: [enblend]
   fedora: [enblend]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -799,7 +799,10 @@ embree:
   arch: [embree]
   debian: [libembree-dev]
   fedora: [embree]
-  ubuntu: [libembree-dev]
+  ubuntu:
+    '*': [libembree-dev]
+    bionic: null
+    xenial: null
 enblend:
   debian: [enblend]
   fedora: [enblend]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -797,7 +797,8 @@ emacs:
   ubuntu: [emacs]
 embree:
   arch: [embree]
-  debian: [libembree-dev]
+  debian:
+    bullseye: [libembree-dev]
   fedora: [embree]
   ubuntu:
     '*': [libembree-dev]


rosdep/base.yaml
```diff
@@ -795,6 +795,11 @@ emacs:
   gentoo: [virtual/emacs]
   openembedded: [emacs@meta-oe]
   ubuntu: [emacs]
+embree:
+  arch: [embree]
+  debian: [libembree-dev]
```
ivanpauno, line 800: The debian package seems to only be bullseye (testing) or sid, but the newer debian version supported is `buster` (ros noetic).

IIUC, the debian `entry` should be `null`:

```suggestion
  debian: null
```

rosdep/base.yaml
```diff
@@ -795,6 +795,11 @@ emacs:
...
+  fedora: [embree]
```
ivanpauno, line 801: This is only available from 32, while we also support 28 (melodic), 23 and 24 (kinetic).
I see in ubuntu we sometimes specify in which distro versions a package is available, but IDK if that's necessary in fedora.
Maybe @nuclearsandwich knows (?)

rosdep/base.yaml
```diff
@@ -795,6 +795,11 @@ emacs:
...
+  ubuntu: [libembree-dev]
```
ivanpauno, line 802: the package is not available in xenial and bionic, maybe:

```suggestion
  ubuntu:
    '*': [libembree-dev]
    bionic: null
    xenial: null
```

ivanpauno (changes requested):

rosdep/base.yaml
nuclearsandwich: Rather than an explicit `null` removing the definition is usually preferred and gives the same result. In this case however we could also specify the dependency for the upcoming bullseye distribution.
```suggestion
  debian:
    bullseye: [libembree-dev]
```

Not specifying other distributions will leave them as implicit nulls.

rosdep/base.yaml
nuclearsandwich: Fedora releases themselves are only supported for ~13 months I don't think we need to block on past releases here.

nuclearsandwich (approved):
@hakuturu583  once @ivanpauno's comments are addressed this should be ready to move forward.

rosdep/base.yaml
hakuturu583: Thank you for pointint out this problem, I fixed in https://github.com/ros/rosdistro/pull/28065/commits/29174c447e57ff9e009f0dc77039ba45cc314207

rosdep/base.yaml
hakuturu583: Thank you for pointing out this problem, I fixed in https://github.com/ros/rosdistro/pull/28065/commits/09f9beec1ac20d1c155cff070f549bc34258f2fb

rosdep/base.yaml
hakuturu583: Will you support fedora 32 in the future?
If you will do so, I will stay these lines, if not, I will remove these lines.

ivanpauno: I think keeping it is fine

ivanpauno: > Will you support fedora 32 in the future?

Noetic targets fedora 32 btw

ivanpauno (approved):
Thanks for addressing the comments @hakuturu583 !!

PR #28067, opened 2021-01-23
add python-mechanize-pip

fmessmer:
this adds a rosdep key for the pip version of `mechanize` (https://pypi.org/project/mechanize/)

officital `python-mechanize`/`python3-mechanize` are already available, but `python3-mechanize` is not available for `debian:buster` and thus results in problematic release of e.g. https://github.com/ipa320/cob_command_tools
(read more about it in https://github.com/ipa320/cob_command_tools/issues/289

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2496,6 +2496,19 @@ python-mechanize:
   fedora: [python-mechanize]
   gentoo: [dev-python/mechanize]
   ubuntu: [python-mechanize]
+python-mechanize-pip:
+  debian:
+    pip:
+      packages: [mechanize]
+  fedora:
+    pip:
+      packages: [mechanize]
+  gentoo:
+    pip:
+      packages: [mechanize]
+  ubuntu:
+    pip:
+      packages: [mechanize]
 python-mock:
   alpine: [py-mock]
   arch: [python2-mock]


tfoote (changes requested):
We shouldn't add pip rules that will conflict/override existing debian rules. If any package in the workspace or system depends on the pip rule it will override the version that is available on the system and potentially break anyone using the apt based key. If we want to add this key it should only be valid for platforms where there is no resolution for the conflicting apt/debian based key.

nuclearsandwich:
In addition to @tfoote's points above about not mixing keys. Pip packages can't be used to resolve dependencies for packages released via bloom. Based on the comment here: https://github.com/ipa320/cob_command_tools/issues/289#issuecomment-766696766 it seems like this dependency has been worked around. If that's the case @fmessmer can we close this PR as unnecessary?

Thanks for iterating on the solution with us!

fmessmer:
yes, I guess I won't need this anymore once the following PRs are merged:
- https://github.com/ros/rosdistro/pull/28071
- https://github.com/ros/rosdistro/pull/28072
- https://github.com/ros-infrastructure/ros_buildfarm_config/pull/196

fmessmer:
https://github.com/ros-infrastructure/ros_buildfarm_config/pull/196 is merged and the release does not issue any warnings anymore...so I guess this key is not needed anymore....

PR #28124, opened 2021-01-26
Change Heifu branch for documentation and source

BV-OpenSource:
Is it possible to change this information using the `bloom-release` command?


mabelzhang:
Not sure. Maybe Bloom needs to bump the version?
I think this is okay, but I'll let @jacobperron say for sure.

jacobperron (approved):
Updating doc and source branches manually is okay.

You can also use bloom with the `--edit` option. It will still make new release artifacts, bumping the revision. I'm not sure if there is a way to avoid bumping the version with bloom.

PR #28148, opened 2021-01-27
add key for GStreamer plugins for GL

zflat:
Provides the library `libgstopengl.so` for gstreamer GL plugins

* Arch: https://archlinux.org/packages/extra/x86_64/gst-plugins-base/
* Debian: https://packages.debian.org/sid/gstreamer1.0-gl
* Fedora: https://src.fedoraproject.org/rpms/gstreamer1-plugins-base
* Ubuntu: https://packages.ubuntu.com/bionic/gstreamer1.0-gl

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1319,6 +1319,11 @@ gstreamer1.0-alsa:
   fedora: [gstreamer1-plugins-base]
   rhel: [gstreamer1-plugins-base]
   ubuntu: [gstreamer1.0-alsa]
+gstreamer1.0-gl:
+  arch: [gst-plugins-base]
+  debian: [gstreamer1.0-gl]
+  fedora: [gstreamer1-plugins-base]
+  ubuntu: [gstreamer1.0-gl]
 gstreamer1.0-libav:
   arch: [gst-libav]
   debian: [gstreamer1.0-libav]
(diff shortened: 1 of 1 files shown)


mabelzhang (commented):
The `ubuntu` entry needs `xenial: null`.

The package used to exist on Gentoo but has been moved to the `bad` category, meaning something is missing. Apparently gstreamer names their plugins in the categories of the good, the bad, and the ugly? So I guess it isn't required to include it.
https://packages.gentoo.org/packages/media-libs/gst-plugins-bad

mabelzhang (commented):
Ehh `xenial` shouldn't have quotes around it. It didn't break the checks, but we should follow the convention.

zflat:
@mabelzhang thanks for catching the inconsistency with the quotes. fixed in 30527cb

mabelzhang (approved):

rosdep/base.yaml
```diff
@@ -1319,6 +1319,13 @@ gstreamer1.0-alsa:
   fedora: [gstreamer1-plugins-base]
   rhel: [gstreamer1-plugins-base]
   ubuntu: [gstreamer1.0-alsa]
+gstreamer1.0-gl:
+  arch: [gst-plugins-base]
+  debian: [gstreamer1.0-gl]
```
jacobperron, line 1324: I think we should add an entry for `stretch`, which is still a supported LTS but does not have `gstreamer1.0-gl`:
```suggestion
  debian:
    '*': [gstreamer1.0-gl]
    stretch: null
```

jacobperron (changes requested):

jacobperron (approved):

jacobperron:
@zflat Thanks for iterating!

PR #28160, opened 2021-01-28
Add python-nose rosdep entry for ubuntu focal

stwirth:
The package has the same name in focal: https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=python-nose


tfoote (changes requested):
Since we're looking at this for future compatibility lets just use a wildcard rule, `'*'` for `python-nose` and then we can remove all the other duplicate rules and just leave the legacy python3 rules that will get cleaned up later. This will also make it cover other unenumberated releases.

stwirth:
@tfoote done.

tfoote (approved):
Thanks, much cleaner!

PR #28164, opened 2021-01-29
Enable PR testing for laser_geometry

mabelzhang:
Not using Bloom because build is failing on Kdev, hence enabling PR testing to check Kpr builds before releasing.

I forget whether I can approve my own PR while I'm on rosdistro duty, probably not :)

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -6573,6 +6573,7 @@ repositories:
       url: https://github.com/ros-gbp/laser_geometry-release.git
       version: 1.6.5-1
     source:
+      test_pull_requests: true
       type: git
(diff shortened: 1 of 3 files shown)


mabelzhang:
Let me know if I need to split into 3 PRs. I got lazy. No idea what turning on PR checks involve.

tfoote (approved):

tfoote:
I guess you could use bloom to do this but manual is easier for just adding this metadata. 

This is just going to cause new jobs to be generated on the buildfarm during the overnight reconfigure process. And that will also register webhooks with github to trigger the builds. (This organization should already have permissions setup so you don't need to check on that.)

More docs here: http://wiki.ros.org/buildfarm/Pull%20request%20testing

PR #28181, opened 2021-01-29
ros_industrial_cmake_boilerplate: 0.2.7-1 in 'noetic/distribution.yaml' [bloom]

Levi-Armstrong:
Increasing version of package(s) in repository `ros_industrial_cmake_boilerplate` to `0.2.7-1`:

- upstream repository: https://github.com/ros-industrial/ros_industrial_cmake_boilerplate.git
- release repository: https://github.com/ros-industrial-release/ros_industrial_cmake_boilerplate-release.git
- bloom version: `0.10.0`

## ros_industrial_cmake_boilerplate

```
* Add contributing file
* Add license files and update documentation
* Contributors: Levi Armstrong
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5476,6 +5476,17 @@ repositories:
       url: https://github.com/ignitionrobotics/ros_ign.git
       version: noetic
     status: developed
+  ros_industrial_cmake_boilerplate:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/ros-industrial-release/ros_industrial_cmake_boilerplate-release.git
+      version: 0.2.7-1
+    source:
+      type: git
+      url: https://github.com/ros-industrial/ros_industrial_cmake_boilerplate.git
+      version: master
+    status: developed
   ros_numpy:
     doc:
       type: git


mabelzhang (commented):
Thanks for adding the LICENSE files! What about the `gtest` package?
"Either each package in the repository should have a LICENSE file, or the source repository should have a top-level LICENSE file."
You can either put the LICENSE files at the root of the repo, or make another for the `gtest` package.

Levi-Armstrong:
> Thanks for adding the LICENSE files! What about the `gtest` package?
> "Either each package in the repository should have a LICENSE file, or the source repository should have a top-level LICENSE file."
> You can either put the LICENSE files at the root of the repo, or make another for the `gtest` package.

I am not sure I would consider this a normal package. It is only for development where it uses cmake's `ExternalProject_Add` utility to clone down the gtest source and build it in your workspace. It is an alternative approach to suing a rosinstall but with this approach you have the ability to set compiler flags, etc.. The gtest component of the repository is not intended to be released and has been added to the noetic.ignore.

mabelzhang:
Oh yeah, I see the ignored package in the README now, thanks for explaining!
I also see this peculiar branch `release/noetic/cmake_common_scripts` and other branches for this `cmake_common_scripts` package that isn't in the repo. Do you know if that's supposed to be there?

Levi-Armstrong:
> Oh yeah, I see the ignored package in the README now, thanks for explaining!
> I also see this peculiar branch `release/noetic/cmake_common_scripts` and other branches for this `cmake_common_scripts` package that isn't in the repo. Do you know if that's supposed to be there?

Sorry, I thought I had removed all of those but I missed a few. The package was renamed per a request of this repo.

mabelzhang:
Thanks!

For future reference, this is a rename of #27906.
Can I get a second opinion from the maintainers on the new name based on the discussion on that previous PR?

`ros_industrial` here is interpreted as an organization name and suggested in the previous PR.
Can it be an exception to [REP 144](https://ros.org/reps/rep-0144.html) Global Rule number 3 "A package name should not contain `ros` as it is redundant. Exceptions include core packages and ROS bindings of an upstream library (e.g. moveit_ros)"?

I don't mean to be a troll. Just don't want to make this decision alone.

Levi-Armstrong:
> I don't mean to be a troll. Just don't want to make this decision alone.

Understand. Thank you.

jacobperron (approved):

wjwwood:
> Can it be an exception to REP 144 Global Rule number 3 "A package name should not contain ros as it is redundant. Exceptions include core packages and ROS bindings of an upstream library (e.g. moveit_ros)"?

Yes, I think so.

wjwwood (approved):

PR #28264, opened 2021-02-08
Python3 version of Keras and Tensorflow pip targets

MatthijsBurgh:


tfoote (changes requested):
We already have `python-keras-pip` and `python-tensorflow-pip` that are functionally equivalent. I'm a bit mixed as to whether to duplicate them to the new standard or to keep the reduced ruleset with the old name. We don't want to drop the old keys that people may already be using successfully. Thus I think I'd rather err on the side of sticking with the existing rules. If we duplicate them now at some future point we'll then have two sets of rules to retire/migrate away from. And as these aren't expected to be packaged into debian packages in the near future they shouldn't conflict.

MatthijsBurgh:
@tfoote i agree with you, as these are not needed and the python version of these entries would also work, but the current documentation is pretty clear that these new keys should be added, see https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1.

Also there are already a few `python3-*-pip` entries.

So either we should merge this or change the documentation and optionally remove these `python3-*-pip` entries.

hidmic (commented):
This is indeed consistent with the guidelines. @tfoote thoughts?

tfoote:
The rule was written to guide the creation of new rules. If we have a rule that exists it's legacy and it's not saying that we should duplicate it and move it. It's there as a legacy, and we don't want to just create churn without having a significant benefit. That's why it says they're legacy. If we wanted to homogenize everything we could write a script and duplicate every `python-*-pip` rule to be `python3-*-pip` but we're not doing that despite the fact that it would be "cleaner". This would be one of the iterations of that. Doing that migration would cause a lot of work for the greater community. When we create new rules we follow the new guildelines and let the legacy ones continue to exist until they are no longer relevant or need to change. A clarification of the legacy status could be provided, it's been somewhat implicit in how we process things where we apply the new guidelines to new rules.

MatthijsBurgh:
@tfoote I missed the conclusion. (Could be just me). Do we keep the current guidelines and go for adding the `python3-*-pip` keys? Or do we change the the guidelines to only have `python-*-pip`?

tfoote:
We keep the current guides. And can add more specifically the clarification that we won't update old entries or create duplicates to match the new rules unless they're going to be changed for another reason.

PR #28280, opened 2021-02-10
Added python packages gputil and jetson-stats

maarufvazifdar:
There is no pip package for gputil and jetson-stats.
1) gputil: To get the gpu stats from our system. [https://pypi.org/project/GPUtil/]
2) jetson-stats: To get the stats from jetson develpoer boards. [https://pypi.org/project/jetson-stats/]

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1919,6 +1919,10 @@ python-gpiozero:
       pip:
         packages: [gpiozero]
     zesty: [python-gpiozero]
+python-gputil-pip:
+  ubuntu:
+    pip:
+      packages: [gputil]
 python-gpxpy:
   debian:
     buster: [python-gpxpy]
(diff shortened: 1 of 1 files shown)


cottsay (approved):

hidmic (commented):
@maarufvazifdar would you mind completing the rules for debian and fedora? macos would be ideal too.

On a slightly related note, `gputil` appears somewhat stale. Is it the only package of its kind?

maarufvazifdar:
@hidmic Hi, I have added the rules for debian, fedora and osx for gputils package. As per our usage gputils works just fine and we haven't explored any other package of its kind.
Regarding jetson-stats, Nvidia Jetson devices only run official Jetpack that comes with various Ubuntu versions. So we have only tested this on Ubuntu 16.04 and 18.04.

hidmic (approved):
Got it. Thanks for your contribution @maarufvazifdar!

PR #28300, opened 2021-02-11
disable ros2_control for rolling

Karsten1987:
The rolling release was made my mistake. The current master branches are not compliant with the rest of the ROS2 rolling stack and for API/ABI conserving reasons, we'd not like branch at this point.

cc/ @bmagyar I couldn't find an entry for ros2_controllers. I assume you didn't release that repo for rolling then?

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1835,30 +1835,6 @@ repositories:
       url: https://github.com/ros2/ros1_bridge.git
       version: master
     status: maintained
-  ros2_control:
-    doc:
-      type: git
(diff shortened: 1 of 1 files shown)


cottsay (approved):
I was seeing these packages fail on the RPM farm pretty catastrophically. I'm glad to hear that I won't need to investigate that.

IIUC, the buildfarm will remove the jobs automatically.

nuclearsandwich (approved):

nuclearsandwich:
I'm sure that we'll post on discourse with some advance warning but if you do want to get into Galactic it would be good to get a rolling-compatible release up soon.

PR #28309, opened 2021-02-12
add flask socketio

tiffca:
Hi, 

For a project I'm working on, we would like to do bidirectional streaming of information over the web for which we need WebSockets. 

We tried using python web sockets, but it was too low level for our purposes. Also we are using Flask, so we want to use the Flask framework. 

https://pypi.org/project/Flask-SocketIO/

https://flask-socketio.readthedocs.io/en/latest/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1642,6 +1642,19 @@ python-flask-restful:
         packages: [flask-restful]
     xenial: [python-flask-restful]
     yakkety: [python-flask-restful]
+python-flask-socketio:
+   debian:
+     pip:
+       packages: [flask-socketio]
+   fedora:
+     pip:
+       packages: [flask-socketio]
+   osx:
+     pip:
+       packages: [flask-socketio]
+   ubuntu:
+     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1642,6 +1642,19 @@ python-flask-restful:
         packages: [flask-restful]
     xenial: [python-flask-restful]
     yakkety: [python-flask-restful]
+python-flask-socketio:
```
hidmic, line 1645: @tiffca Flask-SocketIO docs suggest it is a Python 3 package only (which means the key should be `python3-flask-socketio` instead) and I can see it listed in platform-specific package managers (which means that the key should be `python3-flask-socketio-pip` instead, though it'd best to not use those instead of pip if possible).

hidmic (commented):
@tiffca check failing CI jobs

tiffca:
Thanks @hidmic for your help on this PR! We are closing it since we were able to find an alternative way to download the library we needed (via Dockerfile).

PR #28330, opened 2021-02-15
ros_led: 0.0.9-1 in 'melodic/distribution.yaml' [bloom]

deadln:
Increasing version of package(s) in repository `ros_led` to `0.0.9-1`:

- upstream repository: https://github.com/CopterExpress/ros_led.git
- release repository: https://github.com/CopterExpress/ros_led-release.git
- bloom version: `0.10.0`

## led_msgs

```
* Version bump
```

## ws281x

```
* Restore lost rpi_ws281x library files
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -10250,6 +10250,24 @@ repositories:
       type: git
       url: https://github.com/shadow-robot/ros_ethercat_eml.git
       version: melodic-devel
+  ros_led:
+    doc:
+      type: git
+      url: https://github.com/CopterExpress/ros_led.git
+      version: master
+    release:
+      packages:
+      - led_msgs
+      - ws281x
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/CopterExpress/ros_led-release.git
+      version: 0.0.9-1
+    source:
+      type: git
+      url: https://github.com/CopterExpress/ros_led.git
+      version: master
+    status: maintained
   ros_monitoring_msgs:
     doc:
       type: git


ahcorde:
I think the name is too generic, this package control a `ws281x LED` for a Raspberry Pi. What do you think @clalancette and/or @tfoote ?

Another note, I think that this package is the first time that it's released, you should change the version to 0.0.1.

clalancette:
> I think the name is too generic, this package control a `ws281x LED` for a Raspberry Pi. What do you think @clalancette and/or @tfoote ?

Agreed, I think this is a little too generic.  I'd go with `ws281x_led` or something like that.

> Another note, I think that this package is the first time that it's released, you should change the version to 0.0.1.

We don't really have any rules around this; some things that come in are more mature than others.  In this case, I think the 0.0.9 that is already here is fine.

okalachev:
@ahcorde, @clalancette, the package can be extended later to support platforms other then RPi. Also, in practice, RPi+ ws281x  is the most common combination. `_led` suffix is also not nessesary, because ws281x *is* a LED strip.

Another thing is this package is (as well as [`vl53l1x`](https://github.com/okalachev/vl53l1x_ros), and others) is already widely used for a long time as a part of the [Clover Drone Kit software bundle](https://github.com/CopterExpress/clover). So renaming the packages would be very inconvenient.

deadln:
> > I think the name is too generic, this package control a `ws281x LED` for a Raspberry Pi. What do you think @clalancette and/or @tfoote ?
> 
> Agreed, I think this is a little too generic.  I'd go with `ws281x_led` or something like that.
> 
> > Another note, I think that this package is the first time that it's released, you should change the version to 0.0.1.
> 
> We don't really have any rules around this; some things that come in are more mature than others.  In this case, I think the 0.0.9 that is already here is fine.

Dear @clalancette I think "generic name" isn't an issue at all because this name presents the relation of this package pretty well. There is no other thing in the world that have a name "ws281x" besides this LED strip

sloretz:
> Agreed, I think this is a little too generic. I'd go with `ws281x_led` or something like that.

@clalancette which package name is too generic? `led_msgs` or `ws281x`?


> Another note, I think that this package is the first time that it's released, you should change the version to 0.0.1.

I disagree. I think that would add confusion to people who have used older versions of this software since [other tags already exist](https://github.com/CopterExpress/ros_led/tags).

sloretz (approved):

ivanpauno:
You have some bsd licensed files in `ws281_x`, https://github.com/CopterExpress/ros_led/blob/master/ws281x/vendor/rpi_ws281x/dma.c.

I think that if you combine bsd and mit the `package.xml` must say `mit`, but I don't recall exactly (and in that case, you can add a header with the license to the files that you want to be mit licensed).
Maybe @clalancette can clarify.

About naming, `ws281x` sounds ok to me.

okalachev:
Hi, @ivanpauno, according to [REP-140](https://www.ros.org/reps/rep-0140.html#license-multiple-but-at-least-one):

> A package will have multiple licenses if different source files have different licenses. Every license occurring in the source files should have a corresponding <license> tag.

I [added](https://github.com/CopterExpress/ros_led/commit/d15f1ac818461074380c3359f090e6fda88b61d1) BSD license to package.xml as well as MIT, so now everything should be fine.

ivanpauno:
> I [added](https://github.com/CopterExpress/ros_led/commit/d15f1ac818461074380c3359f090e6fda88b61d1) BSD license to package.xml as well as MIT, so now everything should be fine.

Thanks!

Though it's clear than the files in `ws281x/vendor/` are licensed under the BSD license, it's not super clear that all the others are licensed under MIT.
I guess that adding a note in the top level `LICENSE` file explaining that everything except `ws281x/vendor/` is licensed under MIT and also pointing to the `ws281x/vendor/` license file will make it clear, @okalachev does that sound good to you?

okalachev:
@ivandariojr, I added a header to the main source file and a note to the top LICENSE file for clarifying.

ivanpauno (approved):

ivanpauno:
That looks better to me, thanks @okalachev !

PR #28340, opened 2021-02-18
Added focal key for python-serial

lauralindzey:
I need to have a ROS package depend on python-serial and also be compatible with both melodic and noetic. 

https://packages.ubuntu.com/focal/python3-serial

(Tested locally by changing /opt/ros/rosdep/sources.list.d)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4285,6 +4285,7 @@ python-serial:
     artful_python3: [python3-serial]
     bionic: [python-serial]
     bionic_python3: [python3-serial]
+    focal: [python3-serial]
     lucid: [python-serial]
     maverick: [python-serial]
     natty: [python-serial]


matthew-reynolds:
(Not a reviewer, just curious. This question is mostly aimed at the maintainers of the package, not at @lauralindzey)

What is the preference to handling these py2-vs-py3 packages? Currently, the following would work:
```
  <depend condition="$ROS_PYTHON_VERSION == 2">python-serial</depend>
  <depend condition="$ROS_PYTHON_VERSION == 3">python3-serial</depend>
```
But the existing `<version>_python3` tags within the `python-serial` rule muddies the water for me a little. It seems like a slightly smarter `<depend>python-serial</depend>`, but only when using Ubuntu, as other distros still always get the py2 version. When should one approach be favoured over the other?

tfoote (changes requested):
The postfix `_python3` approach was determined to not provide a good user experience and should not be used going forward. We're leaving existing rules for backwards compatibility until they are on EOL platforms. But unless you explicitly override your rospkg OS detection they will never be used. 


Our new approach is to use separate `python3-*` keys

So instead of adding the python3 rule to the python2 key. The approach mentioned by @matthew-reynolds  is the recommended approach.

Since I think the rules currently provide coverage with that approach I'm going to close this.

lauralindzey:
OK, thanks for the explanation!

PR #28351, opened 2021-02-18
Adding ifxsensor to documentation index for noetic

Yang-IFX:
I'd like ifxsensor to be indexed and documented on ros.org.


Yang-IFX:
I'd like ifxsensor to be indexed and documented on ros.org.

noetic/distribution.yaml
```diff
@@ -2483,6 +2483,24 @@ repositories:
       url: https://github.com/ethz-adrl/ifopt.git
       version: master
     status: maintained
+  ifxsensor:
+    doc:
+      type: git
+      url: https://github.com/Yang-IFX/ifxsensor.git
```
clalancette, line 2489: The license listed in the package.xml seems to be wrong.  The LICENSE file at the top-level says MIT, and the headers on the files say the same, so that should be changed to MIT.

clalancette (changes requested):

noetic/distribution.yaml
Yang-IFX: Thanks very much. Already changed.

clalancette: Great, thanks.

I'm going to close out this PR, and I'll suggest you do a new patch or minor release of the package.  Otherwise, the license that people will see on the https://wiki.ros.org page will be wrong.

PR #28385, opened 2021-02-22
Opensuse leap

neotinker:
I forked rosdistro for some development work I was doing for openSUSE. Since this represents a lot of little changes @tfoote recommended opening a draft PR to give people some time to look over it.

These changes were used by my rossuse.py script to automate the generations of rpm packages for opensuse Leap 15.1 on the opensuse build service platform.

https://github.com/neotinker/rossuse
https://build.opensuse.org/project/show/home:neotinker3:ROS:melodic_test


tfoote (commented):
Overall this is great to get coverage. That's a lot of rules to fill in. As a side comment  I'm surprised how different the rules are for opensuse vs fedora. I'm also surprised by the number of versions that have to be pinned to specific packages in the packages. 

I noticed your focus on 15.1 which from my quick searching appears to be EOL as of last month.15.2 might seem like a better target, though the 18 month cycle is very short to target.

As we'd like this effort to be sustainable it would be great to extend the CONTRIBUTING.md and REVIEW_GUIDELINES to explain how to submit rules for SUSE as well as how to review them. 
It will still be best effort, but a pointer where to find an index of the packages for comparison is important for reviewing as well as non-suse users to find equivalent packages quickly since the easier it is the more likely people will voluntarily take the time to fill it in.

Once we've reviewed that process updates then the individual rules can be processed in small batches with links to the packages for reviewing.

neotinker:
> Overall this is great to get coverage. That's a lot of rules to fill in. As a side comment I'm surprised how different the rules are for opensuse vs fedora. I'm also surprised by the number of versions that have to be pinned to specific packages in the packages.
> 
> I noticed your focus on 15.1 which from my quick searching appears to be EOL as of last month.15.2 might seem like a better target, though the 18 month cycle is very short to target.

I started this work around the time that openSUSE Leap 15.1 was released but if you look at OBS, you will see that I am using the same spec file generated for 15.1 to also build 15.2 packages.

Since there were already some opensuse definitions, I tried not to break what someone else might have been doing. That is why you will notice some definitions like this.

  opensuse:
    '*': [libassimp3]
    '15.1': [assimp-devel]

Since I didnt know what release they were targeting, I tried to be explicit about what I was targeting. If there is no risk of breaking the work of others, I can change that to be more generic. I did try to be more generic, such as the libboost definitions.

  opensuse: [libboost_atomic1_66_0]

But libraries are one place where we will most likely require OS release specific entries since the packaging guidelines for openSUSE/SLES specify putting version information as part of the name to allow having multiple versions installed in parallel.
https://en.opensuse.org/openSUSE:Shared_library_packaging_policy#Package_naming

With the upcoming release of openSUSE 15.3, the boost libraries will all be jumped to version 75. meaning it will need to be changed to something like

  opensuse:
    '15.1': [libboost_atomic1_66_0]
    '15.2': [libboost_atomic1_66_0]
    '*': [libboost_atomic1_75_0]

Assuming that we should be using the wildcard to refer to all future versions.

Any entries I created that specify a specific os release but don't have version information in the package name can most likely be changed to a more generic entry. For instance,

  opensuse:
    '15.1': [python3-CairoSVG]

Could most likely be changed to 

  opensuse: [python3-CairoSVG]

Even though it does technically have version info in the name, it isn't likely to change much.

> 
> As we'd like this effort to be sustainable it would be great to extend the CONTRIBUTING.md and REVIEW_GUIDELINES to explain how to submit rules for SUSE as well as how to review them.

I can look into updating these.

> It will still be best effort, but a pointer where to find an index of the packages for comparison is important for reviewing as well as non-suse users to find equivalent packages quickly since the easier it is the more likely people will voluntarily take the time to fill it in.
> 
> Once we've reviewed that process updates then the individual rules can be processed in small batches with links to the packages for reviewing.

tfoote:
Yes the wildcards are usually used for future compatibility. The boost example is a good best practice.

With respect to the older targets and persistence. The most recent commits in [a quick search](https://github.com/ros/rosdistro/search?o=desc&q=opensuse&s=author-date&type=commits) are from 2017 and before that most were from 2014 which suggests that they're long EOL.  If you can guess what target they are for, our usual approach is to use explicit rules for older versions and keep the wildcard for the expected newest/future version. With the existing rules age I think that modernizing them w/o backwards compatibility makes more sense.

If you have the believe that something is unlikely to change sticking with the generic tag is recommended. Otherwise the rules will have to be updated for each upcoming release.

clalancette:
@tfoote makes some good points; I'll suggest updating your changes to take them into account.

In terms of actually getting this committed, it would be much easier to do this in smaller chunks.  My suggestion is that you take small groups of rosdep rules (maybe no more than 20 at a time), and open a PR for that.  Also make sure to add links to the Opensuse packages so it is easy for reviewers to look at them.

That way we can slowly get the full support into the rosdep database.

nuclearsandwich:
> I started this work around the time that openSUSE Leap 15.1 was released but if you look at OBS, you will see that I am using the same spec file generated for 15.1 to also build 15.2 packages.

Are 15.x releases compatible? Can we target openSUSE Leap 15 rather than 15.x?

neotinker:
> > I started this work around the time that openSUSE Leap 15.1 was released but if you look at OBS, you will see that I am using the same spec file generated for 15.1 to also build 15.2 packages.
> 
> Are 15.x releases compatible? Can we target openSUSE Leap 15 rather than 15.x?

If you mean can you compile it once for '15' and have it run on any '15.x', the short answer is no. They try to maintain compatibility between minor releases but there is no guarantee. The situation should get better in the future with SUSE's push to make openSUSE follow SLES more closely. 
https://en.opensuse.org/Portal:Leap/FAQ/ClosingTheLeapGap

The goal is SLES + SUSE Packagehub = openSUSE

I just got lucky between '15.1' and '15.2'. The spec files I generated for opensuse 15.1 work to compile and package rpms for opensuse 15.2, so far. And as I mentioned earlier, there will be some incompatible file names for '15.3'.

I intend to update everything to 15.2 and 15.3 once I pushed my current changes. Updating for '15.2' and '15.3' should require minimal additional changes.

tfoote:
> In terms of actually getting this committed, it would be much easier to do this in smaller chunks. My suggestion is that you take small groups of rosdep rules (maybe no more than 20 at a time), and open a PR for that. Also make sure to add links to the Opensuse packages so it is easy for reviewers to look at them.

For others reference this is a draft with the intention to split it up following my suggestion at https://discourse.ros.org/t/general-rosdistro-pr-question/19101

I found the chart at the bottom of https://en.wikipedia.org/wiki/OpenSUSE_version_history helpful for understanding the releases. Each 15.x is equivalent to a full ubuntu distro basically.

clalancette:
I'm going to go ahead and close this one, as the discussion has seemed to have stopped and in its current form it can't be merged.  Feel free to reopen, keep commenting, or open smaller PRs to start feeding this into rosdistro.  Thanks for the contribution.

PR #28411, opened 2021-02-25
Disable Python 2.7 tests.

clalancette:
There is no need to run them anymore; Python 2.7 is out of
support for over a year now.  While we are in here, also
upgrade our runners to use Ubuntu 20.04

### DIFF
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -4,10 +4,10 @@ on: [push, pull_request]
 jobs:
   nosetests:
     name: Nosetests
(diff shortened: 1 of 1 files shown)


audrow (approved):
Not sure if you were looking for a review - but looks good to me.

tfoote (approved):

clalancette:
> Not sure if you were looking for a review - but looks good to me.

Yep, always!  Thanks for the review.  I'm going to go ahead and merge despite the fact that the 2.7 Nosetests didn't run.  That is expected, and shouldn't happen in future PRs.

PR #28419, opened 2021-02-26
Add python-opencv for Ubuntu focal (20.04)

marcoesposito1988:
As of 20.04, python-opencv is no longer present in the APT repositories, and was superseded by python3-opencv. 

This is also coherent with the switch from python2 to python3 that happened between ROS Melodic and Noetic.

Solves https://github.com/IFL-CAMP/easy_handeye/issues/74

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2873,6 +2873,9 @@ python-opencv:
     '7': [opencv-python]
   slackware: [opencv]
   ubuntu: [python-opencv]
+python3-opencv:
+  ubuntu:
+    focal: [python3-opencv]
 python-opengl:
   arch: [python2-opengl]
   debian: [python-opengl]


tfoote (changes requested):
python2 and python3 packages are not interchangeable as they do not provide the same content please create a new key `python3-opencv` for focal and newer so maintainers can be confident about which version they are going to get.

Also for review of rosdep submissions please link to the package listings online for others faster reference.

And you have a trailing whitespace which is why you're failing CI.

marcoesposito1988:
Alright, thanks!

marcoesposito1988:
Apparently it was already present. Sorry for the inconvenience.

tfoote:
No problem, glad it's already working.

If you need to span a package across distros you can use the the conditional dependency in package.xml https://www.ros.org/reps/rep-0151.html


```
  <depend condition="$ROS_PYTHON_VERSION == 2">python-opencv</depend>
  <depend condition="$ROS_PYTHON_VERSION == 3">python3-opencv</depend>
```

PR #28420, opened 2021-02-26
Adding ifxsensor to documentation index for noetic

Yang-IFX:
​I'd like ifxsensor to be indexed and documented on ros.org.


noetic/distribution.yaml
```diff
@@ -2483,6 +2483,24 @@ repositories:
       url: https://github.com/ethz-adrl/ifopt.git
       version: master
     status: maintained
+  ifxsensor:
+    doc:
+      type: git
+      url: https://github.com/Yang-IFX/ifxsensor.git
```
clalancette, line 2489: I noticed one other weird thing when looking through the repository.  https://github.com/Yang-IFX/ifxsensor/tree/master/ifx_bgt60_device/lib/sensor_sdk/libs contains a bunch of pre-built libraries.  The problem with it is that we have no idea whether we have the rights to redistribute those libraries.  Can you talk more about what those libraries are, why they are needed, and where they come from?

Yang-IFX: Hi clalancette, 
first of all, thank you very much for checking all these details. 
For the moment, I can answer your question about:
1. There libraries are build from same source code for different platforms;
2. They are used for config/control the radar chip.
3. The library comes from Infineon. No 3rd party code inside.

So back to your first question: whether you have the rights to redistribute. 
From Infineon side, we are willing to share these binary files to public. That's also the reason we put it on GitHub as public repo for everybody to download. But from legal point of view, if you need some official documents to guarantee the right authorization, please contact me by: Yang.Wu@infineon.com . I am very willing to be supportive here. 
Thanks again.
Best Regards,
Yang

Yang-IFX: One correction:
          **_3. The library comes from Infineon. No 3rd party code inside._**
          Actually there is fftw library we used inside. But Infineon has got the authorization to redistribute it.

ivanpauno: @clalancette friendly ping, do you think that's ok?

Yang-IFX: Hi @clalancette @ivanpauno , it's a long time since last comment. I just came back from a long vacation. How can we proceed here about this Pull-Request? Thanks very much.

Yang-IFX: @sloretz , could you please take a look at this PR? Thanks very much.

hidmic: > From Infineon side, we are willing to share these binary files to public. That's also the reason we put it on GitHub as public repo for everybody to download. But from legal point of view, if you need some official documents to guarantee the right authorization, please contact me by: Yang.Wu@infineon.com . I am very willing to be supportive here.

It'd be ideal for such legal documents to be distributed along with the binaries. But I'll defer to @clalancette.

noetic/distribution.yaml
```diff
@@ -2483,6 +2483,24 @@ repositories:
...
+      version: master
+    release:
+      packages:
+      - ifx_bgt60_device
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/Yang-IFX/ifxsensor-release.git
+      version: 0.0.1-2
+    source:
+      test_pull_requests: true
```
hidmic, line 2499: @Yang-IFX have you followed [all the steps](http://wiki.ros.org/buildfarm/Pull%20request%20testing) to enable pull request testing?

hidmic (commented):
@Yang-IFX why wasn't `bloom` used to release this package?

noetic/distribution.yaml
Yang-IFX: Hi @hidmic, thanks very much for your comments. I just double checked my repo setting. Do miss the "Collaborators" of "ros-pull-request-builder". Long vacation lost things. Thank you.

Yang-IFX: Hi @hidmic , thanks for your last comment. I added "ros-pull-request-builder" from my repo: https://github.com/Yang-IFX/ifxsensor/settings/access . It's been several days. Still on the "pending invite" status. Do you have any idea about this? Thank you.

noetic/distribution.yaml
clalancette: The problem here is that anything that is not a strict open-source license as defined by OSI (https://opensource.org/licenses) is basically impossible for us.  We don't have any ways to track the provenance of the code, and whether we are allowed to distribute it, and we really don't want to start doing that.  Also, anyone who is consuming our packages has to do the same due diligence, so putting non-open-source code in here makes their life more difficult as well.

My inclination is to close this out, since we just can't be sure of the legality of distributing the code.  But I'll leave one final ping for @tfoote to look at this.

tfoote: I believe that the MIT license can cover software in both source and binary format which is fine for the libraries owned by Infineon. However I know that the license for FFTW is GPL: http://fftw.org/doc/License-and-Copyright.html though I see that they offer commercial licenses. Documentation of that commercial license should be included in a NOTICES file and also noted in the package.xml. Then if all the other content is Infineon and MIT it should be fine.

noetic/distribution.yaml
tfoote: @nuclearsandwich @clalancette are one of you getting these notifications? Dirk used to get the notifications and accept them.

nuclearsandwich: I don't get notifications by email. I've spiked out a script that signs in and automatically accepts these periodically but have never put it into production.

noetic/distribution.yaml
Yang-IFX: Add a NOTICE file alongside with the binary files to clarify. Thanks.

noetic/distribution.yaml
clalancette: I just got a notification.  I've now logged in and accepted this one.

Yang-IFX: Thank you very much.

nuclearsandwich:
For the other reviewers on this thread:

Now that a [NOTICE file](https://github.com/Yang-IFX/ifxsensor/blob/master/ifx_bgt60_device/lib/sensor_sdk/libs/NOTICE) is present in the repository do we expect that the package.xml should include an additional [license tag](https://ros.org/reps/rep-0140.html#license-multiple-but-at-least-one) to describe the additional license those files are distributed under?

tfoote:
What we and the community need is the right to redistribute the binaries. In the NOTICES file there should be the actual text of the license. I can't answer the question of if I have the right to redistribute the content w/o the license. And similarly others will not have the knowlege of that either. This is why it should also be noted in the package.xml. So that if you're auditing packages on your system you know that there's extra content in this package that needs attention that's not purely open source.

nuclearsandwich:
> In the NOTICES file there should be the actual text of the license.


>  This is why it should also be noted in the package.xml


@Yang-IFX please make changes to address the above points. Thanks.

Yang-IFX:
> > In the NOTICES file there should be the actual text of the license.
> 
> > This is why it should also be noted in the package.xml
> 
> @Yang-IFX please make changes to address the above points. Thanks.

Yang-IFX:
> > In the NOTICES file there should be the actual text of the license.
> 
> > This is why it should also be noted in the package.xml
> 
> @Yang-IFX please make changes to address the above points. Thanks.

Sure. Thanks for reviewing.

ivanpauno:
> In the NOTICES file there should be the actual text of the license.

> This is why it should also be noted in the package.xml

@Yang-IFX friendly ping, these comments don't seem to have been addressed yet.
If you're not planning to address them soon-ish, consider closing the PR and reopening it when ready.

PR #28425, opened 2021-02-26
Replace ament_acado with acado_vendor

JWhitleyWork:
`ament_acado` does not match the naming conventions of other vendor wrapper packages. Additionally, `ament_` makes it sound like this package is part of `ament` rather than being `ament`-ified. The source and release repositories have already been renamed and these changes reflect those naming changes.

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -6,6 +6,17 @@ release_platforms:
   ubuntu:
   - bionic
 repositories:
+  acado_vendor:
+    release:
+      tags:
+        release: release/dashing/{package}/{version}
+      url: https://gitlab.com/autowarefoundation/autoware.auto/acado_vendor-release.git
+      version: 1.0.0-1
+    source:
+      type: git
+      url: https://gitlab.com/autowarefoundation/autoware.auto/acado_vendor.git
+      version: main
+    status: maintained
   ackermann_msgs:
     doc:
       type: git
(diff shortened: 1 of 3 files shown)


audrow (approved):

clalancette:
Holding for the Foxy [sync freeze](https://discourse.ros.org/t/preparing-for-foxy-sync-2021-03-01/19178/2)

shr-project:
@clalancette today's foxy sync is broken without this change, because it looks like the repos were already renamed and
https://gitlab.com/autowarefoundation/autoware.auto/ament_acado.git don't exist anymore (returns 500).

rosdistro_build_cache is now asking for gitlab username/password and fails:
```
DEBUG:rosdistro:Load package.xml file from url "https://raw.githubusercontent.com/ros2-gbp/common_interfaces-release/release/foxy/actionlib_msgs/2.0.3-1/package.xml"
DEBUG:rosdistro:Skipped "github_manifest_provider()": VCS url "https://gitlab.com/autowarefoundation/autoware.auto/ament_acado-release.git" does not match expected format.
DEBUG:rosdistro:Skipped "bitbucket_manifest_provider()": VCS url "https://gitlab.com/autowarefoundation/autoware.auto/ament_acado-release.git" does not match expected format.
Username for 'https://gitlab.com':
```

While https://gitlab.com/autowarefoundation/autoware.auto/acado_vendor and https://gitlab.com/autowarefoundation/autoware.auto/acado_vendor-release do work.

JWhitleyWork:
@shr-project Very sorry about that. Didn't realize it would break the sync. There should be an automatic redirect for the `ament_acado` repo but changing all of the contents of `ament_acado-release` would have been too much of a pain-in-the-butt so it was re-created, which does not have a redirect.

PR #28439, opened 2021-03-01
Updated python-can rosdep rules for Ubuntu

braamBeresford:
The Ubuntu package is out of date and no longer the recommended way to install.

PyPi: https://pypi.org/project/python-can/

This is my first open source contribution, I apologize if this doesn't meet all the rules (I did run unit tests).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -818,49 +818,8 @@ python-can:
     pip:
       packages: [python-can]
   ubuntu:
-    '*': [python-can]
-    lucid:
-      pip:
-        packages: [python-can]
-    maverick:
-      pip:
-        packages: [python-can]
-    natty:
-      pip:
-        packages: [python-can]
-    oneiric:
-      pip:
-        packages: [python-can]
-    precise:
-      pip:
-        packages: [python-can]
-    quantal:
-      pip:
-        packages: [python-can]
-    raring:
-      pip:
-        packages: [python-can]
-    saucy:
-      pip:
-        packages: [python-can]
-    trusty:
-      pip:
-        packages: [python-can]
-    utopic:
-      pip:
-        packages: [python-can]
-    vivid:
-      pip:
-        packages: [python-can]
-    wily:
-      pip:
-        packages: [python-can]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Thank you for your suggestion. However [our policy](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules) is to use the natively packaged versions when they are available.

I don't see any documentation there saying that the ubuntu version should not be used. In general yes the packaged versions will lag behind the latest and greatest available. And the upstream development is expected to move forward and have new exciting features. However if everyone moves forward and gets the new exciting features we loose stability. If you'd like newer versions to be available I'd encourage you to push on the debian packaging to get newer versions available. However they appear to be pretty close to up to date with 3.3.2 in focal and newer, which is only 2 patch releases behind the latest release. 

https://packages.ubuntu.com/focal/python3-can
vs https://pypi.org/project/python-can/#history

You're welcome to override the rules for your project, but know if you do that you'll potentially break comatibility with anyone else who's using those packages and potentially relying on the stable release.

braamBeresford:
> Thank you for your suggestion. However [our policy](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules) is to use the natively packaged versions when they are available.
> 
> I don't see any documentation there saying that the ubuntu version should not be used. In general yes the packaged versions will lag behind the latest and greatest available. And the upstream development is expected to move forward and have new exciting features. However if everyone moves forward and gets the new exciting features we loose stability. If you'd like newer versions to be available I'd encourage you to push on the debian packaging to get newer versions available. However they appear to be pretty close to up to date with 3.3.2 in focal and newer, which is only 2 patch releases behind the latest release.
> 
> https://packages.ubuntu.com/focal/python3-can
> vs https://pypi.org/project/python-can/#history
> 
> You're welcome to override the rules for your project, but know if you do that you'll potentially break comatibility with anyone else who's using those packages and potentially relying on the stable release.

I see, I apologize. Thanks for the info!

PR #28473, opened 2021-03-04
Use apt packages for python-more-itertools

Timple:
In follow-up of the discussion here: https://github.com/ros/rosdistro/pull/19058

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6285,7 +6285,7 @@ python3-more-itertools:
   debian: [python3-more-itertools]
   fedora: [python3-more-itertools]
   gentoo: [dev-python/more-itertools]
-  rhel: [python3-more-itertools]
+  rhel: ['python%{python3_pkgversion}-more-itertools']
   ubuntu: [python3-more-itertools]
 python3-msgpack:
   debian:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2569,10 +2569,27 @@ python-more-itertools:
   debian: [python-more-itertools]
   fedora: [python-more-itertools]
   ubuntu:
-    bionic: [python-more-itertools]
+    '*': [python-more-itertools]
+    xenial: null
+python-more-itertools-pip:
+  arch:
+    pip:
+      packages: [more-itertools]
+  debian:
+    pip:
+      packages: [more-itertools]
+  fedora:
+    pip:
+      packages: [more-itertools]
+  ubuntu:
+    bionic:
+      pip: [more-itertools]
+    focal:
+      pip: [more-itertools]
+    groovy:
+      pip: [more-itertools]
     xenial:
-      pip:
-        packages: [more-itertools]
+      pip: [more-itertools]
 python-moviepy-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2564,19 +2564,15 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
-python-more-itertools-pip:
-  arch:
-    pip:
-      packages: [more-itertools]
-  debian:
-    pip:
-      packages: [more-itertools]
-  fedora:
-    pip:
-      packages: [more-itertools]
+python-more-itertools:
+  arch: [python-more-itertools]
+  debian: [python-more-itertools]
+  fedora: [python-more-itertools]
   ubuntu:
-    pip:
-      packages: [more-itertools]
+    xenial:
+      pip:
+        packages: [more-itertools]
+    bionic: [python-more-itertools]
 python-moviepy-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2564,19 +2564,15 @@ python-mock:
...
+  arch: [python-more-itertools]
+  debian: [python-more-itertools]
+  fedora: [python-more-itertools]
   ubuntu:
-    pip:
-      packages: [more-itertools]
+    xenial:
+      pip:
+        packages: [more-itertools]
+    bionic: [python-more-itertools]
```
clalancette, line 2575: This is failing CI because the list has to be in alphabetical order.

clalancette (dismissed):

rosdep/python.yaml
```diff
@@ -2564,19 +2564,15 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
-python-more-itertools-pip:
```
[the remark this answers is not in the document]
Timple: @tfoote had some arguments against that: https://github.com/ros/rosdistro/pull/19058#issuecomment-789358134

> Maintaining parallel rules with -pip and regular leads to many more problems because the pip version will mask the debian based one. This means that if any package on the system asks for the pip version all packages on the system will get the pip version, even if they asked for the apt version.

I don't have an opinion about this since we simply don't have packages with the `-pip` prefix.

So I will let you two come to a consensus and implement that :+1:

rosdep/python.yaml
Timple: Aha! I just couldn't figure out why, but this list also needs to be alphabetical. I'll fix!

rosdep/python.yaml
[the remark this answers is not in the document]
tfoote: Sorry I wasn't specific in my description. I meant sunsetting as a planned phase out. We should keep the old key, but have it not resolve on the newer platforms on which the apt key is available. To that end we should call out for Ubuntu xenial version for the pip key and not define any other versions as it's the only supported version of ubuntu still available. This will only force users of bionic or newer to switch over as they would have the conflicting definitions.

rosdep/python.yaml
```diff
@@ -2564,19 +2564,15 @@ python-mock:
...
-    pip:
-      packages: [more-itertools]
+python-more-itertools:
+  arch: [python-more-itertools]
+  debian: [python-more-itertools]
+  fedora: [python-more-itertools]
   ubuntu:
-    pip:
-      packages: [more-itertools]
+    bionic: [python-more-itertools]
```
tfoote, line 2572: This should be a wildcard rule to work for focal and other newer platforms.

Timple: But focal and newer do not use python2 anymore. So those would never be used?

rosdep/python.yaml
```diff
@@ -6273,6 +6280,9 @@ python3-mock:
   openembedded: [python3-mock@meta-ros-common]
   rhel: ['python%{python3_pkgversion}-mock']
   ubuntu: [python3-mock]
+python3-more-itertools:
+  debian: [python3-more-itertools]
+  ubuntu: [python3-more-itertools]
```
ivanpauno, line 6285: could you add the following platforms as well?

- fedora `python3-more-itertools` https://fedora.pkgs.org/33/fedora-aarch64/python3-more-itertools-8.4.0-1.fc33.noarch.rpm.html
- arch `python-more-itertools` https://archlinux.org/packages/community/any/python-more-itertools/
- rhel `'python%{python3_pkgversion}-more-itertools'` https://centos.pkgs.org/8/epel-x86_64/python3-more-itertools-7.2.0-3.el8.noarch.rpm.html
- gentoo `dev-python/more-itertools` https://packages.gentoo.org/packages/dev-python/more-itertools

ivanpauno (approved):

rosdep/python.yaml
ivanpauno: Based on the previous comment https://github.com/ros/rosdistro/pull/28473#discussion_r588783207 this seems to be fine to me.

rosdep/python.yaml
```diff
@@ -6273,6 +6280,13 @@ python3-mock:
   openembedded: [python3-mock@meta-ros-common]
   rhel: ['python%{python3_pkgversion}-mock']
   ubuntu: [python3-mock]
+python3-more-itertools:
+  arch: [python-more-itertools]
+  debian: [python3-more-itertools]
+  fedora: [python3-more-itertools]
+  gentoo: [dev-python/more-itertools]
+  rhel: [python3-more-itertools]
```
ivanpauno, line 6288: based on the others rhel rules, I think this must be:

```suggestion
  rhel: ['python%{python3_pkgversion}-more-itertools']
```

Timple: The link you included showed `python3-more-itertools`. But happy to stay in sync with the rest!

jacobperron (approved):

tfoote (approved):
Thanks for iterating on this it looks good now.

PR #28488, opened 2021-03-05
added trac_ik to noetic distribution list

swhart115:
added trac_ik to noetic distribution list

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -7308,7 +7308,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/traclabs/trac_ik-release.git
-      version: 1.5.1-1
+      version: 1.6.1-1
     source:
       type: git
       url: https://bitbucket.org/traclabs/trac_ik.git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -7293,6 +7293,27 @@ repositories:
       url: https://gitlab.com/toposens/public/ros-packages.git
       version: master
     status: developed
+  trac_ik:
+    doc:
+      type: git
+      url: https://bitbucket.org/traclabs/trac_ik.git
+      version: master
+    release:
+      packages:
+      - trac_ik
+      - trac_ik_examples
+      - trac_ik_kinematics_plugin
+      - trac_ik_lib
+      - trac_ik_python
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/traclabs/trac_ik-release.git
+      version: 1.5.1-1
+    source:
+      type: git
+      url: https://bitbucket.org/traclabs/trac_ik.git
+      version: master
+    status: developed
   tsid:
     doc:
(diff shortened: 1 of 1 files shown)


swhart115 (commented):
forgot to update the version number

clalancette:
Out of curiousity, why did you not let `bloom-release` open the PR for you?

Besides the usual reason of wondering, it also looks like there is an inconsistency here.  The `version` listed is 1.6.1-1, but in the release repo it looks like it maybe should be 1.6.1-2?

ivanpauno:
@swhart115 friendly ping, could you check the previous comment?

swhart115:
> @swhart115 friendly ping, could you check the previous comment?

Apologies for that. This was the first time i wen through the process and i didnt see where the PR was made, so i think i jumped the gun and made it myself before i ran the bloom-release script.  But since i did it myself first, and then ran that, i think i got an extra version increment.  I'm fine with either number.  what's the best fix?

ivanpauno:
> Apologies for that. This was the first time i wen through the process and i didnt see where the PR was made, so i think i jumped the gun and made it myself before i ran the bloom-release script. But since i did it myself first, and then ran that, i think i got an extra version increment. I'm fine with either number. what's the best fix?

Hi @swhart115 no problem, if both are the same release this PR is fine.
Sometimes the PR fails to be opened because of authentication problems, check this tutorial http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization.

It would also be great if you can add in the future a top level `LICENSE` file to the upstream repo, that would make licensing more obvious (though it's currently correctly specified in the package.xml files).

swhart115:
> > Apologies for that. This was the first time i wen through the process and i didnt see where the PR was made, so i think i jumped the gun and made it myself before i ran the bloom-release script. But since i did it myself first, and then ran that, i think i got an extra version increment. I'm fine with either number. what's the best fix?
> 
> Hi @swhart115 no problem, if both are the same release this PR is fine.
> Sometimes the PR fails to be opened because of authentication problems, check this tutorial http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization.
> 
> It would also be great if you can add in the future a top level `LICENSE` file to the upstream repo, that would make licensing more obvious (though it's currently correctly specified in the package.xml files).

Great, thanks for merging.  We'll get the LICENSE file in there asap.

Now just need to figure out why the buildfarm build is failing....

swhart115:
@clalancette so i'm getting build errors according (and corresponding nightly buildfarm emails).  The errors seem related specifically to the fix that we pushed into the code and resolved the issue on our local systems.  

The error seems to be wrt not finding the new 20.04 libnlopt-cxx-dev dependency that causes the error of it not being able to find the nlopt.h header.  Our change was to add a conditioninal in the package.xml of trac_ik_lib to add this depends if ROS_DISTRO=noetic.  but again, this seems to not be going through on the build server. Not sure the best way to debug this. 

Any pointers on who to ping to help work this issue out?

PR #28495, opened 2021-03-06
add python3-bluez rosdep key

mikaelarguedas:
Debian: https://packages.debian.org/buster/python3-bluez
Fedora: https://fedora.pkgs.org/32/fedora-x86_64/python3-bluez-0.22-18.fc32.x86_64.rpm.html
Gentoo: https://packages.gentoo.org/packages/dev-python/pybluez
Ubuntu: https://packages.ubuntu.com/focal/python3-bluez

Some widely used packages like [ps3joy](https://github.com/ros-drivers/joystick_drivers/blob/0eb32c2eebb68a7a0066ef288f1a1aa793d08650/ps3joy/package.xml#L33) rely on pybluez

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5579,6 +5579,11 @@ python3-bitarray:
   fedora: [python3-bitarray]
   gentoo: [dev-python/bitarray]
   ubuntu: [python3-bitarray]
+python3-bluez:
+  debian: [python3-bluez]
+  fedora: [python3-bluez]
+  gentoo: [dev-python/pybluez]
+  ubuntu: [python3-bluez]
 python3-bokeh-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5583,7 +5583,10 @@ python3-bluez:
   debian: [python3-bluez]
   fedora: [python3-bluez]
   gentoo: [dev-python/pybluez]
-  ubuntu: [python3-bluez]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5579,6 +5579,11 @@ python3-bitarray:
   fedora: [python3-bitarray]
   gentoo: [dev-python/bitarray]
   ubuntu: [python3-bitarray]
+python3-bluez:
+  debian: [python3-bluez]
+  fedora: [python3-bluez]
+  gentoo: [dev-python/pybluez]
+  ubuntu: [python3-bluez]
```
clalancette, line 5586: It looks like `python3-bluez` is only supported on `focal` forward, so I'll suggest:

```
  ubuntu:
    '*': [python3-bluez]
    'bionic': null
    'xenial': null
```

clalancette (changes requested):

rosdep/python.yaml
mikaelarguedas, line 5586:
```suggestion
  ubuntu:
    '*': [python3-bluez]
    'bionic': null
    'xenial': null
```

mikaelarguedas: Good point, added as a suggestion in https://github.com/ros/rosdistro/pull/28495#discussion_r589637724

clalancette (approved):

nuclearsandwich (approved):

PR #28384, opened 2021-02-22
t_mapping: 0.1.3-1 in 'kinetic/distribution.yaml' [bloom]

tbostic32:
Increasing version of package(s) in repository `t_mapping` to `0.1.3-1`:

- upstream repository: https://github.com/mitre/t_mapping.git
- release repository: https://github.com/tbostic32/t_mapping-release.git
- distro file: `kinetic/distribution.yaml`
- bloom version: `0.10.1`
- previous version for package: `null`

PR #28505, opened 2021-03-08
mitre_fast_layered_map: 0.1.4-2 in 'kinetic/distribution.yaml' [bloom]

tbostic32:
Increasing version of package(s) in repository `mitre_fast_layered_map` to `0.1.4-2`:

- upstream repository: https://github.com/mitre/mitre_fast_layered_map.git
- release repository: https://github.com/mitre/mitre_fast_layered_map-release.git
- bloom version: `0.10.1`

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -7701,6 +7701,21 @@ repositories:
       url: https://github.com/dfki-ric/mir_robot.git
       version: kinetic
     status: developed
+  mitre_fast_layered_map:
+    doc:
+      type: git
+      url: https://github.com/mitre/mitre_fast_layered_map.git
+      version: kinetic-devel
+    release:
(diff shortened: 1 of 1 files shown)


tbostic32:
This MR fixed the issues raised in https://github.com/ros/rosdistro/pull/28384. 

Also, I don't know if others have run into this, but in order for bloom to submit this PR I had to give the access token workflows permissions. That had not been the case other times I created a PR.

tfoote:
@tbostic32 The workflows option/control is new. Github may have reorganized some of the API access under that. If you or anyone can confirm that it would be good to look at what API call is blocked and see if we can avoid that since we don't really need to change any workflow contents. Could you open a ticket on bloom with more details? https://github.com/ros-infrastructure/bloom/issues

PR #28514, opened 2021-03-09
Added Pyrr Maths Library

belorenz:
Added Pyrr Maths Library.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -382,8 +382,8 @@ pyros-setup-pip:
       packages: [pyros-setup]
 pyrr-pip:
   ubuntu:
-  pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -380,6 +380,10 @@ pyros-setup-pip:
   ubuntu:
     pip:
       packages: [pyros-setup]
+pyrr-pip:
```
clalancette, line 383: Can you please make this key follow the guidance here: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1 ?

clalancette (changes requested):

PR #28532, opened 2021-03-10
Adding new repository for indexing: rclada

mosteo:
Adding Ada client library main repository for initial documentation indexing.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2676,8 +2676,32 @@ repositories:
   rclada:
     doc:
       type: git
-      url: https://github.com/ada-ros/ada4ros2.git
-      version: foxy    
+      url: https://github.com/ada-ros/rclada.git
+      version: foxy
+    source:
+      type: git
+      url: https://github.com/ada-ros/rclada.git
+      version: foxy
+    status: developed
+  rclada_common:
+    doc:
+      type: git
+      url: https://github.com/ada-ros/rclada_common.git
+      version: foxy
+    source:
+      type: git
+      url: https://github.com/ada-ros/rclada_common.git
+      version: foxy
+    status: developed
+  rclada_examples:
+    doc:
+      type: git
+      url: https://github.com/ada-ros/rclada_examples.git
+      version: foxy
+    source:
+      type: git
+      url: https://github.com/ada-ros/rclada_examples.git
+      version: foxy
     status: developed
   rclc:
     doc:
@@ -3454,6 +3478,15 @@ repositories:
       url: https://github.com/ros2/rosidl_defaults.git
       version: foxy
     status: maintained
+  rosidl_generator_ada:
+    doc:
+      type: git
+      url: https://github.com/ada-ros/rosidl_generator_ada.git
+      version: foxy
+    source:
+      type: git
+      url: https://github.com/ada-ros/rosidl_generator_ada.git
+      version: foxy
   rosidl_python:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


mosteo:
I'm reading in the review guidelines:

"The source repository should contain one or more ROS packages (meaning they have a package.xml in the source repository)"

The repository I intend to submit has the packages as submodules. Is this acceptable? Thanks.

ivanpauno:
> The repository I intend to submit has the packages as submodules. Is this acceptable? Thanks.

AFAIK we don't have any policy for that.
It sounds fine to me, submodules are automatically cloned when cloning the repo, but IDK what others think.

@nuclearsandwich what do you think?

foxy/distribution.yaml
```diff
@@ -2673,6 +2673,12 @@ repositories:
       url: https://github.com/ros2/rcl_logging.git
       version: foxy
     status: maintained
+  rclada:
+    doc:
```
ivanpauno, line 2677: I would add a `src` entry as well

foxy/distribution.yaml
```diff
@@ -2673,6 +2673,12 @@ repositories:
       url: https://github.com/ros2/rcl_logging.git
       version: foxy
     status: maintained
+  rclada:
```
ivanpauno, line 2676: shouldn't this match the repo name?

foxy/distribution.yaml
```diff
@@ -2673,6 +2673,12 @@ repositories:
...
+    doc:
+      type: git
+      url: https://github.com/ada-ros/ada4ros2.git
+      version: foxy
```
clalancette, line 2680: There's trailing whitespace at the end of this line, which is why CI is failing.

clalancette:
> @nuclearsandwich what do you think?
> 
> edit: if this repo will have an entry, I think than entries for the submodules shouldn't be added, but IDK.

This setup isn't really how the buildfarm works.

What you seem to have here is a "meta" repository that contains all of the other repositories, plus some scripts and other things.  But the buildfarm here really only deals in repositories with one or more packages in it.  So I don't think this is a good way to go about this.

Instead, I'll suggest adding one entry for each of the submodule entries, i.e. https://github.com/ada-ros/rosidl_generator_ada, https://github.com/ada-ros/rclada_common , etc.  That matches much more closely how everything else in this file is setup.

ivanpauno:
> Instead, I'll suggest adding one entry for each of the submodule entries, i.e. https://github.com/ada-ros/rosidl_generator_ada, https://github.com/ada-ros/rclada_common , etc. That matches much more closely how everything else in this file is setup.

:+1: that matches more closely what we do in other places.

> But the buildfarm here really only deals in repositories with one or more packages in it. So I don't think this is a good way to go about this.

I don't think the buildfarm will realize, submodules are transparent (i.e. the buildfarm will still think these are many packages in one repo). But I might be wrong and there's also no precedent of doing this.

nuclearsandwich:
> > 
> 
> AFAIK we don't have any policy for that.
> It sounds fine to me, submodules are automatically cloned when cloning the repo, but IDK what others think.

I think it makes the best sense to add the package repositories directly. Submodule support if it exists everywhere is going to be somewhat accidental and the layer of indirection would be unexpected for most of the ROS community.

mosteo:
Thanks for the review, I will do as suggested.

ivanpauno (approved):

ivanpauno:
Thanks @mosteo !

PR #28563, opened 2021-03-11
Add python3-pyside2.qtopengl for Ubuntu focal

raultron:
Package in Noetic: 
https://packages.ubuntu.com/focal/python/python3-pyside2.qtopengl

This package allows the creation of an opengl widget in rqt when using the pyside backend.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6719,6 +6719,9 @@ python3-pyrr-pip:
   ubuntu:
     pip:
       packages: [pyrr]
+python3-pyside2.qtopengl:
+  ubuntu:
+    noetic: [python3-pyside2.qtopengl]
 python3-pystemd:
   debian:
     bullseye: [python3-pystemd]
@@ -6730,9 +6733,6 @@ python3-pystemd:
     xenial:
       pip:
         packages: [pystemd]
-python3-pyside2.qtopengl:
-  ubuntu:
-    noetic: [python3-pyside2.qtopengl]
 python3-pyswarms-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6720,6 +6720,9 @@ python3-pyrr-pip:
     pip:
       packages: [pyrr]
 python3-pyside2.qtopengl:
+  debian: [python3-pyside2.qtopengl]
+  fedora: [python3-pyside2]
+  gentoo: [dev-python/pyside2]
   ubuntu:
     '*': [python3-pyside2.qtopengl]
     bionic: null
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6892,7 +6892,7 @@ python3-rosdistro-modules:
   rhel: ['python%{python3_pkgversion}-rosdistro']
   ubuntu: [python3-rosdistro-modules]
 python3-rosinstall-generator:
-  ubuntu: [python3-rosinstall-generator]raultron:patch-1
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6719,6 +6719,9 @@ python3-pyrr-pip:
   ubuntu:
     pip:
       packages: [pyrr]
+python3-pyside2.qtopengl:
+  ubuntu:
+    focal: [python3-pyside2.qtopengl]
```
ivanpauno, line 6724:
```suggestion
    '*': [python3-pyside2.qtopengl]
    bionic: null
    xenial: null
```

ivanpauno (changes requested):
is this package available in `arch`, `debian`, `fedora`, `rhel`, `gentoo`?

If yes, it would be great if you can add that to the rules and provide links, e.g.: https://fedora.pkgs.org/32/fedora-x86_64/python3-pyside2-devel-5.13.2-2.fc32.x86_64.rpm.html (though that's equivalent to all pyside2.* packages).

mabelzhang:
The comment above stands. Please add fedora, debian, and gentoo entries if those packages exist. See https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md for details.

The Fedora link is above. I believe it's okay to just take all of `python-pyside` for Fedora, as the `python-pyside.qtuitools` entry has done:
https://github.com/ros/rosdistro/blob/36b9d77b343a5f751557d797fd3c2d212554d485/rosdep/python.yaml#L3647

Similarly for Gentoo, `dev-python/pyside2` seems to have a dependency on `dev-qt/qtopengl`
https://packages.gentoo.org/packages/dev-python/pyside2/dependencies

Package also exists on Debian https://packages.debian.org/buster/python3-pyside2.qtopengl

Please add the three entries above.

rosdep/python.yaml
```diff
@@ -6719,6 +6719,11 @@ python3-pyrr-pip:
   ubuntu:
     pip:
       packages: [pyrr]
+python3-pyside2.qtopengl:
+  ubuntu:
```
mabelzhang, line 6723: Fedora 32 https://fedora.pkgs.org/32/fedora-x86_64/python3-pyside2-5.13.2-2.fc32.x86_64.rpm.html
Fedora 33 https://fedora.pkgs.org/33/fedora-x86_64/python3-pyside2-5.15.0-4.fc33.x86_64.rpm.html
I think this might do:
```suggestion
  debian: [python3-pyside2.qtopengl]
  fedora: [python3-pyside2]
  gentoo: [dev-python/pyside2]
  ubuntu:
```

raultron: Thanks for your help, I was a little bit blocked since I only know the Ubuntu packages.

rosdep/python.yaml
```diff
@@ -6719,6 +6719,15 @@ python3-pyrr-pip:
   ubuntu:
     pip:
       packages: [pyrr]
+python3-pyside2.qtopengl:
+  arch: [pyside2]
+  debian: [python3-pyside2.qtopengl]
```
ivanpauno, line 6724:
```suggestion
  debian:
    '*': [python3-pyside2.qtopengl]
    jessie: null
    stretch: null
```

seems to be available from debian buster

ivanpauno (approved):

mabelzhang (approved):
Jessie is out of support now, and we're no longer building for it.
Stretch is also out of support.
So we don't actually need those anymore.
I'll let a +1 confirm that and merge.

PR #28559, opened 2021-03-11
rmw_connextdds: 0.1.1-1 in 'rolling/distribution.yaml' [non-bloom]

asorbini:
Introduce [rticommunity/rmw_connextdds](https://github.com/rticommunity/rmw_connextdds) (`0.1.1-1`) to `rolling`.

This PR is linked to work being done to phase out [ros2/rmw_connext](https://github.com/ros2/rmw_connext) (see [rticommunity/rmw_connextdds#9](https://github.com/rticommunity/rmw_connextdds/issues/9)).

The following packages will be added:
- `rmw_connextdds`
- `rmw_connextddsmicro`
- `rmw_connextdds_common`
- `rti_connext_dds_cmake_module`

The packages depend on RTI Connext DDS 5.3.1+.
Package `rmw_connextddsmicro` will not actually include an RMW implementation, since the library generation is disabled if RTI Connext Micro is not available (which will be the case).

The release repository currently points to [rticommunity/rmw_connextdds-release](https://github.com/rticommunity/rmw_connextdds-release), only because the name of the [repository in the `ros2-gbp`](https://github.com/ros2-gbp/rmw_connextdds) org needs to be updated with the `-release` suffix. Once the name is fixed, I will push the existing commits to the new repository, or re-run bloom with the new repository. @clalancette (or anyone with access) can you update the repository name?

I'm not an expert on this release process (first time going through it), so I welcome any input.

While reviewing the packages to file this PR, I noticed that `rti_connext_dds_cmake_module` is missing a build dependency from `rti-connext-dds-5.3.1`, so I'll probably have to create a new release (which I can do with the release repository in the `ros2-gbp` org).

PR #28572, opened 2021-03-11
rmw_connextdds: 0.2.1-1 in 'rolling/distribution.yaml' [non-bloom]

asorbini:
Follow up to #28559 using release repository [ros2-gbp/rmw_connextdds-release](https://github.com/ros2-gbp/rmw_connextdds-release).

I'm creating the PR by hand because I wasn't able to create an OAuth token with bloom, even though I used the correct password (I had received the same error several times while preparing for #28559).

```
Would you like to create an OAuth token now [Y/n]? 
GitHub username [asorbini]: 
GitHub password (never stored): 
HTTP Error 404: Not Found (https://api.github.com/authorizations): None

This sometimes fails when the username or password are incorrect, try again?
Continue [Y/n]? n
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1728,6 +1728,26 @@ repositories:
       url: https://github.com/ros2/rmw_connext.git
       version: master
     status: maintained
+  rmw_connextdds:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):
lgtm!

PR #28593, opened 2021-03-12
rmw_gurumdds: 0.7.10-1 in 'dashing/distribution.yaml' [bloom]

YoungJin-gurum:


nuclearsandwich:
Hi @YoungJin-gurum, these pull requests are normally created automatically using Bloom and update one ROS distribution at a time. Is there something preventing you from using Bloom's automatic pull request generation which is the preferred way to make rosdistro pull requests.

If you get an error when creating the pull request you can add the `--pull-request-only` flag to bloom-release in order to retry.

YoungJin-gurum:
Hi @nuclearsandwich, thank you for kind reply. I got error when creating the pull request through bloom, HTTP Error 401: Unauthorized.
I check http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization, edit ~/.config/bloom, .gitconfig, but failed.
Even though I fork the repository manually, I get an error saying that the fork could not be found.
If I create pull request containing only one commit, is it okay?

PR #28595, opened 2021-03-12
Adding ROSXBee to documentation index for foxy

Sudharsan10:
I'd like my repo to be indexed and documented on ros.org


Sudharsan10:
I made a mistake would like to do it again once i get it right.

PR #28596, opened 2021-03-12
Adding ROSXBee to documentation index for foxy

Sudharsan10:
I'd like ROSXBee to be indexed and documented on the ros.org

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -3617,6 +3617,18 @@ repositories:
       type: git
       url: https://github.com/Sudharsan10/ROSXBee.git
       version: foxy
+    release:
+      packages:
+      - rosxbeepy
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/Sudharsan10/ROSXBee-release.git
+      version: 0.0.1-2
+    source:
+      type: git
+      url: https://github.com/Sudharsan10/ROSXBee.git
+      version: foxy
+    status: developed
   roverrobotics_ros2:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -3612,6 +3612,11 @@ repositories:
       url: https://github.com/ros2/rosidl_typesupport_gurumdds.git
       version: foxy
     status: developed
+  rosxbee:
+    doc:
```
ivanpauno, line 3616: could you add a `source` entry?

ivanpauno (changes requested):
the package.xml must have the license entry filled https://github.com/Sudharsan10/ROSXBee/blob/65c24381f598f6bcdf41587f0fdf488138bd96dd/code/src/rosxbeepy/package.xml#L9.

ivanpauno:
do you need both PRs or can this be closed in favor of https://github.com/ros/rosdistro/pull/28597/?

foxy/distribution.yaml
Sudharsan10: Thanks @ivanpauno I will make an entry and will update here.
Also will fill the license information in the package.

Sudharsan10:
> do you need both PRs or can this be closed in favor of #28597?

yes you can close this one. Also would love to know why my pull request failed in the automated test so that I can learn resubmit my pull request again for the #28597.

foxy/distribution.yaml
Sudharsan10: @ivanpauno I have made both the changes, Let me know if I should anything else. I'm new to this so dont mind if I ask dumb questions

PR #28599, opened 2021-03-12
Added simple-pid pip package

joel-mb:
Added simple-pid pip package: https://pypi.org/project/simple-pid/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7036,6 +7036,19 @@ python3-shapely:
     pip:
       packages: [shapely]
   ubuntu: [python3-shapely]
+python3-simple-pid-pip:
+  debian:
+    pip:
+      packages: [simple-pid]
+  fedora:
+    pip:
+      packages: [simple-pid]
+  osx:
+    pip:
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):

rosdep/python.yaml
```diff
@@ -7036,6 +7036,19 @@ python3-shapely:
...
       packages: [shapely]
   ubuntu: [python3-shapely]
+python3-simple-pid-pip:
+  debian:
+    pip:
+      packages: [simple-pid]
+  fedora:
+    pip:
+      packages: [simple-pid]
+  osx:
```
mabelzhang, line 7046:
```suggestion
  gentoo:
    pip:
      packages: [simple-pid]
  osx:
```

mabelzhang (commented):
Gentoo is suggested if the package exists. In this case, since it's `pip`, I think it exists on Gentoo.

rosdep/python.yaml
joel-mb: Done!

jacobperron (approved):

PR #28620, opened 2021-03-12
Make the libspnav-dev key depend on libx11-dev.

clalancette:
The header files depend on it, but it is missing an explicit
dependency on it.  Add that into the rosdep database.

This should fix the build of spacenav in Rolling that is currently failing: https://build.ros2.org/job/Rbin_uF64__spacenav__ubuntu_focal_amd64__binary/1/

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4175,11 +4175,11 @@ libspatialite:
     trusty: [libspatialite5]
 libspnav-dev:
   arch: [libspnav]
-  debian: [libspnav-dev]
+  debian: [libspnav-dev, libx11-dev]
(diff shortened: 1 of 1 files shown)


cottsay (approved):
It doesn't look like there is a bug filed against Ubuntu to fix this, but that shouldn't block this PR: https://launchpad.net/ubuntu/+source/libspnav/+bugs

ivanpauno (approved):

clalancette:
Thanks for the reviews!  I'll attempt to open a bug against libspnav (if I can figure out my Launchpad password).  I'm going to merge this for now.

PR #28642, opened 2021-03-15
Add pydantic pip package to rosdep

ipa-fez:
We're using pydantic (https://pydantic-docs.helpmanual.io/) to generate json schemata.

On focal, there's a python3-pydantic package available, but previous versions don't seem to have one yet, so I specified pip instead.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6592,6 +6592,15 @@ python3-pycryptodome:
       packages: [pycryptodome]
   rhel: ['python%{python3_pkgversion}-pycryptodomex']
   ubuntu: [python3-pycryptodome]
+python3-pydantic:
+  arch: [python-pydantic]
+  debian: null
+  fedora: [python3-pydantic]
+  gentoo: [dev-python/pydantic]
+  ubuntu:
+    '*': [python3-pydantic]
+    bionic: null
+    xenial: null
 python3-pydot:
   alpine: [py3-pydot]
   arch: [python-pydot]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6596,15 +6596,20 @@ python3-pydantic:
   arch: [python-pydantic]
   debian:
     '*': [python3-pydantic]
-    buster: null
-    jessie: null
-    stretch: null
+    buster:
+      pip: [pydantic]
+    jessie:
+      pip: [pydantic]
+    stretch:
+      pip: [pydantic]
   fedora: [python3-pydantic]
   gentoo: [dev-python/pydantic]
   ubuntu:
     '*': [python3-pydantic]
-    bionic: null
-    xenial: null
+    bionic:
+      pip: [pydantic]
+    xenial:
+      pip: [pydantic]
 python3-pydot:
   alpine: [py3-pydot]
   arch: [python-pydot]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6594,7 +6594,11 @@ python3-pycryptodome:
   ubuntu: [python3-pycryptodome]
 python3-pydantic:
   arch: [python-pydantic]
-  debian: null
+  debian:
+    '*': [python3-pydantic]
+    buster: null
+    jessie: null
+    stretch: null
   fedora: [python3-pydantic]
   gentoo: [dev-python/pydantic]
   ubuntu:


rosdep/python.yaml
```diff
@@ -269,6 +269,13 @@ pika:
     pip:
       packages: [pika]
   ubuntu: [python-pika]
+pydantic:
```
ivanpauno, line 272: I think that according to the naming conventions this should be `python-pydantic-pip`

ipa-fez: yea, you're probably right. I've renamed it.

ivanpauno (approved):

mabelzhang (changes requested):
Do you anticipate running on platforms other than Focal? System dependencies are much preferred to `pip` ones, as explained here https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md

If it doesn't make much difference to you, then it should be added as a system dependency, at least for Focal. There are other examples of using system dependencies for newer distributions and pip packages for older ones: https://github.com/ros/rosdistro/blob/36b9d77b343a5f751557d797fd3c2d212554d485/rosdep/python.yaml#L4375-L4391
That way when the old distros go out of support, we have the system dependencies for newer distros.

In addition, the system dependencies do exist on Fedora and Gentoo. Please add those entries.
https://src.fedoraproject.org/rpms/python-pydantic/
https://packages.gentoo.org/packages/dev-python/pydantic

ipa-fez:
No, focal support is all we need at the moment. Unfortunately the native package is quite a bit older than the pip version, but we can work around that.

I've also moved it to the python3 section since it requires python3, I hope that is the correct place to put it.

ivanpauno:
> If it doesn't make much difference to you, then it should be added as a system dependency, at least for Focal. There are other examples of using system dependencies for newer distributions and pip packages for older ones:

I think the preference now is to not mix pip and system dependencies under the same key, so where system dependencies aren't available the entry should be set to `null` instead (I'm not sure why I didn't notice this in the first review :joy:).

rosdep/python.yaml
```diff
@@ -6585,6 +6585,21 @@ python3-pycryptodome:
       packages: [pycryptodome]
   rhel: ['python%{python3_pkgversion}-pycryptodomex']
   ubuntu: [python3-pycryptodome]
+python3-pydantic:
+  arch: [python-pydantic]
+  debian:
+    pip:
+      packages: [pydantic]
```
ivanpauno, line 6592:
```suggestion
  debian: null
```

rosdep/python.yaml
```diff
@@ -6585,6 +6585,21 @@ python3-pycryptodome:
...
+  fedora: [python-pydantic]
+  gentoo: [dev-python/pydantic]
+  ubuntu:
+    '*': [python3-pydantic]
+    bionic:
+      pip:
+        packages: [pydantic]
+    xenial:
+      pip:
+        packages: [pydantic]
```
ivanpauno, line 6602:
```suggestion
    bionic: null
    xenial: null
```

ivanpauno (changes requested):

ipa-fez:
Thanks for the info, I've removed the pip entries. I hope it's correct now.

rosdep/python.yaml
```diff
@@ -6592,6 +6592,15 @@ python3-pycryptodome:
       packages: [pycryptodome]
   rhel: ['python%{python3_pkgversion}-pycryptodomex']
   ubuntu: [python3-pycryptodome]
+python3-pydantic:
+  arch: [python-pydantic]
+  debian: null
+  fedora: [python-pydantic]
```
ivanpauno, line 6598:
```suggestion
  fedora: [python3-pydantic]
```

https://fedora.pkgs.org/33/fedora-aarch64/python3-pydantic-1.6.1-1.fc33.noarch.rpm.html

`python-pydantic` key is obsolete AFAIS

ipa-fez: thanks! i've updated it.

ivanpauno (approved):

rosdep/python.yaml
```diff
@@ -6592,6 +6592,15 @@ python3-pycryptodome:
       packages: [pycryptodome]
   rhel: ['python%{python3_pkgversion}-pycryptodomex']
   ubuntu: [python3-pycryptodome]
+python3-pydantic:
+  arch: [python-pydantic]
+  debian: null
```
jacobperron, line 6597: The package is available in the upcoming Bullseye: https://packages.debian.org/bullseye/python3-pydantic
```suggestion
  debian:
    bullseye: [python3-pydantic]
```

ivanpauno: or to make it future proof:
```suggestion
  debian:
    '*': [python3-pydantic]
    buster: null
    jessie: null
    stretch: null
```

ipa-fez: I added the latter.

jacobperron (approved):

tfoote (changes requested):
> I think the preference now is to not mix pip and system dependencies under the same key, so where system dependencies aren't available the entry should be set to null instead.

Sorry for the back and forth. We haven't changed this policy. The pip rules are fine to fill in for platforms on which we do not ever expect native packages, as long as the pip version is expected to be compatible. This will enable usage from source on the older platforms and not potentially collide with a system installation.

So I'd suggest restoring the versions including the pip backfilled versions.

mabelzhang (approved):
I THINK the latest comment has been addressed. Just removing my request for change.

ipa-fez:
yep, I've re-added the pip entries.

tfoote (approved):

PR #28672, opened 2021-03-17
Add python3-dlib-pip

MatthijsBurgh:
This matches the guidelines (https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1), keep the non pip entry to prevent any issues

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1409,13 +1409,6 @@ python-dlib:
     pip: [dlib]
   ubuntu:
     pip: [dlib]
-python-dlib-pip:
-  debian:
-    pip: [dlib]
-  fedora:
-    pip: [dlib]
-  ubuntu:
-    pip: [dlib]
 python-docker:
   arch: [python-docker]
   debian: [python-docker]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Please edit your description provide resource links following the CONTRIBUTING guide for review. 

As this is a new pip key it should use `python3` as the prefix not just python.

tfoote:
Looking at this a little more deeply this appears to already be packaged for ubuntu:

https://packages.ubuntu.com/source/bionic/dlib

The pypi project just points to dlib.net which is the same thing that the above package points to: https://pypi.org/project/dlib/#description

And pulling the files from pypi also shows all the c++ headers etc. As such we should use the debian package and not the pypi version. I saw that the main release have the python bindings: https://learnopencv.com/install-dlib-on-ubuntu/ But the debian package may not generate the python bindings though. But if it doesn't these pip rules will need dependencies on at least `cmake` `build-essential` and `python3-dev`

hidmic:
@MatthijsBurgh mind to comment on @tfoote observations?

MatthijsBurgh:
When I remove the pip package and install the apt packages. I don't have a dlib python module. So I think the pip entry is needed.

MatthijsBurgh:
I don't know how to add the deps on cmake, etc

tfoote:
Dependency declarations are documented here: https://www.ros.org/reps/rep-0111.html#dependencies

Here's an example: https://github.com/ros/rosdistro/blob/6e6d58afce680f64e1cd06ef749237adb3ce513f/rosdep/python.yaml#L979

MatthijsBurgh:
@tfoote friendly ping ;)

clalancette (approved):

jacobperron (approved):

PR #28794, opened 2021-03-22
Adding python3-evdev rosdep

jstiefel:

## Package name:

python3-evdev

## Package Upstream Source:

https://github.com/gvalkov/python-evdev

## Purpose of using this:

Adding the Python 3 version of evdev to rosdep. The package provides bindings to the generic input event interface in Linux. 

Distro packaging links:

  - https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=python3-evdev
   - https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-evdev&searchon=names
  - Service Unavailable
  - Python3 package not available
  - Python3 package not available
  - added pip

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1505,6 +1505,7 @@ python-espeak:
     trusty: [python-espeak]
     trusty_python3: [python3-espeak]
 python-evdev:
+  gentoo: [dev-python/python-evdev]
   ubuntu: [python-evdev]
 python-expiringdict:
   debian: [python-expiringdict]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5906,6 +5906,7 @@ python3-evdev:
     jessie: null
     stretch: null
   fedora: [python3-evdev]
+  gentoo: [dev-python/python-evdev]
   osx:
     pip:
       packages: [evdev]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5900,7 +5900,10 @@ python3-empy:
   rhel: ['python%{python3_pkgversion}-empy']
   ubuntu: [python3-empy]
 python3-evdev:
-  debian: [python3-evdev]
+  debian:
+    '*': [python3-evdev]
+    jessie: null
+    stretch: null
   fedora: [python3-evdev]
   osx:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5900,6 +5900,7 @@ python3-empy:
   rhel: ['python%{python3_pkgversion}-empy']
   ubuntu: [python3-empy]
 python3-evdev:
+  arch: [python-evdev]
   debian:
     '*': [python3-evdev]
     jessie: null
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1510,10 +1510,10 @@ python-evdev:
 python3-evdev:
   debian: [python3-evdev]
   fedora: [python3-evdev]
-  ubuntu: [python3-evdev]
   osx:
     pip:
       packages: [evdev]
+  ubuntu: [python3-evdev]
 python-expiringdict:
   debian: [python-expiringdict]
   fedora:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1507,6 +1507,13 @@ python-espeak:
 python-evdev:
   gentoo: [dev-python/python-evdev]
   ubuntu: [python-evdev]
+python3-evdev:
+  debian: [python3-evdev]
+  fedora: [python3-evdev]
+  ubuntu: [python3-evdev]
+  osx:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5899,6 +5899,13 @@ python3-empy:
   openembedded: [python3-empy@meta-ros-common]
   rhel: ['python%{python3_pkgversion}-empy']
   ubuntu: [python3-empy]
+python3-evdev:
+  debian: [python3-evdev]
```
jacobperron, line 5903: We should null out supported distros where this package is not available. I would add entries for jessie and stretch as they are recommended support for Kinetic and Melodic which are not EOL.
```suggestion
  debian:
    '*': [python3-evdev]
    jessie: null
    stretch: null
```

rosdep/python.yaml
```diff
@@ -5899,6 +5899,13 @@ python3-empy:
...
+  fedora: [python3-evdev]
```
jacobperron, line 5904: Here's the equivalent Gentoo package: https://packages.gentoo.org/packages/dev-python/python-evdev
```suggestion
  fedora: [python3-evdev]
  gentoo: [dev-python/python-evdev]
```

rosdep/python.yaml
```diff
@@ -5899,6 +5899,13 @@ python3-empy:
   openembedded: [python3-empy@meta-ros-common]
   rhel: ['python%{python3_pkgversion}-empy']
   ubuntu: [python3-empy]
+python3-evdev:
```
jacobperron, line 5902: Here's the equivalent Arch package: https://archlinux.org/packages/extra/x86_64/python-evdev/
```suggestion
python3-evdev:
  arch: [python-evdev]
```

rosdep/python.yaml
```diff
@@ -5899,6 +5899,13 @@ python3-empy:
...
+  debian: [python3-evdev]
+  fedora: [python3-evdev]
+  osx:
+    pip:
```
jacobperron, line 5906: I think we prefer to avoid mixing pip packages with system packages under the same rosdep key. So I think this should be moved to it's own rule named `python3-evdev-pip`. It would be good to get someone else to confirm.

jacobperron (changes requested):
Note, Arch and Gentoo package don't typically have the `python3` prefix.

rosdep/python.yaml
tfoote: It's fine to have pip rules to fill in alternative platforms and older distros where there's not expected to be a conflicting native package in the future.

rosdep/python.yaml
jstiefel: This is the same as the one for the Python 2 package then? Should we remove this from this one?

https://github.com/ros/rosdistro/blob/45fdf02ae88ea2116fba6f390f5d61fd70a626b7/rosdep/python.yaml#L1507

rosdep/python.yaml
jstiefel: How do you know that this is Python 3?

tfoote: The python version matches the system. Arch switch to python3 a while ago.

rosdep/python.yaml
tfoote: No like Arch Gentoo moved to python3 by default already. All python rules on Arch and Gentoo actually resolve to python3.

tfoote (commented):
This looks like it will be good with the suggestions from Jacob if you could accept them.

jstiefel:
Thanks @jacobperron and @tfoote for the reviews.

rosdep/python.yaml
```diff
@@ -1505,7 +1505,6 @@ python-espeak:
     trusty: [python-espeak]
     trusty_python3: [python3-espeak]
 python-evdev:
-  gentoo: [dev-python/python-evdev]
```
jacobperron, line 1508: I think we should still keep the non-python3 rule (at least I don't think it's hurting anyone to leave it).

jstiefel: But if I understood correctly, this Python 2 rule is then linking to a Python 3 package?

jacobperron: I think that's correct. I'm just hesitant removing an existing key. @tfoote Could there be any negative consequences to removing the old entry?

tfoote: We should not remove this rule. This is an intricacy of the Gentoo deployment.

All python keys are inherently the system version on Gentoo now. They're not "incorrect" as such, code that is running on Gentoo is going to run with the system version of python, whichever is selected by the system maintainer. This works with our policy that best policy is that code should be written to be 2/3 compatible if it's 2 compatible. After Noetic and in ROS 2 everything is expected to be python3.

jacobperron (approved):
Thanks for iterating!

PR #28782, opened 2021-03-22
Added python package smbus2

maarufvazifdar:
1) I have added the rules for smbus2 package in python.yaml file. [https://pypi.org/project/smbus2/].

2) Earlier I added jetson-stats package [https://pypi.org/project/jetson-stats/] but found that it is not working with rosdep. Also to install the package manually "pip install jetson-stats" does not work but "sudo -H pip install -U jetson-stats" works. So, how should I define it in the python.yaml file?

@cottsay @hidmic

PR #28799, opened 2021-03-23
Add jetson-stats package 

maarufvazifdar:
Hi, earlier I added jetson-stats package [https://pypi.org/project/jetson-stats/] but found that it is not working with rosdep. Also to install the package manually "pip install jetson-stats" does not work but "sudo -H pip install -U jetson-stats" works. Please guide me on how to define it in the python.yaml file?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1919,6 +1919,10 @@ python-gpiozero:
       pip:
         packages: [gpiozero]
     zesty: [python-gpiozero]
+python-gputil-pip:
+  ubuntu:
+    pip:
+      packages: [gputil]
 python-gpxpy:
   debian:
     buster: [python-gpxpy]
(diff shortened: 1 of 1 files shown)


mabelzhang:
For reference, this is the previous PR mentioned https://github.com/ros/rosdistro/pull/28280

Could you merge from `main` so that the code is updated? It looks like one of the blocks has already merged in the PR above. The other one is waiting for approval here https://github.com/ros/rosdistro/pull/28782/files

In the current state, this is a question and not a PR, and it's more suited if you open an issue instead.

As for the question itself, I'm not familiar enough with pip keys to give an answer...

tfoote:
rosdep installs all python packages via sudo automatically. 

Related you can test the rosdep rules by adding your proposed updated rosdep keys to the rosdep sources locally.

Since this appears to have the same content as #28782 I'm going to close this to avoid clutter. Please open a new PR with the jetsons-stats proposal.

PR #28802, opened 2021-03-23
Update libcoin80-dev system dependency for noetic

fspindle:
Since Ubuntu focal, `libcoin80-dev` package was renamed in `libcoin-dev`

See: https://packages.ubuntu.com/search?keywords=libcoin-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2285,7 +2285,11 @@ libcoin80-dev:
     jessie: [libcoin80-dev]
     stretch: [libcoin80-dev]
   fedora: [Coin2-devel]
-  ubuntu: [libcoin80-dev]
+  ubuntu:
+    xenial: [libcoin80-dev]
+    bionic: [libcoin80-dev]
+    focal: [libcoin-dev]
+    groovy: [libcoin-dev]
 libconfig++-dev:
   debian: [libconfig++-dev]
   gentoo: ['dev-libs/libconfig[cxx]']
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -2285,7 +2285,11 @@ libcoin80-dev:
     jessie: [libcoin80-dev]
     stretch: [libcoin80-dev]
   fedora: [Coin2-devel]
-  ubuntu: [libcoin80-dev]
+  ubuntu:
+    xenial: [libcoin80-dev]
+    bionic: [libcoin80-dev]
+    focal: [libcoin-dev]
+    groovy: [libcoin-dev]
```
mabelzhang, line 2292: I think to be future-proof, it can just be
```suggestion
    '*': [libcoin-dev]
    xenial: [libcoin80-dev]
    bionic: [libcoin80-dev]
```

fspindle: Thanks, changes done as suggested.

tfoote (approved):

mabelzhang (approved):

fspindle:
Introducing this PR doesn't fix this build issue: [Build failed in Jenkins: Nbin_ufv8_uFv8__visp__ubuntu_focal_arm64__binary #8](https://build.ros.org/job/Nbin_ufv8_uFv8__visp__ubuntu_focal_arm64__binary/8/console)

I don't understand why. Changes in this PR where introduced such as `focal` should install `libcoin-dev`, while it continues to try to install `libcoin80-dev` that doesn't exist on focal. I suspect that the changes in rosdep are not taken into account. Have I missed something ?

tfoote:
The keys are resolved by bloom at release time. You need to rerun bloom to get the new resolution baked into the debian package. The debian package doesn't know about rosdep, it just has apt keys in the control file.

fspindle:
Thanks

PR #28817, opened 2021-03-24
Adding python3-seaborn for Ubuntu focal

tylernewnoise:
Please add the following dependency for python3-seaborn for Ubuntu Focal 20.04 to the rosdep database.

## Package name:

python3-seaborn

## Package Upstream Source:

https://github.com/mwaskom/seaborn

## Purpose of using this:

This dependency is being used in a www.dlr.de research project. 

Distro packaging links:

## Links to Distribution Package

  - https://packages.ubuntu.com/focal/python3-seaborn

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4270,7 +4270,6 @@ python-seaborn:
   gentoo: [dev-python/seaborn]
   ubuntu:
     '*': [python-seaborn]
-    focal_python3: [python3-seaborn]
     trusty:
       pip:
         packages: [seaborn]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4270,6 +4270,7 @@ python-seaborn:
   gentoo: [dev-python/seaborn]
   ubuntu:
     '*': [python-seaborn]
+    focal: [python3-seaborn]
     trusty:
       pip:
         packages: [seaborn]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4262,7 +4262,7 @@ python-scp:
   fedora: [python-scp]
   ubuntu: [python-scp]
 python-seaborn:
-  arch: [python2-seaborn]
+  arch: [python-seaborn]
   debian: [python-seaborn]
   fedora:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6994,8 +6994,7 @@ python3-seaborn:
   debian: [python3-seaborn]
   fedora: [python3-seaborn]
   gentoo: [dev-python/seaborn]
-  ubuntu:
-    '*': [python3-seaborn]
+  ubuntu: [python3-seaborn]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Have you tested and validated that this does what you want? This syntax has been deprecated for years and requires manual environment variable override arguments passed to rosdep.

The correct way to do this is to create a new `python3-seaborn` key in parallel with the existing one. It looks like the ubuntu package has been around for a long time so the equivalent deian package should be available. Gentoo and Arch can use the same package for the py2 and py3 rules. And I'd suggest looking if Fedora has a package for it now however many years later.

tylernewnoise:
Hm, you're right, for whatever reason it does not what I want, although I tested it before this PR, sry. I also wasn't aware that the syntax is deprecated, as I commited something similar to this one [here](https://github.com/ros/rosdistro/pull/28282) and I couldn't see any difference? :man_shrugging: 

Anyway, I can create the extra entry for the python3 package. Should this be closed here or can I make the changes in this PR?
Thank you.

tfoote:
You can update this branch and it will update the PR. Just comment back here so we get a notification when you push the updates.

tfoote:
Note that the difference is between `focal` and `focal_python3` you have to override your system identification to use that key.

tylernewnoise:
Added the new entry. Not sure if it makes sense to remove the arch and gentoo ones from ``python-seaborn`` as they are for Python3 only?

## Links to Distribution Package

  -  https://packages.debian.org/buster/python3-seaborn
  - https://packages.ubuntu.com/focal/python3-seaborn
  - Service unabailable, checked on  https://fedora.pkgs.org/32/fedora-x86_64/python3-seaborn-0.9.0-9.fc32.noarch.rpm.html instead
  - https://archlinux.org/packages/community/any/python-seaborn/
  - https://packages.gentoo.org/packages/dev-python/seaborn

tylernewnoise:
I'm not sure why the pipeline fails. Couldn't find whitespaces and it's alphabetically ordered - can some please help me out on how to fix it?

tfoote:
If you scroll up in the console output it says:

`  ERR: list out of alphabetical order line 4283.  'python-selectors2-pip' should come before 'python3-seaborn'`

The python3 rules will not be adjacent to the python rules.

> Added the new entry. Not sure if it makes sense to remove the arch and gentoo ones from python-seaborn as they are for Python3 only?

No you shouldn't remove gentoo and arch rules. THey have been using all the python keys under python3 for a long time. They don't differentiate python2 and python3, it goes with the system.

tylernewnoise:
Thx @tfoote for clearing things up. :+1:  :nerd_face:

wjwwood (approved):
lgtm

tfoote (approved):
Thanks for iterating. I did one small fixup to simplify the ubuntu rule to the minimum too. 

Ready for merge once CI for that fixup passes.

PR #28827, opened 2021-03-24
Move rmw_connextdds to ros2 organization

asorbini:
This PR updates the URL for `rmw_connextdds` now that the repository was transfered to the `ros2` GitHub organization.


asorbini:
Please let me know if I should have made this change with `bloom` instead of manually.

clalancette:
> Please let me know if I should have made this change with `bloom` instead of manually.

In this case, manually is fine.

PR #28830, opened 2021-03-25
added smbus2 python key

prat1kbhujbal:

Package name:
smbus2

Package Upstream Source:
`https://pypi.org/project/smbus2/`

Purpose of using this:
Our current robot stack is working on smbus2 for i2c communication to peripheral boards, earlier we faced problems with smbus and they still not resolved but using smbus2 worked for our stack. 

Please Add This Package to be indexed in the rosdistro.
Melodic

@mabelzhang @cottsay @hidmic

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4554,7 +4554,7 @@ python-smbus2-pip:
       packages: [smbus2]
   ubuntu:
     pip:
-      packages: [smbus2]  
+      packages: [smbus2]
 python-sortedcontainers-pip:
   ubuntu:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4542,6 +4542,19 @@ python-slycot-pip:
 python-smbus:
   debian: [python-smbus]
   ubuntu: [python-smbus]
+python-smbus2-pip:
+  debian:
+    pip:
+      packages: [smbus2]
+  fedora:
+    pip:
+      packages: [smbus2]
+  osx:
+    pip:
+      packages: [smbus2]
+  ubuntu:
+    pip:
+      packages: [smbus2]  
 python-sortedcontainers-pip:
   ubuntu:
     pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Please edit your description using the template that we provide for rosdep keys: https://github.com/ros/rosdistro/blob/master/.github/pull_request_template.md

It should have been presented to you when you opened the PR. 

And please justify why this should be used instead of the python-smbus key. The description for the project is that it's "A drop-in replacement for smbus-cffi/smbus-python in pure Python" as such it would seem that using the packaged version might make more sense.

Also from the [contributing guide](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md) assuming the above is correct it should have a `python3` prefix

Prat33k-dev:
> Please edit your description using the template that we provide for rosdep keys: https://github.com/ros/rosdistro/blob/master/.github/pull_request_template.md
> 
> It should have been presented to you when you opened the PR.
> 
> And please justify why this should be used instead of the python-smbus key. The description for the project is that it's "A drop-in replacement for smbus-cffi/smbus-python in pure Python" as such it would seem that using the packaged version might make more sense.
> 
> Also from the [contributing guide](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md) assuming the above is correct it should have a `python3` prefix

I have iterated the pr description according to the guidelines and provided justification as to why we require this package.

rosdep/python.yaml
```diff
@@ -4542,6 +4542,19 @@ python-slycot-pip:
 python-smbus:
   debian: [python-smbus]
   ubuntu: [python-smbus]
+python-smbus2-pip:
```
tfoote, line 4545:
```suggestion
python3-smbus2-pip:
```

pip rules should start with python3
Accepting this will then also require fixing the alphabetization too.

tfoote (changes requested):
Thanks for adding the justification. That makes sense and I don't see non-pip packaging.

One small syntax fix for the rule though.

tfoote (approved):

tfoote:
Thanks for the updates.

hidmic (approved):
LGTM!

PR #28854, opened 2021-03-26
Adding meshio to rosdep

FSund:

## Package name:

python3-meshio

## Package Upstream Source:

https://github.com/nschloe/meshio

## Purpose of using this:

We are using meshio to load some meshes into our MoveIt planning scene via Python.

Meshio can read and write most commonly used mesh formats for representing unstructured meshes, and smoothly converts between them.

Distro packaging links:

- Debian: https://packages.debian.org/buster/python3-meshio
- Ubuntu: https://packages.ubuntu.com/focal/python3-meshio
  - "Service Unavailable" (couldn't find any matches for meshio on [https://src.fedoraproject.org/](https://src.fedoraproject.org/) either)
- Arch: https://aur.archlinux.org/packages/python-meshio
  - not available
  - not available (see comment below)

I found something for macOS here, but not sure if I would go about adding it to rosdep: https://github.com/macports/macports-ports/blob/master/python/py-meshio/Portfile

It's also on pip: https://pypi.org/project/meshio/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6294,6 +6294,7 @@ python3-mechanize:
 python3-meshio:
   debian: [python3-meshio]
   ubuntu: [python3-meshio]
+  arch: [python-meshio]
 python3-mock:
   alpine: [py3-mock]
   debian: [python3-mock]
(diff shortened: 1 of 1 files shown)


wjwwood (approved):
lgtm

PR #28870, opened 2021-03-26
Make the python3-lark-parser key forwards compatible.

clalancette:
That is, on Focal and later we use the python3-lark key.

Here's the link to the package on Ubuntu: https://packages.ubuntu.com/search?keywords=python3-lark&searchon=names&suite=all&section=all .

Note that prior to Focal, we hosted our own version of lark parser from the bootstrap repository: http://repos.ros.org/repos/ros_bootstrap/pool/main/l/lark-parser/ .  That's why the key changed names.


wjwwood (approved):
lgtm

PR #28890, opened 2021-03-29
Remove rmw_connext_cpp and rosidl_typesupport_connext.

clalancette:
Rolling no longer needs these as we've switched to
rmw_connextdds.


hidmic (approved):
Looks reasonable to me.

PR #28920, opened 2021-03-31
franka_ros: 0.7.1-2 in 'noetic/distribution.yaml' [bloom]

fwalch:
Increasing version of package(s) in repository `franka_ros` to `0.7.1-2`:

- upstream repository: https://github.com/frankaemika/franka_ros.git
- release repository: https://github.com/frankaemika/franka_ros-release.git
- bloom version: `0.10.3`
- previous version for package: `0.7.1-1`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2053,14 +2053,16 @@ repositories:
       packages:
       - franka_control
       - franka_description
+      - franka_example_controllers
       - franka_gripper
       - franka_hw
       - franka_msgs
+      - franka_ros
       - franka_visualization
       tags:
         release: release/noetic/{package}/{version}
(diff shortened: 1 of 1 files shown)


tfoote:
This only has a debian increment change, but I see what appears to be upstream changes. Most of the time this will work but the correct thing to do is a patch release of the upstream content so that a new original tar ball is created, instead of overwriting the current one which has pitfalls relating to caching in the release pipeline.

fwalch:
The only change was that the `noetic.ignored` file was removed from the release repo. Previously, we ignored some packages because a dependency had not been published yet, but this was resolved recently (https://github.com/ros/rosdistro/pull/28851).

tfoote:
ok great, that's reasonable for a debian increment. Thanks for the clarification.

PR #28952, opened 2021-04-05
rosdep: OpenEmbedded updates

shr-project:
https://layers.openembedded.org/layerindex/recipe/555/
https://layers.openembedded.org/layerindex/recipe/102135/
https://layers.openembedded.org/layerindex/recipe/5626/
https://layers.openembedded.org/layerindex/recipe/23543/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6947,7 +6947,7 @@ python3-pycryptodome:
   fedora: [python3-pycryptodomex]
   gentoo: [dev-python/pycryptodome]
   nixos: [python3Packages.pycryptodomex]
-  openembedded: [python3-pycryptodome@meta-python]
+  openembedded: [python3-pycryptodomex@meta-python]
   osx:
     pip:
       packages: [pycryptodome]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -23,6 +23,7 @@ acl:
   fedora: [acl, libacl-devel]
   gentoo: [sys-apps/acl]
   nixos: [acl]
+  openembedded: [acl@openembedded-core]
   rhel: [acl, libacl-devel]
   ubuntu: [acl, libacl1-dev]
 acpi:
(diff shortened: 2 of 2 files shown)


clalancette:
What's the status with this one?  Should we review it for possible inclusion?  If so, can you add links to the description so we can look at where all of the package names are coming from for OpenEmbedded?  Thanks.

shr-project:
> What's the status with this one? Should we review it for possible inclusion? If so, can you add links to the description so we can look at where all of the package names are coming from for OpenEmbedded? Thanks.

It's ready for review and merge, the links are in individual commit messages which added them:
    https://layers.openembedded.org/layerindex/recipe/555/
    https://layers.openembedded.org/layerindex/recipe/102135/
    https://layers.openembedded.org/layerindex/recipe/5626/
    https://layers.openembedded.org/layerindex/recipe/23543/

clalancette (approved):

rosdep/python.yaml
```diff
@@ -6670,7 +6670,7 @@ python3-pycryptodome:
   debian: [python3-pycryptodome]
   fedora: [python3-pycryptodomex]
   gentoo: [dev-python/pycryptodome]
-  openembedded: [python3-pycryptodome@meta-python]
+  openembedded: [python3-pycryptodomex@meta-python]
```
jacobperron, line 6673: I can't find this package in meta-python: https://layers.openembedded.org/layerindex/branch/master/layer/meta-python

It seems to only be in meta-python2: https://layers.openembedded.org/layerindex/recipe/114445/

shr-project: That's because in gatesgarth release it was moved from meta-python to oe-core, see:

https://git.openembedded.org/meta-openembedded/commit/?id=a8f3c00d8d113b46a49584682e10435157d516ca
https://git.openembedded.org/openembedded-core/commit/?id=a96f815c53364b119b5743b8b7100eb5588d5cf5

So @meta-python is correct for LTS release dunfell and for newer releases @openembedded-core would be more correct, but luckily these @ suffixes aren't used for anything (dropped by superflore), so it doesn't matter.

ivanpauno (approved):
LGTM!

PR #28955, opened 2021-04-05
[rosdep] Added util-linux.

tonybaltovski:

## Package name:
util-linux

## Package Upstream Source:

https://github.com/karelzak/util-linux

## Purpose of using this:

Use for setpriv for robot_upstart.

Distro packaging links:

  - https://packages.debian.org/search?keywords=util-linux&searchon=names&suite=stable&section=all
   - https://packages.ubuntu.com/search?keywords=util-linux&searchon=names&suite=focal&section=all
  - https://src.fedoraproject.org/rpms/util-linux
- Arch: https://archlinux.org/packages/core/x86_64/util-linux/
  - IF AVAILABLE
  - https://packages.gentoo.org/packages/sys-apps/util-linux
  - https://formulae.brew.sh/formula/util-linux#default

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6578,6 +6578,13 @@ usbutils:
     homebrew:
       packages: [mikhailai/misc/usbutils]
   ubuntu: [usbutils]
+util-linux:
+  arch: [util-linux]
+  debian: [util-linux]
+  fedora: [util-linux]
+  gentoo: [util-linux]
+  macports: [util-linux]
+  ubuntu: [util-linux]
 uuid:
   arch: [util-linux]
   debian: [uuid-dev]


rosdep/base.yaml
```diff
@@ -6578,6 +6578,15 @@ usbutils:
     homebrew:
       packages: [mikhailai/misc/usbutils]
   ubuntu: [usbutils]
+util-linux:
+  arch: [util-linux]
+  debian:
+    buster: [util-linux]
```
tfoote, line 6584: Why is this only listed for buster? 

It appears to be available for all active distros: 
https://packages.debian.org/buster/util-linux

rosdep/base.yaml
```diff
@@ -6578,6 +6578,15 @@ usbutils:
...
+  fedora: [util-linux]
+  gentoo: [util-linux]
+  macports: [util-linux]
+  ubuntu:
+    focal: [util-linux]
```
tfoote, line 6589: Likewise this is available on all ubuntu distros: https://packages.ubuntu.com/focal/util-linux

tfoote (changes requested):
Please make the debian and ubuntu rules generic not version locked. 
Screenshot from 2021-04-06 12-07-17
Screenshot from 2021-04-06 12-06-03

rosdep/base.yaml
tonybaltovski: I totally missed those @tfoote, thanks!

tfoote (approved):
thanks for the fixup.

looks good to me.

hidmic (approved):
LGTM !

PR #28966, opened 2021-04-06
Update libcoin80-dev debian buster system dependency

fspindle:
Since debian buster, libcoin80-dev package was renamed in libcoin-dev

See: https://packages.debian.org/buster/libcoin-dev


tfoote (approved):
+1 matching the already fixed ubuntu rules

hidmic (approved):

PR #29017, opened 2021-04-06
[dep] rocker

130s:
## Package name:

rocker https://github.com/osrf/rocker

## Purpose of using this:

Convenient `docker` CLI wrapper IINM.

- Debian: buster http://packages.ros.org/ros/ubuntu/dists/buster/main/binary-amd64/Packages
- Ubuntu: xenial http://packages.ros.org/ros/ubuntu/dists/xenial/main/binary-amd64/Packages
- Fedora: https://apps.fedoraproject.org/packages/ --> None? Also `apps.fedoraproject.org/packages` leads me to useless webpage for searching a pkg.
- Arch: https://www.archlinux.org/packages/ --> None
- Gentoo: https://packages.gentoo.org/packages/ --> None. Also `https://packages.gentoo.org/packages/` not found.
- macOS: https://formulae.brew.sh/ --> None

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6965,6 +6965,9 @@ python3-retrying:
     pip:
       packages: [retrying]
   ubuntu: [python3-retrying]
+python3-rocker:
+  debian: [python3-rocker]
+  ubuntu: [python3-rocker]
 python3-rosdep:
   alpine: [rosdep]
   debian: [python3-rosdep]


rosdep/python.yaml
```diff
@@ -6965,6 +6965,13 @@ python3-retrying:
     pip:
       packages: [retrying]
   ubuntu: [python3-retrying]
+python3-rocker:
+  alpine: [rocker]
+  debian: [python3-rocker]
+  fedora: [python3-rocker]
+  gentoo: [dev-util/rocker]
```
tfoote, line 6972: I don't see this available on gentoo: https://packages.gentoo.org/packages/search?q=rocker

rosdep/python.yaml
```diff
@@ -6965,6 +6965,13 @@ python3-retrying:
     pip:
       packages: [retrying]
   ubuntu: [python3-retrying]
+python3-rocker:
+  alpine: [rocker]
+  debian: [python3-rocker]
+  fedora: [python3-rocker]
```
tfoote, line 6971: I don't think that we have RPMs currently for fedora or rhel yet either.

https://src.fedoraproject.org/

rosdep/python.yaml
```diff
@@ -6965,6 +6965,13 @@ python3-retrying:
     pip:
       packages: [retrying]
   ubuntu: [python3-retrying]
+python3-rocker:
+  alpine: [rocker]
```
tfoote, line 6969: I don't see this on alpine either: https://pkgs.alpinelinux.org/packages?name=rocker&branch=edge

tfoote (changes requested):
There's a bunch of packages listed here that don't appear to exist. 

One those alternative platforms if this is a dependency needed it could be made available via pip. However until someone needs them I'd suggest for now leaving it unresolved.

130s:
My bad, looks like I forgot to update the commit after doing the research for each distro and finalized the PR description. Updated now.

tfoote (approved):
thanks for the cleanup.

cottsay (approved):

PR #29062, opened 2021-04-09
adding python3-bidict as rosdep key

crthilakraj:

## Package name:

python3-bidict

## Package Upstream Source:

https://github.com/jab/bidict

## Purpose of using this:

The bidirectional mapping library for Python.

Distro packaging links:

  - https://packages.debian.org/sid/python3-bidict
   - https://packages.ubuntu.com/groovy/python3-bidict

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5601,6 +5601,10 @@ python3-backoff-pip:
   ubuntu:
     pip:
       packages: [backoff]
+python3-bidict:
+  debian: [python3-bidict]
+  fedora: [python3-bidict]
+  ubuntu: [python3-bidict]
 python3-bitarray:
   debian: [python3-bitarray]
   fedora: [python3-bitarray]


cottsay (approved):

rosdep/python.yaml
```diff
@@ -5601,6 +5601,10 @@ python3-backoff-pip:
   ubuntu:
     pip:
       packages: [backoff]
+python3-bidict:
+  debian: [python3-bidict]
```
clalancette, line 5605: Since this package is only available since bullseye, but we still support buster, we need to do something slightly more complicated here.  In particular:

```
debian:
  '*': [python3-bidict]
  buster: null
```

rosdep/python.yaml
```diff
@@ -5601,6 +5601,10 @@ python3-backoff-pip:
...
+  fedora: [python3-bidict]
+  ubuntu: [python3-bidict]
```
clalancette, line 5607: Same goes for Ubuntu.  This is only in Ubuntu since Focal, and we still support Xenial and Bionic:

```
ubuntu:
  '*': [python3-bidict]
  bionic: null
  xenial: null
```

clalancette (changes requested):

ivanpauno:
@crthilakraj friendly ping

hidmic:
@crthilakraj friendly ping

tfoote:
With no response for 4 weeks I'm going to close this. 

@crthilakraj please comment back if you update this or open a new PR with the new diff when you're ready to update it.

PR #29070, opened 2021-04-10
Add many Nix/NixOS packages.

lopsided98:
Adds rosdep keys for a large number of packages in nixpkgs. I use this with my [fork of superflore](https://github.com/lopsided98/superflore) to generate [nix-ros-overlay](https://github.com/lopsided98/nix-ros-overlay).

The following command can be used to verify that all the packages names are valid:
```
yq -r '[.[].nixos | select(.) | .[]] | unique | map("\(.).drvPath") | .[]' rosdep/*.yaml | xargs -n1 nix-instantiate --eval https://github.com/lopsided98/nix-ros-overlay/archive/refs/heads/master.tar.gz -A
```
There should be no errors like the following: 
```
error: attribute '<name>' in selection path '<name>' not found
```
There will be some errors due to packages that are currently broken in nixpkgs, or due to some packages no longer supporting Python 2. Including packages without Python 2 support is still useful for the case where I override the Python version in order to build melodic with Python 3.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -41,7 +41,8 @@ Distro packaging links:
   - IF AVAILABLE
 - Alpine: https://pkgs.alpinelinux.org/packages
   - IF AVAILABLE
-
+- NixOS/nixpkgs: https://search.nixos.org/packages
+  - OPTIONAL
 
 
 <!-- DOC_INDEX_TEMPLATE: add package to rosdistro for documentation indexing -->
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,6 +76,7 @@ Guidelines for rosdep rules
     * Ubuntu and Debian are top priority.
     * Fedora and Gentoo have packaging jobs and should be filled in if they are available.
     * OSX is also nice to have.
+    * NixOS is not required, but may be added if desired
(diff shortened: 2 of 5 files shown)


tfoote (commented):
Thanks for putting in the time to identify all these packages. It's fine to add them. However I don't necessarily want to add more burden/requests to contributors especially for a smaller platform. Could you add to the Contributing Guide and pull request template comments about where to look if people are feeling helpful but also note the rules for nixos are optional when reviewing. And also links to the package manager will help facilitate contributions. The easier it is to add them the more likely contributors are to take the extra effort.

lopsided98:
I added some documentation, let me know what you think.

I'm fine with mostly maintaining the `nixos` keys myself, as I have been doing for the last two years. I normally just keep an eye on the missing dependencies list in the superflore PRs and add new keys as necessary.

clalancette (approved):
I didn't go over all of the keys; there are way too many for that.  I'm going to assume that those work for the most part.

The changes to the documentation seems fine to me, so I'll go ahead and approve.

jacobperron (approved):
I spot checked the additions to `base.yaml` and `python.yaml`. LGTM

jacobperron:
I've resolved small conflict.

PR #29086, opened 2021-04-12
Add python3-flask-cors and python3-flask-socketio 

youliangtan:
## Package name:

`flask-cors` and `flask-socketio`

## Purpose of using this:

This is used by https://github.com/open-rmf/rmf_demos/ 's  `rmf_demos_panel` web server

### python3-flask-cors
- Debian: https://packages.debian.org/buster/python3-flask-cors
- Ubuntu: https://packages.ubuntu.com/groovy/python3-flask-cors

### flask-socketio
- Pip3: https://pypi.org/project/Flask-SocketIO/
- Note: the ubuntu package of `flask-socketio` is version 4.x . A latest version of 5.X is prefered as it not backward compatible.

Edit: After feedback, will switch back `flask-socketio` to deb and ubuntu package for consistency

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6012,6 +6012,16 @@ python3-flask:
   fedora: [python3-flask]
   gentoo: [dev-python/flask]
   ubuntu: [python3-flask]
+python3-flask-cors:
+  debian: [python3-flask-cors]
+  ubuntu: [python3-flask-cors]
+python3-flask-socketio:
+  debian:
+    pip:
+      packages: [flask-socketio]
+  ubuntu:
+    pip:
+      packages: [flask-socketio]
 python3-flask-restplus-pip:
   ubuntu:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6020,12 +6020,8 @@ python3-flask-restplus-pip:
     pip:
       packages: [flask-restplus]
 python3-flask-socketio:
-  debian:
-    pip:
-      packages: [flask-socketio]
-  ubuntu:
-    pip:
-      packages: [flask-socketio]
+  debian: [python3-flask-socketio]
+  ubuntu: [python3-flask-socketio]
 python3-future:
   debian: [python3-future]
   fedora: [python3-future]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6012,10 +6012,20 @@ python3-flask:
...
   ubuntu:
     pip:
       packages: [flask-restplus]
+python3-flask-socketio:
+  debian:
+    pip:
+      packages: [flask-socketio]
+  ubuntu:
+    pip:
+      packages: [flask-socketio]
```
tfoote, line 6028: This is not a parallel implelementation and will conflict with a system install. 5.0 is already packaged for hirsute: https://packages.ubuntu.com/source/hirsute/flask-socketio

rosdep/python.yaml
```diff
@@ -6012,10 +6012,20 @@ python3-flask:
...
+  debian: [python3-flask-cors]
+  ubuntu: [python3-flask-cors]
 python3-flask-restplus-pip:
   ubuntu:
     pip:
       packages: [flask-restplus]
+python3-flask-socketio:
+  debian:
+    pip:
+      packages: [flask-socketio]
```
tfoote, line 6025: The debian version has 5.0 already: https://packages.debian.org/bullseye/python3-flask-socketio

tfoote (changes requested):
It is expected that pip versions will be ahead of the apt package versions. However for consistency we [use the system packaged versions if they are available](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules) as that is what other users on the system expect if they are available and has been cross validated with other libraries.

rosdep/python.yaml
tanyouliang95: Make sense, fixed. https://github.com/ros/rosdistro/pull/29086/commits/c9f227ff7f785fdf657adcf9c5014939ce98b341

tfoote (approved):
thanks for the quick fixup. It's always frustrating waiting for newer versions to land in distros.

cottsay (approved):

PR #29103, opened 2021-04-12
Update Rolling rqt_msg branch to match release repo.

clalancette:


ivanpauno (approved):
I should also update the `track` file in the release repo.
I will `bloom-release --edit` it so we have the last patch in and the track file updated

ivanpauno:
See https://github.com/ros/rosdistro/pull/29105 (rolling) and https://github.com/ros/rosdistro/pull/29106 (foxy).

PR #29135, opened 2021-04-14
New version Nerian ROS node, including first version for ROS2

nerian-vision:
Unfortunately bloom refuses to generate automatic pull requests at the moment, so here's a manual submission. We released the first version of the node for our 3D cameras for ROS2. Also we updated the packages for ROS1.

## Package name:

nerian_stereo

## Package Upstream Source:

https://github.com/nerian-vision/nerian_stereo_ros2.git

## Purpose of using this:

This is the ROS2 version of the node for Nerian's 3D cameras
Distro packaging links:


cottsay (approved):

tfoote:
In the future please submit these as separate pull requests for each distro. Different distros have different release timelines and sync freeze times. As well as a potential necessary rollback is much harder to track if it's cross coupled. You got lucky, but if you were off by one day Foxy sync freeze would be blocking this merge.

Related the Kinetic update is highly urgent because we are preparing for the final kinetic sync. All releases should be prepared by April 20th and we will have a long soak time until April 26th. Please make sure to test from the [testing repository](http://wiki.ros.org/TestingRepository) to make sure that there's no regressions after this is released. You can get more info from [this thread on ROS Discourse](https://discourse.ros.org/t/preparing-for-final-kinetic-sync/19805)

ros-discourse:

https://discourse.ros.org/t/releasing-a-package-for-ros2/19965/3

PR #29187, opened 2021-04-16
add python-albumentations-pip

fmessmer:

## Package name:

`python-albumentations-pip`

## Package Upstream Source:
https://pypi.org/project/albumentations/
(pip only package - no release found, see below)

## Purpose of using this:

https://pypi.org/project/albumentations/#description

Distro packaging links:

  - NOT_FOUND
   - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -434,19 +434,6 @@ python-adafruit-bno055-pip:
   ubuntu:
     pip:
       packages: [adafruit_bno055]
-python-albumentations-pip:
-  debian:
-    pip:
-      packages: [albumentations]
-  fedora:
-    pip:
-      packages: [albumentations]
-  osx:
-    pip:
-      packages: [albumentations]
-  ubuntu:
-    pip:
-      packages: [albumentations]
 python-alembic:
   debian:
     buster: [python-alembic]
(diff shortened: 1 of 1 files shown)


fmessmer:
friendly ping :wink: 
anything wrong with this PR? anything blocking the merge of this PR?

fmessmer:
@clalancette 
are `rosdep` prs not merged anymore/atm?

rosdep/python.yaml
```diff
@@ -434,6 +434,19 @@ python-adafruit-bno055-pip:
   ubuntu:
     pip:
       packages: [adafruit_bno055]
+python-albumentations-pip:
```
clalancette, line 437:
```suggestion
python3-albumentations-pip:
```

clalancette (changes requested):

rosdep/python.yaml
fmessmer: is this due to `kinetic` no longer being supported? so any `rosdep` key in the `python.yaml` is now being considered `python3`?

clalancette: > so any `rosdep` key in the `python.yaml` is now being considered `python3`?

It's a compromise amongst a number of bad options.  Essentially we always prefer the distribution/system key, and those are always named `python3-<something>`.  In order to make the pip key live "near" where the distribution key would live, we are naming all pip keys `python3-<something>-pip`.  This also represents the fact that python2 is no longer really supported, and will completely go out of support once the current host of LTS distributions goes EOL (probably ~2023, when Bionic goes EOL).

clalancette (approved):

jacobperron (approved):

PR #29236, opened 2021-04-21
Set Galactic's status to prerelease.

nuclearsandwich:
Galactic is currently a prerelease distribution and is not yet active.

`prerelease` is one of the semantic statuses described in https://www.ros.org/reps/rep-0153.html#index-file


cottsay (approved):
Didn't even know that was an option!

PR #29257, opened 2021-04-22
add python3 dependency

christian-rauch:

## Package name:

`python3-dev`

## Package Upstream Source:

https://www.python.org/download/releases/

## Purpose of using this:

Use key `python3-dev` besides `python-dev` for Python 3 dependencies.

Distro packaging links:

- Debian: https://packages.debian.org/buster/python3-dev
- Ubuntu: https://packages.ubuntu.com/focal/python3-dev
- Fedora: https://src.fedoraproject.org/rpms/python3

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5441,6 +5441,10 @@ python-dev:
   ubuntu:
     '*': [python-dev]
     'focal': [python2-dev]
+python3-dev:
+  debian: [python3-dev]
+  fedora: [python3-devel]
+  ubuntu: [python3-dev]
 qhull-bin:
   arch: [qhull]
   debian: [qhull-bin]


christian-rauch:
There is a redundancy here: Some python distribution packages are defined in [`python.yaml`](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml) and some in the [`base.yaml`](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml). I previously assumed that the former is for python packages (i.e. not python itself) that are either fetched via `pip` or the distribution, and the latter is for distribution specific package management, e.g. `apt`, `dnf`, etc. But this does not seem to be the case here.

The `python-dev` key, which points to package `python2-dev` (or `python-dev`) is defined in the `base.yaml`. But the `python.yaml` also defines the key `python` which points to the same packages, but only for a subset of Ubuntu distributions. Further, in `python.yaml` the "python2" key is called `python` (without "dev"), but the "Python 3" key is called `python3-dev`.

That's a bit confusing because I don't think you want to change the key in the package.xml depending on the distribution. How is this supposed to work?

clalancette:
> That's a bit confusing because I don't think you want to change the key in the package.xml depending on the distribution. How is this supposed to work?

I don't understand this question.  While I agree that the current situation is a bit messy, that's because a) the rules have evolved over time, and b) we tend not to remove old keys when the rules change, to not break people.

But that being said, the fact that it is messy doesn't mean the key depends on the distribution.  It just depends on what you want to depend on.  And further, for this particular key, it is already defined in python.yaml here: https://github.com/ros/rosdistro/blob/030129cf5e0bcc9ff15f18ca3f66440c0ac583bc/rosdep/python.yaml#L5870-L5876 .  At the end of the day, rosdep doesn't care where the dependency comes from, just that it exists somewhere.

So the question is: what about the current situation isn't working for you?

christian-rauch:
> But that being said, the fact that it is messy doesn't mean the key depends on the distribution.

How a key is resolved depends on the distribution. E.g. key `python` has no package for Ubuntu focal, neither in [`python.yaml`](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml) nor in [`base.yaml`](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml). This key only works for older Ubuntu distributions. If I want package `python3-dev` for focal, I have to use key `python3` instead. Therefore the key that has to be used is distribution dependant.

clalancette:
> How a key is resolved depends on the distribution. E.g. key `python` has no package for Ubuntu focal, neither in [`python.yaml`](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml) nor in [`base.yaml`](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml). This key only works for older Ubuntu distributions. 

Correct, that is by design.  There is no Python 2 in Focal, so there is no corresponding version of the `python` key.

> If I want package `python3-dev` for focal, I have to use key `python3` instead. Therefore the key that has to be used is distribution dependant.

That's right, to get python3, you need to use `python3-dev` key.  That is also by design.

christian-rauch:
> Correct, that is by design. There is no Python 2 in Focal

This forces focal users to update their dep keys for all packages for no reason. I cannot follow why this is by design when all of ROS depends on Python 3 anyway and there is no way to use Python 2 with ROS. The `python` key should just stay a python dependency that resolves to version 2 or 3 depending on the Ubuntu distribution. This would only make sense if python version 2 and 3 would be supported in parallel and a package maintainer has to decide which version to use - but this is not the case for any of the packages. Or am I missing something here?

Is there a concrete application where it makes sense to explicitly define the keys like this? What would be the drawback of resolving the `python` key to python2 on Ubuntu bionic and to python3 on Ubuntu focal?

clalancette:
> This forces focal users to update their dep keys for all packages for no reason. I cannot follow why this is by design when all of ROS depends on Python 3 anyway and there is no way to use Python 2 with ROS.

But that's not the case.  The rosdep database is shared by ROS 1 and ROS 2, and in ROS 1, both Kinetic and Melodic still use Python 2.

christian-rauch:
> The rosdep database is shared by ROS 1 and ROS 2, and in ROS 1, both Kinetic and Melodic still use Python 2.

Not sure what I am missing here. These ROS 1 versions (Kinetic and Melodic) are not supported on Ubuntu focal. So there is no possibility that python versions and ROS versions will clash. Kinetic and Melodic (hence Ubuntu melodic) support Python 2 only, Noetic (hence Ubuntu focal) support Python 3 only.

clalancette:
> Not sure what I am missing here. These ROS 1 versions (Kinetic and Melodic) are not supported on Ubuntu focal. So there is no possibility that python versions and ROS versions will clash. Kinetic and Melodic (hence Ubuntu melodic) support Python 2 only, Noetic (hence Ubuntu focal) support Python 3 only.

Correct.  At this point, it comes down to a philosophical debate on what the rosdep key represents.  We've said that the keys represent resources.  In this case, the "python" key, to me, represents the "Python 2" resource, while the "python3-dev" key represents the "Python 3" resource.  We could conflate the two, but they are not interchangeable.  Making the same key represent two different resources seems confusing to me.

I will also point out that the keys have been this way for several years now; this is not new at this point.

Overall I'm not in favor of that change, but others can weigh in and add their opinion.

ivanpauno:
> Not sure what I am missing here. These ROS 1 versions (Kinetic and Melodic) are not supported on Ubuntu focal. So there is no possibility that python versions and ROS versions will clash. Kinetic and Melodic (hence Ubuntu melodic) support Python 2 only, Noetic (hence Ubuntu focal) support Python 3 only.

ROS 2 dashing is supported in Bionic, same as ROS melodic.
That would've been a problem, because the former would want `python-dev` to be python3 and the later python2.

The decision of explicitly using a `python3` prefix for all python packages is documented in the [contributing guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-3).

---

The contributed rule is a duplicate of an [already existing rule](https://github.com/ros/rosdistro/blob/2cb117c41e1124db42f408c48ef0166f28e9309e/rosdep/python.yaml#L6060), so I think we can close this one.
Thanks for the contribution @christian-rauch !

PR #29279, opened 2021-04-24
Update dartsim source version

mxgrey:
Since dartsim's `release-6.3` branch was retired, we should switch rosdistro to using the `v6.3.1` tag instead, which gives a snapshot of where the `release-6.3` branch left off.

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -2399,7 +2399,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/dartsim/dart.git
-      version: release-6.3
+      version: v6.3.1
   dataspeed_can:
     doc:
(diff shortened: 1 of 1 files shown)


mxgrey:
I see from this

```
AssertionError: ["In file '''/home/runner/work/rosdistro/rosdistro/kinetic/distribution.yaml''': Could not validate source entry for repo dartsim with error [[[Could not validate repository with url https://github.com/dartsim/dart.git and version v6.3.1 from entry at line 2398. Error reason: Tags are not valid, but a tag v6.3.1 was found. Re: https://github.com/ros/rosdistro/pull/20286]]]"] is not false
```

that tags are not allowed. I'll fix this issue by just reopening the original `release-6.3` branch in dartsim.

PR #29316, opened 2021-04-27
Fix libav rosdep on Ubuntu

peci1:
There is no package `libav-tools` in any other distro than Xenial: https://packages.ubuntu.com/xenial/libav-tools . All other releases have only `ffmpeg`, which should do the same job.

I fixed the rosdep key to install `libav-tools` on Xenial and `ffmpeg` on all other Ubuntus.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1930,7 +1930,7 @@ libav:
   fedora: []
   gentoo: [virtual/ffmpeg]
   ubuntu:
-    '*': [ffmpeg]
+    '*': null
     xenial: [libav-tools]
 libavahi-client-dev:
   arch: [avahi]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1929,7 +1929,9 @@ libav:
   debian: [libav-tools]
   fedora: []
   gentoo: [virtual/ffmpeg]
-  ubuntu: [libav-tools]
+  ubuntu:
+    '*': [ffmpeg]
+    xenial: [libav-tools]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -1929,7 +1929,9 @@ libav:
   debian: [libav-tools]
   fedora: []
   gentoo: [virtual/ffmpeg]
-  ubuntu: [libav-tools]
+  ubuntu:
+    '*': [ffmpeg]
+    xenial: [libav-tools]
```
clalancette, line 1934: I understand where you are coming from, but as far as I can tell `ffmpeg` and `libav-tools` don't provide the same resources.  That is, on Xenial it looks like `libav-tools` provides the binaries `avconv`, `avprobe`, and `avplay`, while the `ffmpeg` package on newer distributions provides the binaries `ffmpeg`, `ffplay`, and `ffprobe`.  While those are the same in spirit, anything depending on being able to call `avplay` is going to fail.

What I'd suggest here is nulling out this entry for anything newer than Xenial.  That way packages will have to switch their package.xml to the `ffmpeg` key and update their code at the same time.

```suggestion
  ubuntu:
    '*': null
    xenial: [libav-tools]
```

clalancette (changes requested):

rosdep/base.yaml
peci1: You're right, null entry is better describing the reality.

clalancette (approved):

jacobperron (approved):

PR #29318, opened 2021-04-27
Update python.yaml added python-is-python3

belorenz:

## Package name:
python-is-python3

## Package Upstream Source:

apt?

## Purpose of using this:

In Ubuntu, all python packages use explicit python3 or python2
interpreter and do not use unversioned /usr/bin/python at all. Some
third-party code is now predominantly python3 based, yet may use
/usr/bin/python.
This is a convenience package which ships a symlink to point
the /usr/bin/python interpreter at the current default python3. It may
improve compatibility with other modern systems, whilst breaking some
obsolete or 3rd-party software.

Distro packaging links:
https://ubuntu.pkgs.org/20.04/ubuntu-main-amd64/python-is-python3_3.8.2-4_all.deb.html

- Debian: -
- Ubuntu: https://packages.ubuntu.com/focal/python-is-python3

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2240,7 +2240,10 @@ python-ipdb:
   gentoo: [dev-python/ipdb]
   ubuntu: [python-ipdb]
 python-is-python3:
-  ubuntu: [python-is-python3]
+  ubuntu: 
+    '*': [python-is-python3]
+    bionic: null
+    xenial: null
 python-itsdangerous:
   debian:
     buster: [python-itsdangerous]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2240,7 +2240,7 @@ python-ipdb:
   gentoo: [dev-python/ipdb]
   ubuntu: [python-ipdb]
 python-is-python3:
-  ubuntu: 
+  ubuntu:
     '*': [python-is-python3]
     bionic: null
     xenial: null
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2239,6 +2239,8 @@ python-ipdb:
   debian: [python-ipdb]
   gentoo: [dev-python/ipdb]
   ubuntu: [python-ipdb]
+python-is-python3:
+  ubuntu: [python-is-python3]
```
clalancette, line 2243: Since this is only available since focal, and we still support older distributions, we would need this to be:

```suggestion
  ubuntu: 
    '*': [python-is-python3]
    bionic: null
    xenial: null
```

clalancette (changes requested):
One change inline.  Also, does this package exist for Debian?

belorenz:
> One change inline. Also, does this package exist for Debian?

Was just able to test it on Debian 10.9. It is not available there.
After some research I think it is not available in general for Debian.

clalancette (approved):

rosdep/python.yaml
```diff
@@ -2239,6 +2239,11 @@ python-ipdb:
   debian: [python-ipdb]
   gentoo: [dev-python/ipdb]
   ubuntu: [python-ipdb]
+python-is-python3:
+  ubuntu:
```
clalancette, line 2243: There is trailing whitespace on this line, which is why CI is unhappy.

belorenz: Removed the whitespace. Thanks for the hint!

jacobperron (approved):

PR #29356, opened 2021-04-28
add pip package environs in 'rosdep/python.yaml'

yanrayw:

## Package name:

pip packages: `environs`

## Package Upstream Source:

`environs`: https://pypi.org/project/environs/

## Purpose of using this:

For some [autoware.auto](https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto) users, if they'd like to use [lgsvl-pythonAPI](https://github.com/lgsvl/PythonAPI) to control vehicle. There are some required packages to be installed first. Pip package `environs` is one of them. To simplify dependency installation process, adding them in rosdep would help a lot. Typically, only Ubuntu and debian system are needed. So I only add them in this pull request.

This dependency is being used for this reason. This is why I think it's valuable to be added to the rosdep database. If there are any questions, please let me know. Thanks a lot!

[rosdep/python.yaml](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml)

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5951,6 +5951,13 @@ python3-empy:
   openembedded: [python3-empy@meta-ros-common]
   rhel: ['python%{python3_pkgversion}-empy']
   ubuntu: [python3-empy]
+python3-environs-pip:
+  debian:
+    pip:
+      packages: [environs]
+  ubuntu:
+    pip:
+      packages: [environs]
 python3-evdev:
   arch: [python-evdev]
   debian:


YanrayWang:
Hi, @clalancette , I created a new pull request to describe in more detail of my pull request. But it looks like it requires an approval from maintainer to run two checks. Please have a look. Thanks a lot!

rosdep/python.yaml
```diff
@@ -5329,6 +5329,9 @@ python-webrtcvad-pip:
     pip:
       packages: [webrtcvad]
 python-websocket:
+  arch:
+    pip:
+      packages: [websocket-client]
```
clalancette, line 5334: So we already have the [python3-websocket](https://github.com/ros/rosdistro/blob/67cea41a94e36ab0b43c85cddf15a12a7681bb67/rosdep/python.yaml#L7456-L7462) key.  Since we generally prefer the system packages, I think it would be much better to add the arch entry (`python-websocket-client`) there.

rosdep/python.yaml
```diff
@@ -5938,6 +5944,13 @@ python3-empy:
   openembedded: [python3-empy@meta-ros-common]
   rhel: ['python%{python3_pkgversion}-empy']
   ubuntu: [python3-empy]
+python3-environs-pip:
+  arch:
+    pip:
+      packages: [environs]
+  ubuntu:
+    pip:
+      packages: [environs]
```
clalancette, line 5953: Please also add at least `debian` here.

clalancette (changes requested):

rosdep/python.yaml
YanrayWang: Thanks, added.

YanrayWang:
Hi, @clalancette . I think `websocket-client` has been satisfied by `python3-websocket`. So I don't have to add them anymore. In addition, `ubuntu` is enough for me, no `arch` needed. Thus the only one needed to be installed is `envirions` of pip package. Thanks a lot!

clalancette (approved):

jacobperron (approved):

PR #29376, opened 2021-04-30
genpyi: 0.2.0-1 in 'melodic/distribution.yaml' [bloom]

bonprosoft:
Increasing version of package(s) in repository `genpyi` to `0.2.0-1`:

- upstream repository: https://github.com/rospypi/genpyi.git
- release repository: https://github.com/rospypi/genpyi-release.git
- bloom version: `0.10.7`

## genpyi

```
* Generate py.typed in catkin built packages (#30 <https://github.com/rospypi/genpy_stubgen/issues/30>)
* Make package catkin compatible (#29 <https://github.com/rospypi/genpy_stubgen/issues/29>)
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3530,6 +3530,16 @@ repositories:
       url: https://github.com/ros/genpy.git
       version: kinetic-devel
     status: maintained
+  genpyi:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/rospypi/genpyi-release.git
+      version: 0.2.0-1
+    source:
+      type: git
+      url: https://github.com/rospypi/genpyi.git
+      version: master
   geographic_info:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -3530,6 +3530,16 @@ repositories:
...
     status: maintained
+  genpyi:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/rospypi/genpyi-release.git
+      version: 0.2.0-1
+    source:
+      type: git
+      url: https://github.com/rospypi/genpyi.git
```
clalancette, line 3541: I have to say that I am not a fan of the name of this package.  It could very easily be confused with `genpy`, which is a core ROS package.

What does this package do?

bonprosoft: This package generates `.pyi` files for ROS message/service files.
`.pyi` is [a Python stub file](https://www.python.org/dev/peps/pep-0484/#stub-files) and mainly used for static type checking tool, like mypy.
Since this package depends `genmsg` and provides the cmake file to extend its supported languages, I followed the naming rule of other packages, like genpy, gencpp.
https://github.com/rospypi/genpyi

But it might be confusing and I don't want to cause inconvenience to ROS users, please let me change this package name if needed. How about `genpystub`? Does it sound better?

bonprosoft: I would like to update the package name to `genmypy` in order to avoid confusion with `genpy`.
But before that, It would be really appreciated if I could have your feedback on that name. Is the package name `genmypy` acceptable for ros-distro?

bonprosoft:
As I have opened PR #29409, let me close this PR instead.
I understand that you are busy. I'm sorry for asking you to confirm the name many times...

PR #29381, opened 2021-05-02
Add python3-pyside2.qtuitools

lFatality:

## Package name:

`python3-pyside2.qtuitools`

## Package Upstream Source:

https://doc.qt.io/qtforpython-5/PySide2/QtUiTools/index.html
https://doc.qt.io/qt-5/qtuitools-index.html

## Purpose of using this:

Allows loading UI files to generate a graphical interface from QtDesigner forms at run-time.

Distro packaging links:

  - REQUIRED
  - https://packages.debian.org/buster/python3-pyside2.qtuitools
   - REQUIRED
   - https://packages.ubuntu.com/focal/python/python3-pyside2.qtuitools
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6863,6 +6863,9 @@ python3-pyside2.qtopengl:
     '*': [python3-pyside2.qtopengl]
     bionic: null
     xenial: null
+python3-pyside2.qtuitools:
+  debian: [python3-pyside2.qtuitools]
+  ubuntu: [python3-pyside2.qtuitools]
 python3-pystemd:
   debian:
     bullseye: [python3-pystemd]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6865,7 +6865,10 @@ python3-pyside2.qtopengl:
     xenial: null
 python3-pyside2.qtuitools:
   debian: [python3-pyside2.qtuitools]
-  ubuntu: [python3-pyside2.qtuitools]
+  ubuntu:
+    '*': [python3-pyside2.qtuitools]
+    bionic: null
+    xenial: null
 python3-pystemd:
   debian:
     bullseye: [python3-pystemd]


rosdep/python.yaml
```diff
@@ -6863,6 +6863,9 @@ python3-pyside2.qtopengl:
     '*': [python3-pyside2.qtopengl]
     bionic: null
     xenial: null
+python3-pyside2.qtuitools:
+  debian: [python3-pyside2.qtuitools]
+  ubuntu: [python3-pyside2.qtuitools]
```
clalancette, line 6868: Since this package is only available since `focal`, we need to change this entry a bit.  It should look like:

```
ubuntu:
  '*': [python3-pyside2.qtuitools]
  bionic: null
  xenial: null
```

clalancette (changes requested):

rosdep/python.yaml
lFatality: I've adjusted the availability in Ubuntu as you suggested.

clalancette: Thanks for iterating, looks good.

clalancette (approved):

rosdep/python.yaml
```diff
@@ -6863,6 +6863,12 @@ python3-pyside2.qtopengl:
     '*': [python3-pyside2.qtopengl]
     bionic: null
     xenial: null
+python3-pyside2.qtuitools:
+  debian: [python3-pyside2.qtuitools]
```
jacobperron, line 6867: @clalancette Can you confirm?
I think we should do something similar here, since it is not available in Debian Stretch or Jessie.
```suggestion
  debian:
    '*':  [python3-pyside2.qtuitools]
    stretch: null
    jessie: null
```

clalancette: Both Debian Stretch and Jessie are out-of-support (see https://wiki.debian.org/LTS), so we don't need to add new entries for them.

jacobperron: I see. But Melodic is not EOL and it lists recommended support for Stretch: https://ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023

clalancette: > I see. But Melodic is not EOL and it lists recommended support for Stretch: https://ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023

Ah, yeah, I can see where the confusion comes from.  I ended up disabling the Stretch jobs for Melodic back in December: https://discourse.ros.org/t/preparing-for-melodic-sync-2020-12-01/17681 .  I did not end up updating REP-3 along with that; maybe I should.

> If I understand the debian release cycle, Stretch is still supported by the LTS team.

Historically, we've only supported Debian (and Ubuntu, for that matter) until the end of the "normal" release cycle.  Typically the LTS beyond that requires setting up additional repositories and/or paying for support.  So from that perspective, Stretch is EOL.  But I'm not sure if this is documented anywhere.

jacobperron (approved):

lFatality:
Thanks for merging and the quick response time, you are awesome! :)
Cheers

PR #29422, opened 2021-05-05
mavlink: 2021.5.5-1 in 'galactic/distribution.yaml' [bloom]

vooon:
Increasing version of package(s) in repository `mavlink` to `2021.5.5-1`:

- bloom version: `0.10.7`

### DIFF
--- a/galactic/distribution.yaml
+++ b/galactic/distribution.yaml
@@ -1215,6 +1215,21 @@ repositories:
       url: https://github.com/swri-robotics/marti_messages.git
       version: dashing-devel
     status: developed
+  mavlink:
+    doc:
+      type: git
+      url: https://github.com/mavlink/mavlink-gbp-release.git
+      version: release/galactic/mavlink
+    release:
+      tags:
+        release: release/galactic/{package}/{version}
+      url: https://github.com/mavlink/mavlink-gbp-release.git
+      version: 2021.5.5-1
+    source:
+      type: git
+      url: https://github.com/mavlink/mavlink-gbp-release.git
+      version: release/galactic/mavlink
+    status: developed
   message_filters:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


galactic/distribution.yaml
```diff
@@ -1215,6 +1215,21 @@ repositories:
       url: https://github.com/swri-robotics/marti_messages.git
       version: dashing-devel
     status: developed
+  mavlink:
+    doc:
+      type: git
+      url: https://github.com/mavlink/mavlink-gbp-release.git
```
clalancette, line 1221: I just realized that this is pretty odd; we usually have this point to the *source* repository, not the release repository.

It actually ends up working because that particular branch of the release repository has a clone of the source, but that is not the place you want to point people to make changes.  So I'd suggest changing this to the source repository.

(all of this goes for the source entry down below as well, and also goes for the release in Foxy as well)

clalancette (changes requested):

galactic/distribution.yaml
vooon: Upstream is not ROS package, so it do not have package.xml.
Because of that ROS1 tools cannot build docs or properly add to source index.
Is it changed for ROS2?

vooon: Let's try to use upstream for source and docs.

hidmic (approved):
LGTM! @clalancette ?

On a related note, @vooon is it documented anywhere that the `mavlink` package sources are patched for ROS release? May avoid some confusion.

vooon:
@hidmic well, probably only release repo has some notes and actual patches on master branch: https://github.com/mavlink/mavlink-gbp-release/tree/master/patch

galactic/distribution.yaml
clalancette: Oh, I see.  I'm sorry, I didn't realize that's how it was setup.

In that case, I think what you had before was better.  We'll run into the same trouble with ROS 2 if there is no package.xml, as we have no way to figure out where a package begins.

hidmic:
Thanks for the revert @vooon. I didn't get to do it yesterday afternoon.

Merging now!

PR #29434, opened 2021-05-06
azure-iot-sdk-c: 1.7.0-3 in 'noetic/distribution.yaml' [bloom]

Timple:
Increasing version of package(s) in repository `azure-iot-sdk-c` to `1.7.0-3`:

- upstream repository: https://github.com/Azure/azure-iot-sdk-c.git
- release repository: https://github.com/nobleo/azure-iot-sdk-c-release.git
- bloom version: `0.10.7`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -342,10 +342,10 @@ repositories:
       url: https://github.com/nobleo/azure-iot-sdk-c-release.git
       version: 1.7.0-3
     source:
+      test_commits: false
       type: git
       url: https://github.com/Azure/azure-iot-sdk-c.git
       version: master
-      test_commits: false
     status: maintained
   backward_ros:
     doc:
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -335,6 +335,21 @@ repositories:
       url: https://github.com/astuff/avt_vimba_camera.git
       version: master
     status: maintained
+  azure-iot-sdk-c:
+    doc:
+      type: git
+      url: https://github.com/Azure/azure-iot-sdk-c.git
+      version: master
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/nobleo/azure-iot-sdk-c-release.git
+      version: 1.7.0-3
+    source:
+      type: git
+      url: https://github.com/Azure/azure-iot-sdk-c.git
+      version: master
+    status: maintained
   backward_ros:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -336,10 +336,6 @@ repositories:
       version: master
     status: maintained
   azure-iot-sdk-c:
-    doc:
-      type: git
-      url: https://github.com/Azure/azure-iot-sdk-c.git
-      version: master
     release:
       tags:
         release: release/noetic/{package}/{version}
@@ -349,6 +345,7 @@ repositories:
       type: git
       url: https://github.com/Azure/azure-iot-sdk-c.git
       version: master
+      test_commits: false
     status: maintained
   backward_ros:
     doc:


nuclearsandwich:
:heavy_check_mark: Package has a LICENSE file and package.xml contains a matching license tag (MIT)
:heavy_check_mark: Source repository is publicly accessible

@Timple because the source repository does not have a ROS package.xml the source and doc jobs for this package will fail to run.

One solution would be to remove the doc entry (so that the ROS build system does not try to build documentation and add `test_commits: false` to the source entry so that our devel jobs will not try to run on pushes to the source repository.

Timple:
For my information, there is nothing I can check locally first for these kind of releases right?

nuclearsandwich:
> For my information, there is nothing I can check locally first for these kind of releases right?

"Nothing" isn't strictly correct but in order to run the binarydeb jobs locally you need to release the package into a rosdistro, if not the official one. So it is possible but it's not common or straightforward. Which is why we don't expect packages to always build cleanly the first time they're released.

Timple:
Would it be possible to give me a pointer on what could have gone wrong? Build output is [here](https://build.ros.org/job/Nbin_uF64__azure-iot-sdk-c__ubuntu_focal_amd64__binary/1/console).
But I don't see any concrete error about a missing dependency or a compilation failure.

The project compiles fine with that specific [package.xml](https://github.com/nobleo/azure-iot-sdk-c-release/blob/master/noetic/package.xml) in a catkin workspace.

nuclearsandwich:
> Would it be possible to give me a pointer on what could have gone wrong? Build output is [here](https://build.ros.org/job/Nbin_uF64__azure-iot-sdk-c__ubuntu_focal_amd64__binary/1/console).
> But I don't see any concrete error about a missing dependency or a compilation failure.

Looking at the build logs I see this at the root of the CMake error message
```
00:01:08.507 CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
00:01:08.521   Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
00:01:08.521   system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY
00:01:08.521   OPENSSL_INCLUDE_DIR)
00:01:08.521 Call Stack (most recent call first):
00:01:08.521   /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
00:01:08.521   /usr/share/cmake-3.16/Modules/FindOpenSSL.cmake:447 (find_package_handle_standard_args)
00:01:08.521   c-utility/CMakeLists.txt:116 (find_package)
00:01:08.521 
00:01:08.521 
00:01:08.521 -- Configuring incomplete, errors occurred!
```

Based on the above and your linked package.xml it looks like you have an undeclared dependency on `libssl-dev`. The curl dependency is transitively getting you the libraries but not the development headers. The reason it works in a catkin workspace is that even catkin_make_isolated does not isolate system dependencies so if libssl-dev is present nothing complains about the missing dependency. These are exactly the sort of growing pains we expect from new packages so you're in good company.

Timple:
Thanks, looked right over that!

Tried to recreate the issue with a docker but even that failed:
```Dockerfile
FROM ros:noetic
RUN apt-get update -qq && apt-get install -qqy git wget python3-catkin-tools python3-rosdep python3-osrf-pycommon
RUN mkdir -p ws/src
WORKDIR ws
RUN catkin init
RUN git clone https://github.com/Azure/azure-iot-sdk-c.git src/azure-iot-sdk-c
RUN git -C src/azure-iot-sdk-c submodule update --init
RUN wget https://raw.githubusercontent.com/nobleo/azure-iot-sdk-c-release/master/noetic/package.xml -O src/azure-iot-sdk-c/package.xml
RUN sed -i 's/:{version}/1.7.0/g' src/azure-iot-sdk-c/package.xml
RUN cat src/azure-iot-sdk-c/package.xml
RUN rosdep install -y --from-paths .
RUN /bin/bash -c 'source "/opt/ros/$ROS_DISTRO/setup.bash" && catkin build'
```

So libssl-dev is probably already a dependency of ROS somewhere. I'll add it to the package.xml!

Timple:
https://github.com/ros/rosdistro/pull/29464

nuclearsandwich:
> Tried to recreate the issue with a docker but even that failed:

You can use the `generate_release_script.py` tool from `ros_buildfarm` to create a script that will build and run the package creation jobs in docker containers on your local machine. It's quite difficult to inject changes into those containers but you can at least see the failure locally.

nuclearsandwich:
> This mail was not received from my side.

I haven't had a chance to investigate this but thanks for the report all the same.

PR #29446, opened 2021-05-06
Adds the osrf-pycommon python rule

rickstaa:

## Package name:

[osrf_pycommon](https://osrf-pycommon.readthedocs.io/en/latest/)

## Package Upstream Source:

[osrf_pycommon](https://github.com/osrf/osrf_pycommon)

## Purpose of using this:

This dependency is needed for the catkin-tools package due to a bug that is present (see this [ros answer](https://answers.ros.org/question/353113/catkin-build-in-ubuntu-2004-noetic/) and [this issue](https://github.com/catkin/catkin_tools/issues/594)).

  - I could not find the right package
   - Strange enough I can not find it in the package index but I am able to install it on ubuntu 20.04
- Fedora: https://fedora.pkgs.org/34/fedora-x86_64/python3-osrf-pycommon-0.2.1-1.fc34.noarch.rpm.html
- Gentoo: https://packages.gentoo.org/packages/dev-python/osrf_pycommon

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6722,6 +6722,10 @@ python3-opengl:
   gentoo: [dev-python/pyopengl]
   nixos: [python3Packages.pyopengl]
   ubuntu: [python3-opengl]
+python3-osrf-pycommon:
+  fedora: [python3-osrf-pycommon]
+  gentoo: [dev-python/osrf_pycommon]
+  ubuntu: [python3-osrf-pycommon]
 python3-owlready2-pip:
   debian:
     pip:


nuclearsandwich:
@rickstaa is the rosdep key needed for the catkin-tools release? I didn't think that rosdep was part of the release process for that project.

The reason that you could not find an upstream package for osrf-pycommon in Debian and Ubuntu is that the package is one that we provide via the ROS repositories. I can confirm that there is some python3 version in the package manifests for all supported Debian and Ubuntu versions.

However it is also worth noting that due to quirks of history, there is a separate version of the osrf-pycommon package that is released directly into ROS 2 distributions via distribution-specific packaging. So it is possible for ROS 2 packages to depend on `osrf_pycommon` and rosdep will fulfill that dependency with the appropriate ROS 2 package. For example: ros-foxy-osrf-pycommon. However that wouldn't be the correct approach for a rosdistro agnostic package like catkin-tools.

rickstaa:
@nuclearsandwich Thanks a lot for your answer. No, the rosdep key is not needed for the catkin-tools release. This pull request was just because some of my packages depend on the [catkin_tools](ps://github.com/catkin/catkin_tools) package currently has an upstream bug that causes me to install the osrf_pycommon package manually. As this bug has been open for a long time and I was tired of manually installing it, I added it to the rosdep/python.yaml file.

However, I see that a pull request was created a few days ago that fixes this upstream bug [catkin_tools/648](https://github.com/catkin/catkin_tools/pull/648). As this pull request is likely to be merged this month, we can close this pull request.

PR #29451, opened 2021-05-06
ur_client_library: 0.2.1-1 in 'foxy/distribution.yaml' [bloom]

fmauch:
Increasing version of package(s) in repository `ur_client_library` to `0.2.1-1`:

- upstream repository: https://github.com/UniversalRobots/Universal_Robots_Client_Library
- release repository: https://github.com/UniversalRobots/Universal_Robots_Client_Library-release.git
- bloom version: `0.10.7`

## ur_client_library

```
* Run ci also for ROS2 foxy
* Prepare package.xml and cmakelists for ROS2
* Add downstream workspace
* Contributors: Felix Exner, Lennart Puck
```

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -4719,6 +4719,21 @@ repositories:
       url: https://github.com/ros2/unique_identifier_msgs.git
       version: foxy
     status: maintained
+  ur_client_library:
+    doc:
+      type: git
+      url: https://github.com/UniversalRobots/Universal_Robots_Client_Library.git
+      version: boost
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/UniversalRobots/Universal_Robots_Client_Library-release.git
+      version: 0.2.1-1
+    source:
+      type: git
+      url: https://github.com/UniversalRobots/Universal_Robots_Client_Library.git
+      version: boost
+    status: developed
   urdf:
     doc:
       type: git


nuclearsandwich:
@fmauch because of the [complicated license structure](https://github.com/UniversalRobots/Universal_Robots_Client_Library#license) of this repository it makes sense to me to add `file=` attributes to the license tags in the package.xml (as described in [REP-149](https://www.ros.org/reps/rep-0149.html#license-multiple-but-at-least-one) to pinpoint where the license files for the subprojects are located.

Other than the above this meets the review criteria for a new package in Foxy.

fmauch:
That seems to be a good idea. What would be the release workflow for this? I would probably make a new release v0.2.2 with those changes?

jacobperron:
Unless there's a different opinion, I think it's okay to hold off on making a new release. As long as the license-related changes are in the source branch, folks should be able to find that information.

nuclearsandwich:
> Unless there's a different opinion, I think it's okay to hold off on making a new release. As long as the license-related changes are in the source branch, folks should be able to find that information.

My soft objection would be that the rosdistro_cache.yaml only stores the release package.xml files not those for the source repository and that the package.xml installed with the package would also be missing that added information but given that the current changes are already released in other ROS distributions I'm actually going to turn that soft objection into a recommendation for the next round of releases. :shipit:

PR #29489, opened 2021-05-11
paho-mqtt-c: 1.3.8-2 in 'noetic/distribution.yaml' [bloom]

Timple:
Increasing version of package(s) in repository `paho-mqtt-c` to `1.3.8-2`:

- upstream repository: https://github.com/eclipse/paho.mqtt.c.git
- release repository: https://github.com/nobleo/paho.mqtt.c-release.git
- bloom version: `0.10.7`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4774,7 +4774,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/nobleo/paho.mqtt.c-release.git
-      version: 1.3.8-2
+      version: 1.3.8-3
     source:
       test_commits: false
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4769,6 +4769,17 @@ repositories:
       url: https://github.com/allenh1/p2os.git
       version: main
     status: maintained
+  paho-mqtt-c:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/nobleo/paho.mqtt.c-release.git
+      version: 1.3.8-2
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@Timple this is the second non-ROS package I've reviewed from you recently. Is there a ROS package that you intend to release which depends on these packages? Otherwise it might actually make sense not to include them in ROS distributions.

As a nitpick on the [license](https://github.com/nobleo/paho.mqtt.c-release/blob/master/noetic/package.xml#L12) field in your overlaid package.xml it looks like the [OSI license string](https://opensource.org/licenses/alphabetical) for the Eclipse Public License 2.0 is `Eclipse Public License 2.0` or `EPL-2.0` for short whereas you've put `Eclipse Public License - v 2.0`.

Timple:
The decision of open sourcing the cloud connectivity client is an agenda-point here. To that end I thought I would take a head-start and release two of the dependencies already.

I was hoping the mqtt iothub and azure sdk would be generic enough for other developers to use as well. Let me know if this is not the right approach.

(PS. fixed license)

mabelzhang:
@nuclearsandwich @tfoote or whoever doing rosdistro reviews, I'm reading in the review guidelines that non-ROS packages "require special handling in the release repository." Other points seem satisfied. Do we approve this and the other ticket referenced above or is there something else needed?

tfoote:
Since this is building towards a ROS package I think it resolves @nuclearsandwich 's concern. We mostly don't want to just replace our upstream distros.

mabelzhang:
Oh were we still holding for Noetic because of the resync? https://discourse.ros.org/t/preparing-for-noetic-sync-2021-05-24/20540

tfoote:
@sloretz Sorry I forgot the freeze, but as this is new it shouldn't effect much.

PR #29508, opened 2021-05-12
Add flake8 plugins to python.yaml

kenji-miyake:
## Package name:

- python3-flake8-blind-except-pip
- python3-flake8-builtins-pip
- python3-flake8-class-newline-pip
- python3-flake8-comprehensions-pip
- python3-flake8-deprecated-pip
- python3-flake8-docstrings-pip
- python3-flake8-import-order-pip
- python3-flake8-quotes-pip

## Package Upstream Source:

- https://github.com/elijahandrews/flake8-blind-except
- https://github.com/gforcada/flake8-builtins
- https://github.com/AlexanderVanEck/flake8-class-newline
- https://github.com/adamchainz/flake8-comprehensions
- https://github.com/gforcada/flake8-deprecated
- https://gitlab.com/pycqa/flake8-docstrings
- https://github.com/PyCQA/flake8-import-order
- https://github.com/zheller/flake8-quotes/

## Purpose of using this:

As I wrote in https://github.com/osrf/docker_images/issues/531, it would be nice to make installing flake8 plugins possible with `rosdep`.

These plugins are what listed [here](https://docs.ros.org/en/rolling/Installation/Ubuntu-Development-Setup.html#install-development-tools-and-ros-tools).
If we don't install these plugins, it causes differences in `colcon test`.
Currently, we have to install them manually in CI scripts, but it's useful if we can use `rosdep`.

- https://pypi.org/project/flake8-blind-except/
- https://pypi.org/project/flake8-builtins/
- https://pypi.org/project/flake8-class-newline/
- https://pypi.org/project/flake8-comprehensions/
- https://pypi.org/project/flake8-deprecated/
- https://pypi.org/project/flake8-docstrings/
- https://pypi.org/project/flake8-import-order/
- https://pypi.org/project/flake8-quotes/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6257,6 +6257,38 @@ python3-flake8:
     '*': ['python%{python3_pkgversion}-flake8']
     '7': null
   ubuntu: [python3-flake8]
+python3-flake8-blind-except-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-blind-except]
+python3-flake8-builtins-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-builtins]
+python3-flake8-class-newline-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-class-newline]
+python3-flake8-comprehensions-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-comprehensions]
+python3-flake8-deprecated-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-deprecated]
+python3-flake8-docstrings-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-docstrings]
+python3-flake8-import-order-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-import-order]
+python3-flake8-quotes-pip:
+  ubuntu:
+    pip:
+      packages: [flake8-quotes]
 python3-flask:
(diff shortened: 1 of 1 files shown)


hidmic (commented):
This seems reasonable in principle. I'd think it isn't this way already because (a) you'd rather avoid cluttering the package.xml of every ROS 2 package that bears Python code and (b) `ament_flake8` is not strictly coupled to ROS 2 and thus there's no common place to put these. 

That's not a good reason not to include these here though. One should be free to trade clutter for convenience. But I will point out that you'll have to manually install these for the majority of ROS 2 packages anyhow (for the reasons exposed above). What would be the use case?

hidmic:
Another thought towards reducing the amount of install instructions, an `ament_flake8_ros` package could perhaps help.

wjwwood (approved):
> an ament_flake8_ros package could help.

👍, but as you said, having these rules in place doesn't pose a problem, and if we made that package we'd want these rules to depend on anyway, I guess.

kenji-miyake:
Thank you for your comments.
I agree with the idea of `ament_flake8_ros`.

I'd like to add flake8 plugins to `rosdep` so that someone can create `ament_flake8_ros` and that it can depend on flake8 plugins.

hidmic (approved):

PR #29514, opened 2021-05-13
Add python3-mlflow to rosdep

n-meunier:

## Package name:

`python3-mlflow-pip`
Edit: since it is not only available with `pip`, new name is `python3-mlflow`.

## Package Upstream Source:

https://pypi.org/project/mlflow/
https://build.opensuse.org/package/show/home:jayvdb:py-new/python-mlflow

## Purpose of using this:

https://pypi.org/project/mlflow/#description

A platform to manage Machine Learning models and experiments.

Distro packaging links:

  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
  - NOT_FOUND
- openSUSE: https://build.opensuse.org/search
  - AVAILABLE

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6627,9 +6627,6 @@ python3-mlflow:
   fedora:
     pip:
       packages: [mlflow]
-  gentoo:
-    pip:
-      packages: [mlflow]
   opensuse: [python-mlflow]
   osx:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6620,14 +6620,18 @@ python3-meshio:
   arch: [python-meshio]
   debian: [python3-meshio]
   ubuntu: [python3-meshio]
-python3-mlflow-pip:
+python3-mlflow:
   debian:
     pip:
       packages: [mlflow]
   fedora:
     pip:
       packages: [mlflow]
-  ubuntu: 
+  opensuse: [python-mlflow]
+  osx:
+    pip:
+      packages: [mlflow]
+  ubuntu:
     pip:
       packages: [mlflow]
 python3-mock:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6627,6 +6627,9 @@ python3-mlflow:
   fedora:
     pip:
       packages: [mlflow]
+  gentoo:
+    pip:
+      packages: [mlflow]
   opensuse: [python-mlflow]
   osx:
     pip:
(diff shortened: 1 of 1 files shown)


n-meunier:
FYI @yotabits

rosdep/python.yaml
```diff
@@ -6620,6 +6620,16 @@ python3-meshio:
...
+python3-mlflow-pip:
+  debian:
+    pip:
+      packages: [mlflow]
+  fedora:
+    pip:
+      packages: [mlflow]
+  ubuntu: 
+    pip:
+      packages: [mlflow]
```
hidmic, line 6632: @n-meunier mind to add an entry for OSX and Gentoo? BTW it appears there's `mlflow` 1.11.0 [on OpenSUSE](https://build.opensuse.org/package/show/home:jayvdb:py-new/python-mlflow).

n-meunier: Thanks for the review. I'm gonna add an entry for OSX and OpenSUSE. 
For Gentoo, I could not find the package in https://packages.gentoo.org/. Isn't it the condition for this entry?

hidmic: Yeah, it doesn't look like we can use anything but `pip` on Gentoo.

n-meunier: I haven't found any other example with:
```
  gentoo:
    pip:
      packages: 
```
But I added it following your comment.

hidmic (approved):
This seems reasonable to me. @wjwwood ?

rosdep/python.yaml
tfoote: My impression is that gentoo doesn't like alternative installers like pip. @allenh1  Can you confirm?

allenh1: > My impression is that gentoo doesn't like alternative installers like pip.

Yeah, you shouldn't install things with pip on gentoo. These days, if you attempt to run pip install you get an error message.

```
ERROR: (Gentoo) Please run pip with the --user option to avoid breaking python-exec
```

Previously it would just break your package manager :upside_down_face:.

What we can do, though, is create a package in the overlay repo. We've done this for a few deps. I'll make an issue there for it.

allenh1: I made a ticket to release the dependency [here](https://github.com/ros/ros-overlay/issues/1019). In the meantime, feel free to merge without a gentoo resolution, as there unfortunately isn't one.

n-meunier: Thank you for the ticket and your explanation.

n-meunier:
@tfoote @allenh1 thank you for the review, I reverted the Gentoo entry.

mabelzhang (approved):

PR #29534, opened 2021-05-15
rosdep: OpenEmbedded updates

shr-project:
libsndfile1: https://layers.openembedded.org/layerindex/recipe/733/
portaudio-v1: https://layers.openembedded.org/layerindex/recipe/1080/
tar: https://layers.openembedded.org/layerindex/recipe/71/
qttools: https://layers.openembedded.org/layerindex/recipe/24103/
libx11: https://layers.openembedded.org/layerindex/recipe/5883/
libxt: https://layers.openembedded.org/layerindex/recipe/5128/
python3-termcolor: https://layers.openembedded.org/layerindex/recipe/96040/
python3-pytest-cov: will be added in meta-ros with https://github.com/ros/meta-ros/pull/849

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4784,6 +4784,7 @@ libsndfile1-dev:
   fedora: [libsndfile-devel]
   gentoo: [media-libs/libsndfile]
   nixos: [libsndfile]
+  openembedded: [libsndfile1@openembedded-core]
   rhel: [libsndfile-devel]
   ubuntu: [libsndfile1-dev]
 libsoqt4-dev:
@@ -5364,6 +5365,7 @@ libx11:
   gentoo: [x11-libs/libX11]
   macports: [xorg-libX11]
   nixos: [xorg.libX11]
+  openembedded: [libx11@openembedded-core]
   opensuse: [libX11-6]
   ubuntu: [libx11-dev]
 libx11-dev:
@@ -5521,6 +5523,7 @@ libxt-dev:
   fedora: [libXt-devel]
   gentoo: [x11-libs/libXt]
   nixos: [xorg.libXtst]
+  openembedded: [libxt@openembedded-core]
   ubuntu: [libxt-dev]
 libxtst-dev:
   arch: [libxtst]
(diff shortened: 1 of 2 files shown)


tfoote:
@shr-project Any reason not to merge this as is? You've left it in draft state. We generally like to keep PRs on this repo small and incremental so we can keep the merge rate up.

shr-project:
@tfoote I've left it in draft state, because I'm preparing the PR early to be able to reference it from meta-ros changes I'm working on (in this case here https://github.com/ros/meta-ros/pull/844/commits/f05229d9f59bfc97bc68900156d5a5a9b5fae424#diff-8dcc2f1a452ff6be41e5c37d616e87487325dda175be26876724ea2e7a76b8f9R235).

But because meta-ros1-noetic/generated-recipes/picovoice-ros/picovoice-driver_0.0.4-1.bb doesn't even build yet, it wasn't properly tested end-to-end (in this case not very likely due to wrong assignment here). Also when the values are properly tested in the build I need to re-generate the meta-ros recipes with this PR included in local rosdistro checkout, so that the recipes contain the right dependencies directly instead of these ROS_UNRESOLVED_DEP* variables set in ros-distro.inc file. And sometimes I leave this re-generation for integration of the next distro upgrade (or across all ROS distributions like in https://github.com/ros/meta-ros/pull/832 when there are more ROS_UNRESOLVED_DEP* variables)

So it takes a while and I was hoping that in draft state nobody will look at it here until it's marked as ready. Let me know if I should stop creating the drafts and wait until it's completely ready.

@herb-kuta-lge FYI

tfoote:
They can stay here in draft state. It would be helpful to put a comment in about what it's waiting for or supporting while in draft state so new reviewers can understand it quickly and know when it can be passed over.

shr-project:
I'm done with testing, verified with latest noetic release in meta-ros. Ready for review now.

jacobperron (approved):

nuclearsandwich (approved):

nuclearsandwich:
@shr-project I don't know if you've seen the [rosdep checker](https://github.com/ros/rosdistro/tree/master/test/rosdep_repo_check) that we recently added but if there's a way to integrate openembedded checks it would encourage others to help maintain the openembedded rules. No pressure, but if you had not seen it I expect you'd like to know it's there.

PR #29553, opened 2021-05-17
locator_ros_bridge: 2.1.0-1 in 'rolling/distribution.yaml' [bloom]

slaible:
Increasing version of package(s) in repository `locator_ros_bridge` to `2.1.0-1`:

- upstream repository: https://github.com/boschglobal/locator_ros_bridge.git
- release repository: https://github.com/ros2-gbp/locator_ros_bridge-release.git
- bloom version: `0.10.7`

## bosch_locator_bridge

```
* initial version
* Contributors: Stefan Laible
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1177,6 +1177,23 @@ repositories:
       url: https://github.com/ros2/libyaml_vendor.git
       version: master
     status: maintained
+  locator_ros_bridge:
+    doc:
+      type: git
+      url: https://github.com/boschglobal/locator_ros_bridge.git
+      version: rolling
+    release:
+      packages:
+      - bosch_locator_bridge
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/locator_ros_bridge-release.git
+      version: 2.1.0-1
+    source:
+      type: git
+      url: https://github.com/boschglobal/locator_ros_bridge.git
+      version: rolling
+    status: maintained
   marti_common:
     doc:
       type: git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1181,7 +1181,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/boschglobal/locator_ros_bridge.git
-      version: rolling
+      version: main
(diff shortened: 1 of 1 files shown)


rolling/distribution.yaml
```diff
@@ -1177,6 +1177,23 @@ repositories:
...
+      packages:
+      - bosch_locator_bridge
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/locator_ros_bridge-release.git
+      version: 2.1.0-1
+    source:
+      type: git
+      url: https://github.com/boschglobal/locator_ros_bridge.git
+      version: rolling
```
hidmic, line 1195: @slaible I see no `rolling` branch in the source repository.

hidmic: There's a 2.1.0 tag in the `main` branch, so maybe you meant to use `main`?

slaible: @hidmic Yes, the rolling version is in the main branch. Is this a problem?

wjwwood: Yeah, the value here needs to be a tag or branch in the upstream repository. Which appears to not be the case for 'rolling'?

slaible: @wjwwood Ok, so what do I have to do? Is it sufficient if I change version to main in line 1195 of distribution.yaml? (The main branch contains the rolling version.)

tfoote: Yes, if you update 1195 as well as 1184 to be `main` that should let it pass CI.

wjwwood: That is not covered in the tutorial.

The tool prompts you, asking if you want to have a source and/or doc entry. At that point you put in the repository url, vcs type (e.g. `git`), and the version. I think you just misunderstood that `version` in that context means a tree-ish or commit like thing, either a branch or tag or commit hash for git. Basically you're telling us where to get the source code for documentation or building from source. This might be different for each one case. This version is not for "which ros distribution", as we e already know it's for rolling at this point.

slaible: Ah ok, thank you for the explanation!

Issue #27851, opened 2020-12-28
Inconsistent Gazebo versions in Debian Buster


StefanFabian:
I've encountered this issue in this [failed build log](https://build.ros.org/job/Nbin_db_dB64__gazebo_ros_control_select_joints__debian_buster_amd64__binary/6/console) on Debian Buster.
Apparently, the gazebo rosdep key on Debian Buster resolves to gazebo9
```bash
$ rosdep resolve --os=debian:buster gazebo
#apt
gazebo9
```
However, the noetic version of gazebo-dev `ros-noetic-gazebo-dev` for [Debian Buster](http://packages.ros.org/ros/ubuntu/dists/buster/main/binary-amd64/Packages) depends on `gazebo11`.

What is the recommended approach for dealing with this?
While in theory it would possibly break some Debian Buster setups to switch the `gazebo` key to `gazebo11`, I'd argue that it already is broken since some packages depend on `gazebo11` and others via `gazebo` on `gazebo9`.
There are no debian packages available for Buster that depend on gazebo9 as far as a simple search could tell (5 matches all Conflicts declarations).

StefanFabian:
@tfoote @wjwwood @sloretz
Any opinion on this?
Should I just submit a patch for buster that changes the key to gazebo11?
Seems like a hacky solution which is why I haven't done it so far.

tfoote:
Yeah, this is kind of ending up in a noman's land where there are two versions of gazebo available on that platform that conflict. I think that since this is the keys for the ROS installation it makes sense to target the version that ROS is expecting. We ran into this problem even more so with stretch where the stretch was officially targeted by two distros with conflicting version requirements. (#19843) 

I think that homogenizing these rules to version 11 makes sense on buster to make sure that we're consistent and don't end up with people having incompatible dependencies.

StefanFabian:
Hm, the build is still failing. From the docker commands, I read that the install command is cached?
Do I have to make another release to make sure it uses the new value?

PR #29578, opened 2021-05-19
Update default gazebo version on debian buster to 11

StefanFabian:
This would fix the inconsistency described in #27851

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1140,7 +1140,7 @@ gawk:
 gazebo:
   arch: [gazebo]
   debian:
-    buster: [gazebo9]
+    buster: [gazebo11]
     jessie: [gazebo7]
     stretch: [gazebo9]
   fedora: [gazebo]


tfoote (approved):
Thanks for the long hanging followup.

mabelzhang:
For the 2nd approval, I tried to follow the conversations but it looks complicated. Is it normal that gazebo11 isn't actually listed as an available package in Buster? https://packages.debian.org/search?keywords=gazebo

StefanFabian:
> Thanks for the long hanging followup.

I guess there was some miscommunication. I thought you guys would consider it and change it if there isn't any reason not to.
I'm always hesitant to make single-line change PRs for systems I can't even test on.
However, I've made an exception since the build server kept emailing me to remind me.

> For the 2nd approval, I tried to follow the conversations but it looks complicated. Is it normal that gazebo11 isn't actually listed as an available package in Buster? https://packages.debian.org/search?keywords=gazebo

I think it's available from the ros build farm / debian repo: http://packages.ros.org/ros/ubuntu/dists/buster/main/binary-amd64/Packages
but I'm not 100% sure and I don't have a debian system to confirm this.

tfoote:
There's two versions of gazebo available on Buster gazebo9 is upstream gazebo11 is available from the ROS sources. 

Eloquent used gazebo9 on buster but is EOL: https://www.ros.org/reps/rep-2000.html#id36

Noetic uses gazebo11 on buster and is still supported. https://www.ros.org/reps/rep-0003.html#id23
Foxy uses gazebo11 on buster and is still supported. https://www.ros.org/reps/rep-2000.html#id37

Luckily Melodic only supports stretch so it doesn't confound this issue.

This will mean that the gazebo that you install on buster by the generic `gazebo` key will match the one in both of our active LTS releases instead of conflict. If you need gazebo 9 specifically you can use the `gazebo9` key. The key was originally picked for Eloquent

mabelzhang (approved):
Thanks for the explanation guys!

PR #29586, opened 2021-05-20
imu_from_ios_sensorlog: 0.0.1-1 in 'noetic/distribution.yaml' [bloom]

pietrocolombo:
Increasing version of package(s) in repository `imu_from_ios_sensorlog` to `0.0.1-1`:

- upstream repository: https://github.com/pietrocolombo/imu_from_ios_sensorlog.git
- release repository: https://github.com/pietrocolombo/imu_from_ios_sensorlog-release.git
- bloom version: `0.10.7`

## imu_from_ios_sensorlog

```
* initial version
* Contributors: Pietro Colombo, pietrocolombo
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2820,7 +2820,7 @@ repositories:
         release: release/noetic/{package}/{version}
       url: https://github.com/pietrocolombo/imu_from_ios_sensorlog-release.git
       version: 0.0.1-1
-    status: unmaintained
+    status: maintained
   imu_pipeline:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2810,6 +2810,13 @@ repositories:
       url: https://github.com/ros-perception/image_transport_plugins.git
       version: noetic-devel
     status: unmaintained
+  imu_from_ios_sensorlog:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/pietrocolombo/imu_from_ios_sensorlog-release.git
+      version: 0.0.1-1
+    status: unmaintained
   imu_pipeline:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2820,7 +2820,6 @@ repositories:
         release: release/noetic/{package}/{version}
       url: https://github.com/pietrocolombo/imu_from_ios_sensorlog-release.git
       version: 0.0.1-1
-    status: maintained
     source:
       type: git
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -2810,6 +2810,13 @@ repositories:
       url: https://github.com/ros-perception/image_transport_plugins.git
       version: noetic-devel
     status: unmaintained
+  imu_from_ios_sensorlog:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/pietrocolombo/imu_from_ios_sensorlog-release.git
+      version: 0.0.1-1
+    status: unmaintained
```
tfoote, line 2819: If this is unmaintained, we probably shouldn't release it into the distro.

tfoote (commented):
In addition to my question about whether this should be released.

If so it would also be valuable to add doc and source entries pointing to the `main` or `noetic-devel` branch depending on which one is more appropriate.

noetic/distribution.yaml
jacobperron: @pietrocolombo Friendly ping. If this is not actually unmaintained, then this status should be changed to `developed` (if it is actively being worked on) or `maintained`. Otherwise, I agree that we probably shouldn't release this package.

I noticed that the package does not declare a license. Please update the `package.xml` to declare a license and add a LICENSE file (see [CONTRIBUTING](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-package-contents) for more details).

pietrocolombo: Thanks, I update the status.

pietrocolombo: Thanks, I have also updated the license.

jacobperron (commented):
Thanks for adding the LICENSE file :slightly_smiling_face:

For completeness, I think we should also add a `source` entry here (which should look similar to the `doc` entry).

pietrocolombo:
> Thanks for adding the LICENSE file 🙂
> 
> For completeness, I think we should also add a `source` entry here (which should look similar to the `doc` entry).

just added

nuclearsandwich:
Thanks for iterating with us @pietrocolombo is there a reason the doc version is `main` but the source version is `noetic-devel`.

pietrocolombo:
> Thanks for iterating with us @pietrocolombo is there a reason the doc version is `main` but the source version is `noetic-devel`?

You are right it makes more sense to put the `noetic-devel` branch

nuclearsandwich (approved):

nuclearsandwich:
[Holding for a Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2021-06-25/21052)

PR #29426, opened 2021-05-05
[rosdep] Adding jupyros

FSund:
Please add the following dependency to the rosdep database.

## Package name:

python3-jupyros-pip

## Package Upstream Source:

https://github.com/RoboStack/jupyter-ros

## Purpose of using this:

jupyros is an extension to Jupyter Notebook that enables easy connections to ROS.

Distro packaging links:

## Links to Distribution Packages

Only available through PyPI: https://pypi.org/project/jupyros/

PR #29587, opened 2021-05-20
Add virtualserialports to python.yaml

ggaessler:

## Package name:

~python-virtualserialports-pip~
python3-virtualserialports-pip

## Package Upstream Source:

https://github.com/ezramorris/PyVirtualSerialPorts

## Purpose of using this:

Testing ROS serial device drivers.  

Distro packaging links:

Only available through PyPI: https://pypi.org/project/PyVirtualSerialPorts/


tfoote (changes requested):
I just approved the CI run. 

One request though. Unless you're going to be using the python2 version we shouldn't add it. I believe all platforms have moved to python3 now and adding the extra rule that's unused will just require cleanup in the future.

rosdep/python.yaml
```diff
@@ -7858,6 +7871,19 @@ python3-venv:
...
   ubuntu: [python3-venv]
+python3-virtualserialports-pip:
+  debian:
+    pip:
+      packages: [PyVirtualSerialPorts]
+  fedora:
+    pip:
+      packages: [PyVirtualSerialPorts]
+  gentoo:
+    pip:
```
mabelzhang, line 7882: Did we want to remove the gentoo entry when it's pip?

ggaessler:
> I just approved the CI run.
> 
> One request though. Unless you're going to be using the python2 version we shouldn't add it. I believe all platforms have moved to python3 now and adding the extra rule that's unused will just require cleanup in the future.

I just put it in for backwards compatibility. I'm not needing it. Just removed the Python2 part.

rosdep/python.yaml
sloretz: @mabelzhang Why should the pip entry on gentoo be removed?

mabelzhang: I was reading this comment at the time https://github.com/ros/rosdistro/pull/29514#discussion_r636348684
It sounds like it's a general gentoo+pip thing, and not specific to that PR?

ggaessler: Shall I remove gentoo? I personally just use ubuntu and don't really care about gentoo, but added it in since it was also included for other pip PRs here, e.g.: https://github.com/ros/rosdistro/pull/29426/files

tfoote: Yes please remove pip for gentoo: https://github.com/ros/rosdistro/pull/29514#discussion_r636348684

ggaessler: Gentoo is removed

mabelzhang: Thank you for iterating!

mabelzhang:
Not sure why the test is failing. Looks like a new test. Could you do a merge from `master` and see if that fixes it?

ggaessler:
> Not sure why the test is failing. Looks like a new test. Could you do a merge from `master` and see if that fixes it?

Merged it

mabelzhang (approved):
Checks passed

PR #29624, opened 2021-05-21
c_wrapper: 0.0.1-3 in 'foxy/distribution.yaml' [bloom]

OTL:
Increasing version of package(s) in repository `c_wrapper` to `0.0.1-3`:

- upstream repository: https://github.com/smilerobotics/c_wrapper.git
- release repository: https://github.com/smilerobotics/c_wrapper-release.git
- bloom version: `0.10.7`

## c_wrapper

```
* Start develop
* Contributors: Takashi Ogura
```

## c_wrapper_nav2

```
* Start develop
* Contributors: Takashi Ogura
```

## c_wrapper_ros2

```
* Start develop
* Contributors: Takashi Ogura
```

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -395,6 +395,26 @@ repositories:
       url: https://github.com/ros/bond_core.git
       version: dashing-devel
     status: maintained
+  c_wrapper:
+    doc:
+      type: git
+      url: https://github.com/smilerobotics/c_wrapper.git
+      version: main
+    release:
+      packages:
+      - c_wrapper
+      - c_wrapper_nav2
+      - c_wrapper_ros2
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/smilerobotics/c_wrapper-release.git
+      version: 0.0.1-3
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/smilerobotics/c_wrapper.git
+      version: main
+    status: developed
   carla_msgs:
     doc:
       type: git


jacobperron:
Note, we have [`rcl`](https://github.com/ros2/rcl) that is the C library containing implementation of core ROS 2 features (used by `rclcpp`, `rclpy`, and other client libraries). There is also [`rclc`](https://github.com/ros2/rclc) (ROS 2 client library written in C). Depending on the goals of your project, it might be better to contribute effort to the existing libraries since it seems like there may be some overlap.

OTL:
Thank you for the kind information! (But I could not find action examples. )

The actual purpose of this package is to be used from Rust code.
(Especially for [OpenRR](https://github.com/openrr/openrr) project)

There are no ROS2 rust clients which implement actions today, and linking ROS2 C++ library directly is very difficult.
But if we write a C-lang wrapper, the link process will be super easy for Rust libraries.
So, yes, rclc will be the candidates as you mention. (But the Rust code will become more complecated)

If you think c_wrapper sounds too generic, and make bad effect for the community, I can change the name of this package.
(Then, i'll use something like `arci-ros2-sys`, `arci-nav2-sys`)

What do you think?

OTL:
ping for a reviewer!

jacobperron:
@OTL Apologies for the delay.

`rcl_action` exists alongside `rcl` to provide a base set of C functions for implementing ROS actions: https://github.com/ros2/rcl/tree/master/rcl_action
Other language-specific client libraries implement actions by calling out to `rcl_action`. Though, a lot of logic related to the execution model remains in the higher level client libraries (rclcpp and rclpy) since it's often easier to write this logic there. I imagine a Rust implementation might try to do the same (i.e. directly depend on `rcl` and `rcl_action`).

Though it probably doesn't support actions (yet), I guess you've already found the ROS 2 Rust client library: https://github.com/ros2-rust/ros2_rust
IMO, this would be a good place to contribute for work related to ROS 2  + Rust.

---

If you still want to release your C wrapper library, I think it would be better to rename it as you suggest since "c_wrapper" is a very generic name that doesn't really describe the package.

OTL:
OK, thank you for the great feedback!!

FYI: I'm already using [r2r](https://github.com/sequenceplanner/r2r) (yet another rust ros2 client) (The approach is fit for our purpose more than ros2_rust), but it takes some more time to implement `action` functionality as same as ros2_rust.
I want to contribute to `rclc`, too. 

Thanks, again. And see you again in another PR.

PR #29671, opened 2021-05-25
Adding riptide_controllers to documentation index for noetic

blaine141:
I'd like riptide_controllers to be indexed and documented on ros.org.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5870,6 +5870,12 @@ repositories:
       url: https://github.com/ros-drivers/rgbd_launch.git
       version: noetic-devel
     status: maintained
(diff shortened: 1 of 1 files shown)


mabelzhang (commented):
I see that the license is BSD, could you add a `LICENSE` file as described in 3.ii in the review guidelines here? https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md

Also [holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2021-05-24/20540).

blaine141:
The LICENSE file has been added.

sloretz (approved):

PR #29716, opened 2021-05-29
Update the GPG public keys with new expiration dates

tfoote:
The ros.asc and ros.key public keys expired. We have reexported them and they now expire in 2025-06-01


cottsay (approved):
I tested the binary key and it seems to have been converted correctly.

tonypascal:
hello,
when trying to access http://packages.ros.org/ros2/ubuntu jammy InRelease with apt, I get : 
Error connecting: Could not connect: Connection refused
can you explain why ?

PR #29721, opened 2021-05-30
rosdep: nixos: use SIP 4

lopsided98:
nixpkgs recently changed the `pythonPackages.sip` attribute to point at SIP 5 (soon to be upgraded to 6). On most other other distributions the python-sip (or similar) package is SIP 4, therefore all ROS packages I know of currently only work with SIP 4. Newer SIP versions have significant changes, making upgrading somewhat involved.

nixpkgs source: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/sip/4.x.nix

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4566,7 +4566,7 @@ python-sip:
   freebsd: [py27-sip]
   gentoo: [dev-python/sip]
   macports: [py27-sip]
-  nixos: [pythonPackages.sip]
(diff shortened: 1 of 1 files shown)


audrow (approved):

tfoote (approved):
I confirmed that Ubuntu all the way out to impish is sip 4 https://packages.ubuntu.com/impish/python3-sip-dev

PR #29748, opened 2021-06-02
rmf_battery: 0.1.0-1 in 'rolling/distribution.yaml' [bloom]

marcoag:
Increasing version of package(s) in repository `rmf_battery` to `0.1.0-1`:

- upstream repository: https://github.com/open-rmf/rmf_battery.git
- release repository: https://github.com/ros2-gbp/rmf_battery-release.git
- bloom version: `0.10.7`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2036,7 +2036,7 @@ repositories:
     source:
       type: git
       url: https://github.com/open-rmf/rmf_battery.git
-      version: 0.1.0
+      version: master
     status: developed
   rmf_building_map_msgs:
     doc:
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2023,6 +2023,21 @@ repositories:
       url: https://github.com/ros/resource_retriever.git
       version: ros2
     status: maintained
+  rmf_battery:
+    doc:
+      type: git
+      url: https://github.com/open-rmf/rmf_battery.git
+      version: rolling
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/rmf_battery-release.git
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/open-rmf/rmf_battery.git
+      version: 0.1.0
+    status: developed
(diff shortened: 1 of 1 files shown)


rolling/distribution.yaml
```diff
@@ -2023,6 +2023,21 @@ repositories:
       url: https://github.com/ros/resource_retriever.git
       version: ros2
     status: maintained
+  rmf_battery:
+    doc:
+      type: git
+      url: https://github.com/open-rmf/rmf_battery.git
+      version: rolling
```
sloretz, line 2030: It looks like the `rolling` branch doesn't exist. That said, I would recommend changing this to `main`. Releasing from a distro specific branch like `galactic`, `foxy`, etc is useful for maintaining stability and backwards compatibility, but it comes with some overhead. For ROS Rolling releasing from `main`/`master`/`ros2` is fine because it's expected to be unstable.

rolling/distribution.yaml
```diff
@@ -2023,6 +2023,21 @@ repositories:
...
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/rmf_battery-release.git
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/open-rmf/rmf_battery.git
+      version: 0.1.0
```
sloretz, line 2039: I think the source entry version is expected to be the development branch. Tools like `rosinstall_generator` use this info to checkout the upstream development version.

sloretz (changes requested):

rolling/distribution.yaml
marcoag: Sorry the `rolling` branch got deleted after a PR was merged, I just restored it. 

Thanks for the tip regarding the branch, I agree and would move `rolling` to `main` on the next release, for this one I guess I would rather keep it on the `rolling` branch as we internally agreed before, and for consistency with the other packages.

rolling/distribution.yaml
marcoag: Ok, just changed it to `main`, hope it's fine now.

sloretz (approved):

PR #29777, opened 2021-06-04
Revert "control_msgs: 3.0.0-1 in 'rolling/distribution.yaml' [bloom]"

nuclearsandwich:
This revert is proposed in order to resolve package build failures due to an outdated bloom version used for the release. An alternative to this PR would be a new release of control_msgs 3.0.0 using bloom 0.10.7.

Reverts ros/rosdistro#29691


nuclearsandwich:
Holding for review of https://github.com/ros-infrastructure/ros_buildfarm/pull/889 which will require a deep rebuild of Rolling on RHEL8.

PR #29817, opened 2021-06-09
control_toolbox: 2.0.2-1 in 'rolling/distribution.yaml' [bloom]

bmagyar:
Increasing version of package(s) in repository `control_toolbox` to `2.0.2-1`:

- upstream repository: https://github.com/ros-controls/control_toolbox.git
- release repository: https://github.com/ros-gbp/control_toolbox-release.git
- bloom version: `0.10.7`

## control_toolbox

```
* remove unused variables
* Update visibility_control.hpp
* Windows bringup.
* Contributors: Karsten Knese, Sean Yen, Bence Magyar
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -461,6 +461,21 @@ repositories:
       url: https://github.com/ros-controls/control_msgs.git
       version: foxy-devel
     status: maintained
+  control_toolbox:
+    doc:
+      type: git
+      url: https://github.com/ros-controls/control_toolbox.git
+      version: ros2-master
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros-gbp/control_toolbox-release.git
+      version: 2.0.2-1
+    source:
+      type: git
+      url: https://github.com/ros-controls/control_toolbox.git
+      version: ros2-master
+    status: maintained
   cyclonedds:
     release:
       tags:


ahcorde (approved):

gbalke:
I believe this is missing a `LICENSE` file?

bmagyar:
Can you elaborate on that? I was expecting it being held for a sync. This
same version was released into Foxy and Galactic.

ahcorde:
According with the [guidelines](https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md)

 > Either each package in the repository should have a LICENSE file, or the source repository should have a top-level LICENSE file. If the repository is under multiple licenses, then it is encouraged (but not required) to have one LICENSE. file per license.

bmagyar:
> For packages that are brand-new to the ROS ecosystem, some due diligence will be done on the source and release repositories: 
The first ROS release control_toolbox was featured in was ROS Hydro. It's been used in a fairly similar state ever since. 
As I said before, the exact same version was released to Foxy and Galactic roughly at the same time as this PR, both merged and out since then. Nobody complained about the lack of license.

Considering this, I'd argue that applying this rule now to this package is a wee bit misplaced. This PR is currently holding up releasing `ros2_control`, `ros2_controllers` and `moveit2` to Rolling.

gbalke:
While it may have been missed in prior releases, it is our current guidelines. If you could add a `LICENSE` in the development branch so we know it will be available in future releases, we can merge this release as-is.

bmagyar:
As I said before, this package is *very far* from being anything new to the ecosystem.

tfoote (approved):
@gbalke and @ahcorde The reviewing guidelines only actually requires the LICENSE file for packages that are "brand-new to the ROS ecosystem" which this definitely is not. There's a large codebase of packages with a legacy from before a LICENSE file became common standard. And we don't to block all their releases on this nice to have. We enforce it for new packages to improve our status quo.

@bmagyar Although this package has been effectively grandfathered in. I'd strongly encourage adding the LICENSE file when you get a moment. Looking at the package.xml it's only BSD so it hopefully won't be a significant overhead to copy one instance of the BSD header from a file into a new file named LICENSE. Or from: https://opensource.org/licenses/BSD-3-Clause excluding the copyright line.

PR #29822, opened 2021-06-09
rosdep: add python3-pydbus package

dandedrick:

## Package name:

python3-pydbus

## Package Upstream Source:

https://github.com/LEW21/pydbus

## Purpose of using this:

This provides a python3 version of the already provided python-pydbus,
which is useful for interacting with dbus for things like systemd unit
manipulation.

  - https://packages.debian.org/buster/python3-pydbus
   - https://packages.ubuntu.com/focal/python3-pydbus
  - https://src.fedoraproject.org/rpms/python-pydbus

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7044,6 +7044,14 @@ python3-pydantic:
       pip: [pydantic]
     xenial:
       pip: [pydantic]
+python3-pydbus:
+  debian: [python3-pydbus]
+  fedora: [python3-pydbus]
+  opensuse: [python3-pydbus]
+  rhel:
+    '*': ['python%{python3_pkgversion}-pydbus']
+    '7': null
+  ubuntu: [python3-pydbus]
 python3-pydot:
   alpine: [py3-pydot]
   arch: [python-pydot]


gbalke (approved):

jacobperron (commented):
Please add entries for [opensuse](https://software.opensuse.org/package/python3-pydbus) and [centos](https://centos.pkgs.org/8/centos-appstream-x86_64/python3-pydbus-0.6.0-5.el8.noarch.rpm.html)

dandedrick:
I added opensuse and RHEL. Though for opensuse the appropriate packages was python3-pydbus, since the indicated python-dbus is the python2 version, which is not what we want, and for RHEL python3-pydbus isn't present on RHEL 7 so it has to be limited to 8.

rosdep/python.yaml
```diff
@@ -7044,6 +7044,13 @@ python3-pydantic:
       pip: [pydantic]
     xenial:
       pip: [pydantic]
+python3-pydbus:
+  debian: [python3-pydbus]
+  fedora: [python3-pydbus]
+  opensuse: [python3-pydbus]
+  rhel:
+    '8': [python3-pydbus]
```
jacobperron, line 7052: For future proofing, I think we could write:
```suggestion
    '*': [python3-pydbus]
    '7': null
```

@wjwwood Can you double-check?

dandedrick: Looking at other rhel examples I found that they use `python%{python3_pkgversion}-` and that lead me down the path to find that for rhel7 this package is available but with the name `python36-pydbus`. I think that setting this as `python%{python3_pkgversion}-pydbus` for all versions might make this work.

dandedrick: It seems that wasn't actually the case. python36-pydbus doesn't actually exist for RHEL 7. I've updated it to the commended null for 7.

jacobperron (approved):

PR #29909, opened 2021-06-15
Add pip specific content contribution rosdep rules for Gentoo

### LINKED ISSUE #29908: Extend Gentoo CONTRIBUTING guides to clarify about pip rules
This should be codified in the [CONTRIBUTING guide with respect to gentoo](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#gentoo) so that it's clear to contributors. As well as adding a comment/reference in the REVIEW_GUIDELINES to make it clear for the reviewers too.

https://github.com/ros/rosdistro/pull/29514#discussion_r636348684

@allenh1 Would you have a moment to suggest verbiage along the lines above so we don't have to ping you or find old references in cases like: https://github.com/ros/rosdistro/pull/29587

tfoote:
Fixes: #29908

@allenh1 From your suggestion.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -25,7 +25,7 @@ Distro packaging links:
 
 ## Links to Distribution Packages
 
-<!-- Replace the REQUIRED areas and state not available for IF AVAILABLE 
+<!-- Replace the REQUIRED areas with the URL to the package.  For IF AVAILABLE areas, either put in the URL to the package or state 'not available'.
 More info at https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules -->
 
 - Debian: https://packages.debian.org/
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -25,7 +25,8 @@ Distro packaging links:
 
 ## Links to Distribution Packages
 
-<!-- Replace the REQUIRED areas and state not available for IF AVAILABLE -->
+<!-- Replace the REQUIRED areas and state not available for IF AVAILABLE 
+More info at https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules -->
 
 - Debian: https://packages.debian.org/
   - REQUIRED
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,6 +124,9 @@ TODO
 If the ebuild you are referencing is not in either of those locations, please
 file a PR into ROS-Overlay to add it and any needed dependencies to the tree.
 
+Note that `pip` cannot be used to install system packages on Gentoo.
+If the package is not present in the Gentoo package repository, please create an issue on the ros/ros-overlay repository so that it may be created at a later date, omit the key from the new rosdep rule, and link the created issue in the PR which is adding the rule.
+
 #### Arch Linux
 
 Packages must be in the official Archlinux core, extra, or community repositories at the time they are contributed.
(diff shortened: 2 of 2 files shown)


allenh1:
lgtm

.github/pull_request_template.md
```diff
@@ -25,7 +25,8 @@ Distro packaging links:
 
 ## Links to Distribution Packages
 
-<!-- Replace the REQUIRED areas and state not available for IF AVAILABLE -->
+<!-- Replace the REQUIRED areas and state not available for IF AVAILABLE
```
clalancette, line 28: I find this sentence confusing.  How about:

```suggestion
<!-- Replace the REQUIRED areas with the URL to the package.  For IF AVAILABLE areas, either put in the URL to the package or state 'not available'.
```

CONTRIBUTING.md
```diff
@@ -124,6 +124,9 @@ TODO
 If the ebuild you are referencing is not in either of those locations, please
 file a PR into ROS-Overlay to add it and any needed dependencies to the tree.
 
+Note that `pip` cannot be used to install system packages on Gentoo.
+If the package is not present in the Gentoo package repository, please create an issue on the ros/ros-overlay repository so that it may be created at a later date, omit the key from the new rosdep rule, and link the created issue in the PR which is adding the rule.
```
clalancette, line 128:
```suggestion
If the package is not present in the Gentoo package repository, please create an issue on the https://github.com/ros/ros-overlay repository so that it may be created at a later date, omit the key from the new rosdep rule, and link the created issue in the PR which is adding the rule.
```

clalancette (changes requested):
A few minor rewording changes.

clalancette (approved):

CONTRIBUTING.md
```diff
@@ -124,6 +124,9 @@ TODO
 If the ebuild you are referencing is not in either of those locations, please
 file a PR into ROS-Overlay to add it and any needed dependencies to the tree.
 
+Note that `pip` cannot be used to install system packages on Gentoo.
+If the package is not present in the Gentoo package repository, please create an issue on the https://github.com/ros/ros-overlay repository so that it may be created at a later date, omit the key from the new rosdep rule, and link the created issue in the PR which is adding the rule.
```
jacobperron, line 128: It might be good to reference an example ([for example](https://github.com/ros/ros-overlay/issues/1019)).

jacobperron (approved):

mabelzhang:
Did this PR also mean to add a note about pip in the REVIEW_GUIDELINES (line 63)? Maybe something like, for Gentoo entries, check that it does not use pip, and if it is, refer to CONTRIBUTING.

Related, but not to this PR, I did a search in rosdep/python.yaml and found that we do have like a dozen gentoo+pip entries, like https://github.com/ros/rosdistro/pull/29426/files . I don't know if we can/want to do anything about them.

tfoote:
> Did this PR also mean to add a note about pip in the REVIEW_GUIDELINES (line 63)? Maybe something like, for Gentoo entries, check that it does not use pip, and if it does, refer to CONTRIBUTING.

I initially thought to do that. But then I realized that all the distro related rules are here in the CONTRIBUTING document and that these rules are directly linked in the REVIEWING section. There's several other rules per distro that are at the same level. This one is new but doesn't seem to have any distinction compared to all the others. And duplicating the requirements in a second location I think will lead to more maintenance or inconsistency in statements. 

> Related, but not to this PR, I did a search in rosdep/python.yaml and found that we do have like a dozen gentoo+pip entries, like https://github.com/ros/rosdistro/pull/29426/files . I don't know if we can/want to do anything about them.

A PR to clean them up/remove them would be great, it's mostly a matter of having the time to clean it up. We can definitely remove them as they're not expected to work and as such won't break people to remove them and we don't need a sunsetting process.

mabelzhang:
I could do that PR. But does that mean we also want to create ros/ros-overlay tickets for all of them?

CONTRIBUTING.md
tfoote:
```suggestion
If the package is not present in the Gentoo package repository, please create an issue on the https://github.com/ros/ros-overlay repository so that it may be created at a later date, omit the key from the new rosdep rule, and link the created issue in the PR which is adding the rule. [Here's a simple example](https://github.com/ros/ros-overlay/issues/1019)
```

tfoote:
> I could do that PR. But does that mean we also want to create ros/ros-overlay tickets for all of them?

Great. Ticketing it over there too would be great. To be quicker maybe just a single ticket with each of the dependencies you cleared as a checkbox instead of opening a bunch of tickets.

PR #29913, opened 2021-06-15
add verbiage about overriding 3rdparty content

tfoote:
The majority of the content is from: https://discourse.ros.org/t/release-package-requiring-custom-version-of-system-dependency/18200/6?u=tfoote

Following up: https://github.com/ros/rosdistro/pull/29614#issuecomment-850466694

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,7 +76,8 @@ This will allow people to opt into the potentially disruptive change.
 An example of this is Ubuntu’s PPAs or just instructions for end users to compile from source.
 
 In the case that a package is released into a ROS distribution that is then later available in upstream system distributions when the ROS distribution rolls forward the ROS packages should be removed in favor of relying on the system version.
-To provide a gentle transition a placeholder package can be provided which will just depend on the rosdep key but this should be considered a deprecated package and dependent packages should switch to use the rosdep key.
+To provide a gentle transition, a placeholder package can be provided that only depends on the rosdep key.
+This should be considered a deprecated package and dependent packages should switch to use the rosdep key.
 This transitional package is likely something that should be scoped to be released into rolling only and not be in a subsequent released distribution.
 It is up to the judgement of the ROS Distro's release manager to make an exception.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,7 +50,7 @@ Or they may take [other actions](https://github.com/ros-infrastructure/ros_build
 
 In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
 This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
-If you truly need a newer versions of underlying libraries then you can consider targeting a newer ROS distributionthat targets a newer platform which hopefully will have a higher version of your necessary dependency.
+If you truly need newer versions of underlying libraries then you can consider targeting a newer ROS distributionthat targets a newer platform which hopefully will have a higher version of your necessary dependency.
 If the necessary version of your dependency isn't available even on the latest platforms, please reach out and encourage the upstream operating systems to move forward to at least your minimum version for the next release of the upstream distributions.
 By doing this you will make sure that it's available for everyone in the future.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,7 +50,9 @@ Or they may take [other actions](https://github.com/ros-infrastructure/ros_build
 
 In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
 This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
-If you need newer versions of underlying libraries the you should either target newer versions with the minimum version available or if that’s not available help and encourage the upstream process to move forward to at least your minimum version for the next release of the distribution.
+If you truly need a newer versions of underlying libraries the you can consider targeting a newer rosdistro with which will target newer platforms which hopefully will have a higher version of your necessary dependency.
+If the necessary version of your dependency isn't available even on the latest platforms, please reach out  and encourage the upstream platforms to move forward to at least your minimum version for the next release of the upstream distributions.
+By doing this you will make sure that it's available for everyone in the future.
 
 Obviously this is a slow process and doesn’t fix things immediately.
 There are workarounds such as embedding copies of the libraries into your package or making another package with the newer library version.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ If you need newer versions of underlying libraries the you should either target 
 Obviously this is a slow process and doesn’t fix things immediately.
 There are workarounds such as embedding copies of the libraries into your package or making another package with the newer library version.
 This usually can be made to work well in your local development environment.
-However if you want to release this style of workarounds you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
+However if you want to release this style of workarounds into a ROS distribution you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
 
 To release a ROS package which will overlay a system dependency it must:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,19 +50,19 @@ Or they may take [other actions](https://github.com/ros-infrastructure/ros_build
 
 In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
 This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
-If you truly need a newer versions of underlying libraries the you can consider targeting a newer rosdistro with which will target newer platforms which hopefully will have a higher version of your necessary dependency.
-If the necessary version of your dependency isn't available even on the latest platforms, please reach out  and encourage the upstream platforms to move forward to at least your minimum version for the next release of the upstream distributions.
+If you truly need a newer versions of underlying libraries then you can consider targeting a newer ROS distributionthat targets a newer platform which hopefully will have a higher version of your necessary dependency.
+If the necessary version of your dependency isn't available even on the latest platforms, please reach out and encourage the upstream operating systems to move forward to at least your minimum version for the next release of the upstream distributions.
 By doing this you will make sure that it's available for everyone in the future.
 
 Obviously this is a slow process and doesn’t fix things immediately.
 There are workarounds such as embedding copies of the libraries into your package or making another package with the newer library version.
 This usually can be made to work well in your local development environment.
-However if you want to release this style of workarounds into a ROS distribution you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
+However if you want to release this style of workaround into a ROS distribution you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
 
 To release a ROS package which will overlay a system dependency it must:
 
 * Install side by side with the system one, with a different name.
-* Not cause compile errors for any packages including downstream ones
+* Not cause compile errors for any packages including downstream ones.
 * Not cause linking errors or crashes at runtime for yours or any downstream packages, including ones that use the system version on any of our target platforms.
 
 The release pipeline will only catch the first issue automatically.
@@ -72,7 +72,7 @@ But in general if you want to do this without being disruptive you need to:
 2. Change all namespaces to avoid symbol collisions
 3. Change header paths to not collide with the system ones if you export the include directories
 
-Embedding a copy inside your package, potentially statically linking a version into your executables is also a potential way to do it as well.
+Embedding a copy inside your package, potentially statically linking a version into your executables is another way to do it as well.
 It’s going to require a non-trivial amount of engineering and modifications to make this possible and most people decide it’s not worth it.
 
 In many situations a common approach for this is to not actually release it onto the public distro and distribute it separately through a side channel.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ In general you should not override or replace system dependencies from the syste
 This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
 If you need newer versions of underlying libraries the you should either target newer versions with the minimum version available or if that’s not available help and encourage the upstream process to move forward to at least your minimum version for the next release of the distribution.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -34,6 +34,37 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party packages into a ROS distribution
+
+In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
+This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent withing the distribution.
+If you need newer versions of underlying libraries the you should either target newer versions with the minimum version available or if that’s not available help and encourage the upstream process to move forward to at least your minimum version for the next release of the distribution.
+
+Obviously this is a slow process and doesn’t fix things immediately for a short term there are workarounds, however if you want to release packages which build on the workarounds you must make sure not to break other users who are expecting the system version.
```
clalancette, line 43:
```suggestion
Obviously this is a slow process and doesn’t fix things immediately for the short term.  If you install the updated package locally and introduce workarounds into your package, you must make sure not to break other users who are expecting the system version when you release the package.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,37 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
+* Install side by side with the system one, with a different name.
+* Not cause compile errors for any packages including downstream ones
+* Not cause linking errors or crashes at runtime for yours or any downstream packages, including ones that use the system version on any of our target platforms.
+
+The release pipeline will only catch the first issue automatically.
+Manual review is required for the latter two.
+But in general if you want to do this without being disruptive you need to adjust paths to avoid installation collisions.
+Change all namespaces to avoid symbol collisions. And change header paths to not collide with the system ones if you export the include directories.
+
+Embedding a copy inside your package, potentially a statically linking a version into your executables is also a potential way to do it as well.
```
clalancette, line 56:
```suggestion
Embedding a copy inside your package, potentially statically linking a version into your executables is also a potential way to do it as well.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,37 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
 If a package continuously fails to build, the ROS Boss for that distribution may choose to revert the pull request that introduced it.
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party packages into a ROS distribution
```
clalancette, line 37:
```suggestion
### Releasing third party system packages into a ROS distribution
```

clalancette (changes requested):
Minor wording changes.

CONTRIBUTING.md
tfoote: Hmm, clearly not a properly formed sentence. Here's a slightly more complete suggestion.
```suggestion
Obviously this is a slow process and doesn’t fix things immediately for a short term.
There are workarounds such as embedding copies of the libraries into your package or making another package with the newer library version.
This usually can be made to work well in your local development environment.
However if you want to release this style of workarounds you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,37 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
 If a package continuously fails to build, the ROS Boss for that distribution may choose to revert the pull request that introduced it.
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party system packages into a ROS distribution
+
+In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
+This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent withing the distribution.
```
jacobperron, line 40:
```suggestion
This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,37 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
+
+Embedding a copy inside your package, potentially statically linking a version into your executables is also a potential way to do it as well.
+It’s going to require a non-trivial amount of engineering and modifications to make this possible and most people decide it’s not worth it.
+
+In many situations a common approach for this is to not actually release it onto the public distro and distribute it separately through a side channel.
+This will allow people to opt into the potentially disruptive change.
+An example of this is Ubuntu’s PPAs or just instructions for end users to compile from source.
+
+In the case that a package is released into a ROS distribution that is then later available in upstream system distributions when the ROS distribution rolls forward the ROS packages should be removed in favor of relying on the system version.
+To provide a gentle transition a placeholder package can be provided which will just depend on the rosdep key but this should be considered a deprecated package and dependent packages should switch to use the rosdep key.
```
jacobperron, line 64: minor wording suggesting
```suggestion
To provide a gentle transition, a placeholder package can be provided that only depends on the rosdep key.
This should be considered a deprecated package and dependent packages should switch to use the rosdep key.
```

jacobperron (approved):

CONTRIBUTING.md
```diff
@@ -34,6 +34,41 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
 If a package continuously fails to build, the ROS Boss for that distribution may choose to revert the pull request that introduced it.
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party system packages into a ROS distribution
+
+In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
+This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
+If you need newer versions of underlying libraries the you should either target newer versions with the minimum version available or if that’s not available help and encourage the upstream process to move forward to at least your minimum version for the next release of the distribution.
```
clalancette, line 41:
```suggestion
If you need newer versions of underlying libraries then you should either target newer versions with the minimum version available or if that’s not available help and encourage the upstream process to move forward to at least your minimum version for the next release of the distribution.
```

Besides the typo, actually, what are we trying to say here?  That is, what does "target newer versions with the minimum version available mean"?  That your ROS package should only depend on the minimum version from the OS?  Something else?

CONTRIBUTING.md
```diff
@@ -34,6 +34,41 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
+
+Obviously this is a slow process and doesn’t fix things immediately for a short term.
+There are workarounds such as embedding copies of the libraries into your package or making another package with the newer library version.
+This usually can be made to work well in your local development environment.
+However if you want to release this style of workarounds you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
```
clalancette, line 46:
```suggestion
However if you want to release this style of workarounds into a ROS distribution you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,41 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
+To release a ROS package which will overlay a system dependency it must:
+
+* Install side by side with the system one, with a different name.
+* Not cause compile errors for any packages including downstream ones
+* Not cause linking errors or crashes at runtime for yours or any downstream packages, including ones that use the system version on any of our target platforms.
+
+The release pipeline will only catch the first issue automatically.
+Manual review is required for the latter two.
+But in general if you want to do this without being disruptive you need to adjust paths to avoid installation collisions.
+Change all namespaces to avoid symbol collisions. And change header paths to not collide with the system ones if you export the include directories.
```
clalancette, line 57:
```suggestion
But in general if you want to do this without being disruptive you need to:
1. Adjust paths to avoid installation collisions.
2. Change all namespaces to avoid symbol collisions
3. Change header paths to not collide with the system ones if you export the include directories
```

clalancette (changes requested):
A few more things I noticed while going through this.

CONTRIBUTING.md
```diff
@@ -34,6 +34,41 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
 If a package continuously fails to build, the ROS Boss for that distribution may choose to revert the pull request that introduced it.
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party system packages into a ROS distribution
+
+In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
+This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
+If you need newer versions of underlying libraries the you should either target newer versions with the minimum version available or if that’s not available help and encourage the upstream process to move forward to at least your minimum version for the next release of the distribution.
+
+Obviously this is a slow process and doesn’t fix things immediately for a short term.
```
nuclearsandwich, line 43: The end of this sentence seems redundant, I'd suggest either 

```suggestion
Obviously this is a slow process and doesn’t fix things immediately.
```

```suggestion
Obviously this is a slow process and doesn’t fix things in the short term.
```

nuclearsandwich (commented):
Overall this is a great addition to our contributing docs and it's going to help contributors and reviewers alike. I've added some suggestions and there are a few in flight from other reviewers. Otherwise it is looking good!

ivanpauno (approved):
LGTM, with clalancette and nuclearsandwich comments addressed.

CONTRIBUTING.md
```diff
@@ -34,6 +34,45 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
 If a package continuously fails to build, the ROS Boss for that distribution may choose to revert the pull request that introduced it.
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party system packages into a ROS distribution
+
+In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
+This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
+If you truly need a newer versions of underlying libraries the you can consider targeting a newer rosdistro with which will target newer platforms which hopefully will have a higher version of your necessary dependency.
```
clalancette, line 41:
```suggestion
If you truly need a newer versions of underlying libraries then you can consider targeting a newer ROS distributionthat targets a newer platform which hopefully will have a higher version of your necessary dependency.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,45 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
+If the necessary version of your dependency isn't available even on the latest platforms, please reach out  and encourage the upstream platforms to move forward to at least your minimum version for the next release of the upstream distributions.
```
clalancette, line 42:
```suggestion
If the necessary version of your dependency isn't available even on the latest platforms, please reach out and encourage the upstream operating systems to move forward to at least your minimum version for the next release of the upstream distributions.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,45 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
+By doing this you will make sure that it's available for everyone in the future.
+
+Obviously this is a slow process and doesn’t fix things immediately.
+There are workarounds such as embedding copies of the libraries into your package or making another package with the newer library version.
+This usually can be made to work well in your local development environment.
+However if you want to release this style of workarounds into a ROS distribution you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
```
clalancette, line 48:
```suggestion
However if you want to release this style of workaround into a ROS distribution you must make sure not to break other users who are expecting the system version either in downstream packages or elsewhere on the system.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,45 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
...
+
+To release a ROS package which will overlay a system dependency it must:
+
+* Install side by side with the system one, with a different name.
+* Not cause compile errors for any packages including downstream ones
```
clalancette, line 53:
```suggestion
* Not cause compile errors for any packages including downstream ones.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,45 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
 If a package continuously fails to build, the ROS Boss for that distribution may choose to revert the pull request that introduced it.
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party system packages into a ROS distribution
+
+In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
+This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
+If you truly need a newer versions of underlying libraries then you can consider targeting a newer ROS distributionthat targets a newer platform which hopefully will have a higher version of your necessary dependency.
```
clalancette, line 41:
```suggestion
If you truly need newer versions of underlying libraries then you can consider targeting a newer ROS distributionthat targets a newer platform which hopefully will have a higher version of your necessary dependency.
```

CONTRIBUTING.md
```diff
@@ -34,6 +34,45 @@ Additionally, it may be necessary to determine build failures, these jobs are lo
 If a package continuously fails to build, the ROS Boss for that distribution may choose to revert the pull request that introduced it.
 Or they may take [other actions](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst) to avoid repeated failures.
 
+### Releasing third party system packages into a ROS distribution
+
+In general you should not override or replace system dependencies from the system distribution with a package in the ROS distribution.
+This is part of the software engineering process of maintaining the distribution that all developers can rely on specific versions of software to be available and remain consistent within the distribution.
+If you truly need newer versions of underlying libraries then you can consider targeting a newer ROS distributionthat targets a newer platform which hopefully will have a higher version of your necessary dependency.
```
clalancette, line 41:
```suggestion
If you truly need newer versions of underlying libraries then you can consider targeting a newer ROS distribution that targets a newer platform which hopefully will have a higher version of your necessary dependency.
```

clalancette (approved):

PR #29991, opened 2021-06-22
rmf_battery: 0.1.0-2 in 'rolling/distribution.yaml' [bloom]

marcoag:
Increasing version of package(s) in repository `rmf_battery` to `0.1.0-2`:

- upstream repository: https://github.com/open-rmf/rmf_battery.git
- release repository: https://github.com/ros2-gbp/rmf_battery-release.git
- bloom version: `0.10.7`
- previous version for package: `0.1.0-1`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2189,7 +2189,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/open-rmf/rmf_battery.git
-      version: rolling
+      version: main
     release:
       tags:
         release: release/rolling/{package}/{version}
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2194,7 +2194,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rmf_battery-release.git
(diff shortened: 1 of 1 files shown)


rolling/distribution.yaml
```diff
@@ -2189,12 +2189,12 @@ repositories:
...
       type: git
       url: https://github.com/open-rmf/rmf_battery.git
-      version: rolling
+      version: main
     release:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rmf_battery-release.git
-      version: 0.1.0-1
+      version: 0.1.0-2
```
wjwwood, line 2197: Is this the version you expect? This kind of bump usually only happens when you re-run bloom but no new upstream version exists. It could be what you meant to do, just double checking.

marcoag: Thanks for the double check. Yes, initially we were using `rolling` branches for the `rolling` distribution but in the end using directly `main` makes more sense so we are slowly moving them all to `main`.

nuclearsandwich: @marcoag in the future you can open PRs to change the doc and source branches without bloom and skip the bloom revision.

PR #30011, opened 2021-06-23
Explicitly enable yamllint trailing spaces option.

nuclearsandwich:
When investigating #30010 I realized that this is something yamllint should also be catching, which would have helped add corroborating signal to offset the noise in the nosetests.

Looking at the [yamllint docs](https://yamllint.readthedocs.io/en/stable/configuration.html), trailing whitespace checks should be enabled by default but aren't behaving that way in 1.26.1. This explicitly enables trailing whitespace checks which might be a good idea while I figure out what's happening in yamllint and share my discovery upstream.

### DIFF
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -48,4 +48,4 @@ jobs:
         python -m pip install --upgrade pip setuptools wheel
         pip install yamllint
     - name: Run yamllint
-      run: yamllint */
+      run: "yamllint -d'rules: { trailing-spaces: enable}' */"
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -48,4 +48,4 @@ jobs:
         python -m pip install --upgrade pip setuptools wheel
         pip install yamllint
     - name: Run yamllint
-      run: "yamllint -d'rules: { trailing-spaces: enable}' */"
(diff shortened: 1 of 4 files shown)


nuclearsandwich:
So it's not _just_ the trailing-spaces rule but also several others, however via experimentation I was able to determine that key-duplicates is at least enabled by default, however it's then disabled by my cli config above and so this PR will need to change.

nuclearsandwich:
Okay after a lot of time being _VERY_ confused. The problem appears to be the `.yamllint` configuration file in this repository. It seems to be disabling by virtue of omission quite a few rules that are probably valueable and this PR should be amended to update the .yamllint configuration.

nuclearsandwich:
Okay, I did some more fooling around and figured out that the .yamllint config we had was not doing much because it was not explicitly extending the default yamllint configuration. I've put in a new yamllint config which does extend the default and then tuned it to match our conventions. I've also set it to ignore certain rules for contents which do not meet the requirements currently. We can discuss if it's best to ignore those rules long term or update the contents to conform.

For the old releases/ files I think we can leave them be but there are some rosdep definition lines that are ~500 chars long and it may make sense to split those up for readability.

clalancette (approved):

hidmic (approved):

PR #30034, opened 2021-06-26
add python3-fcn-pip key

knorth55:
I add `python3-fcn-pip` key for python3.
I follow the instruction in https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1 and use anchor/alias for yaml.

pypi URL: https://pypi.org/project/fcn/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1620,7 +1620,7 @@ python-fcl-pip:
     pip:
       depends: [libfcl-dev]
       packages: [python-fcl]
-python-fcn-pip:
+python-fcn-pip: &migrate_eol_2025_04_30_python3_fcn_pip
   debian:
     pip:
       depends: [liblapack-dev]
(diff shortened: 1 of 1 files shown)


knorth55:
I read the discussion in #28891, but maybe I don't understand it correctly.
We should use the anchor and alias style and avoid copy&pasting the `python-*-pip` field, is my understanding correct?

nuclearsandwich:
> We should use the anchor and alias style and avoid copy&pasting the `python-*-pip` field, is my understanding correct?

As far as I can see your changes implement the anchoring correctly. The objectives that we are balancing are: not removing existing `python-*-pip` keys until the last python2-supporting distributions are no longer supported, not duplicating the definitions between `python-*-pip` and `python3-*-pip` keys, and being minimally disruptive to the existing rosdep documents. So adding the anchor to the `python-*-pip` and the alias to `python3-*-pip` is correct. Thanks.

nuclearsandwich (approved):

knorth55:
@nuclearsandwich thank you for your review and answer. That is a relief, and I'm glad that I'm the first person who uses the anchor style in `rosdistro/python.yaml` :)

clalancette (approved):

PR #30056, opened 2021-06-29
Update python-gdal key for focal and bullseye.

nuclearsandwich:
Future Debian derived distributions do not have Python 2 (python-)
packages for python-gdal. This approach will cause key errors on older
unsupported platforms but is forwards compatible vs a '*'-and-null
approach for undefining the key explicitly for new platforms.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1809,10 +1809,14 @@ python-gcloud-pip:
     pip:
       packages: [gcloud]
 python-gdal:
-  debian: [python-gdal]
+  debian:
+    buster: [python-gdal]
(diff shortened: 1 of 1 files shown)


mabelzhang (approved):
Linking to packages for confirmation
Debian stretch & buster: https://packages.debian.org/stretch/python-gdal
Ubuntu bionic: https://packages.ubuntu.com/search?keywords=python-gdal&searchon=names
Ubuntu doesn't list xenial anymore as it's out of support, but some searching found that the package does exist for xenial.

clalancette (approved):

PR #30153, opened 2021-07-11
tf2_2d: 0.6.1-1 in 'melodic/distribution.yaml' [bloom]

ayrton04:
Increasing version of package(s) in repository `tf2_2d` to `0.6.1-1`:

- upstream repository: https://github.com/locusrobotics/tf2_2d.git
- release repository: https://github.com/locusrobotics/tf2_2d-release.git
- bloom version: `0.10.7`

## tf2_2d

```
* Contributors: Stephen Williams
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13070,6 +13070,18 @@ repositories:
       url: https://github.com/Terabee/teraranger_array.git
       version: master
     status: maintained
+  tf2_2d:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/locusrobotics/tf2_2d-release.git
+      version: 0.6.1-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/locusrobotics/tf2_2d.git
+      version: devel
+    status: developed
   tf2_server:
     doc:
       type: git


ayrton04:
Question: we have another package that we are going to release that depends on this one. Once the build farm builds this package, will we be able to release the other package, or do we have to wait for a sync? I can't recall whether the ROS buildfarm gets its dependencies from the synced packages or the ones that are waiting to be synced.

clalancette:
> Question: we have another package that we are going to release that depends on this one. Once the build farm builds this package, will we be able to release the other package, or do we have to wait for a sync? I can't recall whether the ROS buildfarm gets its dependencies from the synced packages or the ones that are waiting to be synced.

Once it successfully builds on the buildfarm, you can release a package that depends on this one.  No need to wait for a sync.

ivanpauno:
The repo has a top level BSD licence, but the tag in the package.xml says something else?
https://github.com/locusrobotics/tf2_2d/blob/323015b6c50f357bfa494fdad38fe3723ec2976e/package.xml#L12

Could you fix that?

ayrton04:
Yes, yes I can. Sorry about that. That package was internal before we released it. Should I just fix and bloom, and then update this PR with the new version number?

ivanpauno:
> Yes, yes I can. Sorry about that. That package was internal before we released it. Should I just fix and bloom, and then update this PR with the new version number?

that sounds fine, but you can also open a new PR with bloom directly

ayrton04:
Actually, we should either merge this and I can immediately patch it, or we can close this PR and I can submit a new PR. This package isn't in the distro yet, so the new PR would have the entire stanza in it. Is the latter preferable?

ivanpauno:
> Actually, we should either merge this and I can immediately patch it, or we can close this PR and I can submit a new PR. This package isn't in the distro yet, so the new PR would have the entire stanza in it. Is the latter preferable?

Last option sounds good to me.

ayrton04:
https://github.com/ros/rosdistro/pull/30181

PR #30344, opened 2021-07-28
Revert rosidl typesupport dependency export releases.

nuclearsandwich:
Reverts the following releases related to the export of rosidl typesupport
dependencies since more discussion is required to determine how best to
make these work correctly with build farm releases.

* https://github.com/ros/rosdistro/pull/30321
* https://github.com/ros/rosdistro/pull/30343

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2978,7 +2978,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rosidl-release.git
-      version: 2.4.0-1
+      version: 2.3.0-1
     source:
       test_pull_requests: true
       type: git
@@ -3068,7 +3068,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rosidl_typesupport-release.git
-      version: 1.4.0-1
+      version: 1.3.0-1
     source:
       test_pull_requests: true
       type: git
@@ -3088,7 +3088,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rosidl_typesupport_fastrtps-release.git
-      version: 2.0.1-1
+      version: 1.2.1-1
     source:
       test_pull_requests: true
       type: git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2978,7 +2978,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rosidl-release.git
-      version: 2.3.0-1
+      version: 2.4.0-1
     source:
       test_pull_requests: true
       type: git
@@ -3068,7 +3068,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rosidl_typesupport-release.git
-      version: 1.3.0-1
+      version: 1.4.0-1
     source:
       test_pull_requests: true
       type: git


rolling/distribution.yaml
```diff
@@ -3068,7 +3068,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rosidl_typesupport-release.git
-      version: 1.4.0-1
+      version: 1.3.0-1
```
tfoote, line 3071: This change is already in main and cannot be just rolled back. It needs a higher version released with the old content.

rolling/distribution.yaml
```diff
@@ -2978,7 +2978,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/rosidl-release.git
-      version: 2.4.0-1
+      version: 2.3.0-1
```
tfoote, line 2981: This one is also in main and consequently can't be rolled back.

tfoote (changes requested):
This change does roll those elements back. But 2 of the 3 packages have already synced to main. They consequently cannot be rolled back, they need a higher versioned release so that people get the rolled back version.

rolling/distribution.yaml
nuclearsandwich: If these releases are in main then I'd just as soon not revert them at all since they're clearly not causing the regressions.

hidmic: Hmm, let's have a look at those debians. Dependency exportation changes were meant to be used together.

nuclearsandwich:
I removed the reverts that are already in Rolling's main repository. @hidmic what's the best way to validate that the debs currently in main are usable?

I'm currently of the opinion that having packages which _mostly_ work in Rolling's testing repository is better than having no packages in Rolling's testing repository so depending on the impact of leaving these release unreverted but functional (compared with the non-functional `rosidl_typesupport_fastrtps` release) seems like the best short-term correction to me.

hidmic:
> I removed the reverts that are already in Rolling's main repository. @hidmic what's the best way to validate that the debs currently in main are usable?

Downloading and checking debians of packages that bear interfaces. 

I just took a look at `ros-rolling-builtin-interfaces` (version `1.0.3-1`) in `main`. It exports all targets but `rosidl_typesupport_fastrtps_c*` related ones, as expected:

```bash
$ more opt/ros/rolling/share/builtin_interfaces/cmake/builtin_interfacesExport.cmake 
# ...
foreach(_expectedTarget builtin_interfaces::builtin_interfaces__rosidl_generator_c builtin_interfaces::builtin_interfaces_
_rosidl_typesupport_introspection_c builtin_interfaces::builtin_interfaces__rosidl_typesupport_c builtin_interfaces::built
in_interfaces__rosidl_generator_cpp builtin_interfaces::builtin_interfaces__rosidl_typesupport_introspection_cpp builtin_i
nterfaces::builtin_interfaces__rosidl_typesupport_cpp)
# ...
```

Most dependencies are satisfied through `ros-rolling-rosidl-default-runtime`:

```bash
$ apt-cache depends ros-rolling-rosidl-default-runtime
ros-rolling-rosidl-default-runtime
  Depends: ros-rolling-rosidl-generator-py
  Depends: ros-rolling-rosidl-runtime-c
  Depends: ros-rolling-rosidl-runtime-cpp
  Depends: ros-rolling-rosidl-typesupport-c
  Depends: ros-rolling-rosidl-typesupport-cpp
  Depends: ros-rolling-rosidl-typesupport-introspection-c
  Depends: ros-rolling-rosidl-typesupport-introspection-cpp
  Depends: ros-rolling-ros-workspace
```

We're good. Let's revert that `rosidl_typesupport_fastrtps` release only.

tfoote (approved):
lgtm with the reduced scope.

PR #30345, opened 2021-07-28
[Python] Voluptous Python 3

mr337:

## Package name:

Voluptous for python3, update to https://github.com/ros/rosdistro/pull/17079

## Package Upstream Source:

https://github.com/alecthomas/voluptuous

## Purpose of using this:

PR to add the JSON validation library Voluptuous. We use this to validate JSON commands coming from external sources (laptop, cloud etc).

Ubuntu: https://packages.ubuntu.com/focal/python3-voluptuous
Fedora: https://src.fedoraproject.org/rpms/python-voluptuous
Debian: https://packages.debian.org/buster/python-voluptuous
Gentoo: https://packages.gentoo.org/packages/dev-python/voluptuous

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5449,6 +5449,11 @@ python-voluptuous:
   gentoo: [dev-python/voluptuous]
   nixos: [pythonPackages.voluptuous]
   ubuntu: [python-voluptuous]
+python3-voluptuous:
+  debian: [python-voluptuous]
+  fedora: [python-voluptuous]
+  gentoo: [dev-python/voluptuous]
+  ubuntu: [python3-voluptuous]
 python-vtk:
   arch: [vtk]
   debian: [python-vtk]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8041,8 +8041,8 @@ python3-virtualserialports-pip:
     pip:
       packages: [PyVirtualSerialPorts]
 python3-voluptuous:
-  debian: [python-voluptuous]
-  fedora: [python-voluptuous]
+  debian: [python3-voluptuous]
+  fedora: [python3-voluptuous]
   gentoo: [dev-python/voluptuous]
   opensuse: [python-voluptuous]
   rhel:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8045,7 +8045,9 @@ python3-voluptuous:
   fedora: [python-voluptuous]
   gentoo: [dev-python/voluptuous]
   opensuse: [python-voluptuous]
-  rhel: [python3-voluptuous]
+  rhel:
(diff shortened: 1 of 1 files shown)


mr337:
I am not fully clear if this is the best way to approach the python3 upgrade. For ubuntu there is a distinct `python-voluptous` and `python3-voluptous`, but for the other distros they ship the updated version which is compatible with python2 and python3. Happy to make any corrections to make this more sane.

gbalke:
It appears there's a `rhel` source as well. It would be great if you could add that as well! Otherwise looks good.

gbalke:
@mr337 I think you have to omit RHEL 7. See https://github.com/ros/rosdistro/pull/30357/files

mr337:
> @mr337 I think you have to omit RHEL 7. See https://github.com/ros/rosdistro/pull/30357/files

I learned the hard way :(, new changes being tested

mr337:
Before merging is it better to do it this way as you linked? From RHEL 8 and on the package should be available. So will the below method do a wildcard for anything RHEL 8+?

```
  rhel:
    '*': [python3-flask-cors]
    '7': null
```

gbalke:
> Before merging is it better to do it this way as you linked? From RHEL 8 and on the package should be available. So will the below method do a wildcard for anything RHEL 8+?
> 
> ```
>   rhel:
>     '*': [python3-flask-cors]
>     '7': null
> ```

Yeah, I think the idea is we want to have forwards compatibility in the case of RHEL 9, etc. Would be a bit of a headache to have to update every package when that happens haha.

gbalke (approved):
Thanks for making the changes!

mr337:
Thanks for the help!

rosdep/python.yaml
```diff
@@ -8040,6 +8040,15 @@ python3-virtualserialports-pip:
   ubuntu:
     pip:
       packages: [PyVirtualSerialPorts]
+python3-voluptuous:
+  debian: [python-voluptuous]
+  fedora: [python-voluptuous]
```
clalancette, line 8045: I feel like in both of these cases, this should be `python3-voluptuous`.  Certainly on Debian that is the correct package name: https://packages.debian.org/buster/python3-voluptuous .  On Fedora, the situation is a little more murky, but if I do `dnf -y install python3-voluptuous` there, I get a valid package (and it looks like `python-voluptuos` is just a link to that).  So I'd suggest changing both of those.

clalancette (dismissed):

rosdep/python.yaml
mr337: OK I think I understand your recommendation clearly, moved debian and fedora to `python3-voluptuous`.

nuclearsandwich (approved):

PR #30358, opened 2021-07-30
[python] Adding python3 of pynmea2

mr337:

## Package name:

pynmea2

## Package Upstream Source:

https://github.com/Knio/pynmea2

## Purpose of using this:

This library is used to parse raw NMEA strings coming from GNSS receivers.

Distro packaging links:

  - via pip, https://packages.debian.org/buster/python-nmea2 is way to old and has only python2 support
   - https://packages.ubuntu.com/focal/python3-nmea2
  - via pip

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7282,6 +7282,14 @@ python3-pymongo:
   nixos: [python3Packages.pymongo]
   openembedded: [python3-pymongo@meta-python]
   ubuntu: [python3-pymongo]
+python3-pynmea2:
+  debian:
+    pip:
+      packages: [pynmea2]
+  fedora:
+    pip:
+      packages: [pynmea2]
+  ubuntu: [python3-nmea2]
 python3-pyosmium:
   debian: [python3-pyosmium]
   fedora: [python3-osmium]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7283,9 +7283,7 @@ python3-pymongo:
   openembedded: [python3-pymongo@meta-python]
   ubuntu: [python3-pymongo]
 python3-pynmea2:
-  debian:
-    pip:
-      packages: [pynmea2]
+  debian: [python3-nmea2]
   fedora:
     pip:
       packages: [pynmea2]


mr337:
I decided not to reference the original python2-pynmea anchor since the Ubuntu now has this in the standard packages vs pip.

rosdep/python.yaml
```diff
@@ -7267,6 +7267,14 @@ python3-pymongo:
   nixos: [python3Packages.pymongo]
   openembedded: [python3-pymongo@meta-python]
   ubuntu: [python3-pymongo]
+python3-pynmea2:
+  debian:
+    pip:
+      packages: [pynmea2]
+  fedora:
+    pip:
+      packages: [pynmea2]
```
nuclearsandwich, line 7276: New keys should not mix pip and system packages.

Please only add definitions for system packages in non-pip keys.

rosdep/python.yaml
```diff
@@ -7267,6 +7267,14 @@ python3-pymongo:
   nixos: [python3Packages.pymongo]
   openembedded: [python3-pymongo@meta-python]
   ubuntu: [python3-pymongo]
+python3-pynmea2:
+  debian:
+    pip:
+      packages: [pynmea2]
```
nuclearsandwich, line 7273:
```suggestion
  debian: [python3-nmea2]
```

This package is available in Debian since at least Debian Stretch so a pip key is not required.

nuclearsandwich (changes requested):

rosdep/python.yaml
mr337: I am a little confused by this. Based on the above comment Debian can be moved to a system package but still leaves Fedora using pip.

Would adding a new package such as `python3-pynmea2-pip` that addresses Fedora work this way there isn't a mixture of pip and system packages?

mr337: I decided to add the `-pip` key to cover fedora. Pushing on next commit.

tfoote: @nuclearsandwich We can allow pip to backfill unsupported platforms and versions for the regular key that we believe will never have conflicting native packages. Splitting the key into two like this means that you can't use the same dependencies in the package.xml on two platforms. The `-pip` suffix is to distinguish keys which are pip only and on no other installer.

mr337: I reverted the last commit and updated the single key to support both pip and system packages.

tfoote (approved):

ivanpauno (approved):

PR #30359, opened 2021-07-30
ROS-MSCL: 1.1.4-1 in 'noetic/distribution.yaml' [bloom]

robbiefish:
Increasing version of package(s) in repository `ROS-MSCL` to `1.1.4-1`:

- upstream repository: https://github.com/LORD-MicroStrain/ROS-MSCL.git
- release repository: https://github.com/LORD-MicroStrain/ROS-MSCL-release.git
- bloom version: `0.9.1`

## mscl_msgs

```
* Please see changelog
* Added aiding measurement summary for each GNSS (GQ7 only)
  Added MSCL version output when node starts
* Merge pull request #54 <https://github.com/LORD-MicroStrain/ROS-MSCL/issues/54> from civerachb-cpr/master
  Explicitly name the mscl_msgs package
* Explicitly name the mscl_msgs package instead of using the directory name; the dir name isn't reliable on some build-farms that extract each package into a separate working directory
* Contributors: Chris Iverach-Brereton, Nathan Miller, nathanmillerparker
```

## ros_mscl

```
* Installs MSCL from CMake to hopefully allow this package to be built in the buildfarm
* Merge pull request #70 <https://github.com/LORD-MicroStrain/ROS-MSCL/issues/70> from ori-drs/master
  [ros_mscl] Turn filter_data_rate and imu_data_rate into an argument
* [ros_mscl] Turn filter_data_rate and imu_data_rate into an argument
* Eliminated build warnings
* Fixed a bug that wouldn't allow the rtk dongle to be enabled as it was using the wrong variable to enable it.
* See changelog
* Added aiding measurement summary for each GNSS (GQ7 only)
  Added MSCL version output when node starts
* Merge pull request #50 <https://github.com/LORD-MicroStrain/ROS-MSCL/issues/50> from civerachb-cpr/master
  Add an arg to enable setting NED/ENU frame parameter
* Add an arg to enable setting NED/ENU frame parameter
* Contributors: Chris Iverach-Brereton, Nathan Miller, Wolfgang Merkt, nathanmillerparker, robbiefish
```

## ros_mscl_cpp_example

```
* Merge branch 'master' of https://github.com/LORD-MicroStrain/ROS-MSCL into develop
* Properly installs examples to match instructions
* Contributors: robbiefish
```

## ros_mscl_py_example

```
* Merge branch 'master' of https://github.com/LORD-MicroStrain/ROS-MSCL into develop
* Properly installs examples to match instructions
* Contributors: robbiefish
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -10,6 +10,23 @@ release_platforms:
   ubuntu:
   - focal
 repositories:
+  ROS-MSCL:
+    release:
+      packages:
+      - mscl_msgs
+      - ros_mscl
+      - ros_mscl_cpp_example
+      - ros_mscl_py_example
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/LORD-MicroStrain/ROS-MSCL-release.git
+      version: 1.1.4-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/LORD-MicroStrain/ROS-MSCL.git
+      version: develop
+    status: developed
   abb_robot_driver_interfaces:
     doc:
       type: git


clalancette:
This PR has a few problems.

1. This is failing CI because it is not in alphabetical order.  That should be easily fixed.
2. This is the only package to use ALL CAPS for the package key, and that should not be.  Please use lower case for the package key.
3. The name of this package also seems to violate https://www.ros.org/reps/rep-0144.html , specifically the global rule where a package name should not contain "ros".  That said, it could be argued that this is "binding of an upstream library", in which case the package name is OK.  If you want to argue that, please let us know.
4. Looking at the source repository, there are a lot of TODO comments in https://github.com/LORD-MicroStrain/ROS-MSCL/blob/master/mscl_msgs/package.xml .  Those should be fixed.
5. Looking at https://github.com/LORD-MicroStrain/ROS-MSCL/blob/master/ros_mscl/CMakeLists.txt#L23-L29 , I'm quite concerned that this can't possibly work on the ROS buildfarm.  Where is the MSCL library coming from?  Why would it be installed into `/usr/share` ?  Can you please explain more about how this package works, how it interacts with MSCL, and how you expect the MSCL library to be installed?

robbiefish:
Thanks for the quick response. As noted, number 1, 2, and 4 are easy fixes, but I had some questions on the other points

> 3\. The name of this package also seems to violate https://www.ros.org/reps/rep-0144.html , specifically the global rule where a package name should not contain "ros".  That said, it could be argued that this is "binding of an upstream library", in which case the package name is OK.  If you want to argue that, please let us know.

The purpose of these packages is to wrap a library that we maintain to interact with our devices called MSCL. I am not sure if this qualifies as `binding of an upstream library`, but if we did not have to rename the packages that would be convenient for us. What would we need to do to make that argument?

> 5\. Looking at https://github.com/LORD-MicroStrain/ROS-MSCL/blob/master/ros_mscl/CMakeLists.txt#L23-L29 , I'm quite concerned that this can't possibly work on the ROS buildfarm.  Where is the MSCL library coming from?  Why would it be installed into `/usr/share` ?  Can you please explain more about how this package works, how it interacts with MSCL, and how you expect the MSCL library to be installed?

This is likely another problem with this PR, but this release is based off of the [develop](https://github.com/LORD-MicroStrain/ROS-MSCL/tree/develop) branch. On develop, we have added the installation of the library in the [CMakeLists.txt](https://github.com/LORD-MicroStrain/ROS-MSCL/blob/develop/ros_mscl/CMakeLists.txt), and we install in the ROS package workspace instead of `/usr/share`. Is there a better way to do this? We had thought about standing up our own deb package server to host the MSCL deb files, but were not sure how that would work with the buildfarm as you would need to add the new server to your sources list.

clalancette:
> This is likely another problem with this PR, but this release is based off of the [develop](https://github.com/LORD-MicroStrain/ROS-MSCL/tree/develop) branch. On develop, we have added the installation of the library in the [CMakeLists.txt](https://github.com/LORD-MicroStrain/ROS-MSCL/blob/develop/ros_mscl/CMakeLists.txt), and we install in the ROS package workspace instead of `/usr/share`. Is there a better way to do this? We had thought about standing up our own deb package server to host the MSCL deb files, but were not sure how that would work with the buildfarm as you would need to add the new server to your sources list.

We definitely cannot add a new server to our sources list.

In terms of building and installing the MSCL library, there are a few ways to do this:
1. Use the version in the base distribution (assuming the library is available in Ubuntu, for instance).  This is preferred if it is available, as all of the updates will come from the base distribution.
2. Bundle the library into the package (as you seem to have done on the `develop` branch).
3. Release a separate package into the ROS distribution that just contains the library.  Once that is in there, then this package can depend on that package.  This is nice if more than one package needs to use the functionality from the library.

nuclearsandwich:
>     3\. Release a separate package into the ROS distribution that just contains the library.  Once that is in there, then this package can depend on that package.  This is nice if more than one package needs to use the functionality from the library.

From a release management perspective, this is the approach I strongly recommend. There is a [bloom tutorial for releasing third party packages](https://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty). I suggest starting with the release of MSCL into ROS and commenting on the bloom PR that this library will shortly be used as a dependency for the ros bindings to MSCL package.

robbiefish:
Thanks for all the advice. We decided to go with option 2 outlined above because of the way our releases are currently orchestrated.

A new PR has been opened at #30558 that hopefully addresses these issues, so I am closing this

PR #30407, opened 2021-08-05
oled_display_node: 1.0.0-1 in 'noetic/distribution.yaml' [bloom]

rohbotics:
Increasing version of package(s) in repository `oled_display_node` to `1.0.0-1`:

- upstream repository: https://github.com/UbiquityRobotics/oled_display_node.git
- release repository: https://github.com/UbiquityRobotics-release/oled_display_node-release.git
- bloom version: `0.10.7`

## oled_display_node

```
* Remove /* in comment that can cause warnings
  Bug: https://github.com/UbiquityRobotics/oled_display_node/issues/4
* Refresh the IP addr and machine name lines every 10 seconds.  This allows for a changed IP address due to networking changes to show up worse case in 10 seconds.  Any other lines the app outside of the oled display driver is concerned must be refreshed by sending messages to the display node over the input topic as this auto-refresh does not want to make assumptions about user lines in use on the display
* README.md
* Refactoring to become oled_display_node
* Added CMake install rules to allow for package release
  This should install the main executable, the launch files and the
  demo script.
* Cleanup of package and make and more
* trying to fix the section on starting from launch file
* Cleanup and enhance README to point out we intend this display node to support 1.3 inch OLED display using the SH1106 controller chip
* Initial push of OLED display node with README
* Contributors: Mark Johnston, Mark Johston, Rohan Agrawal
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4847,6 +4847,17 @@ repositories:
       url: https://github.com/ros-drivers/odva_ethernetip.git
       version: kinetic-devel
     status: maintained
+  oled_display_node:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/UbiquityRobotics-release/oled_display_node-release.git
+      version: 1.0.0-1
+    source:
+      type: git
+      url: https://github.com/UbiquityRobotics/oled_display_node.git
+      version: master
+    status: maintained
   ompl:
     release:
       tags:


jacobperron (commented):
Please add a LICENSE file to the source repository. A new release is not required, we just want to see the LICENSE in the source repository.

rohbotics:
Added, thanks for pointing that out.

clalancette (changes requested):
I have a few concerns with this package.

The first is that it "builds" on x86_64, but doesn't actually produce anything useful.  But I don't really understand why; basically everything it does should, in theory, work on any architecture, not just the Raspberry Pi.

The other problem I have is that this package is named `oled_display_node`, but is actually a really specific node for the SH1106 controller chip.  From that perspective, I think this should really be renamed to `sh1106_oled_display_node` or something like that.

rohbotics:
Hi @clalancette 

This package is also currently released on kinetic, so I am going to say that changing the name doesn't make sense right now. I am also not happy with the name honestly, but I think we could address that in a future ROS2 release.

Agree on the x86_64 issue, we were initially going for a bit-banging approach that may have use some Pi specifics, but now that we are just using the kernel interface we should get rid of that ugly "fake build".

We are going to be releasing a new version of the package soon (in a week or so), that would address the x86_64 issue as well as add some additional features. Should I just close this PR and we can discuss on that release, or can we merge this in and have the next version have meaningful x86_64 build? 

Note I am trying to get the magni_robot package released on noetic, and this is the last remaining dependency, so I would prefer the first option if that works for you.

clalancette:
> This package is also currently released on kinetic, so I am going to say that changing the name doesn't make sense right now. I am also not happy with the name honestly, but I think we could address that in a future ROS2 release.

Yeah, that makes sense.  Again, I just forgot to check previous ROS releases, apologies for that.

clalancette (approved):

rohbotics:
No problem, thanks for the merge!

PR #30409, opened 2021-08-05
Add python-monotonic

Timple:

## Package name:

python-monotonic

## Package Upstream Source:

python-monotonic

## Purpose of using this:

monotonic time is not available for python2. It can be usefull for dt calculations on systems where the clock can be modified by the operating system.

Distro packaging links:

  - `python-monotonic`
   - `python-monotonic`
  - `python-monotonic`
  - `python-monotonic`
  - `dev-python/monotonic`
  - not available
  - not available
  - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2673,6 +2673,7 @@ python-monotonic:
   debian: [python-monotonic]
   fedora: [python-monotonic]
   nixos: [pythonPackages.monotonic]
+  opensuse: [python2-monotonic]
   ubuntu: [python-monotonic]
 python-more-itertools:
   arch: [python-more-itertools]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2668,6 +2668,11 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
+python-monotonic:
+  arch: [python-monotonic]
+  debian: [python-monotonic]
+  fedora: [python-monotonic]
+  ubuntu: [python-monotonic]
 python-more-itertools:
   arch: [python-more-itertools]
   debian: [python-more-itertools]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2672,7 +2672,6 @@ python-monotonic:
   arch: [python-monotonic]
   debian: [python-monotonic]
   fedora: [python-monotonic]
-  gentoo: [dev-python/monotonic]
   ubuntu: [python-monotonic]
 python-more-itertools:
   arch: [python-more-itertools]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2672,6 +2672,7 @@ python-monotonic:
   arch: [python-monotonic]
   debian: [python-monotonic]
   fedora: [python-monotonic]
+  gentoo: [dev-python/monotonic]
   ubuntu: [python-monotonic]
 python-more-itertools:
   arch: [python-more-itertools]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2671,7 +2671,7 @@ python-mock:
 python-monotonic:
   arch: [python-monotonic]
   debian: [python-monotonic]
-  fedora: [python-monotonic]
+  nixos: [pythonPackages.monotonic]
   ubuntu: [python-monotonic]
 python-more-itertools:
   arch: [python-more-itertools]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2668,6 +2668,12 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
+python-monotonic:
+  arch: [python-monotonic]
+  debian: [python-monotonic]
+  fedora: [python-monotonic]
+  gentoo: [dev-python/monotonic]
+  ubuntu: [python-monotonic]
```
jacobperron, line 2676: Let's add an entry for opensuse too:
```suggestion
  opensuse: [python2-monotonic]
  ubuntu: [python-monotonic]
```

rosdep/python.yaml
```diff
@@ -2668,6 +2668,12 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
+python-monotonic:
+  arch: [python-monotonic]
+  debian: [python-monotonic]
+  fedora: [python-monotonic]
+  gentoo: [dev-python/monotonic]
```
jacobperron, line 2675: I couldn't find that this package exists on Gentoo: https://packages.gentoo.org/packages/dev-python/monotonic
Maybe you could point me to the place?

jacobperron (commented):
It looks like there's a NixOS package here: https://search.nixos.org/packages?channel=21.05&show=python39Packages.monotonic&from=0&size=50&sort=relevance&query=monotonic

I think we can use the alias `pythonPackages.monotonic` (or maybe it's `python3Packages.monotonic`).

@lopsided98 please advise

lopsided98:
`pythonPackages.monotonic` is correct. `pythonPackages` should be used for Python 2 packages.

rosdep/python.yaml
Timple: A quick google-search pointed me to it. But perhaps in another flavor, I'll remove.

Timple:
For python3 the package is not required. You can simply `from time import monotomic`

So this is 'just' a python2 release for rosdistro.

jacobperron:
Thanks! Sounds like we can add `nixos: [pythonPackages.monotonic]` for Python 2 then.

rosdep/python.yaml
```diff
@@ -2668,6 +2668,11 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
+python-monotonic:
+  arch: [python-monotonic]
+  debian: [python-monotonic]
+  fedora: [python-monotonic]
```
nuclearsandwich, line 2674:
```suggestion
  nixos: [pythonPackages.monotonic]
```

Based on the discussion I propose this Nix package addition.

nuclearsandwich:
> It looks like there's a NixOS package here: [search.nixos.org/packages?channel=21.05&show=python39Packages.monotonic&from=0&size=50&sort=relevance&query=monotonic](https://search.nixos.org/packages?channel=21.05&show=python39Packages.monotonic&from=0&size=50&sort=relevance&query=monotonic)

Is there a specific motivation for adding nixOS package definitions for this PR?

jacobperron:
> Is there a specific motivation for adding nixOS package definitions for this PR?

No specific motivation. I just noticed the package is available :see_no_evil:

nuclearsandwich:
@Timple if you could please review the suggested changes and incorporate at least the OpenSUSE package that will help move this through review. Thanks.

rosdep/python.yaml
Timple: This helps :)
Didn't have access to a pc

jacobperron: This suggestion accidentally removed the Fedora entry.

rosdep/python.yaml
```diff
@@ -2668,6 +2668,11 @@ python-mock:
     xenial_python3: [python3-mock]
     yakkety: [python-mock]
     zesty: [python-mock]
+python-monotonic:
+  arch: [python-monotonic]
+  debian: [python-monotonic]
```
jacobperron, line 2673:
```suggestion
  debian: [python-monotonic]
  fedora: [python-monotonic]
```

mabelzhang:
We're still missing the opensuse key suggested above. Could you accept that suggestion?

mabelzhang (approved):

ivanpauno (approved):

ivanpauno:
@Timple please consider copying an actual link to the package in the `Links to Distribution Packages` section, and not only the package name for that OS. e.g.: https://archlinux.org/packages/community/any/python-monotonic/ for Arch.
That makes the reviewing process faster.

Thanks for the contribution!

PR #30436, opened 2021-08-09
python3-pyproj-pip: add dep

russkel:

## Package name:

pyproj via pip

## Package Upstream Source:

https://github.com/pyproj4/pyproj

## Purpose of using this:

Focal's version of pyproj is out of date (the proj dependency is multiple major point releases out of date) and is missing features such as `database`. 

https://pypi.org/project/pyproj/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7309,6 +7309,16 @@ python3-pyproj:
     '*': ['python%{python3_pkgversion}-pyproj']
     '7': null
   ubuntu: [python3-pyproj]
+python3-pyproj-pip:
+  debian:
+    pip:
+      packages: [pyproj]
+  fedora:
+    pip:
+      packages: [pyproj]
+  ubuntu:
+    pip:
+      packages: [pyproj]
 python3-pyqrcode:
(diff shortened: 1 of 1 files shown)


jacobperron (approved):
I'm not sure if we want to merge in a pip key given system packages are available.

We highly prefer maintainers to rely on the system packages over packages from pip since packages depending on pip packages cannot be released into a ROS distribution.

Though, I understand the pip key is convenient for from-source builds, so I'd be okay with merging this. I'd like to get thoughts from others (@clalancette, @tfoote).

tfoote (changes requested):
As Jacob mentions the system packages take priority. We should not include a pip package which will mask a debian package. The debian package has been tested and cross validated and we want to stick with the released managed version. It is expected to be an older version than what's available from pip. But the stability of the distro is what people rely upon and why we use the distro instead of building everything from source to get the latest.

russkel:
I was under the impression that there were keys for both system package and pip package. Further inspection of the `python3` keys show that not to be the case.

I'll maintain this key locally in that case.

PR #30441, opened 2021-08-09
update doc and source branches for libstatistics_collector in Galactic

emersonknapp:
Followup to #30411 - update doc and source branches as well


emersonknapp:
Nevermind - already done in https://github.com/ros/rosdistro/pull/30426

PR #30449, opened 2021-08-10
ruckig: 0.3.3-1 in 'melodic/distribution.yaml' [bloom]

pantor:
Increasing version of package(s) in repository `ruckig` to `0.3.3-1`:

- upstream repository: https://github.com/pantor/ruckig.git
- release repository: https://github.com/pantor/ruckig-release.git
- bloom version: `0.10.7`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -12186,7 +12186,7 @@ repositories:
     source:
       type: git
       url: https://github.com/pantor/ruckig.git
-      version: v0.3.3
+      version: master
     status: developed
   rviz:
     doc:
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -12177,6 +12177,17 @@ repositories:
       url: https://github.com/gbiggs/rtsprofile.git
       version: master
     status: maintained
+  ruckig:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/pantor/ruckig-release.git
+      version: 0.3.3-1
+    source:
+      type: git
+      url: https://github.com/pantor/ruckig.git
+      version: v0.3.3
+    status: developed
   rviz:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -12177,6 +12177,17 @@ repositories:
...
     status: maintained
+  ruckig:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/pantor/ruckig-release.git
+      version: 0.3.3-1
+    source:
+      type: git
+      url: https://github.com/pantor/ruckig.git
```
clalancette, line 12188: The [LICENSE](https://github.com/pantor/ruckig/blob/master/LICENSE) file seems to contain both the MIT and BSD licenses, but the [package.xml](https://github.com/pantor/ruckig/blob/master/package.xml) only lists MIT.  If BSD code is used in this package, it should have a second `<license>` tag in the `package.xml`.

melodic/distribution.yaml
```diff
@@ -12177,6 +12177,17 @@ repositories:
...
+      version: v0.3.3
```
clalancette, line 12189: A tag is not going to work here; the source needs to point to a branch.  I'll suggest changing this to `main`.

clalancette (changes requested):
Besides what I pointed out inline, this package also seems to require that the `Reflexxes` library be installed.  However, that will not not be available on the buildfarm.  What will happen if the build can't find that library?

melodic/distribution.yaml
pantor: Ruckig is only licensed under MIT, the BSD text was just a leftover... I've removed that now, thanks for pointing out!

melodic/distribution.yaml
pantor: As I'm not really familiar with the ROS release process: Can I change that within this PR or do I need to repeat the whole process with `bloom-release`?

clalancette: You can just change it in this PR.  That is actually preferred.

pantor: Changed that to `master`.

clalancette (approved):

PR #30454, opened 2021-08-10
Add python3-gpiozero to rosdep

RyanHir:

## Package name:

python3-gpiozero

## Purpose of using this:

This library is able to read CPU temperature data from the Raspberry Pi and is useful for measuring sensor changes.

- Debian: https://packages.debian.org/buster/python3-gpiozero
- Ubuntu: https://packages.ubuntu.com/bionic/python3-gpiozero
- Fedora: https://src.fedoraproject.org/rpms/python-gpiozero

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6460,6 +6460,10 @@ python3-gnupg:
   opensuse: [python3-python-gnupg]
   ubuntu: [python3-gnupg]
 python3-google-cloud-texttospeech-pip: *migrate_eol_2025_04_30_python3_google_cloud_texttospeech_pip
+python3-gpiozero:
+  debian: [python3-gpiozero]
+  fedora: [python3-gpiozero]
+  ubuntu: [python3-gpiozero]
 python3-gpxpy:
   debian: [python3-gpxpy]
   fedora: [python3-gpxpy]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6460,6 +6460,10 @@ python3-gnupg:
   opensuse: [python3-python-gnupg]
   ubuntu: [python3-gnupg]
 python3-google-cloud-texttospeech-pip: *migrate_eol_2025_04_30_python3_google_cloud_texttospeech_pip
+python3-gpiozero:
+  debian: [python3-gpiozero]
+  fedora: [python3-gpiozero]
+  ubuntu: [python3-gpiozero]
```
nuclearsandwich, line 6466: Per the check failure, python3-gpiozero does not appear to be available in Ubuntu Bionic.

```suggestion
  ubuntu:
    '*': [python3-gpiozero]
    bionic: null
```

nuclearsandwich (changes requested):

RyanHir:
It appears that python3-gpiozero is only in the arm ports for Ubuntu Bionic, I have accepted your requested changes.

nuclearsandwich:
> It appears that python3-gpiozero is only in the arm ports for Ubuntu Bionic, I have accepted your requested changes.

Ah yeah, rosdep does not consider architecture and so our checks only run on the primary Ubuntu repositories. Thanks for clarifying and iterating.

nuclearsandwich (approved):

mabelzhang (approved):

PR #30474, opened 2021-08-12
Add python3-mediapipe-pip

viktorlorentz:

## Package name:

python-mediapipe-pip

## Package Upstream Source:

https://github.com/google/mediapipe

## Purpose of using this:

This adds mediapipe solutions to rosdep. Mediapipe solutions offer an easy and performant way to build pose tracking and other ml based applications.

Distro packaging links:

PyPi: https://pypi.org/project/mediapipe/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2632,19 +2632,6 @@ python-mechanize:
   gentoo: [dev-python/mechanize]
   nixos: [pythonPackages.mechanize]
   ubuntu: [python-mechanize]
-python3-mediapipe-pip:
-  debian:
-    pip:
-      packages: [mediapipe]
-  fedora:
-    pip:
-      packages: [mediapipe]
-  osx:
-    pip:
-      packages: [mediapipe]
-  ubuntu:
-    pip:
-      packages: [mediapipe]
 python-mock:
   alpine: [py-mock]
   arch: [python2-mock]
@@ -6753,6 +6740,19 @@ python3-mechanize:
     '*': [python3-mechanize]
     bionic: null
     xenial: null
+python3-mediapipe-pip:
+  debian:
+    pip:
+      packages: [mediapipe]
+  fedora:
+    pip:
+      packages: [mediapipe]
+  osx:
+    pip:
+      packages: [mediapipe]
+  ubuntu:
+    pip:
+      packages: [mediapipe]
 python3-meshio:
   arch: [python-meshio]
   debian: [python3-meshio]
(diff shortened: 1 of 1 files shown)


mabelzhang (approved):

mabelzhang:
Looks like there's an error in the tests from `test/rosdep_repo_check/test_rosdep_repo_check.py`. Could you fix the error? I'm not able to run the test locally for some weird yaml reason. Maybe try merging from `master`.

mabelzhang:
Okay merged from `master`. That should fix the tests.

viktorlorentz:
@mabelzhang thx for fixing the error. LGTM now!

rosdep/python.yaml
```diff
@@ -2632,6 +2632,19 @@ python-mechanize:
   gentoo: [dev-python/mechanize]
   nixos: [pythonPackages.mechanize]
   ubuntu: [python-mechanize]
+python-mediapipe-pip:
```
ivanpauno, line 2635: did you wanted to add the python2 version or actually one for python3?
according to the [contributing guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-rules), the prefix should be `python3` and not `python` in that case.

AFAIS `mediapipe` never supported python2

ivanpauno (changes requested):

rosdep/python.yaml
viktorlorentz: Thy for the hint. I updated the prefix accordingly.

rosdep/python.yaml
```diff
@@ -2632,6 +2632,19 @@ python-mechanize:
   gentoo: [dev-python/mechanize]
   nixos: [pythonPackages.mechanize]
   ubuntu: [python-mechanize]
+python3-mediapipe-pip:
```
ivanpauno, line 2635: you need to use alpha order to make the tests pass.
You can use rosdistro_reformat, which is available after installing python3-rosdistro (example [here](http://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation#Steps_to_have_your_repository_added_to_our_indexer))

viktorlorentz: Rookie mistake... Its sorted now.

ivanpauno: no problem, thanks for updating!

ivanpauno (approved):

hidmic (approved):

hidmic:
Thanks for your contribution and patience @viktorlorentz.

PR #30475, opened 2021-08-12
Add rqt_paramedit to documentation index for melodic

dornhege:
# Please Add rqt_paramedit to be indexed in the rosdistro.

melodic

# The source is here: 

https://github.com/dornhege/rqt_paramedit/tree/melodic-devel

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -11708,6 +11708,12 @@ repositories:
       url: https://github.com/ros-visualization/rqt_nav_view.git
       version: master
     status: maintained
+  rqt_paramedit:
+    doc:
+      type: git
+      url: https://github.com/dornhege/rqt_paramedit.git
+      version: melodic-devel
+    status: maintained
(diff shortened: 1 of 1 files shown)


mabelzhang (approved):

melodic/distribution.yaml
```diff
@@ -11680,6 +11680,12 @@ repositories:
       url: https://github.com/ros-visualization/rqt_nav_view.git
       version: master
     status: maintained
+  rqt_paramedit:
+    doc:
```
ivanpauno, line 11684: Could you add also a `source` entry?
Having only a docs entry is fine though, but it's more common to have both.

ivanpauno (changes requested):

melodic/distribution.yaml
dornhege: I just added a source tag.

ivanpauno (approved):

PR #30588, opened 2021-08-28
cpu_temperature_diagnostics: Initial release

yotabits:

Melodic

# The source is here: 

https://github.com/yotabits/cpu_temperature_diagnostics/tree/melodic-release

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2018,6 +2018,10 @@ repositories:
       version: kinetic-devel
     status: maintained
   cpu_temperature_diagnostics:
+    doc:
+      type: git
+      url: https://github.com/yotabits/cpu_temperature_diagnostics.git
+      version: melodic-release
     source:
       type: git
       url: https://github.com/yotabits/cpu_temperature_diagnostics.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2017,6 +2017,12 @@ repositories:
       url: https://github.com/clearpathrobotics/cpr_multimaster_tools.git
       version: kinetic-devel
     status: maintained
+  cpu_temperature_diagnostics:
+    source:
+      type: git
+      url: https://github.com/yotabits/cpu_temperature_diagnostics.git
+      version: melodic-release
+    status: maintained
   crane_x7:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -2017,6 +2017,12 @@ repositories:
       url: https://github.com/clearpathrobotics/cpr_multimaster_tools.git
       version: kinetic-devel
     status: maintained
+  cpu_temperature_diagnostics:
+    source:
```
ivanpauno, line 2021: could you also add a doc entry?

melodic/distribution.yaml
```diff
@@ -2017,6 +2017,12 @@ repositories:
...
+      type: git
+      url: https://github.com/yotabits/cpu_temperature_diagnostics.git
```
ivanpauno, line 2023: why is the release being done from a fork?

yotabits: Because the release is happening on my own initiative. Original repo organization does not have time to allocate to make and maintain a release.

ivanpauno: This is adding a package as a source entry, so there doesn't seem to be any release associated.
The upstream repo also seems to be responsive to your patches.

If you're planning to start making releases, you only need to have write access to the release repo, you don't need to be using a fork of the source repo.
Of course, if the upstream maintainer is not responsive we could update the source repo to be your fork.

---

@clalancette  do we have any policy related to when to forks?

yotabits: @ivanpauno @clalancette thank you for the feedback.
The original repo is a repository part of a company that does not need to release the package and will not allocate time to do it or maintain it.
Currently, I have write permission to this repository as I'm part of this company, if I happen to not be part of it anymore I will lose all my rights.
Consequently, I think it's better to use a fork I have control of instead of the original repository that can change independently of my will.
I'm the author of this package.

melodic/distribution.yaml
yotabits: Thank you for the suggestion. I added the doc entry.

ivanpauno (approved):

melodic/distribution.yaml
ivanpauno: Considering that you're the author (which I didn't notice before) and you prefer to add the fork as an upstream repo that sounds fair enough to me.
Though I agree with @clalancette that this is a bit confusing.

sloretz (approved):

PR #30599, opened 2021-08-30
micro_ros_msgs: 1.0.0-2 in 'foxy/distribution.yaml' [bloom]

pablogs9:
Increasing version of package(s) in repository `micro_ros_msgs` to `1.0.0-2`:

- upstream repository: https://github.com/micro-ROS/micro_ros_msgs
- release repository: https://github.com/ros2-gbp/micro_ros_msgs-release.git
- bloom version: `0.10.7`

## micro_ros_msgs

```
* Remove not necessary dependencies (#1 <https://github.com/micro-ROS/micro_ros_msgs/issues/1>)
* Add issue template
* Add msg definitions used by micro-ROS graph manager
* Update README.md
```

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1919,11 +1919,20 @@ repositories:
       version: foxy
     status: maintained
   micro_ros_msgs:
+    doc:
+      type: git
+      url: https://github.com/micro-ROS/micro_ros_msgs.git
+      version: foxy
     release:
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/ros2-gbp/micro_ros_msgs-release.git
       version: 1.0.0-2
+    source:
+      type: git
+      url: https://github.com/micro-ROS/micro_ros_msgs.git
+      version: foxy
+    status: maintained
   mimick_vendor:
     doc:
       type: git
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1918,6 +1918,12 @@ repositories:
       url: https://github.com/ros2/message_filters.git
       version: foxy
     status: maintained
+  micro_ros_msgs:
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/ros2-gbp/micro_ros_msgs-release.git
+      version: 1.0.0-2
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -1918,6 +1918,12 @@ repositories:
       url: https://github.com/ros2/message_filters.git
       version: foxy
     status: maintained
+  micro_ros_msgs:
+    release:
```
ivanpauno, line 1922: could you add source and doc entries?
it's not a requirement, but it's usually done

pablogs9: Should I add it manually here or do I have to redo it using the bloom command?

Sorry, but it is our first time releasing with bloom and I have some doubts. One of them is: is normal to release the same version (1.0.0 in this case) for different ROS distributions (Rolling, Foxy, and Galactic)?

ivanpauno: > Should I add it manually here or do I have to redo it using the bloom command?

Both ways are fine.

> One of them is: is normal to release the same version (1.0.0 in this case) for different ROS distributions (Rolling, Foxy, and Galactic)?

If your code doesn't change between the distributions, then yes, it's normal to use the same version.

pablogs9: Done. One more question: I'm expecting that this generates a binary installable via `apt` that's right?

Then, if I have this package as a dependency of another will the `apt` package manager solve the interdependency automatically?

ivanpauno: > Done. One more question: I'm expecting that this generates a binary installable via apt that's right?

Yes, this generates binaries that in ubuntu can be installed via `apt` (you have to add the ros2 apt repository to have them available).

> Then, if I have this package as a dependency of another will the apt package manager solve the interdependency automatically?

I'm not sure if this is what you're asking, but when you install the package it will also install all its required dependencies.

Maybe related, but you can also use rosdep to install dependencies of the packages in your workspace, see http://wiki.ros.org/rosdep#Install_dependency_of_all_packages_in_the_workspace.

---

The binary won't be available until next sync though, syncs roughly happen each 15 days (sometimes it can take a month).

PR #30611, opened 2021-08-31
zenoh-bridge-dds: 0.5.0-1 in 'rolling/distribution.yaml' [bloom]

JEnoch:
Increasing version of package(s) in repository `zenoh-bridge-dds` to `0.5.0-1`:

- upstream repository: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
- release repository: https://github.com/atolab/zenoh-bridge-dds-release.git
- bloom version: `0.10.7`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4430,5 +4430,16 @@ repositories:
       url: https://github.com/ros2/yaml_cpp_vendor.git
       version: master
     status: maintained
+  zenoh-bridge-dds:
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/atolab/zenoh-bridge-dds-release.git
+      version: 0.5.0-1
+    source:
+      type: git
+      url: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
+      version: a466f226765cb4051a368870ba34fad5d5b9f523
+    status: developed
 type: distribution
 version: 2
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4431,6 +4431,10 @@ repositories:
       version: master
     status: maintained
   zenoh-bridge-dds:
+    doc:
+      type: git
+      url: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
+      version: master
     release:
       tags:
         release: release/rolling/{package}/{version}
@@ -4439,7 +4443,7 @@ repositories:
     source:
       type: git
       url: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
-      version: a466f226765cb4051a368870ba34fad5d5b9f523
+      version: master
     status: developed
 type: distribution
 version: 2


rolling/distribution.yaml
```diff
@@ -4430,5 +4430,16 @@ repositories:
       url: https://github.com/ros2/yaml_cpp_vendor.git
       version: master
     status: maintained
+  zenoh-bridge-dds:
+    release:
```
hidmic, line 4434: @JEnoch no `doc` entry?

rolling/distribution.yaml
```diff
@@ -4430,5 +4430,16 @@ repositories:
...
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/atolab/zenoh-bridge-dds-release.git
+      version: 0.5.0-1
+    source:
+      type: git
+      url: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
+      version: a466f226765cb4051a368870ba34fad5d5b9f523
```
hidmic, line 4442: @JEnoch hmm, it's the first I see a commit SHA for a `source` version. I'm inclined to think it won't fly.

rolling/distribution.yaml
JEnoch: @hidmic I was following the [bloom 1st release tuto](http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease) that doesn't refer to the doc. I was surprised when `bloom-release` asked for. But as it proposes a `y/n` answer, I assumed the doc was not required for a 1st release and thought to push it later in another PR.
The doc content is ready as it will be a copy of this [README.md](https://github.com/eclipse-zenoh/zenoh-plugin-dds#readme). I just need to check if I have to manually copy/adapt it for ROS2 doc, or if I can automatize this.

rolling/distribution.yaml
JEnoch: `bloom-release` did propose me to use a tag for `VCS version`:
```bash
==> Generating pull request to distro file located at 'https://raw.githubusercontent.com/ros/rosdistro/5192944639c35d498b155790e6d9231efb8f85cf/rolling/distribution.yaml'
Would you like to add documentation information for this repository? [Y/n]? n
Would you like to add source information for this repository? [Y/n]? 
==> Looking for a source entry for this repository in a different distribution...
No existing source entries found for use as defaults.
Please enter information which points to the active development branch for this repository.
This information is used to run continuous integration jobs and for developers to checkout from.
VCS Type must be one of git, svn, hg, or bzr.
VCS type: git
VCS url: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
VCS version must be a branch, tag, or commit, e.g. master or 0.1.0
VCS version: a466f226765cb4051a368870ba34fad5d5b9f523
Since you are on github we can add a job to run your tests on each pull request.If you would like to turn this on please see http://wiki.ros.org/buildfarm/Pull%20request%20testing for more information. There is more setup required to setup the hooks correctly. 
```

But I'm confused when I see most of other packages to use a branch name for `source.version`. What is used by a ROS2 user or by the build farm to checkout/build a specific version of a package ? For the build to be reproductible with the exact same sources, it needs to be a VCS tag or a commit reference, not a branch name, right ?

rolling/distribution.yaml
JEnoch: @hidmic: I just pushed a commit adding the `doc` entry.
Can you please correct me if I'm misunderstood those points:
 - the [doc indexing tutorial](https://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation) only refers to ROS packages and is obsolete for ROS2, right ? If I look at http://wiki.ros.org/turtlesim it lists only ROS distros. While looking at https://index.ros.org/p/turtlesim/ it lists only to ROS2 distros
 - As a consequence there is no longer need to [create a wiki page](https://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation#Create_a_wiki_page_for_your_package_in_very_short_steps). The package page at index.ros.org will copy the README.md from the `doc` repo, correct ?

rolling/distribution.yaml
JEnoch: My last commit also replaced the commit SHA with `master`. 
I guess for the release the build farm will checkout the sources referred as `last_release` in my [track.yaml](https://github.com/atolab/zenoh-bridge-dds-release/blob/70a82fdfa593f0a560d43a3823bb489f575c1243/tracks.yaml#L18), right ?

hidmic: > bloom-release did propose me to use a commit for VCS version:

Uh, that's true. But as you've seen, it's fairly unusual.

> I guess for the release the build farm will checkout the sources referred as last_release in my `track.yaml`, right ?

Hmm, no, the process is somewhat more involved. IIRC `track.yaml` only assists `bloom-release`, which then populates branches in your release repository with both source code and packaging specific bits. 

---

Now that I take a closer look at the [`debian/rolling/zenoh-bridge-dds`](https://github.com/atolab/zenoh-bridge-dds-release/tree/debian/rolling%2Fzenoh-bridge-dds) branch in your release repository, there're a few assumptions there that don't hold. Fetching source files out-of-tree with `..` paths will not work -- they're simply not there (as `bloom` couldn't possibly know it had to pick them up).

I will also point out that pulling additional dependencies through `cargo` is not great. Both `rosdep` and `bloom` generated packaging are unaware of them, so if linkage is anything but static you'll end up with a broken binary. 

I don't know if and how we've handled with Rust software in the build farm before. CC @nuclearsandwich.

JEnoch: Thanks for your enlightening explanations!

> Now that I take a closer look at the debian/rolling/zenoh-bridge-dds branch in your release repository, there're a few assumptions there that don't hold. Fetching source files out-of-tree with .. paths will not work -- they're simply not there (as bloom couldn't possibly know it had to pick them up).

🤔 I see the complete sources have been copied in [upstream](https://github.com/atolab/zenoh-bridge-dds-release/tree/upstream) branch. But indeed in [debian/rolling/zenoh-bridge-dds](https://github.com/atolab/zenoh-bridge-dds-release/tree/debian/rolling%2Fzenoh-bridge-dds) only the content of `ros2_pkg` has been copied.
Thus it means I have no choice but to put `package.xml` and `CMakeLists.txt` at the root of the repo ?
If I fix this, can I do a `bloom-release` again even if this one has not been merged and released ?

> I will also point out that pulling additional dependencies through cargo is not great. Both rosdep and bloom generated packages are unaware of them, so if linkage is anything but static you'll end up with a broken binary.

Actually by default Rust builds static standalone binaries. And that's the case for `zenoh-bridge-dds`. All dependencies sources are downloaded, built and statically linked by `cargo`.

hidmic: > Thus it means I have no choice but to put package.xml and CMakeLists.txt at the root of the repo ?

Well, looking at https://github.com/eclipse-zenoh/zenoh-plugin-dds, I'd personally make `zplugin-dds` and `zenoh-bridge-dds` each one a package of their own (like in https://github.com/ros2-rust/ros2_rust) and make the latter depend on the former.

That correctly acknowledges the fact that they are indeed separate Rust packages.

> Actually by default Rust builds static standalone binaries. And that's the case for zenoh-bridge-dds. All dependencies sources are downloaded, built and statically linked by cargo.

:+1:

hidmic: > If I fix this, can I do a bloom-release again even if this one has not been merged and released ?

We can close this PR and you can source release and bloom release once you've sorted things out in your source repository.

JEnoch: > Well, looking at https://github.com/eclipse-zenoh/zenoh-plugin-dds, I'd personally make zplugin-dds (would zenoh-plugin-dds be a more appropriate name?) and zenoh-bridge-dds each one a package of their own (like in https://github.com/ros2-rust/ros2_rust) and make the latter depend on the former.

We made 2 distinct crates for zenoh specific purpose:
 - `zplugin-dds` is a [zenoh-specific plugin](https://zenoh.io/docs/manual/plugins) format: a dynamic library loaded by the [zenoh router](https://zenoh.io/docs/getting-started/key-concepts/#zenoh-router).
 - `zenoh-bridge-dds` run the exact same code than the plugin, but as a standalone zenoh application (in [peer mode](https://zenoh.io/docs/getting-started/key-concepts/#peer-application) or in [client mode](https://zenoh.io/docs/getting-started/key-concepts/#client-application)). Rather than dynamically linking or loading the plugin, we made a static link for several reasons:
    - Even if Rust supports the creation of dynamic libraries, their usage is really constrained because [Rust doesn't have a stable ABI](https://people.gnome.org/~federico/blog/rust-stable-abi.html#rust_does_not_have_a_stable_abi). Leading to crash if some type's memory representation differ between the library and the executable. Making it work requires to have the exact same dependencies and Rust compiler for the lib and the exe.
    - It's simpler to deploy a standalone exe that doesn't require any lib except the basic ones (libc, libpthread...)

In ROS2 context, we don't think a zenoh router will be deployed within a robot. It makes more sense to deploy a standalone `zenoh-bridge-dds` that might connect to an external zenoh router.

JEnoch:
Following comments, I close this PR. It's replaced by #30688

rolling/distribution.yaml
hidmic: Thanks for the clarification.

PR #30687, opened 2021-09-03
Adding amr_interop_bridge to documentation index for melodic

kokosabu:

noetic
melodic

# The source is here: 

https://github.com/LexxPluss/amr_interop_bridge

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -168,6 +168,12 @@ repositories:
       url: https://github.com/AinsteinAI/ainstein_radar.git
       version: master
     status: maintained
+  amr_interop_bridge:
+    source:
+      type: git
+      url: https://github.com/LexxPluss/amr_interop_bridge.git
+      version: main
+    status: developed
   angles:
     doc:
       type: git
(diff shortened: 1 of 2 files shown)


sloretz (approved):

kokosabu:
The error log shows that there was a possibility that the error was caused by unregistered noetic, so noetic was added. I've tested it with melodic, but it should probably work with noetic as well.

hidmic (commented):
@kokosabu may I suggest a more descriptive name for the packages in the `amr_interop_bridge` repository? IMHO `interop_bridge` is a bit generic and vague for a name. Perhaps an `amr_` prefix would do.

kokosabu:
> @kokosabu may I suggest a more descriptive name for the packages in the `amr_interop_bridge` repository? IMHO `interop_bridge` is a bit generic and vague for a name. Perhaps an `amr_` prefix would do.

@hidmic Thanks for the suggestion. The package name has been changed as follows.

- `interop_bridge` -> `amr_interop_bridge`
- `interop_msgs` -> `amr_interop_msgs`

wjwwood (approved):

PR #30688, opened 2021-09-03
zenoh-bridge-dds: 0.5.0-2 in 'rolling/distribution.yaml' [bloom]

JEnoch:
Increasing version of package(s) in repository `zenoh-bridge-dds` to `0.5.0-2`:

- upstream repository: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
- release repository: https://github.com/atolab/zenoh-bridge-dds-release.git
- bloom version: `0.10.7`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4463,5 +4463,21 @@ repositories:
       url: https://github.com/ros2/yaml_cpp_vendor.git
       version: master
     status: maintained
+  zenoh-bridge-dds:
+    doc:
+      type: git
+      url: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
+      version: master
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/atolab/zenoh-bridge-dds-release.git
+      version: 0.5.0-2
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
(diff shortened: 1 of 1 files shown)


rolling/distribution.yaml
```diff
@@ -4463,5 +4463,21 @@ repositories:
       url: https://github.com/ros2/yaml_cpp_vendor.git
       version: master
     status: maintained
+  zenoh-bridge-dds:
```
sloretz, line 4466: Small problem, [REP 144](https://ros.org/reps/rep-0144.html) only allows package names to use underscores `_`, not dashes `-`. Would you be willing to rename the package inside the repository? I don't think the repository name on github needs to change, but you may need to use underscores too when running bloom after changing the name in the `CMakeLists.txt` and `package.xml`.

`bloom-release --rosdistro rolling --track rolling zenoh_bridge_dds`

sloretz (changes requested):

rolling/distribution.yaml
JEnoch: OK, I see.
I can't rename the source repo. But for coherency, I'll rename the release repo according to the package name: `zenoh_bridge_dds-release`. Then I'll `bloom-release` again which will probably create another PR. Thus I'll close this one.

That's unfortunate the `ros2 pkg create` command doesn't print at least some warning for this 😕

JEnoch:
Following comments, I close this PR. It's replaced by #30720

PR #30690, opened 2021-09-03
add rom_msgs package to documentation index for noetic

ROM-robotics:
I'd like my rom_msgs package to be indexed and document on ros.org.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -6604,24 +6604,10 @@ repositories:
     status: maintained
   rom_msgs:
     doc:
-      depends:
-      - actionlib
-      - robot_state_publisher
-      - std_msgs
-      type: git
-      url: https://github.com/greenghostman/rom_msgs.git
-      version: master
-    release:
-      packages:
-      - rom_motor_msgs
-      tags:
-        release: release/noetic/{package}/{version}
-      url: https://github.com/greenghostman/rom_msgs.git
-      version: 0.1.1-1
     source:
       type: git
       url: https://github.com/greenghostman/rom_msgs.git
-      version: master
+      version: main
     status: developed
   ros:
     doc:
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -6602,6 +6602,27 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/robotis_manipulator.git
       version: noetic-devel
     status: maintained
+  rom_msgs:
+    doc:
+      depends:
+      - actionlib
+      - robot_state_publisher
+      - std_msgs
+      type: git
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: master
+    release:
+      packages:
+      - rom_motor_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: 0.1.1-1
+    source:
+      type: git
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: master
+    status: developed
   ros:
     doc:
       type: git


noetic/distribution.yaml
```diff
@@ -6602,6 +6602,27 @@ repositories:
...
+      type: git
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: master
+    release:
+      packages:
+      - rom_motor_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: 0.1.1-1
```
sloretz, line 6620:
```suggestion
```

Please remove the `release` block. It's only used when doing a binary release.

noetic/distribution.yaml
```diff
@@ -6602,6 +6602,27 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/robotis_manipulator.git
       version: noetic-devel
     status: maintained
+  rom_msgs:
+    doc:
+      depends:
+      - actionlib
+      - robot_state_publisher
+      - std_msgs
```
sloretz, line 6610:
```suggestion
```

Please remove this `depends` block.

noetic/distribution.yaml
```diff
@@ -6602,6 +6602,27 @@ repositories:
...
+      type: git
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: master
```
sloretz, line 6613:
```suggestion
      version: main
```

I don't see a `master` branch on the `rom_msg` repo; I think this should be [`main`](https://github.com/GreenGhostMan/rom_msgs/branches).

noetic/distribution.yaml
```diff
@@ -6602,6 +6602,27 @@ repositories:
...
+      packages:
+      - rom_motor_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: 0.1.1-1
+    source:
+      type: git
+      url: https://github.com/greenghostman/rom_msgs.git
+      version: master
```
sloretz, line 6624:
```suggestion
      version: main
```

sloretz (changes requested):
Please add a `LICENSE` file to the `main` branch of the `rom_msgs` repo and also put the license name in the `package.xml`. It's currently set as [`<license>TODO</license>`](https://github.com/GreenGhostMan/rom_msgs/blob/ad88b9f5780b01d583df88417957d5d0f73ab088/rom_motor_msgs/package.xml#L16). See [this repo's `LICENSE` file](https://github.com/ros-controls/ros_control/blob/noetic-devel/LICENSE)  and [its `package.xml`](https://github.com/ros-controls/ros_control/blob/53c2487d1b56a40f1d00b06c49512aa7fd2bf465/transmission_interface/package.xml#L14) as an example.

I also notice the `rom_motor_msgs` `CMakeLists.txt` and `package.xml` files list a lot of dependencies, but the `.msg` definitions themselves don't have any. I would recommend using `std_msgs` as a template for your `package.xml` and `CMakeLists.txt` so users who download this repo and use `rosdep` to install its dependencies don't get a bunch of extra software they don't need installed.

https://github.com/ros/std_msgs

noetic/distribution.yaml
GreenGhostMan: ok done!

hidmic:
@GreenGhostMan I see you're still missing a LICENSE file. There's also a dependency on `roscpp` and `rospy` that is not such?

noetic/distribution.yaml
GreenGhostMan: done

noetic/distribution.yaml
GreenGhostMan: done

noetic/distribution.yaml
```diff
@@ -6602,6 +6602,13 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/robotis_manipulator.git
       version: noetic-devel
     status: maintained
+  rom_msgs:
+    doc:
```
tfoote, line 6606: The doc entry should not be empty it needs the same entries as source

PR #30698, opened 2021-09-04
Adding innopolis_vtol_dynamics to documentation index for noetic and melodic

frontw:
# Please Add These Packages to be indexed in the rosdistro.

innopolis_vtol_dynamics
uavcan_communicator
inno_sim_interface

(melodic, noetic)

# The source is here: 

https://github.com/InnopolisAero/innopolis_vtol_dynamics
https://github.com/InnopolisAero/uavcan_communicator
https://github.com/InnopolisAero/inno_sim_interface

# Checks
 - [+] All packages have a declared license in the package.xml
 - [+] This repository has a LICENSE file
 - [+] This package is expected to build on the submitted rosdistro

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4687,6 +4687,16 @@ repositories:
       type: git
       url: https://github.com/inertialsense/inertial_sense_ros.git
       version: master
+  innopolis_vtol_dynamics:
+    doc:
+      type: git
+      url: https://github.com/InnopolisAero/innopolis_vtol_dynamics.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/InnopolisAero/innopolis_vtol_dynamics.git
+      version: master
+    status: maintained
   interactive_marker_proxy:
     doc:
       type: git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -4687,6 +4687,16 @@ repositories:
       type: git
       url: https://github.com/inertialsense/inertial_sense_ros.git
       version: master
+  inno_sim_interface:
+    doc:
+      type: git
+      url: https://github.com/InnopolisAero/inno_sim_interface.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/InnopolisAero/inno_sim_interface.git
+      version: master
+    status: maintained
   innopolis_vtol_dynamics:
     doc:
       type: git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13746,6 +13746,16 @@ repositories:
       url: https://github.com/osrf/uav_testing.git
       version: master
     status: maintained
+  uavcan_communicator:
+    doc:
+      type: git
+      url: https://github.com/InnopolisAero/uavcan_communicator.git
+      version: master
+    source:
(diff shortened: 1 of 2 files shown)


melodic/distribution.yaml
```diff
@@ -4687,6 +4687,26 @@ repositories:
...
       url: https://github.com/inertialsense/inertial_sense_ros.git
       version: master
+  inno_sim_interface:
+    doc:
+      type: git
+      url: https://github.com/InnopolisAero/inno_sim_interface.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/InnopolisAero/inno_sim_interface.git
```
hidmic, line 4697: @frontw please add a LICENSE file at the root of the repository and update its package.xml, which still shows `<license>TODO</license>` for license metadata.  

I've also noticed that dependencies declared in the package.xml do not match dependencies actually used either by the CMakeLists.txt or the [`innosim_relay.launch`](https://github.com/InnopolisAero/inno_sim_interface/blob/master/launch/innosim_relay.launch) file. Consider cleaning things up a bit.

melodic/distribution.yaml
```diff
@@ -4687,6 +4687,26 @@ repositories:
...
+      version: master
+    status: maintained
+  innopolis_vtol_dynamics:
+    doc:
+      type: git
+      url: https://github.com/InnopolisAero/innopolis_vtol_dynamics.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/InnopolisAero/innopolis_vtol_dynamics.git
```
hidmic, line 4707: @frontw [`innopolis_vtol_dynamics`](https://github.com/InnopolisAero/innopolis_vtol_dynamics/tree/master/uav_dynamics/inno_vtol_dynamics)'s package.xml also needs an update: it declares an MIT license on a GPLv3 licensed repository, it does not reflect all dependencies actually used.

hidmic: There's also something to be said about a repository this full of submodules, with ROS packages themselves, some from third-party sources. IIRC those will not be fetched during package scrapping, but I'm not so sure about the licensing. CC @clalancette @tfoote for policy in these cases.

melodic/distribution.yaml
```diff
@@ -13736,6 +13756,16 @@ repositories:
...
       version: master
     status: maintained
+  uavcan_communicator:
+    doc:
+      type: git
+      url: https://github.com/InnopolisAero/uavcan_communicator.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/InnopolisAero/uavcan_communicator.git
```
hidmic, line 13766: @frontw same comment about package.xml files needing an update.

melodic/distribution.yaml
tfoote: Yes, please submit the actual upstream package individually instead of a meta repository with a lot of submodules so that everyone can be aware of where sources are coming from upon inspection w/o having to resolve and traverse the submodules. (There are also tools in our toolchain which will fail on submodules too so that's better too.)

ivanpauno:
@frontw I will close the PR for the moment, when it's ready for review again feel free to reopen it or open a new one.

PR #30712, opened 2021-09-06
Adding dynamixel_interface to documentation index for melodic and noetic

TJMolnar:

Melodic
Noetic

# The source is here: 

https://github.com/csiro-robotics/dynamixel_interface/tree/noetic-devel

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2605,6 +2605,16 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/dynamixel-workbench-msgs.git
       version: melodic-devel
     status: developed
+  dynamixel_interface:
+    doc:
+      type: git
+      url: https://github.com/csiro-robotics/dynamixel_interface
+      version: noetic-devel
+    source:
+      type: git
+      url: https://github.com/csiro-robotics/dynamixel_interface
+      version: noetic-devel
+    status: maintained
   dynamixel_sdk:
     doc:
       type: git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2608,11 +2608,11 @@ repositories:
   dynamixel_interface:
     doc:
       type: git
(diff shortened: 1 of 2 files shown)


melodic/distribution.yaml
```diff
@@ -2605,6 +2605,16 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/dynamixel-workbench-msgs.git
       version: melodic-devel
     status: developed
+  dynamixel_interface:
+    doc:
+      type: git
+      url: https://github.com/csiro-robotics/dynamixel_interface.git
+      version: noetic-devel
```
hidmic, line 2612: @TJMolnar `noetic-devel` for Melodic?

hidmic (commented):
I've noticed this package is requiring C++17 at the CMakeLists.txt level. Bear in mind that any exported includes should be C++14 compliant, as that's the target C++ version for both Melodic and Noetic.

melodic/distribution.yaml
TJMolnar: @hidmic I have updated the noetic-devel branch to be targeting c++14 explicitly. The branch is specifically compatible with both noetic and melodic so I went with the latest version as a branch name, please advise if a different convention is preferred.

TJMolnar:
@hidmic can you provide some clarity on why nosetests are failing? They are passing on my local checkout of my fork so I'm not sure what needs fixing.

wjwwood:
It kind of looks like this needs to be rebased.

@mergifyio rebase

mabelzhang (commented):
Checks are passing.
LICENSE looks okay to me, even though it isn't one of the [OSI Approved Licenses](https://opensource.org/licenses). I don't know enough about licensing to say for sure, so I've asked for internal input.

In the meantime, anyone else more confident about licensing can approve and merge.

ivanpauno (approved):

PR #30763, opened 2021-09-11
Adding python2 and 3 J1939 library

mr337:

## Package name:

j1939

## Package Upstream Source:

https://github.com/benkfra/j1939
https://pypi.org/project/j1939/ (There are a few similarly named libraries but this is registered as j1939 in pypi index)

## Purpose of using this:

We are using this library as a convenience wrapped to talk to ECUs and a few other devices via J1939.

Distro packaging links:

  - not available
   - not available
  - not available
  - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2353,7 +2353,7 @@ python-itsdangerous:
   ubuntu:
     '*': null
     bionic: [python-itsdangerous]
-python-j1939-pip:
+python-j1939-pip: &migrate_eol_2025_04_30_python3_j1939_pip
   debian:
     pip:
       packages: [j1939]
@@ -6679,19 +6679,7 @@ python3-interpreter:
   rhel: [python3]
   slackware: [python3]
   ubuntu: [python3-minimal]
-python3-j1939-pip:
-  debian:
-    pip:
-      packages: [j1939]
-  fedora:
-    pip:
-      packages: [j1939]
-  gentoo:
-    pip:
-      packages: [j1939]
-  ubuntu:
-    pip:
-      packages: [j1939]
+python3-j1939-pip: *migrate_eol_2025_04_30_python3_j1939_pip
 python3-jinja2:
   alpine: [py3-jinja2]
   debian: [python3-jinja2]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2353,6 +2353,19 @@ python-itsdangerous:
   ubuntu:
     '*': null
     bionic: [python-itsdangerous]
+python-j1939-pip:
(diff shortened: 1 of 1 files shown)


wjwwood (approved):

rosdep/python.yaml
```diff
@@ -6590,6 +6603,19 @@ python3-interpreter:
   rhel: [python3]
   slackware: [python3]
   ubuntu: [python3-minimal]
+python3-j1939-pip:
```
jacobperron, line 6606: Should we use an anchor/alias instead of repeated the same information for both Python 2 and 3? See [CONTRIBUTING](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1).

jacobperron: @wjwwood What do you think?

wjwwood: Oh yeah, I knew you could do that but I didn't realize it was recommended in the contributing file. It would be good to use the anchor.

mr337: Thanks for the input, I didn't realize Anchors was preferred as well. I have pushed new anchor changes.

mr337:
A little confused about the failed nosetests

`Package 'python-itsdangerous' could not be found for ubuntu focal on amd64` even though it is true

```
$ lsb_release -a              
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal
$ aptitude search itsdangerous
p   python-itsdangerous-doc                                                                           - Various helpers to pass trusted data to untrusted environment - doc                                        
p   python3-itsdangerous                                                                              - Various helpers to pass trusted data to untrusted environment - Python 3.x  
```

I guess the linter picked up on that as part of the diff since??? Any recommendation? I am happy to remove that for focal on the package `python-itsdangerous` but seems like that should be addressed in a new PR. 
```
 Error: Package 'python-itsdangerous' could not be found for ubuntu focal on amd64
Detecting changed rules with 'git diff --unified=0 unittest_upstream_comparision/master -- rosdep'
{'python-itsdangerous': {'ubuntu': ['python-itsdangerous']},
 'python3-j1939-pip': {'ubuntu': {'pip': {'packages': ['j1939']}}}}
```

jacobperron:
> I guess the linter picked up on that as part of the diff

Yeah, it's not related to this PR. I've proposed a fix in https://github.com/ros/rosdistro/pull/30763#issuecomment-926192678
I think this PR is good to go, thanks!

jacobperron (approved):

wjwwood (approved):

PR #30773, opened 2021-09-13
neo_simulation2: 1.0.0-2 in 'galactic/distribution.yaml' [bloom]

padhupradheep:
Increasing version of package(s) in repository `neo_simulation2` to `1.0.0-2`:

- upstream repository: https://github.com/neobotix/neo_simulation2.git
- release repository: https://github.com/neobotix/neo_simulation2-release.git
- bloom version: `0.10.7`

### DIFF
--- a/galactic/distribution.yaml
+++ b/galactic/distribution.yaml
@@ -1794,6 +1794,21 @@ repositories:
       url: https://github.com/ros-planning/navigation_msgs.git
       version: galactic
     status: maintained
+  neo_simulation2:
+    doc:
+      type: git
+      url: https://github.com/neobotix/neo_simulation2.git
+      version: galactic
+    release:
+      tags:
+        release: release/galactic/{package}/{version}
+      url: https://github.com/neobotix/neo_simulation2-release.git
+      version: 1.0.0-2
+    source:
+      type: git
+      url: https://github.com/neobotix/neo_simulation2.git
+      version: galactic
+    status: maintained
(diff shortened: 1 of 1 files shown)


mjeronimo:
In the source repo, the version number in setup.py (0.0.0) doesn't match that in the package.xml (1.0.0). Also, a couple TODO items in setup.py.

padhupradheep:
Thanks for the comment. Is it okay if we do this update for the next release ? Because, we did not actually update the setup.pay for the release in rolling which was made last week.

wjwwood:
@padhupradheep I would recommend addressing it now, doing a `1.0.1` release, as having the versions mismatched can cause issues.

padhupradheep:
Alright, thanks ! I will let you know once it is updated

padhupradheep:
Updated the version number. Let me know if there is something that I need to do further!

jacobperron:
@padhupradheep I think @wjwwood is suggesting to make a new release (`1.0.1`) and closing this PR. Then the release repo will also contain the change to the setup.py.

padhupradheep:
I also ran the bloom once again, so that the changes are applied to the release repos (Just skipped the PR). I thought that there will be a problem, once when we update the package to "1.0.1". 

> @padhupradheep I think @wjwwood is suggesting to make a new release (`1.0.1`) and closing this PR. Then the release repo will also contain the change to the setup.py.

Okay, I will then do it.

PR #30861, opened 2021-09-24
Disable PR and commit testing for MoveIt 2

gavanderhoorn:
These jobs keep failing / timing out, and we already have CI based on GHA over at `ros-planning/moveit2` (targetting the different active ROS 2 versions, including Rolling).

As MoveIt builds can take a *really* long time (without the extensive caching we do at `ros-planning/moveit2`), this PR suggests to save resources by removing the `source` entries for Foxy, Galactic and Rolling respectively, so the buildfarm can do something else.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2056,6 +2056,8 @@ repositories:
       url: https://github.com/moveit/moveit2-release.git
       version: 2.2.1-1
     source:
+      test_commits: false
+      test_pull_requests: false
       type: git
       url: https://github.com/ros-planning/moveit2.git
       version: main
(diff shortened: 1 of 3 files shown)


jacobperron (approved):
Fine by me. I'd like a second opinion.

mikeferguson:
Would it not be better to set:

```
      test_commits: false
      test_pull_requests: false
```

For both entries - that way the source can still be found by automated tools

gavanderhoorn:
Good suggestion @mikeferguson.

Updated.

audrow (approved):

jacobperron (approved):
LGTM with a +1 from a moveit2 maintainer, maybe @henningkayser?

henningkayser (approved):
Good idea to just disable the build jobs explicitly instead of removing the entries altogether. LGTM

PR #30862, opened 2021-09-24
Switch rolling urdfdom to use the master branch

clalancette:
We've merged all of the necessary changes to the master branch,
so we should start using that instead of the ros2 branch.


clalancette:
We've now switched to the master branch everywhere else, so going ahead and merging this one.

PR #30891, opened 2021-09-29
geometry_tutorials: Remove turtle_tf2_cpp from package builds.

nuclearsandwich:
This package is failing to build on Rolling.

A future release of geometry_tutorials, which hopefully addresses the
build failure, will re-enable turtle_tf2_cpp.

The issue is ticketed upstream as https://github.com/ros/geometry_tutorials/issues/64


mabelzhang:
(I'm adding the label so that I don't merge by mistake.)

PR #30895, opened 2021-09-30
Add rviz_camera_stream to noetic

lucasw:

noetic

# The source is here: 

https://github.com/lucasw/rviz_camera_stream

# Checks

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -8046,6 +8046,21 @@ repositories:
       url: https://github.com/ros-visualization/rviz_animated_view_controller.git
       version: noetic-devel
     status: maintained
+  rviz_camera_stream:
+    doc:
+      type: git
+      url: https://github.com/lucasw/rviz_camera_stream.git
+      version: noetic-devel
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/lucasw/rviz_camera_stream-release.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/lucasw/rviz_camera_stream.git
+      version: noetic-devel
+    status: maintained
   rviz_satellite:
     doc:
       type: git


lucasw:
It's failing on

```
AssertionError: ["In file '''/home/runner/work/rosdistro/rosdistro/noetic/distribution.yaml''': Could not validate source entry for repo rviz_camera_stream with error [[[Could not validate repository with url https://github.com/lucasw/rviz_camera_stream.git and version 1.0.1 from entry at line 8059. Error reason: Tags are not valid, but a tag 1.0.1 was found. Re: https://github.com/ros/rosdistro/pull/20286]]]"] is not false
```

Did I enter the version into bloom incorrectly?

audrow:
I'm not sure why it shows different versions for the docs and release, maybe try redoing this? Also, is there any reason that you didn't use bloom?

lucasw:
> Also, is there any reason that you didn't use bloom?

I got tripped up with the github authentication, couldn't figure out why I couldn't use git@github... for urls involved with pushing instead of https://... and having to jump through extra authentication steps.  But I can try it again.  (It would be really nice if it were possible to specify all the options in a config file instead of the interactive bash script- or maybe it is but the main documentation steered me away from that?)

lucasw:
> I'm not sure why it shows different versions for the docs and release

I figured it out from the other entries in the distibution.yaml, the `version` for doc and source is the branch name, not a tag like it is with `release`.

mabelzhang:
You should be able to use the SSH URL. Have you tried this? http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization
We recommend using Bloom, and it would be good for future releases if you get it set up correctly now.

lucasw:
> You should be able to use the SSH URL

I ran into this initially

https://github.com/ros-infrastructure/bloom/blob/cca7ec80c0cb2234e280bb23741b5fe02345db0d/bloom/commands/release.py#L179-L190

but then later in the process it asks about a push url, and I thought the early url goes through validation and has to be https as required by the above code, but then the push url could be git@github.com- but it didn't work, possibly something else I manually typed in tripped it up.  Does a file get saved after all the interactive steps, I could look at a working example or show what I already tried, or even better just hand edit that and avoid the text entry altogether?

mabelzhang:
Bloom by default pushes with HTTPS, which is what the page I linked above does; it reroutes HTTPS through SSH.
When you say initially it didn't work, which step on that page did you run into an error, and what was the error message?

PR #30920, opened 2021-10-04
REVIEWER_GUIDELINES combine bullets 2 and 3

mabelzhang:
Documentation tweak.
Because REP 144 applies to both binary and source/doc packages, but it's only listed in the former.
Combining the binary and source/doc bullets to reduce confusion.

Please read revised entry to make sure it's still clear that a PR of any combo of bin/src/doc is valid.
I didn't change the wording, so none of that meaning should have changed. I only removed the bulleted list in the src/doc section and said to look at the binary section.


REVIEW_GUIDELINES.md
```diff
@@ -36,7 +36,7 @@ There are a few different types of pull requests that are opened against this re
...
 
 1.  A new binary package in a ROS distribution.  An example of this kind of PR is [25857](https://github.com/ros/rosdistro/pull/25857).  An entry is considered a new binary package if it has a `release` field.  These PRs should generally have been opened via bloom; if they aren’t, the reviewer will ask why the submitter hasn't used bloom.  There are two sub-categories of this kind of PR:
```
clalancette, line 38: I wonder if we can get rid of bullet point 3 completely by just rewording this introduction.  Maybe something like:

```suggestion
1.  A new binary, source, or documentation package in a ROS distribution.  An example of this kind of PR is [25857](https://github.com/ros/rosdistro/pull/25857).  In this type of PR, any combination of (`release`, `doc`, and `source`) fields are permitted.  If an entry has a `release` field, the PR should generally have been opened via bloom; if it wasn't, the reviewer will ask why.  There are two sub-categories of this kind of PR:
```

REVIEW_GUIDELINES.md
```diff
@@ -46,17 +46,7 @@ There are a few different types of pull requests that are opened against this re
...
-        * The license must be reflected in the package.xml file of all sub-packages in the repository.
-        * The source repository must be publicly accessible.
-        * The source repository should contain one or more ROS packages (meaning they have a `package.xml` in the source repository). Packages that are not ROS packages can be accepted, but they are rare and require special handling in the release repository.
-        * All the package names should be compliant with [REP 144](https://ros.org/reps/rep-0144.html).
-
-    Once the above criteria are satisfied, and the ROS distribution isn’t in a "sync freeze", then the PR will be merged.
+1.  A new source or documentation package in a ROS distribution.  An example of this kind of PR is [26383](https://github.com/ros/rosdistro/pull/26383).  An entry is considered a source or documentation package if it has a `source` or `doc` field (or both), but no `release` field.  These PRs need not have been opened with bloom.  The reviewer is to see the bullet above for binary packages; the same rules apply here, except this kind of PRs will not have a release repository.
```
clalancette, line 49: If we accept the above suggestion, then this bullet point would be completely removed.

mabelzhang: Removed in ff2a2c5

clalancette (approved):

PR #30935, opened 2021-10-07
add aravis-tools to rosdep

sgermanserrano:
Signed-off-by: Servando German Serrano <servando.german.serrano@gmail.com>

## Package name:

aravis-tools

  - https://packages.debian.org/bullseye/aravis-tools
   - https://packages.ubuntu.com/focal/aravis-tools

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -110,11 +110,11 @@ apt-transport-https:
   debian: [apt-transport-https]
   ubuntu: [apt-transport-https]
 aravis:
-  debian: [libaravis-0.6-0, aravis-tools, aravis-tools-cli]
+  debian: [libaravis-0.6-0, aravis-tools]
   gentoo: [media-video/aravis]
   nixos: [aravis]
   ubuntu:
-    '*': [libaravis-0.6-0, aravis-tools, aravis-tools-cli]
+    '*': [libaravis-0.6-0, aravis-tools]
     bionic: null
     xenial: null
 aravis-dev:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -125,6 +125,12 @@ aravis-dev:
     '*': [libaravis-dev]
     bionic: null
     xenial: null
+aravis-tools:
+  debian: [aravis-tools]
+  ubuntu:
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -125,6 +125,12 @@ aravis-dev:
     '*': [libaravis-dev]
     bionic: null
     xenial: null
+aravis-tools:
+  debian: [aravis-tools]
+  ubuntu:
+    '*': [aravis-tools]
+    bionic: null
+    xenial: null
```
clalancette, line 133: You can drop the `xenial` key now; Xenial is End-of-Life.

clalancette (changes requested):
Minor nit, but otherwise looks good to me.

rosdep/base.yaml
sgermanserrano: @clalancette  amended

clalancette (approved):

rosdep/base.yaml
```diff
@@ -125,6 +125,11 @@ aravis-dev:
     '*': [libaravis-dev]
     bionic: null
     xenial: null
+aravis-tools:
+  debian: [aravis-tools]
+  ubuntu:
+    '*': [aravis-tools]
+    bionic: null
```
hidmic, line 132: @sgermanserrano hmm, I have the impression that `aravis-tools` and `aravis-tools-cli` packages in Debian can be found as a single `aravis` package in other distros (along with the rest). 

Would it make sense to collapse this entry into the existing `aravis` key and complete it? @clalancette for feedback.

ivanpauno: I think that modifying the already existing `aravis` rule so that it also includes ararvis-tools and aravis-tools-cli in all distros is fine.
We could also have four rules, `aravis`, `aravis-dev`, `aravis-tools-cli`, `aravis-tools`, but that seems like too many.

sgermanserrano: sorted, I've collapse them into `aravis`

ivanpauno (approved):

ivanpauno (approved):

sloretz (approved):

PR #30975, opened 2021-10-13
Fix some macOS dependencies

kneep:
Hi all. It seems building on and for macOS is not officially supported in the latest release. Some packages are missing for ```osx```, while a few others are special because they are already shipped with macOS or Xcode. At least these changes need to be made to make ```rosdep``` happy.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1441,6 +1441,9 @@ google-mock:
   nixos: [gtest]
   openembedded: [gtest@meta-oe]
   opensuse: [gmock]
+  osx:
+    homebrew:
+      packages: [googletest]
   rhel: [gmock-devel]
   ubuntu: [google-mock]
 gpac:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -546,9 +546,6 @@ clang-tidy:
   fedora: [clang-tools-extra]
   gentoo: [sys-devel/clang]
   nixos: [clang]
-  osx:
-    homebrew:
-      packages: [llvm]
   rhel:
     '*': [clang-tools-extra]
     '7': null
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7701,6 +7701,9 @@ wget:
   nixos: [wget]
   openembedded: [wget@openembedded-core]
   opensuse: [wget]
+  osx:
+    homebrew:
+      packages: [wget]
   ubuntu: [wget]
 wireless-tools:
   debian: [wireless-tools]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7617,7 +7617,7 @@ usbutils:
   openembedded: [usbutils@openembedded-core]
   osx:
     homebrew:
-      packages: [mikhailai/misc/usbutils]
+      packages: [lsusb]
   ubuntu: [usbutils]
 util-linux:
   arch: [util-linux]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5611,9 +5611,6 @@ libxml2-utils:
   gentoo: [dev-libs/libxml2]
   nixos: [libxml2]
   openembedded: [libxml2@openembedded-core]
-  osx:
-    homebrew:
-      packages: [libxml2]
   rhel: [libxml2]
   ubuntu: [libxml2-utils]
 libxmlrpc-c++:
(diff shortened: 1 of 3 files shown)


rosdep/base.yaml
```diff
@@ -546,9 +546,6 @@ clang-tidy:
   fedora: [clang-tools-extra]
   gentoo: [sys-devel/clang]
   nixos: [clang]
-  osx:
-    homebrew:
-      packages: [llvm]
```
ivanpauno, line 551: why is this deleted?

IIUC, llvm includes clang-tidy when installed

rosdep/base.yaml
```diff
@@ -5608,9 +5608,6 @@ libxml2-utils:
   gentoo: [dev-libs/libxml2]
   nixos: [libxml2]
   openembedded: [libxml2@openembedded-core]
-  osx:
-    homebrew:
-      packages: [libxml2]
```
ivanpauno, line 5613: why is this getting deleted?

rosdep/osx-homebrew.yaml
```diff
@@ -154,7 +158,7 @@ graphviz:
 gtest:
   osx:
     homebrew:
-      packages: [gtest]
+      packages: [googletest]
```
ivanpauno, line 161: why is this repeated in base and here? I'm not sure how that works

kneep: @ivanpauno  I have the same doubt. I see other macOS packages randomly scattering in these two files. Maybe someone should clean all these up.

rosdep/base.yaml
ivanpauno: > So to make clang-tidy work, we need to manually put it to $PATH.

Which one? The one provided by Xcode?
If that's the case, removing the rule sounds fine, but if installing `llvm` through homebrew is still needed I wouldn't delete the rule.

kneep: @ivanpauno I mean the homebrew one. Xcode does not include a ```tidy-clang```. But even after installing ```llvm``` from homebrew, ```rosdep``` still fails because it query the installation with 
```
brew info llvm --json=v1
```
and the check the ```linked_keg``` in the output. For ```keg-only``` packages, ```linked_keg``` key is null so ```rosdep``` fails. You can take a look at this code:
https://github.com/ros-infrastructure/rosdep/blob/9f5829869db6c28d3cb0a536c699b82919a4c358/src/rosdep2/platforms/osx.py#L207

It seems currently there is no easy way to automatically install ```clang-tidy```. As long as we install it with homebrew (either by ```rosdep``` or manually), we have to manually link the executable to ```$PATH```. But removing ```llvm``` from ```base.yaml``` will at least make ```rosdep``` happy and maybe instructions should be added to documentations.

ivanpauno: that sounds fair enough to me

ivanpauno (approved):

ivanpauno (approved):

rosdep/python.yaml
```diff
@@ -6904,6 +6910,9 @@ python3-mypy:
   nixos: [python3Packages.mypy]
   openembedded: [python3-mypy@meta-ros-common]
   opensuse: [mypy]
+  osx:
+    pip:
```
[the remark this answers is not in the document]
sloretz: I don't see any guidelines on which is preferred, so I'd say this is fine as is.

rosdep/python.yaml
```diff
@@ -8270,6 +8291,9 @@ python3-venv:
   fedora: [python3-libs]
   nixos: [python3]
   opensuse: [python3-virtualenv]
+  osx:
+    pip:
+      packages: [virtualenv]
```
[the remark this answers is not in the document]
sloretz: > So leave this entry blank?

Leaving blank sounds right to me!

tfoote:
@kneep Do you have a moment to resolve the final comments on this?

rosdep/osx-homebrew.yaml
```diff
@@ -583,6 +595,10 @@ python-vtk:
   osx:
     homebrew:
       depends: [libvtk]
+python3:
+  osx:
+    homebrew:
+      packages: []
```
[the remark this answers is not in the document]
kneep: I'll remove this patch in the push. I might get this wrong.

kneep:
@tfoote Hi Tully. I just resolved all the conversations and pushed revised patches, removing the python3 patch and leaving the virtualenv blank according to the discussion with @sloretz. Thanks.

clalancette (approved):

PR #31001, opened 2021-10-16
Update rosbridge_suite package list in Foxy, Galactic, and Rolling

jtbandes:
New packages were added to the rosbridge_suite metapackage in https://github.com/RobotWebTools/rosbridge_suite/pull/665.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -4206,10 +4206,12 @@ repositories:
     release:
       packages:
       - rosapi
+      - rosapi_msgs
       - rosbridge_library
       - rosbridge_msgs
       - rosbridge_server
       - rosbridge_suite
+      - rosbridge_test_msgs
       tags:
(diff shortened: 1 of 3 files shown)


jacobperron (commented):
It looks like the new packages were not released, from CI:
```
RuntimeError: foxy: missing package.xml file for package "rosapi_msgs"
foxy: missing package.xml file for package "rosbridge_test_msgs"
galactic: missing package.xml file for package "rosapi_msgs"
galactic: missing package.xml file for package "rosbridge_test_msgs"
rolling: missing package.xml file for package "rosapi_msgs"
rolling: missing package.xml file for package "rosbridge_test_msgs"
```

clalancette:
Also, it is not a good idea to open a PR targeting multiple ROS distributions at once.  Since they are all on separate release schedules, it is hard to find a time when they are all unfrozen.

PR #31058, opened 2021-10-26
[dep] Update: ruamel.yaml for newer Ubuntu distro

130s:
The package is [available on Ubuntu 20.04 (packages.ubuntu.com)](https://packages.ubuntu.com/source/focal/ruamel.yaml) and on newer distros. But the current rosdep key doesn't seem to cover those distros.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4391,6 +4391,7 @@ python-ruamel.yaml:
   fedora: [python-ruamel-yaml]
   nixos: [pythonPackages.ruamel_yaml]
   ubuntu:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -4391,6 +4391,7 @@ python-ruamel.yaml:
   fedora: [python-ruamel-yaml]
   nixos: [pythonPackages.ruamel_yaml]
   ubuntu:
+    '*': [python-ruamel.yaml]
```
sloretz, line 4394: Since it's not available on focal, is there a replacement package that should be used instead?

PR #31099, opened 2021-11-02
Adding sciurus17 to documentation index for noetic

ShotaAk:
I'd like my repository to be indexed and documented on ros.org

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -8335,6 +8335,12 @@ repositories:
       url: https://github.com/ipa320/schunk_modular_robotics.git
       version: kinetic_dev
     status: maintained
+  sciurus17:
+    doc:
+      type: git
+      url: https://github.com/rt-net/sciurus17_ros.git
+      version: master
+    status: maintained
   sdhlibrary_cpp:
     doc:
       type: git


noetic/distribution.yaml
```diff
@@ -8335,6 +8335,12 @@ repositories:
       url: https://github.com/ipa320/schunk_modular_robotics.git
       version: kinetic_dev
     status: maintained
+  sciurus17:
+    doc:
+      type: git
+      url: https://github.com/rt-net/sciurus17_ros.git
```
clalancette, line 8341: The LICENSE in the repository is a non-open source license.  That means it is unclear to us on whether we can distribute the content in it, so I would be extremely hesitant to take in a `source` or `release` entry for it.

But this is strictly a `doc` entry, so it is less problematic.  @tfoote what's the policy here?

ShotaAk: I see.. I also would like to know the policy.
By the way, the PR for Melodic has been merged. (There doesn't seem to be any discussion about it.)
https://github.com/ros/rosdistro/pull/22111

Tiryoh: Thanks for your review.
We had this discussion before about another NON-COMMERCIAL-LICENSE robot.
https://github.com/ros/rosdistro/pull/19477

clalancette (approved):

PR #31108, opened 2021-11-02
Adding swri-console to foxy release

danthony06:
We would like to release swri-console for foxy. We're switching to a new ros2-devel branch to release from, and are also trying to switch to the ros2-gbp release repo.

## Package name:

swri-console

## Package Upstream Source:

https://github.com/swri-robotics/swri_console/tree/ros2-devel

## Purpose of using this:

We have previously released for dashing and multiple ROS 1 versions, and would like to release for foxy, galactic, and rolling now.

Distro packaging links:

  - REQUIRED
   - REQUIRED

foxy

# The source is here: 

https://github.com/swri-robotics/swri_console/tree/ros2-devel

# Checks

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -5150,6 +5150,22 @@ repositories:
       url: https://github.com/open-rmf/stubborn_buddies.git
       version: foxy
     status: developed
+  swri_console:
+    doc:
+      type: git
+      url: https://github.com/swri-robotics/swri_console.git
+      version: ros2-devel
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/ros2-gbp/swri_console-release.git
+      version: 2.0.0-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/swri-robotics/swri_console.git
+      version: ros2-devel
+    status: developed
   system_metrics_collector:
     doc:
       type: git
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -5155,11 +5155,6 @@ repositories:
       type: git
       url: https://github.com/swri-robotics/swri_console.git
       version: ros2-devel
-    release:
-      tags:
-        release: release/foxy/{package}/{version}
-      url: https://github.com/ros2-gbp/swri_console-release.git
-      version: 2.0.0-1
     source:
       test_pull_requests: true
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -5150,6 +5150,22 @@ repositories:
...
+  swri_console:
+    doc:
+      type: git
+      url: https://github.com/swri-robotics/swri_console.git
+      version: ros2-devel
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/ros2-gbp/swri_console-release.git
+      version: 2.0.0-1
```
nuclearsandwich, line 5162:
```suggestion
```

There's not currently any foxy metadata in the ros2-gbp release repository for swri_console-release. If there's foxy content in a prior release repository I can help migrate it to the ros2-gbp repo but I would suggest waiting to add release info until it can be generated by a `bloom-release` automated PR.

Alternatively you can enter all of this data during a `bloom-release` automated PR and close this manual one.

nuclearsandwich (dismissed):

foxy/distribution.yaml
danthony06: @nuclearsandwich I'm having troubles running bloom-release wit this, I think possibly because we're switching from the swri-release-gbp repo with dashing, to ros2-gbp with foxy and later.

nuclearsandwich: > I'm having troubles running bloom-release wit this, I think possibly because we're switching from the swri-release-gbp repo with dashing, to ros2-gbp with foxy and later.

I just double checked the ros2-gbp repository and noticed that the default branch was not set to `master` it's possible that this was at least part of the problem. I also noticed that this release repository also makes use of the `release_repo_url` track configuration field and that could be part of the problem as well.

It looks like the same issue as https://github.com/swri-robotics/marti_messages/issues/112#issuecomment-950020835 could be affecting this repository. Because the foxy track is not yet created it's probably being copied from a previous release and including the `release_repo_url` field pointing to the swri-robotics-gbp repository.

If you open an issue on https://github.com/ros2-gbp/swri_console-release with details from your bloom error and @-mention me I can take a closer look.

tfoote (approved):
I'm going to merge this with the reduced source and doc content now that the release section is removed. It should be added back automatically by bloom.

PR #31111, opened 2021-11-02
Add pigpiod

felixdivo:

## Package name:

`pigpio`

## Package Upstream Source:

[Github: joan2937/pigpio](https://github.com/joan2937/pigpio)

## Purpose of using this:

It is a GPIO tool available on the [Ubuntu flavors for the Raspberry Pi](https://www.raspberrypi.com/software/operating-systems/). It allows for remote access and easy GPIO interaction.

Is this the correct way of specifying such a dependency?

  - REQUIRED -> None?
   - REQUIRED -> Only on raspberry-pi flavours
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6453,6 +6453,8 @@ php:
     xenial: [php7.0]
     yakkety: [php7.0]
     zesty: [php7.0]
+pigpiod:
+  ubuntu: [pigpiod]
 pigz:
   debian: [pigz]
   fedora: [pigz]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6453,8 +6453,8 @@ php:
     xenial: [php7.0]
     yakkety: [php7.0]
     zesty: [php7.0]
-pigpiod:
-  ubuntu: [pigpiod]
+pigpio:
+  ubuntu: [pigpio]
 pigz:
   debian: [pigz]
   fedora: [pigz]


tfoote (changes requested):
This package does not appear on Ubuntu https://packages.ubuntu.com/search?suite=impish&section=all&arch=any&keywords=pigpio&searchon=names

Nor debian: https://packages.debian.org/search?keywords=pigpio&searchon=names&suite=bullseye&section=all

These are required platforms for submission here to a general databse. Entries here are not useful if they cannot be resolved. 

If there's a specific platform for which you're working you may want to consider maintaining a set of rosdep rules for that platform independently. But platform specific packages here just clutter things up and break expectations.

felixdivo:
That's sad. There isn't a raspberry pi platform yet, right? Seems like it's not worth the hassle creating a new one.

clalancette:
> That's sad. There isn't a raspberry pi platform yet, right? Seems like it's not worth the hassle creating a new one.

No, there isn't a raspberry pi specific platform.  In general, we require that the packages here be available via either the standard Ubuntu, Fedora, Debian, etc, repositories, or via pip.  We don't typically allow packages in the database from third-party repositories.

Because of that, I'm going to close this out.  Do feel free to keep commenting or reopen if you have other thoughts, comments or questions here.  Thanks for the contribution.

PR #31124, opened 2021-11-04
add python-google api bionic, focal support

mqcmd196:

## Package name:

python-googleapi

## Purpose of using this:

I want to install this package with ubuntu ros melodic, noetic environment

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6553,6 +6553,7 @@ python3-gnupg:
 python3-google-cloud-storage-pip: *migrate_eol_2025_04_30_python3_google_cloud_storage_pip
 python3-google-cloud-texttospeech-pip: *migrate_eol_2025_04_30_python3_google_cloud_texttospeech_pip
 python3-googleapi:
+  debian: [python3-googleapi]
   ubuntu:
     bionic: [python3-googleapi]
     focal: [python3-googleapi]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1995,7 +1995,7 @@ python-googleapi:
   debian: [python-googleapi]
   gentoo: [dev-python/google-api-python-client]
   ubuntu:
-    bionic: [google-api-python-client]
+    bionic: [python-googleapi]
     trusty: [python-googleapi]
     utopic: [python-googleapi]
     vivid: [python-googleapi]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1988,6 +1988,9 @@ python-googleapi:
   fedora: [google-api-python-client]
   gentoo: [dev-python/google-api-python-client]
   ubuntu:
+    bionic: [python-googleapi]
+    bionic_python3: [python3-googleapi]
+    focal_python3: [python3-googleapi]
     trusty: [python-googleapi]
     utopic: [python-googleapi]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1988,6 +1988,8 @@ python-googleapi:
   fedora: [google-api-python-client]
   gentoo: [dev-python/google-api-python-client]
   ubuntu:
+    bionic: [python-googleapi]
+    focal: [python3-googleapi]
```
tfoote, line 1992: Following our conventions this key is for Python 2 /any. We should not mix in the explicit python3 version here. It should get a new key `python3-googleapi`

tfoote (changes requested):
I see the python2 version tops out at bionic: https://packages.ubuntu.com/bionic/python-googleapi

And the python3 version starts at least at bionic: https://packages.ubuntu.com/impish/python3-googleapi

rosdep/python.yaml
mqcmd196: Thanks, I'll fix

mqcmd196: fixed

rosdep/python.yaml
```diff
@@ -1988,6 +1988,9 @@ python-googleapi:
   fedora: [google-api-python-client]
   gentoo: [dev-python/google-api-python-client]
   ubuntu:
+    bionic: [python-googleapi]
+    bionic_python3: [python3-googleapi]
+    focal_python3: [python3-googleapi]
```
clalancette, line 1993: This style is deprecated, and we don't use it anymore.  Please make an entirely new rosdep key called `python3-googleapi`, and populate it with these entries.

clalancette (changes requested):

rosdep/python.yaml
jacobperron: @mqcmd196 Friendly ping. Please move these Python3 packages to a new rosdep key as Chris suggests.

mqcmd196: sorry,,, I fixed.

clalancette (changes requested):
It looks like this package is also available for Debian under the same name, so please add a section for Debian as well.

rosdep/python.yaml
```diff
@@ -6551,6 +6552,11 @@ python3-gnupg:
   ubuntu: [python3-gnupg]
 python3-google-cloud-storage-pip: *migrate_eol_2025_04_30_python3_google_cloud_storage_pip
 python3-google-cloud-texttospeech-pip: *migrate_eol_2025_04_30_python3_google_cloud_texttospeech_pip
+python3-googleapi:
+  debian: [python3-googleapi]
+  ubuntu:
+    bionic: [python3-googleapi]
+    focal: [python3-googleapi]
```
clalancette, line 6559: Sorry, I missed this on the last review.  There is no need to split this up, so just make this:

```suggestion
  ubuntu: [python3-googleapi]
```

clalancette (changes requested):

rosdep/python.yaml
mqcmd196: Thank you for your kind review, fixed

clalancette (approved):
Thanks for iterating, looks good to me.

ivanpauno (approved):

PR #31170, opened 2021-11-08
Adding Astrobee repo to index

marinagmoreira:
kinetic/melodic/noetic

# The source is here: 
https://github.com/nasa/astrobee

# Checks
 - [yes] All packages have a declared license in the package.xml
 - [yes] This repository has a LICENSE file
 - [yes] This package is expected to build on the submitted rosdistro

### DIFF
--- a/kinetic/distribution.yaml
+++ b/kinetic/distribution.yaml
@@ -695,6 +695,11 @@ repositories:
       url: https://github.com/orbbec/ros_astra_launch.git
       version: master
     status: developed
+  astrobee:
+    doc:
+      type: git
+      url: https://github.com/nasa/astrobee
+      version: master
   astuff_sensor_msgs:
     doc:
       type: git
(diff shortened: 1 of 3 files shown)


kinetic/distribution.yaml
```diff
@@ -695,6 +695,11 @@ repositories:
       url: https://github.com/orbbec/ros_astra_launch.git
       version: master
     status: developed
+  astrobee:
+    doc:
+      type: git
+      url: https://github.com/nasa/astrobee
+      version: master
```
clalancette, line 702: Kinetic is end-of-life, so we aren't allowing any further changes to it.  Please remove this section.

clalancette (changes requested):

kinetic/distribution.yaml
marinagmoreira: done

clalancette (approved):

PR #31201, opened 2021-11-12
Add python3-qwt

mabelzhang:
Eh, I hope I did this right.

## Package name:

python3-qwt

## Package Upstream Source:

https://github.com/PierreRaybaut/PythonQwt

## Purpose of using this:

`qwt_dependency` does nothing for Kinetic and above:
https://github.com/ros-visualization/qwt_dependency/blob/73bc532699b6aa219f9c842d6cc88cf65d9f398f/package.xml#L11

Once added to `qwt_dependency` for Noetic, this will enable the use of qwt with Qt5 on Focal, which will make fixes like this work in Noetic https://github.com/ros-visualization/rqt_plot/pull/70
(It doesn't help Melodic, because on Bionic, `python3-qwt` is for Qt4)

Looks like `python3-qwt` only exists on Debian and Ubuntu: https://pkgs.org/download/python3-qwt
So on the other platforms it has to come from pip: https://pypi.org/project/PythonQwt/

  - https://packages.debian.org/bookworm/python3-qwt
   - https://packages.ubuntu.com/focal/python3-qwt

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7743,13 +7743,14 @@ python3-qtpy:
   ubuntu: [python3-qtpy]
 python3-qwt:
   debian: [python3-qwt]
+  ubuntu: [python3-qwt]
+python3-qwt-pip:
   fedora:
     pip:
       packages: [PythonQwt]
   opensuse:
     pip:
       packages: [PythonQwt]
-  ubuntu: [python3-qwt]
 python3-rafcon-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7745,10 +7745,16 @@ python3-qwt:
   debian: [python3-qwt]
   ubuntu: [python3-qwt]
 python3-qwt-pip:
+  debian:
+    pip:
+      packages: [PythonQwt]
   fedora:
     pip:
       packages: [PythonQwt]
   opensuse:
+    pip:
+      packages: [PythonQwt]
+  ubuntu:
     pip:
       packages: [PythonQwt]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -7691,6 +7691,15 @@ python3-qtpy:
...
   nixos: [python3Packages.qtpy]
   ubuntu: [python3-qtpy]
+python3-qwt:
+  debian: [python3-qwt]
+  fedora:
+    pip:
+      packages: [PythonQwt]
+  opensuse:
+    pip:
+      packages: [PythonQwt]
```
jacobperron, line 7701: I think the preference is to put pip keys in a separate entry (`python3-qwt-pip`) to avoid collisions if the package becomes available on these platforms in the future. [Reference](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-rules).

jacobperron: @wjwwood please advise what you think the right thing to do here is.

wjwwood: Agreed.

mabelzhang: Thanks! Split off into `-pip` in 9a78bfe.
Do the ones with system packages need to be repeated with their pip packages in the `-pip` entry?

jacobperron: I think that wouldn't hurt and is good for completeness.

mabelzhang: Added in c8e5f72

wjwwood (approved):

jacobperron (approved):

PR #31204, opened 2021-11-12
[dep] deepdiff-pip for Python3

130s:
## Package name:

deepdiff

## Package Upstream Source:

https://pypi.org/project/deepdiff/

Py2 version is already added in https://github.com/ros/rosdistro/pull/15606

## Purpose of using this:

- Help Python application to deal with dict, iterables

# Distro packaging links:

  - REQUIRED
  - https://packages.debian.org/search?keywords=%20deepdiff
   - REQUIRED
   - https://packages.ubuntu.com/search?keywords=deepdiff&searchon=names&suite=all&section=all
  - IF AVAILABLE
  - https://src.fedoraproject.org/rpms/python-deepdiff
  - IF AVAILABLE
  - https://archlinux.org/packages/community/any/python-deepdiff/
  - IF AVAILABLE
  - Not found
  - IF AVAILABLE
  - Not found
- OpenSuse https://software.opensuse.org/package/python3-deepdiff

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6195,6 +6195,16 @@ python3-decorator:
   osx:
     pip: [decorator]
   ubuntu: [python3-decorator]
+python3-deepdiff:
+  arch: [python-deepdiff]
+  debian: [python3-deepdiff]
+  fedora: [python-deepdiff]
+  opensuse: [python3-deepdiff]
+  ubuntu:
+    '*': [python3-deepdiff]
+    bionic:
+      pip:
+        packages: [deepdiff]
 python3-defusedxml:
   debian: [python3-defusedxml]
   fedora: [python3-defusedxml]


rosdep/python.yaml
```diff
@@ -6195,6 +6195,13 @@ python3-decorator:
   osx:
     pip: [decorator]
   ubuntu: [python3-decorator]
+python3-deepdiff-pip:
```
clalancette, line 6198: So it looks like this actually has a system package as of Debian Buster: https://packages.debian.org/search?keywords=deepdiff&searchon=names&suite=all&section=all .  It also looks like it has a package in Ubuntu since 20.04: https://packages.ubuntu.com/search?keywords=deepdiff&searchon=names&suite=all&section=all .  I didn't check Fedora, but I think we should switch to using the system-packaged versions instead.

clalancette (changes requested):

rosdep/python.yaml
130s: Ah, indeed! Commit updated, as well as the OP description copied below for email notification purpose but will be deleted after the post).

----

## Package name:

deepdiff

## Package Upstream Source:

https://pypi.org/project/deepdiff/

Py2 version is already added in https://github.com/ros/rosdistro/pull/15606

## Purpose of using this:

- Help Python application to deal with dict, iterables

# Distro packaging links:

  - REQUIRED
  - https://packages.debian.org/search?keywords=%20deepdiff
   - REQUIRED
   - https://packages.ubuntu.com/search?keywords=deepdiff&searchon=names&suite=all&section=all
  - IF AVAILABLE
  - https://src.fedoraproject.org/rpms/python-deepdiff
  - IF AVAILABLE
  - https://archlinux.org/packages/community/any/python-deepdiff/
  - IF AVAILABLE
  - Not found
  - IF AVAILABLE
  - Not found

rosdep/python.yaml
```diff
@@ -6195,6 +6195,15 @@ python3-decorator:
   osx:
     pip: [decorator]
   ubuntu: [python3-decorator]
+python3-deepdiff:
+  arch: [python-deepdiff]
+  debian: [deepdiff]
```
clalancette, line 6200: This should be `python3-deepdiff` for Debian.

clalancette (dismissed):
Looks like the Debian key is wrong, and that we could add an OpenSUSE key.

130s:
@clalancette Thanks for catching. Addressed Debian and OpenSuse.

jacobperron (approved):

wjwwood (approved):
lgtm

PR #31260, opened 2021-11-18
gpu-voxels-release: 0.8.3-1 in 'melodic/distribution.yaml' [bloom]

BV-OpenSource:
Increasing version of package(s) in repository `gpu-voxels-release` to `0.8.3-1`:

- upstream repository: https://github.com/BV-OpenSource/gpu-voxels.git
- release repository: https://github.com/BV-OpenSource/gpu-voxels-release.git
- bloom version: `0.10.7`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3905,7 +3905,7 @@ repositories:
       url: https://github.com/swri-robotics/gps_umd.git
       version: master
     status: developed
-  gpu-voxels-release:
+  gpu_voxels:
     doc:
       type: git
       url: https://github.com/BV-OpenSource/gpu-voxels.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3905,6 +3905,23 @@ repositories:
       url: https://github.com/swri-robotics/gps_umd.git
       version: master
     status: developed
+  gpu-voxels-release:
+    doc:
+      type: git
+      url: https://github.com/BV-OpenSource/gpu-voxels.git
+      version: melodic-dev
+    release:
+      packages:
+      - gpu_voxels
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -3905,6 +3905,23 @@ repositories:
       url: https://github.com/swri-robotics/gps_umd.git
       version: master
     status: developed
+  gpu-voxels-release:
```
jacobperron, line 3908: This name does not follow packaging naming rules ([REP-144](https://www.ros.org/reps/rep-0144.html)). We should use underscores (`_`) instead of dashes (`-`). Also, consider dropping the `-release` suffix, which doesn't add to the package name IMO.

jacobperron (changes requested):

melodic/distribution.yaml
wjwwood: So, I'm not sure what our policy is on _repository_ names, but package names definitely need to use `_`.

And to clear up any confusion this is supposed to be the name of the source repository, but including the `-release` suffix makes it seem like you're confusing it with the release repository.

Either way, I'd recommend following @jacobperron's suggestion and using `gpu_voxels` here.

BV-OpenSource: Changing the name cause an error on validation. Can you help me by detecting why?

jacobperron: I think you should re-bloom the package (with the updated name). I think you can do this with the `--edit` option when running `bloom-release`.

BV-OpenSource: That generated another pull-request #31305.

clalancette:
Closing this in favor of #31305.

PR #31263, opened 2021-11-18
[rosdep] add python3-sila2lib-pip

fmessmer:
this package is only available from PyPi:
https://pypi.org/project/sila2lib/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7915,6 +7915,19 @@ python3-shapely:
       packages: [shapely]
   rhel: [python3-shapely]
   ubuntu: [python3-shapely]
+python3-sila2lib-pip:
+  debian:
+    pip:
+      packages: [sila2lib]
+  fedora:
+    pip:
+      packages: [sila2lib]
+  opensuse:
+    pip:
+      packages: [sila2lib]
+  ubuntu:
+    pip:
+      packages: [sila2lib]
 python3-simple-pid-pip:
   debian:
     pip:


jacobperron (approved):

rosdep/python.yaml
```diff
@@ -4608,6 +4608,19 @@ python-shapely:
     wily_python3: [python3-shapely]
     xenial_python3: [python3-shapely]
     yakkety_python3: [python3-shapely]
+python-sila2lib-pip:
```
clalancette, line 4611: All of our new pip keys should be of the form `python3-sila2lib-pip`.

clalancette (changes requested):

jacobperron (approved):

rosdep/python.yaml
```diff
@@ -7915,6 +7915,19 @@ python3-shapely:
...
+python3-sila2lib-pip:
+  debian:
+    pip:
+      packages: [sila2lib]
+  fedora:
+    pip:
+      packages: [sila2lib]
+  gentoo:
+    pip:
+      packages: [sila2lib]
```
mabelzhang, line 7927: We're not supposed to use pip for gentoo. See https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md where it says "Note that pip cannot be used to install system packages on Gentoo."

However, we do encourage OpenSUSE. See https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md where it says "Ubuntu and Debian are required, Fedora, Gentoo, and openSUSE are encouraged if the package also exists."

So, we could just swap out the `gentoo` entry for `opensuse`.

```suggestion
  opensuse:
    pip:
      packages: [sila2lib]
```

fmessmer: @mabelzhang 
done and rebased

mabelzhang (approved):

PR #31297, opened 2021-11-24
Adding outsight_alb_driver to the ros index 

EclOutsight:
Hello rosdistro-Team,

We would like to add to the index a package to interact with our Augmented Lidar Box (https://www.outsight.ai/)

Thanks in advance

* Melodic
* Noetic

# The source is here: 

https://gitlab.com/outsight-public/outsight-drivers/outsight_alb_driver.git

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -8332,6 +8332,16 @@ repositories:
       url: https://github.com/CPFL/ouster.git
       version: autoware_branch
     status: developed
+  outsight_alb_driver:
+    doc:
+      type: git
+      url: https://gitlab.com/outsight-public/outsight-drivers/outsight_alb_driver.git
+      version: master
+    source:
+      type: git
+      url: https://gitlab.com/outsight-public/outsight-drivers/outsight_alb_driver.git
+      version: master
+    status: maintained
   oxford_gps_eth:
     doc:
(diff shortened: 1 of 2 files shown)


mabelzhang (commented):
Have you tried opening this PR with [Bloom](http://wiki.ros.org/bloom)? That is the recommended way. More details are at https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

In addition, it is preferred to have 1 PR per distro - 1 for Melodic and 1 for Noetic.

EclOutsight:
Correct me if I am wrong, but I thought Bloom was for releasing a binary package, but here it's just to be added in the index, as explained [here](http://wiki.ros.org/rosdistro/Tutorials/Indexing%20Your%20ROS%20Repository%20for%20Documentation%20Generation).
Before releasing with Bloom, the package needs to be listed in the index, right ?

If it's better to have on PR for each distro, then sure I will do it

mabelzhang:
Ah, sorry I missed that this isn't a binary release.

But yeah 1 PR per distro would be better, because for example, right now Noetic is in sync, Melodic isn't, so we can merge the Melodic one, but the Noetic one would have to wait.

EclOutsight:
Ok thanks !

PR #31305, opened 2021-11-25
gpu_voxels: 0.8.3-2 in 'melodic/distribution.yaml' [bloom]

BV-OpenSource:
Increasing version of package(s) in repository `gpu_voxels` to `0.8.3-2`:

- upstream repository: https://github.com/BV-OpenSource/gpu-voxels.git
- release repository: https://github.com/BV-OpenSource/gpu-voxels-release.git
- bloom version: `0.10.7`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3905,6 +3905,17 @@ repositories:
       url: https://github.com/swri-robotics/gps_umd.git
       version: master
     status: developed
+  gpu_voxels:
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/BV-OpenSource/gpu-voxels-release.git
+      version: 0.8.3-2
+    source:
+      type: git
+      url: https://github.com/BV-OpenSource/gpu-voxels.git
+      version: melodic-dev
+    status: developed
   graceful_controller:
     doc:
       type: git


wjwwood:
It looks like multiple licenses are being used but only one is listed in the `package.xml`. You can have more than one `<license>` tag. Should there be more than one in there?

BV-OpenSource:
@wjwwood You meant the package.xml in the release repository, right? In case positive, I think it's fixed.

mabelzhang:
The [guidelines](https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md) don't answer it explicitly, but if I'm interpreting section 2.ii correctly, the source repo might matter more than the release repo, e.g. "The LICENSE file only needs to be in the sources; a new release after adding a LICENSE file is not required." and "one or more ROS packages (meaning they have a package.xml in the source repository)."

Unless someone comments with an actual answer, it doesn't hurt to add it to the source repo.

wjwwood:
I don't think you need a `package.xml` in the source repository, there are lots of projects where this isn't an option. It would be better if there was one upstream, but it's not required as far as I know.

wjwwood:
@BV-OpenSource it looks like you still have just one license entry:

https://github.com/BV-OpenSource/gpu-voxels-release/blob/405e999cc230d03db054633f1f59481cb1590dc1/package.xml#L12-L15

If you all you did was update the file in `master`, then you need to run bloom again afterwards.

wjwwood:
It will create `0.8.3-3` and unless you cancel it, it will open a new pull request too, which is fine.

BV-OpenSource:
New pull-request #31358 opened.

PR #31348, opened 2021-12-01
added python3-pytrinamic-pip to /rosdep/python.yaml

eugenkaltenegger:

## Package name:

python3-pytrinamic-pip

## Package Upstream Source:

[https://github.com/trinamic/PyTrinamic](https://github.com/trinamic/PyTrinamic)

## Purpose of using this:

This dependency is used to communicate with TRINAMIC Modules and Evaluation boards (such as motors and encoders).

Distro packaging links:

[https://pypi.org/project/PyTrinamic/](https://pypi.org/project/PyTrinamic/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7596,6 +7596,16 @@ python3-pytest-timeout:
   gentoo: [dev-python/pytest-timeout]
   opensuse: [python3-pytest-timeout]
   ubuntu: [python3-pytest-timeout]
+python3-pytrinamic-pip:
+  debian:
+    pip:
+      packages: [pytrinamic]
+  fedora:
+    pip:
+      packages: [pytrinamic]
+  ubuntu:
+    pip:
+      packages: [pytrinamic]
 python3-pyvista-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7603,6 +7603,9 @@ python3-pytrinamic-pip:
   fedora:
     pip:
       packages: [pytrinamic]
+  opensuse:
+    pip:
+      packages: [pytrinamic]
   ubuntu:
     pip:
       packages: [pytrinamic]


audrow (approved):

mabelzhang (commented):
Could you add a pip entry for `opensuse` since that's [encouraged now](https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md)?

Nit: I don't know if we require package names to be lowercase. pypi is case insensitive, but since both links officially list the package installation command to be with case `PyTrinamic`, maybe it's more consistent to use the cased name. I don't know if we really care though.

rosdep/python.yaml
```diff
@@ -7596,6 +7596,16 @@ python3-pytest-timeout:
...
+python3-pytrinamic-pip:
+  debian:
+    pip:
+      packages: [pytrinamic]
+  fedora:
+    pip:
+      packages: [pytrinamic]
+  ubuntu:
+    pip:
+      packages: [pytrinamic]
```
nuclearsandwich, line 7608:
```suggestion
python3-pytrinamic-pip:
  debian:
    pip:
      packages: [pytrinamic]
  fedora:
    pip:
      packages: [pytrinamic]
  opensuse:
    pip:
      packages: [pytrinamic]
  ubuntu:
    pip:
      packages: [pytrinamic]
```

kaltenegger: **opensuse:** I added opensuse as requested. @mabelzhang @nuclearsandwich  Thanks for the hint.

**PyTrinamic vs pytrinamic:** @mabelzhang I have had a look at the PyYAML package ([PyPI](https://pypi.org/project/PyYAML/), [pyyaml.org](https://pyyaml.org/)). This package also contains upper case letters in its name and is in the package name with lower case letters. An other example would be `PyVirtualSerialPorts` or `Adafruit-Blinka`, again both habe upper case letters in the name, lower case letters in the package name.
Therefore I would state that lower case letters in the package name are the more consistent choice within the `rosdep.yaml` file.

kaltenegger:
opensuse: I added opensuse as requested. @mabelzhang @nuclearsandwich Thanks for the hint.

PyTrinamic vs pytrinamic: @mabelzhang I have had a look at the PyYAML package (PyPI, pyyaml.org). This package also contains upper case letters in its name and is in the package name with lower case letters. An other example would be PyVirtualSerialPorts or Adafruit-Blinka, again both habe upper case letters in the name, lower case letters in the package name.
Therefore I would state that lower case letters in the package name are the more consistent choice within the rosdep.yaml file.

ivanpauno (approved):

PR #31393, opened 2021-12-06
add ply for python

christianrauch:

## Package name:

python-ply

## Purpose of using this:

"PLY provides most of the standard lex/yacc features including support for empty productions, precedence rules, error recovery, and support for ambiguous grammars."

This is a pip package found at https://pypi.org/project/ply/.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3330,6 +3330,10 @@ python-planar-pip:
   ubuntu:
     pip:
       packages: [planar]
+python-ply:
+  '*':
+    pip:
+      packages: [ply]
 python-plyfile-pip:
   debian:
     pip:


audrow (changes requested):
It seems that `python-ply` can be added to `rosdep/base.yaml`, which is preferred. Could you add the package there?

christianrauch:
> It seems that `python-ply` can be added to `rosdep/base.yaml`, which is preferred. Could you add the package there?

There is only a Python 2 version (https://packages.ubuntu.com/focal/python-ply). So this makes no sense for newer ROS versions, including all of ROS2.

audrow:
> There is only a Python 2 version (https://packages.ubuntu.com/focal/python-ply).

Gotcha, thanks for pointing this out.

audrow (approved):

PR #31424, opened 2021-12-08
Adding quori_ros to documentation index for distro

bmcdorman:

noetic

# The source is here: 

https://github.com/Quori-ROS/quori_ros

# Checks

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -6350,6 +6350,12 @@ repositories:
       url: https://github.com/QuanergySystems/quanergy_client_ros.git
       version: master
     status: developed
+  quori_ros:
+    doc:
+      type: git
+      url: https://github.com/Quori-ROS/quori_ros.git
+      version: master
+    status: maintained
   qwt_dependency:
     doc:
       type: git


nuclearsandwich (changes requested):
@bmcdorman thanks for updating the license entries in each package.xml. The maintainer data is still invalid as seen here:https://github.com/Quori-ROS/quori_ros/blob/cb8b2254a347d95a1528f5388ed9995ce75acc66/src/quori_controller/package.xml#L10

bmcdorman:
Maintainer information corrected.

noetic/distribution.yaml
```diff
@@ -6350,6 +6350,12 @@ repositories:
       url: https://github.com/QuanergySystems/quanergy_client_ros.git
       version: master
     status: developed
+  quori_ros:
+    doc:
+      type: git
+      url: https://github.com/Quori-ROS/quori_ros.git
+      version: master
```
hidmic, line 6357: @bmcdorman no `source` entry?

hidmic (changes requested):
@bmcdorman hmm, I see packages with TODO for a license in their `package.xml` (such as [`quori_launch`](https://github.com/Quori-ROS/quori_ros/tree/master/src/quori_launch)), and third-party packages brought in as submodules. That's not ideal.

The typical repo layout reduces to a collection of packages, your packages, at the repo root. It'd be better for that to be the case here as well.

noetic/distribution.yaml
bmcdorman: No, just looking for doc generation.

bmcdorman:
@hidmic quori_launch has been fixed. The workspace layout of the repository is an artifact of our OS distro (Qubuntu) build process. I can change it to work differently, but it's going to take some time. I can also split up every package into its own repository and have quori_ros use submodules exclusively. That way each package can be submitted in separate PRs (as I had done originally).

sloretz (approved):
License tags in `package.xml` files look OK to me. I'm not sure what the rosdoc tools will do with the submodules, but I think it'd be fine to try it as is and see what the doc job does.

PR #31432, opened 2021-12-09
add python-dt-apriltags-pip dependency for ubuntu

turniphead:

## Package name:

dt-apriltags

## Package Upstream Source:

[github source](https://github.com/duckietown/lib-dt-apriltags)

## Purpose of using this:

Useful for easy apriltag detection in python

Distro packaging links:

  - i do not have a debian machine and cannot test to make sure a similar rule would work for it.
   - [pip](https://pypi.org/project/dt-apriltags/)
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1345,6 +1345,9 @@ python-docx:
   fedora: [python-docx]
   ubuntu:
     pip: [python-docx]
+python-dt-apriltags:
+  ubuntu:
+    pip: [dt-apriltags]
 python-dxfgrabber-pip:
   ubuntu:
     pip: [dxfgrabber]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1345,6 +1345,9 @@ python-docx:
   fedora: [python-docx]
   ubuntu:
     pip: [python-docx]
+python-dt-apriltags-pip:
+  ubuntu:
```
ivanpauno, line 1349: could you include also debian, fedora, opensuse, osx?

do you want a python 2 entry or a python 3 one?
For the second, you need to add a `python3-dt-apriltags-pip` rule instead.

ivanpauno (approved):

hidmic (approved):

PR #31444, opened 2021-12-11
add python-rdflib on ubuntu focal

McModknower:

## Package name:

python-rdflib

## Package Upstream Source:

https://github.com/RDFLib/rdflib

## Purpose of using this:

python-rdflib is already in the file, however ubuntu changed the package name in focal to python-rdflib-tools

Distro packaging links:

- Ubuntu: https://packages.ubuntu.com/focal/python/python-rdflib-tools

I didn't include the other links because the package is already in the current rosdep rules.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4083,7 +4083,9 @@ python-rdflib:
   debian: [python-rdflib]
   fedora: [python-rdflib]
   gentoo: [dev-python/rdflib]
-  ubuntu: [python-rdflib]
+  ubuntu:
+    '*' : [python-rdflib]
+    focal: [python-rdflib-tools]
 python-readchar-pip:
   ubuntu:
     pip:


nuclearsandwich:
There's actually something more going on here than just a package name change.

[python-rdflib-tools](https://packages.ubuntu.com/bionic/python-rdflib-tools) is a separate package from [python-rdflib](https://packages.ubuntu.com/bionic/python-rdflib) in Bionic both of which depend on `python` (2) whereas in Focal and forward the [python-rdflib-tools](https://packages.ubuntu.com/focal/python-rdflib-tools) package depends on `python3` despite the `python-` package name prefix.

There is a `python3-rdflib` package in [bionic](https://packages.ubuntu.com/bionic/python3-rdflib), [focal](https://packages.ubuntu.com/focal/python3-rdflib), and [jammy](https://packages.ubuntu.com/jammy/python3-rdflib). So I think the actual update would be to add a new key for `python3-rdflib` as we've done for other python3 packages.

nuclearsandwich (changes requested):
Based on review of the packages in Ubuntu I don't think the proposed change represents the actual package state. rosdep keys for `python3-rdflib` exist and could be used with a `ROS_PYTHON_VERSION` package condition in tandem with the python2 keys.

If there's an explicit dependency on `python-rdflib-tools` regardless of the internal python version then a separate key for that package should be added and used. Otherwise I think this PR can be closed.

sloretz:
Thanks for the PR! Closing due to @nuclearsandwich's feedback and no response.

PR #31467, opened 2021-12-13
avt_vimba_camera: 2001.1.0-1 in 'foxy/distribution.yaml' [bloom]

icolwell-as:
Increasing version of package(s) in repository `avt_vimba_camera` to `2001.1.0-1`:

- upstream repository: https://github.com/astuff/avt_vimba_camera.git
- release repository: https://github.com/astuff/avt_vimba_camera-release.git
- bloom version: `0.10.7`

## avt_vimba_camera

```
* Port to ROS2 (#56 <https://github.com/astuff/avt_vimba_camera/issues/56>)
* Update README for ROS2 (#58 <https://github.com/astuff/avt_vimba_camera/issues/58>)
* Contributors: icolwell-as
```

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -356,6 +356,21 @@ repositories:
       url: https://gitlab.com/autowarefoundation/autoware.auto/autoware_auto_msgs.git
       version: master
     status: developed
+  avt_vimba_camera:
+    doc:
+      type: git
+      url: https://github.com/astuff/avt_vimba_camera.git
+      version: ros2_master
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/astuff/avt_vimba_camera-release.git
+      version: 2001.1.0-1
+    source:
+      type: git
+      url: https://github.com/astuff/avt_vimba_camera.git
+      version: ros2_master
+    status: maintained
   aws-robomaker-small-warehouse-world:
     doc:
       type: git


foxy/distribution.yaml
```diff
@@ -356,6 +356,21 @@ repositories:
...
+      url: https://github.com/astuff/avt_vimba_camera.git
+      version: ros2_master
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/astuff/avt_vimba_camera-release.git
+      version: 2001.1.0-1
+    source:
+      type: git
+      url: https://github.com/astuff/avt_vimba_camera.git
```
ivanpauno, line 371: could you add a top level license file to the repo?

ivanpauno (dismissed):

foxy/distribution.yaml
icolwell-as: Will do! this was originally forked from https://github.com/srv/avt_vimba_camera, which has no licence file, but the package.xml states "BSD". I'll create a BSD licence file if that sounds like the best approach.

ivanpauno: Sorry, I didn't notice you were releasing from a fork.
Do you have approval from the upstream maintainers to make a release?

In general, why prefer not allowing releases from forks.

icolwell-as: We were never able to make contact and assumed it was a dead project. See here: https://github.com/srv/avt_vimba_camera/issues/39

ivanpauno: I see.

That seems fine to me, I'll wait for an approval of a second reviewer in this case.
Maybe @sloretz or @hidmic (?).

ivanpauno (approved):

foxy/distribution.yaml
```diff
@@ -356,6 +356,21 @@ repositories:
...
+  avt_vimba_camera:
+    doc:
+      type: git
+      url: https://github.com/astuff/avt_vimba_camera.git
+      version: ros2_master
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/astuff/avt_vimba_camera-release.git
+      version: 2001.1.0-1
```
hidmic, line 368: @icolwell-as `2001` is somewhat unusual for a major version.

foxy/distribution.yaml
hidmic: I'm not sure we have a policy for these cases. @clalancette @tfoote ?

hidmic (approved):
Overall LGTM

foxy/distribution.yaml
icolwell-as: See discussion starting from here:
https://discourse.ros.org/t/ros2-transition-strategy/9154/14?u=ian_colwell
It's the best way we could think of for managing both ros1 and ros2 packages coming out of a single git repo.

foxy/distribution.yaml
clalancette: We do have a policy where we encourage people to contact the upstream and get the releases and changes done there.  However, if the original repository maintainers can't be contacted, we can release from a fork.

In this case, it does look like that was done, so :+1: from me on using this fork.

sloretz (approved):

PR #31482, opened 2021-12-15
feat: update python3-ifcfg key

wep21:
Signed-off-by: wep21 <border_goldenmarket@yahoo.co.jp>

## Package name:

python3-ifcfg

## Package Upstream Source:

https://github.com/osrf/python-ifcfg-release

## Purpose of using this:

add latest distribution

Distro packaging links:

  - REQUIRED
   - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL

ROSDISTRO NAME

# The source is here: 

http://sourcecode_url

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6633,15 +6633,11 @@ python3-hypothesis:
   gentoo: [dev-python/hypothesis]
   ubuntu: [python3-hypothesis]
 python3-ifcfg:
-  debian:
-    buster: [python3-ifcfg]
-    stretch: [python3-ifcfg]
+  debian: [python3-ifcfg]
   fedora: [python3-ifcfg]
   openembedded: [python3-ifcfg@meta-ros-common]
   rhel: ['python%{python3_pkgversion}-ifcfg']
-  ubuntu:
-    bionic: [python3-ifcfg]
-    focal: [python3-ifcfg]
+  ubuntu: [python3-ifcfg]
 python3-imageio:
   debian: [python3-imageio]
   gentoo: [dev-python/imageio]


hidmic (commented):
@wep21 hmm, I don't see these packages in [`packages.ros.org`](http://packages.ros.org/ros2/ubuntu/dists/jammy/main/binary-amd64/Packages). 

@nuclearsandwich is there an timeline for this?

rosdep/python.yaml
```diff
@@ -6634,6 +6634,7 @@ python3-hypothesis:
   ubuntu: [python3-hypothesis]
 python3-ifcfg:
   debian:
+    bullseye: [python3-ifcfg]
     buster: [python3-ifcfg]
     stretch: [python3-ifcfg]
```
clalancette, line 6639: I think that we should just change this to a generic key for all debian versions:

```
debian: [python3-ifcfg]
```

That will make it future-proof.  The same goes for the Ubuntu rule below.

clalancette (changes requested):

nuclearsandwich:
> @nuclearsandwich is there an timeline for this?

We've been trying to remove this package from the ROS bootstrap repository and push it upstream into Debian and Ubuntu.
During review, the package was dismissed and after some discussion @clalancette has been working to enhance the network configuration APIs in [psutil](https://pypi.org/project/psutil/) to provide what we need for ros2doctor and eliminate the ifcfg dependency.

In short, providing python3-ifcfg for Ubuntu Jammy and Debian Bullseye is unplanned. We're removing that dependency from ros2doctor in favor of psutil.

rosdep/python.yaml
wep21: addressed at https://github.com/ros/rosdistro/pull/31482/commits/cdc26081043a7bd054869ab7b4caeca2daf0052d.

wep21:
@nuclearsandwich I see. I close this PR.

PR #31483, opened 2021-12-15
grpc: 0.0.10-2 in 'noetic/distribution.yaml' [bloom]

BillWSY:
Increasing version of package(s) in repository `grpc` to `0.0.10-2`:

- bloom version: `0.10.7`

## grpc

```
* Follow upstream grpc 1.15.1 (#33 <https://github.com/CogRob/catkin_grpc/issues/33>)
  * Sync grpc to upstream 1.15.1
  * Typo fix
* Contributors: Shengye Wang
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2800,6 +2800,21 @@ repositories:
       url: https://github.com/anybotics/grid_map-release.git
       version: 1.6.4-1
     status: maintained
+  grpc:
+    doc:
+      type: git
+      url: https://github.com/CogRob/catkin_grpc.git
+      version: master
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/CogRobRelease/catkin_grpc-release.git
+      version: 0.0.10-2
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -2800,6 +2800,21 @@ repositories:
       url: https://github.com/anybotics/grid_map-release.git
       version: 1.6.4-1
     status: maintained
+  grpc:
```
hidmic, line 2803: @BillWSY hmm, both rosdistro entry and package name are likely to clash with (or confuse people using) a future `rosdep` key for actual `grpc` binaries.

May I suggest `catkin_grpc` or `grpc_vendor` or anything along those lines for entry and package name? It also looks a lot like the source repository has some redundant nesting (i.e. `grpc` content could be at the root of the repo).

BillWSY: Thanks for the suggestion. I prefer this name to keep it consistent with older ROS releases (we have released in indigo, kinetic, lunar, melodic). WDYT?

sloretz (approved):

PR #31513, opened 2021-12-19
use pkg-config names where available

christianrauch:
As suggested in https://github.com/ros/rosdistro/pull/31394#issuecomment-991091407, this is a proposal to use the pkg-config names as rosdep keys for packages that support them (i.e. where a *.pc file exists).

Using pkg-config names as keys has a couple of benefits:
1. they are determined by the developers of a library or component
2. are the same across distributions and operation systems
3. the rosdep key will match the dependency name when used via cmake's FindPkgConfig or meson
4. have a unique solution for mapping the rosdep key to a package name
5. the mapping can be automated (searching for the package with file `${ROSDEP_KEY}.pc`)

Since the pkg-config names are encoded in the *.pc file names and are also commonly used in cmake and other build systems, the process of extracting the keys from a build script and finding the corresponding distribution package name could be largely automated.

I did not find any conflicting convention in the documentation about how to select rosdep keys for non-python packages. To avoid conflicts with the old keys, where it was common to use the Debian/Ubuntu packages name, this rule should only apply to new rosdep key entries, where applicable (i.e. where a *.pc file exists).

Over time, keys could be used in parallel and map to the same package, and new ROS packages that are submitted to the build server could be asked to use the new pkg-config convention until eventually all official ROS packages follow this convention.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -208,6 +208,31 @@ Work has been proposed to add a separate installer for AUR packages [ros-infrast
 
 For pip installers they are expected to be in the main PyPI index https://pypi.org/.
 
+### pkg-config rules
+
+For packages that contain a [pkg-config file](https://linux.die.net/man/1/pkg-config) (\*.pc) the pkg-config name must be used as rosdep key.
+This pkg-config name is set by the developers of a library or component and will be the same across distributions.
+Commonly, the very same key will be used in CMake's [`FindPkgConfig`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) to query this dependency.
+
+Example:
+
+The Eigen3 [upstream repo](https://gitlab.com/libeigen/eigen) contains the pkg-config file [`eigen3.pc`](https://gitlab.com/libeigen/eigen/-/blob/master/eigen3.pc.in). Hence, on all systems that have Eigen3 installed, one can query information about Eigen3 using the `eigen3` key:
+```
+pkg-config --modversion eigen3
+```
+
+Querying a package that contains this pkg-config file can then be easily done by searching for the `eigen3.pc` file and should only provide a single package per distribution.
+
+Example:
+
+Searching for exact file name matches for `eigen3.pc` at https://packages.ubuntu.com (`Search the contents of packages` -> `packages that contain files named like this`) will return `libeigen3-dev` https://packages.ubuntu.com/search?suite=focal&arch=any&mode=exactfilename&searchon=contents&keywords=eigen3.pc.
+
+While this process might be automated in the future, for now the index should match this mapping of "pkg-config name" key to the "package name of the package that contains this pkg-config file":
+```yaml
+eigen3:
+  ubuntu: [libeigen3-dev]
+```
+
 ### Python rules
 
 #### Python 3


CONTRIBUTING.md
```diff
@@ -208,6 +208,29 @@ Work has been proposed to add a separate installer for AUR packages [ros-infrast
...
+The Eigen3 [upstream repo](https://gitlab.com/libeigen/eigen) contains the pkg-config file [`eigen3.pc`](https://gitlab.com/libeigen/eigen/-/blob/master/eigen3.pc.in). Hence, on all systems that have Eigen3 installed, one can query information about Eigen3 using the `eigen3` key:
+```
+pkg-config --modversion eigen3
+```
+
+Querying a package that contains this pkg-config file can then be easily done by searching for the `eigen3.pc` file and should only provide a single package per distribution.
+
+Example:
+
+Searching for excact file name matches for `eigen3.pc` at https://packages.ubuntu.com (`Search the contents of packages` -> `packages that contain files named like this`) will return `libeigen3-dev` https://packages.ubuntu.com/search?suite=focal&arch=any&mode=exactfilename&searchon=contents&keywords=eigen3.pc.
```
hidmic, line 226: @christianrauch typo:
```suggestion
Searching for exact file name matches for `eigen3.pc` at https://packages.ubuntu.com (`Search the contents of packages` -> `packages that contain files named like this`) will return `libeigen3-dev` https://packages.ubuntu.com/search?suite=focal&arch=any&mode=exactfilename&searchon=contents&keywords=eigen3.pc.
```

hidmic: Also, consider hyperlinking `libeigen3-dev` instead of tossing in the link i.e.:
```suggestion
Searching for excact file name matches for `eigen3.pc` at https://packages.ubuntu.com (`Search the contents of packages` -> `packages that contain files named like this`) will return [`libeigen3-dev`](https://packages.ubuntu.com/search?suite=focal&arch=any&mode=exactfilename&searchon=contents&keywords=eigen3.pc).
```

CONTRIBUTING.md
```diff
@@ -208,6 +208,29 @@ Work has been proposed to add a separate installer for AUR packages [ros-infrast
...
+
+While this process might be automated in future, for now the index should match this mapping of "pkg-config name" key to the "package name of the package that contains this pkg-config file":
```
hidmic, line 228: @christianrauch nit: 
```suggestion
While this process might be automated in the future, for now the index should match this mapping of "pkg-config name" key to the "package name of the package that contains this pkg-config file":
```

CONTRIBUTING.md
```diff
@@ -208,6 +208,29 @@ Work has been proposed to add a separate installer for AUR packages [ros-infrast
 
 For pip installers they are expected to be in the main PyPI index https://pypi.org/.
 
+### pkg-config rules
```
hidmic, line 211: @christianrauch hmm, what do you mean by `pkg-config rules`? This is just a guideline for `rosdep` rule key naming.

christianrauch: The idea of this PR is that if there is a pkg-config file, the rosdep key should be derived from this, so that dependency names are more standardised across distributions and build systems, and that this could be automated in future.

CONTRIBUTING.md
```diff
@@ -208,6 +208,29 @@ Work has been proposed to add a separate installer for AUR packages [ros-infrast
...
+
+For packages that contain a [pkg-config file](https://linux.die.net/man/1/pkg-config) (\*.pc) the pkg-config name must be used as rosdep key. This pkg-config name is set by the developers of a library or component and will be the same across distributions. Commonly, the very same key will be used in CMake's [`FindPkgConfig`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) to query this dependency.
```
nuclearsandwich, line 213: Although this file isn't perfect, it does generally use semantic line endings (one sentence per line) to make the diff review process easier and less prone to style/formatting changes. Could you please update the added sections to conform to one sentence per line.

```suggestion
For packages that contain a [pkg-config file](https://linux.die.net/man/1/pkg-config) (\*.pc) the pkg-config name must be used as rosdep key.
This pkg-config name is set by the developers of a library or component and will be the same across distributions.
Commonly, the very same key will be used in CMake's [`FindPkgConfig`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) to query this dependency.
```

nuclearsandwich (approved):
Thanks for following through with the proposal @christianrauch. I think it's a sound one. 

My main concern is that until automation to check whether a target package contains pkg-config files exist, checking for pkg-config files to enforce this convention adds an additional review burden. 

I'm giving my approval to the proposal but before we merge I'm going to bring it up at next week's meeting of the ROS core team in order to give others an opportunity to weigh in as well.

christianrauch:
> My main concern is that until automation to check whether a target package contains pkg-config files exist, checking for pkg-config files to enforce this convention adds an additional review burden.

Currently, submitters of new PRs are already asked to provide links to the packages of the respective distributions. They could instead be asked to provide the link to the search result for the given pkg-config file.

This could probably be automated in a script like this (only works on Ubuntu):
```
OS_VERSION=focal
KEY_NAME=eigen3
xdg-open "https://packages.ubuntu.com/search?searchon=contents&keywords=${KEY_NAME}.pc&mode=exactfilename&suite=${OS_VERSION}&arch=any"
```

On an Ubuntu/Debian system (or on CI) this can be automated using [`apt-file`](https://www.linuxuprising.com/2018/11/how-to-find-package-that-provides-file.html):
```
KEY_NAME=eigen3
sudo apt install apt-file
sudo apt-file update
PKG_NAME=$(apt-file search --package-only ${KEY_NAME}.pc)
echo "package containing ${KEY_NAME}: ${PKG_NAME}"
```

This should give you:
```
package containing eigen3: libeigen3-dev
```

tfoote (changes requested):
This is a neat and potentially valuable idea. However I don't think that as it currently stands that this will actually be a net benefit. 

It is trading one type of standardization off for another. And by having the "if it has a pc file" logic in particular, this will cause a dichotomy when people are searching for a rule. For example if you're looking for a package. do you look for the PC name or the library name. And likely even more frustrating is if you find a library for example `libfoo` and then you want the development version, (which has the pc key) it will then potentially be somewhere else in the alphabetization instead of being consistently alongside as `libfoo-dev` following debian naming conventions which are our defaults. 

I completely agree that it would be good to improve our expression of what's expected to be in the packages. I think that leveraging the much more extensive and very well established debian naming policy is still a better solution than trying to reinvent the naming policy.

christianrauch:
> It is trading one type of standardization off for another. And by having the "if it has a pc file" logic in particular, this will cause a dichotomy when people are searching for a rule. For example if you're looking for a package. do you look for the PC name or the library name.

Well, yes ... it is proposing a "new standard" for naming dependencies. But the reason for this is that the pkg-config names are unique and stable across different distributions and distribution versions and hence they can be automatically resolved to the development package name across many distributions. There is no need to maintain a huge database of something that can be automated. If a user searches for a development library, he/she is supposed to use the canonical library name as chosen by the creator of the library and not by the distribution.

> And likely even more frustrating is if you find a library for example libfoo and then you want the development version

The pkg-config files likely only resolve to development packages. I haven't seen many ROS packages that separate their build and runtime dependencies with separate rosdep keys. But then... bloom also does not build separate runtime and development packages. If you really need separate runtime packages, you can do the same automatic resolution with the shared library files, e.g. `libopencv_features2d.so`, because those will likely also be the same across distributions. And... you can get those library files via pkg-config too (`pkg-config --libs opencv4`).

> I think that leveraging the much more extensive and very well established debian naming policy is still a better solution than trying to reinvent the naming policy.

The Debian/Ubuntu naming policy is good. If Debian/Ubuntu would be the only distributions supported by rosdep, or package names would be the same across distributions, you could use the names just like that to install packages. But because rosdep supports other distributions, you have to create this database that maps between package names.

I think you all will agree, searching through the package index of distributions on a website and collecting this in a text file is a dull task for a robot and not for a human. And this has to be done over and over again for every new distribution or package source (conda, vcpkg, ...). There is so much that can be gained from automating this.

ivanpauno:
@nuclearsandwich @tfoote do you have any further feedback on the proposal?

tfoote:
It's a good idea, but we shouldn't change our policy without a full analysis of what would change and what migration paths would be required. And how wet would homogenize this with the other policies.

ivanpauno:
Considering that there hasn't been progress in monthes, @christianrauch does it make sense to open an issue with the proposal including the feedback provided here and close the PR?

tfoote:
I'm going to close this so we don't have to keep it in our triage list as it's a major change and we should be doing a full tradeoff before adopting a new divergent standard. And since this conflicts with our current standard if we were to adopt this would need to have a full migration path too for all potential keys effected. We can entertain a proposal but as this stands we can't merge this PR.

PR #31527, opened 2021-12-21
Add python-staticmap-pip and python3-staticmap-pip to rosdep/python.yaml

sktometometo:

## Package name:

- python-staticmap-pip
- python3-staticmap-pip

## Package Upstream Source:

- https://github.com/komoot/staticmap

## Purpose of using this:

This package will render a map image from map tiles (e.g. openstreetmap).
I create a map publisher with this package in order to visualize robot location.

- https://github.com/k-okada/jsk_demos/pull/108

  - pip: https://pypi.org/project/staticmap/#description
   - pip: https://pypi.org/project/staticmap/#description

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4859,13 +4859,6 @@ python-sqlalchemy:
 python-sqlite:
   debian: [python-sqlite]
   ubuntu: [python-sqlite]
-python-staticmap-pip:
-  debian:
-    pip:
-      packages: [staticmap]
-  ubuntu:
-    pip:
-      packages: [staticmap]
 python-statsd:
   debian: [python-statsd]
   fedora: [python-statsd]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4859,6 +4859,13 @@ python-sqlalchemy:
 python-sqlite:
   debian: [python-sqlite]
   ubuntu: [python-sqlite]
+python-staticmap-pip:
+  debian:
+    pip:
+      packages: [staticmap]
+  ubuntu:
+    pip:
+      packages: [staticmap]
 python-statsd:
   debian: [python-statsd]
   fedora: [python-statsd]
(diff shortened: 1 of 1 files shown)


hidmic (approved):

rosdep/python.yaml
```diff
@@ -4859,6 +4859,13 @@ python-sqlalchemy:
 python-sqlite:
   debian: [python-sqlite]
   ubuntu: [python-sqlite]
+python-staticmap-pip:
+  debian:
+    pip:
+      packages: [staticmap]
+  ubuntu:
+    pip:
+      packages: [staticmap]
```
hidmic, line 4868: @sktometometo it is somewhat unclear which Python versions does `staticmap` support.

Isn't there a better supported alternative we can use instead?

sktometometo: staticmap supports both python 2 and python 3. 
 In this case, pip package names are identical for both python version.

https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1 shows an alternative like.

```yaml
python-foobar-pip: &migrate_eol_2025_04_30_python3_foobar_pip # Anchor
  ubuntu:
    pip:
      packages: [foobar]
python3-foobar-pip: *migrate_eol_2025_04_30_python3_foobar_pip # Alias
```

Should I use this kind of format?

clalancette: So I'd generally prefer if we only add in the `python3-*-pip` key.  Do you need the Python2 version of the key?

tfoote (dismissed):
Please remove the second copy of the key for python2. The pip keys don't care about which version of python you're using. The confusing thing is that how do I associate a frame within ROS with a real sensor mounted on the system and then perform static and dynamic transforms?

The python3 prefix keeps them side by side with the native packages for ease of maintenance. And duplicate keys doubles the maintenance burden and risks them getting out of sync. 

I

jacobperron:
@sktometometo Friendly ping. Please remove the key for Python 2 following Tully's comment.

sktometometo:
@jacobperron Thank you for your remind. I have removed a key for python2.

jacobperron (approved):

mabelzhang (approved):

PR #31552, opened 2021-12-22
Update python-numpy key.

nuclearsandwich:
* Remove out-of-support Ubuntu distributions.
* This package is no longer available in Alpine Edge.
* This package is no longer available in Debian as of Bullseye.
* This package is no longer available in Ubuntu as of Jammy.


cottsay (approved):

smits:
FYI, this broke a lot of eol distro users builds.  Since rosdep has the `--include-eol-distros` flag it makes sense to keep the eol entries.

knorth55:
i dont see the reason why you removed all eol keys in this PR, just for supporting jammy.

PR #31560, opened 2021-12-23
Update gazebo keys.

nuclearsandwich:
* Remove key for gazebo5.
* Remove key for gazebo7.
* Update key for gazebo9.
  * Remove definition for out-of-support Ubuntu Artful.
  * Add definition for Ubuntu Focal.
* Update key for gazebo11.
  * Change Gentoo key to unversioned [gazebo](https://packages.gentoo.org/packages/sci-electronics/gazebo).
  * Add definition for Debian Bullseye.
  * Add definition for Ubuntu Jammy.
* Update key for gazebo.
  * Drop definition for Debian Jessie.
  * Drop definition for Debian Stretch.
  * Remove archlinux definition as it is no longer packaged.
  * Add definition for Debian Bullseye.
  * Add definition for Ubuntu Jammy.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1166,52 +1166,29 @@ gawk:
   openembedded: [gawk@openembedded-core]
   ubuntu: [gawk]
 gazebo:
-  arch: [gazebo]
   debian:
+    bullseye: [gazebo]
     buster: [gazebo11]
-    jessie: [gazebo7]
-    stretch: [gazebo9]
   fedora: [gazebo]
   gentoo: [sci-electronics/gazebo]
   nixos: [gazebo]
   slackware: [gazebo]
   ubuntu:
-    artful: [gazebo9]
     bionic: [gazebo9]
     cosmic: [gazebo9]
     disco: [gazebo9]
     eoan: [gazebo9]
     focal: [gazebo11]
-    precise: [gazebo]
-    quantal: [gazebo]
-    raring: [gazebo]
-    saucy: [gazebo2]
-    trusty: [gazebo2]
-    wily: [gazebo7]
-    xenial: [gazebo7]
-    yakkety: [gazebo7]
-    zesty: [gazebo7]
+    jammy: [gazebo]
 gazebo11:
   debian:
+    bullseye: [gazebo]
     buster: [gazebo11]
-  gentoo: [=sci-electronics/gazebo-11*]
+  gentoo: [=sci-electronics/gazebo]
   nixos: [gazebo_11]
   ubuntu:
     focal: [gazebo11]
-gazebo5:
-  arch: [gazebo]
-  gentoo: [=sci-electronics/gazebo-5*]
-  ubuntu: [gazebo5]
-gazebo7:
-  arch: [gazebo]
-  debian:
-    buster: [gazebo7]
-    jessie: [gazebo7]
-    stretch: [gazebo7]
-  gentoo: [=sci-electronics/gazebo-7*]
-  nixos: [gazebo_7]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@chapulina would you mind providing an extra set of eyes on this one?

rosdep/base.yaml
```diff
@@ -1166,52 +1166,29 @@ gawk:
...
-    precise: [gazebo]
-    quantal: [gazebo]
-    raring: [gazebo]
-    saucy: [gazebo2]
-    trusty: [gazebo2]
-    wily: [gazebo7]
-    xenial: [gazebo7]
-    yakkety: [gazebo7]
-    zesty: [gazebo7]
+    jammy: [gazebo]
```
chapulina, line 1182: I believe this will be the key on upstream Ubuntu, and we'll keep releasing `gazebo11` onto packages.osrfoundation.org. So if we decide to use more up-to-date packages, we'll need to change this to `gazebo11`

chapulina (approved):
I can only really speak for Ubuntu and Debian. @j-rivero may have some context on other distributions, but they're community-supported.

rosdep/base.yaml
nuclearsandwich: The keys here don't cover upstream packages unless they're part of the bootstrap repository.
So until there are `gazebo11` packages importable for Jammy we can't use an invalid definition.
@j-rivero do the `gazebo11` packages from the OSRF repositories use `Conflicts` or `Replaces` for the packages upstream? I'm trying to figure out what the succession strategy is or if we will need to drop gazebo_ros_packages from Rolling when we first jump to Jammy.

chapulina: There are conflicts here for different versions: https://github.com/ignition-release/gazebo11-release/blob/6492b65efa67c862f8f2d57c58ade9056fe27773/ubuntu/debian/control#L68

We may not have handled the unversioned packages yet since that's going to be a new thing in Jammy.

nuclearsandwich:
Pulled this change back into draft for now.

rosdep/base.yaml
nuclearsandwich: Got it. We should probably add that before this PR goes in.

j-rivero: We have the problem of the difference with gazebo naming between our repo and Ubuntu, I'm afraid is not too much I can do about it since Debian prefers to use non-versioned and we need versioned to difference between multiple versions supported.

Debian recommends to handle name changes using `Breaks-Replace` than can be added to our packages to override the Ubuntu packages. Other option is to create a fake transitional package named `gazebo` in our builds of gazebo11 that will pull `gazebo11` and all the friends transparently if the version is greater so we can handle using a single key `gazebo` here.

chapulina: >  Other option is to create a fake transitional package named gazebo in our builds of gazebo11 that will pull gazebo11 and all the friends transparently if the version is greater so we can handle using a single key gazebo here.

I like that. It makes it easier for users who don't know what version they should install :+1:

chapulina: > we'll keep releasing gazebo11 onto packages.osrfoundation.org

I misspoke back then. That's not true for Gazebo classic, see https://community.gazebosim.org/t/gazebo-classic-on-jammy/1248

I believe what we'll need to handle is a conflict between upstream `ligignition-math` and the one provided by packages.o.o if we want users to install Gazebo 11 from Ubuntu alongside Fortress from packages.o.o.

nuclearsandwich:
This has been superseded by https://github.com/ros/rosdistro/pull/32481

PR #31561, opened 2021-12-23
Updates to libpcl keys

nuclearsandwich:
* `libpcl-all` 
  * Add definitions for Debian Bullseye.
  * Add definitions for Ubuntu Jammy.
  * Drop definitions for out-of-support Debian distributions.
  * Drop definitions for out-of-support Ubuntu distributions.
* `libpcl-all-dev`
  * Remove out-of-support Ubuntu distributions and consolidate Ubuntu key.


clalancette (commented):
The changes to `libpcl-all-dev` seem fine; all remotely recent Ubuntu releases use that name.

I suspect the changes to `libpcl-all` are more controversial.  In particular, the removal of the `xenial` and the `stretch` keys may run in to the same objections that we've seen elsewhere.  So I'll suggest we add those back in.

tfoote (approved):
Re xenial: We talked about it in https://github.com/ros/rosdistro/issues/31569 and have guidance for how to use older tags. I'd suggest that we go ahead and remove them.

nuclearsandwich:
> In particular, the removal of the `xenial` and the `stretch` keys may run in to the same objections that we've seen elsewhere. So I'll suggest we add those back in.

> Re xenial: We talked about it in #31569 and have guidance for how to use older tags. I'd suggest that we go ahead and remove them.

I've added f8f8314 which redefines the key for xenial and once that issue is closed I'll sweep through and remove the xenial entries that I maintained. I am not doing the same for Stretch because no one has come out of the woodwork for it.

PR #31620, opened 2021-12-28
Add Pip Installation Option for Networkx

connoranderson:

## Package name:

networkx

## Package Upstream Source:

https://pypi.org/project/networkx/

## Purpose of using this:

Debians are older than what is available via pip, so some users may want access to later versions.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6948,13 +6948,13 @@ python3-networkx:
 python3-networkx-pip:
   debian:
     pip:
-      packages: [python3-networkx]
+      packages: [networkx]
   fedora:
     pip:
-      packages: [python3-networkx]
+      packages: [networkx]
   ubuntu:
     pip:
-      packages: [python3-networkx]
+      packages: [networkx]
 python3-nose:
   debian: [python3-nose]
   fedora: [python3-nose]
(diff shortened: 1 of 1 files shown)


cottsay (dismissed):
The package name as listed for each platform should match the package name as it appears on PyPI, and should therefore be `networkx`.

The key name `python3-networkx-pip` is right, though.

rosdep/python.yaml
```diff
@@ -6945,6 +6945,16 @@ python3-networkx:
   fedora: [python3-networkx]
   nixos: [python3Packages.networkx]
   ubuntu: [python3-networkx]
+python3-networkx-pip:
```
clalancette, line 6948: Why are we adding this key at all?  We already have the preferred non-pip key above, so this seems unnecessary.

clalancette (changes requested):

nuclearsandwich:
@connoranderson our policy is generally not to add pip definitions to the primary rosdep database if system packages exist for our supported and recommended platforms. Only system packages can be used in officially released ROS packages so once there is a version of a package available via the system package manager for our platforms we want to converge on the available version as a community.

There will always be motivations to rely on newer versions of packages than are available via an operating system distribution but in an ecosystem as large as the ROS community being able to rely on consistent stable releases is an overriding concern. If you have projects which are not going to be released on the official build farm you can configure additional custom rosdep sources to rely on the pip version of the package.

---

I'm going to close the pull request for the reasons above but if there's an extraordinary case here for a pip definition for this specific package we can revisit the issue.

connoranderson:
Thanks for the update, as a general policy this makes sense, though I do wonder how many developers require the latest and greatest releases. Thanks for taking the time to write that summary!

PR #31623, opened 2021-12-29
Add python3-googleapi Rule

connoranderson:

## Package name:

python3-googleapi

## Package Upstream Source:

https://src.fedoraproject.org/rpms/google-api-python-client
https://packages.debian.org/bullseye/python3-googleapi
https://packages.ubuntu.com/focal/python3-googleapi
https://packages.gentoo.org/packages/dev-python/google-api-python-client

## Purpose of using this:

Python3 version is the standard library now for this package

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6617,6 +6617,7 @@ python3-google-cloud-storage-pip: *migrate_eol_2025_04_30_python3_google_cloud_s
 python3-google-cloud-texttospeech-pip: *migrate_eol_2025_04_30_python3_google_cloud_texttospeech_pip
 python3-googleapi:
   debian: [python3-googleapi]
+  fedora: [python3-google-api-client]
   gentoo: [dev-python/google-api-python-client]
   ubuntu: [python3-googleapi]
 python3-gpiozero:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2044,11 +2044,6 @@ python-googleapi:
     vivid: [python-googleapi]
     wily: [python-googleapi]
     xenial: [python-googleapi]
-python3-googleapi:
-  debian: [python3-googleapi]
-  fedora: [google-api-python-client]
-  gentoo: [dev-python/google-api-python-client]
-  ubuntu: [python3-googleapi]
 python-gpiozero:
   debian:
     buster: [python-gpiozero]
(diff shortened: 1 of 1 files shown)


cottsay (dismissed):
It looks like there is a already a `python3-googleapi` key:
https://github.com/ros/rosdistro/blob/d573eab3a166e3aa3642b82260108608e5f468b4/rosdep/python.yaml#L6620-L6622

Feel free to add more platform rules for that key.

nuclearsandwich:
@connoranderson as cottsay pointed out, there is an existing key for python3-googleapi. The key that you've added is a duplicate and rosdep keys in each file should be lexically sorted.

tfoote:
It would be great to update the existing rule with the new target platforms found researching this.

nuclearsandwich (approved):

rosdep/python.yaml
```diff
@@ -6619,6 +6617,7 @@ python3-google-cloud-storage-pip: *migrate_eol_2025_04_30_python3_google_cloud_s
 python3-google-cloud-texttospeech-pip: *migrate_eol_2025_04_30_python3_google_cloud_texttospeech_pip
 python3-googleapi:
   debian: [python3-googleapi]
+  gentoo: [dev-python/google-api-python-client]
   ubuntu: [python3-googleapi]
```
nuclearsandwich, line 6621:
```suggestion
python3-googleapi:
  debian: [python3-googleapi]
  fedora: [python3-google-api-client]
  gentoo: [dev-python/google-api-python-client]
  ubuntu: [python3-googleapi]
```

After looking at why the Fedora package was failing I found out that the while google-api-python-client is the name of the source package, the generated noarch binary package uses the conventional python3- prefixed name. 
Here's an example from Fedora 36 https://koji.fedoraproject.org/koji/buildinfo?buildID=1864742

nuclearsandwich (commented):
Consider re-adding Fedora using the binary rather than source package name.

rosdep/python.yaml
connoranderson: Thanks for the help, was struggling to figure out why that package name wasn't working. Checked what you showed here, and confirmed I was able to install. 

> [root@0e194f62ae53 /]# yum install python3-google-api-client
Fedora 35 - x86_64                                                                                                            5.9 MB/s |  79 MB     00:13     Fedora 35 openh264 (From Cisco) - x86_64                                                                                      1.3 kB/s | 2.5 kB     00:01    
Fedora Modular 35 - x86_64                                                                                                    1.3 MB/s | 3.3 MB     00:02    
Fedora 35 - x86_64 - Updates                                                                                                  3.4 MB/s |  20 MB     00:05    
Fedora Modular 35 - x86_64 - Updates                                                                                          1.2 MB/s | 2.8 MB     00:02    
Dependencies resolved.
==============================================================================================================================================================
 Package                                               Architecture                Version                                 Repository                    Size
==============================================================================================================================================================
Installing:
 python3-google-api-client                             noarch                      2:2.32.0-1.fc35                         updates                      2.6 M

tfoote (approved):

PR #31624, opened 2021-12-29
Add pip Installation for fastapi and uvicorn

youliangtan:
## Package name:

- ~~fastapi-pip~~ python3-fastapi
- ~~uvicorn-pip~~ python3-uvicorn

## Package Upstream Source:

https://pypi.org/project/fastapi/
https://pypi.org/project/uvicorn/

## Purpose of using this:

To run a restful api server in python, with uvicorn fast ASGI server implementation. Although `uvicorn` is avail via debian, will prefer api server is running as near "latest", hence using pip. If we should follow the convention, I can make the change to `python3-uvicorn`.

Also, not sure for `fastapi`, should we can mix the installation from various sources (e.g. debian `pip`, ubuntu : `python-*` ....)

### Fastapi
- Debian: ~~NOT AVAILABLE~~ https://packages.debian.org/bullseye/python3-fastapi
- Ubuntu: https://packages.ubuntu.com/impish/python3-fastapi   (Not avail in focal)
- Fedora: https://src.fedoraproject.org/rpms/python-fastapi
- Arch: https://archlinux.org/packages/community/any/python-fastapi/

### Uvicorn
- Debian: https://packages.debian.org/bullseye/python3-uvicorn
- Ubuntu: https://packages.ubuntu.com/focal/python3-uvicorn
- Fedora: https://src.fedoraproject.org/rpms/python-uvicorn
- Arch: https://archlinux.org/packages/community/any/uvicorn/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8747,6 +8747,20 @@ yapf:
     yakkety:
       pip:
         packages: [yapf]
+univorn-pip:
+  debian:
+    pip:
+      packages: [uvicorn]
+  ubuntu:
+    pip:
+      packages: [uvicorn]
+fastapi-pip:
+  debian:
+    pip:
+      packages: [fastapi]
+  ubuntu:
+    pip:
+      packages: [fastapi]
 yapf3:
   debian: [yapf3]
   fedora: [python3-yapf]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6390,9 +6390,20 @@ python3-falcon:
   ubuntu: [python3-falcon]
 python3-fastapi:
   arch: [python-fastapi]
-  debian: [python3-fastapi]
+  debian:
+    '*': [python3-fastapi]
+    buster:
+      pip:
+        packages: [fastapi]
   fedora: [python-fastapi]
-  ubuntu: [python3-fastapi]
+  ubuntu:
+    '*': [python3-fastapi]
+    bionic:
+      pip:
+        packages: [fastapi]
+    focal:
+      pip:
+        packages: [fastapi]
 python3-fastnumbers-pip:
   arch:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8443,7 +8443,11 @@ python3-uvicorn:
   alpine: [uvicorn]
   debian: [python3-uvicorn]
   fedora: [python-uvicorn]
-  ubuntu: [python3-uvicorn]
+  ubuntu:
+    '*': [python3-uvicorn]
+    bionic:
+      pip:
+        packages: [uvicorn]
 python3-vcstool:
   alpine: [vcstool]
   debian:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -186,6 +186,16 @@ exhale-pip:
   ubuntu:
     pip:
       packages: [exhale]
+fastapi-pip:
+  debian:
+    pip:
+      packages: [fastapi]
+  fedora:
+    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -186,6 +186,16 @@ exhale-pip:
   ubuntu:
     pip:
       packages: [exhale]
+fastapi-pip:
```
clalancette, line 189: Pip-only keys should be of the form:

`python3-fastapi-pip`.  The same goes for the key below.  Also be sure to put them in alphabetical order once you've changed it.

clalancette (changes requested):

rosdep/python.yaml
```diff
@@ -8670,6 +8680,16 @@ uavcan-pip:
...
+univorn-pip:
+  debian:
+    pip:
+      packages: [uvicorn]
+  fedora:
+    pip:
+      packages: [univorn]
+  ubuntu:
+    pip:
+      packages: [uvicorn]
```
luca-della-vedova, line 8692: It seems there is a mix of `univorn` and `uvicorn`, is this intended?

youliangtan: Good catch!

clalancette (approved):

tfoote (dismissed):
I'm sorry but our policy is that if the native packages exist we should use them. As you've highlighted the pip version is newer. But there's always "newer" things out there and using the stable tested version that can be expected is much better for our community. 

Also I found that the fastapi package is available on debian too: https://packages.debian.org/bullseye/python3-fastapi

We can use pip keys to fill in older platforms that aren't yet supported by the debian packages.

youliangtan:
Thanks for reply. I have updated it accordingly. 

I noticed that the [python3-fastapi](https://packages.ubuntu.com/impish/python3-fastapi) is only available in Ubuntu 21 Impish and above. What's the convention if a user is using ubuntu 20.04 focal, and decides to use rosdep to download this package?

clalancette:
> I noticed that the python3-fastapi is only available in Ubuntu 21 Impish and above. What's the convention if a user is using ubuntu 20.04 focal, and decides to use rosdep to download this package?

Good question.  The more usual way (which will allow CI to pass here) is to set the key for the distributions that are still supported, but don't have the package to `null`, which the tools will ignore.  So in the case of `fastapi`, this would look something like:

```
python3-fastapi:
  arch: [python-fastapi]
  debian: 
    '*': [python3-fastapi]
    buster: null
  fedora: [python-fastapi]
  ubuntu:
    '*': [python3-fastapi]
    bionic: null
    focal: null
```

It's also possible to use a pip key for the distributions that don't currently support it.  That said, we still can't release packages onto the buildfarm that depend on that key; we only support system packages on the buildfarm.  If you'd still like to have the pip keys for local, source use, you can do something like:

```
python3-fastapi:
  arch: [python-fastapi]
  debian: 
    '*': [python3-fastapi]
    buster:
      pip:
        packages: [fastapi]
  fedora: [python-fastapi]
  ubuntu:
    '*': [python3-fastapi]
    bionic:
      pip:
        packages: [fastapi]
    focal:
      pip:
        packages: [fastapi]
```

Does that make sense?

hidmic:
@youliangtan friendly ping

youliangtan:
Thanks @clalancette for the suggestion. I think supporting other distributions will be handy for our application. I have implemented the second method for `python3-fastapi` and `python3-uvicorn`.

clalancette (approved):

nuclearsandwich (approved):
Thanks for working with us during this long iteration cycle!

PR #31735, opened 2022-01-11
Add webargs and Flask-JWT-Extended

ardabbour:

## Package names

- python3-webargs
- python3-flask-jwt-extended

## Package Upstream Sources

- webargs: https://github.com/marshmallow-code/webargs
- Flask-JWT-Extended: https://github.com/vimalloc/flask-jwt-extended

## Purpose of using these

- Webargs is a great package for simplifying HTTP parse objects and is compatiable with many backends (Flask, Django, etc.)
- Flask-JWT-Extended enables the use of JSON Web Tokens with Flask

#### webargs

- Arch: https://archlinux.org/packages/community/any/python-webargs/
- Debian: https://packages.debian.org/en/sid/python3-webargs
  - REQUIRED
- NixOS: https://hydra.nixos.org/job/nixos/release-21.11/nixpkgs.python39Packages.webargs.x86_64-linux
- PyPi: https://pypi.org/project/webargs/
- Ubuntu: https://packages.ubuntu.com/impish/python3-webargs
   - REQUIRED

#### Flask-JWT-Extended

- Debian: https://packages.debian.org/en/sid/python3-python-flask-jwt-extended
  - REQUIRED
- FreeBSD: https://cgit.freebsd.org/ports/tree/www/py-flask-jwt-extended
- NixOS: https://hydra.nixos.org/job/nixos/release-21.11/nixpkgs.python39Packages.flask-jwt-extended.x86_64-linux
- OpenSuse: https://software.opensuse.org/package/python-flask-jwt-extended
- PyPi: https://pypi.org/project/Flask-JWT-Extended/
- Ubuntu: https://packages.ubuntu.com/impish/python3-python-flask-jwt-extended
   - REQUIRED

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8536,13 +8536,13 @@ python3-weasyprint-pip:
 python3-webargs:
   arch: [python-webargs]
   debian:
+    '*': [python3-webargs]
     bullseye:
       pip:
         packages: [webargs]
     buster:
       pip:
         packages: [webargs]
-    '*': [python3-webargs]
   nixos: [python39Packages.webargs]
   ubuntu:
     '*': [python3-webargs]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8545,6 +8545,7 @@ python3-webargs:
   opensuse: [python-webargs]
   ubuntu:
     impish: [python3-webargs]
+    jammy: [python3-webargs]
     pip:
       packages: [webargs]
 python3-websocket:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8539,7 +8539,7 @@ python3-webargs:
   debian:
     pip:
       packages: [webargs]
-    sid: [python3-webargs]
+    '*': [python3-webargs]
   nixos: [python39Packages.webargs]
   ubuntu:
     impish: [python3-webargs]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6517,7 +6517,7 @@ python3-flask-cors:
     bionic: null
 python3-flask-jwt-extended:
   debian:
-    bullseye: [python3-python-flask-jwt-extended]
+    '*': [python3-python-flask-jwt-extended]
     pip:
       packages: [flask-jwt-extended]
   freebsd: [py-flask-jwt-extended]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8542,7 +8542,6 @@ python3-webargs:
       packages: [webargs]
     sid: [python3-webargs]
   nixos: [python39Packages.webargs]
-  opensuse: [python-webargs]
   ubuntu:
     impish: [python3-webargs]
     jammy: [python3-webargs]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1750,13 +1750,6 @@ python-flask-cors-pip:
   ubuntu:
     pip:
       packages: [flask-cors]
-python-flask-jwt-extended-pip: &migrate_eol_2025_04_30_python3_flask_jwt_extended_pip
-  debian:
-    pip:
-      packages: [flask-jwt-extended]
-  ubuntu:
-    pip:
-      packages: [flask-jwt-extended]
 python-flask-restful:
   debian: [python-flask-restful]
   fedora: [python-flask-restful]
@@ -5532,13 +5525,6 @@ python-watchdog:
     xenial: [python-watchdog]
     yakkety: [python-watchdog]
     zesty: [python-watchdog]
-python-webargs-pip: &migrate_eol_2025_04_30_python3_webargs_pip
-  debian:
-    pip:
-      packages: [webargs]
-  ubuntu:
-    pip:
-      packages: [webargs]
 python-webob:
   debian: [python-webob]
   fedora: [python-webob]
@@ -6529,7 +6515,21 @@ python3-flask-cors:
   ubuntu:
     '*': [python3-flask-cors]
     bionic: null
-python3-flask-jwt-extended-pip: *migrate_eol_2025_04_30_python3_flask_jwt_extended_pip
+python3-flask-jwt-extended:
+  debian:
+    bullseye: [python3-python-flask-jwt-extended]
+    pip:
+      packages: [flask-jwt-extended]
+    sid: [python3-python-flask-jwt-extended]
+  freebsd: [py-flask-jwt-extended]
+  nixos: [python39Packages.flask-jwt-extended]
+  opensuse: [python-flask-jwt-extended]
+  ubuntu:
+    focal: [python3-python-flask-jwt-extended]
+    hirsuite: [python3-python-flask-jwt-extended] 
+    impish: [python3-python-flask-jwt-extended]
+    pip:
+      packages: [flask-jwt-extended]
 python3-flask-restplus-pip:
   ubuntu:
     pip:
@@ -8534,7 +8534,18 @@ python3-weasyprint-pip:
   ubuntu:
     pip:
       packages: [weasyprint]
-python3-webargs-pip: *migrate_eol_2025_04_30_python3_webargs_pip
+python3-webargs:
+  arch: [python-webargs]
+  debian: 
+    pip:
+      packages: [webargs]
+    sid: [python3-webargs]
+  nixos: [python39Packages.webargs]
+  opensuse: [python-webargs]
+  ubuntu:
+    impish: [python3-webargs]
+    pip:
+      packages: [webargs]
 python3-websocket:
   debian: [python3-websocket]
   fedora: [python3-websocket-client]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1750,6 +1750,13 @@ python-flask-cors-pip:
   ubuntu:
     pip:
       packages: [flask-cors]
+python-flask-jwt-extended-pip: &migrate_eol_2025_04_30_python3_flask_jwt_extended_pip
```
tfoote, line 1753: We don't need to do the migration logic for a new key using the old names. That's only needed to avoid breaking existing keys.

tfoote (changes requested):
Thanks for the submission. Please update it to use the system packages instead of pip where available. And we don't need to add backwards compatibility references for new keys.

---

I noticed that webargs is packaged upstream already: https://packages.ubuntu.com/impish/python3-webargs

Please check the other distros as well. Debian definitely should have it too if Ubuntu has it.

We should add the system packages for impish and newer and you can backfill anything with pip earlier which is important to you. 

I fonund jwt-simple packaged but not jwt-extended https://packages.ubuntu.com/jammy/python3-flask-jwt-simple

rosdep/python.yaml
ardabbour: done!

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
...
+    sid: [python3-python-flask-jwt-extended]
+  freebsd: [py-flask-jwt-extended]
+  nixos: [python39Packages.flask-jwt-extended]
+  opensuse: [python-flask-jwt-extended]
+  ubuntu:
+    focal: [python3-python-flask-jwt-extended]
+    hirsuite: [python3-python-flask-jwt-extended]
+    impish: [python3-python-flask-jwt-extended]
+    jammy: [python3-python-flask-jwt-extended]
+    pip:
```
tfoote, line 6532: This pip key currently will be interpreted as a version of ubuntu and never match. This entry should manually call out all older supported platforms on which the pip key should by used to resolve instead of the wildcard. For this it's only `bionic`.

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
...
+      packages: [flask-jwt-extended]
+    sid: [python3-python-flask-jwt-extended]
+  freebsd: [py-flask-jwt-extended]
+  nixos: [python39Packages.flask-jwt-extended]
+  opensuse: [python-flask-jwt-extended]
+  ubuntu:
+    focal: [python3-python-flask-jwt-extended]
+    hirsuite: [python3-python-flask-jwt-extended]
+    impish: [python3-python-flask-jwt-extended]
+    jammy: [python3-python-flask-jwt-extended]
```
tfoote, line 6531:
```suggestion
    '*': [python3-python-flask-jwt-extended]
```

Please change the current key for focal an newer to a wildcard, removing the other identical ones.

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
   ubuntu:
     '*': [python3-flask-cors]
     bionic: null
+python3-flask-jwt-extended:
+  debian:
+    bullseye: [python3-python-flask-jwt-extended]
+    pip:
+      packages: [flask-jwt-extended]
+    sid: [python3-python-flask-jwt-extended]
```
tfoote, line 6523: We don't add keys for sid generally due to it being a rolling/unreleased target.
```suggestion
```

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
   ubuntu:
     '*': [python3-flask-cors]
     bionic: null
+python3-flask-jwt-extended:
+  debian:
+    bullseye: [python3-python-flask-jwt-extended]
+    pip:
```
tfoote, line 6521: This should explicitly call out `buster`. Right now it's resolved as a pip version.

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
   ubuntu:
     '*': [python3-flask-cors]
     bionic: null
+python3-flask-jwt-extended:
+  debian:
+    bullseye: [python3-python-flask-jwt-extended]
```
tfoote, line 6520:
```suggestion
    '*': [python3-python-flask-jwt-extended]
```

Change this to a wildcard for future compatibility.

rosdep/python.yaml
```diff
@@ -8519,6 +8535,18 @@ python3-weasyprint-pip:
   ubuntu:
     pip:
       packages: [weasyprint]
+python3-webargs:
+  arch: [python-webargs]
+  debian:
+    pip:
+      packages: [webargs]
+    sid: [python3-webargs]
```
tfoote, line 8543:
```suggestion
    '*': [python3-webargs]
```

Switching this to * for bookworm and newer

rosdep/python.yaml
```diff
@@ -8519,6 +8535,18 @@ python3-weasyprint-pip:
   ubuntu:
     pip:
       packages: [weasyprint]
+python3-webargs:
+  arch: [python-webargs]
+  debian:
+    pip:
```
tfoote, line 8541: Please change this to explicitly target bullseye and buster.

rosdep/python.yaml
```diff
@@ -8519,6 +8535,18 @@ python3-weasyprint-pip:
...
+      packages: [webargs]
+    sid: [python3-webargs]
+  nixos: [python39Packages.webargs]
+  ubuntu:
+    impish: [python3-webargs]
+    jammy: [python3-webargs]
```
tfoote, line 8547:
```suggestion
    '*': [python3-webargs]
```

Switching to wildcard for impish and newer.

rosdep/python.yaml
```diff
@@ -8519,6 +8535,18 @@ python3-weasyprint-pip:
...
+    pip:
```
tfoote, line 8548: This needs to target bionic and focal explicitly.

rosdep/python.yaml
```diff
@@ -8519,6 +8535,18 @@ python3-weasyprint-pip:
...
       packages: [weasyprint]
+python3-webargs:
+  arch: [python-webargs]
+  debian:
+    pip:
+      packages: [webargs]
+    sid: [python3-webargs]
+  nixos: [python39Packages.webargs]
+  ubuntu:
+    impish: [python3-webargs]
```
tfoote, line 8546:
```suggestion
```

Covered by wildcard.

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
...
+  debian:
+    bullseye: [python3-python-flask-jwt-extended]
+    pip:
+      packages: [flask-jwt-extended]
+    sid: [python3-python-flask-jwt-extended]
+  freebsd: [py-flask-jwt-extended]
+  nixos: [python39Packages.flask-jwt-extended]
+  opensuse: [python-flask-jwt-extended]
+  ubuntu:
+    focal: [python3-python-flask-jwt-extended]
```
tfoote, line 6528:
```suggestion
```
covered by wildcard

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
...
+    hirsuite: [python3-python-flask-jwt-extended]
```
tfoote, line 6529:
```suggestion
```
Covered by wildcard

rosdep/python.yaml
```diff
@@ -6515,6 +6515,22 @@ python3-flask-cors:
...
+    impish: [python3-python-flask-jwt-extended]
```
tfoote, line 6530:
```suggestion
```
covered by wildcard

tfoote (changes requested):
Thanks for the quick update. The syntax we're looking for is quite specific. I've gone through with all the single line suggestions I can make. The older targets using pip need to be called out individually to override the wildcard syntax. 

We explicitly call out the older versions verbosely because it's possible to list all older versions but not possible to list all future versions. And the best bet for any package is that it will be named the same in the next release which usually means it won't need to be touched again when the next release comes out.

rosdep/python.yaml
ardabbour: resolved in bf1169b2bc4b36c3c5fb44d670d279a9a62bee5d

rosdep/python.yaml
ardabbour: resolved in bf1169b2bc4b36c3c5fb44d670d279a9a62bee5d

rosdep/python.yaml
ardabbour: resolved in bf1169b2bc4b36c3c5fb44d670d279a9a62bee5d

rosdep/python.yaml
ardabbour: resolved in bf1169b2bc4b36c3c5fb44d670d279a9a62bee5d

tfoote (approved):
Thanks for all the iterations. This looks good to me now.

jacobperron (approved):

PR #31592, opened 2021-12-24
Automatically label PRs based on files they change

sloretz:
Signed-off-by: Shane Loretz <sloretz@osrfoundation.org>

This uses https://github.com/actions/labeler to automatically label PRs based on which files they change. See proof that this PR works by looking at the PRs on my fork: https://github.com/sloretz/rosdistro/pulls

* sloretz/rosdistro#2
* sloretz/rosdistro#3
* sloretz/rosdistro#4
* sloretz/rosdistro#5
* sloretz/rosdistro#6
* sloretz/rosdistro#7
* sloretz/rosdistro#8

PR #31930, opened 2022-01-26
Label and close PRs only modifying EOL distro files

audrow:
This PR extends https://github.com/ros/rosdistro/pull/31592 to add an `end-of-life` label to PRs targeting EOL ROS distro folders and closes PRs that only target EOL distributions.

For proof that this works, see https://github.com/ros/rosdistro/pull/31930#issuecomment-1068378652.


audrow:
If we like, we could also use Mergify to automatically close `end-of-life` labels.

Here's some reference info:
- https://docs.mergify.com/actions/close/
- https://docs.mergify.com/conditions/#attributes

clalancette:
> If we like, we could also use Mergify to automatically close `end-of-life` labels.

I like this idea, though it would have to be done in moderation.  That is, if somebody opens a PR that *just* touches EOL distros, then I think an automatic close with a note is fine.

However, sometimes people open up PRs that touch a bunch of distros at once (like `kinetic`, `melodic`, and `noetic`).  In those cases I think we wouldn't want to close them, but instead just automatically add a comment about not touching EOL distros.

If we could get that set of logic, I'd be all for it.  If that's too much work, them I'm fine with adding just the label for now.

audrow:
> However, sometimes people open up PRs that touch a bunch of distros at once (like `kinetic`, `melodic`, and `noetic`). In those cases I think we wouldn't want to close them, but instead just automatically add a comment about not touching EOL distros.
> 
> If we could get that set of logic, I'd be all for it. If that's too much work, them I'm fine with adding just the label for now.

Mergify has and/or logic, so I believe it could be done pretty easily. I can add this to this PR or make another.

audrow:
I added Mergify to this PR and updated the description.

clalancette (commented):
I have two general concerns here:

1. The Mergify configuration depends on labels to be set in order to close things.  But what is the order in which labeling happens vs Mergify checking its rules?  That is, if the Mergify action runs before (or concurrent with) the labeler, then it may have false negatives.
2. While the `and` conditions in the Mergify configuration works, it seems like it will auto-close PRs for e.g. `humble` until we add the negative condition to the configuration.  It's not the end of the world, since it will be pretty noticeable when it first happens, but it isn't very nice.  I'm wondering if we can do something else here (though I have no good suggestions at the moment).

audrow:
> 1. The Mergify configuration depends on labels to be set in order to close things.  But what is the order in which labeling happens vs Mergify checking its rules?  That is, if the Mergify action runs before (or concurrent with) the labeler, then it may have false negatives.

In my testing it seemed that Mergify waited for all other tests to run before performing any actions, although I haven't been able to find this in the documentation. Perhaps I can test more to see if this is really true.

> 2. While the `and` conditions in the Mergify configuration works, it seems like it will auto-close PRs for e.g. `humble` until we add the negative condition to the configuration.  It's not the end of the world, since it will be pretty noticeable when it first happens, but it isn't very nice.  I'm wondering if we can do something else here (though I have no good suggestions at the moment).

It shouldn't autoclose `humble` unless it also has the `end-of-life` tag, which it wouldn't have unless the PR also modifies EOL distributions. Once we go ahead and make the first humble releases, we can update the YAML files, or I could do it in this PR.

ivanpauno:
> In my testing it seemed that Mergify waited for all other tests to run before performing any actions, although I haven't been able to find this in the documentation. Perhaps I can test more to see if this is really true.

I was just looking, I haven't found any documentation about that either.
Mergify allows you to check the status of other checks (see [here](https://docs.mergify.com/conditions/#about-status-checks)), so it must be running after all the other ones (?)

.github/mergify.yml
```diff
@@ -0,0 +1,17 @@
...
+        - label!=noetic
+        # Active ROS 2 distros
+        - label!=foxy
+        - label!=galactic
+        - label!=rolling
+        # Other labels
+        - label!=rosdep
+    actions:
+      close:
+        message: This pull request has been automatically closed by Mergify since it only changes files for a ROS distribution that is no longer supported (End Of Life) and the distribution will not have any future syncs to include this change.
```
ivanpauno, line 17: I'm not sure if it's possible, but pointing out in the message what EOL distros are causing the issue would be awesome.

We could also always leave a message when there's a change to a EOL folder, but only close the PR if only EOL distros were modified (I mean, the check also leaves a message in https://github.com/audrow/rosdistro/pull/5, but doesn't close the PR).
I think contributors are much more likely to read a message that to look at the labels.

ivanpauno (commented):
Looks awesome to me!
I think this is good to go with @clalancette concerns resolved.

I have a minor suggestion that can be ignored.

nuclearsandwich (approved):
1. this is rad.
2. I would enjoy this much more if it was able to parse and re-use the distribution statuses from index-v4.yaml instead of requiring us to maintain a separate list of distros. We already maintain several copies of this list and each one hurts us a little. If we can't consume index-v4.yaml directly could we add a script to generate these from index-v4 (and then add another action which runs that script whenever index-v4.yaml is modified.)

audrow:
> 2\. I would enjoy this much more if it was able to parse and re-use the distribution statuses from index-v4.yaml instead of requiring us to maintain a separate list of distros. We already maintain several copies of this list and each one hurts us a little. If we can't consume index-v4.yaml directly could we add a script to generate these from index-v4 (and then add another action which runs that script whenever index-v4.yaml is modified.)

I'll look into it.

nuclearsandwich:
@audrow should we merge this while you look into improving the duplication of distribution statuses?

tfoote (approved):
This generally looks good. There are occasions when we do want to change something in the past, such as hosting of historical content gets changed. Can we reopen successfully or will it auto close again?

We should also consider using this for sync holds, I assume we can add comments/labels automatically in the same way instead of closing.

I'd suggest we go ahead with this as it is for now and see if we can improve it further down the line.

jacobperron (approved):
lgtm

audrow:
Alright, just confirming that this works as intended.

Should have a comment and be closed:
* Changes to one EOL distro: https://github.com/audrow/rosdistro/pull/13
* Changes to two EOL distros: https://github.com/audrow/rosdistro/pull/14

Should be open with a comment (note I fixed a typo in the comment after creating this example):
* Changes to two EOL distros and one active distro: https://github.com/audrow/rosdistro/pull/15

Should be unaffected:
* Changes to an active distro: https://github.com/audrow/rosdistro/pull/16

.github/mergify.yml
audrow: > We could also always leave a message when there's a change to a EOL folder, but only close the PR if only EOL distros were modified (I mean, the check also leaves a message in [audrow#5](https://github.com/audrow/rosdistro/pull/5), but doesn't close the PR).

I've set this up. Thanks for the idea.

audrow:
> This generally looks good. There are occasions when we do want to change something in the past, such as hosting of historical content gets changed. Can we reopen successfully or will it auto close again?

Just tested, if you reopen a PR it stays open, even if additional commits are made.

> We should also consider using this for sync holds, I assume we can add comments/labels automatically in the same way instead of closing.

I like this idea, and that's correct.

> I'd suggest we go ahead with this as it is for now and see if we can improve it further down the line.

🚀

clalancette (approved):
Looks reasonable to me.  I think this is good enough to try out; if we find we don't like it, or it doesn't act like we expect, we can always disable it.

PR #31965, opened 2022-01-29
rtabmap: 0.20.18-3 in 'melodic/distribution.yaml' [bloom]

matlabbe:
Increasing version of package(s) in repository `rtabmap` to `0.20.18-3`:

- upstream repository: https://github.com/introlab/rtabmap.git
- release repository: https://github.com/introlab/rtabmap-release.git
- bloom version: `0.10.7`
- previous version for package: `0.20.16-1`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -12358,7 +12358,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/introlab/rtabmap-release.git
-      version: 0.20.16-1
+      version: 0.20.18-3
     source:
       type: git
       url: https://github.com/introlab/rtabmap.git


clalancette:
@matlabbe FYI, this package is failing to build on arm32: https://build.ros.org/view/Mbin_ubhf_uBhf/job/Mbin_ubhf_uBhf__rtabmap__ubuntu_bionic_armhf__binary/ .

In turn, that is happening because one of its dependencies, libpointmatcher, can't build on arm32: https://build.ros.org/view/Mbin_ubhf_uBhf/job/Mbin_ubhf_uBhf__libpointmatcher__ubuntu_bionic_armhf__binary/54/console (search for "Libpointmatcher is only supported in 64 bits").

I have to assume that this is a relatively new development, since this built on arm32 in the past.  The question is: do you care about this regression?  If so, we'll probably have to revert to an earlier version of libpointmatcher.  If you don't care, then we can just ignore this failure.

clalancette:
> I don't know if there is a mechanism to do so, but the ideal workaround for me would be to specify in package.xml that libpointmatcher is optional like this:

We don't currently have any way to specify optional dependencies in the package.xml.

> Is there an environment variable set in the buildfarm telling which architecture we are building on?

My understanding is that this is actually resolved at bloom time, not on the buildfarm.  So we would have to have some way to do it at bloom time.  But I'm not an expert here; @nuclearsandwich @cottsay @j-rivero could one of you comment if this is possible?

There may also be the possibility to patch out libpointmatcher as a dependency just in the release repository for arm32, but I don't know how this would be done.

nuclearsandwich:
In general our tools don't cover architecture differences per-platform and there's some question about whether rosdep or dependency conditions are a more appropriate venue for how platform (by which I mean the combination of operating system, operating system version, and architecture) differences should be resolved in dependencies.

> Is there an environment variable set in the buildfarm telling which architecture we are building on?

clalancette is correct that the bloom-release command resolves all rosdep dependencies, and by extension package conditions when generating the release data. Furthermore the same platform metadata (RPM spec files for RHEL, debian/control and debian/rules for Debian and Ubuntu are used regardless of architecture so any architecture dependent behavior needs to be representable in those formats. 

Unlike colcon workspaces which can evaluate dependency conditions based on arbitrary environment variables, bloom uses a [fixed set based on the ROS distribution information](https://github.com/ros-infrastructure/bloom/blob/13b921c2222447966474062bfc880842030dc466/bloom/generators/common.py#L124-L161).

I went down a bit of a rabbit hole trying to find things for the deb side but it looks like build dependencies can be qualified with an architecture restriction according to https://www.debian.org/doc/debian-policy/ch-relationships.html but that would have to be done in a patch after blooming and is not currently expressible in a package.xml.

nuclearsandwich:
@matlabbe those are the correct lines but I'm not sure what branch that blob is on, wherever possible we try to apply patches to the templates rather than the generated files but without a ton of empy logic there's not an easy way to do that.

Rather than restricting the dependency to amd64 I would suggest a negative qualification removing it from armhf. For example:
```diff
- ros-melodic-libpointmatcher [amd64]
+ ros-melodic-libpointmatcher [!armhf]
```

I would suggest also adding a comment on the source-level package.xml indicating that the dependency isn't available on and thus isn't required for 32-bit platforms so that people building from source on 32-bit platforms know that they can skip that rosdep key and still build successfully.

If you'd like me to review changes or initiate the PRs implementing my suggestions I would be happy to do so.

matlabbe:
Yes it is better `[!armhf]` thx. I did the change for melodic and noetic:
https://github.com/introlab/rtabmap-release/commit/59ce7410d1fcc00601fd63dd30ce059c13806ff0
https://github.com/introlab/rtabmap-release/commit/0a4dcbc2aed2c5f6a9ca2e7b27bec4b47f2449dd

Will the buildfarm automatically detect the change and start a new build, or we should trigger a new build (I think I don't have permissions do that on jenkins).

clalancette:
> Will the buildfarm automatically detect the change and start a new build, or we should trigger a new build (I think I don't have permissions do that on jenkins)?

You'll have to do a new bloom build, I believe.  It should be enough to just re-run `bloom-release --track melodic --rosdistro melodic rtabmap`, and that should just bump the debinc number.

matlabbe:
Well, it seems not working as the dependencies are checked before the debian build, so not sure that can be fixed. Well, we will have to let armhf fails then... 
https://build.ros.org/job/Mbin_ubhf_uBhf__rtabmap__ubuntu_bionic_armhf__binary/
https://build.ros.org/job/Nbin_ufhf_uFhf__rtabmap__ubuntu_focal_armhf__binary/

nuclearsandwich:
Ack. I failed to think of that. The build farm naturally has no idea that certain dependencies have been patched to be architecture specific. And blocklisting libpointmatcher builds on armhf would also block its downstream so there really isn't a good solution here within the current capabilities of the build farm.

@clalancette is there a reason that libpointmatcher builds aren't already disabled on armhf or should I open a PR to do it?

clalancette:
> @clalancette is there a reason that libpointmatcher builds aren't already disabled on armhf or should I open a PR to do it?

No, feel free to open a PR to do that.  I'm happy to review.

nuclearsandwich:
@cottsay brought up the very good point that he added features to ros_buildfarm to support the very similar case of some non-critical test/static analysis dependencies not being available for packages on RHEL and that if I use `package_ignore_list` rather than the blacklist it won't disable the downstream packages. I don't think this is a slam dunk or generic solution to problems like this because the setting is buildfarm config-specific rather than package specific and it affects a package's entire downstream rather than just the package that "wants" to be built with or without that dependency.

nuclearsandwich:
> No, feel free to open a PR to do that. I'm happy to review.

https://github.com/ros-infrastructure/ros_buildfarm_config/pull/214

PR #31995, opened 2022-01-31
extend rosdep rule for open3d pip package in python.yaml

bbferka:
This PR extends the rosdep definitions for open3d. As per [documentation](http://www.open3d.org/docs/release/getting_started.html#pip-pypi) of open3d ``pip install open3d`` is the preferred way for installing the package. MR does not remove the old open3d-python definition such that  unintended consequences can be avoided.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7116,13 +7116,13 @@ python3-nuscenes-devkit-pip:
 python3-open3d-pip:
   debian:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
   fedora:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
   ubuntu:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
 python3-opencv:
   debian: [python3-opencv]
   fedora: [python3-opencv]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3055,13 +3055,13 @@ python-omniorb:
 python-open3d-pip:
   debian:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
   fedora:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
   ubuntu:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
 python-opencv:
   arch: [opencv, python2-numpy]
   debian: [python-opencv]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3055,13 +3055,13 @@ python-omniorb:
 python-open3d-pip:
   debian:
     pip:
-      packages: [open3d-python, open3d]
+      packages: [open3d-python]
   fedora:
     pip:
-      packages: [open3d-python, open3d]
+      packages: [open3d-python]
   ubuntu:
     pip:
-      packages: [open3d-python, open3d]
+      packages: [open3d-python]
 python-opencv:
   arch: [opencv, python2-numpy]
   debian: [python-opencv]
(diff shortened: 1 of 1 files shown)


ivanpauno:
>  MR does not remove the old open3d-python definition such that unintended consequences can be avoided.

What are the possible unintended consequences?

rosdep/python.yaml
```diff
@@ -3055,13 +3055,13 @@ python-omniorb:
...
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
   fedora:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
   ubuntu:
     pip:
-      packages: [open3d-python]
+      packages: [open3d-python, open3d]
```
ivanpauno, line 3064: It looks like `open3d` is python3 only, while `open3d-python` was python2 compatible.
i.e.: `python-open3d-pip` shouldn't be modified, only `python3-open3d-pip`

ivanpauno:
> MR does not remove the old open3d-python definition such that unintended consequences can be avoided.

Based on https://github.com/isl-org/Open3D/pull/1162:

> We have reclaimed the open3d name on PyPI. We will deprecate the open3d-python PyPI name for pip install.

It seems that `open3d` is the new name of the package.
If they were able to make the imports backwards compatible, there's no need to keep both entries in the list.

bbferka:
@ivanpauno  I was thinking that maybe because the new pip entry will install a way newer version now this might break packages.  But maybe that is not really our concern here;  I made the requested changes

ivanpauno:
> @ivanpauno I was thinking that maybe because the new pip entry will install a way newer version now this might break packages. But maybe that is not really our concern here; I made the requested changes

That's fine, pip always install the latest version anyway, that can include breaking changes.
My only concern is the one I commented above:

> If they were able to make the imports backwards compatible

are the imports of the new package backwards compatible with the old ones?

ivanpauno (approved):
LGTM if the new package imports are backwards compatible with the old ones (yet to confirm).

hidmic:
@bbferka friendly ping

clalancette:
Hm, tough one.  So looking at both https://pypi.org/project/open3d/ and https://pypi.org/project/open3d-python/ , it is clear they are the same project (they have the same homepage and the same pointer to the source code).

Looking at https://github.com/isl-org/Open3D/releases , it is clear that between 0.7.0 and 0.8.0 they renamed the pip library from `open3d-python` to `open3d` (it is listed right there in the release notes).  So far so good.

However the versions that they have available are all over the place.

| Ubuntu version | Python version | pip package name | version installed |
| -------------------- | -------------------- | ------------------------ | ---------------------- |
| Bionic (18.04)   |  3.6.9                |  open3d-python    | 0.7.0.0                 |
| Bionic (18.04)   | 3.6.9                 |  open3d                | 0.11.2                  |
| Focal (22.04)    | 3.8.10               |  open3d-python    | 0.3.0.0                 |
| Focal (22.04)    | 3.8.10               |  open3d                | 0.13.0                  |
| Jammy (22.04) | 3.9.10               |  open3d-python    | 0.3.0.0                 |
| Jammy (22.04) | 3.9.10               |  open3d                | 0.14.1                  |

For extra hilarity, the latest released version on pypi.org for open3d is 0.15.1, which can't seem to be installed on any platform: https://pypi.org/project/open3d/

So with all of that said, I honestly don't know what to do here.  My inclination is to remove it from the rosdep database completely; users are almost always going to be disappointed in what they get.

@bbferka How are you actually using this library, given that its installation candidates are all over the place?

bbferka:
@clalancette we were so far installing it manually with ``pip install open3d`` and only tested on Focal. Contrary to your table for me pip install v0.14.1 so the same as Jammy.  The first time we hit this issue is when we tired automatic dependency installation through rosdep and ended up with the old library.

nuclearsandwich:
It looks like this rule was introduced in #24346. The most conservative thing to do is create a new rule and transition packages to using that and that would be my recommendation except for the fact that the current rule is camping on the preferred name for the rosdep rule.

We can probably go ahead and make this change as-is. Pip keys and their versions are always potentially volatile and this change aligns both with upstream and the rosdep style guide.

nuclearsandwich (approved):

PR #32010, opened 2022-02-02
Add python-modbus-pip to python.yaml 

tgaspar:
Please add the `python-modbus-pip` dependency to the rosdep database.

## Package name:

python-modbus-pip

## Package Upstream Source:

Link to source repository: https://github.com/riptideio/pymodbus

## Purpose of using this:

There already is an entry for `python-modbus` in the YAML. However, the Ubuntu and Debian repositories have outdated versions, [1.3.2](https://packages.ubuntu.com/search?suite=bionic&searchon=names&keywords=pymodbus) and [1.2.0](https://packages.debian.org/stretch/python-pymodbus) respectively.

When installing from PyPI we get a more recent version - [2.5.3](https://pypi.org/project/pymodbus/). 

I suggest to keep the outdated `python-modbus` inside the YAML to not break compatibility.

Distro packaging links:

  - https://pypi.org/project/pymodbus/
  - https://pypi.org/project/pymodbus/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3701,13 +3701,6 @@ python-pymodbus-pip:
   ubuntu:
     pip:
       packages: [pymodbus]
-python3-pymodbus-pip:
-  debian:
-    pip:
-      packages: [pymodbus]
-  ubuntu:
-    pip:
-      packages: [pymodbus]
 python-pymongo:
   arch: [python2-pymongo]
   debian: [python-pymongo]
@@ -7576,6 +7569,13 @@ python3-pymodbus:
   ubuntu:
     '*': [python3-pymodbus]
     xenial: null
+python3-pymodbus-pip:
+  debian:
+    pip:
+      packages: [pymodbus]
+  ubuntu:
+    pip:
+      packages: [pymodbus]
 python3-pymongo:
   arch: [python-pymongo]
   debian: [python3-pymongo]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3701,6 +3701,13 @@ python-pymodbus-pip:
   ubuntu:
     pip:
       packages: [pymodbus]
+python3-pymodbus-pip:
+  debian:
+    pip:
+      packages: [pymodbus]
+  ubuntu:
+    pip:
+      packages: [pymodbus]
 python-pymongo:
   arch: [python2-pymongo]
   debian: [python-pymongo]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3692,6 +3692,9 @@ python-pymavlink:
     pip:
       packages: [pymavlink]
 python-pymodbus:
+  debian: [python-pymodbus]
+  ubuntu: [python-pymodbus]
+python-pymodbus-pip:
   debian:
     pip:
       packages: [pymodbus]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3694,6 +3694,13 @@ python-pymavlink:
 python-pymodbus:
   debian: [python-pymodbus]
   ubuntu: [python-pymodbus]
+python-pymodbus-pip:
```
ivanpauno, line 3697: do you want to add the python3 key?

In that case, please use `python3-pymodbus-pip`

tgaspar: Would it be alright to **also** add the python 3 key? In order to allow both older ROS distributions to use this too.

ivanpauno: > Would it be alright to also add the python 3 key? In order to allow both older ROS distributions to use this too.

yes!

ivanpauno (approved):

rosdep/python.yaml
sloretz, line 3697: One last change, since these are identical please [make the `python3-` key an alias to the `python-` key](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1).

```yaml
python-pymodbus-pip: &migrate_eol_2027_04_30_python3_pymodbus_pip
  debian:
    pip:
      packages: [pymodbus]
  ubuntu:
    pip:
      packages: [pymodbus]
...
python3-pymodbus-pip: *migrate_eol_2027_04_30_python3_pymodbus_pip

```

tgaspar: Ok, will do! Quick question though: how do I determine the EOL? Or better, how did you determine it?

sloretz: > how do I determine the EOL? Or better, how did you determine it?

I know we're transitioning to Ubuntu Jammy for ROS Rolling at the moment, so I assumed both that it counts as a supported platform and that it has the furthest away EOL date of April 2027.

tgaspar:
I could use a little help here. I am not sure why I am getting the following nosetest error:

```Package 'python-pymodbus' could not be found for ubuntu focal on amd64```

and 

```Package 'python-pymodbus' could not be found for ubuntu jammy on amd64```
Mostly because these are the lines of the YAML that remained untouched by my PR.

**Comment by ivanpauno** [comment]
@tgaspar the PR checker sometimes checks more lines than needed (see https://github.com/ros/rosdistro/issues/31439).
We will need to fix the `python-pymodbus` entry in another PR and rebase this one (or you can also fix that here).

**Comment by ivanpauno** [comment]
To be more clear, this probably fixes the issue:

```
python-pymodbus:
  debian: [python-pymodbus]
  ubuntu:
    '*': [python-pymodbus]
    jammy: null
```

**Comment by tgaspar** [comment]
@ivanpauno thanks for the suggestion. Given the errors I received before I added both `jammy` and `focal`. Hope that's fine ...

File: rosdep/python.yaml [hunk 2]
```diff
@@ -3693,7 +3693,17 @@ python-pymavlink:
       packages: [pymavlink]
 python-pymodbus:
   debian: [python-pymodbus]
-  ubuntu: [python-pymodbus]
+  ubuntu:
+    '*': [python-pymodbus]
+    focal: null
+    jammy: null
```
ivanpauno, line 3699: Actually, future ubuntu releases will also not provide python 2 packages, so better use:

```suggestion
    bionic: [python-pymodbus]
```

melodic is currently the only ROS distribution supporting python2, and that uses Ubuntu Bionic.

tgaspar: Thanks for the heads-up.

ivanpauno (approved):

sloretz (approved):

PR #31954, opened 2022-01-29
rcss3d_agent: Temporary Removal from galactic track

ijnek:
Temporary removal of `rcss3d_agent` from the galactic track, before moving it to a separate repo. Once this is merged, a galactic version of https://github.com/ros/rosdistro/pull/31945 will be raised to re-add rcss3d_agent back.

Bloom release fails due to a name conflict with the rcss3d_agent that already exists.

Signed-off-by: ijnek <kenjibrameld@gmail.com>

PR #32057, opened 2022-02-07
nao_button_sim: Temporary Removal from galactic track

ijnek:
Temporary removal of nao_button_sim from the galactic track, before moving it to a separate repo. Once this is merged, a galactic version of https://github.com/ros/rosdistro/pull/32058 will be raised to re-add nao_button_sim back.

Bloom release fails due to a name conflict with the nao_button_sim that already exists.

Signed-off-by: ijnek [kenjibrameld@gmail.com](mailto:kenjibrameld@gmail.com)

### DIFF
--- a/galactic/distribution.yaml
+++ b/galactic/distribution.yaml
@@ -1993,23 +1993,6 @@ repositories:
       url: https://github.com/ijnek/nao_lola.git
       version: main
     status: developed
-  naosoccer_sim:
-    doc:
-      type: git
-      url: https://github.com/ijnek/naosoccer_sim.git
-      version: main
-    release:
-      packages:
-      - nao_button_sim
-      tags:
-        release: release/galactic/{package}/{version}
-      url: https://github.com/ijnek/naosoccer_sim-release.git
-      version: 0.0.2-1
-    source:
-      type: git
-      url: https://github.com/ijnek/naosoccer_sim.git
-      version: main
-    status: developed
   navigation2:
     doc:
       type: git


ivanpauno:
@ijnek removing a package from the galactic list will make all previous versions unavailable until it's added again.
i.e.: Users using it won't be able to download it anymore.
Is that what you want?

> Temporary removal of nao_button_sim from the galactic track, before moving it to a separate repo. Once this is merged, a galactic version of https://github.com/ros/rosdistro/pull/32058 will be raised to re-add nao_button_sim back.

I think you should be able to update that without removing the existing entry first.
Are you trying to modify the source or release repository?
I think that in both cases, running

```bash
bloom-release --edit ...
```

should work.

hidmic:
@ijnek friendly ping

ijnek:
> removing a package from the galactic list will make all previous versions unavailable until it's added again.
> i.e.: Users using it won't be able to download it anymore.
> Is that what you want?

@ivanpauno No, this removal then readd process is something that I've done before and has not been questioned (https://github.com/ros/rosdistro/pull/31954).

For why I do this, if I run
`bloom-release --rosdistro galactic --track galactic nao_button_sim --edit`, bloom initially says that the repository `nao_button_sim` was not found in galactic/disribution.yaml.

```sh
Specified repository 'nao_button_sim' is not in the distribution file located at 'https://raw.githubusercontent.com/ros/rosdistro/f3d0a93fbc11a70833c130c1ea97cbcf10beae25/galactic/distribution.yaml'
```

Bloom is correct, the existing nao_button_sim package lives inside the naosoccer_sim repository.

The bloom release procedure continues and is all fine until it tries to automatically open a PR, where it says

```sh
Failed to open pull request: AssertionError - Duplicate package name 'nao_button_sim' exists in repository 'naosoccer_sim' as well as in repository 'nao_button_sim'
```

because there are two nao_button_sim packages, the old on under naosoccer_sim, and the new nao_button_sim repo, **and exits**.

I couldn't find an easier way to get around this, then to temporarily remove the package, and then quickly readd it.

ijnek:
I see it's been discussed on ROS Discourse before, with no solution:
https://discourse.ros.org/t/repository-consolidation-moving-released-packages-to-another-repo/327/4

ivanpauno:
@ijnek I see there's no easy solution when the name of the repo changed.

I'm not sure if doing that atomically in one PR has some advantages though, maybe that avoids the issue of making all previous versions unavailable (?). It's true that bloom-release doesn't automate that, and has to be done semi-manually.

Maybe @nuclearsandwich or @tfoote can clarify.

ivanpauno (approved):

ivanpauno:
@ijnek we can also merge as-is if you don't care about the potential removal of previous packages.
Or we can update the PR so it adds the new package atomically here as well, which might potentially avoid that issue (if that doesn't avoid the issue, I don't think there's a way to do it).

ijnek:
@ivanpauno My preference would be to merge this PR as-is, and if necessary I could open an issue on https://github.com/ros-infrastructure/bloom

nuclearsandwich:
This is an unfortunately common workflow. I know that in the past it has been possible to remove a package's listing from another repository in a manual follow-up commit in the bloom PR but perhaps that predates bloom's warning about duplicate package names.

nuclearsandwich (approved):

ijnek:
@nuclearsandwich or @ivanpauno Can we merge this since Galactic since is done?

ijnek:
Friendly ping @ivanpauno

sloretz (approved):

ijnek:
Thanks @sloretz

PR #32071, opened 2022-02-07
Update autoware_auto_msgs release repository.

nuclearsandwich:
The release repository was forked into ros2-gbp for the Galactic migration in April 2021.  Since then there have been releases into the upstream release repository.

These releases were merged into the ros2-gbp branch on 2022-02-07.
Since the upcoming Rolling platform migration will again branch the release repositories into ros2-gbp and the ros2-gbp repository is now up-to-date with the external repository I recommend that we update the release repository pre-emptively to reduce churn in the bloom configuration branches.

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -321,7 +321,7 @@ repositories:
     release:
       tags:
         release: release/rolling/{package}/{version}
-      url: https://gitlab.com/autowarefoundation/autoware.auto/autoware_auto_msgs-release.git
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Looks good to me, though we may want to consider pinging the original maintainers as an FYI.

nuclearsandwich:
> though we may want to consider pinging the original maintainers as an FYI.

Yeah I've pinged @JWhitleyWork in several of these and just figured that they'd be tired of them by now. But for completeness here's one more.

PR #32036, opened 2022-02-03
Prepare Rolling for platform transition.

nuclearsandwich:
As part of the transition process using the [rosdistro migration tools](https://github.com/ros/rosdistro/tree/master/migration-tools) we need an empty target distribution with the intended target platforms.

The first commit of this PR empties the repositories list which will get re-populated when the migration tool runs.
The second commit removes ubuntu focal, adds ubuntu jammy, and adds debian bullseye leaving RHEL8 in place.
It is easier to review each commit separately.

I'd like to get review of this pull request in two phases.

I'd like on review phase to approve the platform list and source ref as I prepare to run the migration tool. This PR won't be merged as-is once approved. Instead it will be used as the foundation for a migration using 7eafe041e as the source ref.

Once the migration is complete, I'll update this PR and request a second phase of reviews for merging the completed migration.

---

Migration command:

```
python3 migration-tools/migrate-rosdistro.py --source rolling --dest rolling \
  --source-ref https://github.com/ros/rosdistro/commit/7eafe041ed8058a17f0fcd0a0c70290bf310ea5a \
  --release-org ros2-gbp \
```

Of 299 repositories, 22 failed to bloom successfully.
This is most frequently caused by missing rosdep dependencies.

The following packages failed to bloom:

* aws-robomaker-small-warehouse-world https://github.com/aws-robotics/aws-robomaker-small-warehouse-world/issues/21
* ~~control_box_rst~~ https://github.com/ros/rosdistro/pull/32285
* ~~fmi_adapter~~ https://github.com/ros/rosdistro/pull/32287
* ~~fmilibrary_vendor~~ https://github.com/ros/rosdistro/pull/32288
* ~~gazebo_ros2_control~~ https://github.com/ros/rosdistro/pull/32286
* gazebo_ros_pkgs https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1359
* ign_rviz https://github.com/ignitionrobotics/ign-rviz/issues/74
* ~~ignition_cmake2_vendor~~ https://github.com/ros/rosdistro/pull/32198
* ~~ignition_math6_vendor~~ https://github.com/ros/rosdistro/pull/32199
* ~~locator_ros_bridge~~ https://github.com/ros/rosdistro/pull/32191
* moveit https://github.com/ros-planning/moveit2/issues/1092
* ~~ntpd_driver~~ https://github.com/ros/rosdistro/pull/32291
* ~~ompl~~ https://github.com/ros/rosdistro/pull/32292
[...]

PR #32075, opened 2022-02-07
Update mrpt2 release repository.

nuclearsandwich:
This release repository was forked into ros2-gbp for the Galactic migration in April 2021 and has not been released into the upstream repository since.

Since the [upcoming Rolling platform migration](https://github.com/ros/rosdistro/pull/32036) will again branch the release repository into ros2-gbp I recommend that we update it pre-emptively to reduce churn in the bloom configuration branches.

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1921,7 +1921,7 @@ repositories:
     release:
       tags:
         release: release/rolling/{package}/{version}
-      url: https://github.com/mrpt-ros2-pkg-release/mrpt2-release.git
+      url: https://github.com/ros2-gbp/mrpt2-release.git
       version: 2.1.3-2
     source:
       type: git


nuclearsandwich:
FYI @jlblancoc

jlblancoc:
Thanks for the ping @nuclearsandwich , but I'm afraid I missed the big picture :-) 
Would it be better to delete the https://github.com/mrpt-ros2-pkg-release/mrpt2-release repo and release future versions directly to the ros2-gbp one? (would need write access, I guess?).

nuclearsandwich:
> Would it be better to delete the https://github.com/mrpt-ros2-pkg-release/mrpt2-release repo and release future versions

It's probably not a good idea to delete the repository because people using older copies of the rosdistro repository may still try to find artifacts hosted there but you could archive that repository and stop using it.

> Thanks for the ping @nuclearsandwich , but I'm afraid I missed the big picture :-)

In order to support the automation surrounding the Rolling distribution [REP-2002](https://www.ros.org/reps/rep-2002.html) release repositories will need to be [hosted in the ros2-gbp organization on GitHub](https://www.ros.org/reps/rep-2002.html#managing-build-farm-writable-release-repositories). For repositories that aren't the migration tool will create an empty repository and push release artifacts for the migrated distribution to it. When we created the Galactic distribution from Rolling a release repository for ros2-gbp was created in this manner. As we get ready to migrate Rolling to Ubuntu 22.04 I'm doing some pro-active work to make sure that repositories which have been previously forked by the migration tool are synchronized with upstream so they don't get forked again, further bifurcating the release histories.

> (would need write access, I guess?).

You are already a member of a team within the ros2-gbp organization with push access to the ros2-gbp/mrpt2-release repository so you should be able to make future releases using that releases repository without issues. If you do have any please let me know!

jlblancoc (approved):
:+1:

PR #32077, opened 2022-02-07
Update roboception release repositories.

nuclearsandwich:
These release repositories were forked into ros2-gbp for the Galactic migration in April 2021.
Some packages have not been released into the upstream repository since but some have.

For the packages which have had releases these have been merged into the ros2-gbp branch on 2022-02-07.

Since the [upcoming Rolling platform migration](https://github.com/ros/rosdistro/pull/32036) will again branch the release repositories into ros2-gbp I recommend that we update them pre-emptively to reduce churn in the bloom configuration branches.

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2564,7 +2564,7 @@ repositories:
     release:
       tags:
         release: release/rolling/{package}/{version}
-      url: https://github.com/roboception-gbp/rc_common_msgs_ros2-release.git
+      url: https://github.com/ros2-gbp/rc_common_msgs_ros2-release.git
       version: 0.5.3-2
     source:
       test_pull_requests: true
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
FYI @flixr

flixr:
Sounds good to me!
Let me know if there is something I should do on our side.

flixr:
@nuclearsandwich what about the packages that are still in the roboception gbp repo (rcdiscover and rc_reason_clients)?
Should we somehow move all of them to the ros2-gbp repo?

nuclearsandwich:
> what about the packages that are still in the roboception gbp repo (rcdiscover and rc_reason_clients)?
> Should we somehow move all of them to the ros2-gbp repo?

Yes. The first pass that I've done was specifically to cover repositories which have been moved to ros2-gbp previously but then moved back to their external repositories before we were able to communicate with the maintainers about the changes (that's on us not you). In order to get the Jammy transition done with all reasonable haste I am waiting to update release repositories that aren't in that situation (that is, have not yet been copied into ros2-gbp but will be)

PR #32113, opened 2022-02-09
Use python3-scipy on ROS Noetic/Debian Buster

rgov:
Not adding a new dependency here just correcting how scipy gets installed on Debian Buster / ROS Noetic. Otherwise this will install the Python 2 version.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4438,7 +4438,9 @@ python-schedule:
   ubuntu: [python-schedule]
 python-scipy:
   arch: [python2-scipy]
-  debian: [python-scipy]
+  debian:
+    '*': [python-scipy]
+    buster: [python3-scipy]
   freebsd: [py-scipy]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -4438,7 +4438,9 @@ python-schedule:
   ubuntu: [python-schedule]
 python-scipy:
   arch: [python2-scipy]
-  debian: [python-scipy]
+  debian:
+    '*': [python-scipy]
+    buster: [python3-scipy]
```
hidmic, line 4443: @rgov that is intended. `python-scipy` is a Python 2 dependency. `python3-scipy` is the Python 3 dependency you're looking for. And it is included already.

rgov:
@hidmic Hmm. Why does `python-scipy` install the `python3-scipy` package on Ubuntu then? Shouldn't it just be not available? Why are there any Python 2 packages on Noetic if it's not supported?

PR #32119, opened 2022-02-10
message_synchronizer: 0.0.2-1 in 'foxy/distribution.yaml' [bloom]

hakuturu583:
Increasing version of package(s) in repository `message_synchronizer` to `0.0.2-1`:

- upstream repository: https://github.com/OUXT-Polaris/message_synchronizer.git
- release repository: https://github.com/OUXT-Polaris/message_synchronizer-release.git
- bloom version: `0.10.7`

## message_synchronizer

```
* Setup workflow (#30 <https://github.com/OUXT-Polaris/message_synchronizer/issues/30>)
  Co-authored-by: MasayaKataoka <mailto:ms.kataoka@gmail.com>
* add boost to the depends
* Setup workflow (#29 <https://github.com/OUXT-Polaris/message_synchronizer/issues/29>)
* Contributors: MasayaKataoka, wam-v-tan
```

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2160,6 +2160,21 @@ repositories:
       url: https://github.com/ros2/message_filters.git
       version: foxy
     status: maintained
+  message_synchronizer:
+    doc:
+      type: git
+      url: https://github.com/OUXT-Polaris/message_synchronizer.git
+      version: master
+    release:
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/OUXT-Polaris/message_synchronizer-release.git
+      version: 0.0.2-1
+    source:
+      type: git
+      url: https://github.com/OUXT-Polaris/message_synchronizer.git
+      version: master
+    status: developed
   micro_ros_diagnostics:
     doc:
       type: git


hidmic (commented):
@hakuturu583 this gives me `message_filters` vibes. What is this `message_synchronizer` package adding that is not covered by the core package? The name is also a bit too generic.

hakuturu583:
> this gives me message_filters vibes. What is this message_synchronizer package adding that is not covered by the core package? The name is also a bit too generic.

message filters cannot synchronize topics when the one of the topics are dropped, and my package can synchronize this.
I need this feature ASAP, so I made this package.
I think it is not a too generic name.

sloretz (approved):
> If we accept this, we'd surely have to accept another message synchronizer package that does gap filling with a user-defined message, and another one that performs linear interpolation, and so on.

I think that's fine. IIRC the advice for getting new messages to `common_msgs/common_interfaces` in the past included making similar custom messages in a different package first to prove their worth, so making similar code features available in a different package seems reasonable to me too.

This is a good question for @clalancette though. Thoughts?

clalancette:
> I think that's fine. IIRC the advice for getting new messages to `common_msgs/common_interfaces` in the past included making similar custom messages in a different package first to prove their worth, so making similar code features available in a different package seems reasonable to me too.

Yeah, that seems reasonable to me as well.  If it becomes something that people really are reusing over and over, we can think about integrating it into the core, but until then a separate package is the way to go.

That said, I do share some of @hidmic's concerns with the name.  I think it will be easy to confuse this with `message_filters`.  According to the rules in REP-144 (https://www.ros.org/reps/rep-0144.html#global-rules point 4 and https://www.ros.org/reps/rep-0144.html#naming-rules), we probably want to call this something like `polaris_message_synchronizer`.  Again, if it becomes something widely used, we can rename it.

tfoote (changes requested):
Reading this over, the name doesn't meet our conventions. I believe that `polaris_message_synchronizer` or other prefix would make sense.

tfoote:
Reading through the example this looks like an ExactTime synchronizer with a timeout. And what it appears is that your implementation will return A or B if the timeout passes even if the matching message has not been received. This seems like it should be able to be implemented as a new filter policy which can plug into the standard message_filters. Adding the timeout at https://github.com/ros2/message_filters/blob/210fcb02a39bb2e674d6c4ad6e87526aa3927481/include/message_filters/sync_policies/exact_time.h#L159 instead of only checking for `full` would likely serve the same functionality.

hakuturu583:
@tfoote Thanks. I close this PR.

PR #31860, opened 2022-01-20
xacro: 1.14.11-1 in 'noetic/distribution.yaml' [bloom]

rhaschke:
Increasing version of package(s) in repository `xacro` to `1.14.11-1`:

- upstream repository: https://github.com/ros/xacro.git
- release repository: https://github.com/ros-gbp/xacro-release.git
- distro file: `noetic/distribution.yaml`
- bloom version: `0.10.7`
- previous version for package: `1.14.10-1`

## xacro

```
* Fix resolving of macros and properties declared and used in/from a namespace (#297 <https://github.com/ros/xacro/issues/297>, #306 <https://github.com/ros/xacro/issues/306>)Macros and properties that are declared within a namespaced include shouldn't require the namespace prefix when used within the namespace.
* Perform expression evaluation in comments (#300 <https://github.com/ros/xacro/issues/300>)
* Expose ``xacro.arg()`` to facilitate access to substitution args
* Fix scoped macro evaluation
  
    * Replace ``[Macro|Property]NameSpace`` with common ``NameSpace`` class derived from ``Table``
    * Use the scoped macro Table
  
* Contributors: Robert Haschke
```

PR #32139, opened 2022-02-11
Revert "xacro: 1.14.11-1 in 'noetic/distribution.yaml' [bloom]"

sloretz:
Reverts ros/rosdistro#31860

@rhaschke While options are being discussed in ros/xacro#300, what do you think about reverting to the previous release? I can do a sync over the weekend.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -10446,7 +10446,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/ros-gbp/xacro-release.git
-      version: 1.14.11-1
+      version: 1.14.10-1
     source:
       test_pull_requests: true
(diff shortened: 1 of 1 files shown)


rhaschke:
Reverting to 1.14.10 might be the quickest bet. However, I plan to implement and test a fixup tomorrow. So, if this still suffices for a sync, I suggest to wait, because 1.14.11 includes a fixup for a regression introduced already in 1.14.10 :cry:

sloretz:
> However, I plan to implement and test a fixup tomorrow. So, if this still suffices for a sync, I suggest to wait, because 1.14.11 includes a fixup for a regression introduced already in 1.14.10

Got it. I'll start a sync freeze now in anticipation of the xacro release tomorrow.

PR #32140, opened 2022-02-11
lockmount_description: 0.0.2-2 in 'melodic/distribution.yaml' [bloom]

tonybaltovski:
Increasing version of package(s) in repository `lockmount_description` to `0.0.2-2`:

- upstream repository: https://github.com/clearpathrobotics/lockmount_description.git
- release repository: https://github.com/clearpath-gbp/lockmount_description-release.git
- bloom version: `0.10.7`

## lockmount_description

```
* Remove a copy & paste error in the cmake file that's breaking the install
* Contributors: Chris Iverach-Brereton
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6074,6 +6074,21 @@ repositories:
       url: https://github.com/clearpathrobotics/lms1xx.git
       version: melodic-devel
     status: maintained
+  lockmount_description:
+    doc:
+      type: git
+      url: https://github.com/clearpathrobotics/lockmount_description.git
+      version: main
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/clearpath-gbp/lockmount_description-release.git
+      version: 0.0.2-2
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/lockmount_description.git
+      version: main
+    status: maintained
   log_view:
     doc:
       type: git


hidmic (approved):
@tonybaltovski the release itself looks to good to me, but I'd much rather prefer a less generic name. I don't have a good one to propose though. @sloretz ?

civerachb-cpr:
The specific model of mount this replicates is https://www.fath24.us.com/portfolio/products/pivot-joint-25-1-with-locking-lever

We could rename it to `fath_pivot_mount_description`?  Or if we want to avoid including the brand name would `locking_pivot_mount_description` be sufficiently specific while not being needlessly verbose?

sloretz (approved):
> I'd much rather prefer a less generic name. I don't have a good one to propose though. @sloretz ?

It could be prepended with the name, following https://www.ros.org/reps/rep-0144.html#naming-rules

> if a package is specialized for a hardware piece, prepend its name

It might be `fath_lockmount_description`. That said, `lockmount_description` seems unlikely to conflict with anything else, so this seems fine to me.

Another good question for @clalancette though. Thoughts?

clalancette:
> Another good question for @clalancette though. Thoughts?

I'd go with `fath_lockmount_description` as a balance between length and the naming rules in REP-144.  But `fath_pivot_mount_description` and `locking_pivot_mount_description` would also be fine with me.

tfoote (changes requested):
I'm going to mark this as request changes so that we can track it for the rename instead of giving a sense of pending.

tonybaltovski:
I am about to change this to `locking_pivot_mount`.  It was merged for Noetic: https://github.com/ros/rosdistro/pull/32138 Shall I change both?

tfoote:
Great. A rerelease with the consistent name for noetic would make sense too.

tfoote:
I'll close this as you'll do a new bloom release. Please refer back to this in a comment on the new PR.

PR #32245, opened 2022-02-18
Skip releasing ros1_bridge.

nuclearsandwich:
The ros1_bridge relies on ROS Noetic packages which are not available in Ubuntu Jammy.


clalancette (approved):
In theory, we should be able to build the `ros1_bridge` against any version of ROS 1, including the packages that are in Debian/Ubuntu.  In practice, we've never tested this and we probably would need to do some work on the rosdep keys and in the release repository to make this work.  So I think this is the correct way to disable this for now until we can do that work.

PR #32272, opened 2022-02-22
Add python3-moderngl-pip, python3-scipy-pip, python3-objloader-pip

bcbuckles:

## Package name:

-python3-moderngl-pip
-python3-scipy-pip
-python3-objloader-pip

## Package Upstream Source:

https://github.com/moderngl/moderngl
https://github.com/scipy/scipy
https://github.com/szabolcsdombi/objloader

## Purpose of using this:

python3-moderngl-pip - moderngl is a very helpful python wrapper for opengl that simplifies handling of vertex buffer and array binding.

python3-scipy-pip - the scipy version in apt package manager is outdated.  The newer version in pip was required for my project

python3-objloader-pip - this python package handles loading of .obj files, useful for building 3D applications in opengl, etc.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8768,6 +8768,10 @@ rpy2:
   gentoo: [=dev-python/rpy-2*]
   opensuse: [python3-rpy2]
   ubuntu: [python-rpy2]
+scipy-pip:
+  ubuntu:
+    pip:
+      packages: [scipy]
 sphinxcontrib-bibtex-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8011,6 +8011,10 @@ python3-scipy:
       depends: [gfortran]
       packages: [scipy]
   ubuntu: [python3-scipy]
+python3-scipy-pip:
+  ubuntu:
+    pip:
+      packages: [scipy]
 python3-scp:
   debian: [python3-scp]
   fedora: [python3-scp]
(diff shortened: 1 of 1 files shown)


cottsay (approved):
Please keep in mind that using pip rules in your package manifests will prevent the package from being released in Bloom and built on the buildfarm.

Please add links to the packaging dashboard in the future:
* https://pypi.org/project/moderngl/
* https://pypi.org/project/objloader/
* https://pypi.org/project/scipy/

nuclearsandwich (dismissed):

rosdep/python.yaml
```diff
@@ -8003,6 +8011,10 @@ python3-scipy:
       depends: [gfortran]
       packages: [scipy]
   ubuntu: [python3-scipy]
+python3-scipy-pip:
+  ubuntu:
+    pip:
+      packages: [scipy]
```
nuclearsandwich, line 8017:
```suggestion
```

Please see the conversation in https://github.com/ros/rosdistro/pull/25615 for why we are not adding pip keys for packages which are available from distribution package managers.

nuclearsandwich (changes requested):
Since neither moderngl nor objloader appear to be packaged in Ubuntu at this time I have no objection to those additions.

However, since scipy is available in Debian and Ubuntu repositories we should not add a conflicting pip key there. I've linked a conversation on the line comment with further details.

tfoote:
> The newer version in pip was required for my project

Unfortunately this is part of being in a distribution we all need to use the consistent released version. For your personal project you're welcome to create your own rosdep rules that will override the upstream ones here. But for consistency across the community we don't want them in the public ruleset as it will likely break other users who expect the standard version.

PR #32274, opened 2022-02-23
Add python3-numba

garb4ge:

## Package name:

Python3 numba package.

## Package Upstream Source:

https://github.com/numba/numba

## Purpose of using this:

Numba provides high-performance numerics by generating machine code from python syntax.

## Distro packaging links:

  - https://packages.debian.org/sid/python3-numba
   - https://packages.ubuntu.com/focal/python3-numba

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7106,7 +7106,7 @@ python3-nose-yanc:
   ubuntu: [python3-nose-yanc]
 python3-numba:
   debian: [python3-numba]
-  ubuntu: 
+  ubuntu:
     bionic: [python3-numba]
     focal: [python3-numba]
 python3-numpy:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7104,6 +7104,9 @@ python3-nose-yanc:
   debian: [python3-nose-yanc]
   openembedded: [python3-nose-yanc@meta-ros-common]
   ubuntu: [python3-nose-yanc]
+python3-numba:
+  debian: [python3-numba]
+  ubuntu: [python3-numba]
 python3-numpy:
   alpine: [py3-numpy]
   debian: [python3-numpy]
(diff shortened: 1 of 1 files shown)


clalancette (approved):

rosdep/python.yaml
```diff
@@ -7104,6 +7104,9 @@ python3-nose-yanc:
   debian: [python3-nose-yanc]
   openembedded: [python3-nose-yanc@meta-ros-common]
   ubuntu: [python3-nose-yanc]
+python3-numba:
+  debian: [python3-numba]
+  ubuntu: [python3-numba]
```
nuclearsandwich, line 7109: This package is not currently available in Ubuntu Jammy. Since it is available in Debian Unstable I'm not sure whether using a wildcard and nulling out Jammy or explicitly listing the Ubuntu LTS releases that do contain the package is more likely to remain accurate going forward. I've opted for the more conservative approach.

```suggestion
  ubuntu: 
    bionic: [python3-numba]
    focal: [python3-numba]
```

nuclearsandwich (changes requested):
CI feedback needs to be addressed before this can be merged.

rosdep/python.yaml
```diff
@@ -7104,6 +7104,11 @@ python3-nose-yanc:
   debian: [python3-nose-yanc]
   openembedded: [python3-nose-yanc@meta-ros-common]
   ubuntu: [python3-nose-yanc]
+python3-numba:
+  debian: [python3-numba]
+  ubuntu:
```
nuclearsandwich, line 7109: Every time I use the web editor it is a mistake.

Fix trailing whitespace.

```suggestion
  ubuntu:
```

garb4ge:
Thanks for you work @nuclearsandwich !

nuclearsandwich (approved):
Now that I've fixed the problems I introduced this looks good!

PR #32306, opened 2022-02-26
fix: update libgmock-dev key for bullseye, jammy

wep21:

## Package name:

libgmock-dev

## Package Upstream Source:

https://github.com/google/googletest

## Purpose of using this:

Distro packaging links:

  - https://packages.debian.org/bullseye/libgmock-dev
   - https://packages.ubuntu.com/jammy/libgmock-dev
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL

ROSDISTRO NAME

# The source is here: 

http://sourcecode_url

# Checks

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3328,8 +3328,8 @@ libgmock-dev:
   opensuse: [gmock-devel]
   rhel: [gmock-devel]
   ubuntu:
-    focal: [libgmock-dev]
-    jammy: [libgmock-dev]
+    '*': [libgmock-dev]
+    bionic: null
 libgmp:
   debian: [libgmp-dev]
   fedora: [gmp-devel]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

rosdep/base.yaml
```diff
@@ -3330,6 +3329,7 @@ libgmock-dev:
   rhel: [gmock-devel]
   ubuntu:
     focal: [libgmock-dev]
+    jammy: [libgmock-dev]
```
tfoote, line 3332: This rule looks to be valid for focal and newer. Please update this to use `'*': [libgmock-dev]` and `bionic: null`

tfoote (changes requested):

rosdep/base.yaml
wep21: @tfoote updated at https://github.com/ros/rosdistro/pull/32306/commits/fa06ebc50bdfa64c19db509f033aa29ef212dd39

tfoote (approved):
Thanks for the quick update! It looks good now.

PR #32308, opened 2022-02-27
add libgazebo-dev for bullseye, jammy

wep21:

## Package name:

libgazebo-dev

## Package Upstream Source:

https://github.com/osrf/gazebo

## Purpose of using this:

add libgazebo-dev for bullseye, jammy

Distro packaging links:

  - https://packages.debian.org/bullseye/libgazebo-dev
   - https://packages.ubuntu.com/jammy/libgazebo-dev
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL

ROSDISTRO NAME

# The source is here: 

http://sourcecode_url

# Checks

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3180,6 +3180,11 @@ libftgl-dev:
   fedora: [ftgl-devel]
   gentoo: [media-libs/ftgl]
   ubuntu: [libftgl-dev]
+libgazebo-dev:
+  debian:
+    bullseye: [libgazebo-dev]
+  ubuntu:
+    jammy: [libgazebo-dev]
 libgazebo11-dev:
   debian:
     buster: [libgazebo11-dev]


clalancette (commented):
I'm honestly not sure if this is the thing we want to do.  Up until now, we've been careful to pick out a particular version of gazebo to install (see the keys below like `libgazebo11-dev`, etc).  I would guess that `libgazebo-dev` is a virtual package that installs whatever is latest.  @j-rivero @chapulina any thoughts on whether we should do this?

wep21:
`libgazebo11-dev` will be available in `bullseye` and `jammy`?

chapulina:
The official packages have dropped the `11` upstream on Jammy, so this should be ok. We could also choose to have the `libgazebo11-dev` key point to `libgazebo-dev` on Jammy, so the rosdep key is the same for Focal and Jammy; that should be easier on maintainers that have packages across both distros.

nuclearsandwich:
> We could also choose to have the `libgazebo11-dev` key point to `libgazebo-dev` on Jammy, so the rosdep key is the same for Focal and Jammy; that should be easier on maintainers that have packages across both distros.

Even though some of these packages are available upstream in Ubuntu it will be easier on package maintainers and consumers if we minimize the number of and difference in sources so that all Ignition and Gazebo dependencies for ROS come from the OSRF repositories if any of them do. I've had #31560 open in draft but I think we will likely end up modifying it to instead use packages imported into the bootstrap repository from the Open Source Robotics Foundation's Gazebo repositories as we're currently doing for the Ignition dependencies on Focal.

See https://github.com/ros-infrastructure/reprepro-updater/pull/143 for the first round of imports as the rest of Gazebo and Ignition Fortress become available for Jammy we'll import them as well.

chapulina:
> all Ignition and Gazebo dependencies for ROS come from the OSRF repositories if any of them do

Yeah that's true for Ignition, but we're not releasing Gazebo classic for Jammy into the OSRF repositories, see [this community post](https://community.gazebosim.org/t/gazebo-classic-on-jammy/1248). So ROS will just use Gazebo classic from Ubuntu for Jammy.

nuclearsandwich:
> Yeah that's true for Ignition, but we're not releasing Gazebo classic for Jammy into the OSRF repositories,

Okay, [this](https://github.com/ros/rosdistro/pull/31560/files#r774314697) is the comment which caused me to think otherwise.

In that case with @wep21's permission I'd like to close this in favor of taking #31560 out of draft and reviewing there.

wep21:
> In that case with @wep21's permission I'd like to close this in favor of taking https://github.com/ros/rosdistro/pull/31560 out of draft since it covers more platforms and reviewing there.

Okay, I close this once.

wep21:
> In that case with @wep21's permission I'd like to close this in favor of taking https://github.com/ros/rosdistro/pull/31560 out of draft since it covers more platforms and reviewing there.

All right. I close this once.

PR #32322, opened 2022-03-01
Add pcl_msgs source section to galactic & rolling

paulerikf:
Hopefully I've understood how to do this, please correct me if not.

Adds the source section back for the Galactic and Rolling indexes of pcl_msgs (it's already present in the Foxy distribution.yaml).  Mentioned here: https://github.com/ros-perception/pcl_msgs/issues/17

# Please Update This Package's index in the rosdistro.

Galactic / Rolling

# The source is here: 

https://github.com/ros-perception/pcl_msgs

# Checks

### DIFF
--- a/galactic/distribution.yaml
+++ b/galactic/distribution.yaml
@@ -2471,6 +2471,10 @@ repositories:
         release: release/galactic/{package}/{version}
       url: https://github.com/ros2-gbp/pcl_msgs-release.git
       version: 1.0.0-6
+    source:
+      type: git
+      url: https://github.com/ros-perception/pcl_msgs.git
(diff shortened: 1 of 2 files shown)


tfoote (approved):
This looks correct. And I cross checked the branches are right with the release: https://github.com/ros2-gbp/pcl_msgs-release/blob/ef1bf3f794f9ac738af5c964fca7ef5a6cf239c5/tracks.yaml#L89

I've triggered CI and we'll let it run to make sure there's no syntax issues. 

@paulbovbel FYI I'm assuming that there's not a problem with listing the sources in the rosdistro.

Issue #32362, opened 2022-03-02
qpOASES vendor


mhubii:
qpOASES vendor can't be linked against in ROS2, this includes foxy, galactic, rolling. The issue is caused by Autoware-AI/qpoases_vendor#5. qpOASES vendor should be removed from rosdistro until this issue is fixed

tfoote:
We don't generally deindex packages because they have a bug in them.  There are likely workarounds that people can use in the meantime, and removing it from the distro will remove indexing, discoverability through cross referencing, and make it harder for people to find the bugs etc. If anything could be considered it would be a rollback to a known good version. But either way please reach out to the maintainer to resolve the issue. They will be the ones to fix it. I see that it already has a PR on it: https://github.com/Autoware-AI/qpoases_vendor/pull/6

PR #32363, opened 2022-03-02
removes ros2/qpoases_vendor (#32362)

mhubii:
Remove `qpoases_vendor` until Autoware-AI/qpoases_vendor#5 is fixed. `qpoases_vendor` should not be released in ROS2. Tracked in issue #32362

# Checks

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -3294,17 +3294,6 @@ repositories:
       url: https://github.com/ros-visualization/python_qt_binding.git
       version: crystal-devel
     status: maintained
-  qpoases_vendor:
-    release:
-      tags:
-        release: release/foxy/{package}/{version}
-      url: https://github.com/Autoware-AI/qpoases_vendor-release.git
-      version: 3.2.3-2
-    source:
-      type: git
-      url: https://github.com/Autoware-AI/qpoases_vendor.git
-      version: ros2
-    status: maintained
   qt_gui_core:
     doc:
       type: git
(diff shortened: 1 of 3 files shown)


clalancette:
We don't typically do this without the consent of the original submitter/maintainer.  In this case, it looks like the rolling package was added in https://github.com/ros/rosdistro/pull/30031 .  @JWhitleyWork can you comment here on whether this is something we should do?  Are there other packages that depend on this?

clalancette (changes requested):
Marking as "Request changes" until we hear back from the original submitter.

JWhitleyWork:
@clalancette For some reason, no one had contacted me about this issue until now. I'm not associated with the AWF much anymore but I'm fine fixing this issue. https://github.com/Autoware-AI/qpoases_vendor/pull/6 should fix it which should resolve this and allow it to be released. It still needs a DCO fix but it appears to work.

clalancette:
> @clalancette For some reason, no one had contacted me about this issue until now. I'm not associated with the AWF much anymore but I'm fine fixing this issue. [Autoware-AI/qpoases_vendor#6](https://github.com/Autoware-AI/qpoases_vendor/pull/6) should fix it which should resolve this and allow it to be released. It still needs a DCO fix but it appears to work.

@JWhitleyWork Thanks for getting back to us, and looking into the issue.  It's appreciated.  I'm going to go ahead and close this out as we aren't likely to remove this from the distributions.  We'll keep an eye out for updated packages.  Thanks again.

PR #32418, opened 2022-03-08
Add python3-pyside2-pip to python.yaml

knzo25:

## Package name:

python3-pyside2-pip

## Package Upstream Source:

None

## Purpose of using this:

There are already entries for pyside2 for apt, but the current version of some distros like ubuntu are experimental (e.g., `python3-pyside2.qtcore (5.14.0-1~exp1ubuntu5`) and present issues that do not allow a seamless switch between pyqt5 and pyside2.

Although the [guidelines](https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md) detail the preference of apt over pip, in this case I hope this request can be considered, since does not break any current packages, while allowing the correct execution of pyside.

- pip: https://pypi.org/
  - [Link](https://pypi.org/project/PySide2/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7726,6 +7726,19 @@ python3-pyrr-pip:
   ubuntu:
     pip:
       packages: [pyrr]
+python3-pyside2-pip:
+  debian:
+    pip:
+      packages: [PySide2]
+  fedora:
+    pip:
+      packages: [PySide2]
+  osx:
+    pip:
+      packages: [PySide2]
+  ubuntu:
+    pip:
+      packages: [PySide2]
 python3-pyside2.qtopengl:
   arch: [pyside2]
   debian: [python3-pyside2.qtopengl]


tfoote (changes requested):
Unfortunately this will overlap and replace the version in apt for anyone on the same system as a package which uses this key. It will then cause untested behaviors based on any of the potential changes between the versions released in the distro and this version. There are clearly notably differences as you're requesting to use the newer version. 

The pip keys are also not able to be depended upon by any package which is to be built into binary packages. So unfortunately we cannot accept a pip key. 

Part of being in a distro is to work with the released versions. So unfortunately we need to work with everyone else and stick with the released versions. We don't have the resources to rebuild, test, and validate every potential dependency of pyside2 and potentially repackage it.

tfoote:
Note that you're welcome to use the rules in your own system or deployment, but we can't take them into the main database as there are too many potential side effects.

knzo25:
Hi, thank you very much to the detailed explanation.
We suspected as much but needed confirmation. I'll close the PR :bow:

PR #32422, opened 2022-03-08
A few updates to one of the metrics scripts

tfoote:
This is a script that we use sometimes for visualizing metrics. I needed to run it and updated it for new distros and fixed a few bits of error handling encountered in the Jan 2019 parsing.

PR #32423, opened 2022-03-08
fix mrpt branches failing in CI

tfoote:
The branches were changed but not updated in the distro

@jlblancoc FYI Please make sure to update the index before removing branches. 

This caused #32422 CI to fail and anyone trying to use melodic from source would fail. 

Default branch is ros1:
https://github.com/mrpt-ros-pkg/mrpt_msgs
https://github.com/mrpt-ros-pkg/mrpt_bridge


jlblancoc:
@tfoote Copy that. Totally agree, my bad, sorry for the mess!

tfoote:
Thanks for confirming.

PR #32430, opened 2022-03-10
Add python3-nlopt to python.yaml

knzo25:

## Package name:

python3-nlopt

## Package Upstream Source:

[Link](https://github.com/stevengj/nlopt)

## Purpose of using this:

In the `libnlopt-dev` some distros already include the python bindings for the library, but it is not the case for Ubuntu/Debian, so I would like to include them. 

Distro packaging links:

  - [Link](https://packages.debian.org/bullseye/python3-nlopt)
   - [Link](https://packages.ubuntu.com/focal/python3-nlopt)
  - [Link](https://src.fedoraproject.org/rpms/NLopt)
  - [Link](https://archlinux.org/packages/community/x86_64/nlopt/)
  - [Link](https://packages.gentoo.org/packages/sci-libs/nlopt)
  - None
  - None
  - None

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7101,6 +7101,12 @@ python3-networkx:
   fedora: [python3-networkx]
   nixos: [python3Packages.networkx]
   ubuntu: [python3-networkx]
+python3-nlopt:
+  arch: [nlopt]
+  debian: [python3-nlopt]
+  fedora: [NLopt]
+  gentoo: [sci-libs/nlopt]
+  ubuntu: [python3-nlopt]
 python3-nose:
   debian: [python3-nose]
   fedora: [python3-nose]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7107,7 +7107,7 @@ python3-nlopt:
     '*': [python3-nlopt]
     buster: null
     stretch: null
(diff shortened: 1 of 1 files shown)


jacobperron (approved):

rosdep/python.yaml
```diff
@@ -7101,6 +7101,18 @@ python3-networkx:
   fedora: [python3-networkx]
   nixos: [python3Packages.networkx]
   ubuntu: [python3-networkx]
+python3-nlopt:
+  arch: [nlopt]
+  debian:
+    '*': [python3-nlopt]
+    buster: null
+    stretch: null
+  fedora: [NLopt]
```
sloretz, line 7110: I think this one should be `python3-NLopt`: https://packages.fedoraproject.org/pkgs/NLopt/python3-NLopt/ which comes from here: https://src.fedoraproject.org/rpms/NLopt/blob/rawhide/f/NLopt.spec#_149

sloretz (changes requested):

rosdep/python.yaml
knzo25: Hi, thanks for the catch! I uploaded the updated version now

sloretz (approved):

ivanpauno (approved):

PR #32493, opened 2022-03-21
add python3-transitions rosdep key

zhtroy:

## Package name:

python-transitions

## Package Upstream Source:

https://github.com/pytransitions/transitions

## Purpose of using this:

Pull python transitions from Ubuntu Focal

   - https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-transitions&searchon=names

### DIFF
--- a/galactic/distribution.yaml
+++ b/galactic/distribution.yaml
@@ -4690,6 +4690,21 @@ repositories:
       url: https://github.com/SteveMacenski/slam_toolbox.git
       version: galactic
     status: maintained
+  slider_publisher:
+    doc:
+      type: git
+      url: https://github.com/oKermorgant/slider_publisher.git
+      version: ros2
+    release:
+      tags:
+        release: release/galactic/{package}/{version}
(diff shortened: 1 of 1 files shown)


mabelzhang (commented):
I think we prefer the system package since it exists https://packages.ubuntu.com/source/focal/python-transitions

zhtroy:
sorry, I only searched for "python-transitions", didn't notice the name has changed to "python3-transitions"

rosdep/python.yaml
```diff
@@ -5249,6 +5249,7 @@ python-transitions:
       pip:
         packages: [transitions]
   ubuntu:
+    focal: [python3-transitions]
```
ivanpauno, line 5252: you need to add a key named `python3-transitions` to `python.yaml`, instead of modifying the one for python2.

ivanpauno (changes requested):

rosdep/python.yaml
zhtroy: ok , I will create a new pull request for python3-transitions

PR #32522, opened 2022-03-25
Adding rm_description to documentation index for distro.

mlione:


noetic/distribution.yaml
```diff
@@ -7184,8 +7184,8 @@ repositories:
       - rm_common
       - rm_control
       - rm_dbus
-      - rm_description
       - rm_gazebo
+      - rm_description
```
ivanpauno, line 7188: Could you explain this?

noetic/distribution.yaml
```diff
@@ -7220,6 +7220,12 @@ repositories:
       url: https://github.com/rm-controls/rm_controllers.git
       version: master
     status: developed
+  rm_description:
+    doc:
+      type: git
+      url: https://github.com/rm-controls/rm_description.git
```
ivanpauno, line 7226: Was `rm_description` moved from `rm-controls/rm_control` to `rm-controls/rm_description`?
Are you still planning to release it from `rm-controls/rm_control` and only want to have a doc entry pointing to `rm-controls/rm_description`?

noetic/distribution.yaml
mlione: I dont know whether i should delete it, i want to separate rm_description from rm_control. At first i thought it should be deleted and i deleted it in distribution.yaml, but i haven't deleted it in rm_control. So i add it again in distribution.yaml.

mlione: @ivanpauno

noetic/distribution.yaml
mlione: I just want to separate rm_description from rm-controls/rm_control to rm-controls/rm_description. I plan to release it from rm-controlss/rm_description. @ivanpauno

mlione: @ivanpauno

noetic/distribution.yaml
ivanpauno: Yes, you need to delete it from `rm_control`.
The new entry for `rm_description` can be added in its first `bloom-release`.

The bad thing is that I think all old packages of `rm_description` will be removed, but IIUC there's no way around that.

PR #32534, opened 2022-03-28
rosbag2_storage_mcap: 0.1.0-1 in 'rolling/distribution.yaml' [bloom]

wep21:
Increasing version of package(s) in repository `rosbag2_storage_mcap` to `0.1.0-1`:

- upstream repository: https://github.com/ros-tooling/rosbag2_storage_mcap.git
- release repository: https://github.com/ros2-gbp/rosbag2_storage_mcap-release.git
- bloom version: `0.10.7`

## rosbag2_storage_mcap

```
* [1.0.0] Use Summary section for get_metadata() and seek(), implement remaining methods (#17 <https://github.com/wep21/rosbag2_storage_mcap/issues/17>)
* feat: add play impl (#16 <https://github.com/wep21/rosbag2_storage_mcap/issues/16>)
* chore: refine package.xml (#15 <https://github.com/wep21/rosbag2_storage_mcap/issues/15>)
* Don't throw when READ_WRITE mode is used; add .mcap file extension to recorded files (#14 <https://github.com/wep21/rosbag2_storage_mcap/issues/14>)
  I may be missing something, but from a cursory glance at [this code](https://github.com/ros2/rosbag2/blob/342d8ed3c1c4ae0411a4a92b60e79a728b8974b8/rosbag2_storage/src/rosbag2_storage/impl/storage_factory_impl.hpp#L108-L135), it appears that the APPEND mode is never used. This means we need to support READ_WRITE.
  This also adds a .mcap extension to recorded file names.
* Add dynamic message definition lookup (#13 <https://github.com/wep21/rosbag2_storage_mcap/issues/13>)
  Currently, an exception will be thrown if lookup fails.
* Switch C++ formatter to clang-format (#12 <https://github.com/wep21/rosbag2_storage_mcap/issues/12>)
  Remove uncrustify linter in favor of clang-format, which is easier to configure for use in VS Code format-on-save.
* Merge pull request #7 <https://github.com/wep21/rosbag2_storage_mcap/issues/7> from ros-tooling/jhurliman/reader-writer
  Reader and writer implementation
* uninitialized struct
* lint
* lint
* lint
* Reader and writer implementation
* Merge pull request #6 <https://github.com/wep21/rosbag2_storage_mcap/issues/6> from wep21/add-metadata-impl
  feat: add metadata impl
* feat: add metadata impl
* Merge pull request #5 <https://github.com/wep21/rosbag2_storage_mcap/issues/5> from wep21/mcap-storage-impl
  feat: mcap storage impl
* chore: update cmake minimum version
* chore: install mcap header
* chore: include mcap header
* fix: move fetch content into rosbag2 storage mcap
* Merge pull request #3 <https://github.com/wep21/rosbag2_storage_mcap/issues/3> from ros-tooling/emersonknapp/mcap_plugin_skeleton
  Add mcap storage plugin skeleton and CI
* Add rosbag2_storage_mcap skeleton
* Contributors: Daisuke Nishimatsu, Emerson Knapp, Jacob Bandes-Storch, John Hurliman, wep21
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -3655,6 +3655,21 @@ repositories:
       url: https://github.com/ros2/rosbag2_bag_v2.git
       version: master
     status: maintained
+  rosbag2_storage_mcap:
+    doc:
+      type: git
+      url: https://github.com/ros-tooling/rosbag2_storage_mcap.git
+      version: main
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/rosbag2_storage_mcap-release.git
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/ros-tooling/rosbag2_storage_mcap.git
+      version: main
+    status: developed
   rosbridge_suite:
     doc:
       type: git


wep21:
cc @jtbandes @jhurliman @emersonknapp

ivanpauno (approved):

wep21:
@ivanpauno @nuclearsandwich I found out that this error skips debian generator. Do you have any solution to resolve this?
```
Pre-verifying Debian dependency keys...
Running 'rosdep update'...
ROS Distro index file associate with commit '6b882056a72ab95d31c80b850523105623626c58'
New ROS Distro index url: 'https://raw.githubusercontent.com/ros/rosdistro/6b882056a72ab95d31c80b850523105623626c58/index-v4.yaml'
Key 'python3-conan-pip' resolved to '['conan']' with installer 'pip', which does not match the default installer 'apt'.
Failed to resolve python3-conan-pip on ubuntu:jammy with: Error running generator: The Debian generator does not support dependencies which are installed with the 'pip' installer.
python3-conan-pip is depended on by these packages: ['rosbag2_storage_mcap']
<== Failed
<== The following generator action reported that it is missing one or more
    rosdep keys, but that the key exists in other platforms:
'['/usr/bin/git-bloom-generate', '-y', 'rosdebian', '--prefix', 'release/rolling', 'rolling', '-i', '2', '--os-name', 'ubuntu']'
```
https://build.ros2.org/view/Rsrc_uJ/job/Rsrc_uJ__rosbag2_storage_mcap__ubuntu_jammy__source/6/console

ivanpauno:
> @ivanpauno @nuclearsandwich I found out that this error skips debian generator action. Do you have any solution to resolve this? I use conan in CMakeLists.txt.

You cannot have pip dependencies if you want to `bloom-release` a repository.
If there're debian packages alternatives for `conan`, we should add a key in rosdep for that.
If not, you'll have to avoid using conan in the `CMakeLists.txt`.

wep21:
@ivanpauno Thank you for your detailed explanation. Is it also acceptable to exec `pip install` in CMakeLists.txt?

ivanpauno:
> Is it also acceptable to exec pip install in CMakeLists.txt?

That would be kind of weird.
If you only need the dependency at build-time maybe it works, I really don't know.
It surely won't if you also need the dependency at runtime.
@nuclearsandwich probably knows better.
I would avoid to do that if possible.

jhurliman:
This is only at build-time, to install the package manager [conan](https://conan.io/).

nuclearsandwich:
Below I've written a ton of "don't do this" but I want to start off with "let's do this instead!". @wep21 can I work with you to create an `mcap_vendor` package instead of using conan (via pip) to vendor the mcap dependency in this package? For one, that would allow other tools to share the mcap dependency with rosbag2_storage_mcap and for another it removes the third-party dependency management obfuscating the dependency chain for this package and thus causing resistance from the ROS distribution maintainers to this package.

---

Whenever proscribing behavior, I try to provide a detailed rationale. So I apologize that the text below is on the long side. With that said, if any of it is unclear or could do with further details please let me know as it's important to me that the reasoning is well understood.

> @ivanpauno Thank you for your detailed explanation. Is it also acceptable to exec `pip install` in CMakeLists.txt?

Absolutely not. If you must depend on packages from pip, you _should_ use pip rosdep keys to do so. The trade-off to using pip at all is that the package cannot be released to the ROS build farm (or make use of ROS devel and PR jobs). Circumventing dependency declaration by installing dependencies during a build script is not an appropriate solution.

 The reason that we build ROS on top of distributions like Debian, RHEL, and Ubuntu is so that we can leverage the work that those distributions do to ensure a stable, inter-operable, and secure system. Circumventing that network of trust and quality assurance by installing dependencies during a build process makes it impossible for a user of that package to readily evaluate the package's dependencies and as a result their security posture when using that package.

> If there're debian packages alternatives for `conan`, we should add a key in rosdep for that.

In this case, pip is being used to bootstrap an _additional_ package management framework: conan which is then being used to install dependencies during the build. For the same reason that using pip in officially released packages is not allowed. Using conan to install dependencies is not something a released ROS package should do.

That's a lot of "don't do this". When you release a package for the ROS community one of the first things to look at is the dependencies of that package. It's possible that your package relies on libraries that are not available on all of the platforms supported by ROS 2 (see [REP-2000](https://www.ros.org/reps/rep-2000.html) for details on the platforms supported by each ROS 2 distribution). In that case there are two things you can do, assuming that you still want to release the package.

1. You could try to remove the dependency or find an alternative dependency which is available on the supported platforms.
2. If the dependency would be of general benefit to the ROS community, you could release it into ROS directly as a [third party package](http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty). For ROS 2, we often manage third party packages as vendor packages although we have never formalized what a vendor package is. (Maybe it's time to sit down and do so)

nuclearsandwich:
> 2\. we have never formalized what a vendor package is. (Maybe it's time to sit down and do so)

Here's the TODO: https://github.com/ros-infrastructure/rep/pull/345

jtbandes:
> can I work with you to create an mcap_vendor package instead of using conan (via pip) to vendor the mcap dependency in this package?

Just FYI, since I'm not sure if it was clearly explained in this thread (since there has also been some other out-of-band communication), but conan is a **build time** dependency only. It is used to fetch C/C++ sources for `mcap` (and its dependencies, `fmt`, `lz4`, and `zstd`) but these sources are then compiled into the rosbag2_storage_mcap library and neither conan nor these C/C++ deps are actual runtime dependencies.

Does that change anything about your recommended solution? I'm not sure how we would apply the "make an [x]_vendor package" approach to all of these transitive dependencies (mcap, fmt, lz4, and zstd) since we really only want the _sources_ so they can be built into the rosbag2_storage_mcap library.

nuclearsandwich:
> Just FYI, since I'm not sure if it was clearly explained in this thread (since there has also been some other out-of-band communication), but conan is a **build time** dependency only.

Thanks for making sure that we're all on the same page. I do understand that conan is being used at build time only and while that is certainly less troublesome than a package which executes a pip installation at runtime or installation time, it is still not a supported case on the buildfarm and the resulting package is still including dependencies which aren't reflected when reviewing the package.xml or apt.

> but these sources are then compiled into the rosbag2_storage_mcap library

This is to a certain degree the primary objection and the use of pip/conan is just what made it obvious to the reviewers.

> Does that change anything about your recommended solution?

I still think that vendoring mcap is the only way that this package can continue to be released on the official build farm. There is already a [zstd_vendor](https://github.com/ros2/rosbag2/tree/master/zstd_vendor) package in rosbag2 in addition to being available from apt on Debian and Ubuntu and there are also apt packages for [lz4](https://packages.ubuntu.com/jammy/liblz4-dev) and [fmt](https://packages.ubuntu.com/jammy/libfmt-dev) in the latest releases of Debian and Ubuntu as well which means that building mcap as a vendor package is already possible.

> since we really only want the _sources_ so they can be built into the rosbag2_storage_mcap library. Maybe there is some precedent for this in the ROS package ecosystem?

The precedent, following the same procedures and conventions as the Linux distributions that ROS is based on, is to first package up dependencies and provide a curated set (usually a single version of each library) that the entire distribution may use before packaging software that relies on those libraries.

jhurliman:
@nuclearsandwich thanks for the feedback, we’ll get this fixed to not use conan or pip. The reason this is not using vendor packages for zstd / fmt is we are using the static linking API, methods that only exist when static linking and do not appear in the dynamic libraries. So it’s really not the same version of the libraries that exist in the ROS ecosystem with the same name. I’m looking into solutions using submodules or vendoring the relevant bits of code and will follow up when it’s repackaged to follow the stated guidelines.

PR #32574, opened 2022-03-29
franka_ros: 0.9.0-1 in 'noetic/distribution.yaml

marcbone:
update franka_ros package to version 0.9.0-1

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2507,7 +2507,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/frankaemika/franka_ros-release.git
-      version: 0.8.2-2
+      version: 0.9.0-1
     source:
       type: git
       url: https://github.com/frankaemika/franka_ros.git


ivanpauno (approved):

hidmic (commented):
@marcbone was this released using `bloom`? Ideally, I'd like to see a `bloom`-generated PR description with release details -- for bookkeeping.

marcbone:
So, the release was done by bloom. But not the PR. Since this release, I always get an error (401 unauthorized or something like that) when i want to create that PR automatically. And of course, every time I try, it will increase the "release_inc". Maybe next time I should try to delete my rosdistro fork before I try to run bloom-release. Is there a way to only create the PR without re-running bloom-release?

ivanpauno:
> And of course, every time I try, it will increase the "release_inc"

You can always use the `--pull-request-only` option of `bloom-release` to avoid that.

> Since this release, I always get an error (401 unauthorized or something like that) when i want to create that PR automatically

Did you check http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization?
If you cannot get that working, bloom started from some versions ago to print to the console the PR description and title, and you can manually open the PR with that data.

marcbone:
ah thanks. Token in the config was not valid anymore :facepalm: . Created a new pr #32615

PR #32593, opened 2022-03-30
Enable test Pull Requests for ament_acceleration

vmayoral:


hidmic (approved):
LGTM, as it looks [you have](https://github.com/ros-acceleration/ament_acceleration/issues/1#issuecomment-1081907133) set all the right permissions.

PR #32728, opened 2022-04-11
Update imu_tools and scan_tools URLs

mintar:
The repositories were moved to a new organization.

# Checks


cottsay (approved):
GitHub redirect appears to agree. Thanks for following up with this change.

PR #32737, opened 2022-04-12
Specify how to handle python packages only available in recent distributions.

SubaruArai:
Based on the information I got from #32698, I added instructions to add packages that are pip only in older distributions.

I'm not sure if this file is the right place to write, so I'm open for edits.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -276,7 +276,7 @@ See above for the Gentoo specific guidelines.
 #### pip mixed with system packages
 
 Some packages are only available as system packages in newer distributions.
-In such case, you can specify the default as system package, and pip for older distributions.
+In such case, you can specify the default as system package, and pip for older distributions for which the system packages are not available.
 
 For example:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -273,6 +273,23 @@ The `-pip` key should be removed when the package becomes available on all platf
 As a reminder `pip` rules should not be used on Gentoo.
 See above for the Gentoo specific guidelines.
 
+#### pip mixed with system packages
+
+Some packages are only available as system packages in newer distributions.
+
+In such case, you can specify the default as system package, and pip for older distributions.
+
+For example:
+
+```yaml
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -273,6 +273,23 @@ The `-pip` key should be removed when the package becomes available on all platf
 As a reminder `pip` rules should not be used on Gentoo.
 See above for the Gentoo specific guidelines.
 
+#### pip mixed with system packages
+
+Some packages are only available as system packages in newer distributions.
+
+In such case, you can specify the default as system package, and pip for older distributions.
```
nuclearsandwich, line 280: I don't think there's a need for a paragraph break here.

```suggestion
Some packages are only available as system packages in newer distributions.
In such case, you can specify the default as system package, and pip for older distributions.
```

nuclearsandwich (approved):
Thanks for taking the time to update the contribution guidelines. I think this looks good modulo a formatting suggestion but we should get a second reviewer to approve as well.

CONTRIBUTING.md
```diff
@@ -273,6 +273,22 @@ The `-pip` key should be removed when the package becomes available on all platf
 As a reminder `pip` rules should not be used on Gentoo.
 See above for the Gentoo specific guidelines.
 
+#### pip mixed with system packages
+
+Some packages are only available as system packages in newer distributions.
+In such case, you can specify the default as system package, and pip for older distributions.
```
tfoote, line 279:
```suggestion
In such case, you can specify the default as system package, and pip for older distributions for which the system packages are not available.
```

tfoote (approved):
Thanks that's good to have documented better. I added one suggested clarification to emphasize that the system package should not be available. To explicitly exclude the case that the system package is "too old".

CONTRIBUTING.md
SubaruArai: At first I thought it was a bit too descriptive. But after passing through auto-translators, I got the idea. Thanks!

clalancette (approved):

PR #32787, opened 2022-04-16
rosdep keys to build ignition from source on Ubuntu Jammy

ijnek:
Following keys are needed to satisfy rosdep for the [currently failing ros_ign build for jammy](https://build.ros2.org/job/Rdev__ros_ign__ubuntu_jammy_amd64/3/console)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1861,6 +1861,7 @@ ignition-gazebo6:
     buster: [libignition-gazebo6-dev]
   ubuntu:
     focal: [libignition-gazebo6-dev]
+    jammy: [libignition-gazebo6-dev]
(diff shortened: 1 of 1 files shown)


ijnek:
@nuclearsandwich I see there's actually something happening on https://github.com/ros-infrastructure/reprepro-updater/pull/153,  I'm fine with this PR being closed with the tests failing, but just quick note on when I can expect the rosdeps to be available would be nice!

tfoote:
@j-rivero I think just said that there were packages accepted upstream for Jammy. Maybe he's has the best understanding of where we should go with these.

ijnek:
This PR was superceded by https://github.com/ros/rosdistro/pull/32838

PR #32788, opened 2022-04-18
nlohmann_json_schema_validator_vendor: 0.1.0-2 in 'rolling/distribution.yaml' [bloom]

marcoag:
Increasing version of package(s) in repository `nlohmann_json_schema_validator_vendor` to `0.1.0-2`:

- upstream repository: https://github.com/open-rmf/nlohmann_json_schema_validator_vendor
- release repository: https://github.com/ros2-gbp/nlohmann_json_schema_validator_vendor-release.git
- bloom version: `0.10.7`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2104,7 +2104,7 @@ repositories:
   nlohmann_json_schema_validator_vendor:
     doc:
       type: git
-      url: https://github.com/ros2-gbp/nlohmann_json_schema_validator_vendor-release.git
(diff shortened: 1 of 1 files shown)


nuclearsandwich (dismissed):
It would be best if the package.xml for this vendor package included the license of the vendored library (as in [this example](https://github.com/ros2/tinydir_vendor/blob/f1ee71b2fb02b24ee40123a905b16922938ca80a/package.xml#L9-L10)).

I don't think we need to reject this release and wait for a future one but I'd like to see the devel branch updated with license information so that future releases incorporate it.

PR #32789, opened 2022-04-18
pybind11_json_vendor: 0.1.0-1 in 'rolling/distribution.yaml' [bloom]

marcoag:
Increasing version of package(s) in repository `pybind11_json_vendor` to `0.1.0-1`:

- upstream repository: https://github.com/open-rmf/pybind11_json_vendor
- release repository: https://github.com/ros2-gbp/pybind11_json_vendor-release.git
- bloom version: `0.10.7`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2582,6 +2582,21 @@ repositories:
       url: https://github.com/ros-perception/pointcloud_to_laserscan.git
       version: rolling
     status: maintained
+  pybind11_json_vendor:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@marcoag I merged this before I remembered to check the license of the upstream package. Since vendor packages can include builds of the vendored library it is important to include the vendored library's license in the package.xml.

There isn't a "perfect" way to do this but here is an example of the current convention https://github.com/ros2/tinydir_vendor/blob/f1ee71b2fb02b24ee40123a905b16922938ca80a/package.xml#L9-L10

PR #32792, opened 2022-04-18
add rosdep key: pickleDB

zhtroy:

## Package name:

python-pickleDB

## Package Upstream Source:

https://pypi.org/project/pickleDB/

## Purpose of using this:

I use pickleDB to do lightweight data persistence

Distro packaging links:

  -  not found https://search.debian.org/cgi-bin/omega?DB=zh-cn&P=pickleDB
   - not found https://packages.ubuntu.com/search?suite=impish&section=all&arch=any&keywords=pickleDB&searchon=names

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5257,15 +5257,24 @@ python-transitions:
       pip:
         packages: [transitions]
 python3-transitions:
-  debian: 
-    '*': [python3-transitions]
+  debian:
+    sid: [python3-transitions]
+    bookworm: [python3-transitions]
+    bullseye: [python3-transitions]
+    buster:
+      pip:
+        packages: [transitions]
     stretch:
       pip:
         packages: [transitions]
   ubuntu:
     '*': [python3-transitions]
-    xenial: null
-    trusty: null
+    xenial:
+        pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -3358,6 +3358,16 @@ python-pexpect:
   debian: [python-pexpect]
   gentoo: [dev-python/pexpect]
   ubuntu: [python-pexpect]
+python-pickleDB:
```
clalancette, line 3361: Since this is a pip-only key, please make this `python-pickleDB-pip`.  We may also want to add `python3-pickleDB-pip` for Python3.

clalancette (changes requested):

rosdep/python.yaml
zhtroy: changed name to python-pickleDB-pip

clalancette (approved):

tfoote (changes requested):
Good catch @clalancette  Please do use `python3-` as the prefix so it will be sorted in the same place an potential new debian packages that might conflict.

tfoote (approved):
Thanks for updating the key. lgtm

PR #32794, opened 2022-04-18
Error early on migrating repositories with no current release.

nuclearsandwich:
This uses a "bail out" condition to loudly error on a repository with no
release version (Such as from a previous migration).
I could see a case being made to silently skip these repositories rather
than raising an error but I think that since the script is rarely run
and the list of repositories with errors is one of the major
communication outputs I wanted to start with an error condition (which
still allows the rest of the migration to proceed).

Adding this check will also prevent bugs when trying to release packages
with an empty version since some later operations require a version
string.


audrow (approved):
Looks good to me!

PR #32804, opened 2022-04-19
[migration-tools] Don't assume that every release repository has a tracks.yaml.

nuclearsandwich:
A malformed release repository could be missing its tracks.yaml
configuration file. When this happens it should raise a ValueError as
with any other failure during the release process so that the migration
can continue.


audrow (approved):
Looks good to me!

PR #32808, opened 2022-04-19
[migration-tools] Bump the minimum versions of bloom and rosdep to the latest.

nuclearsandwich:
We want to use bloom 0.11.0 for the Humble migration.

The rosdep release isn't strictly necessary but is coming along for the ride.


audrow (approved):
Looks good to me!

PR #32838, opened 2022-04-21
Ignition Fortress rosdep keys for Jammy

chapulina:

* Part of https://github.com/ignition-tooling/release-tools/issues/459
* Depends on https://github.com/ros-infrastructure/reprepro-updater/pull/153
* Follow up to https://github.com/ros/rosdistro/pull/31591

## Package name:

[Ignition Fortress libraries](https://ignitionrobotics.org/docs/fortress)

## Package Upstream Source:

All libraries are on this org: https://github.com/ignitionrobotics/

## Purpose of using this:

As defined on [REP-2000](https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027), Ignition Fortress will be the official Ignition version to be used with ROS Humble. This adds keys for Ubuntu Jammy.

Once the rosdep keys are available, we'll make a release of [ros_ign](https://github.com/ignitionrobotics/ros_ign/) into Rolling using Fortress. ROS packages can interface with Ignition Gazebo through `ros_ign`, and they can also use Ignition libraries directly.

  - http://packages.osrfoundation.org/gazebo/debian-stable/
   - http://packages.osrfoundation.org/gazebo/ubuntu-stable/
  - https://github.com/osrf/homebrew-simulation/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6780,14 +6780,14 @@ python3-ifcfg:
   ubuntu:
     bionic: [python3-ifcfg]
     focal: [python3-ifcfg]
-python3-ignition-math6:
-  ubuntu:
-    focal: [python3-ignition-math6]
-    jammy: [python3-ignition-math6]
 python3-ignition-gazebo6:
   ubuntu:
     focal: [python3-ignition-gazebo6]
     jammy: [python3-ignition-gazebo6]
+python3-ignition-math6:
+  ubuntu:
+    focal: [python3-ignition-math6]
+    jammy: [python3-ignition-math6]
 python3-imageio:
   debian: [python3-imageio]
   gentoo: [dev-python/imageio]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6780,6 +6780,10 @@ python3-ifcfg:
   ubuntu:
     bionic: [python3-ifcfg]
     focal: [python3-ifcfg]
+python3-ignition-math6:
+  ubuntu:
+    focal: [python3-ignition-math6]
+    jammy: [python3-ignition-math6]
 python3-imageio:
   debian: [python3-imageio]
   gentoo: [dev-python/imageio]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6784,6 +6784,10 @@ python3-ignition-math6:
   ubuntu:
     focal: [python3-ignition-math6]
     jammy: [python3-ignition-math6]
+python3-ignition-gazebo6:
+  ubuntu:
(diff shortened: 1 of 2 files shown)


nuclearsandwich (approved):
Just imported these packages from the OSRF repositories.

clalancette (commented):
So I'll point out that some of these keys I can't resolve currently (on an arm64 box, which is what I've been testing recently):

```
$ sudo apt-cache policy ignition-fortress
N: Unable to locate package ignition-fortress
$ sudo apt-cache policy libignition-gazebo6-dev
libignition-gazebo6-dev:
  Installed: (none)
  Candidate: (none)
  Version table:
$ sudo apt-cache policy libignition-gazebo6-plugins
N: Unable to locate package libignition-gazebo6-plugins
$ sudo apt-cache policy libignition-sensors6-dev
N: Unable to locate package libignition-sensors6-dev
```

The rest of the keys resolve just fine.  Is that expected?

rosdep/python.yaml
```diff
@@ -6780,6 +6780,10 @@ python3-ifcfg:
   ubuntu:
     bionic: [python3-ifcfg]
     focal: [python3-ifcfg]
+python3-ignition-math6:
+  ubuntu:
+    focal: [python3-ignition-math6]
+    jammy: [python3-ignition-math6]
```
j-rivero, line 6786:
```suggestion
    jammy: [python3-ignition-math6]
python3-ignition-gazebo6:
  ubuntu:
    focal: [python3-ignition-gazebo6]
    jammy: [python3-ignition-gazebo6]
```

j-rivero (approved):
Suggestion to add ignition-gazebo python bindings, otherwise looks good.

rosdep/python.yaml
chapulina: Good call, thanks!

nuclearsandwich:
> So I'll point out that some of these keys I can't resolve currently (on an arm64 box, which is what I've been testing recently):

The arm64 packages likely failed to import due to version discrepancies between amd64 and arm64 packages available. 

Since CI is only checking the amd64 repositories at this time I think it makes sense to get these keys up for amd64 usage, thus unblocking ros_ign on Jammy, and address the platform inconsistencies in followup.

rosdep/python.yaml
nuclearsandwich: This introduced an error:
   ERR: list out of alphabetical order line 6787.  'python3-ignition-gazebo6' should come before 'python3-ignition-math6'

chapulina: fixed 🙃

PR #32876, opened 2022-04-25
add python3-kivy, python3-marshmallow-dataclass-pip and python3-tables to to rosdep/python.yaml

3473f:

## Package name:

python3-kivy
python3-marshmallow-dataclass-pip
python3-tables

## Package Upstream Source:

kivy: https://github.com/kivy/kivy
marshmallow-dataclass: https://github.com/lovasoa/marshmallow_dataclass
tables: https://github.com/PyTables/PyTables

## Purpose of using this:

Kivy: open source Python framework for rapid development of applications that make use of innovative user interfaces, such as multi-touch apps.
Marshmallow-dataclass: Automatic generation of marshmallow schemas from dataclasses.
Tables: managing hierarchical datasets and designed to efficiently and easily cope with extremely large amounts of data. 

Distro packaging links:

Kivy:
- Debian: https://www.debian.org/distrib/packages
  - https://packages.debian.org/bullseye/python3-kivy
  - https://packages.ubuntu.com/focal/python3-kivy

marshmallow-dataclass:
- PiPy: https://pypi.org/
  - https://pypi.org/project/marshmallow-dataclass/

PyTables:
- Debian: https://www.debian.org/distrib/packages
  - https://packages.debian.org/bullseye/python3-tables
  - Buster:
    - PiPy: https://pypi.org/
      - https://pypi.org/project/Kivy/
  - https://packages.fedoraproject.org/pkgs/python-tables/python3-tables/
  -  https://packages.ubuntu.com/focal/python3-tables

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8588,7 +8588,11 @@ python3-systemd:
   nixos: [python3Packages.systemd]
   ubuntu: [python3-systemd]
 python3-tables:
-  debian: [python3-tables]
+  debian:
+    '*': [python3-tables]
+    buster:
+      pip:
+        packages: [Kivy]
   fedora: [python3-tables]
   ubuntu: [python3-tables]
 python3-tabulate:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8592,7 +8592,7 @@ python3-tables:
     '*': [python3-tables]
     buster:
       pip:
-        packages: [Kivy]
+        packages: [tables]
   fedora: [python3-tables]
   ubuntu: [python3-tables]
 python3-tabulate:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6956,7 +6956,11 @@ python3-kitchen:
     '*': [python3-kitchen]
     xenial: null
 python3-kivy:
-  debian: [python3-kivy]
+  debian:
+    '*': [python3-kivy]
+    buster:
+      pip:
+        packages: [Kivy]
   ubuntu: [python3-kivy]
 python3-lark-parser:
   alpine: [py3-lark-parser]
@@ -8588,11 +8592,7 @@ python3-systemd:
   nixos: [python3Packages.systemd]
   ubuntu: [python3-systemd]
 python3-tables:
-  debian:
-    '*': [python3-tables]
-    buster:
-      pip:
-        packages: [tables]
+  debian: [python3-tables]
   fedora: [python3-tables]
   ubuntu: [python3-tables]
 python3-tabulate:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -8580,6 +8588,10 @@ python3-systemd:
   debian: [python3-systemd]
   nixos: [python3Packages.systemd]
   ubuntu: [python3-systemd]
+python3-tables-pip:
```
tfoote, line 8591: This is packaged natively for example Ubuntu https://packages.ubuntu.com/jammy/python3-tables

Please update this to use the native packages on as many platforms as you can find, as per our contributing guide.

rosdep/python.yaml
```diff
@@ -6955,6 +6955,10 @@ python3-kitchen:
   ubuntu:
     '*': [python3-kitchen]
     xenial: null
+python3-kivy-pip:
```
tfoote, line 6958: This is also packaged upstream so pip rules should not be used for it. 

https://packages.ubuntu.com/jammy/python-kivy-examples

tfoote (changes requested):
At least two of these packages have native packages available and as such should be used from those sources not from pip.

The contributing guide lays this out here: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

Please check the various potential platforms for availability and updated your PR to use the native packaging keys. Also don't forget to update the key to remove the `-pip` suffix.

3473f:
Hello @tfoote, thanks for your feedback. I updated my pull request accordingly.

tfoote (approved):
Thanks for switching over. It would be great to add more platforms but I didn't find any in a quick search either so lgtm.

3473f:
> Thanks for switching over. It would be great to add more platforms but I didn't find any in a quick search either so lgtm.

I had to add python3-tables to debian buster through pip as it was not found causing the check to fail.

rosdep/python.yaml
```diff
@@ -6955,6 +6955,13 @@ python3-kitchen:
   ubuntu:
     '*': [python3-kitchen]
     xenial: null
+python3-kivy:
+  debian:
+    '*': [python3-kivy]
+    buster:
+      pip:
+        packages: [Kivy]
```
jacobperron, line 6963: I'm not sure we want to mix pip and system packages in the same rosdep rule. We can avoid the warning for buster by making the rule:

```
buster: null
```

jacobperron: @wjwwood  thoughts?

3473f: I was following the contribution guide in this case

> Some packages are only available as system packages in newer distributions. In such case, you can specify the default as system package, and pip for older distributions for which the system packages are not available.

jacobperron: Ah, cool. I found the section: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-mixed-with-system-packages

wjwwood (approved):

PR #32888, opened 2022-04-25
[humble] Update distribution.yaml

audrow:
This PR updates the repos file using the [`ros2-branching`](https://github.com/audrow/ros2-branching) repo for most of its changes and then adds some manual changes.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -86,7 +86,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ament/ament_cmake.git
-      version: master
+      version: humble
     release:
       packages:
       - ament_cmake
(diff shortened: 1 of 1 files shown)


humble/distribution.yaml
```diff
@@ -366,7 +366,8 @@ repositories:
   autoware_auto_msgs:
     doc:
       type: git
-      url: https://gitlab.com/autowarefoundation/autoware.auto/autoware_auto_msgs.git
+      url: >-
+        https://gitlab.com/autowarefoundation/autoware.auto/autoware_auto_msgs.git
```
clalancette, line 370: This change seems unrelated.  There are a couple more down below like this.

clalancette (changes requested):
Besides the comment I've left inline, the source branch for the `fastrtps` key needs to be changed to 2.6.x.  After that, this is probably good to go.

clalancette (approved):

PR #32898, opened 2022-04-26
Add python3-sysv-ipc package

meganetaaan:

## Package name:

python3-sysv-ipc

## Package Upstream Source:

http://semanchuk.com/philip/sysv_ipc/

## Purpose of using this:

Our private package depends on it. We need to get it from pip since the one on apt is old.

- Ubuntu: https://packages.ubuntu.com/focal/python3-sysv-ipc
- Debian: https://packages.debian.org/bullseye/python3-sysv-ipc

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4939,16 +4939,6 @@ python-systemd:
 python-sysv-ipc:
   debian: [python-sysv-ipc]
   ubuntu: [python-sysv-ipc]
-python-sysv-ipc-pip:
-  debian:
-    pip:
-      packages: [sysv-ipc]
-  fedora:
-    pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):

This appears to be packaged natively. https://packages.ubuntu.com/impish/python3-sysv-ipc as such we should use that version not the one from pip.

meganetaaan:
I updated the PR to use native `python3-sysv-ipc` package. Would you re-review it.

jacobperron (approved):

methylDragon (approved):
Will merge in.

Just note that the native version is 1.0.0 while the pip version is 1.1.0. for your internal uses!

PR #32903, opened 2022-04-26
Make humble branch names more consistent

audrow:
This PR makes the names used by humble branches more consistent.


clalancette (approved):
CI is failing because these branches don't exist yet.  Once the branches are created, manually kick off CI and this should be good to go.

I'll approve it in the meantime, assuming all of that will be done.

PR #32942, opened 2022-04-28
Update link to fedora packages

jacobperron:
The old link is no longer valid.

Alternatively, we could point to the package sources (like the PR template): https://src.fedoraproject.org/browse/projects/
I'm not sure if there's a preference were users reference packages from.


methylDragon:
I like https://src.fedoraproject.org/browse/projects/ because it looks less sparse :eyes:

sloretz (approved):

PR #33000, opened 2022-05-02
Add rospy_message_converter doc + source to foxy

mintar:


foxy/distribution.yaml
```diff
@@ -4939,6 +4939,17 @@ repositories:
       url: https://github.com/ros2/rosidl_typesupport_gurumdds.git
       version: foxy
     status: developed
+  rospy_message_converter:
+    doc:
+      type: git
+      url: https://github.com/uos/rospy_message_converter.git
```
jacobperron, line 4945: Please add a top-level LICENSE file to the source repository ([per our contributing guide](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-package-contents)). Thanks in advance!

mintar: Done!

Issue #33053, opened 2022-05-04
Package 'python-scipy' has no installation candidate in galactic AMD docker


RFRIEDM-Trimble:
When I add `<depend>python-scipy</depend>` to my package.xml, it tries to run an apt install of `python-scipy` with does not exist. It seems it could do either of the following:
`python3 -m pip install scipy` or `apt install python3-scipy` in my environment. 

Docker image: `docker pull osrf/ros:galactic-desktop`
Hash: `c78b28a21b49`

RFRIEDM-Trimble:
Instead, use python3-scipy as the candidate.

PR #33070, opened 2022-05-05
Feature/python scipy debian

RFRIEDM-Trimble:
This is to solve #33053 
Not sure if we also need to update the arm build; I don't have that hardware accessible.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4457,7 +4457,7 @@ python-schedule:
   ubuntu: [python-schedule]
 python-scipy:
   arch: [python2-scipy]
-  debian: [python3-scipy]
+  debian: [python-scipy]
   freebsd: [py-scipy]
   gentoo: [sci-libs/scipy]
   macports: [py27-scipy]
@@ -4478,6 +4478,8 @@ python-scipy:
     xenial_python3: [python3-scipy]
     yakkety_python3: [python3-scipy]
     zesty_python3: [python3-scipy]
+python3-scipy:
+  debian: [python3-scipy]
 python-scp:
   debian: [python-scp]
   fedora: [python-scp]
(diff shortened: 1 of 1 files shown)


methylDragon (approved):

sloretz (changes requested):
Thanks for the PR! Please add a new key `python3-scipy` instead of modifying this one.

RFRIEDM-Trimble:
> Thanks for the PR! Please add a new key `python3-scipy` instead of modifying this one.
> 
> https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-3

Fix (I think). I don't see how to test this, and according to documentation, I need to add the ubuntu key, but I'm not sure how to go about testing it all; I don't have environment for every version of ubuntu. 
`When adding a new key, Ubuntu and Debian are required, Fedora, Gentoo, and openSUSE are encouraged if the package also exists on those Linux distributions.`

RFRIEDM-Trimble:
> Thanks for the PR! Please add a new key `python3-scipy` instead of modifying this one.
> 
> https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-3

Fix (I think). I don't see how to test this, and according to documentation, I need to add the ubuntu key, but I'm not sure how to go about testing it all; I don't have environment for every version of ubuntu. 
`When adding a new key, Ubuntu and Debian are required, Fedora, Gentoo, and openSUSE are encouraged if the package also exists on those Linux distributions.`

tfoote:
You don't need to test it on all platforms if you can link to it, like you have to the ubuntu platform for all supported versions. Note that with the key rename it will need to be realphabetized.

RFRIEDM-Trimble:
It's actually already in there as python3-scipy; just another 4000 lines down. My bad, totally missed it. 

Any reason there isn't a pre-commit hook set up to do the following?
```
Keep everything in alphabetical order for better merging.
No trailing whitespace.
No blank lines.
```

tfoote:
Pre commit hooks have to be configured by the user. We can't automatically deploy them on a checkout. We just have the CI run them and document how to run the unit tests too which is also recommended for the user.

PR #33091, opened 2022-05-06
Add in a rosdep key for python3-transforms3d.

clalancette:
It is available in Ubuntu since Jammy, in Fedora since at
least Fedora 34.  It is also available in Debian Sid and Bookworm,
but not any released Debian distribution yet.

## Package name:

transforms3d

## Package Upstream Source:

https://github.com/matthew-brett/transforms3d

## Purpose of using this:

This dependency is being used by https://github.com/DLu/tf_transformations (which is released into all of Foxy, Galactic, Humble, and Rolling), though it is not currently declared as a dependency.  Thus, it doesn't work out-of-the-box.

Distro packaging links:

  - Not available in any released Debian distribution yet: https://packages.debian.org/search?keywords=transforms3d&searchon=names&suite=all&section=all
   - Released from Jammy onwards: https://packages.ubuntu.com/search?keywords=transforms3d&searchon=names&suite=all&section=all
  - Available since at least Fedora 34: https://pkgs.org/search/?q=transforms3d

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5251,8 +5251,13 @@ python-transforms3d-pip:
     pip:
       packages: [transforms3d]
   ubuntu:
-    pip:
-      packages: [transforms3d]
+    '*': null
+    bionic:
+      pip:
+        packages: [transforms3d]
+    focal:
+      pip:
+        packages: [transforms3d]
 python-transitions:
   debian:
     '*': [python-transitions]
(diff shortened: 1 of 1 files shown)


tfoote (approved):

tfoote:
Adding this we should simultaneously remove the old pip rule, otherwise people who use that will mask this rule. Or at least we should null it out for newer versions where the native versions are available.

clalancette:
I've mostly done this in a9bd5df .  Note that debian is a weird case here, because in all stable versions, it is only available via pip.  However, the packaged version is available in sid (the development version) and bookworm (the 'testing' version, that will become the next stable release).  So for now, I just added pip entries for it and we can update it with the packages once they are available in a stable version.

tfoote (approved):
That is weird that Ubuntu's pulling ahead a bit. But with that stated this looks like a good rule. It will need to be updated at some point when people will want to make debian packages on top of it in the future.

ivanpauno (approved):

PR #33106, opened 2022-05-09

jspricke:
## Package Upstream Source:

https://github.com/bombela/backward-cpp

## Purpose of using this:

Used in osrf_testing_tools_cpp.

Distro packaging links:

  - https://packages.debian.org/libbackward-cpp-dev
   - https://packages.ubuntu.com/libbackward-cpp-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2312,6 +2312,13 @@ libb64-dev:
   gentoo: [libb64]
   nixos: [libb64]
   ubuntu: [libb64-dev]
+libbackward-cpp-dev:
+  debian: [libbackward-cpp-dev]
+  opensuse: [backward-cpp]
+  ubuntu:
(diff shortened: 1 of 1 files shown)


methylDragon:
Looks like the package can't be resolved for some reason?

Also it would be good to add in that OpenSUSE key

methylDragon:
For the broken resolution, see the [review guidelines](https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md)

> If a package was added to e.g. Ubuntu Focal but isn’t available in Bionic or Xenial, the key should look like:
> ```
> mykey:
>   ubuntu:
>    '*': [mykey]
>    bionic: null
>    xenial: null
>```

tfoote (approved):

adityapande-1995 (approved):

PR #33248, opened 2022-05-17
[emoji] Add pip emoji package

tkmtnt7000:

## Package name:

python3-emoji
pip/pip3 emoji

## Package Upstream Source:

https://github.com/carpedm20/emoji/

## Purpose of using this:

I want to release the package that depends on this library. 

Distro packaging links: https://github.com/jsk-ros-pkg/jsk_robot

- Ubuntu: https://packages.ubuntu.com/jammy/python3-emoji
- Debian: https://packages.debian.org/bookworm/python3-emoji
- Fedora: https://packages.fedoraproject.org/pkgs/python-emoji/python3-emoji/
- opensuse: https://software.opensuse.org/package/python3-emoji
- rhel: https://src.fedoraproject.org/rpms/python-emoji
- Pip: https://pypi.org/project/emoji/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6381,6 +6381,9 @@ python3-elasticsearch:
 python3-emoji:
   debian:
     '*': [python3-emoji]
+    bullseye:
+      pip:
+        packages: [emoji]
     buster:
       pip:
         packages: [emoji]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1573,6 +1573,16 @@ python-elasticsearch:
   opensuse: [python-elasticsearch]
   ubuntu:
     bionic: [python-elasticsearch]
+python-emoji-pip:
+  debian:
+    pip:
+      packages: [emoji]
+  fedora:
+    pip:
+      packages: [emoji]
+  ubuntu:
+    pip:
+      packages: [emoji]
 python-empy:
   arch: [python2-empy]
   debian: [python-empy]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1573,6 +1573,16 @@ python-elasticsearch:
   opensuse: [python-elasticsearch]
   ubuntu:
     bionic: [python-elasticsearch]
+python-emoji-pip:
```
ivanpauno, line 1576: do you want to make it available for python3?
In that case, the key should be `python3-emoji-pip`.

ivanpauno: There's an ubuntu debian package available: https://packages.ubuntu.com/jammy/python3-emoji.

So consider adding a `python3-emoji` key instead

tfoote (changes requested):
If it's available from debian packages we should use that and not the pip rules. 

tkmtnt7000:
Thank you for quick reply.
Changed to use debian package if debian package is released.

tfoote (commented):
Thanks for the quick update. I see that it's also available for debian as of bookworm. https://packages.debian.org/bookworm/python3-emoji

Please use the same wildcard technique to list older version of debian via pip and the wildcard for goign forwardard. 

And the linter has identified that there's python3-emoji on rhel and python-emoji on suse.

hidmic (approved):
@tkmtnt7000 would you mind updating the PR description with [a package listing](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#rosdep-rules-contributions) for future reference?

tkmtnt7000:
I updated the PR description.

tfoote (approved):
lgtm

PR #33274, opened 2022-05-19
toposens-library: 1.2.4-4 in 'noetic/distribution.yaml' [bloom]

BarisYazici:
Increasing version of package(s) in repository `toposens-library` to `1.2.4-4`:

- upstream repository: https://gitlab.com/toposens/public/toposens-libraries.git
- release repository: https://gitlab.com/toposens/public/toposens-library-release.git
- distro file: `noetic/distribution.yaml`
- bloom version: `0.10.7`
- previous version for package: `1.2.4-3`

PR #33293, opened 2022-05-20
toposens-library: 1.2.4-1 in 'noetic/distribution.yaml' [bloom]

BarisYazici:
Increasing version of package(s) in repository `toposens-library` to `1.2.4-1`:

- upstream repository: https://gitlab.com/toposens/public/toposens-libraries.git
- release repository: https://gitlab.com/toposens/public/toposens-library-release.git
- bloom version: `0.10.7`
- previous version for package: `1.2.4-4`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -10185,7 +10185,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://gitlab.com/toposens/public/toposens-library-release.git
-      version: 1.2.4-4
+      version: 1.2.4-1
     source:
       type: git
       url: https://gitlab.com/toposens/public/toposens-library.git


noetic/distribution.yaml
```diff
@@ -10185,7 +10185,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://gitlab.com/toposens/public/toposens-library-release.git
-      version: 1.2.4-4
+      version: 1.2.4-1
```
tfoote, line 10188: This is a downgrade not an upgrade. Please comment if this was intended? It's still possible because the last upgrade hasn't synced. But I want to confirm this is intended. And anyone using testing will have to force the downgrade.

tfoote (changes requested):

noetic/distribution.yaml
BarisYazici: It is intended. The version number is the same. The build was failing on the ROS-build farm for debian buster, this fixes it. The changes were forced pushed to 1.2.4-1 tag. So it's the latest state of https://gitlab.com/toposens/public/toposens-libraries

hidmic: Not sure I follow the intent here. [1.2.4-4](https://github.com/ros/rosdistro/pull/33274) went in last week. Then [1.2.5-1](https://gitlab.com/toposens/public/toposens-library-release/-/tree/master#toposens-library-noetic-125-1) was bloom released, but not source released (or I can't find the corresponding tag). Then 1.2.4-1 was re-released (?). 

Did you intend to rollout 1.2.4-5 and something went wrong?

BarisYazici: 1.2.5-1 was accidental. I didn't do a MR to ros/rosdistro for that. Therefore, when I fixed the problem and re-released the 1.2.4 version, bloom-release command released for the with 1.2.4-1 build tag. So the version is still 1.2.4. The build tag was wrong due to a mistake. I then forced pushed the latest state to the 1.2.4-1 tag. This content is here : https://gitlab.com/toposens/public/toposens-library-release/-/tree/release/noetic/toposens-sensor-library/1.2.4-1 mirrors the latest state of https://gitlab.com/toposens/public/toposens-libraries.

tfoote: How did you do a force push into the 1.2.4-1? There are a lot of metadata to do that and a lot branches that need to be updated in parallel. It's much better just to rerun bloom and make a `-5` release cleanly. 

It also looks like you force pushed to your upstream? That's going to create a collision in the upstream tarball that is part of the debian packages as the debian repositories assume that tagged versions dont' change. It's possible for our buildfarm to invalidate and recover sometimes, but not always. If you miss something in a release it's best to just keep moving forward on the version numbers. Especially with 1.2.4-4 already on the buildfarm. You should be doing 1.2.5 if you're changing the upstream. This apparent downgrade will potentially break things in unexpected and undiagnosible ways.

BarisYazici: Okay, I understand. I didn't know that it can potentially break things on the ROS-build-farm. Can we do a 1.2.4-5 release instead of 1.2.5? I wanted to also note that none of the 1.2.4 version tags were synced to noetic. We recently did a 1.2.4 release for melodic distro and want to keep the 1.2.4 version tag.

BarisYazici: Thanks a lot for the detailed explanation, I indeed didn't know about the tar.gz file and its assumptions. I am not part of Topsens anymore, but I let the developers know that they need to do another release with 1.2.4-5 tag.

clalancette:
I'm going to close this PR out, since it is very likely a new bloom release is needed.  If this is in error, feel free to reopen.

PR #33327, opened 2022-05-22
Update release repository for mrpt2

jlblancoc:
Just like #32075, for foxy.


jlblancoc:
I'm having bloom-release errors trying to make a new release to `foxy`, even using `-e` or `--override-release-repository-url`, so *I think* it may be caused by rosdistro still pointing to the old repo (?).

jlblancoc:
Superseded by #33331  , correctly generated via bloom-release.

PR #33421, opened 2022-05-31
Add robotraconteur package source to humble/distribution.yaml

johnwason:

## Package name:

robotraconteur

## Package Upstream Source:

https://github.com/robotraconteur/robotraconteur/tree/ros2-humble

## Purpose of using this:

Robot Raconteur is a communication framework with an ecosystem containing a number of standardized industrial robot drivers, and a universal teach pendant under development.

https://github.com/robotraconteur/robotraconteur-directory
https://github.com/pyri-project/pyri-core/blob/master/README.md

Having the core `robotraconteur` package with C++ and Python bindings in rosdep will make integration with ROS easier.

This package was previously accepted to ROS Noetic in #31844

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3570,6 +3570,12 @@ repositories:
       url: https://github.com/ros/robot_state_publisher.git
       version: humble
     status: maintained
+  robotraconteur:
+    source:
+      type: git
+      url: https://github.com/robotraconteur/robotraconteur.git
+      version: ros2-humble
+    status: maintained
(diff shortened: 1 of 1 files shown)


clalancette (commented):
Looking at the `ros2-humble` branch on the repository, it looks like the package.xml still references `catkin` in it.  Is this package something that ROS 2 Humble users can actually use?

johnwason:
The package is pure cmake and doesn't use catkin. The inclusion was an oversight. I have removed it.

PR #33447, opened 2022-06-02
perception_open3d: 0.1.1-1 in 'rolling/distribution.yaml' [bloom]

SteveMacenski:
Increasing version of package(s) in repository `perception_open3d` to `0.1.1-1`:

- upstream repository: https://github.com/ros-perception/perception_open3d.git
- release repository: https://github.com/ros-gbp/perception_open3d-release.git
- bloom version: `0.11.1`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2514,6 +2514,19 @@ repositories:
       url: https://github.com/ros-perception/pcl_msgs.git
       version: ros2
     status: maintained
+  perception_open3d:
+    release:
+      packages:
+      - open3d_conversions
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros-gbp/perception_open3d-release.git
+      version: 0.1.1-1
+    source:
+      type: git
+      url: https://github.com/ros-perception/perception_open3d.git
+      version: ros2
+    status: developed
   perception_pcl:
     doc:
       type: git


SteveMacenski:
Blacklist for RHEL added due to lack of open3d key https://github.com/ros2/ros_buildfarm_config/pull/244

wep21:
As for versions, it may be better to make the rolling one greater than the humble one.

rolling/distribution.yaml
```diff
@@ -2514,6 +2514,19 @@ repositories:
       url: https://github.com/ros-perception/pcl_msgs.git
       version: ros2
     status: maintained
+  perception_open3d:
+    release:
+      packages:
+      - open3d_conversions
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros-gbp/perception_open3d-release.git
```
clalancette, line 2523: FYI, we highly recommend that you use a release repository in https://github.com/ros2-gbp for Rolling.  If you don't do this, then this package can't automatically be rolled over into the next distribution.

I won't outright block on this, but let me know explicitly if you want to stay with your current release repository.

clalancette (commented):
Besides the comment inline, I agree that it does feel a little weird to have Rolling behind Humble.  It's not wrong, per-se, but it is confusing.  Again, if you want to continue anyway, just let me know.

rolling/distribution.yaml
SteveMacenski: https://github.com/ros-perception/perception_open3d/issues/24 conversation is in the works, I need access to the org to do so. I'm a member of the org but not admin to move / create new repositories which I need to transfer this project (and probably several others in the future)

clalancette: While GitHub redirection will work, I'm not sure whether our tools will follow that properly.  @nuclearsandwich can answer that.  The easiest thing to do here is to move it over and just use the `ros2-gbp` repository, that way there is no question.  I'll see what I can do about getting https://github.com/ros-perception/perception_open3d/issues/24 resolved.

SteveMacenski: OK

PR #33460, opened 2022-06-03
navigation2: 1.1.0-1 in 'humble/distribution.yaml' [bloom]

SteveMacenski:
Increasing version of package(s) in repository `navigation2` to `1.1.0-1`:

- upstream repository: https://github.com/ros-planning/navigation2.git
- release repository: https://github.com/SteveMacenski/navigation2-release.git
- bloom version: `0.11.1`

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2226,6 +2226,53 @@ repositories:
       url: https://github.com/ijnek/nao_lola.git
       version: rolling
     status: developed
+  navigation2:
+    release:
+      packages:
+      - costmap_queue
+      - dwb_core
+      - dwb_critics
+      - dwb_msgs
+      - dwb_plugins
+      - nav2_amcl
+      - nav2_behavior_tree
+      - nav2_behaviors
+      - nav2_bringup
+      - nav2_bt_navigator
+      - nav2_common
+      - nav2_constrained_smoother
+      - nav2_controller
+      - nav2_core
+      - nav2_costmap_2d
+      - nav2_dwb_controller
+      - nav2_lifecycle_manager
+      - nav2_map_server
+      - nav2_msgs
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Please hold for slam_toolbox, they should be merged in tandem since they have interdependencies that are met by each other

SteveMacenski:
@clalancette this is failing and I'm not sure why. I ran bloom as I usually do without any issues. Release repo: https://github.com/ros2-gbp/navigation2-release

Every package is failing to build (ex: https://build.ros2.org/job/Hsrc_uJ__nav2_behaviors__ubuntu_jammy__source/1/console#console-section-9) to just clone the debian branch. When I look at a [Galactic eq. job](https://build.ros2.org/job/Gsrc_uF__nav2_recoveries__ubuntu_focal__source/17/console#console-section-9) I see it pulling tags of the same format but those tags also don't exist on the release repo for `Humble`. 

Is this a weird fluke that I should just re-run bloom on?

SteveMacenski:
Slam toolbox failed the same way

SteveMacenski:
I see them being generated in the bloom output running again. Maybe something odd was aborted when the keys were missing that I didn't pick up on - but due to the circular dependency on slam toolbox <> nav2 the first run fails until they're both in. I recall some issue like this last big distribution change requiring it to be released twice but I can't seem to find any notes about it and this doesn't feel familiar. Could also have been a failed push :shrug:. That computer is long since wiped even if there were logs. 

Either way, PR to come which hopefully handles the issue, but I'll check the tags. We'll probably end up needing to iterate on Nav2's release a couple of times since we introduce several new packages that I'm going to gamble probably have some isolated build dependency issues we didn't pick up on & difficult to test locally on isolation builds. Faster to just let the build farm tell us.

PR #33533, opened 2022-06-10
Revert pose_cov_ops to 0.2.1 in melodic

clalancette:
While it seems to work, it actually tries to install files to /opt/ros/melodic/.catkin, which then causes downstream packages to fail.  Revert back to 0.2.1, which is the last known working version.  @jlblancoc FYI

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -9125,7 +9125,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/mrpt-ros-pkg-release/pose_cov_ops-release.git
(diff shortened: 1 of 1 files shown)


jlblancoc:
Thanks, I'll take a look at it.

jlblancoc:
This was due to my "misbehavior" setting `<build_type>cmake</build_type>` in package.xml while trying to avoid what eventually was confirmed as a bug in catkin_pkg. Now that [it's fixed](https://github.com/ros-infrastructure/catkin_pkg/issues/336), I can fix `package.xml` and make a new release. Sorry for the noise!

PR #33717, opened 2022-06-27
add: python-multimethod

hsgwa:
Signed-off-by: hsgwa <hasegawa.isp@gmail.com>

## Package name:

multimethod

## Package Upstream Source:

https://github.com/coady/multimethod

## Purpose of using this:

multimethod provides a decorator for adding multiple argument dispatches to functions in the same style as functools.singledispatsh.

pip

https://pypi.org/project/multimethod/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2824,16 +2824,6 @@ python-multicast:
     pip: [py-multicast]
   ubuntu:
     pip: [py-multicast]
-python-multimethod:
-  debian:
-    pip:
-      packages: [multimethod]
-  fedora:
-    pip:
-      packages: [multimethod]
-  ubuntu:
-    pip:
-      packages: [multimethod]
 python-multiprocess-pip:
   arch:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2824,6 +2824,16 @@ python-multicast:
     pip: [py-multicast]
   ubuntu:
     pip: [py-multicast]
+python-multimethod:
```
ivanpauno, line 2827: You probably want to make it available for python3.
It's also a pip package, so the name should be:

```suggestion
python3-multimethod-pip:
```

ivanpauno (changes requested):
Please, read the contributing guidelines for python3 packages:

https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-rules

rosdep/python.yaml
hsgwa: Thank you for pointing out!
I modified in 8c71ccc.

ivanpauno (approved):

hidmic (approved):

PR #33726, opened 2022-06-28
Add aiortc

Interpause:

## Package name:

`aiortc`

## Package Upstream Source:

<https://github.com/aiortc/aiortc>

## Purpose of using this:

`aiortc` is a Python implementation of the WebRTC protocol, including support for DataChannels. Notably, it doesn't depend on the Chromium WebRTC project, hence avoiding issues such as <https://github.com/RobotWebTools/webrtc_ros/issues/66>. WebRTC is not only useful for sending video data to a browser from a robot, but also low latency realtime data via datachannels. Hence, I think it is valuable to add this to the index.

Distro packaging links:

- Debian:
  - <https://pypi.org/project/aiortc/> or <https://packages.debian.org/bookworm/python3-aiortc>
- Ubuntu:
   - <https://pypi.org/project/aiortc/> or <https://packages.ubuntu.com/jammy/python3-aiortc>
  - <https://pypi.org/project/aiortc/>
  - <https://pypi.org/project/aiortc/> or <https://aur.archlinux.org/packages/python-aiortc>
  - <https://pypi.org/project/aiortc/>
  - <https://pypi.org/project/aiortc/>
- Alpine:
  - <https://pypi.org/project/aiortc/>
- NixOS/nixpkgs:
  - <https://pypi.org/project/aiortc/>

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5880,7 +5880,7 @@ python3-aiohttp-cors:
   fedora: [python-aiohttp-cors]
   gentoo: [dev-python/aiohttp-cors]
   ubuntu: [python3-aiohttp-cors]
-python3-aiortc-pip:
+python3-aiortc:
   debian:
     bookworm: [aiortc]
   osx:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5882,12 +5882,12 @@ python3-aiohttp-cors:
   ubuntu: [python3-aiohttp-cors]
 python3-aiortc:
   debian:
-    bookworm: [aiortc]
+    bookworm: [python3-aiortc]
   osx:
     pip:
       packages: [aiortc]
   ubuntu:
-    jammy: [aiortc]
+    jammy: [python3-aiortc]
 python3-albumentations-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5882,20 +5882,12 @@ python3-aiohttp-cors:
   ubuntu: [python3-aiohttp-cors]
 python3-aiortc-pip:
   debian:
-    pip:
-      packages: [aiortc]
-  fedora:
-    pip:
-      packages: [aiortc]
-  gentoo:
-    pip:
-      packages: [aiortc]
+    bookworm: [aiortc]
   osx:
     pip:
       packages: [aiortc]
   ubuntu:
-    pip:
-      packages: [aiortc]
+    jammy: [aiortc]
 python3-albumentations-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5880,6 +5880,22 @@ python3-aiohttp-cors:
...
+      packages: [aiortc]
+  gentoo:
+    pip:
+      packages: [aiortc]
+  osx:
+    pip:
+      packages: [aiortc]
+  ubuntu:
+    pip:
+      packages: [aiortc]
```
hidmic, line 5898: @Interpause I see `python3-aiortc` available in several platform specific package managers (e.g. Debian, Ubuntu).

Consider adding a `python3-aiortc` key referring those wherever applicable.

adityapande-1995: We generally prefer system packages over pip : https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md

nuclearsandwich, line 5898:
```suggestion
python3-aiortc-pip:
  debian:
    bookworm: [aiortc]
  osx:
    pip:
      packages: [aiortc]
  ubuntu:
    jammy: [aiortc]
```

I would suggest not backfilling with pip packages. They can't be used in buildfarm-released packages so it's better to set up the definitons without pip allowing them to be used for future releases.

nuclearsandwich (dismissed):
Platforms with system packages available should use those. I propose not backfilling earlier versions of those platforms with pip since packages depending on those would not be release-able on the build farm.

rosdep/python.yaml
Interpause: Should it be renamed to `python3-aiortc` if so?

rosdep/python.yaml
```diff
@@ -5880,6 +5880,14 @@ python3-aiohttp-cors:
...
   gentoo: [dev-python/aiohttp-cors]
   ubuntu: [python3-aiohttp-cors]
+python3-aiortc-pip:
+  debian:
+    bookworm: [aiortc]
+  osx:
+    pip:
+      packages: [aiortc]
+  ubuntu:
+    jammy: [aiortc]
```
jacobperron, line 5890: `aiortc` -> `python3-aiortc`

rosdep/python.yaml
jacobperron: Yes, I think this should be renamed to `python3-aiortc`

rosdep/python.yaml
```diff
@@ -5880,6 +5880,14 @@ python3-aiohttp-cors:
   fedora: [python-aiohttp-cors]
   gentoo: [dev-python/aiohttp-cors]
   ubuntu: [python3-aiohttp-cors]
+python3-aiortc-pip:
+  debian:
+    bookworm: [aiortc]
```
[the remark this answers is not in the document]
Interpause: whoops, it is.

jacobperron (approved):
LGTM pending CI

tfoote (approved):

PR #33731, opened 2022-06-28
Add in retry logic for opening the pull request.

clalancette:
That's because we can run into secondary request limiting
with Github's API, so make sure to catch that case and retry
until it succeeds.


audrow (approved):
Looks good to me!

PR #33732, opened 2022-06-29
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    # only notify on security updates and not version updates.
+    open-pull-requests-limit: 0


hidmic (commented):
This makes sense to me, but I'd like someone more knowledgeable on the topic like @nuclearsandwich to chime in.

naveensrinivasan:
> @naveensrinivasan thanks for opening this PR as a way of suggesting this feature. I wasn't aware we could use dependabot to track GitHub Action dependencies.
> 
> Is it possible to configure dependabot to _only_ trigger notifications/automatic pull requests on security updates? I think that I'm only in favor of getting notifications if they can limited to security updates rather than general update notifications.
> 
> It's apt of dependabot to recognize that shared actions are yet another upgrade treadmill that we subject ourselves to when using them. But if I'm honest with myself, I prefer not to bother with infrastructure upgrades unless there's something I actually need in the updated version (like a security fix). Getting notifications just because there is a new version isn't something I am in favor of.
> 
> None of this is to say that actions authors aren't doing valuable work or contributing to the overall elevation of automation practices in public software, but I would much rather put resources into checking out new updates when I have some than being interrupted by a notification about it, which will likely get stale, unless there's some urgency (again, like a security fix) and I'd rather be able to treat such notifications as unambiguously important than have to evaluate each one.

I think we can if the config is set the number of PR's 0 AFAIK.

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
nuclearsandwich, line 6:
```suggestion
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    # only notify on security updates and not version updates.
    open-pull-requests-limit: 0
```

nuclearsandwich (dismissed):
Once pull requests are disabled so we're not nagged. I think this is a fine addition.

I think a positive review from at least one other ROS team member would be welcome.

naveensrinivasan:
> Once pull requests are disabled so we're not nagged. I think this is a fine addition.
> 
> I think a positive review from at least one other ROS team member would be welcome.

👍

cottsay (approved):

ivanpauno:
@naveensrinivasan a rebase seems to be needed to make the tests pass.

naveensrinivasan:
I just updated the branch.

ivanpauno (approved):

PR #33763, opened 2022-06-30
performance_test: 1.2.1-1 in 'humble/distribution.yaml' [bloom]

revati-naik:
Increasing version of package(s) in repository `performance_test` to `1.2.1-1`:

- upstream repository: https://gitlab.com/ApexAI/performance_test.git
- release repository: https://github.com/ros2-gbp/performance_test-release.git
- bloom version: `0.9.1`

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2730,13 +2730,10 @@ repositories:
       url: https://gitlab.com/ApexAI/performance_test.git
       version: 1.1.0
     release:
-      packages:
-      - performance_report
-      - performance_test
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/ros2-gbp/performance_test-release.git
-      version: 1.1.0-1
+      version: 1.2.1-1
     source:
       type: git
       url: https://gitlab.com/ApexAI/performance_test.git


humble/distribution.yaml
```diff
@@ -2730,13 +2730,10 @@ repositories:
       url: https://gitlab.com/ApexAI/performance_test.git
       version: 1.1.0
     release:
-      packages:
-      - performance_report
```
hidmic, line 2734: @revati-naik IIUC this means `performance_report` will no longer be released. Is that intended?

humble/distribution.yaml
```diff
@@ -2730,13 +2730,10 @@ repositories:
       url: https://gitlab.com/ApexAI/performance_test.git
       version: 1.1.0
```
tfoote, line 2731: This tag is out of date now. Is there a branch we can track instead? @revati-naik

tfoote (changes requested):
@hidmic If you're raising a concern that should block release, please use the request changes so it is known to need attention.

nuclearsandwich (changes requested):
Once https://github.com/ros/rosdistro/pull/33763/files#r916281228 is resolved this release is ✅

tfoote (approved):
`performance_report` has never built so this isn't a regression, and will just safe time on the farm. I'm going to merge as such. 

@revati-naik Please update the docs as appropriate. I'm not going to consider that a blocker, but they should really be using a branch not the tag which will get out of date.

PR #33783, opened 2022-07-05
Add python3-river-pip

amasciotta:

## Package name:

`River`

## Package Upstream Source:

[River](https://riverml.xyz/0.11.1/)

## Purpose of using this:

River is a library to build online machine learning models. Such models operate on data streams.

- Debian: https://pypi.org/project/river/
- Ubuntu: https://pypi.org/project/river/
- macOS: https://pypi.org/project/river/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8296,6 +8296,13 @@ python3-retrying:
     pip:
       packages: [retrying]
   ubuntu: [python3-retrying]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
It looks to be a relatively small package so pip only makes sense.

nuclearsandwich (approved):

PR #33811, opened 2022-07-10
add opensuse to pr template

ijnek:
For consistency, openSUSE should be added to the PR template.

Nosetests warns if an opensuse package is available but not listed, such as:

> warning file=rosdep/base.yaml,line=3187::Key 'libfreetype-dev' might be satisifed by opensuse package named 'freetype-devel': https://software.opensuse.org/package/freetype

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -19,7 +19,7 @@ TODO link to source repository
 
 ## Purpose of using this:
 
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Thanks adding this makes sense to me.

clalancette (approved):

PR #33824, opened 2022-07-12
Update libjsoncpp version for Ubuntu Jammy in rosdep/base.yaml

ksuszka:
Please fix package version for Ubuntu Jammy. 

## Package name:

libjsoncpp

## Purpose of using this:

In Ubuntu Jammy libjsoncpp library package name was changed from libjsoncpp1 to libjsoncpp25. Incorrect package version affects ROS Humble.

- Ubuntu: https://packages.ubuntu.com/jammy/libjsoncpp25

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3810,13 +3810,11 @@ libjsoncpp:
   opensuse: [libjsoncpp19]
   rhel: [jsoncpp]
   ubuntu:
-    '*': [libjsoncpp1]
-    precise: [libjsoncpp0]
-    saucy: [libjsoncpp0]
+    '*': [libjsoncpp25]
+    bionic: [libjsoncpp1]
+    focal: [libjsoncpp1]
     trusty: [libjsoncpp0]
-    utopic: [libjsoncpp0]
-    vivid: [libjsoncpp0]
-    wily: [libjsoncpp0v5]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -3811,6 +3811,7 @@ libjsoncpp:
   rhel: [jsoncpp]
   ubuntu:
     '*': [libjsoncpp1]
```
tfoote, line 3813: The wildcard rule should be used for the latest value and expected potential future values. (aka jammy and kinetic are 25 and the best chance of a future version being the same are that value.) 

So please update libjsoncpp25 to be the wildcard and fill in all the older supported versions, which looks like libjsoncpp1 for bionic and focal. Older platforms can be pruned as they are EOL.

tfoote (changes requested):

rosdep/base.yaml
ksuszka: I left all "current" Ubuntu LTS distros (so xenial/trusty/bionic/focal as it seems that xenial/trusty are still in extended support mode) and updated wildcard for jammy/future versions.

jacobperron (approved):

tfoote (approved):
Thanks for the update. It lgtm

PR #33980, opened 2022-07-26
Change ros2_tracing repo URL to GitHub

christophebedard:
Also, change Rolling branch to `rolling` and enable PR tests, since the repo is now on GitHub.

See: https://gitlab.com/ros-tracing/ros2_tracing/-/issues/144

The release repo was also updated: https://github.com/ros2-gbp/ros2_tracing-release/pull/4

Signed-off-by: Christophe Bedard <christophe.bedard@apex.ai>

### DIFF
--- a/dashing/distribution.yaml
+++ b/dashing/distribution.yaml
@@ -2718,7 +2718,7 @@ repositories:
   ros2_tracing:
     doc:
       type: git
-      url: https://github.com/ros2/ros2_tracing.git
+      url: https://gitlab.com/ros-tracing/ros2_tracing.git
       version: dashing
     release:
       packages:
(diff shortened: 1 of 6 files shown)


christophebedard:
Should I revert the changes for Dashing and Eloquent?

clalancette:
> Should I revert the changes for Dashing and Eloquent?

Yeah, let's do that in this PR.  That will make the tests happy, and just make it easy to get this in.

We could consider doing a separate PR where we update Eloquent and Dashing, and ignore the test failures.  But I'm not sure it is worth it; the GitLab repository is still there as a mirror, and on the whole we prefer not to change older distributions.

christophebedard:
Sounds good. I've reverted the Dashing and Eloquent changes. We can just leave them as-is and not update them then.

clalancette (approved):

PR #33981, opened 2022-07-26
qt_advanced_docking_system: 3.8.2-1 in 'foxy/distribution.yaml' [bloom]

Levi-Armstrong:
Increasing version of package(s) in repository `qt_advanced_docking_system` to `3.8.2-1`:

- upstream repository: https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System
- release repository: https://github.com/tesseract-robotics-release/qt_advanced_docking_system-release.git
- bloom version: `0.11.2`

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -3828,6 +3828,15 @@ repositories:
       url: https://github.com/Autoware-AI/qpoases_vendor.git
       version: ros2
     status: maintained
+  qt_advanced_docking_system:
+    release:
+      packages:
+      - qt_advanced_docking
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/tesseract-robotics-release/qt_advanced_docking_system-release.git
+      version: 3.8.2-1
+    status: developed
   qt_gui_core:
     doc:
       type: git
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -3829,6 +3829,10 @@ repositories:
       version: ros2
     status: maintained
   qt_advanced_docking_system:
+    doc:
+      type: git
+      url: https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System.git
+      version: master
     release:
       packages:
       - qt_advanced_docking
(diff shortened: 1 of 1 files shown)


clalancette (dismissed):
There are a few problems in here:

1. Can you please add a `source` section?
2. It's kind of weird that there is only a single package, but this entry has a `packages` section.  For single package entries, they generally just use the name of the key.  See for example https://github.com/ros/rosdistro/blob/2d30f04e894a5e1c7659fa7887bf2f9a7b7326c5/foxy/distribution.yaml#L284-L298 (there are many more examples).
3. The release repository should ideally be in `ros2-gbp`, though this is not strictly required for non-Rolling releases.
4. This is adding in some infrastructure, not a robotics or ROS package.  This is allowed, but can you give us a better idea of what you want to use this for?  We don't want to be "vendoring" every package in existence just to have it.

Levi-Armstrong:
@clalancette What are the next steps?

sloretz:
> >  It's kind of weird that there is only a single package, but this entry has a packages section. For single package entries, they generally just use the name of the key. See for example
> 
> I assume this is because the name of the release repository is different from the package? This is currently how it is on the noetic release. Should it be change now?

This is correct. There's no need to change it, while unusual [there are others](https://github.com/ros/rosdistro/blob/d4c5d17aad55e48b00136fed7797fb5c9c1c4f64/foxy/distribution.yaml#L6738-L6755).

> I would be happy to release this somewhere else but as I understand it, either needs to be released on Ubuntu or here to be able to use it within a ros package

I didn't find an existing Debian or Ubuntu package for it, so vendoring it in the ROS distro should be ok.

> Right now we plan to release this on foxy and Galactic so I guess it should not be problem for now. When it comes time to do a Rolling release am I able to transfer the current release to that organization?

Yes, the release repo can be moved later. The advantage of having a `ros2-gbp` release repo is it will automatically roll into the next ROS distribution. @clalancette do you have a link to the instructions for creating a `ros2-gbp` repo and adding people to it?

sloretz (approved):
LGTM, but I'll wait for @clalancette to have a second look at the comments.

Levi-Armstrong:
@clalancette friendly ping :)

Levi-Armstrong:
@sloretz and @clalancette friendly ping :)

tfoote (approved):

Levi-Armstrong:
Thank you!

PR #19652, opened 2018-12-08
add python-gi-cairo

k-okada:
deb: https://pkgs.org/download/python-gi-cairo
fedra : https://apps.fedoraproject.org/packages/python2-gobject/contents/
gentoo: https://packages.gentoo.org/packages/dev-python/pygobject

PR #33996, opened 2022-07-28
add python-gobject

k-okada:

## Package name:

python-gobject

## Purpose of using this:

This is almsot as same as https://github.com/ros/rosdistro/pull/19652, but we want to install `python-gobject` for debian/ubuntu.

https://pkgs.org/download/python-gobject

ubuntu: https://packages.ubuntu.com/bionic/python-gobject
debian: https://packages.debian.org/buster/python-gobject

fedra : https://apps.fedoraproject.org/packages/python2-gobject/contents/
gentoo: https://packages.gentoo.org/packages/dev-python/pygobject

# Checks
 - [x ] This package is expected to build on the submitted rosdistro

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2019,10 +2019,17 @@ python-gnupg:
   ubuntu: [python-gnupg]
 python-gobject:
   arch: [python2-gobject]
-  debian: [python-gobject]
+  debian:
+    '*': null
+    stretch: [python-gobject]
+    buster: [python-gobject]
   gentoo: [dev-python/pygobject]
   nixos: [pythonPackages.pygobject3]
-  ubuntu: [python-gobject]
+  opensuse: [python2-gobject]
+  ubuntu:
+    '*': null
+    bionic: [python-gobject]
+    focal: [python-gobject]
 python-google-cloud-bigquery-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2023,9 +2023,8 @@ python-gobject:
     '*': null
     buster: [python-gobject]
     stretch: [python-gobject]
-  fedora: [python3-gobject]
   gentoo: [dev-python/pygobject]
-  nixos: [pythonPackages.pygobject3]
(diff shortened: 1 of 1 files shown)


sloretz:
CI reports a few issues. For a Python 2 only key, maybe this would work?

* nix: https://github.com/NixOS/nixpkgs/blob/9d452bdebedb1c47adaabd40e61a742f9f07b54c/pkgs/top-level/python-packages.nix#L7546
* opensuse: http://download.opensuse.org/distribution/leap/15.2/repo/oss/x86_64/python2-gobject-3.34.0-lp152.2.3.x86_64.rpm
* Ubuntu Focal: https://packages.ubuntu.com/focal/python-gobject

```yaml
python-gobject:
  nixos: [pythonPackages.pygobject2]
  opensuse: [python2-gobject]
  ubuntu:
    '*': null
    focal: [python-gobject]
```

k-okada:
@sloretz thanks, fixed CI

rosdep/python.yaml
```diff
@@ -2017,6 +2017,20 @@ python-gnupg:
   rhel:
     '7': [python2-gnupg]
   ubuntu: [python-gnupg]
+python-gobject:
+  arch: [python2-gobject]
+  debian:
+    '*': null
+    buster: [python-gobject]
+    stretch: [python-gobject]
+  fedora: [python3-gobject]
```
ivanpauno, line 2026: isn't this the python3 package?
It looks like you want to add the python2 version here, right?

rosdep/python.yaml
```diff
@@ -2017,6 +2017,20 @@ python-gnupg:
...
+  gentoo: [dev-python/pygobject]
+  nixos: [pythonPackages.pygobject3]
```
ivanpauno, line 2028: same here

rosdep/python.yaml
k-okada: @ivanpauno yes, you're right. removed fedra key because it does not provide python2 binding and fix nixos .

ivanpauno (approved):
LGTM!

PR #34044, opened 2022-08-02
foxy: Update eigenpy for pythonpath patch

wxmerkt:


foxy/distribution.yaml
```diff
@@ -1217,7 +1217,7 @@ repositories:
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/stack-of-tasks/eigenpy-ros-release.git
-      version: 2.7.10-1
+      version: 2.7.10-2
```
jacobperron, line 1220: This is just bumping the build number, was this intended?

wxmerkt: Yes, I released the same version but now with an added patch in the release repository (we previously installed the Python packages into `lib/python3/dist-packages` as on ROS1 and despite the dsv hook it wasn't added to the path; now it will install to `lib/python$major.$minor/site-packages` as for ROS2)

PR #34051, opened 2022-08-02
rolling: Update eigenpy for standard Python layout

wxmerkt:


rolling/distribution.yaml
```diff
@@ -924,7 +924,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/eigenpy-release.git
-      version: 2.7.10-2
+      version: 2.7.10-3
```
sloretz, line 927: I see the tag [`debian/ros-rolling-eigenpy_2.7.10-3_jammy`](https://github.com/ros2-gbp/eigenpy-release/tree/debian/ros-rolling-eigenpy_2.7.10-3_jammy), but I don't see an entry from bloom [in the README.md](https://github.com/ros2-gbp/eigenpy-release/tree/master). How did you generate the release?

wxmerkt: `git-bloom-release` to add / export a patch, i.e. I didn't run `bloom-release -r $distro` but instead in the release repository `git-bloom-release $distro`, then pushed the branches and tags `git push --all && git push --tags`. The new release is to create a different patch to add an additional option/variable to the CMakeLists. The background is that the Python code was installed into `lib/python3/dist-packages` as in ROS1 and for it to be installed to `lib/python$major.$minor/site-packages` we need to add an extra option to the CMakeLists

Once this one and #34058 are merged and pass, I will manually patch the same packages in all other distros. Starting from the next minor release it will be back to `bloom-release -r $distro` for all of them

wxmerkt: @sloretz Friendly ping - are this PR and #34058 good to go?

methylDragon: Mmm.. makes sense, I'll approve for now. You might want to update the README to reflect this for posterity

methylDragon (approved):

PR #34058, opened 2022-08-03
rolling: Update hpp-fcl to include standard Python layout

wxmerkt:


rolling/distribution.yaml
```diff
@@ -1404,7 +1404,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/hpp_fcl-release.git
-      version: 2.1.2-2
+      version: 2.1.2-3
```
sloretz, line 1407: I have the same question here https://github.com/ros/rosdistro/pull/34051#pullrequestreview-1062687309

methylDragon (approved):

PR #34093, opened 2022-08-09
Added statistics and sbp modules

jacksongeritz:

## Package name:

sbp
statistics

## Package Upstream Source:

sbp: https://github.com/swift-nav/libsbp
statistics: https://github.com/digitalemagine/py-statistics

## Purpose of using this:

sbp is used to integrate with the swift binary protocol. This is necessary to integrate with Swift GPS units for navigation/autonomy.
statistics is a python2 port of the extremely useful python3 statistics module

  - Not Available
   - Not Available
- Pip:
  - https://pypi.org/project/sbp/
  - https://pypi.org/project/statistics/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4481,6 +4481,13 @@ python-rx-pip:
   ubuntu:
     pip:
       packages: [rx]
+python-sbp-pip:
+  debian:
+    pip:
+      packages: [sbp]
+  ubuntu:
+    pip:
+      packages: [sbp]
 python-scapy:
   debian: [python-scapy]
   fedora: [scapy]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -4481,6 +4481,13 @@ python-rx-pip:
   ubuntu:
     pip:
       packages: [rx]
+python-sbp-pip:
```
ivanpauno, line 4484: do you want to make it available for python3?
Please use `python3-sbp-pip` in that case.
Remember to alpha-order again

rosdep/python.yaml
```diff
@@ -4909,6 +4916,13 @@ python-sqlalchemy:
 python-sqlite:
   debian: [python-sqlite]
   ubuntu: [python-sqlite]
+python-statistics-pip:
```
ivanpauno, line 4919: same

rosdep/python.yaml
jacksongeritz: Just added an alias for python3 to refer to the python2 version

rosdep/python.yaml
jacksongeritz: This is basically a python2 port of the default python 3 statistics module. Unless you disagree, I don't really see any reason to make it available for python3?

ivanpauno (approved):

methylDragon (approved):

PR #34200, opened 2022-08-23
Add sparkfun_ublox_gps

willcbaker:

## Package name:

sparkfun-ublox-gps

## Package Upstream Source:

https://pypi.org/project/sparkfun-ublox-gps/

## Purpose of using this:

Useful for interacting with GPS products based on u-blox GPS modules. See https://github.com/sparkfun/Qwiic_Ublox_Gps_Py

python3 pip installable https://pypi.org/project/sparkfun-ublox-gps/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8835,6 +8835,19 @@ python3-sounddevice-pip:
   ubuntu:
     pip:
       packages: [sounddevice]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This doesn't look like something that will have system packages.

nuclearsandwich (approved):

PR #34202, opened 2022-08-24
Add Change clpe_sdk version

rorosi:
I failed by bloom because didn't update git token sorry

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -669,7 +669,7 @@ repositories:
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/canlab-co/CLPE_G_NVP2650D_SDK-ros2-release.git
-      version: 0.1.0-6
+      version: 0.1.0-7
     source:
       test_pull_requests: true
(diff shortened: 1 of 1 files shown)


tfoote:
This looks like you didn't make a release you were expecting to do in the same way as #34203

I see a changelog for 0.1.7 but this is rereleasing 0.1.0

https://github.com/canlab-co/CLPE_G_NVP2650D_SDK/blob/main/CHANGELOG.rst

You need to make sure to do the release, incrementing the package.xml version and tagging the repository. 
There's a helper script to do that called `catkin_prepare_release` or you can do it manually. 


And in the future if you have problems with opening the pull request. You can ask bloom to retry just the pull request using `--pull-request-only` 
```
  --pull-request-only, -p
                        skips the release actions and only tries to open a
                        pull request

```

tfoote:
It looks like this isn't going to do what you want. You need to tag a new version of the repository following section 2 of http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

PR #34203, opened 2022-08-24
clpe_ros: 0.1.0-3 in 'foxy/distribution.yaml' [bloom]

rorosi:
Increasing version of package(s) in repository `clpe_ros` to `0.1.0-3`:

- upstream repository: https://github.com/canlab-co/clpe_ros.git
- release repository: https://github.com/canlab-co/clpe_ros-ros2-release.git
- bloom version: `0.11.1`
- previous version for package: `0.1.0-2`

## clpe_ros

```
* Provides ROS 2 driver for CANLAB CLPE-G-NVP2650D vision system
* Support ROS 2 Foxy and ROS 2 Galactic
* By default the driver will publish three topics for each camera X.
  * /clpe/cam_X/image_raw: The raw image published as sensor_msgs::msg::Image. The default encoding is yuv422. For other supported encodings, see Configuration.
  * /clpe/cam_X/camera_info: The intrinsics of the camera published as a sensor_msgs::msg::CameraInfo message.
  * /clpe/cam_X/clpe_camera_info: The intrinsics of the camera published as a clpe_ros_msgs::msg::ClpeCameraInfo message.
* Contributors: Teo Koon Peng, Yadunund Vijay
```

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -635,7 +635,7 @@ repositories:
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/canlab-co/clpe_ros-ros2-release.git
-      version: 0.1.0-2
+      version: 0.1.0-3
     source:
       test_pull_requests: true
       type: git


tfoote:
I glanced at the changelog and I see what looks like 0.1.3 there. Did you mean to release that instead of rereleasing 0.1.0?

When looking around I also noticed that you appear to be using the same versions on different branches. 0.1.0 on noetic and 0.1.0 on ROS 2. This is generally a bad idea due to potential confusion. I strongly recommend using different number series between your releases so that you don't have potential confusion when someone says they're working with clpe_ros version 0.1.0 and it's not working on their system and you can't differentiate.

tfoote:
This pull request is trying to rerelease the same version you already released from the same repository. If you want to change the source make sure to tell bloom to edit the information. 

Also before you run bloom you have to actually make a new release with a new version number. See section 2 here: http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

PR #34205, opened 2022-08-24
add python3-breathe dep

giusebar:

## Package name:

python3-breathe

## Package Upstream Source:

[Github Link](https://github.com/michaeljones/breathe)

## Purpose of using this:

Breathe is an extension to reStructuredText and Sphinx to be able to read and
render [Doxygen](http://www.doxygen.org/) xml output. Currently used by ROS2 packages, such as [cyclonedds](https://github.com/eclipse-cyclonedds/cyclonedds). 

Distro packaging links:

- Debian: https://packages.debian.org/stretch/python3-breathe
- Ubuntu: https://packages.ubuntu.com/bionic/python3-breathe
- Fedora: https://packages.fedoraproject.org/pkgs/python-breathe/python3-breathe/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -840,10 +840,6 @@ python-breathe:
   debian: [python-breathe]
   fedora: [python-breathe]
   ubuntu: [python-breathe]
-python3-breathe:
-  debian: [python3-breathe]
-  fedora: [python3-breathe]
-  ubuntu: [python3-breathe]
 python-bs4:
   debian: [python-bs4]
   fedora: [python-beautifulsoup4]
@@ -6108,6 +6104,10 @@ python3-boto3:
     '*': ['python%{python3_pkgversion}-boto3']
     '7': null
   ubuntu: [python3-boto3]
+python3-breathe:
+  debian: [python3-breathe]
+  fedora: [python3-breathe]
+  ubuntu: [python3-breathe]
 python3-bs4:
   arch: [python-beautifulsoup4]
   debian: [python3-bs4]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6104,6 +6104,10 @@ python3-boto3:
     '*': ['python%{python3_pkgversion}-boto3']
     '7': null
   ubuntu: [python3-boto3]
+python3-breathe:
+  debian: [python3-breathe]
+  fedora: [python3-breathe]
```
tfoote, line 6109:
```suggestion
  fedora: [python3-breathe]
  opensuse: [python3-breathe]
  rhel: [python3-breathe]
```

Add rhel and opensuse suggested by linter

tfoote (changes requested):
One requested change to fill it out other platforms. Otherwise it looks good.

rosdep/python.yaml
```diff
@@ -6104,6 +6104,12 @@ python3-boto3:
...
+  opensuse: [python3-breathe]
+  rhel: [python3-breathe]
```
nuclearsandwich, line 6111: This package isn't available in RHEL 7 so that has to be nulled out. I also updated this to use the python3_pkgversion macro for the rhel key which I believe helps us in the situation that the default version of python3 for a RHEL release isn't the version of python3 we're using for ROS on that platform but @cottsay should double check before I make a PR to the contributing and review guidelines.

```suggestion
  rhel:
    '*': ['python%{python3_pkgversion}-breathe']
    '7': null
```

nuclearsandwich (changes requested):
Thanks for iterating. The package checker found that older versions of RHEL are missing this package and so a further change is required.

nuclearsandwich (approved):

tfoote (approved):
Thanks for iterating. Lgtm

PR #34243, opened 2022-08-29
Add python3 pyside2 key

hellantos:

## Package name:

python3-pyside2

## Package Upstream Source:

https://code.qt.io/cgit/pyside/pyside-setup.git/
https://code.qt.io/cgit/pyside/pyside.git/
https://code.qt.io/cgit/pyside/shiboken.git/

## Purpose of using this:

The goal of this key is to eventually replace the pyside2 specific portion of python3-qt5-bindings.
This key will install qt python libraries wrapped with PySide2. The key installs the same libraries as python3-pyqt5 see #34246. This is thought to simplify switching between pyqt and pyside2 as python wrapper providers (see https://github.com/ros-visualization/python_qt_binding/issues/114)

It will at least install wrappers for the following qt libraries:
* qtcore
* qtgui
* qthelp 
* qtnetwork
* qtprintsupport
* qtsvg
* qttest
* qtwidgets
* qtxml

In addition it will install:
* shiboken2 and libraries (generator for pyside2 wrappers)

On Fedora the pyside package is not split so this will install all libraries.

Distro packaging links:

  - https://packages.debian.org/search?searchon=names&keywords=python3-pyside2
   - https://packages.ubuntu.com/search?keywords=python3-pyside2&searchon=names
  - https://packages.fedoraproject.org/pkgs/python-pyside2/python3-pyside2/fedora-rawhide.html
  - https://archlinux.org/packages/extra/x86_64/pyside2/
  - https://packages.gentoo.org/packages/dev-python/pyside2

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8186,10 +8186,13 @@ python3-pyrr-pip:
     pip:
       packages: [pyrr]
 python3-pyside2:
-  debian: [libpyside2-dev, libshiboken2-dev, python3-pyside2.qtcore, python3-pyside2.qtgui, python3-pyside2.qthelp, python3-pyside2.qtnetwork, python3-pyside2.qtprintsupport, python3-pyside2.qttest, python3-pyside2.qtwidgets, python3-pyside2.qtxml, shiboken2]
+  arch: [pyside2, python-shiboken2, shiboken2]
+  debian: [libpyside2-dev, libshiboken2-dev, python3-pyside2.qtcore, python3-pyside2.qtgui, python3-pyside2.qthelp, python3-pyside2.qtnetwork, python3-pyside2.qtprintsupport, python3-pyside2.qttest, python3-pyside2.qtsvg, python3-pyside2.qtwidgets, python3-pyside2.qtxml, shiboken2]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@ipa-cmh is there a reason that this PR is still in draft? Would you like initial review on that?

ipa-cmh:
@nuclearsandwich
Its part of a bigger restrucutring around qt python bindings - see linked  issues. But I think I am ready for review now.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "ros.distro": "humble"
+}
```
jacobperron, line 3: Seems like this file was added accidentally. Please remove.

jacobperron (commented):
Could you also add an entry for opensuse, as suggested by the GitHub Action.

ipa-cmh:
Will do.

.vscode/settings.json
ipa-cmh: Removed

ipa-cmh:
> Could you also add an entry for opensuse, as suggested by the GitHub Action.

Opensuse seems to only distribute shiboken6 and not shiboken2. I did not find a package, that distributes the necessary shiboken binaries. So I would rather not add the key at all for opensuse.

jacobperron (approved):

tfoote (approved):

PR #34246, opened 2022-08-29
Add python3 pyqt5 package

hellantos:

## Package name:

python3-pyqt5

## Package Upstream Source:

https://riverbankcomputing.com/software/pyqt/download

## Purpose of using this:

The goal of this key is to eventually replace python3-qt5-bindings.
This key will install qt python libraries wrapped with PyQt (SIP). The key installs the same libraries as python3-pyside2 see #34243. This is thought to simplify switching between pyqt and pyside2 as python wrapper providers (see https://github.com/ros-visualization/python_qt_binding/issues/114)

It will at least install wrappers for the following qt libraries:

* qtcore
* qtgui
* qthelp
* qtnetwork
* qtprintsupport
* qtsvg
* qttest
* qtwidgets
* qtxml

In addition it will install:

* sip (generator for pyside2 wrappers)
* development libraries

  - https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=python3-pyqt5
  - https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=python3-sip
  - https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=pyqt5-dev
   - https://packages.ubuntu.com/search?keywords=python3-pyqt5&searchon=names
   - https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=python3-sip&searchon=names
   - https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=pyqt5-dev&searchon=names

Keys for other distributions are copied over from python3-qt5-bindings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 *~
 .DS_STORE
 *.pyc
-.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *~
 .DS_STORE
 *.pyc
+.vscode/
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8159,6 +8159,11 @@ python3-pyqrcode:
   gentoo: [dev-python/pyqrcode]
   nixos: [python3Packages.pyqrcode]
   ubuntu: [python3-pyqrcode]
+python3-pyqt5:
+  ubuntu:
+    '*': [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev]
+    artful: [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev]
+    bionic: [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev]
 python3-pyqt5.qtquick:
   debian: [python3-pyqt5.qtquick]
   ubuntu: [python3-pyqt5.qtquick]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8164,8 +8164,6 @@ python3-pyqt5:
   arch: [python-pyqt5]
   debian:
     '*': [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev, qtbase5-dev]
-    jessie: [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev, qtbase5-dev]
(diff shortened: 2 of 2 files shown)


rosdep/python.yaml
```diff
@@ -8159,6 +8159,24 @@ python3-pyqrcode:
   gentoo: [dev-python/pyqrcode]
   nixos: [python3Packages.pyqrcode]
   ubuntu: [python3-pyqrcode]
+python3-pyqt5:
+  alpine: [py3-qt5]
+  arch: [python-pyqt5]
+  debian:
+    '*': [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev, qtbase5-dev]
```
tfoote, line 8166: Why are you using the wildcard notation and calling out jessie and stretch? They appear to be the same keys.

rosdep/python.yaml
```diff
@@ -8159,6 +8159,24 @@ python3-pyqrcode:
...
+    stretch: [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev, qtbase5-dev]
+  fedora: [python3-qt5-devel, python3-sip-devel, libXext-devel]
+  gentoo: [dev-python/PyQt5]
+  nixos: [python3Packages.pyqt5]
+  openembedded: [python3-pyqt5@meta-qt5]
+  opensuse: [python3-qt5]
+  rhel: ['python%{python3_pkgversion}-qt5-devel', 'python%{python3_pkgversion}-sip-devel', libXext-devel, redhat-rpm-config]
+  slackware: [python3-PyQt5]
+  ubuntu:
+    '*': [pyqt5-dev, python3-pyqt5, python3-pyqt5.qtsvg, python3-sip-dev]
```
tfoote, line 8177: Similarly why the wildcard here with artful and bionic called out separately with what looks like the same keys as well.

rosdep/python.yaml
ipa-cmh: This was a copy paste artifact. it is now removed.

rosdep/python.yaml
ipa-cmh: This was a copy paste artifact. it is now removed.

.gitignore
```diff
@@ -1,3 +1,4 @@
 *~
 .DS_STORE
 *.pyc
+.vscode/
```
tfoote, line 4:
```suggestion
```
This looks like an unrelated change that slipped into the last commit.

clalancette (changes requested):
Needs update from the original contributor.

jacobperron (approved):

ivanpauno (approved):

PR #34253, opened 2022-08-29
add colcon-meson

christianrauch:

## Package name:

~~python3-colcon-meson-pip~~

python3-colcon-meson

## Package Upstream Source:

~~https://pypi.org/project/colcon-meson/~~

- https://github.com/colcon/colcon-meson
- https://github.com/ros-infrastructure/reprepro-updater/blob/master/config/colcon.debian.upstream.yaml

## Purpose of using this:

This is an extension for colcon to build plain meson packages with/without `package.xml`.

# The source is here:

https://github.com/colcon/colcon-meson

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6249,6 +6249,9 @@ python3-colcon-common-extensions:
   fedora: [python3-colcon-common-extensions]
   rhel: ['python%{python3_pkgversion}-colcon-common-extensions']
   ubuntu: [python3-colcon-common-extensions]
+python3-colcon-meson:
+  ubuntu:
+    jammy: [python3-colcon-meson]
 python3-collada:
   debian:
     '*': [python3-collada]


tfoote:
Is this not being released as a binary package?

christianrauch:
> Is this not being released as a binary package?

This is just python scripts. There is no cython or similar that needs to be compiled.

tfoote:
When I mean binary packages I mean debian/apt based packages versus the pip based installation.

christianrauch:
I release this only as pip package. There is no bloom release (yet).

There was [some talk about releasing `colcon-meson`](https://github.com/colcon/colcon-common-extensions/issues/17#issuecomment-1154454158) via bloom. @nuclearsandwich What is required to get the colcon-meson package released via bloom? @tfoote Is a bloom release a requirement to get this into rosdep? Eventually, I would like to have colcon-meson automatically installed via rosdep for all meson packages.

tfoote:
Our general policy is that if there is currently or is expected to be a natively packaged version we should always use that one. If we use a pip version on any platform it will override the natively packaged one and there's no way to roll back automatically without user interaction. 

So if deb packages are a goal (and that will enable rosdep and dependencies etc) we should not create a rule for pip packages and wait for the moment until we have the debian pipeline figured out. Note that for non-ROS packages like this it won't go through bloom. There's separate processes for colcon packages.

nuclearsandwich:
colcon packages are generally not released with bloom, my apologies that I gave that impression. Since the colcon project, while built and maintained with ROS in mind and by some members of the ROS team, is a separate project and colcon packages generally release to pypi and the colcon repository on packagecloud.

It is also uncommon for any one package to depend on colcon at all. With the possible recent exception of a devtools metapackage for ROS no one package in the ROS ecosystem depends directly on colcon, (where the name comes from "<em>col</em>lective <em>con</em>struction") is about building workspaces in aggregate and solving the challenges that arise.

There isn't really a formal guide on adding a new independent build system to ROS but I've often tried to think about the general steps.

For basic packages that aren't tightly integrated with ROS, they should be able to use the target build system (in your case, and the example I'll use for the rest of this talkthrough, meson) by adding a build_depend on the package and declaring the build system name as the build system.

Here's an incomplete example
```xml
<package format="3">
  <name>meson_example<</name>
  
  <build_depend>meson</build_depend>

  <export>
    <build_type>meson</build_type>
  </export>
</package>
```

On its own this doesn't help much except to indicate to package consumers what build system is expected and what the dependencies are. But you could build this package standalone just using `meson` as you normally would. To build the package from source in a workspace with other packages, you would need to extend colcon, which is where colcon-meson comes in. That allows meson packages to be built in workspaces alongside other cmake, python, ament_cmake, or catkin packages in a workspace. Whether those packages can interdepend with meson packages will depend on how meson interacts with the other build systems, primarily whether it can expose and/or consume package metadata in a format understood by cmake or python based build systems.

If not that's one possible use case for an ament_meson style library that can augment meson builds with information from the [ament resource index](https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/doc/resource_index.md).

Whether it makes sense to create a dedicated `ament_meson` build type in addition to an ament_meson support package will depend on some thoughts that I haven't finished formming and technical vagaries but either way, if the support package exists it could be used to fill in gaps in meson's base abilities to interoperate with other build systems in the ROS ecosystem.

Lastly, in order to be built into binary packages, support for meson based packages will need to be added to bloom. Binary packages are built individually and thus don't use colcon or any of its modules but rather use a set of templates to create platform-specific metadata for a set of ROS packages in a repository. Each build type will have a set of templates for it that integrate with the platform packaging system.

That's everything I can think of off the top of my head. Work on colcon-meson is the right first step since it allows building meson packages in workspace, but it wouldn't be a direct dependency of any ROS package, rather, something that a developer would install in addition to any other colcon packages that support their workflows.

christianrauch:
@nuclearsandwich Thanks for this detailed explanation. How can I get colcon-meson released as Debian package, so it becomes part of the ROS repo and can be used for the build/bloom server?

nuclearsandwich:
> How can I get colcon-meson released as Debian package

There are two main steps to this process:

1. Use [stdeb] and [publish-python](https://github.com/dirk-thomas/publish-python) to build and publish colcon-meson to the Colcon apt repository on packagecloud.
2. Add colcon-meson to the list of colcon packages imported into the ROS bootstrap repository

These steps have a lot of prerequisites and first time setup but are fairly routine once those are taken care of. A major inhibitor of 1 right now is that not everyone on the active colcon maintenance team can push to the repository on packagecloud and no currently active maintainers can manage those permissions. We're working on sorting that out but in the meantime we should be able to support at least one or two releases per month with our current setup.

I'd suggest taking a look at colcon-core's [stdeb.cfg](https://github.com/colcon/colcon-core/blob/0bf904e205aaae3055f9b8fefbdd4f52806b66e6/stdeb.cfg) and [publish-python.yaml](https://github.com/colcon/colcon-core/blob/0bf904e205aaae3055f9b8fefbdd4f52806b66e6/publish-python.yaml) configuration files and adding them to colcon-meson, if you tag me for review I can give feedback on them.
Once those files are in and you can build the debs successfully locally we'll get them pushed to the colcon apt repo and walk through the PR to add it to the bootstrap repo's import configs.

christianrauch:
The first step has been implemented by https://github.com/colcon/colcon-meson/pull/8 and tested with `publish-python` to generate a `python3-colcon-meson_0.4.1-1_all.deb`.

I assume someone else has to do the `publish-python --upload` now?

nuclearsandwich:
> I assume someone else has to do the `publish-python --upload` now?

Yes, since you've shared access on PyPI to cottsay and myself we can run releases for both pypi and the colcon apt repository, we can also just run the apt repository upload after you've done the pypi release


That is, you could run
```
publish-python wheel:pypi --upload
```

and I would run
```
publish-python stdeb:packagecloud --upload
```

It would be easiest to do this with a new tagged release since that would avoid needing to checkout the config files from main in a 0.4.1 checkout to perform the release..

christianrauch:
I tagged a new [0.4.2 release](https://github.com/colcon/colcon-meson/releases/tag/0.4.2). I am happy for you to do both releases at once.

ivanpauno:
@christianrauch is this PR still needed?

christianrauch:
> @christianrauch is this PR still needed?

Kind of. I will update this PR to use the Debian package once it has been released.

nuclearsandwich:
> Kind of. I will update this PR to use the Debian package once it has been released.

As of https://github.com/ros-infrastructure/reprepro-updater/pull/172 colcon-meson 0.4.2 is available in the ROS apt repositories.

ivanpauno (approved):

PR #34285, opened 2022-08-31
Revert "qb_device: 3.0.4-3 in 'melodic/distribution.yaml' [bloom] (#3…

clalancette:
…3887)"

This reverts commit 04a5dace39651931af56e564b2a8d107d1458f83.

This hasn't compiled since it was updated in July: https://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__qb_device_hardware_interface__ubuntu_bionic_amd64__binary/

@umbertofontana93 FYI

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -9795,7 +9795,6 @@ repositories:
       - qb_device_control
       - qb_device_description
       - qb_device_driver
-      - qb_device_gazebo
       - qb_device_hardware_interface
       - qb_device_msgs
       - qb_device_srvs
@@ -9803,7 +9802,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://bitbucket.org/qbrobotics/qbdevice-ros-release.git
-      version: 3.0.4-3
+      version: 2.0.1-0
     source:
       type: git
       url: https://bitbucket.org/qbrobotics/qbdevice-ros.git


umbertofontana93:
@clalancette what does it mean? Do you revert my recent bloom release version(v3.0.4) of qb_device? What I have to do now to update the released package version?

clalancette:
> @clalancette what does it mean? Do you revert my recent bloom release version(v3.0.4) of qb_device? What I have to do now to update the released package version?

So I reverted the release in the Melodic distribution, because it was failing to build (as in the link above).  That just means the next Melodic sync will go out with the "old" version of your package.

To get your new version in, you need to make the fixes to the source repository and then do a new release.  Looking at https://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__qb_device_hardware_interface__ubuntu_bionic_amd64__binary/101/console , for instance, my guess is that you have forgotten to declare a dependency on `control_msgs` (either in your package.xml or CMakeLists.txt, or both).  Once that is fixed, and a new release is done, we can consider it for inclusion in a future Melodic sync.

umbertofontana93:
Thank you for your answer. The same ploblem is present also in noetic release. 
I will fix this bug and I will release it again.

Why this error does not appear performing prerelease tests with http://prerelease.ros.org ?

Thank you again.

PR #34288, opened 2022-08-31
Revert "libfranka: 0.9.2-1 in 'melodic/distribution.yaml' [bloom] (#3…

clalancette:
…4280)"

This reverts commit 816bdf4417f6a0e8937e88d343cf14bdde93aab3.

When trying to build downstream packages with it, we get errors like:

```
CMake Error in CMakeLists.txt:
  Imported target "Franka::Franka" includes non-existent path

    "/opt/ros/melodic/include/libfranka"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

See https://build.ros.org/job/Mbin_uB64__franka_visualization__ubuntu_bionic_amd64__binary/54/ for more information.

@marcbone FYI

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6028,7 +6028,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/frankaemika/libfranka-release.git
-      version: 0.9.2-1
+      version: 0.9.1-1
     source:
       test_commits: false
       type: git


clalancette:
> we still do not quite know what is wrong with our cmake that it does not build in your jenkins. However, reverting to 0.9.1 will probably not work, as we made the 0.9.2 release especially because 0.9.1 would cause an infinite recursion with cmake on your build server. 0.9.0 was the last version that worked.

Yeah, I had to back up to 0.9.0 to make this work (see https://github.com/ros/rosdistro/pull/34289).

From CI, it looks like something is going wrong with how the cmake exports are done in libfranka.  Downstream targets all show something like:

```
CMake Error in CMakeLists.txt:
  Imported target "Franka::Franka" includes non-existent path

    "/opt/ros/melodic/include/libfranka"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

PR #34293, opened 2022-09-01
Add typeguard-pip to python.yaml

InigoMoreno:

## Package name:

typeguard-pip

## Package Upstream Source:
[link to github repository](https://github.com/agronholm/typeguard)
[link to github repository](https://pypi.org/project/typeguard/)

## Purpose of using this:

There is a python3-typeguard, which points to the apt version of the package. However, those binaries are still on version 2.2.2 of typeguard (which was released on 2018-08-13). I suggest adding this typeguard-pip to be able to use the more recent versions available on pip.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9140,6 +9140,16 @@ python3-typeguard:
     bionic:
       pip:
         packages: [typeguard]
+python3-typeguard-pip:
+  debian:
+    pip:
+      packages: [typeguard]
+  fedora:
+    pip:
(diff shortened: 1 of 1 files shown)


jacobperron (changes requested):
Following the [CONTRIBUTING](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1) guide, the entry should start with the prefix `python3-`.

jacobperron (approved):

audrow (approved):

PR #34376, opened 2022-09-08
naoqi_libqi: 2.9.7-0 in 'foxy/distribution.yaml' [bloom]

mbusy:
Increasing version of package(s) in repository `naoqi_libqi` to `2.9.7-0`:

- upstream repository: https://github.com/ros-naoqi/libqi.git
- release repository: https://github.com/ros-naoqi/libqi-release.git
- bloom version: `0.11.2`

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2945,6 +2945,21 @@ repositories:
       url: https://github.com/KIT-MRT/mrt_cmake_modules.git
       version: master
     status: maintained
+  naoqi_libqi:
+    doc:
+      type: git
+      url: https://github.com/ros-naoqi/libqi.git
+      version: master
+    release:
+      tags:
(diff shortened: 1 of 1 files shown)


mbusy:
Mh. It would seem that the tag from the upstream repository that I'm using is upsetting the test somehow
> Could not validate repository with url https://github.com/ros-naoqi/libqi.git and version v2.5.0-foxy from entry at line 2958.
> Error reason: Tags are not valid, but a tag v2.5.0-foxy was found

audrow:
@mbusy, I think that it's because the tag already exists. I think you can either force the tags (probably an option with bloom), or just bump the patch version, again.

mbusy:
Okay, I'll probably try to bump it and re release it then

mbusy:
Closing in favour of #34399

PR #34417, opened 2022-09-10
linear_sum_assignment: 1.9.2-1 in 'rolling/distribution.yaml' [bloom]

wep21:
Increasing version of package(s) in repository `linear_sum_assignment` to `1.9.2-1`:

- upstream repository: https://github.com/wep21/linear_sum_assignment.git
- release repository: https://github.com/ros2-gbp/linear_sum_assignment-release.git
- bloom version: `0.11.1`

## linear_sum_assignment

```
* docs: add license and CONTRIBUTING.md
* docs: add readme
* feat: initial commit
* Contributors: Daisuke Nishimatsu
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1902,6 +1902,21 @@ repositories:
       url: https://github.com/ros2/libyaml_vendor.git
       version: rolling
     status: maintained
+  linear_sum_assignment:
+    doc:
+      type: git
+      url: https://github.com/wep21/linear_sum_assignment.git
+      version: main
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/linear_sum_assignment-release.git
+      version: 1.9.2-1
+    source:
+      type: git
+      url: https://github.com/wep21/linear_sum_assignment.git
+      version: main
+    status: maintained
   locator_ros_bridge:
     doc:
       type: git


clalancette:
So I guess my overall issue with this package is that it seems like a weird repackaging of scipy, which we already have rosdep keys for.  Can you explain more about what this is being used for, and why we can't just use scipy for this?

wep21:
@clalancette I would like to use hungarian assignment method in C++ and found it in `scipy`. However, I don't know the way to call `scipy` from c++, so I exposed a part of `scipy` as a C library.

clalancette:
> @clalancette I would like to use hungarian assignment method in C++ and found it in `scipy`. However, I don't know the way to call `scipy` from c++, so I exposed a part of `scipy` as a C library.

OK.  I'm honestly not sure that this is a great idea to do here.  Also, this isn't really a ROS-specific package, it's just a generic package.  What's the plan moving forward?  Do you have a package you are going to submit that depends on this?

jacobperron:
[Holding until build farm maintenance is complete.](https://discourse.ros.org/t/sync-hold-2022-09-19-2022-09-20-for-planned-build-ros2-org-maintenance/27302)

wep21:
@clalancette 

> Do you have a package you are going to submit that depends on this?

Yes, I will, but it will take some time to add the package.

wjwwood:
@clalancette should we continue discussing this one or move forward with it?

clalancette:
> @clalancette should we continue discussing this one or move forward with it?

This package still seems extremely weird to me.  I guess technically it doesn't violate our conventions, but I'm just not sure we want to set this precedent.

I guess I'll ask to get @nuclearsandwich 's assessment of it.  If he is OK with it, then I'm OK with it.

wep21:
Okay, I will include a part of scipy function in the package which uses the function. It doesn't matter for me to close this PR.

nuclearsandwich:
> This package still seems extremely weird to me. I guess technically it doesn't violate our conventions, but I'm just not sure we want to set this precedent.
> 
> I guess I'll ask to get @nuclearsandwich 's assessment of it. If he is OK with it, then I'm OK with it.

I want to acknowledge @wep21's efforts. I do see how creating this package fits in with our preferences as expressed in prior discussions and thus would appear to be the solution we would most prefer since it essentially a very scoped vendor package. I appreciate that time and energy were spent in putting this together. Thanks very much for both the efforts you made to conform to expected conventions _and_ your flexibility when the result turned out different.

> Okay, I will include a part of scipy function in the package which uses the function. It doesn't matter for me to close this PR.

To me, this is the important part: all we want is a specific function and that function has use within a single package. Unlike the usual vendor case where we are after an entire upstream project and/or we want to make some code available to a large number of downstream projects.

I think that in this case including the function directly within the package that needs it is the correct answer (with proper citation and copyright acknowledgement as required by scipy's license upstream). There is probably no exact number of functions, level of re-use, or other mechanical criteria which I could identify to separate when you should vendor a function into the package like that versus when to vendor the required code directly into the package which uses it.

If you find yourself pulling more functions from scipy over time or needing them in other projects, it might be worthwhile to work with the scipy project upstream on official C++ bindings (a `libscipycpp` if you will) which would almost certainly be a valid candidate to be packaged in Debian and Ubuntu alongside scipy itself thus making it usable within ROS packages.

ivanpauno:
@wep21 do you want to move forward with this PR or do you prefer to add the function directly to your package?

wep21:
@nuclearsandwich Thank you very much for your kind explanation.
@jacobperron @clalancette I decide to close this PR. I will create a new package if I need more functions from `scipy`.

PR #34422, opened 2022-09-11
add overlay_rviz_plugins repository to rolling

ottojo:
Adding new repository https://github.com/teamspatzenhirn/overlay_rviz_plugins to `rolling` as requested in https://github.com/ros2-gbp/ros2-gbp-github-org/issues/112#issuecomment-1242351309,
for the purpose of releasing the packages
* [overlay_rviz_msgs](https://github.com/teamspatzenhirn/overlay_rviz_plugins/tree/main/overlay_rviz_msgs)
* [overlay_rviz_plugins](https://github.com/teamspatzenhirn/overlay_rviz_plugins/tree/main/overlay_rviz_plugins)

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2683,6 +2683,12 @@ repositories:
       url: https://github.com/OUXT-Polaris/ouxt_common.git
       version: master
     status: developed
+  overlay_rviz_plugins:
+    source:
+      type: git
+      url: https://github.com/teamspatzenhirn/overlay_rviz_plugins.git
+      version: main
+    status: maintained
   pcl_msgs:
     release:
       tags:


rolling/distribution.yaml
```diff
@@ -2683,6 +2683,12 @@ repositories:
       url: https://github.com/OUXT-Polaris/ouxt_common.git
       version: master
     status: developed
+  overlay_rviz_plugins:
```
clalancette, line 2686: So this key doesn't violate any of our naming rules, per-se.  However, the word `overlay` is overloaded in the ROS ecosystem.  In this case you are talking about scene overlays, which is a perfectly valid use, but we also use "overlay" to describe packages layered on top of other packages.  In this context, I think it is easy to conflate the two.

My suggestion is something like `rviz_2d_overlay_plugins`, which I think better disambiguates this term.  But pinging @ros/rosdistro-maintainers for additional feedback.

ottojo: We chose the suffix `rviz_plugins` because some other packages follow that pattern (`nav2_rviz_plugins` etc), but `rviz_*_plugins` seems obvious enough to me... The `overlay` issue didn't come to mind since i don't usually use more than one overlay, at least not knowingly...

Renaming is not a problem, `rviz_2d_overlay_plugins` sounds good to me, will wait for additional feedback!

jacobperron: How about `overlay_2d_rviz_plugins`? I'm also fine with `rviz_2d_overlay_plugins`

jacobperron:
[Holding until build farm maintenance is complete.](https://discourse.ros.org/t/sync-hold-2022-09-19-2022-09-20-for-planned-build-ros2-org-maintenance/27302)

rolling/distribution.yaml
clalancette: So it has been a week, and I think we've gotten most of the feedback we are going to get.  @ottojo I'll suggest going with `rviz_2d_overlay_plugins`.  In that case, I'll suggest closing this PR, making the changes to the sources to rename it, and then opening a new PR with the new package name.  Does that sound reasonable?

ottojo:
closing, will open new PR with updated name

PR #34442, opened 2022-09-12
Add python3-git for rhel

DLu:

## Package name:

`python3-git`

## Package Upstream Source:

[`python3-GitPython (fedora)`](https://packages.fedoraproject.org/pkgs/GitPython/python3-GitPython/)

## Purpose of using this:

Package exists for other distros, just not RHEL. Blocking the Galactic blooming of [`system_fingerprint`](https://github.com/MetroRobots/ros_system_fingerprint)

  - [`python3-GitPython (fedora)`](https://packages.fedoraproject.org/pkgs/GitPython/python3-GitPython/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6855,7 +6855,9 @@ python3-git:
   fedora: [python3-GitPython]
   gentoo: [dev-python/git-python]
   nixos: [python3Packages.GitPython]
-  rhel: [python3-GitPython]
+  rhel:
+    '*': [python3-GitPython]
+    '7': null
   ubuntu: [python3-git]
 python3-github:
   debian: [python3-github]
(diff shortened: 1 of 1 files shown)


DLu:
@clalancette Any idea why this one doesn't work in rolling?

clalancette:
> @clalancette Any idea why this one doesn't work in rolling?

It's the same reason #34443; it is only available for EPEL-8.  The same solution should apply.

rosdep/python.yaml
```diff
@@ -6855,6 +6855,7 @@ python3-git:
   fedora: [python3-GitPython]
   gentoo: [dev-python/git-python]
   nixos: [python3Packages.GitPython]
+  rhel: [python3-GitPython]
```
DLu, line 6858:
```suggestion
  rhel:
    '*': [python3-GitPython]
    '7': null
```

clalancette (approved):

DLu:
Who should give this a second review? @tfoote

PR #34582, opened 2022-09-16
Revert the latest updates to Foxy and Galactic vision_opencv

clalancette:
While that package itself succeeds, it seems to be causing problems in downstream packages:

https://build.ros2.org/job/Fbin_uF64__rqt_image_view__ubuntu_focal_amd64__binary/87/console
https://build.ros2.org/job/Gbin_uF64__rqt_image_view__ubuntu_focal_amd64__binary/43/console

@ijnek FYI.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -7183,7 +7183,7 @@ repositories:
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/ros2-gbp/vision_opencv-release.git
-      version: 3.0.4-1
+      version: 2.2.1-1
     source:
       test_pull_requests: true
       type: git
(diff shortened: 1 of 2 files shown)


ijnek:
Thanks @clalancette , I'll have a look into it.

ijnek:
ps. Despite the major version bump in galactic and foxy, there shouldn't have been any API or ABI breaking changes.

ijnek:
Ok, I can reproduce the problem and I understand why it's happening now. I'm going to track the issue in https://github.com/ros-perception/vision_opencv/issues/485, and try and get a non-breaking release out.

ijnek:
Is there an easy way to test for such breakages before performing another release? I've reverted the API breaking changes in foxy and galactic and want to make another release.

clalancette:
> Is there an easy way to test for such breakages before performing another release? I've reverted the API breaking changes in foxy and galactic and want to make another release.

There's a video about using prerelease to do this testing here: https://www.youtube.com/watch?v=XA20_iRBLrg

ijnek:
Thanks, I will check it out.

PR #34616, opened 2022-09-22
Add python3-mako rule

zflat:
Please update the following dependency to the rosdep database. The `python-mako` package is not available for Ubuntu after focal. This change allows for the `python3-mako` package to be installed for newer versions of Ubuntu. 

## Package name:

python3-mako

## Package Upstream Source:

https://www.makotemplates.org/

## Purpose of using this:

This is an existing package already in the rosdep index.

  - REQUIRED
  - https://packages.debian.org/buster/python-mako
  - https://packages.debian.org/buster/python3-mako
   - REQUIRED
   - https://packages.ubuntu.com/focal/python-mako
   - https://packages.ubuntu.com/focal/python3-mako
  - IF AVAILABLE
  - https://packages.fedoraproject.org/pkgs/python-mako/python3-mako/
  - IF AVAILABLE
  - IF AVAILABLE
  - https://packages.gentoo.org/packages/dev-python/mako
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2564,9 +2564,7 @@ python-mako:
   debian: [python-mako]
   fedora: [python-mako]
   gentoo: [dev-python/mako]
-  ubuntu:
-    '*': [python3-mako]
-    bionic: [python-mako]
+  ubuntu: [python-mako]
 python-mapnik:
   debian: [python-mapnik]
   fedora: [mapnik-python]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2564,7 +2564,9 @@ python-mako:
   debian: [python-mako]
   fedora: [python-mako]
   gentoo: [dev-python/mako]
-  ubuntu: [python-mako]
+  ubuntu:
+    '*': [python3-mako]
+    bionic: [python-mako]
 python-mapnik:
   debian: [python-mapnik]
   fedora: [mapnik-python]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6716,6 +6716,12 @@ python3-lxml:
       packages: [lxml]
   rhel: ['python%{python3_pkgversion}-lxml']
   ubuntu: [python3-lxml]
+python3-mako:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2564,7 +2564,9 @@ python-mako:
   debian: [python-mako]
   fedora: [python-mako]
   gentoo: [dev-python/mako]
-  ubuntu: [python-mako]
+  ubuntu:
+    '*': [python3-mako]
+    bionic: [python-mako]
```
jacobperron, line 2569: Rather than mixing Python 2 and Python 3 into the same rule, we prefer to create a separate rules for each. Instead of modifying this rule, please create a new rule for `python3-mako` ([see guideline](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-3)).

jacobperron (dismissed):

rosdep/python.yaml
zflat: OK, I am leaving the original key untouched and adding a `python3-mako` key. I copied over the fedora and gentoo entries from the `python-mako` key but am not sure if the naming is correct for the following packages that I found:
- https://packages.fedoraproject.org/pkgs/python-mako/python3-mako/
- https://packages.gentoo.org/packages/dev-python/mako

ivanpauno (approved):

ivanpauno:
@zflat could you address this automatic suggestion?

> Key 'python3-mako' might be satisifed by rhel package named 'python3-mako': https://repo.almalinux.org/almalinux/8/AppStream/x86_64/os/Packages/python3-mako-1.0.6-13.el8.noarch.rpm

zflat:
@ivanpauno How does this look for adding the `rehl` entry? 53a6bef

zflat:
@ivanpauno Adding the rehl entry is causing the nosetests to fail. Do you have any suggestions?

ivanpauno (approved):

PR #34630, opened 2022-09-23
webots_ros2: 2022.1.0-1 in 'rolling/distribution.yaml' [bloom]

ad-daniel:
Increasing version of package(s) in repository `webots_ros2` to `2022.1.0-1`:

- upstream repository: https://github.com/cyberbotics/webots_ros2.git
- release repository: https://github.com/ros2-gbp/webots_ros2-release.git
- bloom version: `0.11.2`
- previous version for package: `1.2.3-1`

## webots_ros2

```
* Adapted controllers to communicate with Webots R2022b.
* Added feature to import URDF on the fly.
* Add PointCloud2 support for RangeFinder.
```

## webots_ros2_driver

```
* Added an URDF importer feature to spawn robots from URDF files.
```

## webots_ros2_importer

```
* Upgraded to urdf2webots 2.0.0
```

## webots_ros2_msgs

```
* Added URDF robot messages
```

## webots_ros2_universal_robot

```
* The 'robot' and 'moveit_demo' scenarios now show a use case of the new URDF importer.
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -6131,7 +6131,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/webots_ros2-release.git
-      version: 1.2.3-1
+      version: 2022.1.0-1
     source:
       test_pull_requests: true
       type: git


nuclearsandwich:
@ad-daniel these releases are all failing on the ROS build farm because of the dependency on `urdf2webots-pip`.

Packages cannot be built on the build farm unless all of their dependencies are available in the apt / rpm repositories listed in [expected rosdep sources](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#expected-rosdep-sources).

ad-daniel:
@nuclearsandwich I'm not sure to understand, according to the docs, pip dependencies are allowed https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip and in this case urdf2webots is indeed available at https://pypi.org/project/urdf2webots/. The corresponding keys are available in rosdep https://github.com/ros/rosdistro/blob/ca22bb374d02192e7cf9248e12790269afd7b436/rosdep/python.yaml#L9533-L9542

It isn't a new change on our end, it's how the package has always been structured, not sure why this time around the build-farm has issues with it. The apt route isn't very practical for us, we might instead just refactor the code so that it isn't an external dependency, if no other option is available that is.

Either way, do we need to manually revert the changes that have been merged in rosdistro?

Thank you

jacobperron:
Perhaps not in the most obvious place, but here: https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md#purpose-of-this-repository

> ROS packages that depend on "pip" keys cannot be "released" into a ROS distribution. They can only be depended on by from-source builds. Because of this, system packages are highly preferred to pip packages.

In version 1.2.3, I do not see a declared dependency to `urdf2webots-pip` in any package.xml file, but in the newer `2022.1.0`, I do see it declared here: https://github.com/cyberbotics/webots_ros2/blob/d76c2b15a6595645b173f0940519fb643366400b/webots_ros2_driver/package.xml#L28

This is why there was no issue on the buildfarm with the previously released version.

PR #34675, opened 2022-09-27
rviz_2d_overlay_plugins: 1.2.0-2 in 'rolling/distribution.yaml' [bloom]

ottojo:
Increasing version of package(s) in repository `rviz_2d_overlay_plugins` to `1.2.0-2`:

- upstream repository: https://github.com/teamspatzenhirn/rviz_2d_overlay_plugins.git
- release repository: https://github.com/ros2-gbp/rviz_2d_overlay_plugins-release.git
- distro file: `rolling/distribution.yaml`
- bloom version: `0.11.2`
- previous version for package: `null`

## rviz_2d_overlay_msgs

```
* Rename package from overlay_rviz_msgs to rviz_2d_overlay_msgs
* Contributors: Jonas Otto
```

## rviz_2d_overlay_plugins

```
* Add missing dependencies to package.xml
* Fix various compiler-warnings (NFC)
* Rename package from overlay_rviz_plugins to rviz_2d_overlay_plugins
* Contributors: Jonas Otto
```

PR #34676, opened 2022-09-27
rviz_2d_overlay_plugins: 1.2.0-2 in 'humble/distribution.yaml' [bloom]

ottojo:
Increasing version of package(s) in repository `rviz_2d_overlay_plugins` to `1.2.0-2`:

- upstream repository: https://github.com/teamspatzenhirn/rviz_2d_overlay_plugins.git
- release repository: https://github.com/ros2-gbp/rviz_2d_overlay_plugins-release.git
- bloom version: `0.11.2`

## rviz_2d_overlay_msgs

```
* Rename package from overlay_rviz_msgs to rviz_2d_overlay_msgs
* Contributors: Jonas Otto
```

## rviz_2d_overlay_plugins

```
* Add missing dependencies to package.xml
* Fix various compiler-warnings (NFC)
* Rename package from overlay_rviz_plugins to rviz_2d_overlay_plugins
* Contributors: Jonas Otto
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5176,6 +5176,21 @@ repositories:
       url: https://github.com/ros2/rviz.git
       version: humble
     status: maintained
+  rviz_2d_overlay_plugins:
+    release:
+      packages:
+      - rviz_2d_overlay_msgs
+      - rviz_2d_overlay_plugins
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/ros2-gbp/rviz_2d_overlay_plugins-release.git
+      version: 1.2.0-2
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/teamspatzenhirn/rviz_2d_overlay_plugins.git
+      version: main
+    status: developed
   rviz_visual_tools:
     doc:
       type: git
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5177,6 +5177,10 @@ repositories:
       version: humble
     status: maintained
   rviz_2d_overlay_plugins:
+    doc:
+      type: git
+      url: https://github.com/teamspatzenhirn/rviz_2d_overlay_plugins.git
+      version: main
     release:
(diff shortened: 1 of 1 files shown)


ottojo:
Question: I followed the release process for `rolling` so far (https://github.com/ros/rosdistro/pull/34675), but we currently use/develop this package on `humble`.

I did check that the package builds on `rolling`, but i don't really want to maintain two separate branches for `rolling` and `humble` (or whatever the latest ros2 release is in the future).

Do you suggest not going forward with the `rolling` release in that case, even if the same branch (currently, by accident i suppose) works for both ros2 releases currently?

clalancette:
> edit: Is is required/best practice to release on rolling anyways, as a "staging" environment to catch potential issues with future releases i guess?

It is not required to release into Rolling, but we do consider it to be best practice.  For exactly the reason you state; to catch potential issues early.  If it is in Rolling, it will also be automatically released into the next ROS 2 distribution.

> I did check that the package builds on rolling, but i don't really want to maintain two separate branches for rolling and humble (or whatever the latest ros2 release is in the future).

How you deal with your branching is completely up to you.  The core ROS 2 packages have a separate branch for each ROS distribution; that allows us to evolve them independently.  Some packages in here use a single branch for all ROS 2 distributions.  Some packages in here use a single branch for all ROS 1 *and* ROS 2 distributions (though that takes more work).

In your case, my suggestion would be to keep using the same branch for both Humble and Rolling (with releases in both) as long as you can.  If a breaking change comes to Rolling, then you can consider splitting off into a separate branch for that.

ottojo:
Thank you very much! In that case i will leave the PR as is (using `main` for both `rolling` and `humble')! I will also add the `doc` section in here, as in the other PR.

wjwwood:
[Holding for Humble sync](https://discourse.ros.org/t/preparing-for-humble-sync-2022-09-30)

ottojo:
New version released in the meantime, closing in favor of #34714

PR #34739, opened 2022-10-03
Add psutil-pip

igor-markovic:

## Package name:

pip/pip3 psutil

## Package Upstream Source:

https://github.com/giampaolo/psutil

## Purpose of using this:

Similar to #34293, the `python3-psutil` version is outdated via apt, so this simply adds the pip definition to be able to use the latest `psutil` library.

### Tags for anyone searching for a solution to this kind of issue in the future:
apt outdated old version pip new

- Pip: https://pypi.org/project/psutil/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7888,6 +7888,16 @@ python3-psutil:
   rhel: ['python%{python3_pkgversion}-psutil']
   slackware: [psutil]
   ubuntu: [python3-psutil]
+python3-psutil-pip:
+  debian:
+    pip:
+      packages: [psutil]
+  fedora:
+    pip:
+      packages: [psutil]
+  ubuntu:
+    pip:
+      packages: [psutil]
 python3-py3exiv2-pip:
   debian:
     pip:


ivanpauno (approved):

ivanpauno (changes requested):
Actually, there's already a `python3-psutil` key.
When debian packages are available, the idea is to profer those.
So no need to add the pip key.

igor-markovic:
@ivanpauno yes, typically the standard debian package would work, but in this case it is very out of date. This pull request just adds the pip package separately so we can use the latest version.

Please take a look at the linked pull request #34293, which was approved by @jacobperron and @audrow, and merged.

ivanpauno:
The policy is that we shouldn't be adding both alternatives if a debian package is available, see https://github.com/ros/rosdistro/pull/33017#pullrequestreview-1028893193.

The [CONTRIBUTING](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1) guide is not super explicit, but it says that pip packages can be added if they are **only** available on PyPI (and not in the system package manager).

PR #34756, opened 2022-10-03
Added python3-rich-pip

asasine:

## Package name:

rich

## Package Upstream Source:

https://github.com/Textualize/rich

## Purpose of using this:

Rich is a Python library for rich text and beautiful formatting in the terminal.

The [Rich API](https://rich.readthedocs.io/en/latest/) makes it easy to add color and style to terminal output. Rich can also render pretty tables, progress bars, markdown, syntax highlighted source code, tracebacks, and more — out of the box.

Distro packaging links: https://pypi.org/project/rich/

- pip: https://pypi.org/project/rich/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8608,6 +8608,16 @@ python3-retrying:
     pip:
       packages: [retrying]
   ubuntu: [python3-retrying]
+python3-rich-pip:
+  debian:
+    pip:
+      packages: [rich]
+  osx:
+    pip:
+      packages: [rich]
+  ubuntu:
+    pip:
+      packages: [rich]
 python3-river-pip:
   debian:
     pip:


methylDragon (approved):

rosdep/python.yaml
```diff
@@ -8608,6 +8608,16 @@ python3-retrying:
     pip:
       packages: [retrying]
   ubuntu: [python3-retrying]
+python3-rich-pip:
```
ivanpauno, line 8611: could you add a python3-rich key instead?

- https://packages.ubuntu.com/search?keywords=python3-rich
- https://packages.debian.org/search?keywords=python3-rich

For older distros of ubuntu that don't have a binary package available, you can use pip

ivanpauno (changes requested):

rosdep/python.yaml
ivanpauno: See https://github.com/ros/rosdistro/pull/33017#pullrequestreview-1028893193

asasine: Fair enough, thanks for taking a look at this PR! Going to abandon it and go a different route, perhaps custom rosdep rules installed on the local machine.

PR #34886, opened 2022-10-11
libtiff-dev exists in the generic on all supported platforms

tfoote:
We don't need the specific version numbers, they're aliased by the versionless one.

Ubuntu: https://packages.ubuntu.com/bionic/libtiff-dev

Debian: https://packages.debian.org/buster/libtiff-dev

I needed this to be able to use it on Jammy and newer platforms. I cleaned it up instead of just extending.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5373,11 +5373,7 @@ libtheora:
   ubuntu: [libtheora-dev]
 libtiff-dev:
   arch: [libtiff]
-  debian:
-    buster: [libtiff5-dev]
-    jessie: [libtiff5-dev]
-    stretch: [libtiff5-dev]
-    wheezy: [libtiff5-alt-dev]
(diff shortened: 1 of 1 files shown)


hidmic (approved):
@tfoote I don't see `libtiff-dev` in some of the oldest Debian/Ubuntu (EOL'd) distros. Not that we should care, but I suspect this change might break packages built on embedded systems running Linux distributions based on those older distros.

tfoote:
The expectation for older EOL distros is that you should use the older tags or rosdistro from when the EOL platform was last supported (likely the last sync of the relevant EOL rosdistro). As such we're not responsible for maintaining backwards compatibility beyond our supported platforms for rosdep keys. This needed to be implemented because otherwise the support burden of maintaining the keys grows unbounded.

cottsay (approved):

PR #34978, opened 2022-10-17
update Libdc1394-dev for newer Ubuntu releases

### LINKED ISSUE #34921: libdc1394-22-dev not longer available for Debian and Ubuntu after version 21.10.
After Debian/Ubuntu Version 21.10. the `libdc1394-22-dev` package was replaced by `libdc1394-dev`. (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=963924 and https://askubuntu.com/questions/1407580/unable-to-locate-package-libdc1394-22-dev)
However the entry in `rosdep/base.yaml` still reflects to the removed `libdc1394-22-dev`: https://github.com/ros/rosdistro/blob/cfe61316fcb3af6a4790c1b96edb8f1912a70e9e/rosdep/base.yaml#L3087-L3099
Therefore in ROS Humble (which uses Ubuntu 22 by default), using the rosdep dependency `<depend>libdc1394-22-dev</depend>` will lead to an issue: 
```
E: Unable to locate package libdc1394-22-dev
  apt: command [sudo -H apt-get install libdc1394-22-dev] failed
executing command [sudo -H apt-get install libdc1394-22-dev]
```
I'm not sure how to solve this issue. Because changing the `base.yaml` to the new version would probably lead to errors for older ROS Versions right?
I'm not that familiar with rosdep. Is there any possibility to create different entries for different Ubuntu/Debian Version?

FranekStark:

## Package name:

libdc1394-dev

## Purpose of using this:
Since Ubuntu kinetic the package `libdc1394-22-dev` is replaced by `libdc1394-dev`. 
Therefore I set `libdc1394-dev` as the new standard package for Ubuntu and added rules for the older Ubuntu releases.
See:
- https://askubuntu.com/questions/1407580/unable-to-locate-package-libdc1394-22-dev
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=963924

closes [#34921](https://github.com/ros/rosdistro/issues/34921)

   - https://packages.ubuntu.com/jammy/libdc1394-dev
   - https://packages.ubuntu.com/kinetic/libdc1394-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3101,8 +3101,8 @@ libdc1394-dev:
       packages: [libdc1394]
   ubuntu:
     '*': [libdc1394-dev]
-    focal: [libdc1394-22-dev]
     bionic: [libdc1394-22-dev]
+    focal: [libdc1394-22-dev]
 libdevil-dev:
   arch: [devil]
   debian: [libdevil-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3090,8 +3090,8 @@ libdc1394-dev:
       packages: [libdc1394]
   debian:
     '*': [libdc1394-dev]
-    'buster': [libdc1394-22-dev]
-    'stretch': [libdc1394-22-dev]
+    buster: [libdc1394-22-dev]
+    stretch: [libdc1394-22-dev]
   fedora: [libdc1394-devel]
   gentoo: [media-libs/libdc1394]
   nixos: [libdc1394]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3103,6 +3103,7 @@ libdc1394-dev:
     'impish': [libdc1394-22-dev]
     'focal': [libdc1394-22-dev]
     'bionic': [libdc1394-22-dev]
+    'artful': [libdc1394-22-dev]
 libdevil-dev:
   arch: [devil]
   debian: [libdevil-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3098,12 +3098,8 @@ libdc1394-dev:
       packages: [libdc1394]
   ubuntu:
     '*': [libdc1394-dev]
-    'artful': [libdc1394-22-dev]
-    'bionic': [libdc1394-22-dev]
     'focal': [libdc1394-22-dev]
-    'impish': [libdc1394-22-dev]
-    'trusty': [libdc1394-22-dev]
-    'xenial': [libdc1394-22-dev]
+    'bionic': [libdc1394-22-dev]
 libdevil-dev:
   arch: [devil]
   debian: [libdevil-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3096,7 +3096,13 @@ libdc1394-dev:
   osx:
     macports:
       packages: [libdc1394]
-  ubuntu: [libdc1394-22-dev]
+  ubuntu:
+    '*': [libdc1394-dev]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
I suggest dropping the out-of-support distributions to keep things maintainable. It would also be nice to update the Debian keys equivalently (at least for bullseye and bookworm) but I can't use the suggestion API because those lines weren't touched.

rosdep/base.yaml
```diff
@@ -3096,7 +3096,14 @@ libdc1394-dev:
...
       packages: [libdc1394]
-  ubuntu: [libdc1394-22-dev]
+  ubuntu:
+    '*': [libdc1394-dev]
+    'trusty': [libdc1394-22-dev]
+    'xenial': [libdc1394-22-dev]
+    'impish': [libdc1394-22-dev]
+    'focal': [libdc1394-22-dev]
+    'bionic': [libdc1394-22-dev]
+    'artful': [libdc1394-22-dev]
```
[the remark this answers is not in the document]
FranekStark: ok thanks for the comment.
Yes makes sense, I wasn't sure how outdated distributions are handled. So I tried to add as many old distributions I found to be backwards compatible. 
Also I will add a new commit for debian in a moment

FranekStark:
@nuclearsandwich I removed the outdated ubuntu keys, and added the debian keys.

rosdep/base.yaml
```diff
@@ -3088,15 +3088,21 @@ libdc1394-dev:
   arch:
     pacman:
       packages: [libdc1394]
-  debian: [libdc1394-22-dev]
+  debian:
+    '*': [libdc1394-dev]
+    'buster': [libdc1394-22-dev]
+    'stretch': [libdc1394-22-dev]
```
clalancette, line 3094: While we aren't totally consistent about this, we don't typically quote these:

```suggestion
    buster: [libdc1394-22-dev]
    stretch: [libdc1394-22-dev]
```

rosdep/base.yaml
```diff
@@ -3088,15 +3088,21 @@ libdc1394-dev:
...
   nixos: [libdc1394]
   openembedded: [libdc1394@meta-multimedia]
   osx:
     macports:
       packages: [libdc1394]
-  ubuntu: [libdc1394-22-dev]
+  ubuntu:
+    '*': [libdc1394-dev]
+    'focal': [libdc1394-22-dev]
+    'bionic': [libdc1394-22-dev]
```
clalancette, line 3105: Same here:

```suggestion
    focal: [libdc1394-22-dev]
    bionic: [libdc1394-22-dev]
```

clalancette (changes requested):
Two small things to change, then I'm happy to approve this.

clalancette (approved):

rosdep/base.yaml
```diff
@@ -3088,15 +3088,21 @@ libdc1394-dev:
...
   nixos: [libdc1394]
   openembedded: [libdc1394@meta-multimedia]
   osx:
     macports:
       packages: [libdc1394]
-  ubuntu: [libdc1394-22-dev]
+  ubuntu:
+    '*': [libdc1394-dev]
+    focal: [libdc1394-22-dev]
+    bionic: [libdc1394-22-dev]
```
clalancette, line 3105:
```suggestion
    bionic: [libdc1394-22-dev]
    focal: [libdc1394-22-dev]
```

clalancette (changes requested):
One more change should fix CI.

clalancette (approved):

audrow (approved):

PR #34986, opened 2022-10-18
Add dynamixel_control to Rolling index

youtalk:
Signed-off-by: Yutaka Kondo <yutaka.kondo@youtalk.jp>

Rolling

# The source is here:

https://github.com/youtalk/dynamixel_control/tree/rolling

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -844,6 +844,16 @@ repositories:
       url: https://github.com/ros2/domain_bridge.git
       version: main
     status: developed
+  dynamixel_control:
+    doc:
+      type: git
+      url: https://github.com/youtalk/dynamixel_control.git
+      version: rolling
+    source:
+      type: git
+      url: https://github.com/youtalk/dynamixel_control.git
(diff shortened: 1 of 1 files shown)


clalancette (commented):
Looking through the sources to your package, I noticed that this package depends on https://github.com/ROBOTIS-GIT/dynamixel-workbench/tree/ros2 and https://github.com/ROBOTIS-GIT/dynamixel-workbench-msgs/tree/ros2 , neither of which are currently released into ROS 2.  That means that this package can't possibly be released into ROS 2 at the moment.  See also https://github.com/ROBOTIS-GIT/dynamixel-workbench-msgs/issues/23 , where I requested a release of at least the msgs.

We could still go forward and list this in the index for now.  @ros/rosdeputies what's our policy here?

tfoote (approved):
dynamixel_workbench and dynamixel_workbench_msgs have been released: #35080 and #35077 so this should be good to go once the conflicts are resolved. @youtalk

PR #35010, opened 2022-10-21
Adding new repository for indexing: rviz_ground_image

Plaba:
# Please Add rviz_static_ground_image to be indexed in the rosdistro.

FOXY

# The source is here:

https://github.com/Plaba/rviz_ground_image.git

# Checks

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -6221,6 +6221,11 @@ repositories:
       url: https://github.com/ros2/rviz.git
       version: foxy
     status: maintained
+  rviz_static_ground_image:
+    doc:
+      type: git
+      url: https://github.com/Plaba/rviz-ground-image.git
+      version: v0.0.1
   rviz_visual_tools:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


foxy/distribution.yaml
```diff
@@ -6221,6 +6221,11 @@ repositories:
       url: https://github.com/ros2/rviz.git
       version: foxy
     status: maintained
+  rviz_static_ground_image:
```
clalancette, line 6224: So there are 3 different names I'm seeing:

The key you are using here is `rviz_static_ground_image`.  The repository name is `rviz-ground-image`.  And the package name in the [package.xml](https://github.com/Plaba/rviz-ground-image/blob/main/package.xml) is `ground_image`.

Let's make all of these the same, preferably with an `rviz` prefix.

clalancette (changes requested):

foxy/distribution.yaml
Plaba: Thanks for pointing that out! I've renamed everything to `rviz_ground_image`.

clalancette (approved):
Thanks for doing the renaming, that's helpful.

With that, this looks good to me.

Plaba:
Thanks a lot!

PR #35060, opened 2022-10-26
Revert "libhri: 0.5.3-1 in 'melodic/distribution.yaml' [bloom] (#35053)"

clalancette:
This reverts commit 0a47f870e5dc3993b57390d728539af2f2a37c27.

This update is failing to build; see https://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__hri__ubuntu_bionic_amd64__binary/12/console .

@severin-lemaignan FYI

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -6062,7 +6062,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/ros4hri/libhri-release.git
-      version: 0.5.3-1
+      version: 0.5.1-1
     source:
       type: git
       url: https://github.com/ros4hri/libhri.git


severin-lemaignan:
@clalancette -- thanks for looking into it. I'm a bit puzzled by the error, I must say:

- first the build seems to use the package version 0.5.1 instead of the released 0.5.3:
```
/tmp/binarydeb/ros-melodic-hri-0.5.1/include/hri/face.h:33:10: fatal error: hri_msgs/NormalizedPointOfInterest2D.h: No such file or directory
 #include <hri_msgs/NormalizedPointOfInterest2D.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
- second, the error is about a missing header in `hri_msgs`, but both melodic and noetic have the same version of `hri_msgs` (0.8.0), and the build is fine on noetic...

not sure what is going on here!

PR #35061, opened 2022-10-26
Revert "qb_device: 3.0.5-1 in 'melodic/distribution.yaml' [bloom] (#3…

clalancette:
…4369)"

This reverts commit 16cc21b3883fca5532b10e54e433093fb299c50b.

This is causing downstream packages to fail to build, like: https://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__qb_move_hardware_interface__ubuntu_bionic_amd64__binary/102/console

@umbertofontana93 FYI

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -9806,7 +9806,6 @@ repositories:
       - qb_device_control
       - qb_device_description
       - qb_device_driver
-      - qb_device_gazebo
       - qb_device_hardware_interface
       - qb_device_msgs
       - qb_device_srvs
@@ -9814,7 +9813,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://bitbucket.org/qbrobotics/qbdevice-ros-release.git
-      version: 3.0.5-1
+      version: 2.0.1-0
     source:
       type: git
       url: https://bitbucket.org/qbrobotics/qbdevice-ros.git


umbertofontana93:
from [jenkins console](https://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__qb_move_hardware_interface__ubuntu_bionic_amd64__binary/102/console) it seems that the error is `no known conversion for argument 1 from ‘std::vector<unsigned char>’ to ‘const std::vector<int>&’`. However in the latest version of qb-device pkg (3.0.5) the argument is changed from _const std::vector<int>_ to _const std::vector<uint8_t>_. As you can see, the error is [here](https://bitbucket.org/qbrobotics/qbmove-ros/src/1a2e60a0b529d60be952d016d4261bbc255f0716/qb_move_hardware_interface/src/qb_move_hardware_interface.cpp#lines-75) and is caused by [this](https://bitbucket.org/qbrobotics/qbmove-ros/src/1a2e60a0b529d60be952d016d4261bbc255f0716/qb_move_hardware_interface/include/qb_move_hardware_interface/qb_move_transmission_interface.h#lines-249) function. The variable passed is of the same [type](https://bitbucket.org/qbrobotics/qbdevice-ros/src/35624ad46a877fdd8da0e643fe5c35a73c41a0b8/qb_device_hardware_interface/include/qb_device_hardware_interface/qb_device_hardware_resources.h#lines-77) declared in the new pkg version and no conversion is required.
Can all the errors be related to the fact that qbdevice was not merged and the qb_move package do not compile?

By running the official tool test this error does not appear. Can you suggest how I can do to upgrade my packages? What is the correct path to follow?

clalancette:
> Can all the errors be related to the fact that qbdevice was not merged and the qb_move package do not compile?

What I suspect is going on here is that the qb_move packages were too old, so they were looking for the old API.  My guess is that if you release both qb_move and qb_device again, it should work (as long as it works locally for you).  You may have to do a source release of them both.

umbertofontana93:
Locally they compile without any problem. What you suggest with a source release is to make a **bloom** release of both packages(qb_device and qb_move) at the same time?

clalancette:
Well, maybe first a source release (`catkin_generate_changelog`; `catkin_prepare_release`), followed by a `bloom-release`.

PR #35089, opened 2022-10-28
Adding witmotion_ros to documentation index for Noetic and Melodic

twdragon:
# Please Add Package `witmotion_ros` to be indexed in the rosdistro.

ROSDISTRO NAME: `noetic`, `melodic`

# The source is here:

https://github.com/ElettraSciComp/witmotion_IMU_ros

# The wiki page is here: 

https://wiki.ros.org/witmotion_ros

# Purpose of using this package

It is a ROS driver allowing the IMU sensors manufactured by [WitMotion Shenzhen Co.,Ltd](https://www.wit-motion.com/) to be used as a controllable provider of IMU information.

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -15385,11 +15385,6 @@ repositories:
       type: git
       url: https://github.com/ElettraSciComp/witmotion_IMU_ros.git
       version: main
-    release:
-      tags:
-        release: release/melodic/{package}/{version}
-      url: https://github.com/ElettraSciComp/witmotion_IMU_ros.git
-      version: 0.11.18
     source:
       type: git
       url: https://github.com/ElettraSciComp/witmotion_IMU_ros.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -15380,6 +15380,21 @@ repositories:
       url: https://github.com/clearpathrobotics/wireless.git
       version: master
     status: maintained
+  witmotion_ros:
+    doc:
+      type: git
+      url: https://github.com/ElettraSciComp/witmotion_IMU_ros.git
+      version: main
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/ElettraSciComp/witmotion_IMU_ros.git
+      version: 0.11.18
+    source:
+      type: git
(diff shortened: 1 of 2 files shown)


melodic/distribution.yaml
```diff
@@ -15380,6 +15380,21 @@ repositories:
...
+  witmotion_ros:
+    doc:
+      type: git
+      url: https://github.com/ElettraSciComp/witmotion_IMU_ros.git
+      version: main
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/ElettraSciComp/witmotion_IMU_ros.git
+      version: 0.11.18
```
clalancette, line 15392: This definitely isn't going to work; the release section here needs to be filled out by bloom.  I'll suggest you remove this, and if you want to do a release follow the instructions at http://wiki.ros.org/bloom/Tutorials/FirstTimeRelease

melodic/distribution.yaml
```diff
@@ -15380,6 +15380,21 @@ repositories:
       url: https://github.com/clearpathrobotics/wireless.git
       version: master
     status: maintained
+  witmotion_ros:
```
clalancette, line 15383: The key here is `witmotion_ros`, and the package.xml in the sources say `witmotion_ros`, but the repository name is `witmotion_IMU_ros`.  Which one is correct?  Can we make these all the same?

clalancette (changes requested):
Besides the stuff pointed out inline, please make separate PRs for Noetic and Melodic.  That's because they are on different schedules, so it might be difficult to find a time when they are both open for changes.

melodic/distribution.yaml
twdragon: The repository is maintained by Elettra so it was un upstream decision to choose Github repository name. I will ask for renaming but it will require time (unknown amount, I see). The correctname is definitely `witmotion_ros` and it will never be changed, so here we can refer to this package name.

melodic/distribution.yaml
twdragon: I will remove this section for now, then me and my colleagues consider involving bloom. Thanks for an advice! I will update the PR soon.

twdragon:
@clalancette done, please review now. The renaming request will be in consideration in a few days.

PR #35107, opened 2022-10-30
Adding rosros to noetic and foxy, and to rosdep/python

suurjaak:

## Package name:

`rosros`

## Package Upstream Source:

https://github.com/suurjaak/rosros

## Purpose of using this:

rosros is a wrapper library for providing a convenient unified interface to ROS1 / ROS2 Python API.

Distro packaging links:

- Pip:
  - https://pypi.org/project/rosros

noetic, foxy

# The source is here:

https://github.com/suurjaak/rosros

# Checks

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -5710,6 +5710,21 @@ repositories:
       url: https://github.com/uos/rospy_message_converter.git
       version: foxy
     status: maintained
+  rosros:
+    doc:
+      type: git
+      url: https://github.com/suurjaak/rosros.git
+      version: master
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/suurjaak/rosros-release.git
+      version: 0.1.0-1
+    source:
+      type: git
+      url: https://github.com/suurjaak/rosros.git
+      version: master
+    status: developed
   rosxbee:
     doc:
       type: git
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8800,6 +8800,19 @@ python3-rospkg-modules:
     pip:
       packages: [rospkg]
   ubuntu: [python3-rospkg-modules]
+python3-rosros-pip:
+  debian:
+    pip:
+      packages: [rosros]
+  fedora:
+    pip:
+      packages: [rosros]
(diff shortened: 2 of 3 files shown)


foxy/distribution.yaml
```diff
@@ -5710,6 +5710,21 @@ repositories:
       url: https://github.com/uos/rospy_message_converter.git
       version: foxy
     status: maintained
+  rosros:
```
clalancette, line 5713: I really don't think we should have a project called `rosros` in the official distribution.  It is far too generic, and unclear on what it does.  Please consider renaming this to be more descriptive of what it does.  See https://www.ros.org/reps/rep-0144.html for the rules about naming packages.

clalancette (changes requested):
Besides the name of the package, please open separate PRs for the different distributions.  It is easier to get them in that way.

foxy/distribution.yaml
suurjaak: Understandable.. I assume the package name in the ROS index does not necessarily need to match the package name in its source repository?

Would `python3-rosros` be acceptable for the ROS index?

tfoote: No, that does not resolve any of the issues with the name. Please read the naming guidelines linked above. The name should be at least unique and preferably informative about the contents. Repeating the name of the larger project name twice(which is explicitly said not to use in your package name "A package name should not contain ros as it is redundant.") as a name does not help anyone know what the package is or what it does.

rosdep/python.yaml
```diff
@@ -8800,6 +8800,19 @@ python3-rospkg-modules:
     pip:
       packages: [rospkg]
   ubuntu: [python3-rospkg-modules]
+python3-rosros-pip:
```
tfoote, line 8803: Please choose if you're going to be a rosdep or as a native ROS package.

tfoote (changes requested):

nuclearsandwich:
@suurjaak since this PR hasn't been updated recently I am going to close it. Feel free to re-open it when you have time to address the given feedback. If you open a new PR for the same project with an updated name please link back to this pull request so the discussion history is preserved.

PR #35110, opened 2022-10-31
Added Python bindings for MARISA (trie implementation)

peci1:

## Package name:

python-marisa/python3-marisa

## Package Upstream Source:

https://github.com/s-yata/marisa-trie

## Purpose of using this:

This package is an efficient implementation of tries.

Distro packaging links:

  - https://packages.debian.org/stretch/python-marisa
  - https://packages.debian.org/stretch/python3-marisa
   - https://packages.ubuntu.com/bionic/python-marisa
   - https://packages.ubuntu.com/focal/python3-marisa
  - https://packages.fedoraproject.org/pkgs/marisa/python3-marisa/
  - https://archlinux.org/packages/community/x86_64/python-marisa/
  - [IF AVAILABLE](https://packages.gentoo.org/packages/dev-libs/marisa)
  - N/A
  - N/A
  - https://search.nixos.org/packages?channel=22.05&show=marisa&from=0&size=50&sort=relevance&type=packages&query=marisa
  - https://software.opensuse.org/package/python3-marisa (only available in tumbleweed and I don't know how to correctly write that down, so I ditched this entry)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7421,7 +7421,6 @@ python3-marisa:
   debian: [python3-marisa]
   fedora: [python3-marisa]
   gentoo: [dev-libs/marisa]
-  opensuse: [python3-marisa]
   ubuntu: [python3-marisa]
 python3-markdown:
   debian: [python3-markdown]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2700,6 +2700,14 @@ python-mako:
 python-mapnik:
   debian: [python-mapnik]
   ubuntu: [python-mapnik]
+python-marisa:
+  debian:
+    '*': null
+    buster: [python-marisa]
+    stretch: [python-marisa]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2700,6 +2700,14 @@ python-mako:
 python-mapnik:
   debian: [python-mapnik]
   ubuntu: [python-mapnik]
+python-marisa:
```
tfoote, line 2703: Do you need a python2 version? If not we'd rather not clutter up the rules with an unused key.

tfoote (changes requested):

rosdep/python.yaml
peci1: Yes, we still have systems running Melodic

tfoote (approved):

sloretz (approved):

PR #35193, opened 2022-11-04
add gentoo package for libatomic

tomkimsour:

## Package name:

libatomic

## Package Upstream Source:

https://gcc.gnu.org/

## Purpose of using this:

In order to compile ros-galactic for gentoo with superflore, we need to add the gentoo package for libatomic.

Distro packaging links:

  - ALREADY AVAILABLE
   - ALREADY AVAILABLE
  - https://packages.gentoo.org/packages/dev-libs/libatomic_ops

Most of the packages have already been added in #29030
Related to https://github.com/ros-infrastructure/superflore/issues/292

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2441,7 +2441,7 @@ libatomic:
   arch: [gcc-libs]
   debian: [libatomic1]
   fedora: [libatomic]
-  gentoo: [dev-libs/libatomic_ops]
+  gentoo: [sys-devel/gcc]
   nixos: []
   openembedded: [gcc-runtime@openembedded-core]
   rhel: [libatomic]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2441,6 +2441,7 @@ libatomic:
   arch: [gcc-libs]
   debian: [libatomic1]
   fedora: [libatomic]
+  gentoo: [dev-libs/libatomic_ops]
   nixos: []
   openembedded: [gcc-runtime@openembedded-core]
   rhel: [libatomic]


tfoote (approved):

rosdep/base.yaml
```diff
@@ -2441,6 +2441,7 @@ libatomic:
   arch: [gcc-libs]
   debian: [libatomic1]
   fedora: [libatomic]
+  gentoo: [dev-libs/libatomic_ops]
```
sloretz, line 2444: I'm not very familiar with Gentoo, but I'm not sure `libatomic_ops` is the right package. The [Ubuntu package is only providing `libatomic.so.1` and `libatomic.so.1.2.0`](https://packages.ubuntu.com/jammy/amd64/libatomic1/filelist).

In the gentoo `stage3` docker container I see those libs are provided by `sys-devel/gcc`.

```console
gentoo:root> find / -name "libatomic.so*" 2>/dev/null
/usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so
/usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so.1.2.0
/usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so.1
/usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/libatomic.so
/usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/libatomic.so.1.2.0
/usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/libatomic.so.1
gentoo:root> qfile /usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so
sys-devel/gcc: /usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so
gentoo:root> qfile /usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so.1.2.0 
sys-devel/gcc: /usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so.1.2.0
gentoo:root> qfile /usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so.1
sys-devel/gcc: /usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/32/libatomic.so.1

```

sloretz (changes requested):

rosdep/base.yaml
tomkimsour: You are right it works if we use this dependency instead.
However, the tool superflore while fetching for every dependancy, would look for the name "libatomic" and then take the argument of gentoo in the yaml file.
Would that be alright if I replace libatomic_ops by sys-devel/gcc ?

sloretz (approved):

tfoote (approved):

PR #35203, opened 2022-11-06
euslime: 1.1.4-4 in 'melodic/distribution.yaml' [bloom]

Affonso-Gui:
Increasing version of package(s) in repository `euslime` to `1.1.4-4`:

- upstream repository: https://github.com/jsk-ros-pkg/euslime.git
- release repository: https://github.com/jsk-ros-pkg/euslime-release.git
- bloom version: `0.11.1`
- previous version for package: `1.1.2-1`

## euslime

```
* Fix build
* Contributors: Guilherme Affonso
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2984,7 +2984,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/jsk-ros-pkg/euslime-release.git
-      version: 1.1.2-1
+      version: 1.1.4-4
     source:
       type: git
       url: https://github.com/jsk-ros-pkg/euslime.git


Affonso-Gui:
I am hoping that this fixes the build.

All pre-release tests (http://prerelease.ros.org/) run fine, but I haven't been able to replicate the binary builds locally.

tfoote:
I'll retry it, but as mentioned in my noetic release review. Retagging is strongly recommended against if you change the content in the upstream repository.

tfoote:
I don't think it's necessary in this case. But I would just strongly recommend it for future reference. We've enabled force rebuilds on the buildfarm to work around I believe all internal collisions. So the only ones who might be effected would be people using the source debs on the testing repo. So very limited impact.

PR #35236, opened 2022-11-07
Only run on pushes to master _or_ pull requests.

nuclearsandwich:
The current configuration is causing builds to be run twice, duplicating the results and creating confusion (in addition to burning CPU time needlessly).


tfoote (approved):
lgtm

PR #35438, opened 2022-11-28
Add geojson and kml2geojson pip packages

igor-markovic:

## Package name: `geojson`

python3-geojson

## Package Upstream Source:

https://github.com/jazzband/geojson

## Purpose of using this:

This dependency provides python functions for encoding and decoding [GeoJSON](http://geojson.org/) formatted data

- Arch: https://archlinux.org/packages/community/any/python-geojson/
- Debian: https://packages.debian.org/bullseye/python3-geojson
- Fedora: https://packages.fedoraproject.org/pkgs/python-geojson/python3-geojson/
- openSUSE: https://software.opensuse.org/package/python3-geojson?search_term=python3-geojson
- Ubuntu: https://packages.ubuntu.com/focal/python3-geojson

---

## Package name: `kml2geojson`

pip/pip3 kml2geojson

## Package Upstream Source:

https://gitlab.com/mrcagney/kml2geojson

## Purpose of using this:

This dependency is a Python library to covert KML files to GeoJSON files

- Pip: https://pypi.org/project/kml2geojson/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6990,6 +6990,16 @@ python3-geographiclib:
     '*': [python3-GeographicLib]
     '7': null
   ubuntu: [python3-geographiclib]
+python3-geojson-pip:
+  debian:
+    pip:
+      packages: [geojson]
+  fedora:
+    pip:
+      packages: [geojson]
+  ubuntu:
+    pip:
+      packages: [geojson]
 python3-geomag-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6990,6 +6990,16 @@ python3-geographiclib:
...
+python3-geojson-pip:
+  debian:
+    pip:
+      packages: [geojson]
+  fedora:
+    pip:
+      packages: [geojson]
+  ubuntu:
+    pip:
+      packages: [geojson]
```
clalancette, line 7002: So it looks like geojson is actually available as system packages for all of these distributions:

* Debian: https://packages.debian.org/search?keywords=python3-geojson&searchon=names&suite=all&section=all
* Ubuntu: https://packages.ubuntu.com/search?keywords=python3-geojson&searchon=names&suite=all&section=all
* Fedora: https://packages.fedoraproject.org/pkgs/python-geojson/python3-geojson/

Please update the key to be a regular one (`python3-geojson`), and have it reference the system packages.

clalancette (changes requested):

rosdep/python.yaml
igor-markovic: Okay, I've updated the key to reference the system packages. I also found this package on opensuse and arch. Can you take another look? Thanks

clalancette (approved):

nuclearsandwich (approved):

PR #35449, opened 2022-11-28
simple_term_menu: 1.5.2-1 in 'humble/distribution.yaml' [bloom]

roni-kreinin:
Increasing version of package(s) in repository `simple_term_menu` to `1.5.2-1`:

- upstream repository: https://github.com/clearpathrobotics/simple-term-menu.git
- release repository: https://github.com/clearpath-gbp/simple_term_menu-release.git
- bloom version: `0.11.2`

## simple_term_menu

```
* Initial ROS package
* Contributors: Roni Kreinin
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5706,6 +5706,21 @@ repositories:
       url: https://github.com/oKermorgant/simple_launch.git
       version: devel
     status: maintained
+  simple_term_menu:
+    doc:
+      type: git
+      url: https://github.com/clearpathrobotics/simple-term-menu.git
+      version: humble
+    release:
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/clearpath-gbp/simple_term_menu-release.git
+      version: 1.5.2-1
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/simple-term-menu.git
+      version: humble
+    status: developed
   slam_toolbox:
     doc:
       type: git


clalancette (commented):
In general, we don't want to be packaging the world.  While we definitely have vendor packages elsewhere, they are usually used by other ROS packages.  So before we go ahead with this, what is the goal for this package?

roni-kreinin:
I am planning on using this package as a dependency for a turtlebot4 setup tool. The package is currently only available through pip but I would like to release the setup tool as a debian, so I need this package released too.

nuclearsandwich:
To echo and extend what @clalancette said above, vendoring software that is not part of ROS should only be done when absolutely necessary and ideally only as a temporary measure pending platform support from ROS 2's upstream platforms.

1. Is there no software either in the Python standard library or packages currently available in Ubuntu Jammy and RHEL 8 which provides comparable functionality? That is to say, is this dependency _strictly_ necessary or is there an alternative that is already packaged.

3. Is the vendored software stable enough that it would be considered for a release into Debian? If not, it probably isn't suitable as a ROS dependency either. If it is, we should probably make sure that it's a requested package in Debian so that we can remove the vendored version in a future release of ROS 2 when the package has become available.

We also do not have a standard workflow for vendoring python libraries in ROS 2. I'm not opposed to wrapping it using ament_cmake_python, in some ways it has advantages to vendoring a "plain python" package. If we are going to proceed with this package, I think that the `_vendor` suffix should be added to the name of the package in package.xml.

roni-kreinin:
I will bloom again with the `_vendor` suffix.

PR #35453, opened 2022-11-29
Add intel-opencl-icd

5730289021-NN:
## Purpose of using this:

It's an [Intel's OpenCL Driver](https://github.com/intel/compute-runtime) which enable iGPU on Intel CPU. Useful when using with [OpenVINO](https://docs.openvino.ai/latest/index.html) aka. Intel's Neural Network Inference Engine.

  - https://packages.debian.org/sid/intel-opencl-icd
   - https://packages.ubuntu.com/focal/intel-opencl-icd
  - NOT AVAILABLE
  - NOT AVAILABLE
  - NOT AVAILABLE
  - NOT AVAILABLE
  - NOT AVAILABLE
  - NOT AVAILABLE
  - NOT AVAILABLE

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2197,7 +2197,9 @@ imagemagick:
   nixos: [imagemagick]
   ubuntu: [imagemagick]
 intel-opencl-icd:
-  debian: [intel-opencl-icd]
+  debian:
+    '*': [intel-opencl-icd]
+    buster: null
   ubuntu:
     '*': [intel-opencl-icd]
     bionic: null
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2196,6 +2196,9 @@ imagemagick:
   gentoo: [virtual/imagemagick-tools]
   nixos: [imagemagick]
   ubuntu: [imagemagick]
+intel-opencl-icd:
+  debian: [intel-opencl-icd]
+  ubuntu: [intel-opencl-icd]
 intltool:
   arch: [intltool]
   debian: [intltool]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -2196,6 +2196,9 @@ imagemagick:
   gentoo: [virtual/imagemagick-tools]
   nixos: [imagemagick]
   ubuntu: [imagemagick]
+intel-opencl-icd:
+  debian: [intel-opencl-icd]
```
nuclearsandwich, line 2200: This package is not available in Debian Buster so the distribution version cannot be elided.

```suggestion
  debian:
    '*': [intel-opencl-icd]
    buster: null
```

rosdep/base.yaml
```diff
@@ -2196,6 +2196,9 @@ imagemagick:
...
+  ubuntu: [intel-opencl-icd]
```
nuclearsandwich, line 2201: This package is not available in Ubuntu Bionic so the distribution version cannot be elided.

```suggestion
  ubuntu:
    '*': [intel-opencl-icd]
    bionic: null
```

nuclearsandwich (changes requested):
These packages aren't available in all Debian and Ubuntu distributions that we currently support via rosdep.

rosdep/base.yaml
5730289021-NN: Hadn't checked that thoroughly, thanks for the fix

nuclearsandwich (approved):
Thanks for making the requested changes.

audrow (approved):

audrow:
Looks good to me, but I'll wait to merge for someone from @ros/rosdeputies to approve.

sloretz (approved):

PR #35491, opened 2022-12-01
python3-deprecated added to rosdep dependencies

gv11181989:

## Package name:  
Deprecated 

## Package Upstream Source: 
https://github.com/tantale/deprecated

## Purpose of using this:
This dependency is Python @deprecated decorator to deprecate old python classes, functions, or methods

- Debian: https://packages.debian.org/bullseye/python3-deprecated
- Ubuntu: https://packages.ubuntu.com/jammy/python3-deprecated

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -169,13 +169,6 @@ cython3:
   gentoo: [dev-python/cython]
   nixos: [python3Packages.cython]
   ubuntu: [cython3]
-python3-deprecated:
-  debian:
-    '*': [python3-deprecated]
-    buster: null
-  ubuntu:
-    '*': [python3-deprecated]
-    bionic: null
 dpath-pip:
   ubuntu:
     pip:
@@ -6479,6 +6472,13 @@ python3-defusedxml:
   opensuse: [python3-defusedxml]
   rhel: ['python%{python3_pkgversion}-defusedxml']
   ubuntu: [python3-defusedxml]
+python3-deprecated:
+  debian:
+    '*': [python3-deprecated]
+    buster: null
+  ubuntu:
+    '*': [python3-deprecated]
+    bionic: null
 python3-depthai-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -169,13 +169,13 @@ cython3:
   gentoo: [dev-python/cython]
   nixos: [python3Packages.cython]
   ubuntu: [cython3]
-deprecated-pip:
+python3-deprecated:
   debian:
-    pip:
-      packages: [Deprecated]
+    '*': [python3-deprecated]
+    buster: null
   ubuntu:
-    pip:
-      packages: [Deprecated]
+    '*': [python3-deprecated]
+    bionic: null
 dpath-pip:
   ubuntu:
     pip:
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
There are several changes that need to be made.

This entry does not follow our conventions for pip keys, which should be of the form `python3-deprecated-pip` rather than `deprecated-pip` regardless of which major python versions are supported by the package.

Furthermore, you've linked to the debian and ubuntu repository listings for this package which means that pip keys should _not_ be used and instead keys for the system packages should be added instead. For example:
```yaml
python3-deprecated:
  debian:
    '*': [python3-deprecated]
    buster: null
  ubuntu:
    '*': [python3-deprecated]
    bionic: null
```

gv11181989:
> There are several changes that need to be made.
> 
> This entry does not follow our conventions for pip keys, which should be of the form `python3-deprecated-pip` rather than `deprecated-pip` regardless of which major python versions are supported by the package.
> 
> Furthermore, you've linked to the debian and ubuntu repository listings for this package which means that pip keys should _not_ be used and instead keys for the system packages should be added instead. For example:
> 
> ```yaml
> python3-deprecated:
>   debian:
>     '*': [python3-deprecated]
>     buster: null
>   ubuntu:
>     '*': [python3-deprecated]
>     bionic: null
> ```

Changes made as requested.

cottsay (changes requested):
The tests failed with the following error:
```
ERR: list out of alphabetical order line 179.  'dpath-pip' should come before 'python3-deprecated'
```

Additionally, there are two suggestions on the [Files changed](https://github.com/ros/rosdistro/pull/35491/files) tab of this PR that appear to be valid.

sloretz (approved):

clalancette (changes requested):
There are still 2 suggestions on the Files tab: https://github.com/ros/rosdistro/pull/35491/files

Once those are fixed, I think this can go in.

rosdep/python.yaml
```diff
@@ -6472,6 +6472,15 @@ python3-defusedxml:
   opensuse: [python3-defusedxml]
   rhel: ['python%{python3_pkgversion}-defusedxml']
   ubuntu: [python3-defusedxml]
+python3-deprecated:
+  debian:
+    '*': [python3-deprecated]
+    buster: null
+  fedora: [python3-deprecated]
+  rhel: [python3-deprecated]
```
clalancette, line 6480:
```suggestion
  rhel: 
    '*': [python3-deprecated]
    '7': null
```

clalancette (changes requested):

rosdep/python.yaml
```diff
@@ -6472,6 +6472,17 @@ python3-defusedxml:
   opensuse: [python3-defusedxml]
   rhel: ['python%{python3_pkgversion}-defusedxml']
   ubuntu: [python3-defusedxml]
+python3-deprecated:
+  debian:
+    '*': [python3-deprecated]
+    buster: null
+  fedora: [python3-deprecated]
+  rhel:
```
clalancette, line 6480: Looks like there is trailing whitespace on this line that is causing CI to fail.

clalancette (changes requested):

clalancette (approved):

PR #35497, opened 2022-12-01
simple_term_menu_vendor: 1.5.4-1 in 'humble/distribution.yaml' [bloom]

roni-kreinin:
Increasing version of package(s) in repository `simple_term_menu_vendor` to `1.5.4-1`:

- upstream repository: https://github.com/clearpathrobotics/simple-term-menu.git
- release repository: https://github.com/clearpath-gbp/simple_term_menu_vendor-release.git
- bloom version: `0.11.2`

## simple_term_menu_vendor

```
* Renamed folder to simple_term_menu_vendor
* Contributors: Roni Kreinin
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5689,6 +5689,21 @@ repositories:
       url: https://github.com/oKermorgant/simple_launch.git
       version: devel
     status: maintained
+  simple_term_menu_vendor:
+    doc:
+      type: git
+      url: https://github.com/clearpathrobotics/simple-term-menu.git
+      version: humble
+    release:
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/clearpath-gbp/simple_term_menu_vendor-release.git
+      version: 1.5.4-1
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/simple-term-menu.git
+      version: humble
+    status: developed
   slam_toolbox:
     doc:
       type: git


nuclearsandwich:
This is connected to https://github.com/ros/rosdistro/pull/35449

@roni-kreinin could you please speak to the other concerns raised:

> To echo and extend what @clalancette said above, vendoring software that is not part of ROS should only be done when absolutely necessary and ideally only as a temporary measure pending platform support from ROS 2's upstream platforms.

> 1. Is there no software either in the Python standard library or packages currently available in Ubuntu Jammy and RHEL 8 which provides comparable functionality? That is to say, is this dependency _strictly_ necessary or is there an alternative that is already packaged.

> 2. Is the vendored software stable enough that it would be considered for a release into Debian? If not, it probably isn't suitable as a ROS dependency either. If it is, we should probably make sure that it's a requested package in Debian so that we can remove the vendored version in a future release of ROS 2 when the package has become available.

nuclearsandwich (changes requested):
Awaiting feedback on acceptability for vendoring this library.

roni-kreinin:
@nuclearsandwich 

1.  I have not found any alternatives that are available on Debian. Most are also only available on pip.
2.  The package has had stable releases for a couple years now so I don't think stability is an issue. I can look into requesting it be released in Debian.

hidmic:
> I can look into requesting it be released in Debian.

That'd be ideal @roni-kreinin.

roni-kreinin:
I would still like to have this released as a ROS package in the meantime. The Debian release may take a while.

nuclearsandwich:
@roni-kreinin thanks for your patience. Vendoring software within ROS is something that we try to avoid wherever possible by keeping the software in ROS distributions limited to software written specifically for the ROS ecosystem and pushing other, more general dependencies, down to our supported platforms and using them via system packages.

With the information you've provided it sounds like there isn't an alternative already available. I would really appreciate if you'd consider filing package request tickets with Debian and Fedora/EPEL which both feed into our supported platforms. Once packaged versions are available we can begin transitioning away from this vendored version.

nuclearsandwich (approved):

PR #35555, opened 2022-12-04
casadi: 3.5.5-1 in 'noetic/distribution.yaml' [bloom]

yakunouyang:
Increasing version of package(s) in repository `casadi` to `3.5.5-1`:

- upstream repository: https://github.com/casadi/casadi.git
- release repository: https://github.com/yakunouyang/casadi-release.git
- bloom version: `0.11.2`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -713,6 +713,12 @@ repositories:
       type: git
       url: https://github.com/PickNikRobotics/cartesian_msgs.git
       version: jade-devel
+  casadi:
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/yakunouyang/casadi-release.git
+      version: 3.5.5-3
   catch_ros:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -718,7 +718,7 @@ repositories:
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/yakunouyang/casadi-release.git
-      version: 3.5.5-3
+      version: 3.5.5-1
   catch_ros:
     doc:
       type: git


nuclearsandwich:
New package review:

  - [x] Top level license file: https://github.com/casadi/casadi/blob/master/LICENSE.txt
  - [ ] Per package license files
- [x] License is [OSI-approved](https://opensource.org/licenses): LGPL-3.0
  - [x] casadi (release repo): https://github.com/yakunouyang/casadi-release/blob/master/noetic/package.xml#L11
- [x] Public source repo: https://github.com/casadi/casadi
- [ ] Source repository contains ROS packages and each package meets [REP-144](https://www.ros.org/reps/rep-0144.html) naming conventions
  - The packaged software is not a native ROS package and has a package.xml included in the release repository.
  - Foxy and Dashing have a casadi_vendor package: https://github.com/ros/rosdistro/blob/b5fef677134a551953178b7d9290bafb3d4ef884/foxy/distribution.yaml#L540-L554

sloretz (changes requested):
First, [I don't see an upstream system package in Focal](https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=casadi)

Next, checking if consistent with recommendations in [REP 136](https://ros.org/reps/rep-0136.html)

* Is a package.xml injected via bloom? 
   * Yes there is a [patches directory in the Noetic track](https://github.com/yakunouyang/casadi-release/blob/5cf05a0b97cd04e02ec77fd9c6d8fad7a62a118c/tracks.yaml#L20), and it does [have a package.xml](https://github.com/yakunouyang/casadi-release/blob/master/noetic/package.xml)
* Is there an install rule patch that installs the package.xml
   * [Yes](https://github.com/yakunouyang/casadi-release/blob/patches/release/noetic/casadi/0001-Added-install-rule-for-package.xml.patch)
* Is there an exec_depend or run_depend on catkin?
   * **No**
* Is the build_type set appropriately?
   * Yes, [build_type is set to cmake, and it's a cmake packge](https://github.com/yakunouyang/casadi-release/blob/5cf05a0b97cd04e02ec77fd9c6d8fad7a62a118c/noetic/package.xml#L22)  


If I understand correctly, I think it needs an `<exec_depend>catkin</exec_depend>` in the patched in package.xml, and then a re-bloom.

yakunouyang:
Thank you for your thorough review! I've added `<exec_depend>catkin</exec_depend>` to `noetic/package.xml` according to your comment.

nuclearsandwich:
> I noticed that the release inc was set to 3 before because of several incautious `bloom-release` run, I've reset the release inc back to 1 and re-bloomed the release repository, hopefully this won't cause any problems. smile

I would recommend not modifying that value at all. Most packages released via bloom that do anything other than very standard ROS releases will require a few revisions and package revisions are a normal part of the workflow. You cannot change that value without changing the contents of the release which means that fixes made since the first revision will no longer be present.

If you make a follow up release, please preserve the release increment as-is. Failing to do so will likely create broken packages.

yakunouyang:
> > I noticed that the release inc was set to 3 before because of several incautious `bloom-release` run, I've reset the release inc back to 1 and re-bloomed the release repository, hopefully this won't cause any problems. smile
> 
> I would recommend not modifying that value at all. Most packages released via bloom that do anything other than very standard ROS releases will require a few revisions and package revisions are a normal part of the workflow. You cannot change that value without changing the contents of the release which means that fixes made since the first revision will no longer be present.
> 
> If you make a follow up release, please preserve the release increment as-is. Failing to do so will likely create broken packages.

Will do, thank you for your kind reminder.

sloretz (approved):
Tag ending in deb inc of `1` seems to have the exec_depend on `catkin` now, so this looks good to me if it looks ok to @nuclearsandwich too

https://github.com/yakunouyang/casadi-release/blob/ros-noetic-casadi_3.5.5-1_focal/package.xml#L13

nuclearsandwich:
> Tag ending in deb inc of `1` seems to have the exec_depend on `catkin` now, so this looks good to me if it looks ok to @nuclearsandwich too
> 
> https://github.com/yakunouyang/casadi-release/blob/ros-noetic-casadi_3.5.5-1_focal/package.xml#L13

The above link has an apparent typo but likely refers to https://github.com/yakunouyang/casadi-release/blob/debian/ros-noetic-casadi_3.5.5-1_focal/package.xml#L13 ([permalink](https://github.com/yakunouyang/casadi-release/blob/657ca77de0c66cb58dc0cc95e516208ee985d3c4/package.xml#L13))

I agree that it looks alright there but I think that it would be best for @yakunouyang to close this PR and run a full bloom-release without changing the release increment to make sure that bloom is configured correctly for this repository.

PR #35595, opened 2022-12-09
rename ros2 to rolling branch in image_transport_plugins

ijnek:
The ros2 branch has been renamed to rolling, to match core-package recommendations.


hidmic (approved):
@ijnek LGTM but would you mind adding a BSD license file at the root of the repo?

PR #35597, opened 2022-12-09
Add transforms3d for Jammy

vatanaksoytezer:


hidmic (approved):
OK but I will point out that `transforms3d` has been released to Jammy.

audrow (approved):

PR #35600, opened 2022-12-10
Add ldrobot-lidar-ros2 entry

Myzhar:

## Package name:

TODO

## Package Upstream Source:

## Purpose of using this:

Distro packaging links:

  - REQUIRED
   - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE

ROSDISTRO NAME

# The source is here:

http://sourcecode_url

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2161,6 +2161,25 @@ repositories:
       url: https://github.com/ros2/launch_ros.git
       version: humble
     status: maintained
+  ldrobot-lidar-ros2:
+    doc:
+      type: git
+      url: https://github.com/Myzhar/ldrobot-lidar-ros2.git
+      version: humble
+    release:
+      packages:
+      - ldlidar
+      - ldlidar_node
+      - ldlidar_component
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/ros2-gbp/ldrobot-lidar-ros2.git
+      version: 0.2.0
+    source:
+      type: git
+      url: https://github.com/Myzhar/ldrobot-lidar-ros2.git
+      version: humble
+    status: maintained    
   leo_common:
     doc:
       type: git


humble/distribution.yaml
```diff
@@ -2161,6 +2161,25 @@ repositories:
...
+      - ldlidar_node
+      - ldlidar_component
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/ros2-gbp/ldrobot-lidar-ros2.git
+      version: 0.2.0
+    source:
+      type: git
+      url: https://github.com/Myzhar/ldrobot-lidar-ros2.git
+      version: humble
```
hidmic, line 2181: @Myzhar the `humble` branch doesn't seem to exist in the source (and doc) repository.

humble/distribution.yaml
```diff
@@ -2161,6 +2161,25 @@ repositories:
...
       version: humble
     status: maintained
+  ldrobot-lidar-ros2:
+    doc:
+      type: git
+      url: https://github.com/Myzhar/ldrobot-lidar-ros2.git
+      version: humble
+    release:
+      packages:
+      - ldlidar
```
hidmic, line 2171: @Myzhar hmm, I see an MIT license mentioned in this package while everything points to Apache License, Version 2.0. Which one is it? Is it mixed?

hidmic (changes requested):
@Myzhar was this `bloom` released? We strongly encourage [package releases](https://docs.ros.org/en/humble/How-To-Guides/Releasing/Releasing-a-Package.html) to be done using `bloom` and not manually.

humble/distribution.yaml
Myzhar: Adding it

humble/distribution.yaml
Myzhar: It was initially MIT, but then I changed to Apache-2. I search for remaining mentions to MIT

humble/distribution.yaml
Myzhar: Fixed

humble/distribution.yaml
Myzhar: Fixed

hidmic:
@Myzhar see https://docs.ros.org/en/humble/How-To-Guides/Releasing/First-Time-Release.html

clalancette:
Since this was not created with bloom, it almost certainly will not work.  I'll encourage you to follow the full instructions at https://docs.ros.org/en/humble/How-To-Guides/Releasing/First-Time-Release.html , and then let bloom open a new PR.

With that said, I'm going to close this out.

nuclearsandwich:
@Myzhar thanks for your efforts with the process thus far, there are definitely some rough edges in the initial setup that we're working to both better document and smooth over.

For this first PR, which you correctly created manually, we should leave off the`release` section entirely. That will be added by bloom down the line after the release repository is configured and bloom can be used to create release data.

I would suggest updating this PR (or creating a new one and linking this discussion for context) which only adds the `source` and `doc` sections for your repository. Once that is merged your package will be "indexed" and will have also undergone the naming review required for the release repository setup in https://github.com/ros2-gbp/ros2-gbp-github-org/issues/159

Let me know if you have any other questions.

PR #35657, opened 2022-12-15
Add python3-transformers-pip

greinet:

## Package name:

transformers

## Package Upstream Source:

https://pypi.org/project/transformers/

## Purpose of using this:

Transformers provides thousands of models for various machine-learning tasks like text, image or audio. It is backed by Jax, PyTorch and Tensorflow and provides seamless integration.

Distro packaging links:

- Debian: https://pypi.org/project/transformers/
- Ubuntu: https://pypi.org/project/transformers/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5336,13 +5336,6 @@ python-transforms3d-pip:
     jammy:
       pip:
         packages: [transforms3d]
-python3-transformers-pip:
-  debian:
-    pip:
-      packages: [transformers]
-  ubuntu:
-    pip:
-      packages: [transformers]
 python-transitions:
   debian:
     '*': [python-transitions]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5321,6 +5321,13 @@ python-tqdm:
     xenial:
       pip:
         packages: [tqdm]
+python3-transformers-pip:
```
clalancette, line 5324: This is failing CI because this needs to be after the `python3-transforms3d-pip` key.

clalancette (changes requested):

rosdep/python.yaml
greinet: Oh sorry, fixed it.

clalancette (changes requested):
Ah sorry, I got this wrong.  The key is in the complete wrong part of the file; because of the `python3` prefix, it needs to be among the `python3` keys, not the `python` keys.  If you move it, that should hopefully fix it.

greinet:
I added the 3 to the prefix after creating the PR and didn't think about it, now it should be in the right spot I hope.

clalancette (approved):

methylDragon (approved):
Exciting!

PR #35667, opened 2022-12-19
[noetic] Move rospy_message_converter to DFKI-NI

mintar:


sloretz (approved):
Old URL redirects to new one, so this LGTM

PR #35695, opened 2022-12-22
Split off humble branch for zbar_ros

ijnek:
Splits off [humble branch](https://github.com/ros-drivers/zbar_ros/tree/humble) due to an API breaking change in an underlying library.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -7071,7 +7071,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-drivers/zbar_ros.git
-      version: ros2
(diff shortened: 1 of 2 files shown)


clalancette (approved):
This is fine, but you are also going to want to update https://github.com/ros2-gbp/zbar_ros-release/blob/23a89fe8f76497267c5c82a5b7165f3e39c314c8/tracks.yaml#L17 to point to the `humble` branch.  Otherwise, the next time you go to do a release it won't do what you expect.

ijnek:
Thanks @clalancette , I'll get that done.

PR #35707, opened 2022-12-28
Updated package upstream sources for depthai 

saching13:

## Package name:

TODO

## Package Upstream Source:

## Purpose of using this:

Distro packaging links:

  - REQUIRED
   - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE

ROSDISTRO NAME

# The source is here:

http://sourcecode_url

# Checks

### DIFF
--- a/galactic/distribution.yaml
+++ b/galactic/distribution.yaml
@@ -741,14 +741,13 @@ repositories:
     doc:
       type: git
       url: https://github.com/luxonis/depthai-ros.git
-      version: galactic
+      version: ros-release
     release:
       packages:
       - depthai-ros
       - depthai_bridge
       - depthai_examples
       - depthai_ros_msgs
-      - depthai_ros_driver
       tags:
         release: release/galactic/{package}/{version}
       url: https://github.com/luxonis/depthai-ros-release.git
@@ -757,7 +756,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/luxonis/depthai-ros.git
-      version: galactic
+      version: ros-release
     status: developed
   depthimage_to_laserscan:
     doc:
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):
Galactic is now End-of-Life, so we won't be taking new changes to it.  Please remove that change from this PR (which will also fix the failing CI).

saching13:
Ah got it. Will do.

saching13:
@clalancette any idea why CI still failing ?

clalancette:
> @clalancette any idea why CI still failing ?

CI says: `RuntimeError: humble: missing package.xml file for package "depthai_ros_driver"`.

Indeed, if you look at https://github.com/luxonis/depthai-ros-release/blob/master/README.md , you can see that in the latest humble release, `depthai_ros_driver` is not listed there.  It looks like the bloom for this was not correct.  In particular, my guess is that this line is wrong: https://github.com/luxonis/depthai-ros-release/blob/51567a2d515ac3dbe87fb0bfc49b9e014a71623e/tracks.yaml#L69 .  You should probably change that to `humble`, and then do a new bloom release.

Given all of that, I'm going to close this one out.

saching13:
> https://github.com/luxonis/depthai-ros-release/blob/master/README.md

@clalancette  Please correct me if I'm wrong here. 
The depthai_ros_driver  is the new package. and it gets listed there when I make a new release of it. But to make a release for it. Don't I need to add it to the distro yaml here first to be able to picked up in the build.ros.org ?

clalancette:
> The depthai_ros_driver is the new package. and it gets listed there when I make a new release of it. But to make a release for it. Don't I need to add it to the distro yaml here first to be able to picked up in the build.ros.org ?

No, it is the other way around.  When you use `bloom-release`, it updates the release repository (in this case, https://github.com/luxonis/depthai-ros-release), and then opens a PR to rosdistro here.

PR #35733, opened 2023-01-03
Change doc and source branch of image_transport_plugins in melodic

ijnek:
A [melodic branch](https://github.com/ros-perception/image_transport_plugins/tree/melodic-devel) has been split off in image_transport_plugins due to a backported feature which is untested in previous eol distros.


ijnek:
Thanks @methylDragon !

PR #35758, opened 2023-01-05
behaviortree_cpp: 3.8.2-1 in 'foxy/distribution.yaml' [manual]

facontidavide:
Increasing version of package(s) in repository behaviortree_cpp to 3.8.2-1:

Note that this PR was done by hand instead of blook, because I got this error: 

`Failed to open pull request: AssertionError - Duplicate package name 'behaviortree_cpp_v3' exists in repository 'behaviortree_cpp_v3' as well as in repository 'behaviortree_cpp'`

This changes should fix the mess

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -442,8 +442,6 @@ repositories:
       url: https://github.com/BehaviorTree/BehaviorTree.CPP.git
       version: v3.8
     release:
-      packages:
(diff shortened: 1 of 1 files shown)


methylDragon:
Nosetests are failing. I think you can try deleting the package list, since the package.xml is in the repository root

methylDragon (dismissed):
As above

PR #35767, opened 2023-01-05
Migrate release repos into ros2-gbp

tylerjw:
Per this change to migrate release repos into the ros2-gbp org, here is a change to update the entries to point to the new home of the release repos: https://github.com/ros2-gbp/ros2-gbp-github-org/issues/163


methylDragon (approved):
Holding for sync: https://discourse.ros.org/t/preparing-for-rolling-sync-2023-01-06/29052

clalancette (approved):

PR #35779, opened 2023-01-07
Grace_Proj_Msgs: 0.0.1-4 in 'noetic/distribution.yaml' [bloom]

TheDoctor-JI:
Increasing version of package(s) in repository `Grace_Proj_Msgs` to `0.0.1-4`:

- upstream repository: https://github.com/HKUST-NISL/Grace_Proj_Msgs.git
- release repository: https://github.com/HKUST-NISL/Grace_Proj_Msgs.git
- bloom version: `0.11.2`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -8,6 +8,14 @@ release_platforms:
   ubuntu:
   - focal
 repositories:
+  Grace_Proj_Msgs:
+    release:
+      packages:
+      - grace_attn_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/HKUST-NISL/Grace_Proj_Msgs.git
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
This seems to be using the same repository for both the sources and for the releases.  That should never be the case; there should be a separate repository for the sources and for the release.

Further, it is best (though not strictly required) if the name of the source repository matches the name of the package.  In this case, that means that the repository should be something like https://github.com/HKUST-NISL/grace_attn_msgs , and then a separate release repository something like https://github.com/HKUST-NISL/grace_attn_msgs-release .

Finally, the sources are missing a LICENSE file and a license specified in the package.xml

For all of those reasons, I'm going to close this out.  Once those issues are all fixed, please feel free to resubmit a PR.

PR #35800, opened 2023-01-10
behaviortree_cpp: 3.8.2-1 in 'rolling/distribution.yaml' [manual]

facontidavide:
Increasing version of package(s) in repository behaviortree_cpp to 3.8.2-1

Done manually to fix other issues

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -483,22 +483,20 @@ repositories:
       url: https://github.com/wep21/bag2_to_image.git
       version: main
     status: maintained
-  behaviortree_cpp:
+  behaviortree_cpp_v3:
     doc:
       type: git
       url: https://github.com/BehaviorTree/BehaviorTree.CPP.git
-      version: master
+      version: v3.8
     release:
-      packages:
-      - behaviortree_cpp_v3
       tags:
         release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/behaviortree_cpp-release.git
-      version: 3.8.0-1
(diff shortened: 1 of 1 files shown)


facontidavide:
Does that repo even exist? I can't find it

tfoote (changes requested):
It would be much better to use the ros2-gbp in general, we generally insist for rolling releases.  Otherwise as Chris mentions it will get dropped from rolling the next time that we migrate platforms or fork a new release (Iron is pending soon)

facontidavide:
ok, let me fix that

tfoote:
I wouldn't recommend making a new one for v3 and then continuing to use the default for v4. The contents of the repository are already there with the full history of v3 so making a new one for the latest release will just be confusing as all the old releases will still be in the original repository. And moving the existing repo isn't recommended as it will break any cache dreferences. 

So to that end I'd recommend just using the same release repository and then everything will be in one place. That repo already has the full history of v2 and v3 releases now it will have v4 as well. 

My alternative to that would be to make a new release repo `_v4` which would have the new series. But unless we're running out of storage or anything like that I'd suggest just sticking to the single repository. 

For reference to get a new repository you can request that yourself by opening a PR to extend this list: https://github.com/ros2-gbp/ros2-gbp-github-org/blob/bd80f85c71856bee7499b5e4838f71e34097ec0b/behaviortree.tf#L6

But again I'd strongly suggest just leveraging the existing repository which already has the history until it gets too large and then we just leave it for archival purposes and make a new one.

tfoote (approved):
Thanks for the adjustment in plan. I've pinged Steven about the new repo creation.

PR #35829, opened 2023-01-11
Remove rosbag2_storage_mcap

audrow:
I'm removing this package because I'm not able to release ros2bag because both of these repositories contain a `mcap_vendor` package. Here's the exact error from Bloom:

> Failed to open pull request: AssertionError - Duplicate package name 'mcap_vendor' exists in repository 'rosbag2_storage_mcap' as well as in repository 'rosbag2'


audrow:
FYI @MichaelOrlov and @emersonknapp

emersonknapp (approved):
Yup, right move

PR #35838, opened 2023-01-12
Added package python-mistune

peci1:

## Package name:

python-mistune

## Package Upstream Source:

https://github.com/lepture/mistune

## Purpose of using this:

Markdown parser, useful for documentation tools.

- Debian: https://packages.debian.org/buster/python-mistune
- Ubuntu: https://packages.ubuntu.com/bionic/python-mistune
- Fedora: https://packages.fedoraproject.org/pkgs/python-mistune/python3-mistune/
- Arch: https://archlinux.org/packages/community/any/python-mistune/
- Gentoo: https://packages.gentoo.org/packages/dev-python/mistune
- macOS: not available (via [pip](https://pypi.org/project/mistune/))
- Alpine: not available (via [pip](https://pypi.org/project/mistune/))
- NixOS/nixpkgs: https://search.nixos.org/packages?channel=22.11&show=m2r&from=0&size=50&sort=relevance&type=packages&query=mistune
- openSUSE: https://software.opensuse.org/package/python3-mistune

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2771,6 +2771,24 @@ python-mechanize:
   gentoo: [dev-python/mechanize]
   nixos: [pythonPackages.mechanize]
   ubuntu: [python-mechanize]
+python-mistune:
+  alpine:
+    pip:
+      packages: [mistune]
+  arch: [python-mistune]
+  debian:
+    '*': [python3-mistune]
+    buster: [python-mistune]
+  fedora: [python3-mistune]
+  gentoo: [dev-python/mistune]
+  nixos: [python39Packages.mistune]
+  opensuse: [python3-mistune]
+  osx:
+    pip:
+      packages: [mistune]
+  ubuntu:
+    '*': [python3-mistune]
+    bionic: [python-mistune]
 python-mock:
   alpine: [py-mock]
   arch: [python2-mock]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2772,22 +2772,13 @@ python-mechanize:
   nixos: [pythonPackages.mechanize]
   ubuntu: [python-mechanize]
 python-mistune:
-  alpine:
-    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2771,6 +2771,24 @@ python-mechanize:
...
+  fedora: [python3-mistune]
+  gentoo: [dev-python/mistune]
+  nixos: [python39Packages.mistune]
+  opensuse: [python3-mistune]
+  osx:
+    pip:
+      packages: [mistune]
+  ubuntu:
+    '*': [python3-mistune]
+    bionic: [python-mistune]
```
tfoote, line 2791: We shouldn't be mixing python2 and 3 rules. here.

rosdep/python.yaml
```diff
@@ -2771,6 +2771,24 @@ python-mechanize:
   gentoo: [dev-python/mechanize]
   nixos: [pythonPackages.mechanize]
   ubuntu: [python-mechanize]
+python-mistune:
```
tfoote, line 2774: Most of these are python 3 rules so I'd suggest that this be python3-mistune

For the older releases we can backfill pip for python3 (aka bionic and buster) 

And with that we'd call this python3-mistune

In general we're expecting all python 2 rules to eventually disappear as python2 is now strongly deprecated.

tfoote (changes requested):
Please update this to be python3 primary. And if you need it on python2 create a separate rule. but if you're not needing that I'd suggest skipping that.

peci1:
Okay, I've split the rules for py2 and py3. I have a use for it on Melodic which requires the py2 version, so I need both rules.

tfoote (approved):
Great thanks for splitting it out. 

If you have a moment to add the suggested 'rhel' target from the linter for python3-mistune before the 2nd review that would be great.

clalancette (approved):

PR #35869, opened 2023-01-16
Add joy_tester to index (foxy, galactic, humble, rolling)

joshnewans:

foxy galactic humble rolling

(If it is not OK to submit for multiple distros, please let me know and I will split up the PRs)

# The source is here:

https://github.com/joshnewans/joy_tester

# Checks

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2107,6 +2107,16 @@ repositories:
       url: https://github.com/ros/joint_state_publisher.git
       version: foxy
     status: maintained
+  joy_tester:
+    doc:
+      type: git
+      url: https://github.com/joshnewans/joy_tester.git
+      version: main
+    source:
+      type: git
+      url: https://github.com/joshnewans/joy_tester.git
+      version: main
+    status: maintained
   joystick_drivers:
     doc:
       type: git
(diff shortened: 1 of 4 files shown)


joshnewans:
I see that it's failing on galactic with the below error.

I am happy to just not release for galactic at all if that's the only option but otherwise I would have thought there's no harm in adding it.

Please let me know how to proceed.

```
AssertionError: ["In file '''/home/runner/work/rosdistro/rosdistro/galactic/distribution.yaml''': There is a change to a doc section of an EOLed distribution. Lines: [1750, 1751, 1752, 1753]"] is not false
```

clalancette:
> I am happy to just not release for galactic at all if that's the only option but otherwise I would have thought there's no harm in adding it.

Galactic is EOL, so we no longer accept changes to it.

Besides that, we generally prefer to have separate PRs for separate distributions, since they all have different release schedules.

Because of all of that, I'm going to close this out.  Please feel free to reopen as separate PRs for Foxy, Humble, and Rolling.

joshnewans:
> Besides that, we generally prefer to have separate PRs for separate distributions, since they all have different release schedules.

Yeah that occurred to me just after I submitted it, makes sense.
I have submitted separate PRs now and they have all passed :)

PR #35891, opened 2023-01-19
l3cam_ros: 1.0.2-3 in 'noetic/distribution.yaml' [bloom]

adrisubi:
Increasing version of package(s) in repository `l3cam_ros` to `1.0.2-3`:

- upstream repository: https://github.com/beamaginelidar/l3cam_ros.git
- release repository: https://github.com/beamaginelidar/l3cam_ros-release.git
- bloom version: `0.11.2`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4373,6 +4373,22 @@ repositories:
       url: https://github.com/MITRE/kvh_geo_fog_3d.git
       version: noetic-devel
     status: maintained
+  l3cam_ros:
+    doc:
+      type: git
+      url: https://github.com/beamaginelidar/l3cam_ros.git
+      version: master
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/beamaginelidar/l3cam_ros-release.git
+      version: 1.0.2-3
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/beamaginelidar/l3cam_ros.git
+      version: master
+    status: maintained
   lanelet2:
     doc:
       type: git


noetic/distribution.yaml
```diff
@@ -4373,6 +4373,22 @@ repositories:
       url: https://github.com/MITRE/kvh_geo_fog_3d.git
       version: noetic-devel
     status: maintained
+  l3cam_ros:
+    doc:
+      type: git
+      url: https://github.com/beamaginelidar/l3cam_ros.git
```
clalancette, line 4379: There seem to be a number of problems in the source repository:

1.  The repository is missing a LICENSE file, which should be a 3-clause BSD (from the package.xml)
2.  While it is not strictly required, it is also recommended to have a license header at the top of every source file.
3.  It would be good to put something in the README.md that describes what this package is meant to do.  That would help us figure out whether it is appropriate for the rosindex and whether the name is appropriate.
4.  There seems to be a binary archive checked into the source repository, which probably should not be there: https://github.com/beamaginelidar/l3cam_ros/blob/master/include/libL3cam/libL3Cam.a

clalancette (changes requested):

noetic/distribution.yaml
adrisubi: Working on these.
The binary archive is the SDK the node depends on, which we first thought it to be integrated. Is it good measure having it integrated or should we externalize it and specify the dependency?
Thank you

clalancette: > The binary archive is the SDK the node depends on, which we first thought it to be integrated. Is it good measure having it integrated or should we externalize it and specify the dependency?

You should externalize it as much as possible.  That way the dependency is clear.

mjcarroll:
Friendly ping @adrisubi

adrisubi:
Still working on it, we are making changes on the library we use so it's taking some time. If this thread should be closed, no problem for me

clalancette:
> Still working on it, we are making changes on the library we use so it's taking some time. If this thread should be closed, no problem for me

OK, sounds good.  Will close this PR for now; feel free to open a new one when you are ready.  Thanks.

PR #35908, opened 2023-01-22
feature: add pyjwt to python.yaml

ChrisPaliqaw:

## Package name:

pyjwt

## Package Upstream Source:

https://github.com/jpadilla/pyjwt

## Purpose of using this:

pyjwt is more user-friendly and better documented than python-jwt, which is already in python.yaml.

I am using it at Robotnik for a replacement of rosauth, to add jwt-based security to rosbridge_server, so that we can restrict which topics, services and actions are accessible to users via roslibjs. Currently I use Supbase for generating JWTs, and the JWTs encode a list of user-groups. Once our end-to-end system is running, I will look into contributing these features to Robot Web Tools, and I may translate the JWT validation into C++ and augment the features of the original rosauth package.

Distro packaging links:

  - N/A - pip only
   - N/A - pip only

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3724,6 +3724,16 @@ python-pyinotify:
   fedora: [python-inotify]
   gentoo: [dev-python/pyinotify]
   ubuntu: [python-pyinotify]
+python-pyjwt-pip:
+  debian:
+    pip:
+      packages: [pyjwt]
+  fedora:
+    pip:
+      packages: [pyjwt]
+  ubuntu:
+    pip:
+      packages: [pyjwt]
 python-pykalman:
   debian:
     pip:


nuclearsandwich (approved):
As long as you understand that packages using this rosdep key won't be usable on the build farm due to the pip dependency requirements.

nuclearsandwich:
@ChrisPaliqaw this change has CI failures which need to be addressed before this can be merged.

mjcarroll:
@ChrisPaliqaw friendly ping

ChrisPaliqaw:
Sorry I missed your ping before!

It is ok to close this, thanks! Last week we decided against creating the
component that would require this library.

PR #35940, opened 2023-01-25
added python3-gtsam-pip to rosdep python

mklpiening:

## Package name:

gtsam

## Package Upstream Source:

https://github.com/borglab/gtsam

## Purpose of using this:

Python Wrapper for GTSAM for Smoothing and Mapping (SAM) in robotics.

- PyPi
  - https://pypi.org/project/gtsam/

Please add this package to the rosdep database.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7239,6 +7239,16 @@ python3-grpcio:
     '*': [python3-grpcio]
     bionic: null
     xenial: null
+python3-gtsam-pip:
+  debian:
+    pip:
+      packages: [gtsam]
+  fedora:
+    pip:
+      packages: [gtsam]
+  ubuntu:
+    pip:
+      packages: [gtsam]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (dismissed):
@mklpiening so you are aware: packages with pip dependencies cannot be released on the official ROS build farms.

The licensing for this dependency is pretty messy, which may be why it is not yet packaged on the usual platforms. Since these licensing issues won't make it to the build farm until this is packaged upstream. I do no think that blocks this pip key's inclusion in the database.

nuclearsandwich:
I am withdrawing my review approval. I didn't realize that gtsam is packaged _in ROS_ for some distributions and thus using the pip version could create conflicts. It will remain possible to use the version from pip together with custom rosdep sources but I do not think that pip-alternatives to packaged ROS software should be added to the official rosdep db.

clalancette:
No response to this in a long time, and there are concerns here.  I'm going to close this out for now, but feel free to reopen if you'd like to continue the discussion.

PR #36031, opened 2023-02-02
Add python3-inject-pip alias

igor-markovic:
Simply adds an anchor and alias for the python3 version of the `inject` package, which is already specified in the python.yaml as `python-inject-pip`.

EOL date for ubuntu 20.04 LTS from [here](https://wiki.ubuntu.com/Releases#:~:text=April%202025-,April%202030,-Ubuntu%2020.04.3%20LTS).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2421,7 +2421,7 @@ python-influxdb:
     bionic: [python-influxdb]
     disco: [python-influxdb]
     xenial: [python-influxdb]
-python-inject-pip:
+python-inject-pip: &migrate_eol_2030_04_30_python3_inject_pip
   ubuntu:
     pip:
       packages: [inject]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2421,7 +2421,7 @@ python-influxdb:
     bionic: [python-influxdb]
     disco: [python-influxdb]
     xenial: [python-influxdb]
-python-inject-pip:
+python-inject-pip: &migrate_eol_2030_04_30_python3_inject_pip
```
sloretz, line 2424: I think the date we use is April 2025. The extra 5 years is "Extended Security Maintenance" which requires an "Ubuntu Advantage Subscription"

https://ubuntu.com/about/release-cycle

igor-markovic: Okay, I've updated the date to 2025. Thanks

sloretz (approved):

mjcarroll (approved):

PR #36052, opened 2023-02-07
Added python3-open3d via .deb package

esteve:

## Package name:

python3-open3d

## Package Upstream Source:

https://github.com/isl-org/Open3D

## Purpose of using this:

This package already exists in the index as a PIP dependency, this adds it as a distribution package (for Ubuntu and Debian, I couldn't find it for Fedora)

Distro packaging links:

  - https://packages.debian.org/bullseye/python3-open3d
   - https://packages.ubuntu.com/jammy/python3-open3d

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8010,9 +8010,14 @@ python3-onnxruntime-pip:
       packages: [onnxruntime]
 python3-open3d:
   debian:
-    bullseye: [python3-open3d]
+    '*': [python3-open3d]
+    buster: null
+    stretch: null
   ubuntu:
-    jammy: [python3-open3d]
+    '*': [python3-open3d]
+    artful: null
+    bionic: null
+    focal: null
 python3-open3d-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


audrow (approved):

esteve:
@audrow thanks for the quick review. Shall I merge this or do I wait for one of @ros/rosdeputies ?

rosdep/python.yaml
```diff
@@ -8008,6 +8008,11 @@ python3-onnxruntime-pip:
   ubuntu:
     pip:
       packages: [onnxruntime]
+python3-open3d:
+  debian:
+    bullseye: [python3-open3d]
+  ubuntu:
+    jammy: [python3-open3d]
```
SteveMacenski, line 8015: If we assume this is going to be released into future distributions, wouldn't it be more future-proofed to label this with the name as `'*'` and then null out any active distributions its not available in?

esteve: Good idea, I've pushed https://github.com/ros/rosdistro/pull/36052/commits/e2f538907999db6a3265901ad593f4dcbba65532 so that `python3-open3d` is available for only the Ubuntu and Debian releases that ship with it

SteveMacenski (approved):

PR #36071, opened 2023-02-08
Add python3-sshtunnel

tlipps:

## Package name:

sshtunnel

## Package Upstream Source:

https://github.com/pahaz/sshtunnel

## Purpose of using this:

sshtunnel offers pure python way to ssh on different machines. The following use case is copied from https://github.com/pahaz/sshtunnel/#usage-scenarios

> One of the typical scenarios where sshtunnel is helpful is depicted in the figure below. User may need to connect a port of a remote server (i.e. 8080) where only SSH port (usually port 22) is reachable.

```

    ----------------------------------------------------------------------

                                |
    -------------+              |    +----------+
        LOCAL    |              |    |  REMOTE  | :22 SSH
        CLIENT   | <== SSH ========> |  SERVER  | :8080 web service
    -------------+              |    +----------+
                                |
                             FIREWALL (only port 22 is open)

    ----------------------------------------------------------------------
```

> Fig1: How to connect to a service blocked by a firewall through SSH tunnel.

In my use case it is used to access ROS logs/bags saved in mongo DBs on different machines via a mongo client with help of the sshtunnel's `SSHTunnelForwarder` class.

Distro packaging links:

  - https://packages.debian.org/unstable/python3-sshtunnel
   - https://packages.ubuntu.com/focal/python3-sshtunnel
  - https://packages.fedoraproject.org/pkgs/python-sshtunnel/python3-sshtunnel/
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9524,10 +9524,20 @@ python3-squaternion-pip:
 python3-sshtunnel:
   debian: [python3-sshtunnel]
   fedora: [python3-sshtunnel]
-  ubuntu: [python3-sshtunnel]
   ubuntu:
-    pip:
-      packages: [sshtunnel]
+    '*': [python3-sshtunnel]
+    focal:
+      pip:
+        packages: [sshtunnel]
+    jammy:
+      pip:
+        packages: [sshtunnel]
+    kinetic:
+      pip:
+        packages: [sshtunnel]
+    lunar:
+      pip:
+        packages: [sshtunnel]
 python3-stable-baselines3-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9525,13 +9525,8 @@ python3-sshtunnel:
   debian: [python3-sshtunnel]
   fedora: [python3-sshtunnel]
   ubuntu:
-    '*': [python3-sshtunnel]
-    focal:
-      pip:
-        packages: [sshtunnel]
-    jammy:
-      pip:
-        packages: [sshtunnel]
+    focal: [python3-sshtunnel]
+    jammy: [python3-sshtunnel]
 python3-stable-baselines3-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9532,12 +9532,6 @@ python3-sshtunnel:
     jammy:
       pip:
         packages: [sshtunnel]
-    kinetic:
-      pip:
-        packages: [sshtunnel]
-    lunar:
-      pip:
-        packages: [sshtunnel]
 python3-stable-baselines3-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9521,6 +9521,13 @@ python3-squaternion-pip:
   ubuntu:
     pip:
       packages: [squaternion]
+python3-sshtunnel:
+  debian: [python3-sshtunnel]
+  fedora: [python3-sshtunnel]
+  ubuntu: [python3-sshtunnel]
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Please fill in your TODO fields -- why is this a draft? Is there something more you wish to do before this is reviewed?

tlipps:
Hey @SteveMacenski I hoped that the draft is not directly publicly visible, but now everything is filled out and I changed the state into 'ready to review'.

SteveMacenski:
Nope, drafts are the exact same as normal PRs, just without the ability for me to mergei t

rosdep/python.yaml
```diff
@@ -9521,6 +9521,23 @@ python3-squaternion-pip:
...
+  fedora: [python3-sshtunnel]
+  ubuntu:
+    '*': [python3-sshtunnel]
+    focal:
+      pip:
+        packages: [sshtunnel]
+    jammy:
+      pip:
+        packages: [sshtunnel]
+    kinetic:
```
SteveMacenski, line 9535: You can remove kineitc and lunar, those aren't Ubuntu OS which ROS 2 releases binaries to (or even has a distribution on)

rosdep/python.yaml
```diff
@@ -9521,6 +9521,23 @@ python3-squaternion-pip:
   ubuntu:
     pip:
       packages: [squaternion]
+python3-sshtunnel:
+  debian: [python3-sshtunnel]
+  fedora: [python3-sshtunnel]
+  ubuntu:
+    '*': [python3-sshtunnel]
```
SteveMacenski, line 9528: @mjeronimo I'm not much a python user, but this doesn't look quite right to me. I'm not sure why we have something called `python3-XYZ` and then `XYZ` under pip. Looking over this file, I don't see any analogs to this setup which makes me think its incorrectly formatted.

SteveMacenski (changes requested):

SteveMacenski:
CI: `Error: Package 'python3-sshtunnel' could not be found for ubuntu bionic on amd64, Warning: Key 'python3-sshtunnel' might be satisifed by opensuse package named 'python3-sshtunnel': https://software.opensuse.org/package/python-sshtunnel`

rosdep/python.yaml
tlipps: Frankly, I'm not sure whether this is correct, but I've seen this in the python.yaml. Here is an example with the rosdep key `python-future`: https://github.com/ros/rosdistro/blob/1962b88f2000d984a9bfa48a04670c076482c630/rosdep/python.yaml#L1851

tlipps: @SteveMacenski I've decided to just remove the pull from pip. I hoped to get both the system's package distributor (e.g. apt on ubuntu) and pip inside the python.yaml. However, I see no way to do that without using the wildcard operator, which is too powerful (no package for bionic: https://github.com/ros/rosdistro/actions/runs/4141872809). So I just removed pip: https://github.com/ros/rosdistro/pull/36071/commits/f6e326ac42ae761db1877e8687289bfabd843c09

tlipps:
> CI: `Error: Package 'python3-sshtunnel' could not be found for ubuntu bionic on amd64, Warning: Key 'python3-sshtunnel' might be satisifed by opensuse package named 'python3-sshtunnel': https://software.opensuse.org/package/python-sshtunnel`

Please see https://github.com/ros/rosdistro/pull/36071#discussion_r1102402076

rosdep/python.yaml
tlipps: Done: https://github.com/ros/rosdistro/pull/36071/commits/1962b88f2000d984a9bfa48a04670c076482c630

SteveMacenski (approved):

PR #36076, opened 2023-02-09
Ubuntu jammy entry for python3-udev

boilerbots:
## Package name:

python3-udev

## Package Upstream Source:

https://packages.ubuntu.com/source/jammy/pyudev

## Purpose of using this:

Adding an entry for Ubuntu jammy so the default "python-udev" is not used.

https://packages.ubuntu.com/jammy/python3-pyudev

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4082,7 +4082,12 @@ python-pyudev:
     trusty_python3: [python3-pyudev]
     xenial_python3: [python3-pyudev]
     zesty_python3: [python3-pyudev]
-    jammy: [python3-pyudev]
+python3-pyudev:
+  debian: [python3-pyudev]
+  fedora: [python3-pyudev]
+  gentoo: [dev-python/pyudev]
+  nixos: [pythonPackages.pyudev]
+  ubuntu: [python3-pyudev]
 python-pyusb-pip:
   debian:
     pip: [pyusb]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4082,6 +4082,7 @@ python-pyudev:
     trusty_python3: [python3-pyudev]
     xenial_python3: [python3-pyudev]
     zesty_python3: [python3-pyudev]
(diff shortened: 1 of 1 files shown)


SteveMacenski:
@clalancette this is a bit of an odd situation - how is this handled with the others having `_python3` ?

clalancette:
> @clalancette this is a bit of an odd situation - how is this handled with the others having `_python3` ?

Right, good question.  The `_python3` keys are a historical artifact, and we shouldn't add new ones anymore[1].

Instead, what we should do here is to add an entirely new key called `python3-pyudev`, and populate it with the python3 keys.  It will look something like:

```
python3-pyudev:
  debian: [python3-pyudev]
  fedora: [python3-pyudev]
  gentoo: [dev-python/pyudev]
  nixos: [pythonPackages.pyudev]
  ubuntu: [python3-pyudev]
```

The `python-pyudev` key can remain as-is, mostly for Python 2 and legacy purposes.

[1] The history is that these were our first attempt to handle Python 3 keys alongside Python 2 ones.  But we found that this didn't work well, so we switched to the system we use today.

SteveMacenski:
@boilerbots can you update accordingly?

boilerbots:
Yes let me rework this.

emersonknapp (changes requested):
This looks good now, except that the new key is now not in alphabetical order. It'll need to go down in the `python3-` area. Output from tests:

> list out of alphabetical order line 4091.  'python-pyusb-pip' should come before 'python3-pyudev'

boilerbots:
Hope I didn't break this, I rebased my branch to keep up with the changes.

emersonknapp (approved):
LGTM - I might suggest adding `rhel` and `opensuse` as per automated suggestions, but it's not required.

boilerbots:
sorry I don't want to keep touching these files.

emersonknapp:
@mjcarroll i think we just need another set of eyes on this one

mjcarroll (approved):

Issue #36088, opened 2023-02-10
Missing rosdep key for pymap3d-pip in ubuntu jammy


pariaspe:
Hi all, I'm new here, sorry if I'm not in the right place.

I'm getting a jenkins build [error](https://build.ros2.org/job/Hdev__aerostack2__ubuntu_jammy_amd64/lastBuild/console), not finding pymap3d-pip key. rosdep rule is defined [here](https://github.com/ros/rosdistro/blob/4b7fb961ad829ef1433ff9e1b2b45a615e52ff00/rosdep/python.yaml#L405) and it seems right. Also our own CI action finds it, any help?

Should I open a PR adding ubuntu jammy version? Should I add a new rule for `python3-pymap3d` apt [pkg](https://packages.ubuntu.com/jammy/python3-pymap3d)? Thanks!

SteveMacenski:
The name in your package.xml is wrong `pymap3d`. The key is actually `pymap3d-pip` as you can see as the root name of the rosdep entries. This is what we look for in the package.xml parsing. Updating that should fix your issue.

Separately, we do generally prefer apt packages over pip packages, so submitting updates to use the apt version would be a separate and useful contribution

pariaspe:
Hey, thanks for the help.

We actually are using the key as you mention in the [package.xml](https://github.com/aerostack2/aerostack2/blob/main/as2_python_api/package.xml#L19).

Alright, I'll open a PR in the next days, glad to contribute!

SteveMacenski:
Oh I see (sorry, I scanned quickly and that's the most common reason), that actually looks like the key resolved. It looks like that doesn't actually exist on your OS... which is odd because I can pip install that on my 22.04 machine. That is odd. Can't say I know the reasons for that, might be something pythony (which I don't use extensively). Its perhaps best to get the opinions of another maintainer here that is more familiar with these kinds of issues w.r.t. python

pariaspe:
Solved in #36164 , thanks!

PR #36164, opened 2023-02-14
Adding rosdep rule for python3-pymap3d

pariaspe:

## Package name:

python3-pymap3d

## Package Upstream Source:

https://github.com/geospace-code/pymap3d

## Purpose of using this:

As explained in #36088, `pymap3d-pip` key is not available in Ubuntu Jammy. Following Steve's suggestion, I'm opening this PR. 

Distro packaging links:

- Debian:
  - https://packages.debian.org/bullseye/python3-pymap3d
- Ubuntu:
   - https://packages.ubuntu.com/jammy/python3-pymap3d
- Fedora:
  - Not Available
- Arch:
  - Not Available
- Gentoo:
  - https://packages.gentoo.org/packages/sci-geosciences/pymap3d
- macOS:
  - Not Available
- Alpine:
  - Not Available
- NixOS/nixpkgs:
  - Not Available
- openSUSE:
  - Not Available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8268,7 +8268,8 @@ python3-pylibdmtx:
       pip:
         packages: [pylibdmtx]
 python3-pymap3d:
-  debian: [python3-pymap3d]
+  debian:
+    bullseye: [python3-pymap3d]
   ubuntu:
     jammy: [python3-pymap3d]
 python3-pymesh2-pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8267,6 +8267,8 @@ python3-pylibdmtx:
     focal:
       pip:
         packages: [pylibdmtx]
+python3-pymap3d:
+  ubuntu: [python3-pymap3d]
 python3-pymesh2-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -392,8 +392,16 @@ pyflakes3:
   ubuntu: [pyflakes3]
 pymap3d-pip:
   debian:
-    pip:
-      packages: [pymap3d]
+    '*': null
+    buster:
+      pip:
+        packages: [pymap3d]
+    stretch:
+      pip:
+        packages: [pymap3d]
+    jessie:
+      pip:
+        packages: [pymap3d]
   fedora:
     pip:
       packages: [pymap3d]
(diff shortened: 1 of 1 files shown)


pariaspe:
A better solution would be merging `pymap3d-pip` and `python3-pymap3d` entries, something like:
```yaml
python3-pymap3d:
    debian:
      '*':
        pip:
          packages: [pymap3d]
      bullseye: [python3-pymap3d]
    fedora:
      pip:
        packages: [pymap3d]
    gentoo: [sci-geosciences/pymap3d]
    osx:
      pip:
        packages: [pymap3d]
    ubuntu:
      '*':
        pip:
          packages: [pymap3d]
      jammy: [python3-pymap3d]
    ubuntu:
      pip:
        packages: [pymap3d]
```

I've added it as a different entry, since I don't want to break a existing rule.

rosdep/python.yaml
```diff
@@ -8636,6 +8636,12 @@ python3-pylibdmtx:
     focal:
       pip:
         packages: [pylibdmtx]
+python3-pymap3d:
```
SteveMacenski, line 8639: If we think this will be released into future distributions, we can instead use the wildcard to set the package for all variants and specify `null` for the specific distributions with active ROS distributions which don't have it (e.g. 20.04). That lets this work with the next distribution without later needing to modify again

tfoote (changes requested):
> I've added it as a different entry, since I don't want to break a existing rule. 

We should not support both the pip and apt based rules in parallel on any platform. And the native packages take precedence. 

To that end we should update the pip rule to null out the overlapping platforms. We don't need to break past ones but we want to avoid overlaping rules where the pip version will overlay the apt version. Including nulling out future versions on the pip rule. 

And with that said we should use wildcard for the apt rule to match with expected future versions and null out the older supported platforms. 

We will break existing users who are using pip on the apt supported platforms, but it will give a clear error that the key isn't defined anymore. This is considered an error in the rules as we should have switched to the apt rule already.

pariaspe:
Done
- `pymap3d-pip` --> debian buster, jessie and stretch and ubuntu bionic and focal only
- `python3-pymap3d` --> debian bullseye and ubuntu jammy and newer distributions

emersonknapp (approved):
Thanks for the iteration, this looks a lot better

tfoote (approved):
Thanks for the iteration
 Lgtm

emersonknapp:
@nuclearsandwich Not sure if the branch protection test requirements are fully correct here - it looks like it's expecting a test that won't run. Is this an artifact of the changeover to pytest?

mjcarroll (approved):

nuclearsandwich:
> Not sure if the branch protection test requirements are fully correct here - it looks like it's expecting a test that won't run. Is this an artifact of the changeover to pytest?

Following up, thanks for the catch. I did not realize that branch protections do text matching on repository checks and it took me a bit to find and update this one. It has since been corrected.

PR #36220, opened 2023-02-16
Update names for ignition->gz renaming

mjcarroll:
Connected to: https://github.com/ros2-gbp/ros2-gbp-github-org/pull/200


mjcarroll:
This is going to take additional modifications to the `_vendor` repos, so I'm keeping it in draft for now.

cottsay (changes requested):
I believe that the migration script made changes to the old `ros2-gbp` repositories so we may need to deal with that prior to merging this.

PR #36045, opened 2023-02-06
Add apa102-pi in 'rosdep/python.yaml'

Kotochleb:
## Package name:

python3-apa102-pi-pip

## Package Upstream Source:

https://github.com/tinue/apa102-pi

## Purpose of using this:

APA102 LED driver for Raspberry Pi. Useful for LED animations without use of microcontrollers.

## Links to Distribution Packages

- Debian: not available (via [pip](https://pypi.org/project/apa102-pi/))
- Ubuntu: not available (via [pip](https://pypi.org/project/apa102-pi/))
- Fedora: not available
- Arch:  not available
- Gentoo:  not available
- macOS: not available
- Alpine: not available
- NixOS/nixpkgs: not available
- openSUSE: not available

# Checks
 - [ ] All packages have a declared license in the package.xml
 - [x] This repository has a LICENSE file
 - [ ] This package is expected to build on the submitted rosdistro

PR #36222, opened 2023-02-17
Add apa102-pi in 'rosdep/python.yaml'

Kotochleb:
This PR is in refference to #36045 

## Package name:

python3-apa102-pi-pip

## Package Upstream Source:

https://github.com/tinue/apa102-pi

## Purpose of using this:

APA102 LED driver for Raspberry Pi. Useful for LED animations without use of microcontrollers.

- Debian: not available (via [pip](https://pypi.org/project/apa102-pi/))
- Ubuntu: not available (via [pip](https://pypi.org/project/apa102-pi/))
- Fedora: not available
- Arch:  not available
- Gentoo:  not available
- macOS: not available
- Alpine: not available
- NixOS/nixpkgs: not available
- openSUSE: not available

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6003,7 +6003,7 @@ python3-anytree-pip:
   ubuntu:
     pip:
       packages: [anytree]
-python3-apa102-pi:
+python3-apa102-pi-pip:
   debian:
     pip:
       depends: [python-rpi.gpio-pip]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6003,6 +6003,15 @@ python3-anytree-pip:
   ubuntu:
     pip:
       packages: [anytree]
+python3-apa102-pi:
+  debian:
+    pip:
+      depends: [python-rpi.gpio-pip]
+      packages: [apa102-pi]
+  ubuntu:
+    pip:
+      depends: [python-rpi.gpio-pip]
+      packages: [apa102-pi]
 python3-argcomplete:
   alpine: [py3-argcomplete]
   arch: [python-argcomplete]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6022,6 +6022,15 @@ python3-anytree-pip:
   ubuntu:
     pip:
       packages: [anytree]
+python3-apa102-pi-pip:
+  debian:
+    pip:
+      depends: [python-rpi.gpio-pip]
```
emersonknapp, line 6028: Is this a python2 dependency for this python3 package? I noticed in the https://github.com/tinue/apa102-pi#quick-raspberry-pi-setup section that it suggests `sudo apt install -y python3-pip python3-venv python3-rpi.gpio` - so should we also depend on the native package `python3-rpi.gio`?

Kotochleb: I think you are right. I left it this way because it worked... Probably python3 was downloading required dependency itself in background.

I could add new dependency `python3-rpi.gpio-pip`, but I'm not sure if it is in the scope of this PR.

emersonknapp: I think if it's an implicit dependency that `pip` will download automatically, then you don't need to specify anything at all. The `depends:` field is for native packages that this `pip` package depends on, basically for cross-package-manager usage.

Kotochleb: Thank you for detailed description and help. I removed `depends` filed and left it to download dependency itself.

emersonknapp (approved):

mjcarroll:
Just a friendly reminder that binary released packages won't be able to depend on these `pip`-only keys.

PR #36229, opened 2023-02-19
Adding robosar to documentation index for distro

charvigupta96:
ROSDISTRO - noetic
# The source is here:
https://github.com/MRSD-Team-RoboSAR/robosar.git

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -8256,7 +8256,7 @@ repositories:
       url: https://github.com/rm-controls/rm_description.git
       version: master
     status: developed
-  robosar:
+  robosar_multi_robot_search:
     doc:
       type: git
       url: https://github.com/MRSD-Team-RoboSAR/robosar.git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -8256,6 +8256,11 @@ repositories:
       url: https://github.com/rm-controls/rm_description.git
       version: master
     status: developed
+  robosar:
+    doc:
+      type: git
+      url: https://github.com/MRSD-Team-RoboSAR/robosar.git
+      version: main
   robot_body_filter:
     doc:
       type: git


emersonknapp:
I've noticed that not all packages in the metarepo's submodules have `LICENSE` files - this is a concern from a release perspective. It may not be as much of a concern for a documentation-only package, I will need another rosidistro reviewer to weigh in.

clalancette:
> I've noticed that not all packages in the metarepo's submodules have LICENSE files - this is a concern from a release perspective.

So I think a submodule-only style repository like this would have trouble releasing in any case; I don't know if our buildfarm understands submodules like this.  Maybe @nuclearsandwich can weigh in?

> It may not be as much of a concern for a documentation-only package, I will need another rosidistro reviewer to weigh in.

Regardless, this is a concern.  All new packages should have some LICENSE declaration, including their submodules.

emersonknapp:
I think the `package.xml` description should match a full-text license in a file named `LICENSE` - both parts is the best approach so there is no ambiguity about the licensing.

nuclearsandwich (changes requested):
This metarepo approach isn't conventional within the ROS release process. While bloom does now clone submodules I don't know if package discovery in submodules fully works and since each submodule is itself a repository with multiple packages, I think it would make the most sense to add each repository to the ROS distribution normally rather than via this submodule network. In general the larger the release repository the more difficult it is for the package structure to evolve (say when individual packages split out into their own repositories).

nuclearsandwich:
> Regarding submodule-only style repository, I was referring to the tutorial for indexing and it had this PR - #13688 which has submodules in it.

@charvigupta96 perhaps I am misreading something but when I look at the source repository for your reference PR: https://github.com/uos/katana_driver/tree/kinetic I see packages in sub<em>directories</em> but not sub<em>modules</em>. All of those directories are managed directly in the katana_driver repository.

nuclearsandwich:
I'm closing this PR due to lack of activity @charvigupta96 please feel free to re-open (or create a new PR with links to this one) once you have addressed the above feedback.

PR #36242, opened 2023-02-20
Add gz-transport-12 for Ubuntu

### LINKED ISSUE #385: Missing rosdep key for gz-transport-12
## Environment
* OS Version: Ubuntu 22.04 (Jammy)
* Source or binary build?

## Description
* Expected behavior: Using rosdep resolves all keys
* Actual behavior: ROSdep is missing keys for gz-transport-12

## Steps to reproduce

Follow instructions for compiling ros_gz here: https://github.com/gazebosim/ros_gz/tree/ros2#compile-ros_gz

## Output

```
ryan@ryan-m93p:~/Documents/gz_ws$ rosdep install -r --from-paths src -i -y
to system dependencies:
ros_gz_bridge: Cannot locate rosdep definition for [gz-transport12]
ros_gz_image: Cannot locate rosdep definition for [gz-transport12]
ros_gz_sim: Cannot locate rosdep definition for [gz-math7]
Continuing to install resolvable dependencies...
#All required rosdeps installed successfully
```
## The problem

This package has uploaded debians with a new key name, but has not updated the rosdep index, and now other packages are relying on version 11 that does not exist. 
https://github.com/ros/rosdistro/blob/91349fa4751c7a5b62b1129220148a11c693694f/rosdep/base.yaml#L1869-L1875

This was found in this issue: https://github.com/gazebosim/ros_gz/issues/358

Ryanf55:
* https://gazebosim.org/libs/transport
* Fixes https://github.com/gazebosim/gz-transport/issues/385

## Package name:

gz-transport-12

## Package Upstream Source:

[link to source repository](https://gazebosim.org/libs/transport)

## Purpose of using this:

For using gazebo garden with ROS2 humble

Distro packaging links:

I cannot find the listings since it's a private cache. However, the build status shows it's being built. I have tested this on Ubuntu 22.04. 

Doing any other package is out of scope since that's all that is released that I can tell. 

# Checks

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1870,6 +1870,10 @@ gz-transport11:
   ubuntu:
     focal: [libgz-transport11-dev]
     jammy: [libgz-transport11-dev]
+gz-transport12:
+  ubuntu:
+    focal: [libgz-transport12-dev]
+    jammy: [libgz-transport12-dev]
 gz-transport8:
   ubuntu:
     focal: [libgz-transport8-dev]


emersonknapp (approved):

Ryanf55:
I could use some help with fixing the failing build. 
`Package 'libgz-transport12-dev' could not be found for ubuntu focal on amd64`

I copied verbatim what was done before, and validated on focal this package existed. On a clean slate ubuntu system, the package is not there, till you add the `apt list` file according to the [docs ](https://gazebosim.org/api/sensors/7/installation.html). 

```
sudo sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-stable.list'
wget http://packages.osrfoundation.org/gazebo.key -O - | sudo apt-key add -
```

Now, assuming someone already set this up before, how does it work for version 11, but not 12? I don't see any notes in the contributing guide about adding keys. Any tips?

mjcarroll:
The `gazebo` packages tend to be a little bit tricky.  The rule for rosdep dependencies is that they either need to be available in the upstream Ubuntu repositories, or in the rosdep repository itself.

On Ubuntu Jammy, the upstream version is 11.0 (https://packages.ubuntu.com/source/jammy/ignition-transport), which is why the previous rule worked.

In this case, the `ros_gz` bridge is allowed to build from source with dependencies from the packages.osrfoundation repo, but cannot depend on them for the binary deb builds.

Ryanf55:
> The `gazebo` packages tend to be a little bit tricky. The rule for rosdep dependencies is that they either need to be available in the upstream Ubuntu repositories, or in the rosdep repository itself.
> 
> On Ubuntu Jammy, the upstream version is 11.0 (https://packages.ubuntu.com/source/jammy/ignition-transport), which is why the previous rule worked.
> 
> In this case, the `ros_gz` bridge is allowed to build from source with dependencies from the packages.osrfoundation repo, but cannot depend on them for the binary deb builds.

Thanks for the clarification. How would you recommend I proceed to help to ensure the gazebo packages install their dependencies correctly in Garden?

azeey:
> Thanks for the clarification. How would you recommend I proceed to help to ensure the gazebo packages install their dependencies correctly in Garden?

Please see https://gazebosim.org/docs/garden/ros_installation#using-an-specific-gazebo-version-with-ros-2

PR #36285, opened 2023-02-24
Remove Debian Buster from the repository checks

sloretz:
Debian Buster is EOL

https://github.com/ros-infrastructure/rep/pull/373

Similar to: https://github.com/ros/rosdistro/pull/30094 and https://github.com/ros/rosdistro/pull/27640

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -3,8 +3,6 @@
 # see REP 143: http://ros.org/reps/rep-0143.html
 ---
 release_platforms:
-  debian:
-  - buster
   ubuntu:
   - focal
 repositories:
--- a/test/rosdep_repo_check/config.yaml
+++ b/test/rosdep_repo_check/config.yaml
@@ -61,7 +61,6 @@ supported_versions:
   arch:
   - ''
   debian:
-  - buster
   - bullseye
   fedora:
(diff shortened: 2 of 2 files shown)


clalancette (approved):
Looks reasonable to me, though I'd like a second check from @nuclearsandwich or @cottsay before I merge this.

cottsay (approved):
Thanks!

noetic/distribution.yaml
```diff
@@ -3,8 +3,6 @@
 # see REP 143: http://ros.org/reps/rep-0143.html
 ---
 release_platforms:
-  debian:
-  - buster
```
nuclearsandwich, line 7: This change governs the release platforms that Bloom uses when generating release data for Noetic. I didn't expect to see it here based on the PR title and description.

In my [little checklist](https://github.com/ros-infrastructure/ros_buildfarm_config/pull/221#issuecomment-1444599141), this change happens much later than the removal of the repo checker config. As a result, I think this PR was merged somewhat prematurely.

sloretz: Reverted in https://github.com/ros/rosdistro/pull/36303 and redone in https://github.com/ros/rosdistro/pull/36387

PR #36303, opened 2023-02-24
Re-add debian buster to Noetic distribution.yaml

sloretz:
Partially reverts https://github.com/ros/rosdistro/pull/36285 - after talking with @nuclearsandwich I learned I made a mistake. This needs to be restored until after the regressions in Buster are resolved and the last snapshot is made.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -3,6 +3,8 @@
 # see REP 143: http://ros.org/reps/rep-0143.html
 ---
 release_platforms:
+  debian:
+  - buster
   ubuntu:
   - focal
 repositories:


nuclearsandwich (approved):

nuclearsandwich:
> This needs to be restored until after the regressions in Buster are resolved and the last snapshot is made.

In particular, what we want is for the last sync which includes changes to buster to be a suitable "final" sync. For our existing workflows we stop updating distros after the final sync and although there are no guarantees that we won't remove them from the main repositories at a later date, removing buster from the main repos partially (while also updating some other packages) is unprecedented and would have been done, essentially "accidentally" rather than purposefully, which is what the final snapshots were meant to support.

PR #36323, opened 2023-02-26
ros_2cli: 0.2.0-1 in 'rolling/distribution.yaml' [bloom]

christophebedard:
Increasing version of package(s) in repository `ros_2cli` to `0.2.0-1`:

- upstream repository: https://github.com/christophebedard/ros_2cli.git
- release repository: https://github.com/christophebedard/ros_2cli-release.git
- bloom version: `0.11.2`

## ros_2cli

```
* Switch to Python implementation based on ros2cli for autocompletion
* Contributors: Christophe Bedard
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4513,6 +4513,17 @@ repositories:
       url: https://github.com/osrf/ros2launch_security.git
       version: main
     status: maintained
+  ros_2cli:
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/christophebedard/ros_2cli-release.git
+      version: 0.2.0-1
+    source:
+      type: git
+      url: https://github.com/christophebedard/ros_2cli.git
+      version: rolling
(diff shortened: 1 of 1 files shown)


christophebedard:
> FAILED test/test_build_caches.py::test_build_caches - RuntimeError: rolling: missing package.xml file for package "ros_2cli"

The package does have a `package.xml` file :thinking: https://github.com/christophebedard/ros_2cli/blob/0.2.0/ros_2cli/package.xml

clalancette:
I really don't think we should have this in the repository.  While it is great as a joke, I think the potential for confusion is very high if we have it in the main repositories.  And I really do think that at least for the foreseeable future, we should keep this as `ros2` commands to disambiguate ROS 2 from ROS 1.

That said, others might have different opinions, so maybe we should have a discussion about this.

christophebedard:
That's totally fair. I certainly didn't intend on this being anything more than a joke 😄

Closing this

PR #36324, opened 2023-02-27
Add python3-ultralytics-pip

martinandrovich:

## Package name:

python3-ultralytics

## Package Upstream Source:

https://pypi.org/project/ultralytics/
https://github.com/ultralytics/ultralytics

## Purpose of using this:

YOLO (You Only Look Once) is an object detection and image segmentation model developed by [Ultralytics](https://ultralytics.com/). The `YOLOv8` model can be useful for computer-vision based tasks, where it can be easily integrated into a ROS node.

Distro packaging links:

  - https://pypi.org/project/ultralytics/
   - https://pypi.org/project/ultralytics/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9922,7 +9922,7 @@ python3-ujson:
     '*': [python3-ujson]
     '7': null
   ubuntu: [python3-ujson]
-python3-ultralytics:
+python3-ultralytics-pip:
   debian:
     pip:
       packages: [ultralytics]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -9922,6 +9922,13 @@ python3-ujson:
     '*': [python3-ujson]
     '7': null
   ubuntu: [python3-ujson]
+python3-ultralytics:
```
clalancette, line 9925: Since it looks like this is only available via pip, this key should be `python3-ultralytics-pip`.

clalancette (changes requested):

rosdep/python.yaml
martinandrovich: Did not know that, will change :+1: .

clalancette (approved):

methylDragon (approved):

PR #36375, opened 2023-03-02
VDA 5050 Connector for mobile transport robots in intralogistics

frothm:

melodic vda5050_connector
noetic vda5050_connector

# The source is here:

https://github.com/tum-fml/ros_vda5050_connector.git

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -14815,11 +14815,11 @@ repositories:
     doc:
       type: git
       url: https://github.com/tum-fml/vda5050_connector.git
-      version: v0.0.3
+      version: current_ros_melodic
     source:
       type: git
       url: https://github.com/tum-fml/vda5050_connector.git
-      version: v0.0.3
+      version: current_ros_melodic
     status: maintained
   velo2cam_calibration:
     doc:
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -14811,6 +14811,16 @@ repositories:
       url: https://github.com/anybotics/variant-release.git
       version: 0.1.6-1
     status: maintained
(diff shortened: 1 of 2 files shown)


methylDragon:
Note that since this package was not bloomed, it won't be available to `apt install` and won't build on the buildfarm. Is the intention just to get the docs and source indexed?

clalancette (approved):

melodic/distribution.yaml
```diff
@@ -14811,6 +14811,16 @@ repositories:
       url: https://github.com/anybotics/variant-release.git
       version: 0.1.6-1
     status: maintained
+  ros_vda5050_connector:
```
clalancette, line 14814: According to CI, this is now alphabetically in the wrong place.  Also, this should match the name of your package, so this really should be `vda5050_connector`.  Same goes for the noetic side.

clalancette (changes requested):

melodic/distribution.yaml
clalancette: So the thing is that the new name is not [REP-144 compliant](https://www.ros.org/reps/rep-0144.html), particularly the 3rd global rule which says:

```
A package name should not contain ros as it is redundant. Exceptions include core packages and ROS bindings of an upstream library (e.g. moveit_ros)
```

I guess the question is whether this qualifies for an exception?

frothm: I totally agree... It is difficult making all those marketing/PR people behind me happy.

I now convinced my team to stick with the previous, conform name for the ROS package, but rename the git repository. I hope this is acceptable, even though the package name now differs from the repo name...

clalancette (approved):

PR #36406, opened 2023-03-06
Add base rules for paho-mqtt-c and paho-mqtt-cpp

lreiher:

## Package name:

- libpaho-mqtt
- libpaho-mqtt-dev
- libpaho-mqttpp
- libpaho-mqttpp-dev

## Package Upstream Source:

- https://github.com/eclipse/paho.mqtt.c
- https://github.com/eclipse/paho.mqtt.cpp

## Purpose of using this:

The Paho MQTT C and C++ client libraries are added. The libraries are dependencies of our ROS package [mqtt_client](https://github.com/ika-rwth-aachen/mqtt_client). ROS1 Noetic and ROS2 Galactic already [include a rule to install the libraries from a dedicated release repository](https://github.com/ros/rosdistro/blob/master/noetic/distribution.yaml#L6759). For ROS2 support I'm adding them as system dependencies since they have been officially released as apt dependencies in Ubuntu Jammy. Related to https://github.com/nobleo/paho.mqtt.cpp-release/issues/1.

  - https://packages.debian.org/bullseye/libpaho-mqtt1.3
  - https://packages.debian.org/bullseye/libpaho-mqtt-dev
   - https://packages.ubuntu.com/jammy/libpaho-mqtt1.3
   - https://packages.ubuntu.com/jammy/libpaho-mqtt-dev
   - https://packages.ubuntu.com/jammy/libpaho-mqttpp3-1
   - https://packages.ubuntu.com/jammy/libpaho-mqttpp-dev
  - https://packages.fedoraproject.org/pkgs/paho-c/paho-c/
  - https://packages.fedoraproject.org/pkgs/paho-c/paho-c-devel/
  - not available
  - not available
  - not available
  - https://pkgs.alpinelinux.org/packages?name=paho-mqtt-c
  - https://pkgs.alpinelinux.org/packages?name=paho-mqtt-c-dev
  - https://search.nixos.org/packages?channel=22.11&show=paho-mqtt-cpp&from=0&size=50&sort=relevance&type=packages&query=paho
  - https://search.nixos.org/packages?channel=22.11&show=paho-mqtt-cpp&from=0&size=50&sort=relevance&type=packages&query=paho
  - https://software.opensuse.org/package/libpaho-mqtt1
  - https://software.opensuse.org/package/libpaho-mqtt-devel

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7160,6 +7160,14 @@ osmium:
   fedora: [libosmium-devel]
   nixos: [libosmium]
   ubuntu: [libosmium-dev]
+paho-mqtt-c:
+  debian:
+    bullseye: [libpaho-mqtt1.3]
+  ubuntu:
+    jammy: [libpaho-mqtt1.3]
+paho-mqtt-cpp:
+  ubuntu:
+    jammy: [libpaho-mqttpp3-1]
 pandoc:
   debian: [pandoc]
   fedora: [pandoc]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7162,17 +7162,20 @@ osmium:
   ubuntu: [libosmium-dev]
 paho-mqtt-c:
   alpine: [paho-mqtt-c]
-  debian:
-    bullseye: [libpaho-mqtt1.3]
+  debian: [libpaho-mqtt1.3]
   fedora: [paho-c]
   nixos: [paho-mqtt-c]
   opensuse: [libpaho-mqtt1]
   ubuntu:
-    jammy: [libpaho-mqtt1.3]
+    '*': [libpaho-mqtt1.3]
+    focal: null
+    bionic: null
 paho-mqtt-cpp:
   nixos: [paho-mqtt-cpp]
   ubuntu:
-    jammy: [libpaho-mqttpp3-1]
+    '*': [libpaho-mqttpp3-1]
+    focal: null
+    bionic: null
 pandoc:
   debian: [pandoc]
   fedora: [pandoc]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7163,9 +7163,14 @@ osmium:
 paho-mqtt-c:
   debian:
     bullseye: [libpaho-mqtt1.3]
+  fedora: [paho-c]
+  nixos: [paho-mqtt-c]
+  opensuse: [paho-mqtt]
   ubuntu:
     jammy: [libpaho-mqtt1.3]
 paho-mqtt-cpp:
+  nixos: [paho-mqtt-cpp]
+  opensuse: [libpaho-mqttpp3-1]
   ubuntu:
     jammy: [libpaho-mqttpp3-1]
 pandoc:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7161,16 +7161,16 @@ osmium:
   nixos: [libosmium]
   ubuntu: [libosmium-dev]
 paho-mqtt-c:
+  alpine: [paho-mqtt-c]
   debian:
     bullseye: [libpaho-mqtt1.3]
   fedora: [paho-c]
   nixos: [paho-mqtt-c]
-  opensuse: [paho-mqtt]
+  opensuse: [libpaho-mqtt1]
   ubuntu:
     jammy: [libpaho-mqtt1.3]
 paho-mqtt-cpp:
   nixos: [paho-mqtt-cpp]
-  opensuse: [libpaho-mqttpp3-1]
   ubuntu:
     jammy: [libpaho-mqttpp3-1]
 pandoc:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4639,6 +4639,36 @@ libosmesa6-dev:
   fedora: [mesa-libOSMesa-devel]
   gentoo: ['media-libs/mesa[osmesa]']
   ubuntu: [libosmesa6-dev]
+libpaho-mqtt:
+  alpine: [paho-mqtt-c]
+  debian: [libpaho-mqtt1.3]
+  fedora: [paho-c]
+  nixos: [paho-mqtt-c]
+  opensuse: [libpaho-mqtt1]
+  ubuntu:
+    '*': [libpaho-mqtt1.3]
+    bionic: null
+    focal: null
+libpaho-mqtt-dev:
+  alpine: [paho-mqtt-c-dev]
+  debian: [libpaho-mqtt-dev]
+  fedora: [paho-c-devel]
+  opensuse: [libpaho-mqtt-devel]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -7160,6 +7160,19 @@ osmium:
...
+    bullseye: [libpaho-mqtt1.3]
+  fedora: [paho-c]
+  nixos: [paho-mqtt-c]
+  opensuse: [libpaho-mqtt1]
+  ubuntu:
+    jammy: [libpaho-mqtt1.3]
+paho-mqtt-cpp:
+  nixos: [paho-mqtt-cpp]
+  ubuntu:
+    jammy: [libpaho-mqttpp3-1]
```
adityapande-1995, line 7175: It is generally recommended to have ubuntu and debian keys, and the rest as needed. Is this package available in debian ?

adityapande-1995 (dismissed):

rosdep/base.yaml
lreiher: Yes, I know. Unfortunately only the C library is available for Debian Bullseye as included in this PR. The C++ library has been released for Debian Bookworm though, see [libpaho-mqttpp3-1](https://packages.debian.org/bookworm/libpaho-mqttpp3-1). Bookworm is Bullseye's successor, but still in testing though, so I didn't add it to the PR.

lreiher: @adityapande-1995 Gentle reminder to check on my answer :)

rosdep/base.yaml
```diff
@@ -7160,6 +7160,19 @@ osmium:
   fedora: [libosmium-devel]
   nixos: [libosmium]
   ubuntu: [libosmium-dev]
+paho-mqtt-c:
+  alpine: [paho-mqtt-c]
+  debian:
+    bullseye: [libpaho-mqtt1.3]
```
clalancette, line 7166: Since we expect this package to be available in all debian distributions going forward, this can just be:

```suggestion
  debian: [libpaho-mqtt1.3]
```

rosdep/base.yaml
```diff
@@ -7160,6 +7160,19 @@ osmium:
   fedora: [libosmium-devel]
   nixos: [libosmium]
   ubuntu: [libosmium-dev]
+paho-mqtt-c:
```
clalancette, line 7163: Typically we prefer that these match the names of the Ubuntu/Debian packages.  In this case, that would technically be `libpaho-mqtt1.3`, but I don't think we should encode the version into the package name.  Thus, I would suggest that we make this `libpaho-mqtt`, and the C++ one `libpaho-mqttpp`.

rosdep/base.yaml
```diff
@@ -7160,6 +7160,19 @@ osmium:
...
+  alpine: [paho-mqtt-c]
+  debian:
+    bullseye: [libpaho-mqtt1.3]
+  fedora: [paho-c]
+  nixos: [paho-mqtt-c]
+  opensuse: [libpaho-mqtt1]
+  ubuntu:
+    jammy: [libpaho-mqtt1.3]
```
clalancette, line 7171: Here, we want this key to be forwards-compatible as much as possible.  So this should be:

```suggestion
  ubuntu:
    '*': [libpaho-mqtt1.3]
    focal: null
    bionic: null
```

(same goes for the C++ package below)

clalancette (changes requested):

rosdep/base.yaml
lreiher, line 7175:
```suggestion
    '*': [libpaho-mqttpp3-1]
    focal: null
    bionic: null
```

rosdep/base.yaml
lreiher: I chose the same names that are already used for the releases for Noetic and Galactic, [see here](https://github.com/ros/rosdistro/blob/master/noetic/distribution.yaml#L6759). Do you still suggest the renaming?

lreiher: I have also added the development packages `libpaho-mqtt-dev` and `libpaho-mqttpp-dev` and thus changed the names of the library packages as you suggested.

clalancette (approved):

lreiher:
@clalancette Would you mind merging or do we need to wait for @adityapande-1995?

rosdep/base.yaml
nuclearsandwich: > In this case, that would technically be libpaho-mqtt1.3, but I don't think we should encode the version into the package name. Thus, I would suggest that we make this libpaho-mqtt, and the C++ one libpaho-mqttpp.

I would be interested in figuring out if this would be an acceptable general policy. It's cumbersome to change the version numbers in key names and it's slightly revolting that the key name may include a version number which differs from the version in another package, but what would be done in the case of a debian distribution where there are multiple versions of the library package? Which version would get the unversioned rosdep key and how would we decide that?

I lean heavily towards verbatim debian package names, my internal hesitation is motivated by the fact that the version number is likely incorrect for other platforms much more than it being annoying to update or juggle keys. @tfoote would you be willing to weigh in?

nuclearsandwich (commented):
I'm hesitant to approve because I don't necessarily agree that versions should be dropped from key names when they appear in Debian packages, but all other elements of this PR pass review. Thanks @lreiher for iterating and for your patience while we settle.

rosdep/base.yaml
cmraaron: to ensure the lowest amount of friction for those upgrading from Noetic and Galactic, indeed do we not want to just preserve the stability of these existing names?

lreiher: > to ensure the lowest amount of friction for those upgrading from Noetic and Galactic, indeed do we not want to just preserve the stability of these existing names?

Personally, I am also leaning towards this solution. This would also allow us to easily support ROS2 Galactic and ROS2 Humble+ at the same time with our [mqtt_client[(https://github.com/ika-rwth-aachen/mqtt_clienthttps://github.com/ika-rwth-aachen/mqtt_client), which is currently blocked by this PR.

clalancette: > to ensure the lowest amount of friction for those upgrading from Noetic and Galactic, indeed do we not want to just preserve the stability of these existing names?

I don't think this is something we have to worry about here.  In Noetic, these were released as "vendor" style packages, meaning that they would be something like `ros-noetic-paho-mqtt-c`, with a rosdep key of `paho-mqtt-c`.  These a rosdep keys, which use a different convention (which we are debating here).  I think it is actually a good thing that we are changing that name; it will be clearer that it isn't exactly the same thing as what came from ROS 1.

nuclearsandwich (approved):
My apologies for not getting back to this. There was an internal discussion and the consensus was that omitting the version from the key name was preferred because it:

* introduces an all but guaranteed discrepancy between the key name and the package definition on platforms other than debian/ubuntu and Debian themselves seem to be moving away from always encoding the version in the package name.

PR #36417, opened 2023-03-08
Segno has to be added as a pip package for focal and bionic

gv11181989:
Please modify the following dependency to the rosdep database to add it as a pip package  for focal and bionic.

Package name:
Segno

Package Upstream Source:
https://github.com/heuer/segno

Purpose of using this:
Pure Python QR Code generator with no dependencies

Debian: https://packages.debian.org/bullseye/python3-segno
Ubuntu: https://packages.ubuntu.com/jammy/python3-segno


adityapande-1995 (changes requested):
The review guidelines strongly suggest preffering system packages over pip, is there a specific reason for preferring  pip in this case ? cc @mjcarroll

gv11181989:
> The review guidelines strongly suggest preffering system packages over pip, is there a specific reason for preferring pip in this case ? cc @mjcarroll

Segno has to added as a pip package for focal as there is no debian release for it.

mjcarroll:
If there is no debian release, then this is fine.  Be aware that no other debian releases built in the buildfarm can depend on this rosdep keys.

mjcarroll (approved):

adityapande-1995 (approved):

PR #36453, opened 2023-03-14
stcamera_ros2: 1.0.1

ose-support-ros:

stcamera_ros2

# The source is here:

https://github.com/ose-support-ros/stcamera_ros2.git

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -7717,7 +7717,7 @@ repositories:
       url: https://github.com/tier4/zmqpp_vendor.git
       version: main
     status: developed
-  stcamera_rosw:
+  stcamera_ros2:
     doc:
       type: git
       url: https://github.com/ose-support-ros/stcamera_ros2.git
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -7717,5 +7717,20 @@ repositories:
       url: https://github.com/tier4/zmqpp_vendor.git
       version: main
     status: developed
+  stcamera_rosw:
+    doc:
+      type: git
+      url: https://github.com/ose-support-ros/stcamera_ros2.git
+      version: humble
+    release:
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/ose-support-ros/stcamera_ros2.git
+      version: 1.0.1
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


humble/distribution.yaml
```diff
@@ -6442,6 +6442,21 @@ repositories:
       url: https://github.com/ros2/sros2.git
       version: humble
     status: developed
+  stcamera_components:
+    doc:
+      type: git
+      url: https://github.com/ose-support-ros/stcamera_ros2.git
```
clalancette, line 6448: The source repository is missing a LICENSE file at the top level.  All of the sub-packages seem to define a license of `BSD`, so that is what you should probably put there.

humble/distribution.yaml
```diff
@@ -6442,6 +6442,21 @@ repositories:
...
+      version: humble
+    release:
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/ose-support-ros/stcamera_ros2.git
+      version: 1.0.1
```
clalancette, line 6454: This whole section is not correct for a release.  In order to do a release, you'll need to create a `ros2-gbp` repository and use `bloom-release` to generate the release metadata.  See http://docs.ros.org/en/humble/How-To-Guides/Releasing/Releasing-a-Package.html for detailed information.

If you'd prefer not to do all of that now, you can just add in the `doc` and `source` entries for now, meaning that the package will be available for indexing and document generation.  A binary package won't be available until you follow through with the release instructions, though.

clalancette (dismissed):

humble/distribution.yaml
ose-support-ros: Thank you for your explanation. Since building this package requires a separate manual library installation, we remove the release entry.

humble/distribution.yaml
ose-support-ros: There is a LICENSE file at the top of the repository, but it is not included in each package, so try moving the LICENSE file under each package.

ose-support-ros:
I copied the license file to the top of the repository. Please review it again when you have time.

emersonknapp (approved):
This seems fine now

ose-support-ros:
> This seems fine now

Thank you for your review.

nuclearsandwich (approved):

PR #36469, opened 2023-03-14
Update python-bottle definition for ubuntu focal

jibweb:
Please update the following dependency to the rosdep database: python-bottle

## Package name:

python-bottle

## Package Upstream Source:

System package update for ubuntu

## Purpose of using this:

python-bottle key is already defined in the list, but the package it points to is no longer available on ubuntu focal 

- Ubuntu:
   - https://packages.ubuntu.com/kinetic/python3-bottle
- Debian
   - https://packages.debian.org/stretch/python3-bottle
- Gentoo
   - https://packages.gentoo.org/packages/dev-python/bottle
- Fedora
   - https://packages.fedoraproject.org/pkgs/python-bottle/python3-bottle/
   
# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -866,11 +866,6 @@ python-bottle:
   fedora: [python-bottle]
   gentoo: [dev-python/bottle]
   ubuntu: [python-bottle]
-python3-bottle:
-  debian: [python3-bottle]
-  fedora: [python3-bottle]
-  gentoo: [dev-python/bottle]
-  ubuntu: [python3-bottle]
 python-box2d:
   debian: [python-box2d]
   ubuntu: [python-box2d]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -865,7 +865,9 @@ python-bottle:
   debian: [python-bottle]
   fedora: [python-bottle]
   gentoo: [dev-python/bottle]
-  ubuntu: [python-bottle]
+  ubuntu: 
+    '*': [python-bottle]
+    'focal': [python3-bottle]
```
clalancette, line 870: We shouldn't make this change here, as the `python-` keys are implicitly for Python 2.

Instead, what we should do is to add in a new key which is `python3-bottle`, and then add in the python3 package entries as appropriate.

clalancette (changes requested):

jibweb:
Does that look more appropriate then ?

clalancette (approved):

audrow (approved):

PR #36474, opened 2023-03-14
Add libvtk9-dev rule

sloretz:

## Package name:

`libvtk9.1` and `libvtk9-dev`

## Package Upstream Source:

https://vtk.org/download/

## Purpose of using this:

This is for https://github.com/robotlocomotion/drake-ros . We must use the same version of vtk as Drake: https://github.com/RobotLocomotion/drake/blob/91e3dccea12c6b5fa73ec215d8af0ff4ee7f1876/setup/ubuntu/binary_distribution/packages-jammy.txt#L41

Distro packaging links:

libvtk9-dev
* debian
  * [bullseye](http://deb.debian.org/debian/pool/main/v/vtk9/libvtk9-dev_9.0.1+dfsg1-8_amd64.deb)
* ubuntu
  * [jammy](http://archive.ubuntu.com/ubuntu/pool/universe/v/vtk9/libvtk9-dev_9.1.0+really9.1.0+dfsg2-3build1_amd64.deb)

libvtk9.1
* debian
  * [bookworm](https://packages.debian.org/bookworm/libvtk9.1)
* ubuntu
  * [jammy](http://archive.ubuntu.com/ubuntu/pool/universe/v/vtk9/libvtk9.1_9.1.0+really9.1.0+dfsg2-3build1_amd64.deb)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6208,14 +6208,6 @@ libvtk9-dev:
     '*': [libvtk9-dev]
     bionic: null
     focal: null
-libvtk9.1:
-  debian:
-    '*': [libvtk9.1]
-    bullseye: null
-  ubuntu:
-    '*': [libvtk9.1]
-    bionic: null
-    focal: null
 libvulkan-dev:
   debian: [libvulkan-dev]
   fedora: [vulkan-devel]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6202,6 +6202,20 @@ libvtk-qt:
     xenial: [libvtk6-qt-dev]
     yakkety: [libvtk6-qt-dev]
     zesty: [libvtk6-qt-dev]
+libvtk9-dev:
+  debian: [libvtk9-dev]
+  ubuntu:
+    '*': [libvtk9-dev]
+    bionic: null
+    focal: null
+libvtk9.1:
+  debian:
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -6202,6 +6202,20 @@ libvtk-qt:
     xenial: [libvtk6-qt-dev]
     yakkety: [libvtk6-qt-dev]
     zesty: [libvtk6-qt-dev]
+libvtk9-dev:
+  debian: [libvtk9-dev]
+  ubuntu:
+    '*': [libvtk9-dev]
+    bionic: null
+    focal: null
+libvtk9.1:
```
clalancette, line 6211: Hm.  I know that this is the name of the Debian package, but this seems like far too specific a name.  That is, it is unclear to me that we would ever be able to fill in other entries for this, since a newer Debian/Ubuntu version might add a new version, and Fedora (for instance) might also have a different version.

Could we maybe instead add this to the `libvtk` key above?  That one already has different versions in it.

clalancette: Yeah, that seems like a good compromise here.  I'd be OK with that change.

sloretz: removed the libvtk9.1 rule in 80c2049

audrow (approved):

clalancette (approved):

PR #36278, opened 2023-02-23
swri_console: 2.0.3-1 in 'foxy/distribution.yaml' [bloom]

danthony06:
Increasing version of package(s) in repository `swri_console` to `2.0.3-1`:

- upstream repository: https://github.com/swri-robotics/swri_console.git
- release repository: https://github.com/ros2-gbp/swri_console-release.git
- distro file: `foxy/distribution.yaml`
- bloom version: `0.11.2`
- previous version for package: `2.0.1-1`

## swri_console

```
* Fix QoS on Humble (#55 <https://github.com/swri-robotics/swri_console/issues/55>)
  * Use Humble's Default rosout QoS Settings
  Co-authored-by: David Anthony <mailto:david.anthony@swri.org>
  Co-authored-by: Tony Najjar <mailto:tony.najjar.1997@gmail.com>
* Contributors: David Anthony, Tony Najjar
```

PR #36479, opened 2023-03-15
Reverting swri_console to stable v2.0.1-1

quarkytale:
See #36278 for details.


clalancette (approved):
This is fine, but in the future it is better to use `git revert` for this.  That way, the exact PR that caused the regression will be in the commit message.

clalancette:
@quarkytale I'm going to leave this to you to merge in when you are ready.

PR #36525, opened 2023-03-20
[melodic] Move mir_robot to DFKI-NI

mintar:


clalancette:
I see that https://github.com/dfki-ric/mir_robot.git redirects to https://github.com/DFKI-NI/mir_robot.git , so there is no issue here.  But for posterity, what's the reason for this move?

mintar:
> But for posterity, what's the reason for this move?

Just house cleaning and consolidation of our GitHub repos. Our research group was previously part of DFKI RIC, hence the repo was part of the `dfki-ric` group. Now we split off into our own section called DFKI NI, and we have added several related repos in the `DFKI-NI` GitHub group. To keep everything together, I've decided to move the `mir_driver` repo where it belongs.

clalancette (approved):

PR #36527, opened 2023-03-20
[foxy] Move mir_robot to DFKI-NI

mintar:


quarkytale (approved):
LGTM based on https://github.com/ros/rosdistro/pull/36525#issuecomment-1476574559

PR #36528, opened 2023-03-20
[humble] Move mir_robot to DFKI-NI

mintar:


clalancette:
Same question as https://github.com/ros/rosdistro/pull/36525#issuecomment-1476355766

PR #36530, opened 2023-03-20
[migration-tools] Update release_repo_url value if the release track has one.

nuclearsandwich:
This is an optional configuration option for individual release tracks which is terribly common since most projects get and use the release repo url from the ros/rosdistro distribution.yaml file. The earlier versions of the migration tool did not check for this value (at the time I wrote this migration script I did not even know that such a config setting was possible).

This change should help avoid confusion in the future due to discrepancies between the release repo url in tracks.yaml and in the rosdistro distribution.yaml.

I would consider eliminating this value entirely rather than rewriting it as well, which personally I like better.

### DIFF
--- a/migration-tools/migrate-rosdistro.py
+++ b/migration-tools/migrate-rosdistro.py
@@ -180,6 +180,9 @@ for repo_name in sorted(new_repositories + repositories_to_retry):
                     config = get_patch_config(newref)
                     config['parent'] = config['parent'].replace(args.source, args.dest)
                     set_patch_config(newref, config)
+            # Check for a release repo url in the track configuration
+            if 'release_repo_url' in dest_track:
+                dest_track['release_repo_url'] = new_release_repo_url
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Since I am attempting to resolve all current release repos outside of ros2-gbp before tomorrow's migration I don't consider this a blocker to #36509

cottsay (approved):
I would also rather see the configuration removed than corrected and propagated, but either change is an improvement.

emersonknapp (approved):

nuclearsandwich:
> I would also rather see the configuration removed than corrected and propagated, but either change is an improvement.

I will update this PR to do that then.

nuclearsandwich:
> > I would also rather see the configuration removed than corrected and propagated, but either change is an improvement.
> 
> I will update this PR to do that then.

I've pushed a change which does remove the config value from the destination track when encountering it.

cottsay (approved):

PR #36531, opened 2023-03-20
[foxy] Update schunk_svh urls

stefanscherzinger:
## Background
We developed these libraries in behalf of [Schunk GmbH & Co. KG](https://schunk.com/) (OEM) and now share maintenance. We agreed that the sources should be hosted at [their github space](https://github.com/SCHUNK-GmbH-Co-KG) from now on.

## Steps
This PR updates the respective URLs for *Foxy*.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -6440,7 +6440,7 @@ repositories:
   schunk_svh_library:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


quarkytale (commented):
LGTM, but I'd wait for @tfoote to take a look.

PR #36532, opened 2023-03-20
[humble] Update schunk_svh urls

stefanscherzinger:
## Background
See #36531

## Steps
This PR updates the respective URLs for *Humble*.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -6109,7 +6109,7 @@ repositories:
   schunk_svh_library:
     doc:
       type: git
-      url: https://github.com/fzi-forschungszentrum-informatik/schunk_svh_library.git
+      url: https://github.com/SCHUNK-GmbH-Co-KG/schunk_svh_library.git
       version: main
     release:
       tags:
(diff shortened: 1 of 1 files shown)


humble/distribution.yaml
```diff
@@ -6109,7 +6109,7 @@ repositories:
   schunk_svh_library:
     doc:
       type: git
-      url: https://github.com/fzi-forschungszentrum-informatik/schunk_svh_library.git
+      url: https://github.com/SCHUNK-GmbH-Co-KG/schunk_svh_library.git
```
clalancette, line 6112: This change is fine; no problem to move the repository.

But can I suggest that you actually move it from the `fzi-forschungszentrum-informatik` organization into the `SCHUNK-GmbH-Co-KG` organization?  That way GitHub will automatically do redirects.  Copying it like this makes it confusing for anyone using the old URL (it will still be there, just not be updated).

That said, this is not a blocking comment so if you can't/don't want to do this, we can proceed.

tfoote: +1 for the move which will enable redirects.

tfoote: Though we should still update the urls so I'm going to merge all of these.

PR #36574, opened 2023-03-23
applied pep8 rules

alexanderLinear:
applied style for most of locations reported by this line:

`find . -name '*.py' -not -path '*/.*' -exec pep8 {} --max-line-length 160 --ignore E701 \;`

### DIFF
--- a/doc/scripts/distro_to_rosinstall.py
+++ b/doc/scripts/distro_to_rosinstall.py
@@ -5,16 +5,17 @@ import sys
 import yaml
 from rospkg.distro import load_distro, distro_uri
 
+
 def translate(distro, translate_dir):
     d = load_distro(distro_uri(distro))
     repo_list = d.get_stacks(True)
     for name, item in repo_list.iteritems():
         if item.vcs_config.type == 'svn':
-            rosinstall = [{item.vcs_config.type: \
+            rosinstall = [{item.vcs_config.type:
                            {'local-name': item.name,
                             'uri': item.vcs_config.anon_dev}}]
         else:
-            rosinstall = [{item.vcs_config.type: \
+            rosinstall = [{item.vcs_config.type:
                            {'local-name': item.name,
                             'uri': item.vcs_config.anon_repo_uri,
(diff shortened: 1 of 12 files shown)


alexanderLinear:
@tfoote - can you take a look here?

AlexanderStohr:
hmm - the tooling proposed to merge towards ros:master but the checker seems to choke because there is no ros:master.
galactic might be the default on ros: - but of course there are also e.g. humble or rolling.
:-/

emersonknapp:
Is there any proposal here to add in an enforcer for this choice of formatting style? Doing a pass on the code is all well and good (if one agrees with the choice of formatter, which I have no opinion on in this case) but unless a linter is plugged in  to review PRs moving forward then it won't stick - and if it's not enforced then philosophically I lean against  changing the existing code even if it is a functional no-op.

AlexanderStohr:
> **Validate rosdistro / rosdistro / rosdep checks (3.8) (pull_request) ** Failing after 12m

just a short blink here - not sure if that is "good" already or if this is relevant here... but there are checks:

"Validate rosdistro / rosdistro / rosdep checks (3.8) (pull_request) Failing after 12m"

emersonknapp:
> PEP8 is the base "standard" 

That's fine. It's not what I was trying to comment on. I'm saying - "if you want to enforce a code formatting standard on this repository, then add automated tooling that enforces it". Pytest is being run on the repo in the PR checker, so maybe a pytest that checks pep8 formatting for the repo? Or if that has issues then at least a separate Github Action?

tfoote:
It is useful to conform with the standard. Though there's overhead to general refactors in terms of maintainability. So I agree with Emerson's suggestion that to take this effort we should also add the linter such as flake8 to the unit tests. 

Picking one off the top of my head rosdep has an example: https://github.com/ros-infrastructure/rosdep/blob/master/test/test_flake8.py

sloretz:
Thanks for the PR!

It looks like it's been a little while, and there's now a merge conflict. I'll close this for now, but we can reopen once there's a unit test to make sure the style sticks.

PR #36585, opened 2023-03-23
rsl: 0.2.2-1 in 'humble/distribution.yaml' [bloom]

ChrisThrasher:
Increasing version of package(s) in repository `rsl` to `0.2.2-1`:

- upstream repository: https://github.com/PickNikRobotics/RSL.git
- release repository: https://github.com/ros2-gbp/RSL-release.git
- bloom version: `0.11.2`
- previous version for package: `0.2.1-1`

## rsl

```
* Deprecate rsl::lower_bounds and rsl::upper_bounds
* Depend on ament_cmake_ros for SHARED default
* Upgrade with new pkgs to fix issue with ROS
* Update Catch2
* Contributors: Chris Thrasher, Tyler Weaver
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5967,7 +5967,7 @@ repositories:
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/ros2-gbp/RSL-release.git
-      version: 0.2.1-1
+      version: 0.2.2-1
     source:
       type: git
       url: https://github.com/PickNikRobotics/RSL.git


ChrisThrasher:
> I'm going to revert this because it's creating downstream errors like this: https://build.ros2.org/view/Hbin_uJ64/job/Hbin_uJ64__pick_ik__ubuntu_jammy_amd64__binary/19/.

Are you saying that we can't deprecate anything because it will break downstream projects?

audrow:
Hey @ChrisThrasher, I'm trying to get a humble sync out this week and it looks like this change has caused a few packages to fail to build (because of deprecation warnings).

My primary concern as the maintainer of Humble is to keep regressions out of the distribution. This doesn't mean you can't deprecate things but that you should work with downstream dependencies to fix the fallout. For example, in our buildfarm you can see the downstream packages here: https://build.ros2.org/view/Hbin_uJ64/job/Hbin_uJ64__rsl__ubuntu_jammy_amd64__binary/

One other solution is to do what we do with our core repositories, which is keep separate LTS branches and a rolling branch where deprecation and breaking changes can occur.

PR #36590, opened 2023-03-24
[python.yaml] add pip package pypdf

tlipps:

## Package name:

pypdf

## Package Upstream Source:

https://github.com/py-pdf/pypdf

## Purpose of using this:

pypdf is a free and open-source pure-python PDF library capable of splitting, [merging](https://pypdf.readthedocs.io/en/stable/user/merging-pdfs.html), [cropping, and transforming](https://pypdf.readthedocs.io/en/stable/user/cropping-and-transforming.html) the pages of PDF files. It can also add custom data, viewing options, and [passwords](https://pypdf.readthedocs.io/en/stable/user/encryption-decryption.html) to PDF files. pypdf can [retrieve text](https://pypdf.readthedocs.io/en/stable/user/extract-text.html) and [metadata](https://pypdf.readthedocs.io/en/stable/user/metadata.html) from PDFs as well.

In our case it is utilized to visualize and save ROS data. 

Distro packaging links:

  - bookworm: https://packages.debian.org/bookworm/python3-pypdf 
  - others: via [pip](https://pypi.org/project/pypdf/)
  - lunar: https://packages.ubuntu.com/lunar/python3-pypdf 
  - others: via [pip](https://pypi.org/project/pypdf/)
  - NOT AVAILABLE
  - NOT AVAILABLE
  - AVAILABLE https://packages.gentoo.org/packages/dev-python/pypdf
  - NOT AVAILABLE
  - NOT AVAILABLE
  - NOT AVAILABLE
  - NOT AVAILABLE

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8734,13 +8734,29 @@ python3-pyparsing:
   openembedded: [python3-pyparsing@meta-python]
   opensuse: [python3-pyparsing]
   ubuntu: [python3-pyparsing]
-python3-pypdf-pip:
+python3-pypdf:
   debian:
-    pip:
-      packages: [pypdf]
+    bookworm: [python3-pypdf]
+    stretch:
+      pip:
+        packages: [pypdf]
+    buster:
+      pip:
+        packages: [pypdf]
+    bullseye:
+      pip:
+        packages: [pypdf]
   ubuntu:
-    pip:
-      packages: [pypdf]
+    lunar: [python3-pypdf]
+    bionic:
+      pip:
+        packages: [pypdf]
+    focal:
+      pip:
+        packages: [pypdf]
+    jammy:
+      pip:
+        packages: [pypdf]
 python3-pypng:
   arch: [python-pypng]
   debian: [python3-png]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8734,6 +8734,10 @@ python3-pyparsing:
   openembedded: [python3-pyparsing@meta-python]
   opensuse: [python3-pyparsing]
   ubuntu: [python3-pyparsing]
+python3-pypdf-pip:
+  ubuntu:
+    pip:
+      packages: [pypdf]
 python3-pypng:
   arch: [python-pypng]
   debian: [python3-png]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8735,6 +8735,9 @@ python3-pyparsing:
   opensuse: [python3-pyparsing]
   ubuntu: [python3-pyparsing]
 python3-pypdf-pip:
+  debian:
+    pip:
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
I found https://packages.ubuntu.com/lunar/python3-pypdf2 

We can't accept this as a pip rule. It's already packaged generally upstream and the pip installation would conflict and override the system version. 

If you'd like to research and submit an entry using the system dependencies we'd be happy to review that. 

https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

tlipps:
> I found https://packages.ubuntu.com/lunar/python3-pypdf2
> 
> We can't accept this as a pip rule. It's already packaged generally upstream and the pip installation would conflict and override the system version.
> 
> If you'd like to research and submit an entry using the system dependencies we'd be happy to review that.
> 
> https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules

There are two different `pypdf` packages upstream: `python3-pypdf` and `python3-pypdf2`. The package `python3-pypdf2` contains older versions of `pypdf`: 1.26, 2.10 for ubuntu 18 to 23 , which are very different from the current pip version `pypdf` (3.7.0): 

Quoting from `pypdf`'s GitHub: 
> `NOTE: pypdf>=3.1.0 improved a lot compared to pyPdf<2.0.0 and compared to PyPDF2 < 2.0.0. Please read [the migration guide](https://pypdf.readthedocs.io/en/latest/user/migration-1-to-2.html).` 

So I do not see a reason to create a PR for `python3-pypdf2`, since it is pretty much outdated.

But there still exists a second package called `python3-pypdf` for lunar, which has no ROS build. So I can change it to `python3-pypdf` (referring to https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-mixed-with-system-packages ) instead of `python3-pypdf-pip` but it will only contain pip packages.

tlipps:
> Sorry I assumed pypdf2 was newer than pypdf without looking closer.
> 
> I see that pypdf is in debian as of bookworm too: https://packages.debian.org/bookworm/python3-pypdf
> 
> Please do update it to be lunar and bookworm for the newer platforms and pip for backfilling. These rules will persist long into the future so we need to make sure that we are converging on the native packages.

Done.

tfoote (approved):
Thanks for the update, this now meets the minimum of debian and ubuntu. Is there a package for fedora/rhel or anything else yet?

tlipps:
> Thanks for the update, this now meets the minimum of debian and ubuntu. Is there a package for fedora/rhel or anything else yet?

Found something on gentoo, but otherwise no.

tfoote (approved):

cottsay (approved):

PR #36694, opened 2023-04-04
actuator_msgs: 0.0.1-1 in humble/distribution.yaml

bperseghetti:
# Please Add This Package to be indexed in the rosdistro.

Humble

# The source is here:

https://github.com/rudislabs/actuator_msgs.git

# Checks
 - [x] All packages have a declared license in the package.xml
 - [x] This repository has a LICENSE file
 - [ ] This package is expected to build on the submitted rosdistro

PR #36719, opened 2023-04-06
actuator_msgs: 0.0.1-1 in rolling/distribution.yaml

bperseghetti:

Rolling

# The source is here:

https://github.com/rudislabs/actuator_msgs.git

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -50,6 +50,12 @@ repositories:
       url: https://github.com/ros-drivers/ackermann_msgs.git
       version: ros2
     status: maintained
+  actuator_msgs:
+    source:
+      type: git
+      url: https://github.com/rudislabs/actuator_msgs.git
+      version: main
+    status: maintained
   adaptive_component:
     doc:
       type: git


bperseghetti:
forward port of: https://github.com/ros/rosdistro/pull/36694

emersonknapp:
You only want to put `source` entry in here? No `release` or `doc`?

bperseghetti:
So, I guess that's where I'm slightly confused as I was told to leave the release portion off till it got brought in through: https://github.com/ros2-gbp/ros2-gbp-github-org/pull/239

bperseghetti:
> You only want to put `source` entry in here? No `release` or `doc`?

So, I guess that's where I'm slightly confused as I was told to leave the release portion off till it got brought in through: https://github.com/ros2-gbp/ros2-gbp-github-org/pull/239

bperseghetti:
But whatever the best way is to get this released upstream I'll be more than glad to do.

emersonknapp:
Oh I see, you don't have a release repository yet. If I understand correctly - this `source`  merge will just add this package to the ROS index but it's not going to build any binaries for you on the buildfarm. You could always make a release repo elsewhere, but I'm not 100% sure on what the side effects of doing it outside `ros2-gbp` are. I'll merge this for now and  you can update once you've got the rest of your assets in place.

emersonknapp:
* [ ] New package checks
    * [ ] LICENSE file. 
        * [ ] Either each package in the repository should have a LICENSE file, or 
        * [ ] the source repository should have a top-level LICENSE file.
    * [ ] OSI-approved license [https://opensource.org/licenses](https://opensource.org/licenses) (preferably one of the "Popular licenses").
    * [ ] The license must be reflected in the package.xml file of all sub-packages in the repository.
    * [ ] The source repository must be publicly accessible.
    * [ ] The source repository should contain one or more ROS packages (meaning they have a `package.xml` in the source repository). Packages that are not ROS packages can be accepted, but they are rare and require special handling in the release repository.
    * [ ] The name of the repository and the packages within it must comply with the guidelines in [REP-144](https://www.ros.org/reps/rep-0144.html).

emersonknapp (approved):
This looks good to me, but I'm going to ask to pull in another reviewer to double check on REP-144 naming conventions. @nuclearsandwich @sloretz

bperseghetti:
> 

So is what I was doing with: https://github.com/ros2-gbp/ros2-gbp-github-org/pull/239 not the suggested way?

emersonknapp:
Sorry for any confusion, I think you're doing the right thing. If you followed https://github.com/ros2-gbp/ros2-gbp-github-org/blob/latest/CONTRIBUTING.md then you should be good.

nuclearsandwich:
> So is what I was doing with: [ros2-gbp/ros2-gbp-github-org#239](https://github.com/ros2-gbp/ros2-gbp-github-org/pull/239) not the suggested way?

The actual documentation is still settling. Mostly because the emergent workflow is spread over and pings back and forth between several repositories and so every attempt to document it makes me sad enough to try and fix it and that's not easy either.

But in general before a release repository is created, we want to see a source entry in ros/rosdistro which means that naming review was performed. In most cases I check that on ros2-gbp by first grepping the package name and looking for matches in the recent ROS and ROS 2 distros. In this case I found one added to humble in eee874f940 and noted that when I proceeded with the PR (although I didn't link the commit or PR because I am lazy).

It's definitely not a bad idea to check completely fresh releases for naming but the first/fastest check I use is "is this package already in another distro?" I do this with grep on a local checkout because I find it faster than fooling around in the GitHub webui

nuclearsandwich (approved):

PR #36747, opened 2023-04-10
Adding Python3 to OSX

CursedRock17:
Adding the Python3 package to the OSX version of rosdistro, as it is already required with the ubuntu version and there is a homebrew package.

## Package name:

python3

## Package Upstream Source:

https://www.python.org/

## Purpose of using this:

There are libraries like moveit_ros_planning_interface that utilize python3 as part of rosdep, but it currently doesn't have access. Although the package has a different name it is already on homebrew which makes this simple to add.

- macOS: https://formulae.brew.sh/formula/python@3.11#default

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -283,6 +283,10 @@ libogre-dev:
   osx:
     homebrew:
       packages: [ogre]
+libomp-dev:
+  osx:
+    homebrew:
+      packages: [libomp]
 libopencv-dev:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -283,10 +283,6 @@ libogre-dev:
   osx:
     homebrew:
       packages: [ogre]
-libomp-dev:
-  osx:
-    homebrew:
-      packages: [libomp]
 libopencv-dev:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -650,7 +650,7 @@ protobuf-dev:
 python:
   osx:
     homebrew:
-      packages: []
+      packages: [python, python3]
 python-cairo:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -650,7 +650,7 @@ protobuf-dev:
 python:
   osx:
     homebrew:
-      packages: [python, python3]
+      packages: []
 python-cairo:
   osx:
     homebrew:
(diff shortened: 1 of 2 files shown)


tfoote:
Please rebase this. You appear to have added it on top of #36746

emersonknapp (approved):

rosdep/osx-homebrew.yaml
```diff
@@ -691,6 +691,10 @@ python-vtk:
   osx:
     homebrew:
       depends: [libvtk]
+python3:
+  osx:
+    homebrew:
+      packages: []
```
nuclearsandwich, line 697: Is the package listed intended to be empty?

CursedRock17: No, but I went through and recertified different versions to make it load correctly on python3.

rosdep/osx-homebrew.yaml
```diff
@@ -650,7 +650,7 @@ protobuf-dev:
 python:
   osx:
     homebrew:
-      packages: []
+      packages: [python, python3]
```
nuclearsandwich, line 653: Looking at https://docs.brew.sh/Homebrew-and-Python#python-3y and https://formulae.brew.sh/formula/python@3.11

I actually think we should make the `python` rosdep key, which is implicitly python2, undefined on homebrew, since python2 has been removed.

rosdep/osx-homebrew.yaml
nuclearsandwich: Looking at https://docs.brew.sh/Homebrew-and-Python#python-3y and https://formulae.brew.sh/formula/python@3.11

It seems like the `python@3` or `python3` alias for the latest python3.x formula is an appropriate definition here.

```suggestion
      packages: [python3]
```

nuclearsandwich (dismissed):

rosdep/osx-homebrew.yaml
CursedRock17: Is there an argument to remove it then, or will it just provide users a warning?

nuclearsandwich: The change as proposed is incorrect as both the `python` and `python3` formula names will result in python3 being installed. By removing the definition, any package relying on `python` (implicitly python2, and no longer available on macOS) will see a failure to resolve the rosdep key. Which is the preferred behavior if a dependency is not resolvable.

nuclearsandwich: I pushed [3ae1423](https://github.com/ros/rosdistro/pull/36747/commits/3ae14238aa26b2c3796b169f90def664593a5088) implementing my intended suggestion. There is a difference in rosdep between an "undefined" definition and an "empty" definition. An empty package list means that installing no packages satisfies this key. It is used in situations where the packages listed are part of the core system and therefore not installed by any specific package and other special situations where a package manager may actually be responsible for managing a package. To make the error condition we want for this key on macOS, we have to either assign it a value of `null` or, as I've done here, remove the definition entirely since there are no other definitions for this key in this file.

clalancette (approved):

PR #36750, opened 2023-04-11
Adding fmt on osx

CursedRock17:
Adding fmt to the osx homebrew base of rosdep because ubuntu already has access to this, while osx still needs it.

## Package name:

fmt

## Package Upstream Source:

[TODO link to source repository](https://fmt.dev/latest/index.html)

## Purpose of using this:

fmt is required by many packages such as parameter_traits and ubuntu already uses it in their distribution, so since there's already a homebrew package it should be added.

- macOS: https://formulae.brew.sh/formula/fmt#default

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -111,10 +111,6 @@ fluid:
   osx:
     homebrew:
       packages: [fltk]
-fmt:
-  osx:
-    homebrew:
-      packages: []
 freeimage:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -99,6 +99,10 @@ ffmpeg:
   osx:
     homebrew:
       packages: [ffmpeg]
+fmt:
+  osx:
+    homebrew:
+      packages: []
 flex:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -99,10 +99,6 @@ ffmpeg:
   osx:
     homebrew:
       packages: [ffmpeg]
-fmt:
-  osx:
-    homebrew:
-      packages: []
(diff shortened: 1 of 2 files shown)


emersonknapp (approved):

nuclearsandwich (changes requested):
Please add this definition to the existing key in base.yaml rather than here. See the review comment on https://github.com/ros/rosdistro/pull/36753#pullrequestreview-1386282809 for rationale.

rosdep/osx-homebrew.yaml
```diff
@@ -111,6 +111,10 @@ fluid:
   osx:
     homebrew:
       packages: [fltk]
+fmt:
+  osx:
+    homebrew:
+      packages: []
```
nuclearsandwich, line 117:
```suggestion
```

The changes here are unnecessary and can be dropped.

nuclearsandwich (dismissed):
Thanks for iterating. CI is warning, correctly, of duplicate keys. The empty entry in the rosdep/osx-homebrew.yaml file is spurious. Prior iterations of rosdep maintenance tried to split files up for easier management but it is actually easier (at least for the current maintenance team) to keep all definitions for a key in one place.

CursedRock17:
I appreciate the explanations, I was wondering about package management and the practices this repository followed. I tried to follow other patterns in file after trying to search through the contributing guide. From now on I will attempt to make only clear definitions of packages and their requires.

CursedRock17:
>Prior iterations of rosdep maintenance tried to split files up for easier management but it is actually easier (at least for the current maintenance team) to keep all definitions for a key in one place.

To extend on top of this would it make more sense to just move all definition of osx-homebrew.yaml into their respectable libraries in base.yaml. It would remove this file and prevent more of these duplication issues from happening. I know MacOS is tier 3, but if this is worth doing could I make a general, large PR moving everything over.

clalancette (approved):

PR #36752, opened 2023-04-11
Added range-v3 to OSX

CursedRock17:
This PR looks to add the range-v3 homebrew formula to OSX

## Package name:

range-v3

## Package Upstream Source:

https://ericniebler.github.io/range-v3/

## Purpose of using this:

There are other packages that need this package as there is already a section for it in Ubuntu and others, since there was a homebrew package already, this can be added.

- macOS: https://formulae.brew.sh/formula/range-v3#default

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -703,10 +703,6 @@ qtbase5-dev:
   osx:
     homebrew:
       packages: [qt5]
-range-v3:
-  osx:
-    homebrew:
-      packages: []
 rake:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -703,6 +703,10 @@ qtbase5-dev:
   osx:
     homebrew:
       packages: [qt5]
+range-v3:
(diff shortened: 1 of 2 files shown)


rosdep/osx-homebrew.yaml
```diff
@@ -703,6 +703,10 @@ qtbase5-dev:
   osx:
     homebrew:
       packages: [qt5]
+range-v3:
+  osx:
+    homebrew:
+      packages: []
```
nuclearsandwich, line 709: Is this package list intended to be empty?

nuclearsandwich (changes requested):
Please move this definition to the existing key in base.yaml.

See the review comment in https://github.com/ros/rosdistro/pull/36753#pullrequestreview-1386282809 for rationale.

rosdep/osx-homebrew.yaml
nuclearsandwich, line 709:
```suggestion
```

This section is spurious and the cause of the failed checks in CI

nuclearsandwich (dismissed):

adityapande-1995 (approved):

PR #36753, opened 2023-04-11
Adding libdw to osx

CursedRock17:
This PR aims to add the libdw dependency to osx via homebrew

## Package name:

libdw

## Package Upstream Source:

https://sourceware.org/elfutils/

## Purpose of using this:

This dependency is already needed in packages like backward_ros and is being used on other OS like ubuntu. While it at first seems misleading that it brews with the entire elfutils library instead of just libdw, ubuntu currently uses the exact same tar.bz file via apt that macos would use the homebrew. If needing reference [osx](https://github.com/Homebrew/homebrew-core/blob/b3e60b5e8ef82db4e226184552568c6820a26e14/Formula/elfutils.rb) and [ubuntu](https://packages.ubuntu.com/kinetic/libdw-dev)

- macOS: https://formulae.brew.sh/formula/elfutils#default

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -199,10 +199,6 @@ libdmtx-dev:
   osx:
     homebrew:
       packages: [libdmtx]
-libdw-dev:
-  osx:
-    homebrew:
-      packages: [elfutils]
 libedit:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -199,6 +199,10 @@ libdmtx-dev:
   osx:
     homebrew:
       packages: [libdmtx]
+libdw-dev:
+  osx:
+    homebrew:
+      packages: [elfutils]
 libedit:
   osx:
     homebrew:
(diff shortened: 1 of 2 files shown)


nuclearsandwich (changes requested):
@CursedRock17 it took a few reviews for my macOS memories to come back to me, but our preference for osx/homebrew rosdep keys is now to maintain them in base.yaml and python.yaml alongside the other platforms so that it's easier to track them. Even python.yaml itself may eventually get merged back into base.yaml.

Could you please move this definition into the libdw-dev key in base.yaml.

CursedRock17:
Should all definitions related to homebrew have the osx version completely moved into the python.yaml and base.yaml files? Or can they sit in both?

rosdep/osx-homebrew.yaml
```diff
@@ -199,6 +199,10 @@ libdmtx-dev:
   osx:
     homebrew:
       packages: [libdmtx]
+libdw-dev:
+  osx:
+    homebrew:
+      packages: [elfutils]
```
nuclearsandwich, line 205:
```suggestion
```

This section is spurious ad the cause of the failures in CI checks.

nuclearsandwich (dismissed):

adityapande-1995 (approved):

mjcarroll (approved):

mjcarroll:
@nuclearsandwich any other issues here?

clalancette (approved):

PR #36796, opened 2023-04-12
actuator_msgs: 0.0.1-2 in 'rolling/distribution.yaml' [bloom]

bperseghetti:
Increasing version of package(s) in repository `actuator_msgs` to `0.0.1-2`:

- upstream repository: https://github.com/rudislabs/actuator_msgs.git
- release repository: https://github.com/ros2-gbp/actuator_msgs-release.git
- bloom version: `0.11.2`

## actuator_msgs

```
* Initial commit.
* Contributors: Benjamin Perseghetti
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -51,6 +51,15 @@ repositories:
       version: ros2
     status: maintained
   actuator_msgs:
+    doc:
+      type: git
+      url: https://github.com/rudislabs/actuator_msgs.git
+      version: 0.0.1
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/actuator_msgs-release.git
+      version: 0.0.1-2
     source:
       type: git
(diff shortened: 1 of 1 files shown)


rolling/distribution.yaml
```diff
@@ -51,6 +51,15 @@ repositories:
       version: ros2
     status: maintained
   actuator_msgs:
+    doc:
+      type: git
+      url: https://github.com/rudislabs/actuator_msgs.git
+      version: 0.0.1
```
nuclearsandwich, line 57: This will usually be a branch rather than a tag, since it's reasonable to want to update your documentation without a complete release. That of course relies on your specified branch being compatible with the current release. This isn't ab blocker though just an observation and can be changed in a subsequent manual PR.

nuclearsandwich (approved):

rolling/distribution.yaml
bperseghetti: Ah, good to know for the future, thanks.

PR #36798, opened 2023-04-12
Adding python3_mock to osx

CursedRock17:
This PR adds the python3_mock pack to OSX via pip

## Package name:

python3_mock

## Package Upstream Source:

https://github.com/testing-cabal/mock

## Purpose of using this:

There are packages like urdfdom_py which require python3_mock, also this package is already used in linux builds, with a package on pip.

- macOS: https://pypi.org/project/mock/


nuclearsandwich (approved):
I wonder if this dependency itself should be reconsidered now that mock is in the python3 standard library (and has been since 3.3).

methylDragon (approved):

methylDragon:
> I wonder if this dependency itself should be reconsidered now that mock is in the python3 standard library (and has been since 3.3).

Perhaps, but I don't think it's blocking this particular PR (since the other platforms have it)

PR #36810, opened 2023-04-12
ros_grpc-release: 1.35.0-3 in 'noetic/distribution.yaml' [bloom]

jarkenau:
Increasing version of package(s) in repository `ros_grpc-release` to `1.35.0-3`:

- upstream repository: https://github.com/grpc/grpc
- release repository: https://github.com/DFKI-NI-gbp/ros_grpc-release.git
- bloom version: `0.11.2`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -8810,6 +8810,19 @@ repositories:
       url: https://github.com/shadow-robot/ros_ethercat_eml.git
       version: noetic-devel
     status: maintained
+  ros_grpc-release:
+    doc:
+      type: git
+      url: https://github.com/DFKI-NI-gbp/ros_grpc-release.git
+      version: release/noetic/ros_grpc
+    release:
+      packages:
+      - ros_grpc
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/DFKI-NI-gbp/ros_grpc-release.git
+      version: 1.35.0-3
+    status: maintained
   ros_ign:
     doc:
       type: git


nuclearsandwich (changes requested):
This release of a third party repository appears to conflict with the unconventional third party package already in noetic under the `grpc` name hosted at https://github.com/CogRob/catkin_grpc.git

Were you aware of this existing package and is there a reason that it was unsuitable?

I do not think it makes sense to release this package in Noetic unless there is some reason that the existing one is deficient, and even then, I think that only one instance of this third party package should be in the distribution and thus, we would have to consider removing the existing package for this one to be eligible.

jarkenau:
I saw the existing `grpc` repo. From my point of view, its main purpose is a CMake integration for building the source files from the `.proto` definitions.  However, by looking at the [CMake file](https://github.com/CogRob/catkin_grpc/blob/master/grpc/cmake/generate_proto.cmake#L41-L63), you're still required to install gRPC manually in order to use the repo, since there is no `FetchContent` or `git submodule` of gRPC. 

Moreover, CMake now provides a module for doing this work out of the box: https://cmake.org/cmake/help/latest/module/FindProtobuf.html. So maybe the repo is already outdated? :thinking: 

My intention with this PR was to get gRPC, to install with `rosdep`, so you don't have to install it manually or use the outdated apt packages (which would probably also help the existing gRPC package). I think we could keep both packages since they serve completely different purposes, however the naming would need to change, in order to make their intention/usage clear. 

@BillWSY (maintainer of the existing gRPC packages) can you comment on this?

mintar:
> However, by looking at the [CMake file](https://github.com/CogRob/catkin_grpc/blob/master/grpc/cmake/generate_proto.cmake#L41-L63), you're still required to install gRPC manually in order to use the repo, since there is no FetchContent or git submodule of gRPC.

I thought so too, but actually it fetches and builds version 1.34.1 of `grpc`:

https://github.com/CogRob/catkin_grpc/blob/40c3a270232bf64a519181cad0385bfa8ef5d98b/grpc/CMakeLists.txt#L18-L30

... but it really shouldn't do that; instead, it should use the upstream libraries (i.e., the ones that are packaged with the Linux distro) instead, and only provide `generate_proto` (which can perhaps be replaced by CMake out of the box, as you suggested). Tracked here: https://github.com/CogRob/catkin_grpc/issues/35

PR #36840, opened 2023-04-13
Adding gpgme to OSX

CursedRock17:
This PR adds the libgpgme-dev package to OSX via homebrew.

## Package name:

ibgpgme-dev

## Package Upstream Source:

https://www.gnupg.org/related_software/gpgme/

## Purpose of using this:

There are packages such as rosbag_storage which utilize this library and it is already included in linux builds, so since it's already on homebrew it's worth installing

- macOS: https://formulae.brew.sh/formula/gpgme#default

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -247,6 +247,10 @@ libglew-dev:
   osx:
     homebrew:
       packages: [glew]
+libgpgme-dev:
+  osx:
+    homebrew:
+      packages: [gpgme]
 libgsl:
   osx:
     homebrew:
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -247,10 +247,6 @@ libglew-dev:
   osx:
     homebrew:
       packages: [glew]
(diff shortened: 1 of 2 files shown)


nuclearsandwich (changes requested):
Please move this definition to the existing key in base.yaml.

See the review comment in https://github.com/ros/rosdistro/pull/36753#pullrequestreview-1386282809 for rationale.

rosdep/osx-homebrew.yaml
```diff
@@ -247,6 +247,10 @@ libglew-dev:
   osx:
     homebrew:
       packages: [glew]
+libgpgme-dev:
+  osx:
+    homebrew:
+      packages: [gpgme]
```
nuclearsandwich, line 253:
```suggestion
```

This section is spurious and the cause of failed checks in CI.

nuclearsandwich (dismissed):
Removing the empty definition from the osx-homebrew file should get this PR ready for approval.

methylDragon (approved):

adityapande-1995 (approved):

clalancette (approved):

PR #36857, opened 2023-04-14
[rosdep] add ubuntu-sounds entry

sktometometo:
## Package name:

`ubuntu-sounds` package (Ubuntu's GNOME audio theme)

## Package Upstream Source:

Source package is https://packages.ubuntu.com/source/focal/ubuntu-sounds. I don't where is the original project.

## Purpose of using this:

Our lab have developed [some package](https://github.com/jsk-ros-pkg/jsk_3rdparty/tree/master/ros_speech_recognition) using the sound file in this package. (e.g. `/usr/share/sounds/ubuntu/stereo/bell.ogg`)
And it will be helpful to add the package to rosdep for our package users.

Distro packaging links:

  - Not found
  - https://packages.ubuntu.com/focal/ubuntu-sounds

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8387,6 +8387,8 @@ ttf-sazanami-mincho:
   fedora: [sazanami-mincho-fonts]
   gentoo: [media-fonts/sazanami]
   ubuntu: [ttf-sazanami-mincho]
+ubuntu-sound:
+  ubuntu: [ubuntu-sounds]
 udev:
   arch: [systemd]
   debian: [udev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8387,7 +8387,7 @@ ttf-sazanami-mincho:
   fedora: [sazanami-mincho-fonts]
   gentoo: [media-fonts/sazanami]
   ubuntu: [ttf-sazanami-mincho]
-ubuntu-sound:
+ubuntu-sounds:
   ubuntu: [ubuntu-sounds]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):
I am approving this as what appears to be a previously undeclared dependency for an existing ROS package, but I would say it's very suboptimal to depend on a package that is so specific to just one of the platforms that ROS is available for as it prevents this package from being released on Debian as well as Ubuntu.

If there is another more cross-platform sound library to use, that would be worth considering, or bundling the sounds you need directly in your application if they are appropriately licensed.

sktometometo:
@nuclearsandwich 
Thank you for approving and advise. [sound-theme-freedesktop](https://packages.ubuntu.com/kinetic/sound-theme-freedesktop) package will be more suitable since it is distributed on multiple distribution ([ubuntu](https://packages.ubuntu.com/kinetic/sound-theme-freedesktop), [debian](https://packages.debian.org/bullseye/sound-theme-freedesktop) and already added to [rosdep/base.yaml](https://github.com/ros/rosdistro/blob/39ccc6972a08bc56038dc41a2e0038f8554e65ac/rosdep/base.yaml#L7339-L7343).

tfoote:
Yes that package would be much better to use. If you can use that it would be good to choose this one so that others don't end up going down the same limited path.

audrow:
I'm going to close this since it seems sound-theme-freedesktop is a good solution. Feel free to reopen if need be.

PR #36883, opened 2023-04-17
maliput_ros_interfaces: 0.1.1-1 in 'foxy/distribution.yaml' [bloom]

francocipollone:
Increasing version of package(s) in repository `maliput_ros_interfaces` to `0.1.1-1`:

- upstream repository: https://github.com/maliput/ros2_maliput.git
- release repository: https://github.com/ros2-gbp/maliput_ros_interfaces.git
- bloom version: `0.11.2`

## maliput_ros_interfaces

```
* Adds READMEs to the packages. (#24 <https://github.com/maliput/ros2_maliput/issues/24>)
* Sample lane s route (#21 <https://github.com/maliput/ros2_maliput/issues/21>)
* Routing service (#20 <https://github.com/maliput/ros2_maliput/issues/20>)
* Road position services (#19 <https://github.com/maliput/ros2_maliput/issues/19>)
* Removes tests to avoid breaking CI with generated files. (#13 <https://github.com/maliput/ros2_maliput/issues/13>)
* Adds the translation interface for geometric types. (#4 <https://github.com/maliput/ros2_maliput/issues/4>)
* Adds geometric descriptions and a few basic requests for the entities. (#3 <https://github.com/maliput/ros2_maliput/issues/3>)
* Create a message package (#2 <https://github.com/maliput/ros2_maliput/issues/2>)
* Contributors: Agustin Alba Chicar
```

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2809,6 +2809,26 @@ repositories:
       url: https://github.com/maliput/maliput_py.git
       version: main
     status: developed
+  maliput_ros_interfaces:
+    doc:
+      type: git
+      url: https://github.com/maliput/ros2_maliput.git
+      version: main
+    release:
+      packages:
+      - maliput_ros
+      - maliput_ros_interfaces
+      - maliput_ros_translation
+      tags:
+        release: release/foxy/{package}/{version}
+      url: https://github.com/ros2-gbp/maliput_ros_interfaces.git
+      version: 0.1.1-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/maliput/ros2_maliput.git
+      version: main
+    status: developed
   maliput_sparse:
     doc:
       type: git


nuclearsandwich:
@francocipollone there is something unusual with this release (as well as ) which I should have caught when reviewing https://github.com/ros2-gbp/ros2-gbp-github-org/pull/246

The "unit of release" in bloom (and thus ROS 2) is the repository, rather than the package. Bloom will separate and produce release data for each individual package in the repository and rosdistro and the ros_buildfarm tools will use the per-package artifacts in the release repository when producing binary packages.

I think all of the current release PRs for https://github.com/maliput/ros2_maliput.git should be closed, and I'd even consider deleting the current release repositories and replacing them with a single `maliput_ros-release`

I think it may also be worth naming the source repository `maliput_ros` rather than `ros2_maliput` to match the package naming convention.

nuclearsandwich (changes requested):
The package/repository release conventions aren't completely followed for this release. See https://github.com/ros/rosdistro/pull/36883#issuecomment-1511879034 for recommendations.

francocipollone:
> @francocipollone there is something unusual with this release (as well as ) which I should have caught when reviewing [ros2-gbp/ros2-gbp-github-org#246](https://github.com/ros2-gbp/ros2-gbp-github-org/pull/246)
> 
> The "unit of release" in bloom (and thus ROS 2) is the repository, rather than the package. Bloom will separate and produce release data for each individual package in the repository and rosdistro and the ros_buildfarm tools will use the per-package artifacts in the release repository when producing binary packages.
> 
> I think all of the current release PRs for https://github.com/maliput/ros2_maliput.git should be closed, and I'd even consider deleting the current release repositories and replacing them with a single `maliput_ros-release`
> 
> I think it may also be worth naming the source repository `maliput_ros` rather than `ros2_maliput` to match the package naming convention.

> 

@nuclearsandwich  Oh I see my error. :man_facepalming:  Thanks for the detailed information. Very much appreciated.

Two further question:
 - I now see that under the release repository https://github.com/ros2-gbp/maliput_ros , the three packages' debians are already pushed. Is it necessary then to also close this repository and start over? (Probably needed if I am renaming the source repo to `maliput_ros` rather than `ros2_maliput` to match the package naming convention.)
 
 - For removing/closing the release repositories, I just need to create a PR in `ros2-gbp-github-org` and remove the repositories from the maliput family https://github.com/ros2-gbp/ros2-gbp-github-org/blob/latest/maliput.tf#L22-L24 , right? Or do I have to do something extra later on to actually close the repositories?

nuclearsandwich:
> * now see that under the release repository https://github.com/ros2-gbp/maliput_ros , the three packages' debians are already pushed. Is it necessary then to also close this repository and start over?

If we're going to maintain the name `maliput_ros`, then I do think we could actually leave that repository as is. The "wipe clean and start over" approach is overly broad but it was also conceptually easier than explaining why one of the release repositories could be kept but the others couldn't. However, since you found that same answer on your own. Yes, a more focused resolution would be to keep exactly one of these release repositories online and use it going forward for all releases in this repository.

> (Probably needed if I am renaming the source repo to `maliput_ros` rather than `ros2_maliput` to match the package naming convention.)

The source repository will redirect after a rename for a time, but we could maintain the current `maliput_ros` release repository, but update the bloom configuration in tracks.yaml to use the renamed source repository, then do likewise for the rosdistro release. If we wanted to be extra sure that things were working correctly using the renamed source repository, rather than change the url manually, we could do an additional bloom release which would create a package revision (0.1.1-2) during which we update the source url for bloom and rosdistro. But either the manual edits or the bloom workflow produce the same result. Naming a repository after the "stem" of the package names within is a fairly common practice (i.e. [ros2/rosidl](https://github.com/ros2/rosidl) or [ros2/rclcpp](https://github.com/ros2/rclcpp) so I think `maliput_ros` is a good name for the upstream source repository.

> * For removing/closing the release repositories, I just need to create a PR in `ros2-gbp-github-org` and remove the repositories from the maliput family https://github.com/ros2-gbp/ros2-gbp-github-org/blob/latest/maliput.tf#L22-L24 , right? Or do I have to do something extra later on to actually close the repositories?

It's actually not possible within the terraform scripts to delete a repository (by design) with the rationale that anything which contains a release artifact that was ever part of any rosdistro should be preserved in order to provide artifacts for reproductions of that rosdistro. However, this is a special case as we never merged the releases from the repositories that we won't keep. So they were never part of a rosdistro and thus, in my opinion are eligible for, and indeed recommended for, complete removal in order to avoid future confusion. I'd say the removal process starts with dropping them from the terraform scripts and then I'll do a little behind the scenes magic to release them from the terraform scripts and manually delete them.

PR #36888, opened 2023-04-17
Fix `python-attrs-pip` for Ubuntu

zultron:
Replace Ubuntu in the `python-attrs-pip` key, removed in commit 59517bc9c424.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -620,6 +620,9 @@ python-attrs-pip:
   fedora:
     pip:
       packages: [attrs]
+  ubuntu:
+    pip:
+      packages: [attrs]
 python-autobahn:
   debian: [python-autobahn]
   fedora:


nuclearsandwich:
Hmmm. I think that the status quo is incorrect or at least odd. And opened https://github.com/ros/rosdistro/pull/36893 to address that in a minimally compatible way.

@tfoote was the intent in removing the pip definition for just Ubuntu made under the assumption that the pip key was also only there to backfill for Trusty? If that's the case I think we could consider removing it for Debian as well.

@zultron what platforms are you on that you're relying on this key? All ROS platforms which support python2 are nearing the end of their support, so the best thing to do if you intend to keep using them may be to switch to a fixed copy of the rosdep database.

tfoote:
The fix in #36893 looks like the right fix. The pip rule was used before the debian packages were available. And the debian rule could be extended to support older platforms. But I think that now all the older platforms are eol. So I removed the pip rule because all the expected platforms are now unsupported. So I think that we should keep the pip rule removed.

mjcarroll:
Since there hasn't been a response here, I'm going to close this as fixed by #36893

PR #36755, opened 2023-04-11
Cleanup rosdep rules for long EOL Ubuntu and Debian distros

tfoote:
This is a repository cleanup. I have removed all the specific rules for ubuntu distros older than bionic and debian distros older than stretch. 

* Jessie went [out of LTS support in June 2020](https://www.debian.org/releases/jessie/)
* Xenial [left LTS in April of 2021](https://wiki.ubuntu.com/Releases)
* Ardent was [EOL in 2018](https://wiki.ubuntu.com/Releases)

This will hopefully make maintaining the rosdep database a little bit easier with fewer lines to search through. 

The recommendation if you're using an older EOL platform is to use a tag of the rosdistro from the last sync of the rosdistro.

PR #36893, opened 2023-04-18
Updates to python-attrs after cleanup introduced an oddity.

nuclearsandwich:
This is prompted by https://github.com/ros/rosdistro/pull/36888.

The current state of `master` includes the following stanza:

```
python-attrs:
  # snip...
  ubuntu:
    '*':
      packages: [python-attr]
      pip:
        packages: [attrs]
```

I'm not even sure if this is a valid spelling. I missed this when reviewing https://github.com/ros/rosdistro/pull/36755/ which is probably an object lesson in small reviewable changes. If I'm very responsible then tomorrow morning I'll test what the behavior of this config actually is.

But in any case it seems like an unintended result, or at least one that is worth confirming its intent.

I took a look at the package availability in Debian and Ubuntu. This package is available in Debian Buster and Ubuntu Bionic and Focal but not in later distributions of either. So I've scoped these packages down to those distributions and then likely this entire key can be retired once they're no longer in rosdep's in-scope distros.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -611,8 +611,6 @@ python-attrs:
   ubuntu:
     '*':
       packages: [python-attr]
-      pip:
-        packages: [attrs]
 python-attrs-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This looks like the best solution. It should be on the way out as it is the python 2 package. It must have previously only had specific rules. I thought the linter would catch this.

mjcarroll (approved):

PR #36941, opened 2023-04-25
Adding the Carla install to python.yaml

CursedRock17:
Carla is a dependency that can be installed via pip on osx.

## Package name:

Carla

## Package Upstream Source:

https://github.com/carla-simulator/carla

## Purpose of using this:

Carla is a dependency and may as well be added to osx

- macOS: https://pypi.org/project/carla/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -132,6 +132,9 @@ carla-pip:
   ubuntu:
     pip:
       packages: [carla]
+   osx:
+    pip:
+      packages: [carla]
 casadi-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


adityapande-1995 (approved):

mjcarroll:
@CursedRock17 is it available on other platforms via pip?  Does it make sense to add them now?

CursedRock17:
There is a build distribution for win_amd and Linux 86_64. Since I didn't specify OSX in the pull request name, I don't see any reason not to Windows and a couple other Linux distros to this list.

mjcarroll:
Ah, sorry, I was mistaken, it's already listed for several linux distros here, I think this is good to go.

PR #37099, opened 2023-05-01
Adding rosgpt to documentation index for distro

ZHY109:
i'd like my repo "rosgpt" to be index and documented on ros.rog.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -11633,6 +11633,16 @@ repositories:
       url: https://github.com/xqms/rosfmt.git
       version: master
     status: developed
+  rosgpt:
+    doc:
+      type: git
+      url: https://github.com/ZHY109/rosgpt.git
+      version: main
+    source:
+      type: git
+      url: https://github.com/ZHY109/rosgpt.git
+      version: main
+    status: developed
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
This source repository doesn't meet some of our standards:

1.  It requires the user to `pip` install some packages, which means it can't ever be released on the buildfarm.
1. While it isn't incorrect, having a `src` directory as the top-level directory is strange.  Usually the package is just at the top-level of the repository.
1. The setup in the source repository is pretty strange.  There is a CMakeLists.txt in an intermediate directory (https://github.com/ZHY109/rosgpt/tree/main/src) that shouldn't be there.
1. https://github.com/ZHY109/rosgpt/blob/main/src/rosgpt/scripts/rosgpt.py#L1 is much too specific; this should just be `python3`, otherwise it won't work elsewhere.
1. CI is failing because this PR needs to be rebased onto the latest.

If all of the above issues are fixed, we can consider this one again.  Thanks!

clalancette:
No response in over a month, so I'm going to close this out.  If you'd like to resubmit, please fix the issues identified above.  Thanks.

PR #37118, opened 2023-05-02
 Adding rose2 to documentation index for distro 

Nisici:
I'd like rose2 (https://github.com/aislabunimi/ROSE2) to be indexed and documented on ros.org


mjcarroll:
If you want documentation, you need a `doc:` entry.

Nisici:
> If you want documentation, you need a `doc:` entry.

Thanks, I added `doc:` entry.

Nisici:
> If you want documentation, you need a `doc:` entry.

I think checks failed because I put the wrong indentation after I added doc entry, can you run the checks again? And if it fails again I probably will need your help.

audrow (approved):

PR #36923, opened 2023-04-21
Add Iron distribution as pre-release

Yadunund:
This PR adds the `distribution.yaml` file for `iron` generated by running the migration tool with the command
```bash
python3 migration-tools/migrate-rosdistro.py --dest iron --source rolling --source-ref d7676d5c2ce5a99abb43ac2a3e6b4510fe50a261 --release-org ros2-gbp
```
with `source-ref` being the latest commit to `master` at the time of run.

Some repositories were manually migrated due to issues encountered during the first run.

PR #37119, opened 2023-05-03
[iron] Switch source version to iron for core repos

Yadunund:
PR https://github.com/ros/rosdistro/pull/36923 added a distribution.yaml for `iron` which was generated by running the migration script. However, the source versions for various core ROS 2 repos remained as `rolling`.

For those pkgs with version `iron` in [ros2.repos](https://github.com/ros2/ros2/blob/iron/ros2.repos) file, the corresponding source versions in the iron/distribution.yaml have been updated in this PR.

Once this is merged, we need to update the iron tracks for these packages in the release repos.

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -134,7 +134,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/ament/ament_cmake.git
-      version: rolling
+      version: iron
     status: developed
   ament_cmake_catch2:
     doc:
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
These changes look good, but while we are in here we should also update the `doc` blocks to also point to the `iron` branch.

Yadunund:
Good call. I've updated the doc versions too [9675af8](https://github.com/ros/rosdistro/pull/37119/commits/9675af88c3e6ed51ff219945814d9795fcc6cbc8)

clalancette (approved):

PR #37139, opened 2023-05-05
add libnanopb

Svastits:

## Package name:

libnanop-dev

## Package Upstream Source:

https://github.com/nanopb/nanopb

## Purpose of using this:

nanopb is a commonly used messaging protocol, which is necessary for the ROS2 hardware interface of some KUKA robots

- Debian: 
  - nanopb is still in unstable https://packages.debian.org/sid/libnanopb-dev
- Ubuntu:
  - nanopb: https://packages.ubuntu.com/jammy/libnanopb-dev

UPDATE: I have removed libgrpc, as there is another PR for those packages

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3512,10 +3512,6 @@ libgps:
   opensuse: [gpsd-devel]
   rhel: [gpsd-devel]
   ubuntu: [libgps-dev]
-libgrpc++-dev:
-  debian: [libgrpc++-dev]
-  fedora: [grpc]
-  ubuntu: [libgrpc++-dev]
 libgsl:
   arch: [gsl]
   debian:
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
Please check the failures on the "Files changed" screen, and make appropriate updates.

Svastits:
Added debian bullseye:
https://packages.debian.org/bullseye/libnanopb-dev

rosdep/base.yaml
```diff
@@ -4094,6 +4094,12 @@ libnanoflann-dev:
     '*': [libnanoflann-dev]
     bionic: null
     focal: null
+libnanopb-dev:
+  debian:
+    bullseye: [libnanopb-dev]
```
clalancette, line 4099: Since this is available in all stable versions of Debian, we can just use:

```suggestion
  debian: [libnanopb-dev]
```

clalancette (changes requested):

clalancette (approved):

PR #37147, opened 2023-05-05
mavros: 2.5.0-1 in 'foxy/distribution.yaml' [bloom]

vooon:
Increasing version of package(s) in repository `mavros` to `2.5.0-1`:

- bloom version: `0.11.2`
- previous version for package: `2.4.0-1`

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -2997,10 +2997,11 @@ repositories:
       - mavros
       - mavros_extras
       - mavros_msgs
+      - test_mavros
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/mavlink/mavros-release.git
-      version: 2.4.0-1
+      version: 2.5.0-1
     source:
       test_pull_requests: true
       type: git


foxy/distribution.yaml
```diff
@@ -2997,10 +2997,11 @@ repositories:
       - mavros
       - mavros_extras
       - mavros_msgs
+      - test_mavros
```
mjcarroll, line 3000: This is failing linting due to the `test_mavros` package having a COLCON_IGNORE file in it, I believe.  Did you intend to release the test package, or can it be removed from the release?

vooon: That's a bit unclear to me. Test package provides some useful examples, but it is not ported to ros2 yet. That's why it should be skipped...

mjcarroll: If it should be skipped, then it should not appear in this list.

clalancette: You can also place a `foxy.ignored` file in your release repository to have `bloom` automatically skip the package when releasing.  See https://github.com/ros2-gbp/geometry2-release for an example.

quarkytale:
Hi @vooon, this package is regressing for version 2.4.0 in Foxy, do you want to fix the issues as soon as possible? Foxy is going EOL and we might need to revert it to a stable version if there's no fix for the newest versions.

quarkytale:
Closing this out as reverting to 2.3.0

quarkytale:
Hey @vooon, even 2.3.0 is regressing for `libmavconn`, `mavlink`, `mavros`, `mavros_extras` and `mavros_msgs`, is it possible for you to take a look or I'll probably revert it down further. See https://repo.ros2.org/status_page/ros_foxy_ubv8.html?q=REGRESSION.

clalancette:
So I think the problem here is that `mavlink` also needs to be downgraded.  And it looks like we already synced `2.4.0` out to the main repositories, so we'll want to re-upgrade this to 2.4.0.

@quarkytale I'll contact you this week to try to resolve this.

PR #37158, opened 2023-05-09
add python3-grip key

knorth55:
Add python-grip pip key
## Package name:

grip

## Package Upstream Source:

https://github.com/joeyespo/grip

## Purpose of using this:

launch markdown server

  - pip: https://pypi.org/project/grip/
  - pip: https://pypi.org/project/grip/
  - pip: https://pypi.org/project/grip/
  - pip: https://pypi.org/project/grip/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1843,6 +1843,19 @@ python-graphviz-pip:
 python-gridfs:
   debian: [python-gridfs]
   fedora: [python-pymongo-gridfs]
+python-grip:
+  debian:
+    pip:
+      packages: [grip]
+  fedora:
+    pip:
+      packages: [grip]
+  gentoo:
+    pip:
+      packages: [grip]
+  ubuntu:
+    pip:
+      packages: [grip]
 python-grpc-tools:
   debian:
     '*': [python-grpc-tools]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1844,18 +1844,12 @@ python-gridfs:
   debian: [python-gridfs]
   fedora: [python-pymongo-gridfs]
 python-grip:
-  debian:
-    pip:
-      packages: [grip]
+  debian: [grip]
   fedora:
     pip:
       packages: [grip]
-  gentoo:
-    pip:
-      packages: [grip]
-  ubuntu:
-    pip:
-      packages: [grip]
+  gentoo: [app-text/grip]
+  ubuntu: [grip]
 python-grpc-tools:
   debian:
     '*': [python-grpc-tools]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1843,17 +1843,6 @@ python-graphviz-pip:
 python-gridfs:
   debian: [python-gridfs]
   fedora: [python-pymongo-gridfs]
-python-grip:
-  debian:
-    '*': [grip]
-    bookworm: null
-    bullseye: null
-    buster: [grip]
-  fedora:
-    pip:
-      packages: [grip]
-  gentoo: [app-text/grip]
-  ubuntu: [grip]
 python-grpc-tools:
   debian:
     '*': [python-grpc-tools]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1843,6 +1843,19 @@ python-graphviz-pip:
...
+      packages: [grip]
+  fedora:
+    pip:
+      packages: [grip]
+  gentoo:
+    pip:
+      packages: [grip]
+  ubuntu:
+    pip:
+      packages: [grip]
```
clalancette, line 1858: It looks like this is available as a system package on Ubuntu: https://packages.ubuntu.com/search?keywords=grip&searchon=names&suite=all&section=all .  Please change it to use that package instead.

rosdep/python.yaml
```diff
@@ -1843,6 +1843,19 @@ python-graphviz-pip:
 python-gridfs:
   debian: [python-gridfs]
   fedora: [python-pymongo-gridfs]
+python-grip:
+  debian:
+    pip:
+      packages: [grip]
```
clalancette, line 1849: It looks like this is available on (some) Debian releases as a system package: https://packages.debian.org/search?keywords=grip&searchon=names&suite=all&section=all .  Please fix it to use the system package where possible.

rosdep/python.yaml
```diff
@@ -1843,6 +1843,19 @@ python-graphviz-pip:
...
+  fedora:
+    pip:
+      packages: [grip]
+  gentoo:
+    pip:
+      packages: [grip]
```
clalancette, line 1855: It looks like this is available on gentoo as a system package: https://packages.gentoo.org/packages/app-text/grip

clalancette (changes requested):

rosdep/python.yaml
knorth55: Thank you! I updated.

rosdep/python.yaml
knorth55: Thank you! I updated.

rosdep/python.yaml
knorth55: Thank you! I updated.

rosdep/python.yaml
```diff
@@ -1843,6 +1843,13 @@ python-graphviz-pip:
 python-gridfs:
   debian: [python-gridfs]
   fedora: [python-pymongo-gridfs]
+python-grip:
+  debian: [grip]
```
clalancette, line 1847: Arg, this is annoying.  Looking at https://packages.debian.org/search?keywords=grip&searchon=names&suite=all&section=all , it looks like `grip` is available in debian `buster` (the last release) and `sid` (the development release), but not `bullseye` (the current release) or `bookworm` (the release happening later this year).

In that case, we are going to have to have a more complicated key for debian:

```suggestion
  debian:
    '*': [grip]
    bookworm: null
    bullseye: null
    buster: [grip]
```

clalancette (changes requested):

clalancette (approved):
Thanks, this looks good to me.

rosdep/python.yaml
```diff
@@ -1843,6 +1843,17 @@ python-graphviz-pip:
 python-gridfs:
   debian: [python-gridfs]
   fedora: [python-pymongo-gridfs]
+python-grip:
```
tfoote, line 1846: Sorry one more requested change. We're using the `python3-` prefix to be consistent with python3 being the default. As the `python-` packages are going to go away and we want people to be able to run into things alphabetically. 
Alternatively this could be change to be `grip` as our default is typically that it's the debian/ubuntu package name as the key.

tfoote (dismissed):

rosdep/python.yaml
knorth55: thanks,
i renamed!

clalancette (approved):

nuclearsandwich (approved):

PR #37165, opened 2023-05-10
remote_serial: 1.0.1-1 in 'humble/distribution.yaml' [bloom]

openvmp:
Increasing version of package(s) in repository `remote_serial` to `1.0.1-1`:

- upstream repository: https://github.com/openvmp/serial.git
- release repository: https://github.com/openvmp/serial-release.git
- bloom version: `0.11.2`

## remote_serial

- No changes

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -4615,6 +4615,17 @@ repositories:
       url: https://github.com/ros-controls/realtime_tools.git
       version: master
     status: maintained
+  remote_serial:
+    release:
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/openvmp/serial-release.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/openvmp/serial.git
+      version: main
+    status: developed
   resource_retriever:
     doc:
       type: git


wjwwood:
I talked with @openvmp in person at ROS by the bay and think it's something we should talk about wrt to rep-144 so I've added an agenda item to the next ros 2 meeting.

openvmp:
I shared some feedback that applies to the way all packages are named, from all contributors. Please, don't punish me for that. Either block all submissions from all people or let me submit these changes for now. I have a lot of cool stuff lined up and waiting for this release (such as https://github.com/openvmp/microcontroller/).

Please, note, I'm requesting the package to be called "remote_serial". That's 100% compliant with all existing conventions.

wjwwood:
Ok, follow up from this week's meeting. We've agreed to adjust the language in REP-144 and our recommendation for you is to use your project's namespace in your package name and ideally in your API as well, e.g. `openvmp_serial`, `openvmp_microcontroller`, etc. We think this will be the least confusing and, at the risk of people not knowing these could be (re)used outside of your `openvmp` work, the best alternative.

In the spirit of this, I'll fixup and then rename my serial library as well (may deprecate the existing one to give a transition path for existing users), though in general these things are only applied to new packages rather than retroactively, to avoid unnecessary churn.

If you want to wait for the update to REP-144 you may, but otherwise new bloom pull requests here with the `openvmp_` namespace will welcome.

If you would like to use something else for the name still, e.g. something other than `openvmp_serial`, you may suggest that, but we also agreed this name `remote_serial` was too generic. Ideas off the top of my own head would be something like `serial_ros`, follow the guideline:

> A package name should not contain ros as it is redundant. Exceptions include core packages and ROS bindings of an upstream library (e.g. moveit_ros)

But in that case you really should have a non-ROS base library that's separate, like my `serial` library is currently ROS agnostic.

Or you could consider something like `serial_over_ros`, to better convey that the purpose of the package is to provide access to Serial devices via ROS interfaces.

That being said, those are subject to further discussion and approval, so I'd recommend `openvmp_serial` and if you want to advertise its value as a generic package, make a discourse post announcing its availability and that it can be used outside of the OpenVMP work.

Thanks again for your patience and since this name is not something we'd recommend I will close this pr, but feel free to argue that point if you like, we can always reopen the pr.

wjwwood:
Proposed updates to REP-144: https://github.com/ros-infrastructure/rep/pull/379

PR #37169, opened 2023-05-10
tf2_client: 1.0.0-1 in 'melodic/distribution.yaml' [bloom]

peci1:
Increasing version of package(s) in repository `tf2_client` to `1.0.0-2`:

- upstream repository: https://github.com/tpet/tf2_client.git
- release repository: https://gitlab.fel.cvut.cz/cras/ros-release/tf2_client
- bloom version: `0.11.2`

## tf2_client

```
* Adding default server launcher, with install.
* Cmake refactor, namespaced cpp file, dynamic cast to subclass.
* Fix client/server switch in c++ client.
* Passing private node handle as argument.
* C++ client single node handle, clean up, exporting library.
* Log clearing listener and buffer.
* Allow to clear the listener and buffer if they are not needed any more.
* Fix creating duration object.
* Adding cpp support, clean up of parameter names, conditional param reading.
* Passing Duration object to local Buffer instead of float.
* Initial version (untested).
* Contributors: Martin Pecka, Tomas Petricek
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13851,6 +13851,21 @@ repositories:
       url: https://github.com/locusrobotics/tf2_2d.git
       version: devel
     status: developed
+  tf2_client:
+    doc:
+      type: git
+      url: https://github.com/tpet/tf2_client.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://gitlab.fel.cvut.cz/cras/ros-release/tf2_client.git
+      version: 0.0.0-1
+    source:
+      type: git
+      url: https://github.com/tpet/tf2_client.git
+      version: master
+    status: maintained
   tf2_server:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -13851,6 +13851,21 @@ repositories:
...
+  tf2_client:
+    doc:
+      type: git
+      url: https://github.com/tpet/tf2_client.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://gitlab.fel.cvut.cz/cras/ros-release/tf2_client.git
+      version: 0.0.0-1
```
clalancette, line 13863: We don't usually like to release packages with version 0.0.0; it is pretty meaningless.  Can you please update this to have some kind of non-zero number?

clalancette (dismissed):

melodic/distribution.yaml
peci1: I don't have write access to that package since it's from one of my former colleagues who doesn't respond to my requests to give me write access.

I could probably change the version in the release repo via patches. Would that be okay?

peci1: Okay, I've updated this PR and made the changes as release patches.

quarkytale (approved):

peci1:
I've finally got access to the upstream repo, so I switched to standard release process and dropped the release repo changes. The license file has been added.

nuclearsandwich (approved):

PR #37174, opened 2023-05-10
Remove toposens from kinetic/melodic/noetic distributions

tcptomato:
The toposens nodes lack a maintainer and the git repositories are unfortunately no longer available.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13935,44 +13935,6 @@ repositories:
       url: https://gitlab.com/InstitutMaupertuis/topics_rviz_plugin.git
       version: melodic
     status: maintained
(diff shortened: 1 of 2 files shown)


clalancette:
> This pull request changes files for a ROS distribution that is no longer supported (End Of Life) and the distribution will not have any future syncs to include this change.

Which means that we can't merge this.  Please update this to remove the kinetic changes, then we can include it.

tcptomato:
Updated the PR.

clalancette (approved):
None of these repositories seem to be available anymore, so they wouldn't build.  So I'll go ahead and merge this, thanks.

tcptomato:
Thanks.

PR #37180, opened 2023-05-11
draco: 1.3.6-3 in 'melodic/distribution.yaml' [bloom]

peci1:
Increasing version of package(s) in repository `draco` to `1.3.6-3`:

- upstream repository: https://github.com/google/draco
- release repository: https://gitlab.fel.cvut.cz/cras/ros-release/draco.git
- bloom version: `0.11.2`

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2553,10 +2553,6 @@ repositories:
         release: release/melodic/{package}/{version}
       url: https://gitlab.fel.cvut.cz/cras/ros-release/draco.git
       version: 1.3.6-3
-    source:
-      type: git
-      url: https://github.com/google/draco.git
-      version: 1.3.6
     status: maintained
   driver_common:
     doc:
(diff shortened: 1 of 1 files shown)


peci1:
This is a vendored pointcloud compression library from Google. The released version 1.3.6 is the latest that works with Bionic's CMake.

quarkytale:
It's not optimal to have the source entry removed, please consider getting the tags right in the source repository.

quarkytale (approved):

peci1:
> It's not optimal to have the source entry removed, please consider getting the tags right in the source repository.

That doesn't make sense. There is no branch for the older 1.3.x release, only the master branch pointing to 1.5.x and feature branches. And version 1.5.x cannot be built on U 18.04 because of too old CMake.

I'm about to prepare a Noetic release, where, if possible, I'd like to release the 1.5.x version. In that case, I'd add the source entry, too.

PR #37218, opened 2023-05-11
Add rosdep key for 'semgrep'

florcabral:

## Package name:

`semgrep`

## Package Upstream Source:

https://github.com/returntocorp/semgrep

## Purpose of using this:

Semgrep is an open source static analysis engine. I have developed a [wrapper for the ament build system](https://github.com/florcabral/ament_semgrep) to run Semgrep in ROS projects. For this package to build, the `semgrep` rosdep key is required.

Distro packaging links:

  - https://formulae.brew.sh/formula/semgrep
- pip
  - https://pypi.org/project/semgrep/

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7324,6 +7324,11 @@ sdl2-ttf:
   gentoo: [media-libs/sdl2-ttf]
   nixos: [SDL2_ttf]
   ubuntu: [libsdl2-ttf-dev]
+semgrep:
+  osx:
+    homebrew:
+      packages: [semgrep]
+  pip: [semgrep]
 setserial:
   debian: [setserial]
   fedora: [setserial]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -7324,6 +7324,11 @@ sdl2-ttf:
   gentoo: [media-libs/sdl2-ttf]
   nixos: [SDL2_ttf]
   ubuntu: [libsdl2-ttf-dev]
+semgrep:
+  osx:
+    homebrew:
+      packages: [semgrep]
+  pip: [semgrep]
```
clalancette, line 7331: This is not how the `pip` key works; you'll instead need to split this out for each distribution (debian, ubuntu, etc), and then put a pip key underneath it.

Even then, you should know that you cannot release packages that depend on `pip` keys into ROS.  All of the dependencies must be available via system packages.

clalancette (changes requested):

nuclearsandwich (approved):

clalancette (approved):

PR #37256, opened 2023-05-17
robot_localization: 3.4.2-1 in 'humble/distribution.yaml' [bloom]

SteveMacenski:
Increasing version of package(s) in repository `robot_localization` to `3.4.2-1`:

- upstream repository: https://github.com/cra-ros-pkg/robot_localization.git
- release repository: https://github.com/ros2-gbp/robot_localization-release.git
- bloom version: `0.11.2`
- previous version for package: `3.5.0-1`

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5097,7 +5097,7 @@ repositories:
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/ros2-gbp/robot_localization-release.git
-      version: 3.5.0-1
+      version: 3.4.2-1
     source:
       test_pull_requests: true
       type: git


SteveMacenski:
This fixes a mistake I made in filling in bloom on Humble's final release breaking ABI which needs to be reverted and matched with the correct final Humble version

audrow:
@SteveMacenski, it turns out doing this creates some weirdness for users that did install 3.5.0 with apt. As I understand, anyone with 3.5.0 will have to explicitly downgrade this package since apt will compare the versions and see that the user has a higher version. They can get around this by explicitly installing the newer version, but won't be able to do the normal upgrade path (`apt upgrade` to get the most recent). 

This will also make the update path weird for patch releases, which if users have 3.5.0 installed, won't be upgraded unless they explicitly install an older version.

It looks like there was a period of about a month where 3.5.0 was the latest release so there could be a large number of users who will have a confusing upgrade path. I think the best path is to bump the minor version used by Humble and Rolling twice each (Humble => 3.6.0 and rolling => 3.7.0) so that users have a clear upgrade path with `apt upgrade` and you can make patch releases to Humble without getting in trouble with conflicting versions between Humble and Rolling. I get that there is some weirdness in this option, too, but believe it is the most user friendly option that I can think of.

That being said, if you don't think this is worth it and that many users won't be caught in the weird upgrade path of 3.5.0 or that they'll be able to figure it out, we can proceed as things are. I was planning to do a Humble sync this week and I noticed this, let me know what you think is the best way to proceed. 

If you do make new releases, feel free to @ me to get them in quickly.

PR #37335, opened 2023-05-23
Update python3-tqdm in python.yaml

belorenz:
Moved from apt to pip dependency for `python3-tqdm` in ubuntu focal because last apt version available in focal is from Mai 2019 latest pip version is from March 23.

## Package Upstream Source:

[repo](https://github.com/tqdm/tqdm)

## Purpose of using this:

Instantly make your loops show a smart progress meter. Looking good!

Distro packaging links:

- Debian: https://packages.debian.org/buster/python3-tqdm
  - REQUIRED
- Ubuntu: https://packages.ubuntu.com/focal/python/python3-tqdm
  - REQUIRED
- PyPi: https://pypi.org/project/tqdm/

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8652,7 +8652,7 @@ python3-tqdm:
     '*': [python3-tqdm]
     'focal':
       'pip':
-         packages: [tqdm]
+        packages: [tqdm]
 python3-transformers-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8650,6 +8650,9 @@ python3-tqdm:
       packages: [tqdm]
   ubuntu:
     '*': [python3-tqdm]
+    'focal':
+      'pip':
+         packages: [tqdm]
 python3-transformers-pip:
   debian:
     pip:


rosdep/python.yaml
```diff
@@ -8650,6 +8650,9 @@ python3-tqdm:
       packages: [tqdm]
   ubuntu:
     '*': [python3-tqdm]
+    'focal':
+      'pip':
+         packages: [tqdm]
```
belorenz, line 8655:
```suggestion
        packages: [tqdm]
```

peci1:
I don't support this change. The general rule is that APT packages have strong preference to pip packages unless there is a very good reason. This would also break buildfarm builds of all existing packages that depend on the rosdep key because buildfarm cannot install pip packages.

If you need a newer version, you have to manually install it on the machines.

Alternatively, you could create a new rosdep key `python3-tqdm-pip`, but I'm not sure what's the policy on such dualities...

nuclearsandwich (changes requested):
As a policy, we generally discourage the use of pip dependencies on platforms which provide those via system packages and replacing an existing system dependency with a pip dependency is completely off the table within an existing key.

Sticking with released dependencies for the lifetime of a distribution is part of the trade-off that allows us to leverage that distribution's support and inter-compatibility.

emersonknapp:
I will also recommend just closing out this PR - as noted above it doesn't meet the standards for rosdistro, and the existing code already does without a change.

PR #37384, opened 2023-05-26
urg_node_msgs: update branch names

mikeferguson:
We have a first change in like 5 years in this package - so it was time to branch off "Iron and earlier", and then rolling moving forward. Took the opportunity to also change from master -> main.

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -7589,7 +7589,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-drivers/urg_node_msgs.git
-      version: master
+      version: iron
     release:
       tags:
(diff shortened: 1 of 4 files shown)


mikeferguson:
I added back the master branch (since it seems that broke the validate rosdistro) - someone will have to retrigger that CI

nuclearsandwich:
While Humble is in [sync hold](https://discourse.ros.org/t/preparing-for-humble-sync-2023-05-30/31622) this change won't be eligible for a merge.

nuclearsandwich:
> I added back the master branch (since it seems that broke the validate rosdistro) - someone will have to retrigger that CI

Re-running failed actions.

audrow (approved):

PR #37449, opened 2023-06-02
Removed turtlesim key

ahcorde:
Removed `turtlesim` key. Related with the comment from @clalancette in this PR https://github.com/ros/ros_tutorials/pull/153. 

If I'm not wrong we should remove the key in rolling and then release the package with the right key, isn't it ?

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4838,6 +4838,8 @@ repositories:
       url: https://github.com/ros/ros_tutorials.git
       version: rolling
     release:
+      packages:
+      - turtlesim
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/ros_tutorials-release.git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4832,6 +4832,22 @@ repositories:
       url: https://github.com/ros2/ros_testing.git
       version: rolling
     status: maintained
+  ros_tutorials:
+    doc:
+      type: git
+      url: https://github.com/ros/ros_tutorials.git
+      version: rolling
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/ros_tutorials-release.git
+      version: 1.7.1-1
+    source:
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
I think we might get away here with an inline edit rather than having to remove and replace.

Suggestion

```
  ros_tutorials:
    doc:
      type: git
      url: https://github.com/ros/ros_tutorials.git
      version: rolling
    release:
      packages:
        - turtlesim
      tags:
      url: https://github.com/ros2-gbp/ros_tutorials-release.git
      version: 1.7.1-1
    source:
      test_pull_requests: true
      type: git
      url: https://github.com/ros/ros_tutorials.git
      version: rolling
    status: maintained
```

clalancette:
> I think we might get away here with an inline edit rather than having to remove and replace.

Yeah, I think that would be better.  My suggestion is the following:

1.  Make the update as @nuclearsandwich suggests.
2.  Merge this in.
3.  Immediately do a new `bloom-release ros_tutorials` of the package to verify that this does what we expect.

We may then want to consider backporting this to both `humble` and `iron`.

ahcorde:
Thank you @clalancette and @nuclearsandwich

rolling/distribution.yaml
```diff
@@ -4832,6 +4832,22 @@ repositories:
...
+  ros_tutorials:
+    doc:
+      type: git
+      url: https://github.com/ros/ros_tutorials.git
+      version: rolling
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/ros_tutorials-release.git
+      version: 1.7.1-1
```
clalancette, line 4844: I think this is going to need to be:

```suggestion
    release:
      packages:
        - turtlesim
      tags:
        release: release/rolling/{package}/{version}
      url: https://github.com/ros2-gbp/ros_tutorials-release.git
      version: 1.7.1-1
```

(though I could use confirmation from @nuclearsandwich )

clalancette (changes requested):

clalancette (approved):
This looks good to me.

PR #37621, opened 2023-06-13
Cubicleguy patch 1

cubicleguy:

melodic

# The source is here:

https://github.com/cubicleguy/ess_imu_ros1_spi_driver.git

# Checks

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -3004,7 +3004,7 @@ repositories:
       tags:
         release: release/melodic/{package}/{version}
       url: https://github.com/cubicleguy/ess_imu_ros1_spi_driver-release.git
-      version: 1.3.0-1
+      version: 1.3.0-2
     source:
       type: git
       url: https://github.com/cubicleguy/ess_imu_ros1_spi_driver.git
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -2995,6 +2995,21 @@ repositories:
       url: https://github.com/enabled-robotics/er_public_msgs-release.git
       version: 1.4.0-1
     status: maintained
+  ess_imu_ros1_spi_driver:
+    doc:
+      type: git
+      url: https://github.com/cubicleguy/ess_imu_ros1_spi_driver.git
+      version: main
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://github.com/cubicleguy/ess_imu_ros1_spi_driver-release.git
+      version: 1.3.0-1
+    source:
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -2995,6 +2995,21 @@ repositories:
       url: https://github.com/enabled-robotics/er_public_msgs-release.git
       version: 1.4.0-1
     status: maintained
+  ess_imu_ros1_spi_driver:
```
clalancette, line 2998: I think that this package name likely violates https://www.ros.org/reps/rep-0144.html#global-guidelines, point 3.  That is, it should probably not contain `ros` in the name, and almost certainly should not contain `ros1` in the name.

clalancette (changes requested):

melodic/distribution.yaml
cubicleguy: Yes, I agree to both points. 
If I must make some changes,
One issue is that there is already an pre-existing "ess_imu_ros1_uart_driver" accepted in the melodic distro.
And I have a couple of mirrored pkgs for ROS2 which are in the queue to release.

I'm open to suggestions. Would you recommend to rename to 2 repos "ess_imu_spi_driver" and "ess_imu_uart_driver".
Then within each repo have a ros branch and ros2 branch?
Or consolidate it all into 1 repo with 4 branches: ros uart, ros spi, ros2 uart, ros2 spi

clalancette: Sorry for the delay here.

> I'm open to suggestions. Would you recommend to rename to 2 repos "ess_imu_spi_driver" and "ess_imu_uart_driver". Then within each repo have a ros branch and ros2 branch? Or consolidate it all into 1 repo with 4 branches: ros uart, ros spi, ros2 uart, ros2 spi

I think either is OK.  Having a single repository with a bunch of packages in it will probably make it easier to manage in the future, so I'd suggest that.  But both work.

clalancette: That said, Melodic went [End-of-life](https://discourse.ros.org/t/new-packages-for-melodic-2023-06-27/32121/2) today, so I'm going to close this one out.  I'll suggest that you revamp things as mentioned above, and then open PRs to target noetic (and ROS 2 Humble, Iron, and Rolling) instead.

cubicleguy: Yeah, I guess I need to re-org the repo. 
Sorry for wasting some of your bandwidth

clalancette: No worries at all!

PR #37626, opened 2023-06-13
feat: add  pip MyST-Parser  pip dependency

rickstaa:

## Package name:

MyST-Parser

## Package Upstream Source:

https://github.com/executablebooks/MyST-Parser

## Purpose of using this:

This PR adds the [myst-parser](https://pypi.org/project/myst-parser/) pip package as a python dependency. This package can be used to parse markdown for sphinx-based documentation projects. It can be useful to include as a `<doc_depend>` in the `package.xml`.

**Distro packaging links:**

PIP: https://pypi.org/project/myst-parser/

**N/a**

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6749,6 +6749,19 @@ python3-mypy:
     '7': null
     '8': null
   ubuntu: [python3-mypy]
+python3-myst-parser:
+  debian:
+    pip:
+      packages: [myst-parser]
+  fedora:
+    pip:
+      packages: [myst-parser]
+  osx:
+    pip:
+      packages: [myst-parser]
+  ubuntu:
+    pip:
+      packages: [myst-parser]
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

rosdep/python.yaml
```diff
@@ -6749,6 +6749,19 @@ python3-mypy:
     '7': null
     '8': null
   ubuntu: [python3-mypy]
+python3-myst-parser:
```
clalancette, line 6752: Since this is a pip-only key, this should be:

```suggestion
python3-myst-parser-pip:
```

clalancette (changes requested):

rosdep/python.yaml
rickstaa: Your right thanks!

clalancette (approved):

PR #37646, opened 2023-06-14
roscpp_core: Change branch to melodic-devel

peci1:
I'd like to switch roscpp_core on Melodic to melodic-devel branch for the last sync.


peci1:
Accompanied by https://github.com/ros-gbp/roscpp_core-release/pull/2 in the release repo. Not sure which one should go first (my guess is it doesn't matter).

peci1:
Closing because we have agreed that such changes can't be merged in the last Melodic sync.

PR #37661, opened 2023-06-14
Update libtool for debian in base.yaml

hellantos:

## Package name:

libtool

## Package Upstream Source:

https://www.gnu.org/software/libtool/libtool.html

## Purpose of using this:

Adding bullseye support and shortening debian entry through wildcard.

Distro packaging links:

  - libtool-bin: https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=libtool-bin
  - libtool: https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=libtool
  - libltdl-dev: https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=libltdl-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5638,10 +5638,9 @@ libtins-dev:
 libtool:
   arch: [libtool]
   debian:
-    buster: [libtool, libltdl-dev, libtool-bin]
+    '*': [libtool, libltdl-dev, libtool-bin]
     lenny: [libtool, libltdl3-dev]
     squeeze: [libtool, libltdl-dev]
-    stretch: [libtool, libltdl-dev, libtool-bin]
     wheezy: [libtool, libltdl-dev]
   fedora: [libtool, libtool-ltdl-devel]
   freebsd: [libtool]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5637,11 +5637,7 @@ libtins-dev:
   ubuntu: [libtins-dev]
 libtool:
   arch: [libtool]
-  debian:
-    '*': [libtool, libltdl-dev, libtool-bin]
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

rosdep/base.yaml
```diff
@@ -5638,10 +5638,9 @@ libtins-dev:
 libtool:
   arch: [libtool]
   debian:
-    buster: [libtool, libltdl-dev, libtool-bin]
+    '*': [libtool, libltdl-dev, libtool-bin]
     lenny: [libtool, libltdl3-dev]
     squeeze: [libtool, libltdl-dev]
-    stretch: [libtool, libltdl-dev, libtool-bin]
```
clalancette, line 5644: Actually, I think that all of the specifically-listed Debian distributions here are End-of-Life.  Thus I'd suggest we just simplify this key to:

```
  debian: [libtool, libltdl-dev, libtool-bin]
```

While we are in here, we may as well update the `ubuntu` key the same way:

```
  ubuntu: [libtool, libtldl-dev, libtool-bin]
```

clalancette (changes requested):

rosdep/base.yaml
ipa-cmh: Totally agree.
I thought to do this but did not know the policy for deleting old keys.

clalancette: > I thought to do this but did not know the policy for deleting old keys.

To be honest, we don't have a hard-and-fast rule about removing support for old distributions.  Every time we do, somebody complains.

But in this case, the old distros are *very* old, so I think removing them is fine.

ipa-cmh:
@clalancette 
Updated now to suggested keys.

clalancette (approved):

PR #37702, opened 2023-06-16
add python-qpsolvers-pip to rosdep database

Qrpucp:

## Package name:

qpsolvers

## Package Upstream Source:

https://pypi.org/project/qpsolvers/

## Purpose of using this:

This dependency is being used for solving Quadratic Programming (QP) problem.

Distro packaging links:

- Debian: https://pypi.org/project/qpsolvers/
- Ubuntu: https://pypi.org/project/qpsolvers/
- macOS: https://pypi.org/project/qpsolvers/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3373,6 +3373,16 @@ python-pyzbar-pip:
   ubuntu:
     pip:
       packages: [pyzbar]
+python-qpsolvers-pip:
+  debian:
+    pip:
+      packages: [qpsolvers]
+  osx:
+    pip:
+      packages: [qpsolvers]
+  ubuntu:
+    pip:
+      packages: [qpsolvers]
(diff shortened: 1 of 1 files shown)


adityapande-1995 (approved):

rosdep/python.yaml
```diff
@@ -3373,6 +3373,16 @@ python-pyzbar-pip:
   ubuntu:
     pip:
       packages: [pyzbar]
+python-qpsolvers-pip:
```
clalancette, line 3376: Our current best practice is to make these keys prefixed with `python3`:

```suggestion
python3-qpsolvers-pip:
```

(note that you'll also have to move this block so that it is in alphabetical order)

clalancette (changes requested):

clalancette (approved):

PR #37709, opened 2023-06-18
electronic-io: 1.0.1-1 in 'noetic/distribution.yaml' [bloom]

peci1:
Increasing version of package(s) in repository `electronic-io` to `1.0.1-1`:

- upstream repository: https://github.com/ctu-vras/electronic-io.git
- release repository: https://gitlab.fel.cvut.cz/cras/ros-release/electronic-io.git
- bloom version: `0.11.2`

## electronic_io

```
* Added support for simplified specification of pins in config.
* Added BinaryPWM virtual pin.
* Added virtual pins.
* It topic is set to empty, the topics and services will not be created.
* Added Battery device.
* Export the provided devices.
* Allow setting queue_size and latching for output devices.
* Added support for output groups.
* Robustified IO board client.
* Initial commit
* Contributors: Martin Pecka
```

## electronic_io_msgs

```
* Initial commit
* Contributors: Martin Pecka
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2300,6 +2300,24 @@ repositories:
       url: https://github.com/stack-of-tasks/eiquadprog.git
       version: devel
     status: maintained
+  electronic-io:
+    doc:
+      type: git
+      url: https://github.com/ctu-vras/electronic-io.git
+      version: master
+    release:
+      packages:
+      - electronic_io
+      - electronic_io_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://gitlab.fel.cvut.cz/cras/ros-release/electronic-io.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/ctu-vras/electronic-io.git
+      version: master
+    status: developed
   eml:
     release:
       tags:


clalancette:
See my comments in https://github.com/ros/rosdistro/pull/37708#pullrequestreview-1488672975 about this package.

clalancette:
Reproducing the conversation from #37708 

clalancette:
```
This name seems pretty generic to me, and thus I think it probably violates
https://www.ros.org/reps/rep-0144.html#global-guidelines point 1 or 4. It's kind of subjective,
as this is a generic package, but it isn't clearly used by large parts of the community. Thus I think
this package would be a candidate for prefixing the name of the entity.
```

peci1:
```
The goal was indeed to create a generically usable package.
I wanted to promote it a bit in the coming weeks so the community would get to know it.

I understand why you suggest prepending organization. I have two points which led me to not do so:

1.    There are (imo) really no competing packages trying to achieve this in a generic way. Yet I think some standardization would be beneficial.
2.    Was there actually some package that started with organzation prefix and then dropped it over time? I get that this is a REP recommendation, yet I think it is actually very impractical to force users of an established package to use a different name.

One point for prefixing is the dependency on cras_msgs. However, this package just adds what's missing in sensor_msgs (Voltage, ElectricCurrent...).

If you still think this package should be prefixed, I'll do it.
```

mjcarroll:
```
I think that if it's intended to be generic, then the package name is fine.
```

clalancette:
> 2.    Was there actually some package that started with organzation prefix and then dropped it over time? I get that this is a REP recommendation, yet I think it is actually very impractical to force users of an established package to use a different name.

Yes, that is a big problem with that recommendation.  On the other hand, we really have no other way to do things since we need to have package names be globally unique.  I think the most likely scenario is that prefixed packages will remain prefixed forever, which isn't great but also isn't a huge deal (in my opinion).

As for this particular package, I'm OK with adding it under the current name.  So I'll go ahead and do the rest of the review on it.

clalancette (approved):

PR #37712, opened 2023-06-19
compass: 1.0.1-1 in 'melodic/distribution.yaml' [bloom]

peci1:
Increasing version of package(s) in repository `compass` to `1.0.1-1`:

- upstream repository: https://github.com/ctu-vras/compass.git
- release repository: https://gitlab.fel.cvut.cz/cras/ros-release/compass.git
- bloom version: `0.11.2`

## compass

```
* Made use of cras_cpp_common's node_from_nodelet CMake macro.
* Removed imu_transformer workaround
* Added option to force magnetic declination value.
* Added visualization of azimuth.
* Fixed IMU orientation covariance transformation.
* Renamed magnetometer_compass package to compass.
* Added variance to magnetometer compass outputs. It now also outputs fully valid georeferenced IMU messages instead of just writing the orientation into them.
* Contributors: Martin Pecka
```

## compass_msgs

```
* Added variance to Azimuth message.
* Added improved version of compass.
* Contributors: Martin Pecka
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -1860,6 +1860,24 @@ repositories:
       url: https://github.com/ros/common_tutorials.git
       version: indigo-devel
     status: maintained
+  compass:
+    doc:
+      type: git
+      url: https://github.com/ctu-vras/compass.git
+      version: master
+    release:
+      packages:
+      - compass
+      - compass_msgs
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://gitlab.fel.cvut.cz/cras/ros-release/compass.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/ctu-vras/compass.git
+      version: master
+    status: maintained
   computer_status_msgs:
     doc:
       type: git


melodic/distribution.yaml
```diff
@@ -1860,6 +1860,24 @@ repositories:
       url: https://github.com/ros/common_tutorials.git
       version: indigo-devel
     status: maintained
+  compass:
```
clalancette, line 1863: This name seems pretty generic to me, and thus I think it probably violates https://www.ros.org/reps/rep-0144.html#global-guidelines point 1 or 4. It's kind of subjective, as this is a generic package, but it isn't clearly used by large parts of the community. Thus I think this package would be a candidate for prefixing the name of the entity.

clalancette (changes requested):

melodic/distribution.yaml
peci1: What about keeping compass_msgs and renaming compass to magnetometer_compass? 

There's really no need for org prefix as the package just implements a generic algorithm anybody would use when trying to get azimuth readings from a magnetometer.

peci1: @clalancette I guess this has somehow fallen off the table? There doesn't seem to be much time until the sync...

PR #37720, opened 2023-06-19
relative_positional_controller: 1.0.1-1 in 'melodic/distribution.yaml' [bloom]

peci1:
Increasing version of package(s) in repository `relative_positional_controller` to `1.0.1-1`:

- upstream repository: https://github.com/ctu-vras/relative_positional_controller.git
- release repository: https://gitlab.fel.cvut.cz/cras/ros-release/relative_positional_controller.git
- bloom version: `0.11.2`

## relative_positional_controller

```
* Noetic compatibility.
* Initial commit
* Contributors: Martin Pecka
```

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -10457,6 +10457,21 @@ repositories:
       url: https://github.com/ros-controls/realtime_tools.git
       version: melodic-devel
     status: maintained
+  relative_positional_controller:
+    doc:
+      type: git
+      url: https://github.com/ctu-vras/relative_positional_controller.git
+      version: master
+    release:
+      tags:
+        release: release/melodic/{package}/{version}
+      url: https://gitlab.fel.cvut.cz/cras/ros-release/relative_positional_controller.git
+      version: 1.0.1-1
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


melodic/distribution.yaml
```diff
@@ -10457,6 +10457,21 @@ repositories:
       url: https://github.com/ros-controls/realtime_tools.git
       version: melodic-devel
     status: maintained
+  relative_positional_controller:
```
clalancette, line 10460: This name seems pretty generic to me, and thus it might violate https://www.ros.org/reps/rep-0144.html#global-guidelines point 1 or 4.  This one is really subjective, as I could be convinced otherwise.  But given the fact that we are at the end of life for Melodic, I'm of the opinion to not put this in, and instead have the discussion about the name in https://github.com/ros/rosdistro/pull/37721 .


It's kind of subjective, as this is a generic package, but it isn't clearly used by large parts of the community. Thus I think this package would be a candidate for prefixing the name of the entity.

clalancette (changes requested):

melodic/distribution.yaml
peci1: Okay, I'm leaving this be for Melodic...

PR #37721, opened 2023-06-19
relative_positional_controller: 1.0.1-1 in 'noetic/distribution.yaml' [bloom]

peci1:
Increasing version of package(s) in repository `relative_positional_controller` to `1.0.1-1`:

- upstream repository: https://github.com/ctu-vras/relative_positional_controller.git
- release repository: https://gitlab.fel.cvut.cz/cras/ros-release/relative_positional_controller.git
- bloom version: `0.11.2`

## relative_positional_controller

```
* Noetic compatibility.
* Initial commit
* Contributors: Martin Pecka
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -8291,6 +8291,21 @@ repositories:
       url: https://github.com/ros-controls/realtime_tools.git
       version: melodic-devel
     status: maintained
+  relative_positional_controller:
+    doc:
+      type: git
+      url: https://github.com/ctu-vras/relative_positional_controller.git
+      version: master
+    release:
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://gitlab.fel.cvut.cz/cras/ros-release/relative_positional_controller.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/ctu-vras/relative_positional_controller.git
+      version: master
+    status: maintained
   remote_rosbag_record:
     doc:
       type: git


clalancette:
See my comments in https://github.com/ros/rosdistro/pull/37720#pullrequestreview-1488705615 .  I'd like input from the rest of the rosdistro team on whether this name is too generic or not.

peci1:
My input for discussion: You're right this name doesn't follow REP 144 very well. It implicitly assumes non-holonomic 2D motion, while the name seems very generic. On the other hand, the whole nav stack in ROS 1 and 2 is for 2D robots, so the name might not be that bad... And nobody says holonomic motion action cannot be added to this package.

emersonknapp:
I also think the name is a bit too generic. The package is so small, I wonder is there a lot of benefit to releasing it?

sloretz:
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2023-08-03/)

peci1:
Okay. Has anyone had an idea for a better name for this package? I've only got one - relative_positional_controller_2d.

I could also easily add extension for holonomic motion (motion in `y` direction) to get closer to what a user could expect from a package with this name.

Extension for full 6 DOF motion is also possible.

To stress the potential usefulness of the package, it can be very useful for simple robot testing via keyboard, where the relative positional control is much better than the constant velocity interface in some cases. It can also be used with cheap gamepads which do not have proper analog axes but only a set of 4 digital arrows (see Turtlebot 3). Last, we used this controller in DARPA SubT for manual control of distant robots with high latency and jitter, where velocity control would be completely unusable.

If the package were extended with support for 6 DOF motion, it could also be easily used to control UAVs, e.g. telling it to go 1 meter higher (of course, without feedback, it won't be exactly 1 meter, but it might be more precise than manual velocity control from a distant viewpoint).

clalancette:
First, thanks for your patience here.

We discussed this further, and what we came to is that this package name is a bit generic according to REP-144.  What we suggest is that you add in an organization prefix according to [REP-144](https://www.ros.org/reps/rep-0144.html#global-guidelines), which will definitively make this fine to merge.

moriarty:
Following up on the naming discussion:
Adding the postfix _2d helps but it would it still handle all the generic relative positional control cases that ROS users might do? eg would relative_positional_controller_2d does it work with different DoF manipulators, mobile bases with wheels or legs, drones, boats etc

Prefixing with your org avoids this it is clear that this might not be generic for all use cases.

When a generic name is accepted, it is expected to handle a lot of generic cases or have set plan or roadmap to handle all cases for other potential users.

peci1:
Thank you for keeping this afloat and thinking about it. I agree that adding an org prefix is probably the best thing to do now. If we ever expand the package to cover more cases, we'll consider publishing it under the generic name again.

I'll close this PR and send the prefixed package in a new one.

peci1:
New PR: https://github.com/ros/rosdistro/pull/38368 .

PR #37726, opened 2023-06-19
Add helyos_agent_sdk to rosdep database

jkkolb:

## Package name:

- python3-helyos_agent_sdk-pip

## Package Upstream Source:

[helyOS Agent SDK](https://github.com/FraunhoferIVI/helyOS-agent-sdk)

## Purpose of using this:

The helyos-agent-sdk python package encloses methods and data structure definitions that facilitate the connection of autonomous vehicles to [helyOS Framework](https://helyos-manual.readthedocs.io/en/latest/) through rabbitMQ.

Distro packaging links:

The package is OS independent and distributed via [PyPI](https://pypi.org/project/helyos-agent-sdk/)

  - [not available](https://packages.debian.org/search?suite=stable&section=all&arch=any&searchon=names&keywords=helyos_agent_sdk)
  - [not available](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=helyos_agent_sdk&searchon=names)

We are aware of the restriction that packages depending on pip packages cannot be released to the official build farm

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6232,6 +6232,10 @@ python3-h5py:
   nixos: [python3Packages.h5py]
   openembedded: [python3-h5py@meta-python]
   ubuntu: [python3-h5py]
+python3-helyos_agent_sdk-pip:
+  '*':
+    pip:
+      packages: [helyos_agent_sdk]
 python3-httplib2:
   debian: [python3-httplib2]
   fedora: [python3-httplib2]


clalancette (changes requested):
Hm, there seems to be something odd in this PR.  I know it is unrelated, but the `python-pylint3` key here is different from the one that is in the official file: https://github.com/ros/rosdistro/blob/5eebfb7b4f73f4ce2ee7d6f6706e64e215123922/rosdep/python.yaml#L3097-L3103 .  That is what is causing CI to be unhappy.

@jkkolb Can you please rebase this onto the latest and see if that resolves the issue?

jkkolb (commented):
Thanks for pointing that out, I rebased on the latest right now. Hopefully it solves the issue

clalancette (approved):

clalancette (approved):

audrow (approved):

PR #37728, opened 2023-06-19
added onvif_zeep with pip for ubuntu

ARK3r:
a package for [onvif protocol](https://en.wikipedia.org/wiki/ONVIF) commonly used for [PTZ Cameras](https://en.wikipedia.org/wiki/Pan%E2%80%93tilt%E2%80%93zoom_camera).

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -322,6 +322,10 @@ meson:
 nuitka:
   debian: [nuitka]
   ubuntu: [nuitka]
+onvif_zeep:
+  ubuntu:
+    pip:
+      packages: [onvif_zeep]
 opcua-pip:
   ubuntu:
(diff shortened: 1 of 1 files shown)


adityapande-1995 (approved):

clalancette (commented):
I'll point out that `onvif_zeep` has had no releases since [2018](https://pypi.org/project/onvif-zeep/), and no source activity since [2020](https://github.com/quatanium/python-onvif).  That doesn't automatically disqualify it, but it has the look of an abandoned package.  So I'm not sure that this is something we want to add in.  @ARK3r can you explain what you are using this for?

audrow:
@mergifyio rebase

audrow (approved):

PR #36211, opened 2023-02-15
Stop running tests with nose.

nuclearsandwich:
After a week or so and a few explicit reviews, we should drop the nose test execution. As of #36201 tests are primarily run with pytest.

<details><summary>I should have called this branch tissues</summary>Because it stops running nose</details>

PR #37746, opened 2023-06-21
Mark Foxy as end-of-life.

clalancette:


adityapande-1995 (approved):

mjcarroll (approved):

mjcarroll:
Hold for final rosboss (@quarkytale) approval

quarkytale (approved):

nuclearsandwich:
For future EOL PRs I would recommend disabling build farm jobs before merging this change. `ros_buildfarm` uses the rosdistro index to identify active ROS distributions and so foxy devel jobs have been failing to run since this change was merged.

nuclearsandwich:
## Foxy build farm wind-down

- [x] Stop merging all foxy-related ros/rosdistro PRs
- [x] Tag the last commit of ros2/ros_buildfarm_config which includes the foxy config and push the tag: https://github.com/ros2/ros_buildfarm_config/releases/tag/foxy-20230626-config
- [ ] Delete foxy jobs from ros2/ros_buildfarm_config: https://github.com/ros2/ros_buildfarm_config/pull/275
- [ ] Disable jobs using the method described in https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/ongoing_operations.rst (but with batching)
- [x] This is when I'd recommend a PR like this one be merged: https://github.com/ros/rosdistro/pull/36211
- [ ] Update distro in https://github.com/ros-infrastructure/rosindex/blob/ros2/_config.yml#L87-L115
- [ ] After a few weeks, assuming that no one has asked to resurrect foxy for a fixup sync, the jobs can be deleted using a modified copy of the disabling script.

quarkytale:
Thanks for the super helpful checklist!

PR #37756, opened 2023-06-22
Adding catch_ros2 to rolling and iron distros

ngmor:

rolling, iron

# The source is here:

https://github.com/ngmor/catch_ros2

# Checks


mjcarroll:
I think that this should be `catch2_ros2` to keep the naming of the upstream package?

Also does it make sense to keep the license of your package consistent with the upstream?

ngmor:
This is already indexed for humble under this name/license. There is also some included (modified) code from Open Robotics under the Apache License v2, which is why I originally went with this license.

mjcarroll (approved):

PR #37856, opened 2023-07-03
Add ros_kortex src to Noetic

moriarty:
ROS Kortex is the official ROS package to interact with Kortex and its related products. It is built upon the Kortex API, documentation for which can be found in the GitHub Kortex repository. https://github.com/Kinovarobotics/kortex

ros_kortex has been supported by @felixmaisonneuve of Kinova Robotics since Kinetic. However it apparently was not released, only available from source.

packages include:

- kortex_api
- kortex_control
- kortex_description
- kortex_driver
- kortex_examples
- kortex_gazebo
- kortex_move_it_config

Related to this issue:
https://github.com/ros2-gbp/ros2-gbp-github-org/issues/291
& this PR: https://github.com/ros2-gbp/ros2-gbp-github-org/pull/294

There will be a ros2_kortex repository released soon.

The source code to ros_kortex is here:
https://github.com/Kinovarobotics/ros_kortex

The top level license file is here:
https://github.com/Kinovarobotics/ros_kortex/blob/noetic-devel/LICENSE

PR #37857, opened 2023-07-03
add ros2_robotiq_gripper src entries for naming review

moriarty:
Please add this package to be indexed in the rosdistro for Humble, Iron and Rolling.

Repository:
https://github.com/PickNikRobotics/ros2_robotiq_gripper

Packages:

- robotiq_controllers
- robotiq_description
- robotiq_driver

In ROS 1 there was a ros-industrial/robotiq package here: https://github.com/ros-industrial/robotiq
However as of 2021-05-28 it is marked unmaintained. It has not been released since Kinetic.

This is new repository built on top of ROS 2 Control.

https://github.com/PickNikRobotics/ros2_robotiq_gripper/blob/main/LICENSE

Robotiq is currently not supporting this package.
PickNik Robotics will be maintaining it.
Kinova Robotics will depend on this.

This is related to #37856 as we're preparing to release ROS 2 support for Kinova

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5500,6 +5500,12 @@ repositories:
       url: https://github.com/IntelligentRoboticsLabs/ros2_planning_system.git
       version: humble-devel
     status: developed
+  ros2_robotiq_gripper:
+    source:
+      type: git
+      url: https://github.com/PickNikRobotics/ros2_robotiq_gripper.git
+      version: main
+    status: maintained
   ros2_socketcan:
     doc:
       type: git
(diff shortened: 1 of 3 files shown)


clalancette (approved):
The name seems to meet our naming conventions, so looks good to me.

A couple of minor points here (which don't block merging, at least right now):
1.  It is usually best to do a separate PR per ROS distribution.  Having them all in one like this means that if any of the distributions are in a sync hold, we won't be able to merge it.  That's actually not an issue right now as none of them are in a sync hold, but it is a common problem.
2. You may want to consider adding a `doc` section as well, but it is not required.

@audrow can you also take a look and if you approve, go ahead and merge?

clalancette:
> If they're src only entries would they be blocked by a sync hold?

Typically, yes.  It is just easier to say "no merges into this distribution during a sync hold" than to carve out individual exceptions.

audrow (approved):

PR #37858, opened 2023-07-04
Add msgpack11 and sick_scan to documentation index for distro noetic

rostest:

ROSDISTRO noetic

# The source is here:

https://github.com/SICKAG/msgpack11
https://github.com/SICKAG/sick_scan_xd

# Checks

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -10686,7 +10686,7 @@ repositories:
     source:
       type: git
       url: https://github.com/SICKAG/sick_scan_xd.git
-      version: master
+      version: feature/bloom_pretest
     status: developed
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
So this PR is unreleasing sick_scan from noetic, changing the source repository to a different one, changing the branch, and also adding in a new package.  That is way too many things in a single PR, and we'll need to get agreement from the existing maintainers to do some of them (like take over the source repository).  If you'd like to do all of those things, please open up separate PRs for each of them where we can review them separately.  Thanks.

rostest:
@clalancette Thank you for the advice and feedback! We will do it step by step with separate PRs.

PR #37873, opened 2023-07-05
clarify that naming request PR must be made manually

berndpfrommer:
The PR current template is somewhat confusing because it says that bloom will make the PR automatically, which is not correct for the naming review PR.
See [this issue](https://github.com/ros-infrastructure/bloom/issues/706) filed against bloom.


audrow (commented):
I think this is valuable. Maybe it makes sense to add a paragraph summarizing the contents of this comment to make things ever clearer:

https://github.com/ros-infrastructure/bloom/issues/706#issuecomment-1614964703

sloretz (approved):

PR #37903, opened 2023-07-10
add fusion engine driver node in ros2 humble

BlanchoMartin:


humble/distribution.yaml
```diff
@@ -10,6 +10,17 @@ release_platforms:
...
+    release:
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/PointOneNav/ros2-fusion-engine-driver.git
+      version: 1.0.0
+    source:
+      type: git
+      url: https://github.com/PointOneNav/ros2-fusion-engine-driver.git
+      version: ros2
+    status: maintained
```
clalancette, line 23: This is not going to work as-is.  First of all, this needs to be in alphabetical order by key.

Second, the release repository should never be the same repository as the source repository.  It should almost always be a repository in https://github.com/ros2-gbp .  What I'm going to suggest here is that you just add the `source` repository here; that will allow us to verify the name, and also avoid some of the complexities with the `release` section for now.  Once we have this vetted and in a source repository, we can then turn to trying to get the release repository setup following the documentation in http://docs.ros.org/en/humble/How-To-Guides/Releasing/Releasing-a-Package.html

Turning to the source repository, I see a few problems:
* The package description is empty: https://github.com/PointOneNav/ros2-fusion-engine-driver/blob/df89f727c621b2c52ed5368063b8937e53adc135/fusion-engine-driver/package.xml#L6
* The package license is empty: https://github.com/PointOneNav/ros2-fusion-engine-driver/blob/df89f727c621b2c52ed5368063b8937e53adc135/fusion-engine-driver/package.xml#L8
* There is a `requirements.txt` at the top-level that references `catkin`, which shouldn't be needed (this isn't blocking, but I just thought I would point it out).

With the above things fixed, I think we can consider this for inclusion.

clalancette (changes requested):

PR #37915, opened 2023-07-11
Add action to label inactive PRs as stale

audrow:
I think this should work as it is. This PR does the following:

- Runs a bot everyday in the early morning, US time, to label PRs that have been inactive for 14 days as stale
- When a PR is stale the PR gets a "stale" label with a comment
- Doesn't label issues as "stale"
- Doesn't close issues or PRs that have the "stale" label
- Ignores PRs with the label "don't mark as stale"

Feel free to help me workshop the comment that users will see when their PR is stale. 

I also created the "stale" and "don't mark as stale" labels.

### DIFF
--- a/.github/workflows/stale_labeler.yaml
+++ b/.github/workflows/stale_labeler.yaml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/stale@v8
         with:
           stale-pr-label: 'stale'
-          stale-pr-message: "This PR is being marked as stale because there hasn't been activity in 14 days. It may be closed by a rosdistro maintainer. If this label isn't appropriate, you can remove the label and add the 'dont-mark-as-stale' label. You can find our community contributing guidelines here: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md."
+          stale-pr-message: "This PR is being marked as stale because there hasn't been activity in 14 days. It may be closed by a rosdistro maintainer. If this label isn't appropriate, you can remove the label and add the \"don't mark as stale\" label. You can find our community contributing guidelines here: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md."
           days-before-pr-stale: 14
           days-before-pr-close: -1     # dont close PRs automatically
           days-before-issue-stale: -1  # dont mark issues as stale
--- a/.github/workflows/stale_labeler.yaml
+++ b/.github/workflows/stale_labeler.yaml
@@ -11,9 +11,9 @@ jobs:
       - uses: actions/stale@v8  # https://github.com/marketplace/actions/close-stale-issues
         with:
           stale-pr-label: 'stale'
-          stale-pr-message: "This PR is being marked as stale because there hasn't been activity in 14 days. It may be closed by a rosdistro maintainer. If this label isn't appropriate, you can remove the label and add the \"don't mark as stale\" label. You can find our community contributing guidelines here: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md."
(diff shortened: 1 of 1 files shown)


.github/workflows/stale_labeler.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - cron: '0 11 * * *'  # runs at 11am UTC every day => early morning in US
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8  # https://github.com/marketplace/actions/close-stale-issues
+        with:
+          stale-pr-label: 'stale'
+          stale-pr-message: "This PR is being marked as stale because there hasn't been activity in 14 days. It may be closed by a rosdistro maintainer. If this label isn't appropriate, you can remove the label and add the \"don't mark as stale\" label. You can find our community contributing guidelines here: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md."
```
clalancette, line 14: One thing about this message is that users will typically not have the permissions to manipulate labels on rosdistro PRs.  So that wording should be modified to "ask a maintainer to label it".  Also, I'd prefer if the label were a bit shorter, and didn't have a negation in it; I like @tylerjw's suggestion of "persistent", but I'm open to other suggestions as well.

.github/workflows/stale_labeler.yaml
```diff
@@ -0,0 +1,19 @@
...
+          days-before-pr-stale: 14
```
clalancette, line 15: The other concern I have is what was brought up in the meeting; i.e. that this won't add a new comment every two weeks.  Can you verify whether that is the case or not?

audrow: The docs say this, which doesn't fully answer the question:

> If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart

It seems like if the `stale` label exists, it won't make a comment. It also filters out bots commenting on PRs, so it doesn't consider them.

I've been looking at the code and it looks 
https://github.com/actions/stale/blob/465330b7e2524a0d9b3430795c7970658106bf5c/src/classes/issues-processor.ts#L798-L802

I don't think that we'll see this commenting multiple times on a stale issue.

.github/workflows/stale_labeler.yaml
audrow: Done in 0e782a8.

.github/workflows/stale_labeler.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - cron: '0 11 * * *'  # runs at 11am UTC every day => early morning in US
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8  # https://github.com/marketplace/actions/close-stale-issues
+        with:
+          stale-pr-label: 'stale'
+          stale-pr-message: "This PR is being marked as stale because there hasn't been activity in 14 days. It may be closed by a rosdistro maintainer. If this label isn't appropriate, you can ask a maintainer to remove the label and add the 'persistent' label. You can find our community contributing guidelines here: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md."
```
tfoote, line 14:
```suggestion
          stale-pr-message: "This PR hasn't been activity in 14 days. If you are still are interested in getting it merged please provide an update. Otherwise it will likely be closed by a rosdistro maintainer following our [contributing policy](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md). It's been labeled "stale" for visibility to the maintainers. If this label isn't appropriate, you can ask a maintainer to remove the label and add the 'persistent' label.
```

tfoote (changes requested):
I reworded it a bit to be more focused on the user rather than the process.

.github/workflows/stale_labeler.yaml
audrow: That's much better 👍

tfoote (approved):

clalancette (approved):

PR #37951, opened 2023-07-13
gz_ros2_control: 1.1.1-1 in 'rolling/distribution.yaml' [bloom]

ahcorde:
Increasing version of package(s) in repository `gz_ros2_control` to `1.1.1-1`:

- upstream repository: https://github.com/ros-controls/gz_ros2_control
- release repository: https://github.com/ros2-gbp/ign_ros2_control-release.git
- distro file: `rolling/distribution.yaml`
- bloom version: `0.11.2`
- previous version for package: `null`

## gz_ros2_control

```
* Remove plugin export from ROS 1 (#158 <https://github.com/ros-controls/gz_ros2_control//issues/158>)
* Fixed default gazebo version in CMakeLists.txt (#156 <https://github.com/ros-controls/gz_ros2_control//issues/156>)
* Compile master with iron and rolling (#142 <https://github.com/ros-controls/gz_ros2_control//issues/142>)
* Update package.xml (#141 <https://github.com/ros-controls/gz_ros2_control//issues/141>)
* Contributors: Alejandro Hernández Cordero, Bence Magyar, Christoph Fröhlich
```

## gz_ros2_control_demos

```
* typo fix (#143 <https://github.com/ros-controls/gz_ros2_control//issues/143>)
* Contributors: Reza Kermani
```

## gz_ros2_control_tests

```
* Run end to end test in CI (#152 <https://github.com/ros-controls/gz_ros2_control//issues/152>)
* Add test to check position controller (#134 <https://github.com/ros-controls/gz_ros2_control//issues/134>)
* Contributors: Alejandro Hernández Cordero
```

PR #38054, opened 2023-07-24
Removed ros2_gz_control_tests package from gz_ros2_control

ahcorde:
[Failing in CI](https://build.ros2.org/job/Rbin_ujv8_uJv8__gz_ros2_control_tests__ubuntu_jammy_arm64__binary/2/) but we don't need to release this package.


quarkytale (approved):
Removing a package from `gz_ros2_control` first added here https://github.com/ros/rosdistro/pull/37951 and https://github.com/ros/rosdistro/pull/37988

PR #38119, opened 2023-08-02
add python3-sqlmodel-pip to rosdep/python.yaml

SubaruArai:

## Package name:

`python3-sqlmodel-pip`

## Package Upstream Source:

https://github.com/tiangolo/sqlmodel

## Purpose of using this:

This python package provides a easy-to-use, pythonic and strongly-typed interface to various SQL databases.

SQL databases are used widely even today, and I found them to be very useful to store runtime-configurable data across reboots.
This package should facilitate storing those data.

  - not available
  - not available

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8580,10 +8580,11 @@ python3-sqlalchemy-utils:
   nixos: [python3Packages.sqlalchemy-utils]
   opensuse: [python3-SQLAlchemy-Utils]
   ubuntu: [python3-sqlalchemy-utils]
-python3-sqlmodel-pip:
+python3-sqlmodel:
   debian:
-    pip:
-      packages: [sqlmodel]
+    '*':
+      packages: [python3-sqlmodel]
+    buster: null
   ubuntu:
     pip:
       packages: [sqlmodel]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8582,15 +8582,18 @@ python3-sqlalchemy-utils:
   ubuntu: [python3-sqlalchemy-utils]
 python3-sqlmodel:
   debian:
-    '*':
-      packages: [python3-sqlmodel]
+    '*': [python3-sqlmodel]
     bullseye:
       pip:
         packages: [sqlmodel]
-    buster: null
   ubuntu:
-    pip:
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):

SubaruArai:
@clalancette 
My bad, didn't notice. I've updated, please check.

rosdep/python.yaml
```diff
@@ -8580,6 +8580,17 @@ python3-sqlalchemy-utils:
   nixos: [python3Packages.sqlalchemy-utils]
   opensuse: [python3-SQLAlchemy-Utils]
   ubuntu: [python3-sqlalchemy-utils]
+python3-sqlmodel:
+  debian:
+    '*':
+      packages: [python3-sqlmodel]
```
clalancette, line 8586: It's a little confusing, but the syntax here is slightly different:

```suggestion
    '*': [python3-sqlmodel]
```

rosdep/python.yaml
```diff
@@ -8580,6 +8580,17 @@ python3-sqlalchemy-utils:
...
+    bullseye:
+      pip:
+        packages: [sqlmodel]
+    buster: null
```
clalancette, line 8590: I think we can just drop `buster` for now, it is only on LTS support in Debian anyway.

rosdep/python.yaml
```diff
@@ -8580,6 +8580,17 @@ python3-sqlalchemy-utils:
...
+  ubuntu:
+    pip:
+      packages: [sqlmodel]
```
clalancette, line 8593: It also looks like this is available in the latest Ubuntu (https://packages.ubuntu.com/search?keywords=python3-sqlmodel&searchon=names&suite=all&section=all), so this can be:

```suggestion
  ubuntu:
    '*': [python3-sqlmodel]
    focal: null
    jammy: null
```

clalancette (changes requested):

SubaruArai:
@clalancette Updated, please check!

clalancette (approved):
This looks good to me, thanks for iterating!

audrow (approved):

SubaruArai:
@clalancette No, thank you for your patience!

PR #38123, opened 2023-08-02
Add hebi_cpp_api to humble.yaml

cwbollinger:
Add hebi_cpp_api package to humble

## Package name:

hebi_cpp_api

## Package Upstream Source:

https://github.com/HebiRobotics/hebi_cpp_api_ros.git

Previously released on Noetic

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2107,6 +2107,15 @@ repositories:
       url: https://github.com/tier4/heaphook.git
       version: main
     status: maintained
+  hebi_cpp_api:
+    doc:
+      type: git
+      url: https://github.com/HebiRobotics/hebi_cpp_api_ros/
+      version: ros2
+    source:
+      type: git
+      url: https://github.com/HebiRobotics/hebi_cpp_api_ros/
+      version: ros2
   hey5_description:
     doc:
       type: git
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2110,11 +2110,11 @@ repositories:
   hebi_cpp_api:
     doc:
       type: git
-      url: https://github.com/HebiRobotics/hebi_cpp_api_ros/
+      url: https://github.com/HebiRobotics/hebi_cpp_api_ros.git
       version: ros2
     source:
       type: git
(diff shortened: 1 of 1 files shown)


humble/distribution.yaml
```diff
@@ -2107,6 +2107,15 @@ repositories:
       url: https://github.com/tier4/heaphook.git
       version: main
     status: maintained
+  hebi_cpp_api:
+    doc:
+      type: git
+      url: https://github.com/HebiRobotics/hebi_cpp_api_ros/
```
clalancette, line 2113: This is failing CI because repositories need to end in `.git`.

Besides that, this repository bundles a bunch of `.so` files which don't seem to be under the Apache 2.0 license (if I had to guess, I'd guess they were under the "HEBI C++ Software License").  Because of that, the licensing of this is unclear to me.  Can you clarify what the license of those `.so` files is?

clalancette (changes requested):

humble/distribution.yaml
cwbollinger: Apologies for the delay getting back to this, we've been talking internally about the licensing question. Currently the relevant license can be found at https://www.hebirobotics.com/software-license under "C API terms and conditions". We have not released the source for these libraries with previous releases, and weren't planning on doing that now, but we're not trying to place restrictions on the use of the shared object files.

mjcarroll (approved):

humble/distribution.yaml
clalancette: I'm really sorry about the delay here; I typed up a response, and I thought I clicked "Comment" on it, but it hasn't shown up here.

We wanted one more piece of information for clarification.  In particular, are the binary `.so` files here explicitly licensed under Apache-2.0?  Assuming that is the case, then we can go ahead and merge this in.  Thanks for your patience.

cwbollinger: No worries! Yes, all the files in the package, including the shared object files, are intended to be licensed under Apache 2.0. If there's a standard way to make that more explicitly clear I can add that as well.

clalancette (approved):

PR #38126, opened 2023-08-03
fix: non-raw string for regex in script causing `DeprecationWarning`

SubaruArai:
in scripts/check_rosdep.py, there was a normal string used with `'\s'` escape, which caused `DeprecationWarning`.
This commit fixes it by adding `r` and specifying as a raw string.


SubaruArai:
for more info, see: https://docs.python.org/3/library/re.html

clalancette (approved):

PR #38136, opened 2023-08-03
rosdep: nixos: Add qt5-svg library package

hacker1024:
This adds the NixOS package for `libqt5-svg`, ~~and changes `libqt5-svg-dev` to use the correct package~~.

cc: @lopsided98


lopsided98:
There is usually no need to specifically explicitly specify the `dev` output of a Nix package in rosdep. Did you run into something that requires this?

hacker1024:
> There is usually no need to specifically explicitly specify the `dev` output of a Nix package in rosdep. Did you run into something that requires this?

No, I just searched for other instances of `.dev` to see if it was used and found one.

There's not much of a difference though, right? When a derivation is added to `buildInputs` its `.dev` output is added along with `lib` and `out` if they exist.

Isn't `.dev` slightly more accurate, in case a tool other than Super Flore does need only headers for some reason?

lopsided98:
My reasoning is that you wouldn't explicitly add the `dev` output to `buildInputs` if writing the package by hand, so the auto-generated packages shouldn't either. Additionally, using the `dev` output will cause things to break if split outputs are ever disabled for that package in nixpkgs (this sometimes happens because subtle breakage is discovered due to improperly constructed paths in pkg-config or CMake files). When you truly need to get only the `dev` output, you would normally use the `lib.getDev` function to handle this case without breaking.

I was not aware of the other instance of `.dev` and I would have recommended against it if I had reviewed that PR.

hacker1024:
Gotcha. I reverted the `.dev` addition.

lopsided98 (approved):

clalancette (approved):

audrow (approved):

PR #38125, opened 2023-08-02
Remove the remaining remnants of nose from rosdistro.

clalancette:
This includes the dependency in the requirements.txt, as well as the documentation for it in CONTRIBUTING.md

PR #38143, opened 2023-08-04
Update outdated testing documentation (use pytest)

SubaruArai:
The CONTRIBUTING.md's "How to submit pull requests" specified nosetests, which is no longer used. (see: commit id f27dfbf, github pr #36201)
This commit updates the documentation to use pytest instead, and also encourages the use of a virtual environment.

Config files are also updated to ignore the virtual environment.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -300,7 +300,7 @@ The unit tests enforce alphabetization of elements and a consistent formatting t
 ### Unit Testing
 
 It is recommended to use a virtual environment and pip to install the unit test dependencies.
-They are listed in tests/requirements.txt.
+The test dependencies are listed in tests/requirements.txt.
 
 ```bash
 # create the virtual environment
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -305,6 +305,20 @@ To run the tests, run `pytest` in the root of the repository.
 
 #### Test Dependencies
 
+Using a virtual environment and `pip` is recommended to install the dependencies.
+
+```bash
+# create the virtual environment
+python3 -m venv .venv
+# "activate" the virtual environment
+# this will let pip install dependencies into the virtual environment
+# use activate.zsh if you use zsh, activate.fish if you use fish, etc.
+source .venv/bin/activate
+
+# install the dependencies
+pip3 install -r test/requirements.txt
+```
+
 Tests are written using [pytest](https://docs.pytest.org/).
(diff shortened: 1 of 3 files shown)


clalancette (changes requested):
I'll note that this will conflict with https://github.com/ros/rosdistro/pull/38125 .  There is more information in here than in that PR, but I think we should probably get that one in first and then rebase this on top.

SubaruArai:
@clalancette merged, please check!

SubaruArai:
@clalancette ping

CONTRIBUTING.md
```diff
@@ -318,10 +337,17 @@ These tests require several dependencies that can be installed either from the R
 | unidiff        | python3-unidiff (Zesty and higher) | unidiff        |
 | yamllint       | yamllint                          | yamllint       |
```
clalancette, line 338: This whole thing seems a bit redundant.  We tell the user 3 times to run `pytest` locally.

What I'm going to suggest instead is that we rewrite this to be simpler.  In particular, let's just add in one section called "Unit testing" or something like that.  In there, we say something like:

```
All of the dependencies for testing are listed in tests/requirements.txt.  It is recommended to use a virtual environment and pip to install the dependencies.
```

And then go on to give those instructions on how to install and run.  We can then remove the rest of the duplicated information here, so this is a lot shorter and more to the point.

CONTRIBUTING.md
```diff
@@ -318,10 +337,17 @@ These tests require several dependencies that can be installed either from the R
...
+```bash
+for f in ./rosdep/*.yaml; do echo "$f"; ./scripts/check_rosdep.py "$f"; done
+```
 
-For example, to check a change to `rosdep/base.yaml`:
+Or simply use `pytest` (although the output is quite verbose):
 ```bash
-python3 scripts/check_rosdep.py rosdep/base.yaml
+pytest test/rosdep_formatting_test.py
 ```
```
clalancette, line 353: Honestly, I'm not sure we need this bit at all.  This is already covered by the general testing above, so maybe we can just drop this stanza completely?

clalancette (changes requested):

CONTRIBUTING.md
SubaruArai: lgtm

CONTRIBUTING.md
SubaruArai: Yeah, this was basically a leftover that I didn't know if I should've deleted.
I'm fine for dropping.

SubaruArai:
@clalancette updated, please check!

SubaruArai:
@clalancette ping

SubaruArai:
@clalancette ping

clalancette (approved):
Sorry for the delay here.  This looks good to me now.

PR #38144, opened 2023-08-04
linting: modify yaml files to satisfy yamllint

SubaruArai:
The files modified are:
- `.yamllint` (document-start, truthy)
- `.github/**/*.y[a]ml` (document-start, indentation, line-length, truthy)

Once this PR is merged, we can fix yamllint in github actions (#38145). (currently, it ignores hidden directories)

### DIFF
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -1,6 +1,7 @@
 ---
 name: Validate rosdistro
-"on":
+# this is fine since gha runs with yaml 1.2
+on:  # yamllint disable-line rule:truthy
   push:
     branches: ['master']
   pull_request:
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -1,5 +1,6 @@
+---
 name: Validate rosdistro
-on:
+"on":
   push:
     branches: ['master']
   pull_request:
@@ -22,7 +23,8 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Fetch upstream (to enable diff)
       run: |
-        git remote add unittest_upstream_comparison https://github.com/ros/rosdistro.git || git remote set-url unittest_upstream_comparison https://github.com/ros/rosdistro.git
+        git remote add unittest_upstream_comparison https://github.com/ros/rosdistro.git \
+        || git remote set-url unittest_upstream_comparison https://github.com/ros/rosdistro.git
         git fetch --no-tags --depth=1 unittest_upstream_comparison master
     - name: Install Dependencies
       run: |
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,9 @@
+---
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    # only notify on security updates and not version updates.
-    open-pull-requests-limit: 0
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  # only notify on security updates and not version updates.
+  open-pull-requests-limit: 0
--- a/.github/labeler-config.yaml
+++ b/.github/labeler-config.yaml
@@ -1,38 +1,39 @@
+---
 # Add 'noetic' label if any files in noetic/ folder are changed by a PR
 noetic:
-  - noetic/*
+- noetic/*
 
 # Add 'humble' label if any files in humble/ folder are changed by a PR
 humble:
-  - humble/*
+- humble/*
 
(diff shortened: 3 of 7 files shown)


audrow (approved):
This looks good to me. Thanks for the catch!

@clalancette, any thoughts on this?

.github/workflows/build_test.yaml
```diff
@@ -1,5 +1,6 @@
+---
 name: Validate rosdistro
-on:
+"on":
```
clalancette, line 3: I have to say that this is weird.  The official GitHub docs don't put quotes around this: https://docs.github.com/en/actions/examples/using-scripts-to-test-your-code-on-a-runner .

While I understand we may want the linter to be happy, I also think that we should keep these workflows looking like the official examples.  So maybe we should just ignore these files when using the yaml linter, or change the configuration so that this isn't required?

(same goes for the rest of the workflows below)

clalancette (changes requested):

.github/workflows/build_test.yaml
sloretz: Prior to yaml 1.2, `on` unquoted was interpreted as a boolean: https://yaml.org/spec/1.2.2/ext/changes/#:~:text=Only%20true%20and%20false%20strings%20are%20parsed%20as%20booleans%20(including%20True%20and%20TRUE)%3B%20y%2C%20yes%2C%20on%2C%20and%20their%20negative%20counterparts%20are%20parsed%20as%20strings.

yamllint has an option to ignore dictionary keys, so if you don't want quoted "on" here I think this in `.yamllint` would prevent that check: https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy

```yaml
rules:
  truthy:
    check-keys: false
```

SubaruArai: Looks like there's even an issue (https://github.com/adrienverge/yamllint/issues/559) stating that `check-keys: false` should be `false` as a default.

Also - if we start talking about the official github docs, the indentation could be an issue too.

```yaml
github:
  - a
  - b
rosdistro:
- a
- b
```

Ultimately, I think we have to choose between the 2:
1. keep consistent formatting across all yaml files in the project
2. have 2 types of formatting rules, each for rosdistro and gha

While I personally prefer 1 over 2, it ultimately boils down to personal preference.
I'll let the maintainers decide.

<details><summary>todo list if we choose 2</summary><div>

1. exclude `.github` from `prj_root/.yamllint`
2. make a `.yamllint` file inside `prj_root/.github`
3. make a test script to glob `**/.yamllint` files and execute `yamllint` in every one of those directories
4. update the CI to use 3
</div></details>

<details><summary>sidenote</summary><div>
And just as a sidenote (not saying that we should) - we could go nuclear:

1. make a dockerfile, and update the docs to test using docker-compose
2. switch to prettier or another formatter, not linter
3. since we're already doing all those, why not let prettier force format on pr?
</div></details>

clalancette: > 2\. have 2 types of formatting rules, each for rosdistro and gha

My personal feeling on linters is that they help guide development so that things look consistent.  In this case, the linters are actually guiding us *away* from consistency with the rest of the GitHub workflow-using world.  So I'd prefer to have different formatting rules for the GitHub workflows.  But I'd like to hear from @sloretz and @tfoote as well.

SubaruArai: Just FYI (and anyone reading this the first time), Currently the formatting in `.github` is mixed. Some obey github style and some obey rosdistro style.

sloretz: Subjectively, I don't see the benefit of maintaining style with the rest of the github workflow world, especially if the cost is maintaining two different yaml linting styles in the same repo. If I had a vote I'd vote for consistency.

clalancette: We discussed this in the weekly meeting, and the consensus was that we should let yamllint run over the `.github` directory, but the quoted "on" thing is weird.  It was pointed out that yamllint quotes "on" to distinguish the string "on" with the old-style YAML coercion of on/off to a boolean.  That said, that shouldn't be a problem with GitHub actions, so we think we should *not* have this quoted.

So I think the way to move forward here is to remove the "on" quoting, and add a `noqa` comment to these specific lines.  @SubaruArai if you go ahead and do that, I think we can get this PR in.  Let me know if this makes sense to you.

SubaruArai: @clalancette I'm fine with this decision. Maybe I'll add an explanation to why using `noqa` instead of `check-keys: false`.
I guess it's related to pyyaml still not supporting yaml1.2, and rosdep using pyyaml right?

SubaruArai:
@clalancette ping

clalancette (approved):

Issue #25334, opened 2020-06-04
restore folding logic from Travis or update error messages


dirk-thomas:
See e.g. https://github.com/ros/rosdistro/blob/a8d19c8bd7db8cf720da53882959ac4a5bb53482/test/rosdistro_check_urls_test.py#L22-L29

Regression from #24333.

dirk-thomas:
@tfoote Friendly ping.

tfoote:
This was discussed in the review on that as an acceptable regression with the improved performance. I'm not going to have time to look at it right now. 

If someone has time to look at it . There looks like there's a group and ungroup is an alternative to folding:
- https://github.community/t/has-github-action-somthing-like-travis-fold/16841
- https://github.community/t/travis-fold-equivalent/18256
- https://github.com/actions/toolkit/blob/df7e2c13c89ea93069cad62f1adfe78a87111cc0/docs/commands.md#group-and-ungroup-log-lines

PR #38167, opened 2023-08-07
restore CI output group folding in github actions

SubaruArai:
Related: #25334

See official docs for details:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

### DIFF
--- a/test/fold_block.py
+++ b/test/fold_block.py
@@ -31,22 +31,6 @@ class BaseCiCfg:
         pass
 
 
-class TravisCiCfg(BaseCiCfg):
-    @classmethod
-    def is_ci(cls) -> bool:
-        return os.environ.get("TRAVIS") == "true"
-
-    def get_message(self, msg=""):
-        return self._get_message_folded(msg)
-
-    def __enter__(self):
-        print("\ntravis_fold:start:%s" % self.get_block_name())
-        return self
-
-    def __exit__(self, type, value, traceback):
-        print("\ntravis_fold:end:%s" % self.get_block_name())
-
-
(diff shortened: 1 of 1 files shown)


audrow (approved):
This looks like a good change to me and makes the CI easier to go through. @clalancette, what do you think?

test/fold_block.py
```diff
@@ -1,30 +1,72 @@
...
+        return self._get_message_folded(msg)
+
+    def __enter__(self):
+        print("\ntravis_fold:start:%s" % self.get_block_name())
+        return self
+
+    def __exit__(self, type, value, traceback):
+        print("\ntravis_fold:end:%s" % self.get_block_name())
+
+
```
clalancette, line 49: This bit can be removed; we are not going to go back to `travis`.

clalancette (changes requested):
I've left one thing to change in line, but this looks great.

test/fold_block.py
SubaruArai: I first thought to delete it, but it doesn't hurt being there, so...
@audrow what do you think?

audrow: If we're not using it, I'd favor removing it.

The reason is that I'd rather not have unused code hanging around.

SubaruArai:
Sorry, I just added an update.
I didn't like that `cls_` remained in the module.

clalancette (approved):
This seems good to me, but I'll wait for another review from @audrow before merging in.

audrow (approved):
Looks good to me!

PR #38180, opened 2023-08-08
adding python3-nmcli-pip to python

ARK3r:
Allows network control using python similar to the bash nmcli package.

Two points however:
1. for the nmcli pip package to work, apt needs to install network-manager
2. The package only worked for ubuntu 20.04 and 22.04 as I tested in docker.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2485,7 +2485,7 @@ python-networkx:
   gentoo: [dev-python/networkx]
   ubuntu:
     '*': [python-networkx]
-python3-nmcli:
+python3-nmcli-pip:
   ubuntu:
     pip:
       packages: [nmcli]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2485,11 +2485,6 @@ python-networkx:
   gentoo: [dev-python/networkx]
   ubuntu:
     '*': [python-networkx]
-python3-nmcli-pip:
-  ubuntu:
-    pip:
-      depends: [network-manager]
-      packages: [nmcli]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2485,6 +2485,10 @@ python-networkx:
   gentoo: [dev-python/networkx]
   ubuntu:
     '*': [python-networkx]
+python3-nmcli:
```
sloretz, line 2488: Please name this `python3-nmcli-pip`: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1

sloretz (changes requested):

ARK3r:
Is there any way to enforce the network-manager dependency while installing this package?

sloretz (approved):

tfoote:
Yes you can add a depends line for pip rules

Spec: https://www.ros.org/reps/rep-0111.html#dependencies

Enabled: https://github.com/ros-infrastructure/rosdep/blob/05b49e4d588a4f36f222a02c7e492b46002dca86/src/rosdep2/platforms/pip.py#L130C14-L130C14

Implementation: https://github.com/ros-infrastructure/rosdep/blob/05b49e4d588a4f36f222a02c7e492b46002dca86/src/rosdep2/installers.py#L403C38-L403C38

Example: 

https://github.com/ros/rosdistro/blob/ee18ca7343a7da32b9049ce3171c117166141fe3/rosdep/python.yaml#L963-L966

methylDragon (approved):

PR #38211, opened 2023-08-11
Add libavfilter-dev to ffmpeg package on ubuntu/debian

oysstu:
Add libavfilter-dev to the ffmpeg system dependency. This has for example already been done for opensuse.

## Package name:

ffmpeg

## Package Upstream Source:

[link to source repository](https://git.ffmpeg.org/ffmpeg.git)

## Purpose of using this:

It's a core component of ffmpeg/libav

  - https://packages.debian.org/bookworm/libavfilter-dev
  - https://packages.ubuntu.com/jammy/libavfilter-dev

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -4958,7 +4958,7 @@ repositories:
       tags:
         release: release/iron/{package}/{version}
       url: https://github.com/ros2-gbp/ros2_kortex-release.git
-      version: 0.2.1-1
+      version: 0.2.2-1
     source:
       type: git
       url: https://github.com/Kinovarobotics/ros2_kortex.git


tfoote (changes requested):
Why should this be bundled into the ffmpeg key versus being called out separately?

We usually use the ubuntu/debian packages as canonical and others fill in with multiple. This is likely a legacy package w/o the dev addendum. 

But to that end I'd suggest that we leave this package as is and create a new `ffmpeg-dev` rule which is where all the dev packages from ffmpeg's source package are then collected (https://packages.ubuntu.com/source/lunar/ffmpeg search "-dev") . Rather than continuing to expand the scope of this legacy package which uses the runtime name.

oysstu:
> Why should this be bundled into the ffmpeg key versus being called out separately?
> 
> We usually use the ubuntu/debian packages as canonical and others fill in with multiple. This is likely a legacy package w/o the dev addendum.
> 
> But to that end I'd suggest that we leave this package as is and create a new `ffmpeg-dev` rule which is where all the dev packages from ffmpeg's source package are then collected (https://packages.ubuntu.com/source/lunar/ffmpeg search "-dev") . Rather than continuing to expand the scope of this legacy package which uses the runtime name.

Created a new PR in  #38239 that adds an ffmpeg-dev package.

PR #38228, opened 2023-08-13
feat: add gnu octave to rosdep

Podesta:
Please add the following dependency to the rosdep database.

## Package name:

octave

## Package Upstream Source:

[GNU Mercurial](https://hg.savannah.gnu.org/hgweb/octave)
[Github Mirror](https://github.com/gnu-octave/octave)

## Purpose of using this:

Octave is a (mostly MATLAB® compatible) high-level language, primarily  intended for numerical computations. It provides a convenient command-line interface for solving linear and nonlinear problems numerically.

It is an open source alternative to run code that was originally written for matlab, in and open source framework. We are using it for signal processing.

## Links to Distribution Packages

- Octave Wiki: https://wiki.octave.org/Octave_for_GNU/Linux
- Debian: https://packages.debian.org/bookworm/octave
- Ubuntu: https://packages.ubuntu.com/jammy/octave
- Fedora: https://packages.fedoraproject.org/pkgs/octave/octave/
- Arch: https://archlinux.org/packages/extra/x86_64/octave/
- Gentoo: https://packages.gentoo.org/packages/sci-mathematics/octave
- Alpine: https://pkgs.alpinelinux.org/package/edge/community/x86_64/octave
- openSUSE: https://software.opensuse.org/package/octave

PR #38229, opened 2023-08-13
feat: add pip oct2py to rosdep

Podesta:

## Package name:

oct2py

## Package Upstream Source:

https://github.com/blink1073/oct2py

## Purpose of using this:

Oct2Py allows you to seamlessly call M-files and Octave functions from Python. It manages the Octave session for you, sharing data behind the scenes using MAT files. With this we can call matlab/octave functions from our python ros node. Related to #38228 

Distro packaging links:

- pypi: https://pypi.org/project/oct2py/
- Debian: not available
- Ubuntu: not available
- Fedora: not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6969,6 +6969,16 @@ python3-oauth2client:
     '*': [python3-oauth2client]
     '7': null
   ubuntu: [python3-oauth2client]
+python3-oct2py-pip:
+  debian:
+    pip:
+      packages: [oct2py]
+  fedor:
+    pip:
+      packages: [oct2py]
+  ubuntu:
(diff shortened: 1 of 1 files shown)


tfoote (commented):
Can you confirm that you're not planning to use this in a package that you might release and only ever plan to build from source? pip rules cannot be used in released packages.

Podesta:
Thanks for the warning! Am I correct in assuming that this restriction only applies to distro based packages? Meaning that if we decide to package it as a snap, for example, it should be ok?

But yes, our plans for now is only to build from source.

methylDragon (approved):

mjcarroll (approved):

PR #38244, opened 2023-08-14
feat: add 'gymnasium-robotics' pip package

rickstaa:

## Package name:

https://pypi.org/project/gymnasium-robotics/

## Package Upstream Source:

https://github.com/Farama-Foundation/Gymnasium-Robotics

## Purpose of using this:

This package can be used to train RL on robotics systems.

Distro packaging links: N/A

## Additional comments

I omitted a [python/python3](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1) prefix since this package is unavailable as a python2 package.

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6275,6 +6275,21 @@ python3-gymnasium-pip:
   ubuntu:
     '*':
       pip:
+        packages: [gymnasium]
+    focal: null
+python3-gymnasium-robotics-pip:
+  debian:
+    pip:
+      packages: [gymnasium-robotics]
+  fedora:
+    pip:
+      packages: [gymnasium-robotics]
+  osx:
+    pip:
+      packages: [gymnasium-robotics]
+  ubuntu:
+    '*':
+      pip:
         packages: [gymnasium-robotics]
     focal: null
 python3-gz-math6:


rickstaa:
I made this PR a draft since this should not be merged until a decision is taken on https://github.com/ros/rosdistro/pull/38243#issuecomment-1683584262.

rickstaa:
I think we can safely merge this since system dependency conflicts will only exist on Ubuntu 20.04 (see https://github.com/ros/rosdistro/issues/38332 and https://github.com/ros/rosdistro/pull/38243#issuecomment-1690254962).

tfoote:
Sorry, can you clarify how this is only a python2 package?

Almost every system that these rules will match are going to be python 3. 

And we follow the debian naming convention even if it's not in debian for consistency for readers. 

And please clarify your statement that this will only conflict on 20.04. On that system it will install a newer version of numpy implicitly? But not on the other systems. A summary of the different versions here with links to references would be helpful to get us all on the same page.

rickstaa:
> Sorry, can you clarify how this is only a python2 package?
> 
> Almost every system that these rules will match are going to be python 3.
> 
> And we follow the debian naming convention even if it's not in debian for consistency for readers.

@tfoote Of course. Sorry for being unclear. The [gymnasium-robotics](https://pypi.org/project/gymnasium-robotics/) package is relatively new and therefore requires Python >=3.7. It was never released; therefore, it is a Python3-only package. Judging from the [CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md). I, therefore, think the key would need to be `python3-gymnasium-robotics-pip`. I, however, noticed that a lot of rules don't have the python3:

https://github.com/ros/rosdistro/blob/9c23c70f8a9dc8afe5a6b14f73708fe258300097/rosdep/python.yaml#L1-L21

To name a view. I, therefore, needed clarification about whether the python3 prefix was required. If the python3 prefix is necessary, I think we should also update the [gymnasium](https://gymnasium.farama.org/) key added in https://github.com/ros/rosdistro/pull/37334 not that it hasn't been in the repository for too long.

> And please clarify your statement that this will only conflict on 20.04. On that system, does it will install a newer version of numpy implicitly? But not on the other systems. A summary of the different versions here with links to references would be helpful to get us all on the same page.

The issue is the other way around. If rosdep installs Numpy, then only Ubuntu 20.04 version 1.17.4 (see https://pkgs.org/search/?q=python3-numpy) will be installed, while both the gymnasium and gymnasium-robotics packages require numpy=>1.20.1. Users will only encounter this issue when they work on Ubuntu 20.04. You can checkout https://github.com/ros/rosdistro/issues/38332, which I created for Gymnasium but also applies to gymnasium-robotics.

rickstaa:
The reason why I haven't yet added the Python3 prefix is because the gym and gymnasium now don't have the Python3 prefix.

https://github.com/ros/rosdistro/blob/8637df61552d43091c5b5b49e9ba36a6e5d7ff4a/rosdep/python.yaml#L208-L233

tfoote:
Thanks for putting that all together. You're right that those other rules are not following our naming conventions and as they are just landed we should probably update them now instead of maintaining them as is. And reach out to anyone observed using the key publicly to make sure it's not a major problem. 

@cottsay @nuclearsandwich 

For the incompatibility on 20.04. The best thing to do on that is to simply declare a version specific 'null' entry for that version of Ubuntu and use the wildcard syntax to cover the other versions. Then rosdep will tell you that it can't be installed successfully instead of trying and getting runtime issues later. That could also be propagated.

rickstaa:
> Thanks for putting that all together. You're right that those other rules are not following our naming conventions and as they are just landed we should probably update them now instead of maintaining them as is. And reach out to anyone observed using the key publicly to make sure it's not a major problem.
> 
> @cottsay @nuclearsandwich
> 
> For the incompatibility on 20.04. The best thing to do on that is to simply declare a version specific 'null' entry for that version of Ubuntu and use the wildcard syntax to cover the other versions. Then rosdep will tell you that it can't be installed successfully instead of trying and getting runtime issues later. That could also be propagated.

@tfoote, I updated this pull request to reflect the changes you requested. I also created a pull request to update the `gymnasium-pip` key to `python3-gymnasium-pip` (see https://github.com/ros/rosdistro/pull/38341/files). We can safely merge this pull request since it was introduced relatively recently. 

For the `gym-pip` rule, I don't think it is wise to update this rule since it was introduced four years ago. A gymnasium has replaced further gym, so it will be appreciated in the future (see https://github.com/openai/gym?tab=readme-ov-file#the-team-that-has-been-maintaining-gym-since-2021-has-moved-all-future-development-to-gymnasium-a-drop-in-replacement-for-gym-import-gymnasium-as-gym-and-gym-will-not-be-receiving-any-future-updates-please-switch-over-to-gymnasium-as-soon-as-youre-able-to-do-so-if-youd-like-to-read-more-about-the-story-behind-this-switch-please-check-out-this-blog-post).

tfoote (approved):
Thanks for the iterations, lgtm

mjcarroll (approved):

mjcarroll:
@rickstaa I just merged the other PR and resolved conflicts here.  Can you double check my resolution?

One thing I noticed when resolving conflicts is that the `python3-gymnasium-pip` key resolves to `gymnasium-robotics` based on (https://github.com/ros/rosdistro/pull/38341/files).  Is that desired or was it a typo?

rickstaa:
O wait, it looks like other things went wrong.

rickstaa:
Ah looks like I got confused because of how Vscode displayed the git diff. I checked it here and I think the PR is good to be merged 👍🏻.

mjcarroll:
Thanks for iterating!

PR #38406, opened 2023-09-03
rosdep: Add libfyaml-dev

wentasah:

## Package name:

libfyaml-dev

## Package Upstream Source:

https://github.com/pantoniou/libfyaml

## Purpose of using this:

This dependency is used by [mrpt2](https://index.ros.org/search/?term=mrpt2) package. Currently, it fetches the copy of the library source code by cmake. However, this fails on Nix(OS), because packages are built in a sandbox without network access. Having it in rosdep allows building the package without problems.

Distro packaging links:

- Debian: https://packages.debian.org/search?keywords=libfyaml-dev
  - REQUIRED
- Ubuntu: https://packages.ubuntu.com/search?keywords=libfyaml-dev
  - REQUIRED
- Fedora: https://packages.fedoraproject.org/ (NOT AVAILABLE)
  - IF AVAILABLE
- Arch: https://www.archlinux.org/packages/ (NOT AVAILABLE)
  - IF AVAILABLE
- Gentoo: https://packages.gentoo.org/ (NOT AVAILABLE)
  - IF AVAILABLE
- macOS: https://formulae.brew.sh/ (NOT AVAILABLE)
  - IF AVAILABLE
- Alpine: https://pkgs.alpinelinux.org/package/edge/testing/x86_64/libfyaml-dev
  - IF AVAILABLE
- NixOS/nixpkgs: https://search.nixos.org/packages?channel=unstable&show=libfyaml
  - OPTIONAL
- openSUSE: https://software.opensuse.org/package/libfyaml
  - IF AVAILABLE
- rhel: https://rhel.pkgs.org/ (NOT AVAILABLE)
  - IF AVAILABLE

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3402,6 +3402,19 @@ libftgl-dev:
   fedora: [ftgl-devel]
   gentoo: [media-libs/ftgl]
   ubuntu: [libftgl-dev]
+libfyaml-dev:
+  alpine: [libfyaml-dev]
+  debian:
+    '*': [libfyaml-dev]
+    bullseye: null
+  fedora: [libfyaml-devel]
+  nixos: [libfyaml]
+  opensuse:
+    '*': [libfyaml]
+    '15.2': null
+  ubuntu:
+    '*': [libfyaml-dev]
+    focal: null
 libgazebo-dev:
   debian:
     bullseye: [libgazebo-dev]


emersonknapp (approved):

cottsay (dismissed):
An empty list carries different meaning than omitting the rule entirely. An empty list says that the dependency is implicitly met without installing anything, where omitting the rule or explicitly `null` states that the dependency is not available.

I see no reason that we should have empty lists here.

wentasah:
Thanks for explaining the difference. I've removed the `[]` lines.

clalancette (changes requested):
The bot thinks we can satisfy the dependency on `fedora` as well (take a look at the "Files changed" tab).  Once that is fixed this looks good to me.

wentasah:
I've added the fedora entry. Now, there are no hints from the bot.

clalancette (approved):

PR #38413, opened 2023-09-05
fadecandy_ros: 1.0.0-1 in 'humble/distribution.yaml' [bloom]

reinzor:
Increasing version of package(s) in repository `fadecandy_ros` to `1.0.0-1`:

- upstream repository: https://github.com/eurogroep/fadecandy_ros.git
- release repository: https://github.com/eurogroep/fadecandy_ros-release.git
- bloom version: `0.11.2`

## fadecandy_driver

```
* remove time dependency
* refactor to ros2
* Contributors: Tom de Winter
```

## fadecandy_msgs

```
* refactor to ros2
* Contributors: Tom de Winter
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -1598,6 +1598,24 @@ repositories:
       url: https://github.com/ros2/examples.git
       version: humble
     status: maintained
+  fadecandy_ros:
+    doc:
+      type: git
+      url: https://github.com/eurogroep/fadecandy_ros.git
+      version: ros2
+    release:
+      packages:
+      - fadecandy_driver
+      - fadecandy_msgs
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/eurogroep/fadecandy_ros-release.git
+      version: 1.0.0-1
+    source:
+      type: git
+      url: https://github.com/eurogroep/fadecandy_ros.git
+      version: ros2
+    status: maintained
   fastcdr:
     release:
       tags:


emersonknapp (dismissed):
I notice that the URL https://github.com/eurogroep/fadecandy_ros.git is different than that released to past ROS distributions https://github.com/iron-ox/fadecandy_ros.git - is there a strong reason for the migration, is it possible to continue using the existing repo?

Given that, I'm taking a quick fresh look at this:

  - [ ] Top level license file:
- [X] License is [OSI-approved](https://opensource.org/licenses): Apache 2.0
- [X] License correctly listed in package.xmls
- [X] Public source repo: https://github.com/eurogroep/fadecandy_ros.git
- [X] Source repository contains ROS packages
- [X] Each package meets [REP-144](https://www.ros.org/reps/rep-0144.html) naming conventions

And I've noticed there are no specific `LICENSE` files in the source tree. It'd be best if we added those before moving forward with a new rosdistro release

reinzor:
Thanks for the review, just added a `LICENSE` file to the root of the source repository.

> I notice that the URL https://github.com/eurogroep/fadecandy_ros.git is different than that released to past ROS distributions https://github.com/iron-ox/fadecandy_ros.git - is there a strong reason for the migration, is it possible to continue using the existing repo?

See issue https://github.com/iron-ox/fadecandy_ros/issues/25 . Some main contributors are not working at Iron Ox anymore so it would be better to transfer the maintenance so that we can push the `ros2` developments forward.

clalancette (approved):

reinzor:
@clalancette , I see that the build is failing apparently:

https://build.ros2.org/job/Hbin_uJ64__fadecandy_driver__ubuntu_jammy_amd64__binary/5/

but I have no clue what the exact error message is here. Could you point me in the right direction here?

Could it be that we should not put `rosidl_default_generators` here https://github.com/eurogroep/fadecandy_ros/blob/d0aadfa302cef155f99529a479ce554949b32275/fadecandy_msgs/CMakeLists.txt#L30 but only the dependent interfaces?

clalancette:
> Could it be that we should not put `rosidl_default_generators` here https://github.com/eurogroep/fadecandy_ros/blob/d0aadfa302cef155f99529a479ce554949b32275/fadecandy_msgs/CMakeLists.txt#L30 but only the dependent interfaces?

That's certainly one of the problems.  Another is that you should make `rosidl_default_generators` a `buildtool_depend`, rather than just a `build_depend`.  I'd try fixing both of those things, doing a new release, and seeing what happens.

reinzor:
Thanks for the quick reply. Will try that.

reinzor:
https://github.com/ros/rosdistro/pull/38614

reinzor:
@clalancette , updated the `fadecandy_msgs` accordingly and bumped to `1.0.1` but the build of the `fadecandy_driver` still seems to fail:

https://build.ros2.org/job/Hbin_uJ64__fadecandy_driver__ubuntu_jammy_amd64__binary/8/console
```
08:10:32 -- Found fadecandy_msgs: 1.0.1 (/opt/ros/humble/share/fadecandy_msgs/cmake)
08:10:32 CMake Error at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
08:10:32   Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
08:10:32 Call Stack (most recent call first):
08:10:32   /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
08:10:32   /usr/share/cmake-3.22/Modules/FindPkgConfig.cmake:99 (find_package_handle_standard_args)
08:10:32   CMakeLists.txt:28 (find_package)
```

Is there any way to reproduce these builds on my local machine? I am not experiencing any errors when I build the repository on a clean workspace using `colcon build`. Help would be appreciated.

clalancette:
> Is there any way to reproduce these builds on my local machine? I am not experiencing any errors when I build the repository on a clean workspace using `colcon build`. Help would be appreciated.

In short, the buildfarm installs *only* the things you list in your package.xml into a bare-bones container.  So you have to specify everything you depend on in the package.xml; in this case, it looks like it is missing `pkg-config`.

There is some documentation on how to reproduce the build locally; see http://docs.ros.org/en/rolling/How-To-Guides/Building-a-Custom-Debian-Package.html .  But note that you need to do that in a totally "clean" environment, otherwise you'll run into the same issue.  There is also https://discourse.ros.org/t/colcon-in-container-an-extension-to-build-and-test-ros-2-inside-a-container/33274 (though I haven't used it myself).

reinzor:
Thanks; created a PR with the dependency fix https://github.com/ros/rosdistro/pull/38672

PR #38422, opened 2023-09-06
Added source entry for: point_cloud_transport, point_cloud_transport_plugins and point_cloud_transport_tutorial

ahcorde:
Added source entry for naming review: 
 - point_cloud_transport 
 - point_cloud_transport_plugins
 - point_cloud_transport_tutorial

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -3713,6 +3713,24 @@ repositories:
       url: https://gitlab.com/ApexAI/point_cloud_msg_wrapper
       version: rolling
     status: developed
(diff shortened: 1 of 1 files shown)


emersonknapp:
Well, there is ROS 1 precedent (though I guess it is pretty new?) - I see that the repository is moving from https://github.com/ctu-vras/point_cloud_transport, is that right?

ahcorde:
These repositories were transfered to `ros-perception` and only include the ros2 version

cottsay (approved):

PR #38537, opened 2023-09-13
mlx90640_thermal_camera: 1.0.0-1 in 'noetic/distribution.yaml' [bloom]

vakshit:
Increasing version of package(s) in repository `mlx90640_thermal_camera` to `1.0.0-1`:

- upstream repository: https://github.com/vakshit/mlx90640_thermal_camera.git
- release repository: https://github.com/vakshit/mlx90640_thermal_camera-release.git
- bloom version: `0.11.2`

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5599,7 +5599,7 @@ repositories:
       version: master
     release:
       packages:
-      - mlx90640-thermal-camera
+      - mlx90640_thermal_camera
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/vakshit/mlx90640_thermal_camera-release.git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5592,6 +5592,24 @@ repositories:
       type: git
       url: https://github.com/221eROS/mitch_v2_driver.git
       version: main
+  mlx90640_thermal_camera:
+    doc:
+      type: git
+      url: https://github.com/vakshit/mlx90640_thermal_camera.git
+      version: master
+    release:
+      packages:
+      - mlx90640-thermal-camera
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/vakshit/mlx90640_thermal_camera-release.git
+      version: 1.0.0-1
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/vakshit/mlx90640_thermal_camera.git
+      version: master
+    status: maintained
   mobile_robot_simulator:
     doc:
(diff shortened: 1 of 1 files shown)


noetic/distribution.yaml
```diff
@@ -5592,6 +5592,24 @@ repositories:
...
       url: https://github.com/221eROS/mitch_v2_driver.git
       version: main
+  mlx90640_thermal_camera:
+    doc:
+      type: git
+      url: https://github.com/vakshit/mlx90640_thermal_camera.git
+      version: master
+    release:
+      packages:
+      - mlx90640-thermal-camera
```
clalancette, line 5602: This is fairly unusual.  The name of the package as specified in https://github.com/vakshit/mlx90640_thermal_camera/blob/master/package.xml is `mlx90640_thermal_camera`, but here it has the dashes instead of the underscores.  Do you know how that happened?

vakshit: Actually, i followed the documentation on ROS Wiki. The package is meant to be installed as ros-noetic-mlx90640-thermal-camera.

clalancette: > Actually, i followed the documentation on ROS Wiki. The package is meant to be installed as ros-noetic-mlx90640-thermal-camera.

That makes sense, but actually the buildfarm will take care of that bit.  The package name here should match what is in the package.xml (in this case, use underscore instead of dashes).

vakshit: so, a commit changing that name suffices.?

clalancette: > so, a commit changing that name suffices.?

I think that should work.

vakshit:
I have updated the package name. and also merged the master branch of the main repo.

noetic/distribution.yaml
clalancette: OK, it turns out I was wrong.  You'll need to re-bloom in order to make this work, because otherwise the wrong branches and package names are in the release repository.

vakshit:
So, i need to do the bloom-release part again right?
Also, just to keep things consistent, I need to use `_` instead of `-` in the package name and the debian name would be taken care of by the buildfarm.

clalancette:
> So, i need to do the bloom-release part again right? Also, just to keep things consistent, I need to use `_` instead of `-` in the package name and the debian name would be taken care of by the buildfarm.

Yes, exactly right on both counts.

vakshit:
I guess i need to raise a new PR because the name is with `-` in the current branch name.

clalancette:
Closing this in favor of #38558

PR #38538, opened 2023-09-13
Remove hebi_cpp_api_ros from Noetic.

clalancette:
The license on the code is a non-standard license, not approved by the OSI.  Thus we can neither confirm nor deny whether we have the right to redistribute the sources and binaries contained within it.  Until that situation is clarified, we should remove it from the index.

@cwbollinger If we can quickly change the license in https://github.com/HebiRobotics/hebi_cpp_api_ros/blob/master/package.xml , then we don't need to do this revert.  If it is going to take time, however, we'll revert this for now and readd it once the license situation is cleared up.

@gbiggs @nuclearsandwich FYI.

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -3544,23 +3544,6 @@ repositories:
       url: https://github.com/crigroup/handeye.git
       version: master
     status: maintained
-  hebi_cpp_api_ros:
-    doc:
-      type: git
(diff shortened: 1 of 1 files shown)


cwbollinger:
@clalancette I have just updated `package.xml` and added a `LICENSE` file, let me know if anything else is out of order. Thanks!

PR #38539, opened 2023-09-13
nixos: Use generic python3Packages instead of specific python versions

kjeremy:
Correct nixos packages to use the generic Python 3 package set and not tie to a specific minor version.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4947,7 +4947,7 @@ python3-bcrypt:
   debian: [python3-bcrypt]
   fedora: [python3-bcrypt]
   gentoo: [dev-python/bcrypt]
-  nixos: [python39Packages.bcrypt]
+  nixos: [python3Packages.bcrypt]
   opensuse: [python3-bcrypt]
   rhel: ['python%{python3_pkgversion}-bcrypt']
   ubuntu: [python3-bcrypt]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This makes sense to me. 

For mine and others edification, can you link to any docs about how these are differentiated and recommended to be used in nix?

hacker1024:
There's also a near-even split between `pythonPackages` and `python3Packages`. nix-ros-overlay uses the same package set for both, so I'm not sure which is more appropriate. @lopsided98 do you have any thoughts on this?

lopsided98 (approved):
Yes, this is definitely correct.

lopsided98:
Regarding `pythonPackages` vs. `python3Packages`: in nixpkgs `pythonPackages` has always been Python 2 for backwards compatibility reasons, similar to the Debian/Ubuntu package names. I always used `pythonPackages` for Python 2 packages in rosdep, with the added complication that I would sometimes override `pythonPackages` in nix-ros-overlay for a particular distro to point at `python3Packages` to allow me to use Python 3 for distros that were technically supposed to use Python 2. This was particularly important at the tail end of melodic's life because by that point Python 2 was EOL and mostly removed from nixpkgs. Now that melodic is EOL, all the distros use Python 3 and this is mostly irrelevant.

hacker1024 (approved):

clalancette (approved):

PR #38559, opened 2023-09-15
Update techmagic repositories to use ros2-gbp release repo.

nuclearsandwich:
These repositories have been mirrored into the ros2-gbp organization and the ros2-gbp copies should be used for releases going forward.

@Kansea please give a 👍🏼 or Approving review before the ROS team merges this PR.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -18,7 +18,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/TechMagicKK/aandd_ekew_driver_py-release.git
+      url: https://github.com/ros2-gbp/aandd_ekew_driver_py-release.git
       version: 0.0.2-1
     source:
       test_pull_requests: true
@@ -8843,7 +8843,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/TechMagicKK/weight_scale_interfaces-release.git
+      url: https://github.com/ros2-gbp/weight_scale_interfaces-release.git
       version: 0.0.3-1
     source:
       type: git


Kansea (approved):

Kansea:
👍🏼

humble/distribution.yaml
```diff
@@ -18,7 +18,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/TechMagicKK/aandd_ekew_driver_py-release.git
+      url: https://github.com/ros2-gbp/aandd_ekew_driver_py-release.git
```
clalancette, line 21: There is something wrong with the either the source repository or the release repository corresponding to this package.  In particular, the package name in the [package.xml](https://github.com/TechMagicKK/aandd_ekew_driver_py/blob/338b66f831794793ba0658a53b8067a9a902022d/package.xml#L4) is currently `aandd_ekew_driver`, but the package name as specified here is `aandd_ekew_driver_py` (we use the key name as the package name if there is no `packages` subkey).  This is causing CI here to fail.

@Kansea can you take a look?

clalancette (changes requested):

humble/distribution.yaml
Kansea: @clalancette 
The reason for this error is likely due to an incorrect default branch in the source repository.
We have now changed the default branch to the correct one called "humble" that corresponds to `aandd_ekew_driver_py`. Please try again. Thank you.

Kansea: @clalancette 
The reason for this error is likely due to an incorrect default branch in the source repository. 
We have now changed the default branch to the correct one called "humble" that corresponds to it, the package name in the [package.xml](https://github.com/TechMagicKK/aandd_ekew_driver_py/blob/d3139437c1c06d132d2b0e5ae1c074ae65a233cb/package.xml#L4) is currently `aandd_ekew_driver_py` now. 
Please try again. Thank you.

clalancette: CI is showing the same error.

Kansea: Could you tell me which branch CI uses when executing?

clalancette: The branch name itself is whatever is specified in here, so the `humble` branch.

But the package name is encoded into the release repository, as well as here.  Since that has been changed, I think you'll need to do a new bloom-release.  Though that is a bit tricky since `bloom` is probably going to target the old release repository, not the new one.

I'm actually not sure how to handle this; @nuclearsandwich could you give some advice?

longjie0723: The new release repository(https://github.com/ros2-gbp/weight_scale_interfaces-release.git and https://github.com/ros2-gbp/aandd_ekew_driver_py-release.git) contains the files for the older version, so I guess we need to do bloom-release once with the new release repository to pass the test. We can override the release repository by using these options.
@Kansea Could you try to do it on the both packages?
```
  --override-release-repository-url OVERRIDE_RELEASE_REPOSITORY_URL
                        override the release repository url; the 'Release Repository Push
                        URL' track configuration is ignored
  --override-release-repository-push-url OVERRIDE_RELEASE_REPOSITORY_PUSH_URL
                        override the 'Release Repository Push URL' track configuration; can
                        be used in conjunction with --override-release-repository-url
```

longjie0723: @clalancette Could you give a shot to run the CI check one more time? The release repo of aandd_ekew_driver_py was  updated yesterday, so I guess it would be solved.

longjie0723: @clalancette Thank you for the advice!
@Kansea Could you do it on your side?
We have once[ successfully build](https://build.ros2.org/job/Hbin_uJ64__aandd_ekew_driver_py__ubuntu_jammy_amd64__binary/8/) our packages with previous release repo, so I believe it is a problem of the inconsistency of the new release repo.

Kansea: Ok. I will do it today.

Kansea: @clalancette 
I have opened up a new pull request, could you please continue in the new [PR](https://github.com/ros/rosdistro/pull/38641), and close this PR.

PR #38572, opened 2023-09-18
point_cloud_transport: 1.0.12-1 in 'humble/distribution.yaml' [bloom]

ahcorde:
Increasing version of package(s) in repository `point_cloud_transport` to `1.0.12-1`:

- upstream repository: https://github.com/ros-perception/point_cloud_transport
- release repository: https://github.com/ros2-gbp/point_cloud_transport-release.git
- bloom version: `0.11.2`

## point_cloud_transport

```
* ROS2 port
* Contributors: Alejandro Hernández Cordero, John D'Angelo
```

## point_cloud_transport_py

```
* Added point_cloud_transport_py (#26 <https://github.com/ros-perception/point_cloud_transport/issues/26>)
* Contributors: Alejandro Hernández Cordero
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -4553,6 +4553,24 @@ repositories:
       url: https://gitlab.com/ApexAI/point_cloud_msg_wrapper
       version: rolling
     status: developed
+  point_cloud_transport:
+    doc:
+      type: git
+      url: https://github.com/ros-perception/point_cloud_transport_tutorial.git
+      version: rolling
+    release:
+      packages:
+      - point_cloud_transport
+      - point_cloud_transport_py
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/ros2-gbp/point_cloud_transport-release.git
+      version: 1.0.12-1
+    source:
+      type: git
+      url: https://github.com/ros-perception/point_cloud_transport.git
+      version: rolling
+    status: maintained
   pointcloud_to_laserscan:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


humble/distribution.yaml
```diff
@@ -4553,6 +4553,24 @@ repositories:
       url: https://gitlab.com/ApexAI/point_cloud_msg_wrapper
       version: rolling
     status: developed
+  point_cloud_transport:
+    doc:
+      type: git
+      url: https://github.com/ros-perception/point_cloud_transport_tutorial.git
+      version: rolling
```
clalancette, line 4560: I'm just noticing that the branch here is `rolling`, which is the same as in [rolling/distribution.yaml](https://github.com/ros/rosdistro/blob/676cfdb79ca36a2cc5d9745a48e8e8b0768a9765/rolling/distribution.yaml#L3731-L3749).  However, the released version is 1.0.12 here, but 2.0.0 in Rolling.  Is this all correct?

ahcorde: For the documentation we only have `rolling`

clalancette: Using a different repository for the documentation is unusual.  It's not disallowed, but what is listed here in `doc` is what we will use when generating API documentation for the package.  So what you are specifying here is that we should generate API documentation for `point_cloud_transport_tutorial`, not for `point_cloud_transport`.

My suggestion here is that you change this to https://github.com/ros-perception/point_cloud_transport.git; the documentation for the tutorial will be specified via the `point_cloud_transport_tutorial` key below.  This should probably be fixed in rolling/distribution.yaml as well.

Further, the source branch below also points to `rolling`, which I think should be `humble` instead.

clalancette (approved):

PR #38658, opened 2023-09-29
Adding python-gsmmodem-new

DavidD121:
added python-gsmmodem-new

## Package name:

python-gsmmodem-new

## Package Upstream Source:

https://github.com/babca/python-gsmmodem/tree/master

## Purpose of using this:

Control an attached GSM modem: send/receive SMS messages, handle calls, etc

Distro packaging links:

  - Pip: https://pypi.org/project/python-gsmmodem-new/
  - Pip: https://pypi.org/project/python-gsmmodem-new/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1881,11 +1881,13 @@ python-grpcio:
       pip: [grpcio]
 python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
   debian:
-    pip:
-      packages: [python-gsmmodem-new]
+    '*':
+      pip:
+        packages: [python-gsmmodem-new]
   ubuntu:
-    pip:
-      packages: [python-gsmmodem-new]
+    '*':
+      pip:
+        packages: [python-gsmmodem-new]
 python-gst:
   arch: [gstreamer0.10-python]
   debian:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1880,6 +1880,9 @@ python-grpcio:
     bionic:
       pip: [grpcio]
 python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
+  debian:
+    pip:
+      packages: [python-gsmmodem-new]
   ubuntu:
     pip:
       packages: [python-gsmmodem-new]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1879,6 +1879,10 @@ python-grpcio:
     '*': [python-grpcio]
     bionic:
       pip: [grpcio]
+python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
+  ubuntu:
+    pip:
+      packages: [python-gsmmodem-new]
 python-gst:
   arch: [gstreamer0.10-python]
   debian:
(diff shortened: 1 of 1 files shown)


methylDragon (changes requested):
Thanks for the contribution! Could you add the pip packages under the "*" match rules like some of the rosdep bindings below?

DavidD121:
> Thanks for the contribution! Could you add the pip packages under the "*" match rules like some of the rosdep bindings below?

Just to be clear, are you looking for something more general like this:
```
python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
  '*':
    pip:
      packages: [python-gsmmodem-new]
```
or to use '*' matching within the distros? i.e:
```
python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
  debian:
    '*':
      pip:
        packages: [python-gsmmodem-new]
  ubuntu:
    '*':
      pip:
        packages: [python-gsmmodem-new]
```

methylDragon:
> > Thanks for the contribution! Could you add the pip packages under the "*" match rules like some of the rosdep bindings below?
> 
> Just to be clear, are you looking for something more general like this:
> 
> ```
> python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
>   '*':
>     pip:
>       packages: [python-gsmmodem-new]
> ```
> 
> or to use '*' matching within the distros? i.e:
> 
> ```
> python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
>   debian:
>     '*':
>       pip:
>         packages: [python-gsmmodem-new]
>   ubuntu:
>     '*':
>       pip:
>         packages: [python-gsmmodem-new]
> ```

I think I would go with the latter, matching the distros!

quarkytale (approved):

nuclearsandwich:
@methylDragon I'm slightly perplexed by your suggestions as they contradict the ones I would make and I want to make sure that I'm not misunderstanding the rules or out of date on current preferences.

Is there a reason _not_ to elide the ubuntu and debian version field and to prefer an explicit wildcard there? I've always advocated eliding where possible in order to keep the rules as readable as possible.

rosdep/python.yaml
```diff
@@ -6319,6 +6328,7 @@ python3-grpcio:
   ubuntu:
     '*': [python3-grpcio]
     bionic: null
+python3-gsmmodem-pip: *migrate_eol_2025_04_30_python3_gsmmodem_pip
```
clalancette, line 6331: Personally, I don't really think we need the migration key.  Python 2 is dead, so adding in new keys for it doesn't seem to make much sense.  So I would suggest just making the key `python3-gsmmodem-pip`, and dispensing with the `python-gsmmodem-pip` key and the migration anchor altogether.

rosdep/python.yaml
```diff
@@ -1879,6 +1879,15 @@ python-grpcio:
     '*': [python-grpcio]
     bionic:
       pip: [grpcio]
+python-gsmmodem-pip: &migrate_eol_2025_04_30_python3_gsmmodem_pip
+  debian:
+    '*':
+      pip:
+        packages: [python-gsmmodem-new]
```
clalancette, line 1886: It's a little odd that this is a `-new` version of this package.  Is this a fork of the original pip package?  If so, should we name the key `python3-gsmmodem-new-pip` instead?

clalancette:
There's been no response in 3 weeks here, so I'll close this out.  Please feel free to respond to the latest feedback and reopen.

PR #38721, opened 2023-10-05
Add robot_module_msgs and digital_interface_msgs

smihael:
## Package name:

1. robot_module_msgs
2. digital_interface_msgs

## Package Upstream Source:

1. https://github.com/ReconCycle/robot_module_msgs
2. https://github.com/ReconCycle/digital_interface_msgs

## Purpose of using this:

1. Provides message, service, and action definitions for general purpose robot-agnostic controllers. Notably, it includes JointCommand.msg and CartesianCommand.msg, which include fields to specify positions, velocities, accelerations and additional force applied by the robot as well as an option to set robot compliance. it is used in https://github.com/abr-ijs/ijs_controllers
3. Provides message and service definitions for controlling GPIO pins. It is used for https://github.com/ReconCycle/raspi_ros and https://github.com/ReconCycle/rqt_raspi_ros

# Checks

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1977,6 +1977,16 @@ repositories:
       url: https://github.com/ros-mobile-robots/diffbot.git
       version: noetic-devel
     status: developed
+  digital_interface_msgs:
+    doc:
+      type: git
+      url: https://github.com/ReconCycle/digital_interface_msgs
+      version: release
+    source:
+      type: git
+      url: https://github.com/ReconCycle/digital_interface_msgs
+      version: release
+    status: maintained
   dingo:
     doc:
       type: git
@@ -8781,6 +8791,16 @@ repositories:
       url: https://github.com/cra-ros-pkg/robot_localization.git
       version: noetic-devel
     status: developed
+  robot_module_msgs:
+    doc:
+      type: git
+      url: https://github.com/ReconCycle/robot_module_msgs.git
+      version: release
+    source:
+      type: git
+      url: https://github.com/ReconCycle/robot_module_msgs.git
+      version: release
+    status: maintained
   robot_navigation:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1980,11 +1980,11 @@ repositories:
   digital_interface_msgs:
     doc:
       type: git
-      url: https://github.com/ReconCycle/digital_interface_msgs
+      url: https://github.com/ReconCycle/digital_interface_msgs.git
       version: release
     source:
       type: git
-      url: https://github.com/ReconCycle/digital_interface_msgs
+      url: https://github.com/ReconCycle/digital_interface_msgs.git
       version: release
     status: maintained
   dingo:


methylDragon:
I am unsure if the message definitions are general enough to be taking a package name in the global namespace.
It's generally a good idea to [prefix the package names](https://ros.org/reps/rep-0144.html).

This is especially because some messages are:
- [Specific to raspberry pis](https://github.com/ReconCycle/digital_interface_msgs/blob/release/srv/ConfigRead.srv)
- Or have potentially more general analogues in major, pre-existing prefixed packages: [JointCommand](https://github.com/ReconCycle/robot_module_msgs/blob/release/msg/JointCommand.msg) ([moveit_msgs/RobotTrajectory](http://docs.ros.org/en/noetic/api/moveit_msgs/html/msg/RobotTrajectory.html))

methylDragon:
I'd wait for someone else to chime in. Maybe @tfoote ?

tfoote (changes requested):
Yes, these package names are not specific enough. 

My first reaction is digital interface to what? This is not the package that will provide all digital interfaces.  Likewise what robot's module are these messages for? This is not provideing all messages possible for any robot module. 

Based on the use in raspi_ros the `raspi_` prefix might make sense. 

Though I'll note that that package doesn't follow the naming guides either. It should be something more like `raspi_deployment` If I read the README correctly. `ros` in a ROS package name is redundant as it is a ROS package.

smihael:
Thank you for your constructive comments. The package names are inherited from various previous projects, but we will consider renaming them (@BorisKuster).

`digital_interface_msgs` are used in conjunction with [raspi_ros](https://github.com/ReconCycle/raspi_ros) package to expose GPIOs as named ROS services, but the messages are more general and could be used with any other board providing GPIOs. @Rok-Pahic, perhaps `gpio_msgs` would be better?

`robot_module_msgs` are used in conjunction with internally developed controllers within a broader modular software architecture. We could go with some non-descriptive prefix `ijs_` (institution name). I don't agree that the messages can be replaced with "more general analogues in major, pre-existing prefixed packages".

Both [moveit_msgs/CartesianTrajectoryPoint and JointTrajectoryPoint](http://docs.ros.org/en/noetic/api/moveit_msgs/html/index-msg.html) are fairly new (robot_module_msgs existed since 2016)don't support setting compliance. 

Furthermore, there are no action server definitions and messages for executing
- joint space point-to-point trajectory with trapezoidal velocity profile,
- Cartesian space straight line point-to-point motion & quaternion SLERP trajectory with minimum jerk
time evolution,
- joint space point-to-point trajectory with trapezoidal velocity profile, with the initial and final pose provided
in Cartesian space and transformed into joint space using inverse kinematics,
- dynamic movement primitive (DMP) in joint space,
- Cartesian space DMP.

Alternatively, the messages could be split into multiple packages, I'm open to naming suggestions.

sloretz:
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2023-10-12/)

tfoote:
An `ijk_robot_module_msgs` makes complete sense. 

The name `gpio_msgs` is again very broad claim implicitly suggesting that it covers all gpio for the whole ROS ecosystem. I would suggest `raspi_gpio_msgs` as it appears to be very specific to one implementation. If it were to be a generic implementation it should be an abstraction of the gpio that can be self documenting.

quarkytale:
@smihael friendly ping, are you planning on taking this forward?

smihael:
> @smihael friendly ping, are you planning on taking this forward?

We discussed this internally and will proceed with refactoring. Tentatively, this should be done by the end of this month.

nuclearsandwich:
@smihael thanks for the update. Since a new release will be required for updated package names, I am going to close this PR. It would also be very helpful if you would link it and this conversation when opening the follow-up so that the reviewers there have the full discussion context.

PR #38794, opened 2023-10-21
Add gazebo_video_monitors in humble and iron

nlamprian:

Humble and Iron

# The source is here:

https://github.com/nlamprian/gazebo_video_monitors

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -1970,6 +1970,16 @@ repositories:
       url: https://github.com/ros-simulation/gazebo_ros_pkgs.git
       version: ros2
     status: maintained
+  gazebo_video_monitors:
+    doc:
+      type: git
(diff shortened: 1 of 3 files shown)


quarkytale (approved):

quarkytale:
It turns out that all 3 distros are currently not in a sync hold, so we can do this all at once. But keep in mind for the future that cross-distribution changes can be hard to get in since we can only merge when all of them are not held.

quarkytale:
@nlamprian missing LICENSE files!!! Merged it by mistake before adding the comment, can you please add a top-level license?

nlamprian:
It's been a while, but isn't a [COPYING](https://github.com/nlamprian/gazebo_video_monitors/blob/ros2/COPYING) file what is required by GPLv3?

quarkytale:
Ooh okay wasn't aware of that, thanks for the update, all good!

PR #38859, opened 2023-11-05
Add rocky_tf_monitor in noetic/distribution.yaml

rkoyama1623:

tf_monitor

# The source is here:

https://github.com/rkoyama1623/tf_monitor.git

# Checks

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -9048,6 +9048,15 @@ repositories:
       url: https://github.com/robotraconteur/robotraconteur.git
       version: ros-noetic
     status: maintained
+  rocky_tf_monitor:
+    doc:
+      type: git
+      url: https://github.com/rkoyama1623/rocky_tf_monitor.git
+      version: main
+    source:
+      type: git
+      url: https://github.com/rkoyama1623/rocky_tf_monitor.git
+      version: main
   ros:
     doc:
       type: git


nuclearsandwich:
New package review:

  - [x] Top level license file: https://github.com/rkoyama1623/tf_monitor/blob/main/LICENSE
  - [ ] Per package license files
  - [x] tf_monitor: https://github.com/rkoyama1623/tf_monitor/blob/main/package.xml
- [x] Public source repo: https://github.com/rkoyama1623/tf_monitor/tree/main/
- [ ] Source repository contains ROS packages and each package meets [REP-144](https://www.ros.org/reps/rep-0144.html) naming conventions
  - [ ] tf_monitor


@audrow I could a second opinion on this package name before approving the name.
The name is accurate but I'm not sure if it counts as too general.

clalancette:
I think the package name here could be confusing, because we already have a `tf2_monitor` command available from geometry2: https://github.com/ros2/geometry2/blob/rolling/tf2_ros/src/tf2_monitor.cpp .  I know that is for ROS 2, but I still think this could be confusing.

How about `rkoyama_tf_monitor` instead?

sloretz:
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2023-11-09/)

rkoyama1623:
@nuclearsandwich, @clalancette, Thank you for your considerations and naming suggestion.
rkoyama is my personal name:laughing:. I'd like to use a little more generic name.
I named this ros package rocky_tf_monitor, since this tool is used to inspect rocky tf topics, and rocky is a bit like my name.

clalancette:
> I named this ros package rocky_tf_monitor, since this tool is used to inspect rocky tf topics, and rocky is a bit like my name.

That sounds good to me.

clalancette (approved):

PR #38913, opened 2023-11-08
added libcaer to humble/iron/rolling

berndpfrommer:
package name: libcaer
distros: humble, iron, rolling

# The source is here:
https://github.com/ros-event-camera/libcaer

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2978,6 +2978,16 @@ repositories:
       type: git
       url: https://github.com/lgsvl/lgsvl_msgs.git
       version: foxy-devel
+  libcaer:
+    doc:
+      type: git
+      url: https://github.com/ros-event-camera/libcaer.git
+      version: ros_event_camera
+    source:
+      type: git
+      url: https://github.com/ros-event-camera/libcaer.git
+      version: ros_event_camera
+    status: developed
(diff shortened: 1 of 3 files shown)


humble/distribution.yaml
```diff
@@ -2978,6 +2978,16 @@ repositories:
       type: git
       url: https://github.com/lgsvl/lgsvl_msgs.git
       version: foxy-devel
+  libcaer:
```
clalancette, line 2981: Can you please explain the relationship between `libcaer` here and `libcaer_driver` below?

berndpfrommer: libcaer_driver depends on libcaer (a library that provides access to all of Inilab's devices).
I first tried to compile and install libcaer directly inside the libcaer_driver workspace but couldn't get it to work so I decided to provide it as a separate package. That will also make it easier if somebody wants to access libcaer directly for other ROS projects.

clalancette (approved):
This looks OK to me, but we can't merge it because `iron` is in a sync-hold.

@audrow I'd appreciate a look from you as well.

berndpfrommer:
Is some input missing from my side for this PR to move forward? I think iron is no longer in sync hold.

PR #38926, opened 2023-11-10
Update distribution.yaml

RoBeau:

## Package name:

boeing_gazebo_model_attachment_plugin (humble entry)
boeing_set_joint_positions_plugin

## Package Upstream Source:

https://github.com/Boeing/gazebo_model_attachment_plugin
https://github.com/Boeing/gazebo_set_joint_positions_plugin

## Purpose of using this:

This package allows for models to be linked and un linked during runtime using a ros service
Provides 

This plugin sets a RobotModel's joint values to the latest published values on a speicifed JointState topic from ROS.

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -705,6 +705,15 @@ repositories:
       url: https://github.com/flynneva/bno055.git
       version: main
     status: maintained
+  boeing_gazebo_model_attachment_plugin:
+    doc:
+      type: git
+      url: https://github.com/Boeing/gazebo_model_attachment_plugin.git
+      version: humble
+    source:
+      type: git
+      url: https://github.com/Boeing/gazebo_model_attachment_plugin.git
+      version: humble
   bond_core:
     doc:
       type: git
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -714,6 +714,15 @@ repositories:
       type: git
       url: https://github.com/Boeing/gazebo_model_attachment_plugin.git
       version: humble
+  boeing_gazebo_set_joint_positions_plugin:
+    doc:
+      type: git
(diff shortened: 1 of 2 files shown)


humble/distribution.yaml
```diff
@@ -705,6 +705,24 @@ repositories:
...
       version: main
     status: maintained
+  boeing_gazebo_model_attachment_plugin:
+    doc:
+      type: git
+      url: https://github.com/Boeing/gazebo_model_attachment_plugin.git
+      version: humble
+    source:
+      type: git
+      url: https://github.com/Boeing/gazebo_model_attachment_plugin.git
```
clalancette, line 715: So the big problem with this is the modified license on the repository.  Because it is not one of the standard OSI-approved licenses, we can't say that we are allowed to redistribute the results, and we are not lawyers.

I think the only ways forward here are to change this to an OSI-approved license (i.e. plain Apache-2.0), or close this PR.

clalancette (dismissed):

humble/distribution.yaml
RoBeau: We will investigate changing this to plain Apache-2.0.

mjcarroll: @RoBeau friendly ping

RoBeau: Hi, we have been given verbal approval to change to the plain Apache-2.0 license. Just waiting for final sign off in the next week or so.

sloretz:
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2023-12-11/)

humble/distribution.yaml
RoBeau: @clalancette thanks for you patients, I have updated both packages to use plain Apache-2.0.

quarkytale (approved):

PR #38962, opened 2023-11-13
Add adi_tmcl to humble index

jmacagba:

humble

# The source is here:

https://github.com/analogdevicesinc/tmcl_ros2

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -79,6 +79,16 @@ repositories:
       url: https://github.com/ros-acceleration/adaptive_component.git
       version: rolling
     status: developed
+  adi_tmcl:
+    doc:
+      type: git
+      url: https://github.com/analogdevicesinc/tmcl_ros2.git
+      version: humble
+    source:
+      type: git
+      url: https://github.com/analogdevicesinc/tmcl_ros2.git
+      version: humble
+    status: maintained
   aerostack2:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


humble/distribution.yaml
```diff
@@ -7991,6 +7991,16 @@ repositories:
       url: https://github.com/ros2/tlsf.git
       version: humble
     status: maintained
+  tmcl_ros2:
```
clalancette, line 7994: I think that the name of this package violates our naming guidelines in https://www.ros.org/reps/rep-0144.html .  In particular, it shouldn't have `ros2` in the title, as that would lead to package names like:

```
ros-humble-tmcl-ros2
```

Further, I think that it would be best to prepend the name of your organization to the package name.  So I'll suggest renaming the package to `adi_tmcl`, which I think should fix both of these issues.

clalancette (changes requested):

humble/distribution.yaml
jmacagba: As you suggested, I will update the package name to `adi_tmcl` within the week.
I will inform you once updated.

jmacagba: Hi @clalancette , I have now updated the package name to `adi_tmcl`.

clalancette (approved):
Thanks for doing the rename, it is much appreciated.

jmacagba:
Thank you @clalancette

PR #38989, opened 2023-11-15
Update distribution.yaml with rrt_planner

david-dorf:

## Package name:

rrt_planner

## Package Upstream Source:

https://github.com/daviddorf2023/rrt_planner.git

## Purpose of using this:

There used to be a package in ROS 1 for doing path planning with Rapidly Exploring Random Trees (RRTs), called rrt_exploration. That package has not been added to ROS 2, and features such as 3D path planning and marker obstacle avoidance. These features are present in rrt_planner, where a user can publish an occupancy grid or marker in 2D or 3D, and the algorithm will avoid the walls/obstacles. It then publishes the nodes as markers during runtime, and publishes a path if one is found within the node limit.

Distro packaging links:

  - https://github.com/daviddorf2023/rrt_planner.git
  - https://github.com/daviddorf2023/rrt_planner.git
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available

Iron

# The source is here:

https://github.com/daviddorf2023/rrt_planner.git

# Checks

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -6225,6 +6225,22 @@ repositories:
       url: https://github.com/ros-visualization/rqt_topic.git
       version: iron
     status: maintained
+  rrt_planner:
+    doc:
+      type: git
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: iron
+    release:
+      tags:
+        release: release/iron/{package}/{version}
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: 1.0.0
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: iron
+    status: maintained
   rsl:
     doc:
       type: git


iron/distribution.yaml
```diff
@@ -6225,6 +6225,22 @@ repositories:
...
+  rrt_planner:
+    doc:
+      type: git
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: iron
+    release:
+      tags:
+        release: release/iron/{package}/{version}
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: 1.0.0
```
clalancette, line 6237: You haven't setup a release repository, or done a proper bloom-release yet, so this section shouldn't exist:

```suggestion
```

iron/distribution.yaml
```diff
@@ -6225,6 +6225,22 @@ repositories:
       url: https://github.com/ros-visualization/rqt_topic.git
       version: iron
     status: maintained
+  rrt_planner:
+    doc:
+      type: git
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: iron
```
clalancette, line 6232: There's no `iron` branch on the github repository.  This should be `main`.

iron/distribution.yaml
```diff
@@ -6225,6 +6225,22 @@ repositories:
...
+    release:
+      tags:
+        release: release/iron/{package}/{version}
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: 1.0.0
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/daviddorf2023/rrt_planner.git
+      version: iron
```
clalancette, line 6242: There's no `iron` branch on the github repository.  This should be `main`.

iron/distribution.yaml
```diff
@@ -6225,6 +6225,22 @@ repositories:
       url: https://github.com/ros-visualization/rqt_topic.git
       version: iron
     status: maintained
+  rrt_planner:
+    doc:
+      type: git
+      url: https://github.com/daviddorf2023/rrt_planner.git
```
clalancette, line 6231: One problem I have with this is that the name `rrt_planner` is too generic.  I can imagine there being multiple implementations of an RRT planner in ROS.  This comes under our naming policies in https://www.ros.org/reps/rep-0144.html#global-guidelines .

What I'll suggest here is that this package gets renamed to something like `daviddorf_rrt_planner`, which I think would satisfy REP-144 above.

clalancette (changes requested):

clalancette:
No response in two weeks, so I'm going to close this out.  If you'd still like to see this in, please make the changes suggested in this PR and make a new PR.  Thanks.

PR #39010, opened 2023-11-17
rosdep: nixos: Add transforms3d packages

hacker1024:


rosdep/python.yaml
```diff
@@ -4276,6 +4276,7 @@ python-transforms3d-pip:
   debian:
     pip:
       packages: [transforms3d]
+  nixos: [python3Packages.transforms3d]
```
clalancette, line 4279: There's no real reason to add to this one; this is implicitly Python 2, which is no longer supported.

```suggestion
```

rosdep/python.yaml
```diff
@@ -9113,6 +9114,7 @@ python3-transforms3d:
...
   fedora: [python3-transforms3d]
```
clalancette, line 9118: This needs to be in alphabetical order:

```suggestion
  fedora: [python3-transforms3d]
  nixos: [python3Packages.transforms3d]
```

clalancette (changes requested):

tfoote:
I merged the fix for the formatting but I don't have a link to validate the package. Please fill out the description with more information so we can validate the package linked is correct.

sloretz (approved):
https://search.nixos.org/packages?channel=23.11&show=python311Packages.transforms3d&from=0&size=50&sort=relevance&type=packages&query=transforms3d

clalancette (approved):

PR #39021, opened 2023-11-17
Set topic_tools branches to distro branches

christophebedard:
We're switching from using a `main` branch for all distros to using distro-specific branches, like `humble`, `iron`, and `rolling`. See https://github.com/ros-tooling/topic_tools/issues/62. The branches all currently have the same version (and tag) and have not really diverged. New releases will be made soon to make new improvements available.

I've updated the branch names in the release repo: https://github.com/ros2-gbp/topic_tools-release/commit/6ae603ac0b5ecd039e99839aceae646da3ddfa5a

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -8029,7 +8029,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-tooling/topic_tools.git
-      version: main
+      version: humble
     release:
       packages:
       - topic_tools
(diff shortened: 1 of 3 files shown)


christophebedard:
We'll have to wait for the end of the Humble hold, which should be on November 20th: https://discourse.ros.org/t/preparing-for-humble-sync-and-patch-release-2023-11-20/34628/5

christophebedard:
I think all sync holds are over.

clalancette (approved):

PR #39063, opened 2023-11-22
Add rosdep key for python3-flask-restful

waldgaenger:

## Package name:
python3-flask-restful

## Package Upstream Source:
https://packages.ubuntu.com/jammy/python3-flask-restful

## Purpose of using this:
Flask-RESTful is an extension for Flask that adds support for quickly building REST APIs.

Distro packaging links:

  - https://packages.ubuntu.com/jammy/python3-flask-restful
  - https://packages.debian.org/bookworm/python3-flask-restful
  - not available
  -  not available
  - not available
  - not available'
  - not available
  - not available'
  - not available'
  - not available'

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6001,6 +6001,8 @@ python3-flask:
   nixos: [python3Packages.flask]
   rhel: [python3-flask]
   ubuntu: [python3-flask]
+python3-flask-restful:
+  ubuntu: [python3-flask-restful]
 python3-flask-bcrypt:
   debian: [python3-flask-bcrypt]
   nixos: [python3Packages.flask-bcrypt]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6001,6 +6001,9 @@ python3-flask:
   nixos: [python3Packages.flask]
   rhel: [python3-flask]
   ubuntu: [python3-flask]
+python3-flask-restful:
```
clalancette, line 6004: This is going to fail CI since it should be in alphabetical order.

clalancette (changes requested):

rosdep/python.yaml
sloretz: @waldgaenger friendly ping, mind making this alphabetical so the linter passes?

PR #39111, opened 2023-11-30
python-omniorb is not available on debian, use omniorb, see https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=omniidl

k-okada:

## Package name:

python-omniorb

## Package Upstream Source:

`[python-omniorb, python-omniorb-omg, omniidl-python]` is not avilable on debian.

https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=omniidl

This fix should resolve build failure on https://build.ros.org/job/Nbin_db_dB64__openhrp3__debian_buster_amd64__binary/289/console ()

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2562,7 +2562,7 @@ python-objectpath-pip:
       packages: [objectpath]
 python-omniorb:
   arch: [omniorbpy]
-  debian: [python-omniorb, python-omniorb-omg, omniidl-python]
+  debian: [omniorb, omniidl]
   fedora: [python-omniORB, omniORB-devel]
   gentoo: ['net-misc/omniORB[python]']
   ubuntu:


rosdep/python.yaml
```diff
@@ -2562,7 +2562,7 @@ python-objectpath-pip:
       packages: [objectpath]
 python-omniorb:
   arch: [omniorbpy]
-  debian: [python-omniorb, python-omniorb-omg, omniidl-python]
+  debian: [omniorb, omniidl]
```
sloretz, line 2565: I don't see any Python files in `omniorb`. Is `omniidl` enough here?

https://packages.debian.org/sid/amd64/omniorb/filelist
https://packages.debian.org/sid/amd64/omniidl/filelist

It looks like `python-omniorb` was a key in older versions of debian: https://tracker.debian.org/pkg/python-omniorb

I'm not really seeing the key in fedora either: https://src.fedoraproject.org/rpms/omniORB/blob/rawhide/f/omniORB.spec

Maybe it's time we drop this key in favor of https://github.com/ros/rosdistro/blob/10c5cc83d334bc32d1c614a0ef341b3b6ca0f343/rosdep/base.yaml#L6857 ?

clalancette: Looking at the link from @sloretz , it looks like this key was dropped from Debian in 2018.  So I think we can just remove this key completely.

k-okada:
thank you for comments, I'll use `omniidl` `omniorb` rosdep keys insted.

https://github.com/tork-a/openhrp3-release/commit/507893d16dcfc7900a0e9d4d1b68194752aa5b27

PR #39112, opened 2023-11-30
Update distribution.yaml

Tacha-S:

humble

# The source is here:

https://github.com/Tacha-S/sync_parameter_server

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -7889,6 +7889,11 @@ repositories:
       url: https://github.com/swri-robotics/swri_console.git
       version: ros2-devel
     status: developed
+  sync_parameter_server:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


mjcarroll:
Per https://www.ros.org/reps/rep-0144.html, I think that `sync_parameter_server` may be too generic without widespread community adoption.

Can you change the name to be more specific to your organization/application (for instance `tacha_sync_parameter_server`.

mjcarroll:
Since the name here is generic, we discussed it as a team. Since the name accurately reflects what you are doing in the package, and the package is generically useful, there is no need to do a renaming here.

mjcarroll (approved):

Tacha-S:
Thank you for considering the package positively. Please feel free to contact me if there are any issues.

PR #39172, opened 2023-12-07
Switch the RHEL-9 python3-qt5-bindings key to SIP.

clalancette:
Pyside2 no longer works on RHEL, so instead switch to SIP, which now works.

This is the last piece of switching RHEL-9 to using SIP, after https://github.com/ros2/ci/pull/730, https://github.com/ros-visualization/python_qt_binding/pull/129, and https://github.com/ros2/ci/pull/732 .

Marked as a draft for now; I'd like to get @cottsay's approval before we merge this.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8241,7 +8241,7 @@ python3-qt5-bindings:
   openembedded: [python3-pyqt5@meta-qt5]
   opensuse: [python3-qt5]
   rhel:
-    '*': [clang, python3-devel, python3-qt5-devel, python3-sip-devel]
+    '*': [python3-qt5-devel, python3-sip-devel, libXext-devel]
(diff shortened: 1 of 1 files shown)


cottsay (approved):
LGTM. There may be other packages that we need to add to this list like `libXext-devel`, but I'd rather not guess.

We should be able to drop `clang` now that we're not using Shiboken2 anymore, but that's up to you.

clalancette:
> We should be able to drop `clang` now that we're not using Shiboken2 anymore, but that's up to you.

@cottsay Actually, I went and made this look the same as the `fedora` key.  Either way, they should be more-or-less equivalent.  Please take another look; if you are still happy with this, then I'll mark it as non-draft.  Thanks!

cottsay (approved):
Let's give it a shot!

mjcarroll (approved):

PR #39181, opened 2023-12-10
add Pangolin to humble and rolling

christianrauch:

## Package name:

pangolin

## Package Upstream Source:

https://github.com/stevenlovegrove/Pangolin.git

## Purpose of using this:

Pangolin is a UI toolkit that is widely used in the robotics and specifically the SLAM community to render user interfaces.

`humble`, `rolling`

# The source is here:

https://github.com/stevenlovegrove/Pangolin.git

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -4583,6 +4583,11 @@ repositories:
       url: https://github.com/pal-robotics/pal_urdf_utils.git
       version: humble-devel
     status: developed
+  pangolin:
+    source:
+      type: git
+      url: https://github.com/stevenlovegrove/Pangolin.git
+      version: master
   pcl_msgs:
     release:
       tags:
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -3653,6 +3653,11 @@ repositories:
       url: https://github.com/pal-robotics/pal_statistics.git
       version: humble-devel
     status: maintained
+  pangolin:
+    source:
+      type: git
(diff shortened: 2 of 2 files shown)


audrow:
@christianrauch, would you like to add a docs entry?

christianrauch:
> @christianrauch, would you like to add a docs entry?

I am just following the ["Create a new release repository"](https://docs.ros.org/en/foxy/How-To-Guides/Releasing/Release-Team-Repository.html#create-a-new-release-repository) which says
> you should first open a pull request on ros/rosdistro adding a `source` entry for your repository

It does not mention the `doc` section and I also would not know what I should add there. Glimpsing over some of the other entries in the yaml, they just seem to use the same source repository url. That's redundant but I guess I could do the same if it is required.

nuclearsandwich:
> It does not mention the `doc` section and I also would not know what I should add there. Glimpsing over some of the other entries in the yaml, they just seem to use the same source repository url. That's redundant but I guess I could do the same if it is required.

The doc section is generally meant to support documentation builds for the package, which often do rely on the source repository (e.g. for doxygen based API docs) it's encouraged for libraries that have API docs but not strictly required.

One thing that stands out is that the project's author is not the one submitting it but I do see @christian-rauch as a contributor of the `package.xml` file upstream. There's no restriction to submitting a library written by someone else as long as it does not conflict with a project already present in the rosdistro database but it is worth an FYI to @stevenlovegrove that their package is being proposed for inclusion in the rosdistro index.

nuclearsandwich (approved):

PR #39203, opened 2023-12-12
rosdep: nixos: Add OpenCV cxxdev output

hacker1024:
Nixpkgs (the Nix package collection) has recently [changed](https://github.com/NixOS/nixpkgs/commit/ada3991349beb5880e3994f25c65a0cf68941b83) the way CUDA dependency resolution works.

Packages that use OpenCV as a C++ dependency should now also include `opencv.cxxdev` along with `opencv` in order to allow CUDA to work correctly. When not using CUDA, this change is harmless.


nuclearsandwich (approved):

nuclearsandwich:
rosdep PRs require two approvals before merging.

nuclearsandwich:
@audrow can you provide the second review?

audrow (approved):

PR #39209, opened 2023-12-14
Add mp2p_icp for indexing

jlblancoc:

iron

# The source is here:

https://github.com/MOLAorg/mp2p_icp

# Checks

NOTE: This package is right now already published within another package ("mola") but in the next release it will be released as an independently package & git repository, hence this PR.

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -3181,6 +3181,16 @@ repositories:
       url: https://github.com/ros-planning/moveit_visual_tools.git
       version: ros2
     status: maintained
+  mp2p_icp:
+    doc:
+      type: git
+      url: https://github.com/MOLAorg/mp2p_icp.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/MOLAorg/mp2p_icp.git
+      version: master
+    status: developed
   mqtt_client:
     doc:
       type: git
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3431,6 +3431,16 @@ repositories:
       url: https://github.com/mrpt-ros-pkg/mrpt_msgs.git
       version: master
     status: maintained
+  mrpt_navigation:
+    doc:
+      type: git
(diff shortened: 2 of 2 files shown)


audrow:
Same as https://github.com/ros/rosdistro/pull/39210#issuecomment-1856575288.

jlblancoc:
Comments solved in original thread; plus `<license file=...>` attribute [added upstream](https://github.com/MOLAorg/mp2p_icp/commit/a487073e0053738fbcd483ca393c57c275ebe2b5).

iron/distribution.yaml
```diff
@@ -3181,6 +3181,16 @@ repositories:
       url: https://github.com/ros-planning/moveit_visual_tools.git
       version: ros2
     status: maintained
+  mp2p_icp:
```
clalancette, line 3184: I have a question on how this package is going to interact with https://github.com/ros/rosdistro/blob/5bb368ac3182c13773638f8a9fee63ad1202c1c1/iron/distribution.yaml#L3078 , which is already released.  As far as I understand, this new package has the same name as that package, so you won't be able to release it here.  What is your plan for dealing with this?

jlblancoc: Hi @clalancette !

Yes, you are correct. Long story short: 
- Current state: one git repo (with submodules) named "mola".
- New state: 3 git repos (no submodules anymore)
  - mola_common : Already merged for indexing in #39230  (Read the "NOTE" section at the end of that PR).
  - mp2p_icp: This PR
  - mola: New releases to be made, now of course without the other two packages above, once they are merged and bloom-released.

If it's an issue to have the same package released twice, even during 1 day, I can create a PR to remove the "release:" section of "mola". 

Thanks, and sorry for the mess, this new state of things makes more sense thinking of the modularity of new forthcoming packages.

jlblancoc:
I'll reopen this one after #39311 to avoid problems.

PR #39223, opened 2023-12-15
Revert "Remove the python-omniorb key."

k-okada:
Reverts ros/rosdistro#39174

@clalancette python-omniorb is removed from debian, but not from ubuntu yet.
- ubuntu: https://packages.ubuntu.com/focal/python-omniorb
- gento: https://packages.gentoo.org/packages/net-misc/omniORB
- fedora : https://packages.fedoraproject.org/pkgs/omniORBpy/python3-omniORB/ (removed...)
- arch: https://aur.archlinux.org/packages/omniorbpy

### DIFF
--- a/rosdep/osx-homebrew.yaml
+++ b/rosdep/osx-homebrew.yaml
@@ -659,6 +659,10 @@ python-cairo:
   osx:
     homebrew:
       packages: [py2cairo]
+python-omniorb:
+  osx:
+    homebrew:
+      packages: [omniorb]
 python-opencv:
   osx:
     homebrew:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2562,12 +2562,10 @@ python-objectpath-pip:
     pip:
       packages: [objectpath]
 python-omniorb:
-  arch: [omniorbpy]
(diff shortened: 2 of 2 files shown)


nuclearsandwich:
@k-okada I don't think it makes sense to revert the entire PR, especially since the checks have revealed places where the package is _not_ available.

If you're still using the ubuntu support for this key, a scoped restoration of that key makes sense:

```yaml
python-omniorb:
  ubuntu:
    focal: [python-omniorb, python-omniorb-omg, omniidl-python]
```

Focal is the only active Ubuntu distribution still supported by rosdep. The dependency itself is targeting Python 2 as well, for which we do not currently have any active ROS distributions, so it may be that these changes should be introduced in personal rosdep sources rather than the official database, but if you're encountering errors with this change when releasing packages into the official rosdistros that should be investigated first.

k-okada:
@nuclearsandwich ok, i have removed debian/arch from keys and specify 'focal' to ubuntu keys,

nuclearsandwich (approved):
rosdep PRs require a second reviewer.

clalancette (approved):

k-okada:
thank you

Issue #39208, opened 2023-12-14
Old python3-mypy version on Ubuntu


Wojcik98:
`python3-mypy` on Ubuntu is configured to be installed using system package [python3-mypy](https://packages.ubuntu.com/jammy/python3-mypy) which contains `mypy` in version `0.942`. That version has a bug that appears when using python version `>=3.10.7` (https://github.com/python/mypy/issues/13627), it was fixed in version `0.981`. Ubuntu's current python's version is `3.10.12`.

This prevents me from using `ament_mypy` in packages as mypy crashes.

`mypy` on pip is in the latest version `1.7.1`.

Should the rules here in `rosdistro` be changed to use `pip` installer for ubuntu, or should this be on mypy/ubuntu maintainers to release new version for ubuntu?

nuclearsandwich:
> Should the rules here in rosdistro be changed to use pip installer for ubuntu, or should this be on mypy/ubuntu maintainers to release new version for ubuntu?

Version stability is part of the necessary trade-off for long term support distributions. See this comment https://github.com/ros/rosdistro/pull/39226#issuecomment-1858402401 for why pip cannot provide this dependency.

PR #39226, opened 2023-12-15
python3-mypy: change Ubuntu rule to use pip

Wojcik98:
Please change the `python3-mypy` dependency in the rosdep database.

## Package name:

python3-mypy

## Package Upstream Source:

https://pypi.org/project/mypy/

## Purpose of using this:

Initially explained in #39208:
Currently, the rule for Ubuntu is to use the system's package [python3-mypy](https://packages.ubuntu.com/jammy/python3-mypy), which has a `mypy` in version `0.942` (on Ubuntu 22.04). That version contains a bug (https://github.com/python/mypy/issues/13627, fixed in `0.981`) that appears with Python's versions `>=3.10.7` (current Python's version on Ubuntu 22.04 is `3.10.12`). On PyPI, `mypy` version is the newest `1.7.1`.

Because of that mypy's bug, `ament_mypy` is crashing on my ROS packages (but works after installing `mypy` from `pip`).

This PR changes rosdep rule for Ubuntu only to use `pip` instead of the system package in order to use the latest version and ensure this and other bug fixes are implemented.

Distro packaging links:

  - ~https://packages.ubuntu.com/jammy/python3-mypy~
  - https://pypi.org/project/mypy/
- Other
  - not changed

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6980,7 +6980,9 @@ python3-mypy:
     '*': [python3-mypy]
     '7': null
     '8': null
-  ubuntu: [python3-mypy]
+  ubuntu:
+    pip:
+      packages: [mypy]
 python3-mysqldb:
   arch: [python-mysqlclient]
   debian: [python3-mysqldb]


nuclearsandwich:
@Wojcik98 thanks for opening this pull request.

Packages that are released on the build farm require all dependencies to be satisfied from the platform package manager rather than from apt. So changing the key in the official rosdep database is not an option: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1

If Ubuntu's current versions of Python and mypy are indeed broken for everyone, you can try and use Ubuntu's [stable release update](https://wiki.ubuntu.com/StableReleaseUpdates) process but it's usually very conservative in what changes it will make in a stable distribution. In your own projects and systems you can use whichever version of mypy best suits your needs but the official definitions cannot accept this change.

PR #39236, opened 2023-12-15
Added python3-build

m-elwin:

## Package name:

python3-build

## Package Upstream Source:

https://github.com/pypa/build

## Purpose of using this:

This dependency is being used for a package that makes it easier to add python code to an  ament_cmake project.
It may be use to others who want to work on various python-related build processes with ROS.

Distro packaging links:

  - https://packages.debian.org/bullseye/main/python3-build
  - https://packages.ubuntu.com/jammy/python3-build
  - Not available in focal or earlier, added pip dependency
  - https://packages.fedoraproject.org/pkgs/python-build/python3-build/
  - https://archlinux.org/packages/extra/any/python-build/
  - https://packages.gentoo.org/packages/dev-python/build
  - https://formulae.brew.sh/formula/python-build
  - https://pkgs.alpinelinux.org/package/edge/community/x86_64/py3-build
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5104,6 +5104,19 @@ python3-bson:
       packages: [bson]
   rhel: ['python%{python3_pkgversion}-bson']
   ubuntu: [python3-bson]
+python3-build:
+  alpine: [py3-build]
+  arch: [python-build]
+  debian: [python3-build]
+  fedora: [python3-build]
+  gentoo: [dev-python/build]
+  osx: [python-build]
+  rhel:
+    '*': [python3-build]
+    '8': null
+  ubuntu:
+    '*': [python3-build]
+    focal: null
 python3-cairo:
   arch: [python-cairo]
   debian: [python3-cairo]


adityapande-1995 (approved):

audrow:
@m-elwin, it looks like the package is available for RHEL, too. Would you mind adding it?

m-elwin:
@audrow added for rhel.

rosdep/python.yaml
```diff
@@ -5104,6 +5104,23 @@ python3-bson:
...
+  osx: [python-build]
+  rhel:
+    rhel8:
+      pip:
+        packages: [build]
+    rhel9: [python3-build]
+  ubuntu:
+    focal:
+      pip:
+        packages: [build]
```
nuclearsandwich, line 5122: Unless you specifically need the pip packages on Focal it's better not to add them.

```suggestion
```

rosdep/python.yaml
```diff
@@ -5104,6 +5104,23 @@ python3-bson:
...
+  alpine: [py3-build]
+  arch: [python-build]
+  debian: [python3-build]
+  fedora: [python3-build]
+  gentoo: [dev-python/build]
+  osx: [python-build]
+  rhel:
+    rhel8:
+      pip:
+        packages: [build]
```
nuclearsandwich, line 5117: Unless you specifically need the pip packages for RHEL 8 (Thanks for adding RHEL 9 in response to earlier feedback!) it's better not to add them.

```suggestion
```

nuclearsandwich (commented):
I'd like to drop the pip packages unless they're explicitly needed by a contributor in order to reduce the use of mixed pip/platform keys overall. If these pip keys are in use by the contributor in transition, then they don't need to be removed. But I'm against adding them without a concrete use case.

m-elwin:
Switched to '9' for rhel and removed pip dependencies in focal/rhel8. This package is not available via apt/yum for Focal/rhel8 as far as I can tell, but I also don't know of a need for this dependency on those platforms.

rosdep/python.yaml
```diff
@@ -5104,6 +5104,17 @@ python3-bson:
...
+  alpine: [py3-build]
+  arch: [python-build]
+  debian: [python3-build]
+  fedora: [python3-build]
+  gentoo: [dev-python/build]
+  osx: [python-build]
+  rhel:
+    '9': [python3-build]
+  ubuntu:
+    jammy: [python3-build]
```
clalancette, line 5117: Generally it is better to mark things as '*' so that they are automatically used for newer versions, and then to "null" out the older ones that aren't supported.  So in this case this would be better:

```suggestion
  ubuntu:
    '*': [python3-build]
    focal: null
```

rosdep/python.yaml
```diff
@@ -5104,6 +5104,17 @@ python3-bson:
...
   ubuntu: [python3-bson]
+python3-build:
+  alpine: [py3-build]
+  arch: [python-build]
+  debian: [python3-build]
+  fedora: [python3-build]
+  gentoo: [dev-python/build]
+  osx: [python-build]
+  rhel:
+    '9': [python3-build]
```
clalancette, line 5115: Same here:

```suggestion
  rhel:
    '*': [python3-build]
    '8': null
```

clalancette (changes requested):

rosdep/python.yaml
m-elwin: Fixed

rosdep/python.yaml
m-elwin: Fixed

clalancette (approved):

PR #39267, opened 2023-12-18
add motion_capture_tracking to humble/iron/rolling

whoenig:

Humble, Iron, Rolling

# The source is here:

https://github.com/IMRCLab/motion_capture_tracking/tree/ros2

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3693,6 +3693,16 @@ repositories:
       url: https://github.com/MOLAorg/mola.git
       version: develop
     status: developed
+  motion_capture_tracking:
+    doc:
+      type: git
+      url: https://github.com/IMRCLab/motion_capture_tracking.git
+      version: ros2
+    source:
+      type: git
+      url: https://github.com/IMRCLab/motion_capture_tracking.git
+      version: ros2
+    status: developed
   moveit:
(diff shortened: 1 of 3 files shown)


audrow (approved):

audrow (commented):
Actually, I'd like a second opinion that this package doesn't have too generic of a name.

whoenig:
Note that it's a generic name, because it supports VICON, OptiTrack, Qualisys, and several other smaller vendors in the backend. The "_tracking" part comes from the fact that we provide a custom rigid body tracker that works with single markers and identical marker configurations.

mikeferguson:
Is there any documentation on how to use this package? (the README is pretty sparse). Good documentation would probably go a long way to justifying the generic name.

whoenig:
Thanks for your swift and constructive feedback. I have updated the README (see https://github.com/IMRCLab/motion_capture_tracking/blob/ros2/README.md) as well as the comments in the `cfg.yaml` file to explain the various usage options.

audrow (approved):
The new readme provides better context, and the comments on the configuration file will probably make this much easier to use.

Thanks for adding documentation!

clalancette (approved):

PR #39311, opened 2023-12-20
Temporarily remove 'mola' package (transitional)

jlblancoc:
This PR removes the entire section for the `mola` package (and all their released packages), as a transition towards the new organization of packages as discussed in https://github.com/ros/rosdistro/pull/39209

Once this one is merged, I'll open new PRs for the corresponding new packages, now without risk of conflicts of packages with the same name coexisting.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3673,35 +3673,6 @@ repositories:
       url: https://github.com/OrebroUniversity/mod.git
       version: master
     status: developed
-  mola:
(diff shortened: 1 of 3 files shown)


clalancette (approved):

jlblancoc:
@clalancette  Thanks! The next step after this removal would be https://github.com/ros2-gbp/ros2-gbp-github-org/issues/402 so we can release the new alternatives.

PR #39350, opened 2023-12-26
Added source entry for rqt_gauges (naming review)

ahcorde:
Added source entry for rqt_gauges (naming review)


clalancette (approved):
The package name looks reasonable (we also went over this during the ROS 2 weekly call).  So going ahead and merging.

PR #39400, opened 2024-01-02
fix python3-psycopg2 for jammy

l-schilling:

## Package name:

python3-psycopg2

## Package Upstream Source:
https://pypi.org/project/psycopg2/

## Purpose of using this:

Add python3-psycopg2 for jammy instead of default python-psycopg2 as this results in following error:

```
E: Unable to locate package python-psycopg2
  apt: command [apt-get install -y python-psycopg2] failed
```

Distro packaging links:

  - https://packages.debian.org/bookworm/python3-psycopg2
  - https://packages.ubuntu.com/focal/python3-psycopg2
  - https://packages.ubuntu.com/jammy/python3-psycopg2
  - https://packages.fedoraproject.org/pkgs/python3-psycopg2/python3-psycopg2/
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2912,10 +2912,11 @@ python-psycopg2:
   debian: [python-psycopg2]
   fedora: [python-psycopg2]
   gentoo: [=dev-python/psycopg-2*]
-  ubuntu:
-    '*': [python-psycopg2]
-    focal: [python3-psycopg2]
-    jammy: [python3-psycopg2]
+  ubuntu: [python-psycopg2]
+python3-psycopg2:
+  debian: [python3-psycopg2]
+  fedora: [python3-psycopg2]
+  ubuntu: [python3-psycopg2]
 python-pulsectl-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2913,10 +2913,6 @@ python-psycopg2:
   fedora: [python-psycopg2]
   gentoo: [=dev-python/psycopg-2*]
   ubuntu: [python-psycopg2]
-python3-psycopg2:
-  debian: [python3-psycopg2]
-  fedora: [python3-psycopg2]
-  ubuntu: [python3-psycopg2]
 python-pulsectl-pip:
   debian:
     pip:
@@ -7617,6 +7613,10 @@ python3-psutil:
   rhel: ['python%{python3_pkgversion}-psutil']
   slackware: [psutil]
   ubuntu: [python3-psutil]
+python3-psycopg2:
+  debian: [python3-psycopg2]
+  fedora: [python3-psycopg2]
+  ubuntu: [python3-psycopg2]
 python3-py3exiv2-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2912,7 +2912,6 @@ python-psycopg2:
   debian: [python-psycopg2]
   fedora: [python-psycopg2]
   gentoo: [=dev-python/psycopg-2*]
-  ubuntu: [python-psycopg2]
 python-pulsectl-pip:
   debian:
     pip:
@@ -7617,6 +7616,8 @@ python3-psycopg2:
   debian: [python3-psycopg2]
   fedora: [python3-psycopg2]
   ubuntu: [python3-psycopg2]
+  opensuse: [python3-psycopg2]
+  rhel: [python3-psycopg2]
 python3-py3exiv2-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2915,6 +2915,7 @@ python-psycopg2:
   ubuntu:
     '*': [python-psycopg2]
     focal: [python3-psycopg2]
+    jammy: [python3-psycopg2]
```
clalancette, line 2918: Actually, it turns out that the `focal` entry is wrong.  All of our rosdep keys for python3 should start with `python3-`; the `python-` ones are implicitly Python 2.

So my suggestion here is to fix this key to remove the `focal` entry, and then add a new `python3-psycopg2` key with the appropriate entries:

```suggestion
  ubuntu: [python-psycopg2]
```

clalancette (changes requested):

rosdep/python.yaml
l-schilling: Seems to be the correct way according to the docs. I was not sure if python3 and python2 are still separated like this, because all distros using python2 are EOL.

rosdep/python.yaml
```diff
@@ -2912,9 +2912,11 @@ python-psycopg2:
   debian: [python-psycopg2]
   fedora: [python-psycopg2]
   gentoo: [=dev-python/psycopg-2*]
-  ubuntu:
-    '*': [python-psycopg2]
-    focal: [python3-psycopg2]
+  ubuntu: [python-psycopg2]
+python3-psycopg2:
```
clalancette, line 2916: This is the right key, but you are going to have to move it down with the other `python3-` keys in order for it to pass the alphabetic linters.

clalancette (changes requested):

rosdep/python.yaml
```diff
@@ -2912,9 +2912,7 @@ python-psycopg2:
   debian: [python-psycopg2]
   fedora: [python-psycopg2]
   gentoo: [=dev-python/psycopg-2*]
-  ubuntu:
-    '*': [python-psycopg2]
-    focal: [python3-psycopg2]
+  ubuntu: [python-psycopg2]
```
clalancette, line 2915: CI is failing because it can't find this package in Ubuntu, which I should have guessed.  In that case, I'm going to suggest we just remove the key completely since it isn't available in any of our supported distributions:

```suggestion
```

clalancette (changes requested):
Besides the fix for the Ubuntu key, there are also two suggestions to add the new `python3-psycopg2` key for RHEL and OpenSUSE.  Once all of that is fixed, I'll run CI again.

rosdep/python.yaml
```diff
@@ -7615,6 +7612,12 @@ python3-psutil:
   rhel: ['python%{python3_pkgversion}-psutil']
   slackware: [psutil]
   ubuntu: [python3-psutil]
+python3-psycopg2:
+  debian: [python3-psycopg2]
+  fedora: [python3-psycopg2]
+  ubuntu: [python3-psycopg2]
+  opensuse: [python3-psycopg2]
+  rhel: [python3-psycopg2]
```
clalancette, line 7620: These also need to be in alphabetical order.

clalancette (changes requested):

rosdep/python.yaml
```diff
@@ -7615,6 +7612,12 @@ python3-psutil:
   rhel: ['python%{python3_pkgversion}-psutil']
   slackware: [psutil]
   ubuntu: [python3-psutil]
+python3-psycopg2:
+  debian: [python3-psycopg2]
+  fedora: [python3-psycopg2]
```
clalancette, line 7617: Remove trailing whitespace:

```suggestion
  fedora: [python3-psycopg2]
```

clalancette (changes requested):

clalancette (approved):
Thanks for all of the iterations!  This looks good to me now.

rosdep/python.yaml
```diff
@@ -2912,9 +2912,6 @@ python-psycopg2:
   debian: [python-psycopg2]
   fedora: [python-psycopg2]
   gentoo: [=dev-python/psycopg-2*]
-  ubuntu:
-    '*': [python-psycopg2]
-    focal: [python3-psycopg2]
```
l-schilling, line 2917: I'm wondering if this is really right?
`ubuntu: [python-psycopg2]`
Would not work for foxy and jammy, but maybe this would be needed for some legacy distros?

clalancette: > Would not work for foxy and jammy, but maybe this would be needed for some legacy distros?

We don't typically worry about distributions that are end-of-life (though we also don't actively remove those keys either).  So in this case, I think it is OK.

methylDragon (approved):

PR #39402, opened 2024-01-02
Skip end-of-life distributions when checking rosdep duplicates.

clalancette:
We want to allow new rosdep keys that may overlap with things that may have been released in end-of-life distributions.  To do that, make sure we ignore end-of-life distributions in our rosdep duplicate checking.


methylDragon (approved):

methylDragon:
I'll wait for one more +1 just in case I missed something

adityapande-1995 (approved):

PR #39408, opened 2024-01-02
Fix ouster source and doc links.

clalancette:
Make sure they point to repos that exist, and the correct branch.

@Samahu FYI.


methylDragon (approved):

Samahu:
@clalancette apologies. I don't recall why `master` got selected.

PR #39409, opened 2024-01-02
Update locator_ros_bridge branches.

clalancette:
@syyuen FYI

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2601,7 +2601,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/boschglobal/locator_ros_bridge.git
-      version: main
+      version: humble
     release:
       packages:
       - bosch_locator_bridge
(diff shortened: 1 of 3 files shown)


rolling/distribution.yaml
```diff
@@ -2601,7 +2601,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/boschglobal/locator_ros_bridge.git
-      version: main
```
methylDragon, line 2604: Would these be better as https://github.com/boschglobal/locator_ros_bridge/tree/main-v1.4 ?

methylDragon (approved):
Just a question, otherwise looks okay.

rolling/distribution.yaml
clalancette: Yeah, good question.  I honestly don't know.  @syyuen will have to answer.

clalancette:
I'm going to go ahead and merge this; if this is the wrong branch, we can update it again later.

PR #39411, opened 2024-01-02
Switch irobot_create_msgs source branch to humble.

clalancette:
@alsora FYI


alsora (approved):
Thanks for catching this!

PR #39418, opened 2024-01-03
Added linux-isolate-process rosdep key

adityapande-1995:

## Package name:

linux-isolate-process

## Package Upstream Source:

Link to source : https://github.com/adityapande-1995/linux_isolate_process

## Purpose of using this:

This tool is useful to isolate and contain ROS nodes, so that they cannot communicate or discover other ROS nodes, kind of like ROS DOMAIN IDs. This approach is more scalable than domain ids, and isn't bound to ports. Instead, this uses the ``unshare()`` system call, and is linux specific. Primary use case would be isolate tests, so that a large number of tests can be run in parallel without them interfering with each other. Working on a colcon extension to use this tool, so that tests can be isolated.

Distro packaging links:

- Pip:
  - https://pypi.org/project/linux-isolate-process
  - Use the pip link above
  - Use the pip link above

Note : Ideally this should work for any linux distribution, but I haven't tested this, hence just including ubuntu and debian for now.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6716,6 +6716,13 @@ python3-lingua-franca-pip:
   ubuntu:
     pip:
       packages: [lingua-franca]
+python3-linux-isolate-process-pip:
+  debian:
+    pip:
+      packages: [linux-isolate-process]
+  ubuntu:
+    pip:
+      packages: [linux-isolate-process]
 python3-lockfile:
   alpine: [py3-lockfile]
   arch: [python-lockfile]


adityapande-1995:
cc @cottsay @nuclearsandwich

wjwwood:
nitpick: if it's `linux_isolate_process` in the `setup.py` and on the command-line, then it really shouldn't be `linux-isolate-process` in the repo or on pypi... It's an unnecessary inconsistency. It will need `_` replaced with `-` in Debian, but that's a well understood Debian requirement that affects lots of packages, plus you don't have a Debian yet.

methylDragon (changes requested):
See William's comment

adityapande-1995:
Updated the name to include underscores in the repo. For Pypi, I'm not sure if underscores are allowed. Are there any advantages to this being a ros package rather than a pip package with a rosdep key ? cc @cottsay

methylDragon (dismissed):
Ok pending the rest

Edit: Were the changes made to the repo? I don't see it :o

adityapande-1995:
As per discussion in the ROS2 meeting earlier today, closing this out in favor of a pure python ROS package.

PR #39475, opened 2024-01-09
mocap: 0.0.2-1 in 'humble/distribution.yaml' [bloom]

fmrico:
Increasing version of package(s) in repository `mocap` to `0.0.2-1`:

- upstream repository: https://github.com/MOCAP4ROS2-Project/mocap.git
- release repository: https://github.com/MOCAP4ROS2-Project/mocap-release.git
- bloom version: `0.11.2`

## mocap_control

```
* Added node_options parameter to constructor with default value, to be used with mocap4ros2 vicon driver
* Initial commit
* Contributors: Francisco Martín Rico, MFernandezCarmona
```

## mocap_control_msgs

```
* Initial commit
* Contributors: Francisco Martín Rico
```

## mocap_marker_publisher

```
* Update to msgs changes
* Initial commit
* Contributors: Francisco Martín Rico
```

## mocap_marker_viz

```
* Add launch rviz
* Multiple rigid bodies and launcher
* Adding rigid body and setting markers axis
* Fix header
* Initial commit
* Contributors: Francisco Martín Rico, Juancams, jmguerreroh
```

## mocap_marker_viz_srvs

```
* Initial commit
* Contributors: Francisco Martín Rico
```

## mocap_robot_gt

```
* Multiple rigid bodies and launcher
* Param to mocap frame
* Package for GT services
* Contributors: Francisco Martín Rico, jmguerreroh
```

## mocap_robot_gt_msgs

```
* Package for GT services
* Contributors: Francisco Martín Rico
```

## rqt_mocap_control

```
* Initial commit
* Contributors: Francisco Martín Rico, jmguerreroh
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3707,6 +3707,30 @@ repositories:
       url: https://github.com/DFKI-NI/mir_robot.git
       version: humble
     status: developed
+  mocap:
+    doc:
+      type: git
+      url: https://github.com/MOCAP4ROS2-Project/mocap.git
+      version: humble-devel
+    release:
+      packages:
+      - mocap_control
+      - mocap_control_msgs
+      - mocap_marker_publisher
+      - mocap_marker_viz
+      - mocap_marker_viz_srvs
+      - mocap_robot_gt
+      - mocap_robot_gt_msgs
+      - rqt_mocap_control
+      tags:
+        release: release/humble/{package}/{version}
+      url: https://github.com/MOCAP4ROS2-Project/mocap-release.git
+      version: 0.0.2-1
+    source:
+      type: git
+      url: https://github.com/MOCAP4ROS2-Project/mocap.git
+      version: humble-devel
+    status: developed
   mod:
     release:
       tags:


audrow (approved):
Looks good to me once merge conflicts are resolved.

mjcarroll:
Just a note that if you eventually want to release this in rolling, that we will need a ros2-gbp release repo.

fmrico:
Hi @mjcarroll 

Yes, actually these releases are related to this https://github.com/ros2-gbp/ros2-gbp-github-org/issues/385#issuecomment-1879200603

Thanks!!

tfoote:
As this is ambiguous enough we escalted this [REP 144](https://ros.org/reps/rep-0144.html) naming review to the ROS 2 meeting. And on review we've determined that these package names are too general. In particular `mocap` is a standard shortening of a term for a general class of technologies. We see that this is looking to be a generalizable solution for motion capture. But with mocap being so broad there's likely one or more approaches possible. 

This is already evidenced by there existing very similar names such as [mocap_optitrack](https://index.ros.org/p/mocap_optitrack/) And this project looks like it's going to progress next to releasing `mocap_optitrack_driver` https://github.com/MOCAP4ROS2-Project/mocap4ros2_optitrack/blob/main/mocap_optitrack_driver/package.xml  Which is confusing as it will have the same name as the already released package but is not expected to be compatible. 

This approach is a whole system that works together and what it really needs is a name in the sense of branding to be clear about it's compatibility. Looking at https://github.com/MOCAP4ROS2-Project you can see the effective use of naming as `mocap4ros2` is very clear. I would suggest it except that it goes against the policy of using ROS in the name of products and in package names.

The important part of the naming is that it clearly defines the scope. This is where `mocap` is too broad. Everything in that whole domain would potentially fall into that package scope and as there are definitely other people doing other things in the ecosystem that are not compatible and in this scope this clearly doesn't encompass everything possible in mocap. So to that end a more specific name or descriptor should be used. A potential approach would be to capture a philosophy or approach that this grouping is using.

tfoote (changes requested):
Officially reviewing, please see for the requested changes https://github.com/ros/rosdistro/pull/39475#issuecomment-1894443133

tfoote:
Yeah, that would work well in my opinion. It's unique, but moderately understandable without context. It makes me think of v4l as a good analogy which contrasts to uvc cameras.

fmrico:
Closing for preparing a new PR with different package names

PR #39536, opened 2024-01-12
generate_parameter_library: remove example packages

tylerjw:
Dropping example packages from the distribution. There was no good reason for these to be packaged. I tried putting ignored files in the release repo and running a release but that didn't seem to work as I expected.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2224,11 +2224,8 @@ repositories:
       version: main
     release:
       packages:
-      - cmake_generate_parameter_module_example
       - generate_parameter_library
-      - generate_parameter_library_example
       - generate_parameter_library_py
-      - generate_parameter_module_example
       - parameter_traits
       tags:
         release: release/humble/{package}/{version}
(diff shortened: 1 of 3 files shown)


tylerjw:
So, this didn't work as I hoped. I want to remove the example packages from the distributions. There is no reason to package them. I tried pushing ignored files to the release repo and then re-running bloom: https://github.com/ros2-gbp/generate_parameter_library-release/blob/main/rolling.ignored

That did not seem to work, as is evident by the content of this PR. Is there another step I should take?

tfoote:
I think that you're running into an assumption in bloom that the default branch is 'master' versus the more modern policy of 'main' 

I found https://robotics.stackexchange.com/q/97981/161  and from there 

https://github.com/ros-infrastructure/bloom/blob/d3d21ae9e816500154798d254422ac30e6bbee0f/bloom/config.py#L62

Where I think that bloom isn't detecting the default branch and is simply checking for a branch named `master` so it works for all the older repositories where `master` is the default, but won't work for newer ones where `main` is the default. 

Just removing the entries here will do what you want to remove it from the buildfarm. 

I made a new issue here: https://github.com/ros-infrastructure/bloom/issues/721

tylerjw:
That makes a ton of sense. Thank you for explaining that. I had just found old examples and copied them without noticing the branch.

PR #39542, opened 2024-01-14
Add mola_test_datasets for indexing

jlblancoc:

humble, iron, rolling

# The source is here:

https://github.com/MOLAorg/mola_test_datasets

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3787,6 +3787,16 @@ repositories:
       url: https://github.com/MOLAorg/mola_common.git
       version: develop
     status: developed
+  mola_test_datasets:
+    doc:
+      type: git
+      url: https://github.com/MOLAorg/mola_test_datasets.git
+      version: develop
+    source:
+      type: git
+      url: https://github.com/MOLAorg/mola_test_datasets.git
+      version: develop
+    status: developed
   motion_capture_tracking:
     doc:
       type: git
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -3104,6 +3104,16 @@ repositories:
       url: https://github.com/MOLAorg/mola_common.git
       version: develop
     status: developed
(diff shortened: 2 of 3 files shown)


Yadunund (approved):

Yadunund:
Will defer to @tfoote and @adityapande-1995 for final approval since it's a new package.

tfoote:
The name looks ok with the prefix. However I notice the readme says "Each dataset file holds the license of their original authors. Refer to each subdirectory." But the package.xml only says BSD. And I didn't find license files in the subdirectories.

jlblancoc:
@tfoote Please, take a [look at the repo now](https://github.com/MOLAorg/mola_test_datasets) and let me know if it's clearer via the `*/README.md` files, or if I should reflect this somehow in the package.xml too. Thanks.

tfoote:
There should be multiple license lines in the package.xml to provide a complete picture of the licenses included in the package so that if someone is auditing packages on their system the package.xml will give them the quick summary. 

And the Readme is good for readability and discovery. I'd recommend a link to the upstream projects. 

And please place a LICENSE file in each of the directories copied from the upstream repository as the short names are only categories and the actual license text is what is enforceable. 

You can reference those files from each license tag in the xml.

jlblancoc:
@tfoote Ok, it makes sense. It's now done upstream via [this commit](https://github.com/MOLAorg/mola_test_datasets/commit/504b249c566b134ecea28d22e778bf285f75cd20).

tfoote (approved):
Thanks for the extra followups. I know that it can feel like a pain, but these things are important to make sure we keep clear.

jlblancoc:
I know, no problem! I'm used to similar issues with `debian/copyright` in other packages :-) Thanks for the quick follow up.

tfoote:
Yeah, we actually generate those for the debian packages from the package.xml

PR #39621, opened 2024-01-22
multidim_rrt_planner-release: 0.0.6-1 in 'iron/distribution.yaml' [bloom]

david-dorf:
Increasing version of package(s) in repository `multidim_rrt_planner-release` to `0.0.6-1`:

- upstream repository: https://github.com/ros2-gbp/multidim_rrt_planner-release.git
- release repository: https://github.com/ros2-gbp/multidim_rrt_planner-release.git
- distro file: `iron/distribution.yaml`
- bloom version: `0.11.2`

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -3392,24 +3392,16 @@ repositories:
       version: iron
     release:
       packages:
-      - rrt_planner
-      tags:
-        release: release/iron/{package}/{version}
-      url: https://github.com/ros2-gbp/multidim_rrt_planner-release.git
-      version: 0.0.4-1
-    source:
-      type: git
-      url: https://github.com/daviddorf2023/multidim_rrt_planner.git
-      version: iron
-    status: maintained
-  multidim_rrt_planner-release:
-    release:
-      packages:
       - multidim_rrt_planner
       tags:
         release: release/iron/{package}/{version}
       url: https://github.com/ros2-gbp/multidim_rrt_planner-release.git
       version: 0.0.6-1
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


iron/distribution.yaml
```diff
@@ -3402,6 +3402,14 @@ repositories:
       url: https://github.com/daviddorf2023/multidim_rrt_planner.git
       version: iron
     status: maintained
+  multidim_rrt_planner-release:
```
mjcarroll, line 3405: I don't believe that you should have a `-release` suffix here, it should replace the section above?

mjcarroll (changes requested):

iron/distribution.yaml
mjcarroll: You should be able to hand-modify the yaml file to what you expect it to look like and update this PR.  I believe next time you run bloom (for your next release), you may need to edit the tracks, as well.

daviddorf2023: Thank you for your help! Didn't realize bloom was opening a new branch on my rosdistro fork. I'm getting a better understanding of the tool now. Should be fixed now.

clalancette:
So this one is a bit difficult to track down, but CI is failing because the name of the package exactly matches the name of the top-level "key" that is being used.  In that specific case, there should be no `-packages` section at all, and the name of the package is inferred from the name of the key.  So I think if you just remove the `-packages` section here, it will be happier.

mjcarroll (approved):

mjcarroll:
I think we are good here, @clalancette want one more look?

clalancette (approved):

PR #39702, opened 2024-01-24
Add google-cloud-secret-manager

safijari:

## Package Upstream Source: google-cloud-secret-manager from pip

https://pypi.org/project/google-cloud-secret-manager/

## Purpose of using this:

Google cloud secret manager is useful for managing secrets when using google cloud :eyes:. Looks like all the other google-cloud packages are present except this one.

Distro packaging links:

Don't think there are any? Like the other google-cloud packages this comes from pip.

  - https://pypi.org/project/google-cloud-secret-manager/
- Ubuntu: https://packages.ubuntu.com/j
  - https://pypi.org/project/google-cloud-secret-manager/
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

python-google-cloud-secret-manager-pip

# The source is here:

https://pypi.org/project/google-cloud-secret-manager/

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1811,7 +1811,7 @@ python-google-cloud-vision-pip:
   ubuntu:
     pip:
       packages: [google-cloud-vision]
-python-google-cloud-secret-manager:
+python3-google-cloud-secret-manager-pip:
   debian:
     pip:
       packages: [google-cloud-secret-manager]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1811,16 +1811,6 @@ python-google-cloud-vision-pip:
   ubuntu:
     pip:
       packages: [google-cloud-vision]
-python3-google-cloud-secret-manager-pip:
-  debian:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -1811,6 +1811,16 @@ python-google-cloud-vision-pip:
   ubuntu:
     pip:
       packages: [google-cloud-vision]
+python-google-cloud-secret-manager:
```
clalancette, line 1814: Because this is a pip-only key, this should be:

```suggestion
python3-google-cloud-secret-manager-pip:
```

(also you'll have to move it in the correct place so it is alphabetically sorted)

clalancette (changes requested):

clalancette (approved):

wjwwood (approved):

PR #39719, opened 2024-01-25
added apriltag_detector

berndpfrommer:
# Please add apriltag_detector to be indexed in the rosdistro.
ROS distros: humble, iron, rolling

Note: there is already a [apriltag_ros](https://github.com/christianrauch/apriltag_ros) package, so why add another one?
The existing apriltag_ros package not only detects apriltags, but also uses them to compute the pose of the tags (transforms). For that reason it subscribes to calibration (camera_info) messages which are *not available* until the camera has been calibrated. However, this makes the existing  package difficult to use for camera calibration.
The new package proposed here focuses only on the elementary task of perception: detect the tag's ID and corners in the image. Note that it uses the same message format as the apriltag_ros package, except that it does not utilize the homography field.

NAME of package: apriltag_detector

# The source is here:
https://github.com/ros-misc-utilities/apriltag_detector

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -358,6 +358,16 @@ repositories:
       url: https://github.com/AprilRobotics/apriltag.git
       version: master
     status: maintained
+  apriltag_detector:
+    doc:
+      type: git
+      url: https://github.com/ros-misc-utilities/apriltag_detector.git
+      version: master
+    source:
+      type: git
+      url: https://github.com/ros-misc-utilities/apriltag_detector.git
+      version: master
+    status: developed
   apriltag_msgs:
     release:
       tags:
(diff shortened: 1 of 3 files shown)


wjwwood (approved):
Sounds ok to me, waiting on another review to confirm the naming is ok with others too.

mjcarroll (approved):
Makes sense to me.

rolling/distribution.yaml
```diff
@@ -358,6 +358,16 @@ repositories:
       url: https://github.com/AprilRobotics/apriltag.git
       version: master
     status: maintained
+  apriltag_detector:
```
mjcarroll, line 361: Note that your next step here will be to get an entry in the ros2-gbp org: https://github.com/ros2-gbp/ros2-gbp-github-org

clalancette (approved):

PR #39799, opened 2024-02-06
Drop debian bullseye from Iron distribution.

nuclearsandwich:
Bullseye is no longer the current Debian Stable release and so we will stop generating release metadata for it after the next Iron sync.
This PR should be merged just after tagging this repository for the [next Iron sync](https://discourse.ros.org/t/preparing-for-iron-patch-and-sync-2024-02-09/35920).


Yadunund (approved):
The sync was tagged [iron/2024-02-09](https://github.com/ros/rosdistro/tree/iron/2024-02-09).

PR #39798, opened 2024-02-06
Drop old stable debian release from Humble.

nuclearsandwich:
This will stop bloom from attempting to generate Debian Bullseye release metadata when running releases for these distributions.

This will not prevent past releases from being available in the ros2-gbp future releases will not include metadata for these platforms.

I am opening one PR for the initial review but I think it would be best to merge this change distribution-by-distribution after each one has completed a sync so that there is a tag which represents the last set of of releases where Debian metadata was attempted. I'll use this PR to coordinate those changes.

Last sync with Debian metadata:
- [ ] Humble
- [x] Iron https://github.com/ros/rosdistro/pull/39799
- [x] Rolling https://github.com/ros/rosdistro/pull/39802

PR #39802, opened 2024-02-07
Drop debian bullseye from Rolling distribution.

nuclearsandwich:
Bullseye is no longer the current Debian Stable release and so we will stop generating release metadata for it after the next Rolling sync.


nuclearsandwich:
This isn't technically a yolomerge because this change is approved over here https://github.com/ros/rosdistro/pull/39798

PR #39858, opened 2024-02-08
Add sqlite-utils pip source

ghost:

## Package name:

python3-sqlite-utils-pip

## Package Upstream Source:

Package from pypi
[link to source repository](https://github.com/simonw/sqlite-utils)

## Purpose of using this:

This dependency is being used for handling sql databases using sqlite. I think it is interesting as a utils package to manage sql databases since it eases the usage of sqlite based databases given an abstraction in python.

Distro packaging links:

- Python: https://pypi.org
  - [REQUIRED](https://pypi.org/project/sqlite-utils/)

# Checks
 - [] This package is expected to build on the submitted rosdistro


cottsay (approved):
Please bear in mind that pip-only dependencies like this cannot be used in packages which are released to be built on build.ros2.org.

sloretz (approved):

PR #39868, opened 2024-02-11
Add rosdep key for playsound alpine

RomainL972:

## Package name:

python3-playsound-pip

## Package Upstream Source:

https://pypi.org/project/playsound/

## Purpose of using this:

The package is already in the database but I want to use it with Alpine.

N/A


cottsay (changes requested):
The keys need to be alphabetized (`alpine` should be listed first under `python3-playsound-pip`)

cottsay (approved):

clalancette (approved):

PR #39888, opened 2024-02-13
marine_msgs: 2.0.1-2 in 'noetic/distribution.yaml' [bloom]

rolker:
Increasing version of package(s) in repository `marine_msgs` to `2.0.1-2`:

- upstream repository: https://github.com/CCOMJHC/marine_msgs
- release repository: https://github.com/CCOMjhc/marine_msgs-release.git
- distro file: `noetic/distribution.yaml`
- bloom version: `0.11.2`
- previous version for package: `null`

## marine_acoustic_msgs

- No changes

## marine_sensor_msgs

```
* Add radar message and migration rules. (#40 <https://github.com/rolker/marine_msgs/issues/40>)
  Add radar messages and migration rules.
* Contributors: Roland Arsenault
```

PR #39899, opened 2024-02-14
marine_msgs: 2.0.1-1 in 'noetic/distribution.yaml' [bloom]

rolker:
Increasing version of package(s) in repository `marine_msgs` to `2.0.1-1`:

- upstream repository: https://github.com/apl-ocean-engineering/marine_msgs.git
- release repository: https://github.com/CCOMJHC/marine_msgs-release.git
- bloom version: `0.11.2`

## marine_acoustic_msgs

- No changes

## marine_sensor_msgs

```
* Add radar message and migration rules. (#40 <https://github.com/rolker/marine_msgs/issues/40>)
  Add radar messages and migration rules.
* Contributors: Roland Arsenault
```

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5424,6 +5424,23 @@ repositories:
       url: https://github.com/swri-robotics/mapviz.git
       version: master
     status: developed
+  marine_msgs:
+    doc:
+      type: git
+      url: https://github.com/apl-ocean-engineering/marine_msgs.git
+      version: 2.0.1
+    release:
+      packages:
+      - marine_acoustic_msgs
+      - marine_sensor_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/CCOMJHC/marine_msgs-release.git
+      version: 2.0.1-1
+    source:
+      type: git
+      url: https://github.com/apl-ocean-engineering/marine_msgs.git
+      version: main
   marker_msgs:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5428,7 +5428,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/apl-ocean-engineering/marine_msgs.git
-      version: 2.0.1
+      version: main
     release:
       packages:
       - marine_acoustic_msgs


rolker:
This is a resubmission of #39888

clalancette:
New package review:

  - [x] Top level license file:
- [x] License is [OSI-approved](https://opensource.org/licenses):
- [x] Public source repo: https://github.com/CCOMJHC/marine_msgs

noetic/distribution.yaml
```diff
@@ -5424,6 +5424,23 @@ repositories:
       url: https://github.com/swri-robotics/mapviz.git
       version: master
     status: developed
+  marine_msgs:
+    doc:
+      type: git
+      url: https://github.com/apl-ocean-engineering/marine_msgs.git
+      version: 2.0.1
```
clalancette, line 5431: The branch name here should typically be the same as the one in the `source` entry.  If you really don't want them to be the same they don't have to be, but that is recommended.

clalancette (changes requested):

noetic/distribution.yaml
rolker: Thanks, change commited.

noetic/distribution.yaml
```diff
@@ -5424,6 +5424,23 @@ repositories:
...
+      type: git
+      url: https://github.com/apl-ocean-engineering/marine_msgs.git
+      version: main
+    release:
+      packages:
+      - marine_acoustic_msgs
+      - marine_sensor_msgs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/CCOMJHC/marine_msgs-release.git
```
lauralindzey, line 5438: I'm not familiar with the release process, but seems weird to have both a CCOMJHC and an apl-ocean-engineering repo listed here. Was this line an oversight?

lauralindzey, line 5438: I'm definitely not familiar with the release process ... is there a reason to use both the apl-ocean-engineering repo and the CCOMJHC version in this file?

clalancette: So the way that this works is that there has to be two repositories: one is the source repository, and where the work is generally done.  The second repository is what we call the `release` repository, and is where bloom pushes all kinds of metadata related to building packages.

Those two repositories can reside anywhere you can host a git repository, though we typically prefer github.com for some of the other automation that we can do there.  As far as the GitHub organization is concerned, the two repositories can be in separate organizations, as it is here.  For Noetic, at least, we don't have a preference, so this should be sorted out between @lauralindzey and @rolker .

lauralindzey: Thanks for the explanation! I have no objection to the release repository living under the CCOMJHC org.

ros-discourse:

https://discourse.ros.org/t/new-ros-enhancement-proposal-for-marine-robotics/36218/6

clalancette:
We took a quick look at this in the ROS 2 weekly.  Our main concern is the name of the package, since it is a pretty generic name.

However, it looks like this has at least a few different parties contributing to the messages, so from that perspective it seems like this is a community effort.  With that in mind, we are going to accept this PR, but please keep in mind that the `marine_msgs` name is a "shared" resource, and thus there may be others who want to contribute to them.

Thanks for your effort here!

clalancette (approved):

rolker:
Yes, this is definitly meant to be a shared name! Thanks for helping us with this first release.

ros-discourse:

https://discourse.ros.org/t/new-packages-for-noetic-2024-03-07/36514/2

PR #39905, opened 2024-02-15
Add `overrides` to python rosdeps

amalnanavati:

## Package name:

`overrides`

## Package Upstream Source:

https://github.com/mkorpela/overrides

## Purpose of using this:

This dependency makes it easier and cleaner to manage docstrings when using class inheritance in Python3.

- pip: https://pypi.org
    - https://pypi.org/project/overrides
  - Use `pip` package linked above. Not available in the Debian repositories.
  - Use `pip` package linked above. Not available in the Ubuntu repositories.
  - Use `pip` package linked above. Not available in the Fedora repositories.
  -https://archlinux.org/packages/extra/any/python-overrides/
  - Not available
  - Use `pip` package linked above. Not available in the Homebrew repositories.
  - Not available
  - Not available
  - Not available
  - Not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7327,7 +7327,7 @@ python3-openhsi-pip:
   ubuntu:
     pip:
       packages: [openhsi]
-python3-overrides-pip:
+python3-overrides:
   arch: [python-overrides]
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7317,6 +7317,16 @@ python3-opengl:
   nixos: [python3Packages.pyopengl]
   opensuse: [python3-opengl]
   ubuntu: [python3-opengl]
+python3-openhsi-pip:
+  debian:
+    pip:
+      packages: [openhsi]
+  fedora:
+    pip:
+      packages: [openhsi]
+  ubuntu:
+    pip:
+      packages: [openhsi]
 python3-overrides:
   arch: [python-overrides]
   debian:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -7327,6 +7327,20 @@ python3-openhsi-pip:
   ubuntu:
     pip:
       packages: [openhsi]
+python3-overrides:
```
clalancette, line 7330: Since this is a pip-only key, this should have a `-pip` suffix:

```suggestion
python3-overrides-pip:
```

clalancette (changes requested):

amalnanavati:
Done. The reason I originally kept it with a non-pip ID is because the arch package is non-pip. However, given that that is a less commonly used OS, I understand the logic in keeping `pip` in the ID.

clalancette:
> Done. The reason I originally kept it with a non-pip ID is because the arch package is non-pip. However, given that that is a less commonly used OS, I understand the logic in keeping `pip` in the ID.

Ah, sorry, I totally missed that it was available as a package in `arch`.  Sorry about that.

At this point, we can either:
1.  Go back to the non-pip key, given that it does have at least one OS where it is packaged.
2.  Remove the `arch` entry, and leave it as a pip key.

I'm fine with either one, I'll leave it up to you.

amalnanavati:
Done, I reverted back to the non-pip key.

mjcarroll (approved):

clalancette (approved):

PR #39984, opened 2024-02-21
added spinnaker_synchronized_camera_driver package

berndpfrommer:
# Please add spinnaker_synchronized_camera_driver package to be indexed in the rosdistro.
This package is part of the flir_camera_driver family. It supports synchronized cameras by associating frames from multiple cameras with each other, and forcing identical time stamps for frames that have been triggered by the same sync pulse.

Note: I also updated which branches should be used for the documentation.

rosdistros: Humble and Iron

# The source is here:
https://github.com/ros-drivers/flir_camera_driver.git

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2003,12 +2003,13 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-drivers/flir_camera_driver.git
-      version: humble-devel
+      version: humble-release
     release:
       packages:
       - flir_camera_description
       - flir_camera_msgs
       - spinnaker_camera_driver
+      - spinnaker_synchronized_camera_driver
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/ros-drivers-gbp/flir_camera_driver-release.git
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -1553,12 +1553,13 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-drivers/flir_camera_driver.git
-      version: humble-devel
+      version: iron-release
     release:
       packages:
       - flir_camera_description
       - flir_camera_msgs
       - spinnaker_camera_driver
+      - spinnaker_synchronized_camera_driver
       tags:
         release: release/iron/{package}/{version}
       url: https://github.com/ros-drivers-gbp/flir_camera_driver-release.git


berndpfrommer:
I don't get why it's not working. I'm just adding a new package to an existing source repository, which already has a release repository. The package certainly has a package.xml file and builds and tests just fine locally (although the rosdistro test above fails as well when I run it locally). Why is this failing?

clalancette:
> I don't get why it's not working. I'm just adding a new package to an existing source repository, which already has a release repository. The package certainly has a package.xml file and builds and tests just fine locally (although the rosdistro test above fails as well when I run it locally). Why is this failing?

You can't "just" add a new package to a release; you need to to a source release, and then a `bloom-release` of that package.  That will generate the correct metadata in https://github.com/ros-drivers-gbp/flir_camera_driver-release, which is what is missing.

Given that a new bloom-release is going to have to be done anyway, I'm going to close this one out.  But feel free to keep commenting.

berndpfrommer:
I was under the impression that the first step before releasing a package is to seek approval for the naming of it, and that the process for that is submitting a PR against the rosdistro. Apparently that's not the right process.
I went ahead and bloom-released all packages in the repo, including the new spinnaker-synchronized-camera-driver. Let's see what happens.

berndpfrommer:
Sorry, I don't know how to make it work. The spinnaker_synchronized_camera_driver package does not show up in the distribution.yaml even after I did a bloom release. I usually just bloom-release like this:
```
bloom-release --rosdistro ${distro} ${my_repo}
```
but somehow that does not add the new package to the rosdistro either.

berndpfrommer:
@clalancette I guess I don't know what the "source release" is that you mentioned above. Is there any documentation on how to add a new package to an existing repository? I found plenty of documentation on how to add a new repo to rosdistro, but not on how to add a new package to an existing one.

clalancette:
> @clalancette I guess I don't know what the "source release" is that you mentioned above. Is there any documentation on how to add a new package to an existing repository? I found plenty of documentation on how to add a new repo to rosdistro, but not on how to add a new package to an existing one.

Adding a new package to an existing repository is pretty much the same as doing any release.  The only difference is that `bloom` will discover the new package and add it.  That is, you need to follow all of the instructions starting at http://docs.ros.org/en/rolling/How-To-Guides/Releasing/Subsequent-Releases.html .

berndpfrommer:
That's what I did, but the "track" was screwed up so somehow bloom picked up an old version of the repo which did not have the new package yet. After fixing the "track" the messages I get from bloom look much more encouraging, i.e. the new package is mentioned there.

PR #39987, opened 2024-02-21
Update pull_request_template.md to include package name review template

ijnek:
Updates issue template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,5 @@
-<!-- Thank you for contributing a change to the rosdistro. There are two primary types of submissions.
-Please select the appropriate template from below: ROSDEP_RULE_TEMPLATE or DOC_INDEX_TEMPLATE
+<!-- Thank you for contributing a change to the rosdistro. There are three primary types of submissions.
+Please select the appropriate template from below: ROSDEP_RULE_TEMPLATE, NAMING_REVIEW_TEMPLATE or DOC_INDEX_TEMPLATE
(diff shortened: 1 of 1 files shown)


ijnek:
@nuclearsandwich fyi

.github/pull_request_template.md
```diff
@@ -47,6 +47,21 @@ More info at https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidel
...
+## Package name
+
+package_name
+
+## Package source code
+
+http://sourcecode_url
+
+# Checks
+- [ ] Does the package naming comply with [REP-144](https://ros.org/reps/rep-0144.html) ?
```
audrow, line 64:
```suggestion
- [ ] Does the package naming comply with [REP-144](https://ros.org/reps/rep-0144.html)? 
```

audrow (commented):
Seems okay to me. I'd like to get some other eyes on it before merging.

nuclearsandwich (changes requested):
Thanks for making the suggestion here! I brought this up in today's ROS core meeting and we're concerned with adding an additional tree to the "choose your own adventure" template. Do you think you could instead make the REP-144 requirement part of the DOC INDEX review, which already serves a similar purpose of registering a source repository to be indexed in rosdistro?

ijnek:
Thanks for the review @audrow and @nuclearsandwich , i have addressed your feedback in #40518. Closing this PR since it is superceded.

PR #40024, opened 2024-02-28
Use ros2-gbp release repository for message_tf_frame_transformer.

nuclearsandwich:
This package was previously copied into the ros2-gbp repository during the Iron branching but not used for subsequent releases.

As a result, there is divergent bloom info between the ros2-gbp repository and the upstream one. But all tags, which are the important artifacts for release integrity, have been copied over to the ros2-gbp repository.

Future releases in all ROS 2 distributions should use the ros2-gbp repository which is configured in https://github.com/ros2-gbp/ros2-gbp-github-org/pull/458

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3674,7 +3674,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/ika-rwth-aachen/message_tf_frame_transformer-release.git
+      url: https://github.com/ros2-gbp/message_tf_frame_transformer-release.git
       version: 1.1.0-1
     source:
       type: git
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -3033,7 +3033,7 @@ repositories:
     release:
       tags:
         release: release/iron/{package}/{version}
-      url: https://github.com/ika-rwth-aachen/message_tf_frame_transformer-release.git
+      url: https://github.com/ros2-gbp/message_tf_frame_transformer-release.git
       version: 1.1.0-1
     source:
       type: git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2861,7 +2861,7 @@ repositories:
     release:
       tags:
         release: release/rolling/{package}/{version}
-      url: https://github.com/ika-rwth-aachen/message_tf_frame_transformer-release.git
+      url: https://github.com/ros2-gbp/message_tf_frame_transformer-release.git
       version: 1.1.0-1
     source:
       type: git


nuclearsandwich:
FYI @lreiher Per [REP-2002](https://www.ros.org/reps/rep-2002.html#managing-build-farm-writable-release-repositories) release repositories for ROS 2 are centralized in the [ros2-gbp](https://github.com/ros2-gbp) organization on GitHub. The release repository was previously mirrored into ros2-gbp for an earlier migration. I've updated the ros2-gbp repository with tags from the previous release repository so that it contains all release artifacts used for ROS 2 distributions.

Your access to the release repository in ros2-gbp will be available once you accept the invitation to join the organization which will be deploying shortly now that https://github.com/ros2-gbp/ros2-gbp-github-org/pull/458 is merged. Please open an issue on https://github.com/ros2-gbp/ros2-gbp-github-org if you have any questions or trouble.

marcoag (approved):

humble/distribution.yaml
```diff
@@ -3674,7 +3674,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/ika-rwth-aachen/message_tf_frame_transformer-release.git
+      url: https://github.com/ros2-gbp/message_tf_frame_transformer-release.git
       version: 1.1.0-1
```
clalancette, line 3678: Looking at https://github.com/ros2-gbp/message_tf_frame_transformer-release/blob/master/tracks.yaml, it is clear that a release of this package was done into the old release repository that is not reflected in the new one.  How do we want to handle this?

nuclearsandwich: What precisely do you mean by "not reflected"?

I specifically documented a discrepancy in the PR description:

> there is divergent bloom info between the ros2-gbp repository and the upstream one. But all tags, which are the important artifacts for release integrity, have been copied over to the ros2-gbp repository.

I'll double check whether the changes from the external release repository's bloom configuration need to be incorporated for the migration tool to work as intended before merging.

> Looking at https://github.com/ros2-gbp/message_tf_frame_transformer-release/blob/master/tracks.yaml, it is clear that a release of this package was done into the old release repository that is not reflected in the new one

Furthermore, it is _not_ clear to me what specifically you're trying to highlight as  a difference between the external tracks.yaml and the ros2-gbp one. A line-specific link or description of the discrepancy would be appreciated in the future.

nuclearsandwich: As previously stated, all tags were propagated so any past release artifacts are available in the ros2-gbp repository whether the release was originally performed using the external repository or the ros2-gbp one. I am fairly certain that the script would function correctly since it uses the rosdistro metadata as its primary source of truth, but that doesn't mean that future manual releases to the other ROS 2 distributions would behave as expected. So I went ahead and integrated the changes from the external repository's tracks.yaml in https://github.com/ros2-gbp/message_tf_frame_transformer-release/commit/3b498040041fe067ffa9cf9e496d50cfb385b108

PR #40026, opened 2024-02-28
Use ros2-gbp release repository for tuw_geometry.

nuclearsandwich:
This package was previously copied into the ros2-gbp repository during the Iron branching but not used for subsequent releases.

As a result, there is divergent bloom info between the ros2-gbp repository and the upstream one. But all tags, which are the important artifacts for release integrity, have been copied over to the ros2-gbp repository.

Future releases in all ROS 2 distributions should use the ros2-gbp repository which is configured in ros2-gbp/ros2-gbp-github-org#243

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -9021,7 +9021,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/tuw-robotics/tuw_geometry-release.git
+      url: https://github.com/ros2-gbp/tuw_geometry-release.git
       version: 0.1.1-1
     source:
       type: git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -7150,7 +7150,7 @@ repositories:
     release:
       tags:
         release: release/rolling/{package}/{version}
-      url: https://github.com/tuw-robotics/tuw_geometry-release.git
+      url: https://github.com/ros2-gbp/tuw_geometry-release.git
       version: 0.0.7-2
     source:
       type: git


nuclearsandwich:
FYI @maxbader Per [REP-2002](https://www.ros.org/reps/rep-2002.html#managing-build-farm-writable-release-repositories) release repositories for ROS 2 are centralized in the [ros2-gbp](https://github.com/ros2-gbp) organization on GitHub. The release repository was previously mirrored into ros2-gbp for an earlier migration. I've updated the ros2-gbp repository with tags from the previous release repository so that it contains all release artifacts used for ROS 2 distributions.

Your access to the release repository in ros2-gbp is already available. Please open an issue on https://github.com/ros2-gbp/ros2-gbp-github-org if you have any questions or trouble.

marcoag (approved):

humble/distribution.yaml
```diff
@@ -9021,7 +9021,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/tuw-robotics/tuw_geometry-release.git
+      url: https://github.com/ros2-gbp/tuw_geometry-release.git
       version: 0.1.1-1
```
clalancette, line 9025: Looking at https://github.com/ros2-gbp/tuw_geometry-release/blob/master/tracks.yaml , it is clear that a release has been done to the old https://github.com/tuw-robotics/tuw_geometry-release.git which is not reflected in https://github.com/ros2-gbp/tuw_geometry-release.git (the same goes for `rolling` below).  This means that the version listed here no longer reflects what is available in the `ros2-gbp` repository.  What should we do here?

nuclearsandwich: Thanks for flagging this. I'm confident that the script would use the appropriate version when updating Rolling because it uses the rosdistro index as the source-of-truth but that doesn't cover manual releases for tracks that aren't getting migrated. So I went ahead and integrated the changes from the upstream release repository in https://github.com/ros2-gbp/tuw_geometry-release/commit/2497e69ab48833ca35c79e2f949f461886f51edb

PR #40030, opened 2024-02-29
Use ros2-gbp repository for magic_enum release.

nuclearsandwich:
The repository has been mirrored into ros2-gbp here: ros2-gbp/ros2-gbp-github-org#456.

I did not change the release repository for noetic at this time since it is not required for ROS 1 distributions, but if the maintainer prefers to use a single release repository for all distributions that is perfectly acceptable.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3483,7 +3483,7 @@ repositories:
     release:
       tags:
         release: release/humble/{package}/{version}
-      url: https://github.com/nobleo/magic_enum-release.git
+      url: https://github.com/ros2-gbp/magic_enum-release.git
       version: 0.9.5-1
     source:
       type: git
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
FYI @Timple

Timple:
I'm good. 
Is there a reason this is mirrored and not forked? Now the default branch is kind of random :slightly_smiling_face:

marcoag (approved):

clalancette:
> Is there a reason this is mirrored and not forked? Now the default branch is kind of random 🙂

I've fixed it now; the default branch is now `master`.  We don't fork these because these are supposed to be the authoritative source of truth for the buildfarm.

clalancette (approved):

Timple:
Learning every day :slightly_smiling_face:

PR #40091, opened 2024-03-06
Make the ros1_bridge and rosbag2_bag_v2 source-only in Rolling

clalancette:
Both of these require ROS 1 packages, which are not available in Ubuntu Noble, so we cannot build binary packages for them.  They are both still available to build from source.

@marcoag @nuclearsandwich FYI

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -5184,20 +5184,13 @@ repositories:
       url: https://github.com/robotont/robotont_msgs.git
       version: ros2-rolling-devel
   ros1_bridge:
-    doc:
-      type: git
-      url: https://github.com/ros2/ros1_bridge.git
-      version: master
-    release:
-      tags:
-        release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/ros1_bridge-release.git
     source:
       test_commits: false
       type: git
       url: https://github.com/ros2/ros1_bridge.git
       version: master
-    status: maintained
+    status_description: Maintained in source form only since no ROS 1 packages available
+      in Rolling
   ros2_canopen:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

nuclearsandwich:
@clalancette an alternative would be to only remove the `doc` and `release` stanzas and add
```
test_pull_requests: false
test_commits: false
```

to the source stanza, but keep it in the distribution. That way it remains indexed but is not built on the build farm.

clalancette:
> to the source stanza, but keep it in the distribution. That way it remains indexed but is not built on the build farm.

Oh, I like it.  I'll update this PR.

rolling/distribution.yaml
```diff
@@ -5177,16 +5177,6 @@ repositories:
...
-    release:
-      tags:
-        release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/ros1_bridge-release.git
-    source:
-      test_commits: false
-      type: git
-      url: https://github.com/ros2/ros1_bridge.git
-      version: master
-    status: maintained
```
nuclearsandwich, line 5189: This is actually not quite what I intended. The doc stanza has been retained (but I'm not sure we can build documentation for ros1_bridge without ROS 1 nor does it make sense to me to provide docs for a package we don't make available and the source stanza has been removed.

What I think should remain is:
```yaml
  ros1_bridge:
    source:
      test_commits: false
      type: git
      url: https://github.com/ros2/ros1_bridge.git
      version: master
    status: <could we put something here about the deprecation due to no ROS 1 build?>
```

nuclearsandwich (changes requested):

rolling/distribution.yaml
clalancette: > ```yaml
> 
>     status: <could we put something here about the deprecation due to no ROS 1 build?>
> ```

Apparently we have both `status`, which has to be one of `developed`, `maintained`, `unmaintained`, and `end-of-life`, as well as `status_description`, which is a free-form field (see https://www.ros.org/reps/rep-0141.html#distribution-file).  So I'm going to add in a `status_description` here, saying something about why this is not released.

nuclearsandwich (approved):

PR #40101, opened 2024-03-06
Add in rosdep key for nlohmann-json-dev for noble.

clalancette:
Please update the following dependency in the rosdep database.

## Package name:

nlohmann-json-dev

## Package Upstream Source:

https://nlohmann.github.io/json/

## Purpose of using this:

Update nlohmann-json-dev to have a noble key. This is needed to release `foxglove_bridge` into Rolling on Noble.

  - https://packages.ubuntu.com/noble/nlohmann-json3-dev

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6917,6 +6917,7 @@ nlohmann-json-dev:
     bionic: [nlohmann-json-dev]
     focal: [nlohmann-json3-dev]
     jammy: [nlohmann-json3-dev]
+    noble: [nlohmann-json3-dev]
 nmap:
   archlinux: [nmap]
   debian: [nmap]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6904,9 +6904,7 @@ nkf:
   ubuntu: [nkf]
 nlohmann-json-dev:
   debian:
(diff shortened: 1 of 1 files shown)


cottsay (approved):

rosdep/base.yaml
```diff
@@ -6906,6 +6906,7 @@ nlohmann-json-dev:
     bionic: [nlohmann-json-dev]
     focal: [nlohmann-json3-dev]
     jammy: [nlohmann-json3-dev]
+    noble: [nlohmann-json3-dev]
```
nuclearsandwich, line 6909: Could we elide this rather than add to it with:
```diff
+     '*': [nlohmann-json3-dev]
      bionic: [nlohmann-json-dev]
-     focal: [nlohmann-json-dev]
-     jammy: [nlohmann-json-dev]

```

nuclearsandwich (approved):
We could avoid having to do this again in two years with a slightly different change but however the author prefers to proceed is fine by me.

rosdep/base.yaml
clalancette: Yeah, that's fair.  I notice that `debian` is in much the same boat, so I'll fix them both.

PR #40144, opened 2024-03-08
Add python3-pyzbar-pip

MatthijsBurgh:
Anchored to `python-pyzbar-pip` (https://github.com/ros/rosdistro/pull/23278)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3362,7 +3362,7 @@ python-pyxhook-pip:
   ubuntu:
     pip:
       packages: [pyxhook]
-python-pyzbar-pip:
+python-pyzbar-pip: &migrate_eol_2025_04_30_python3_pyzbar_pip
   debian:
     pip:
       packages: [pyzbar]
@@ -8311,6 +8311,7 @@ python3-pyxdg-pip:
   ubuntu:
     pip:
       packages: [pyxdg]
+python3-pyzbar-pip: *migrate_eol_2025_04_30_python3_pyzbar_pip
 python3-qpsolvers-pip:
   debian:
     pip:


rosdep/python.yaml
```diff
@@ -8311,6 +8311,7 @@ python3-pyxdg-pip:
   ubuntu:
     pip:
       packages: [pyxdg]
+python3-pyzbar-pip: *migrate_eol_2025_04_30_python3_pyzbar_pip
```
clalancette, line 8314: In my opinion, I think we should do this differently.  In particular, it looks like `python3-pyzbar` is now available in Ubuntu and Debian, since `jammy` and `bullseye`, respectively: https://pkgs.org/search/?q=python3-pyzbar.  So I think we should instead make this a full key:

```suggestion
python3-pyzbar:
  debian: [python3-pyzbar]
  ubuntu:
    '*': [python3-pyzbar]
    focal: null
```

clalancette (changes requested):

rosdep/python.yaml
MatthijsBurgh: I think we should add both.

nuclearsandwich: @MatthijsBurgh our current policy is not to add new pip keys for packages that are available in distribution repositories. pip-installed dependencies cannot be used in officially distributed packages. So pip packages in the official sources are only used to backfill dependencies on platforms where they are unavailable in the distribution.

If you specifically need a pip package for some reason the recommendation is now to use custom rosdep sources locally.

mjcarroll:
@MatthijsBurgh friendly ping

MatthijsBurgh:
You don't want to merge this, so I am closing this PR. For tidyness of the PR and git, I will not convert this to a non pip target. And as I don't need it, I will not add an APT target for now.

PR #40170, opened 2024-03-11
feat: add yoctopuce pip to python.yaml

nikosmitrop1995:
## Package name:

`yoctopuce-pip`

## Package Upstream Source:

[yoctolib_api](https://github.com/yoctopuce/yoctolib_python)

## Purpose of using this:

This dependency is being used for controlling the [Yoctopuce](https://www.yoctopuce.com/) devices using their Python API. This makes easier to use and control actuators and sensors.

Distro packaging links:

  - [https://pypi.org/project/yoctopuce/](https://pypi.org/project/yoctopuce/)
  - [https://pypi.org/project/yoctopuce/](https://pypi.org/project/yoctopuce/)
  - Not available
  - Not available
  - Not available
  - Not available
  - Not available
  - Not available
  - Not available
  - Not available

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9904,6 +9904,10 @@ yapf3:
   nixos: [python3Packages.yapf]
   ubuntu:
     '*': [yapf3]
+yoctopuce-pip:
+  ubuntu:
+    pip:
+      packages: [yoctopuce]
 zulip-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9760,6 +9760,10 @@ python3-yappi:
   ubuntu:
     '*': [python3-yappi]
     bionic: null
+python3-yoctopuce:
+  ubuntu:
+    pip:
+      packages: [yoctopuce]
 python3-yolov5:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -9904,6 +9904,10 @@ yapf3:
   nixos: [python3Packages.yapf]
   ubuntu:
     '*': [yapf3]
+yoctopuce-pip:
```
clalancette, line 9907: Our current best practice is to put `python3-` in front of the key:

```suggestion
python3-yoctopuce-pip:
```

(don't forget to re-alphabetize this once you do that)

clalancette (dismissed):

rosdep/python.yaml
nikosmitrop1995: I have renamed it to `python3-yoctopuce` and placed it to the correct order alphabetically. Do you think it's better to add it at `base.yaml` ?

clalancette: > I have renamed it to `python3-yoctopuce` and placed it to the correct order alphabetically. Do you think it's better to add it at `base.yaml` ?

No, it should be in `python.yaml`, but it should be called `python3-yocotopuce-pip`.

nikosmitrop1995: Done

quarkytale (approved):

audrow (approved):

PR #40183, opened 2024-03-12
Remove mqtt_client from Rolling.

clalancette:
As far as I can tell, it has never successfully even attempted to be built on Rolling: https://build.ros2.org/search/?q=mqtt_client

It probably needs a new source release, along with a new bloom release, using the correct release repository. For now, just remove it from Rolling completely; we can always add it back later.

@lreiher FYI.

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -3164,23 +3164,6 @@ repositories:
       url: https://github.com/MOLAorg/mp2p_icp.git
       version: master
     status: developed
-  mqtt_client:
-    doc:
-      type: git
-      url: https://github.com/ika-rwth-aachen/mqtt_client.git
-      version: main
-    release:
-      packages:
-      - mqtt_client
-      - mqtt_client_interfaces
-      tags:
-        release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/mqtt_client-release.git
-    source:
-      type: git
-      url: https://github.com/ika-rwth-aachen/mqtt_client.git
-      version: main
-    status: maintained
   mrpt2:
     doc:
       type: git


lreiher:
Thank you for bringing this to my attention. To be honest, I have never even realized that the rolling release has never worked.

I guess that we should also remove `message_tf_frame_transformer` with the same argument. Here, [someone recently brought the build failures to my attention](https://github.com/ika-rwth-aachen/message_tf_frame_transformer/issues/2). I wasn't aware that likely the problem is only that I need to switch to the ros2-gbp release repositories.

I will try to re-release both packages using the correct release repository soon!

clalancette:
> To be honest, I have never even realized that the rolling release has never worked.

An email is sent daily to the maintainer.  In this case, it looks like that is configured to be `lennart.reiher@rwth-aachen.de`

> I guess that we should also remove `message_tf_frame_transformer` with the same argument.

Sort of.  The difference between them is that while `message_tf_frame_transformer` fails to build, it doesn't fail to even generate a binary job, like `mqtt_client`.

> I will try to re-release both packages using the correct release repository soon!

If that happens soon, we can end up closing this PR.  Thanks for taking a look at it.

lreiher:
I have re-released new versions of both `message_tf_frame_transformer` and `mqtt_client`. All ROS 2 releases are using the ros2-gbp release repositories. Hopefully this resolves all build issues and this PR can be closed.
- https://github.com/ros/rosdistro/pull/40255
- https://github.com/ros/rosdistro/pull/40257
- https://github.com/ros/rosdistro/pull/40254
- https://github.com/ros/rosdistro/pull/40256
- https://github.com/ros/rosdistro/pull/40253
- https://github.com/ros/rosdistro/pull/40251
- https://github.com/ros/rosdistro/pull/40250
- https://github.com/ros/rosdistro/pull/40252

clalancette:
Thank you for the work here!  Closing this in favor of #40256

PR #40185, opened 2024-03-12
Remove warehouse_ros_mongo.

clalancette:
As far as I can tell, it has not had a successful build on the buildfarm in a long time: https://build.ros2.org/search/?q=warehouse_ros_mongo

Just remove it for now.

@henningkayser FYI

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -7744,20 +7744,6 @@ repositories:
       url: https://github.com/ros-planning/warehouse_ros.git
       version: ros2
     status: maintained
-  warehouse_ros_mongo:
-    doc:
-      type: git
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):
I would also be fine if it was left source-indexed but doc and release stanzas were removed in order to let it continue to be indexed (it could be buildable from source using custom rosdep sources to provide mongo.

clalancette:
> I would also be fine if it was left source-indexed but doc and release stanzas were removed in order to let it continue to be indexed (it could be buildable from source using custom rosdep sources to provide mongo.

I'm going to merge this and remove it for now, since it hasn't built in quite a while.  If we want to add it back as source-only, we can do that later.

PR #40227, opened 2024-03-16
Fix rosdep keys for wxwidgets in modern Ubuntu/Debian

jlblancoc:
Updated the package names for Debian and Ubuntu for wxWidgets. 

References:
- https://packages.ubuntu.com/search?keywords=wxwidgets&searchon=sourcenames
- https://tracker.debian.org/pkg/wxwidgets3.0
- https://tracker.debian.org/pkg/wxwidgets3.2

I have the doubt on the preferred way to do this: either use the wildcard '*' for the new current distros (with the idea of it will be working in the future), or for the older ones. I picked the first. 
Please, let me know if it's desired to do it the other way around.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8601,9 +8601,9 @@ wxwidgets:
     '7': [wxGTK-devel]
   ubuntu:
     '*': [libwxgtk3.2-dev]
+    bionic: [libwxgtk3.0-dev]
     focal: [libwxgtk3.0-gtk3-dev]
     jammy: [libwxgtk3.0-gtk3-dev]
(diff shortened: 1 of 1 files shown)


jlblancoc:
I don't understand the reason of the CI failure from the error, please let me know how to fix it.

BTW: this PR is needed to handle new package names in Debian sid & Ubuntu 24.04 Noble, which leads to error for example [here](https://build.ros2.org/job/Rdev__mrpt2__ubuntu_noble_amd64/10/console).

jlblancoc:
It was the keys order, it's fixed now.

PR #40241, opened 2024-03-18
libsoup2.4-dev in 'rosdep/base.yml'

fbe555:

## Package name:

libsoup2.4-dev

## Package Upstream Source:

https://gitlab.gnome.org/GNOME/libsoup

## Purpose of using this:

Needed as a system dependency for the [ros-gst-bridge](https://github.com/BrettRD/ros-gst-bridge) package 

  - https://packages.debian.org/bookworm/libsoup2.4-dev
  - https://packages.ubuntu.com/mantic/libsoup2.4-dev
  - https://packages.fedoraproject.org/pkgs/libsoup/libsoup-devel/
  - https://archlinux.org/packages/extra/x86_64/libsoup/
  - https://packages.gentoo.org/packages/net-libs/libsoup
  - https://formulae.brew.sh/formula/libsoup@2#default
  - https://pkgs.alpinelinux.org/package/edge/community/x86_64/libsoup-dev
  - https://search.nixos.org/packages?channel=23.11&show=libsoup&from=0&size=50&sort=relevance&type=packages&query=libsoup
  - https://software.opensuse.org/package/libsoup-2_4-1
  - NOT AVAILABLE

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5753,6 +5753,16 @@ libsoundio-dev:
   macports: [libsoundio]
   nixos: [libsoundio]
   ubuntu: [libsoundio-dev]
+libsoup2.4-dev:
+  alpine: [libsoup-dev]
+  arch: [libsoup]
+  debian: [libsoup2.4-dev]
+  fedora: [libsoup-devel]
+  gentoo: [net-libs/libsoup-2.4]
+  macports: [libsoup@2]
+  nixos: [libsoup]
+  opensuse: [libsoup-2_4-1]
+  ubuntu: [libsoup2.4-dev]
 libspatialindex-dev:
   debian: [libspatialindex-dev]
   fedora: [spatialindex-devel]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5753,7 +5753,7 @@ libsoundio-dev:
   macports: [libsoundio]
   nixos: [libsoundio]
   ubuntu: [libsoundio-dev]
-libsoup-dev:
+libsoup2-dev:
   alpine: [libsoup-dev]
   arch: [libsoup]
   debian: [libsoup2.4-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5753,7 +5753,7 @@ libsoundio-dev:
   macports: [libsoundio]
   nixos: [libsoundio]
   ubuntu: [libsoundio-dev]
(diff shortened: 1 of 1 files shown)


audrow (dismissed):

rosdep/base.yaml
```diff
@@ -5753,6 +5753,16 @@ libsoundio-dev:
   macports: [libsoundio]
   nixos: [libsoundio]
   ubuntu: [libsoundio-dev]
+libsoup2.4-dev:
```
clalancette, line 5756: I know that our typical policy is to make the key match the Ubuntu package name, but in this case I don't think it makes sense.  In particular, `libsoup-devel` on Fedora, for instance, is already libsoup 2.74.  So I'm going to suggest we call this key `libsoup-dev` instead.

clalancette (dismissed):

rosdep/base.yaml
fbe555: Sounds resonable. I have renamed it.

sloretz: @clalancette a downside of this is we now have a versionless key `libsoup-dev` and have to choose if or when  we'll upgrade it to the [libsoup-3.0-dev package](https://packages.debian.org/search?keywords=libsoup-3.0-dev).

sloretz (approved):
Changes look good to me, though I lean more towards making the rosdep key name match the debian package name for consistency.

rosdep/base.yaml
```diff
@@ -5753,6 +5753,16 @@ libsoundio-dev:
   macports: [libsoundio]
   nixos: [libsoundio]
   ubuntu: [libsoundio-dev]
+libsoup-dev:
```
sloretz, line 5756: @fbe555 -  thank you for your patience, would you be willing to change this name to `libsoup2-dev`?

Per offline discussion with @clalancette & @cottsay it seems like the best compromise we could come up with for handling multiple OS's packaging different versions of 2.x and 3.x of libsoup.

sloretz (changes requested):

rosdep/base.yaml
fbe555: Sure - that does make better sense. 
If you wanna reward my patience, pop by my other PR #40264. I Think it's ready to merge 😉

fbe555:
Renamed it to libsoup2-dev now.

sloretz (approved):

cottsay (approved):

PR #40276, opened 2024-03-20
Add in the gazebo vendor packages.

clalancette:
We are switching to a model where we vendor Gazebo within ROS itself, rather than having it separate.  This will make it very clear which version of Gazebo is supported with which version of ROS.  This PR adds in the necessary packages to do this vendoring.

There are a number of steps to follow to get this completely integrated:
1.  Get this PR reviewed for REP-144 naming, and merged.
2.  Add in ros2-gbp repositories for these new packages (https://github.com/ros2-gbp/ros2-gbp-github-org/pull/474)
3.  Addition of a number of rosdep keys (there is a list in https://github.com/gazebo-tooling/release-tools/issues/1117#issuecomment-2005407791)
4.  Update of https://github.com/ros2/ros2/blob/rolling/ros2.repos to use these new vendor packages.

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1926,6 +1926,114 @@ repositories:
       url: https://github.com/borglab/gtsam.git
       version: develop
     status: developed
+  gz_cmake_vendor:
+    doc:
+      type: git
+      url: https://github.com/gazebo-release/gz_cmake_vendor.git
+      version: main
+    source:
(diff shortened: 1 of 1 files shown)


clalancette:
@azeey @nuclearsandwich FYI

azeey (approved):

nuclearsandwich (approved):

PR #40280, opened 2024-03-20
Remove gazebo_ros_pkgs and dependencies from Rolling

j-rivero:
Gazebo classic aka gazebo11 package is not present in Ubuntu Noble nor is the plan of the Gazebo simulation team to package it anymore since the recommended path is to migrate to the new Gazebo, see https://gazebosim.org/docs/latest/migrating_gazebo_classic_ros2_packages. 

I'm completely removing in this PR the entries from the rolling `distribution.yaml`, not sure if there is any better way of dealing with this kind of "retirement".

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1018,27 +1018,6 @@ repositories:
       url: https://github.com/ros/diagnostics.git
       version: ros2
     status: maintained
-  dolly:
-    doc:
-      type: git
-      url: https://github.com/chapulina/dolly.git
-      version: galactic
-    release:
-      packages:
-      - dolly
-      - dolly_follow
-      - dolly_gazebo
-      - dolly_ignition
-      tags:
-        release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/dolly-release.git
-      version: 0.4.0-5
-    source:
-      test_pull_requests: true
-      type: git
-      url: https://github.com/chapulina/dolly.git
-      version: galactic
-    status: developed
   domain_bridge:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


rolling/distribution.yaml
```diff
@@ -1018,27 +1018,6 @@ repositories:
...
-  dolly:
-    doc:
-      type: git
-      url: https://github.com/chapulina/dolly.git
-      version: galactic
-    release:
-      packages:
-      - dolly
-      - dolly_follow
-      - dolly_gazebo
```
clalancette, line 1030: I'm not 100% sure of this, but is it only `dolly_gazebo` that is failing here?  If that is the case, maybe we should just blacklist that particular package, rather than totally removing the project.

rolling/distribution.yaml
```diff
@@ -7129,7 +7069,6 @@ repositories:
     release:
       packages:
       - turtlebot3_fake_node
-      - turtlebot3_gazebo
```
clalancette, line 7132: I think this isn't going to work how we want, though maybe @nuclearsandwich can comment.  In particular, the next time someone does a `bloom-release`, I believe that this will "come back".  Maybe the right thing to do here is to blacklist the package in https://github.com/ros2/ros_buildfarm_config/blob/ros2/rolling/release-build.yaml ?  Not sure.

rolling/distribution.yaml
```diff
@@ -7588,25 +7527,6 @@ repositories:
...
     status: developed
-  velodyne_simulator:
-    doc:
-      type: git
-      url: https://bitbucket.org/DataspeedInc/velodyne_simulator.git
-      version: foxy-devel
-    release:
-      packages:
-      - velodyne_description
-      - velodyne_gazebo_plugins
```
clalancette, line 7599: Same question here; can we just blacklist this one package, rather than removing the whole thing?

mjcarroll:
@j-rivero friendly ping

clalancette:
I'm actually going to close this one out for now, since we've removed some of this stuff separately.  We can always open other, smaller PRs with the individual things that we need to remove.

PR #40284, opened 2024-03-20
Add glslc key

azeey:

## Package name:

`glslc`

## Package Upstream Source:

https://github.com/google/shaderc

## Purpose of using this:

Needed by [`gz_ogre_next_vendor`](https://github.com/gazebo-release/gz_ogre_next_vendor)

Distro packaging links:

  - debian: https://packages.debian.org/bookworm/glslc
  - https://packages.ubuntu.com/noble/glslc
  - https://packages.fedoraproject.org/pkgs/shaderc/glslc
  - https://archlinux.org/packages/extra/x86_64/shaderc/
  - https://packages.gentoo.org/packages/media-libs/shaderc
  - skipped
  - skipped
  - skipped
  - skipped
  - skipped

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1456,9 +1456,13 @@ glslc:
   debian: [glslc]
   fedora: [glslc]
   gentoo: [media-libs/shaderc]
+  rhel:
+    '*': [glslc]
+    '8': null
   ubuntu:
-    '*': null
-    noble: [glslc]
+    '*': [glslc]
+    focal: null
+    jammy: null
 glut:
   arch: [freeglut]
   debian: [freeglut3-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1451,6 +1451,12 @@ glpk:
   gentoo: [sci-mathematics/glpk]
   rhel: [glpk-devel]
   ubuntu: [libglpk-dev]
+glslc:
+  arch: [shaderc]
+  debian: [glslc]
+  fedora: [glslc]
+  gentoo: [media-libs/shaderc]
+  ubuntu: [glslc]
 glut:
   arch: [freeglut]
   debian: [freeglut3-dev]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

rosdep/base.yaml
```diff
@@ -1451,6 +1451,14 @@ glpk:
...
   rhel: [glpk-devel]
   ubuntu: [libglpk-dev]
+glslc:
+  arch: [shaderc]
+  debian: [glslc]
+  fedora: [glslc]
+  gentoo: [media-libs/shaderc]
+  ubuntu:
+    '*': null
+    noble: [glslc]
```
clalancette, line 1461: We tend to do this the other way, with the thinking that if it is in `noble`, it will also be in later versions.  This means we won't have to update the key for the next release:

```suggestion
    '*': [glslc]
    focal: null
    jammy: null
```

clalancette (changes requested):
Besides the inline change, the bot thinks we can satisfy this package on rhel as well.

azeey:
> Besides the inline change, the bot thinks we can satisfy this package on rhel as well.

Done in https://github.com/ros/rosdistro/pull/40284/commits/659c6a77a1849df2e41607c7c3cbe3ca264fd7af

mjcarroll (approved):

rosdep/base.yaml
```diff
@@ -1451,6 +1451,18 @@ glpk:
   gentoo: [sci-mathematics/glpk]
   rhel: [glpk-devel]
   ubuntu: [libglpk-dev]
+glslc:
+  arch: [shaderc]
+  debian: [glslc]
```
clalancette, line 1456:
```suggestion
  debian:
    '*': [glslc]
    bullseye: null
```

clalancette (changes requested):

rosdep/base.yaml
azeey: Done in https://github.com/ros/rosdistro/pull/40284/commits/5e286d1de9b0b0d637fe83067964e040278f41d8

clalancette (approved):

PR #40304, opened 2024-03-21
Update `zziplib` key for Ubuntu Noble

azeey:

## Package name:

`zziplib`

## Purpose of using this:

On Ubuntu Noble, the package `libzzip-dev` now depends on `libzzip-0-13t64` which breaks `libzzip-0-13`. We believe this name change is due to the 64bit time  transition (see https://wiki.debian.org/ReleaseGoals/64bit-time)

Distro packaging links:

  - https://packages.ubuntu.com/noble/libzzip-dev
  - https://packages.ubuntu.com/noble/libzzip-0-13t64

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8966,7 +8966,4 @@ zziplib:
   nixos: [zziplib]
   opensuse: [zziplib-devel]
   rhel: [zziplib-devel]
-  ubuntu:
-    '*': [libzzip-0-13t64, libzzip-dev]
-    focal: [libzzip-0-13, libzzip-dev]
-    jammy: [libzzip-0-13, libzzip-dev]
+  ubuntu: [libzzip-dev]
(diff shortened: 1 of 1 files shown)


clalancette (approved):

rosdep/base.yaml
```diff
@@ -8966,4 +8966,4 @@ zziplib:
   nixos: [zziplib]
   opensuse: [zziplib-devel]
   rhel: [zziplib-devel]
-  ubuntu: [libzzip-0-13, libzzip-dev]
+  ubuntu: [libzzip-dev]
```
clalancette, line 8969: While we are here, we should do the same for the `debian` key above.

clalancette (changes requested):

rosdep/base.yaml
azeey: Done https://github.com/ros/rosdistro/pull/40304/commits/56f87f672afa34f140eb7e6401e25d98582f068c

clalancette (approved):
Looks good to me, thanks.

PR #40305, opened 2024-03-22
Revert "generate_parameter_library: 0.3.8-1 in 'iron/distribution.yaml' [bloom]"

Yadunund:
Reverts ros/rosdistro#40149

The changes here result in regressions in moveit2 which block the upcoming Iron sync.
See https://github.com/ros-planning/moveit2/issues/2758#issuecomment-2013696411

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -1797,16 +1797,13 @@ repositories:
       version: main
     release:
       packages:
-      - cmake_generate_parameter_module_example
       - generate_parameter_library
-      - generate_parameter_library_example
       - generate_parameter_library_py
-      - generate_parameter_module_example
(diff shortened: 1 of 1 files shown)


christian-rauch:
@clalancette Can we please also revert this for `humble` (i.e. revert https://github.com/ros/rosdistro/pull/40148)? If this version of `generate_parameter_library` has an issue, this will also be present in `humble` and should be reverted there too.

clalancette:
> @clalancette Can we please also revert this for `humble` (i.e. revert #40148)? If this version of `generate_parameter_library` has an issue, this will also be present in `humble` and should be reverted there too.

Please feel free to open a PR to do that.  @sea-bass FYI

PR #40407, opened 2024-03-28
[Iron] Removing opennav-coverage

SteveMacenski:
This is still after months waiting on an update to F2C binaries to fix a build issue. At this point if its not fixed yet, it may not be in any reasonable timeframe so removing from here to save the compute time on the daily builds that are failing (and emails) 

I'll resubmit when I have reason to think its resolved

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -4003,24 +4003,6 @@ repositories:
         release: release/iron/{package}/{version}
       url: https://github.com/ros2-gbp/ompl-release.git
       version: 1.5.2-4
-  opennav_coverage:
-    release:
-      packages:
-      - opennav_coverage
-      - opennav_coverage_bt
-      - opennav_coverage_demo
-      - opennav_coverage_msgs
-      - opennav_coverage_navigator
-      - opennav_row_coverage
-      tags:
-        release: release/iron/{package}/{version}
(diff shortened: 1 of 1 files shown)


sloretz (changes requested):
It looks like some of the packages successfully build:

* https://build.ros2.org/job/Ibin_uJ64__opennav_coverage_bt__ubuntu_jammy_amd64__binary/
* https://build.ros2.org/job/Ibin_rhel_el964__opennav_coverage_bt__rhel_9_x86_64__binary/
* https://build.ros2.org/job/Ibin_uJ64__opennav_coverage_msgs__ubuntu_jammy_amd64__binary/
* https://build.ros2.org/job/Ibin_rhel_el964__opennav_coverage_msgs__rhel_9_x86_64__binary/
* https://build.ros2.org/job/Ibin_ujv8_uJv8__opennav_coverage_bt__ubuntu_jammy_arm64__binary/
* https://build.ros2.org/job/Ibin_ujv8_uJv8__opennav_coverage_msgs__ubuntu_jammy_arm64__binary/

What do you think about keeping the packages `opennav_coverage_bt` and `opennav_coverage_msgs` and the source entry?

SteveMacenski:
They're meaningless without the actual servers that can be called using those interfaces and behavior trees :smil

sloretz (approved):

PR #40413, opened 2024-03-29
Removing euslime

Affonso-Gui:
Binary builds for the euslime package have been failing for some time, so I'm taking them down until I (or someone else) can figure out a fix.

https://build.ros.org/job/Nbin_ufv8_uFv8__euslime__ubuntu_focal_arm64__binary/
https://build.ros.org/job/Nbin_ufhf_uFhf__euslime__ubuntu_focal_armhf__binary/

On a related note, any info on how to replicate the buildfarm results locally?
I've been trying the prerelease tests, but I can't quite get the arm ones running.
(do I really need an arm machine to run the arm tests?)
https://wiki.ros.org/regression_tests

@k-okada

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -2647,17 +2647,6 @@ repositories:
       url: https://github.com/ika-rwth-aachen/etsi_its_messages.git
       version: main
     status: developed
-  euslime:
-    release:
-      tags:
-        release: release/noetic/{package}/{version}
-      url: https://github.com/jsk-ros-pkg/euslime-release.git
-      version: 1.1.4-4
-    source:
-      type: git
-      url: https://github.com/jsk-ros-pkg/euslime.git
-      version: master
-    status: developed
   euslisp:
     doc:
       type: git


sloretz:
> (do I really need an arm machine to run the arm tests?)

You can use qemu-user-static, but I don't know of a good resource that explains it, and I don't understand it sufficiently to give a good short explanation.

I think installing qemu-user-static and then running one of the official docker images under the target architecture is probably the easiest way. Maybe this page has enough info? https://github.com/multiarch/qemu-user-static/blob/master/README.md

sloretz (changes requested):
It looks like it's building on amd64. 

https://build.ros.org/job/Nbin_uF64__euslime__ubuntu_focal_amd64__binary/

Instead of removing it here, I'd recommend adding euslime to the ROS 1 buildfarm's package blacklists for arm64 and armhf. Here's a PR that shows how: https://github.com/ros-infrastructure/ros_buildfarm_config/pull/224/files

Affonso-Gui:
@sloretz 

Blacklisting on ARM does seem like a better solution for this case!
Created a PR in https://github.com/ros-infrastructure/ros_buildfarm_config/pull/228

> I think installing qemu-user-static and then running one of the official docker images under the target architecture is probably the easiest way. Maybe this page has enough info? https://github.com/multiarch/qemu-user-static/blob/master/README.md

Thanks for the hint, I will take a look.

sloretz:
Closing as completed by https://github.com/ros-infrastructure/ros_buildfarm_config/pull/228 🎉

PR #40529, opened 2024-04-09
Add python3-pynput apt package for debian and ubuntu

dnikitop:

## Package name:

pynput (python3-pynput)

## Package Upstream Source:

https://github.com/moses-palmer/pynput

## Purpose of using this:

Uses apt instead of pip.

  - https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=pynput)
  - https://packages.ubuntu.com/search?keywords=pynput&searchon=names
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -438,11 +438,6 @@ pymodbustcp-pip:
   ubuntu:
     pip:
       packages: [pyModbusTCP]
-pynput:
-  ubuntu:
-    jammy: [python3-pynput]
-    noble: [python3-pynput]
-  debian: [python3-pynput]
 pynput-pip:
   debian:
     pip:
@@ -8021,6 +8016,11 @@ python3-pynmeagps-pip:
   ubuntu:
     pip:
       packages: [pynmeagps]
+python3-pynput:
+  debian: [python3-pynput]
+  ubuntu:
+    jammy: [python3-pynput]
+    noble: [python3-pynput]
 python3-pyosmium:
   debian: [python3-pyosmium]
   fedora: [python3-osmium]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -438,6 +438,11 @@ pymodbustcp-pip:
   ubuntu:
     pip:
       packages: [pyModbusTCP]
+pynput:
+  ubuntu:
+    jammy: [python3-pynput]
+    noble: [python3-pynput]
+  debian: [python3-pynput]
```
clalancette, line 445: This should generally match the name of the debian key.  Also, there is no reason to call out the Ubuntu names explicitly.  And finally, the distribution names need to be in alphabetical order.  So this should be:

```suggestion
python3-pynput:
  debian: [python3-pynput]
  ubuntu: [python3-pynput]
```

(you'll also need to move this around within this file to make the whole thing alphabetical)

clalancette (changes requested):

rosdep/python.yaml
clalancette: > Focal for ubuntu doesn't host python3-pynput, so only have nobel and jammy specified. We can combine `pynput-pip` and `python3-pynput` and specify pip/non pip for specific releases, but will break backwards compatibility the the name 'pynput-pip'

OK, good point about Focal.  In that case, let's change this to be forwards-compatible:

```
  ubuntu:
    '*': [python3-pynput]
    focal: null
```

mjcarroll (approved):

clalancette (approved):

PR #40543, opened 2024-04-10
add toppra pip dependency to python.yaml

kbrameld:

## Package name:

**python3-toppra-pip**

## Package Upstream Source:

https://github.com/hungpham2511/toppra

## Purpose of using this:

The library is a **Time-Optimal Path Parameterization** and is a popular library for motion planning.

- pip: https://pypi.org/project/toppra/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9323,6 +9323,19 @@ python3-toml:
   gentoo: [dev-python/toml]
   nixos: [python3Packages.toml]
   ubuntu: [python3-toml]
+python3-toppra:
+  debian:
+    pip:
+      packages: [toppra]
+  fedora:
+    pip:
+      packages: [toppra]
+  osx:
+    pip:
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

rosdep/python.yaml
```diff
@@ -9323,6 +9323,19 @@ python3-toml:
   gentoo: [dev-python/toml]
   nixos: [python3Packages.toml]
   ubuntu: [python3-toml]
+python3-toppra:
```
sloretz, line 9326: Please since this is a pip only key please change its name to `python3-toppra-pip`

sloretz (changes requested):

rosdep/python.yaml
kbrameld: thanks for the review, this has been updated

sloretz (approved):

PR #40552, opened 2024-04-11
Add pyads python lib

ghost:

## Package name:

python3-pyads-pip

## Package Upstream Source:

[https://github.com/chwiede/pyads](https://github.com/chwiede/pyads)
Available through pip in https://pypi.org/project/pyads/

## Purpose of using this:

This dependency is used as a wrapper for the ads library. It allows to communicate with Beckhoff PLC using python3. It is useful in case anyone wants to integrate a Beckhoff PLC with a robot stack using ROS

Distro packaging links:

- Python: https://pypi.org/project/pyads/

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3,8 +3,6 @@
 # see REP 143: http://ros.org/reps/rep-0143.html
 ---
 release_platforms:
-  debian:
-  - bullseye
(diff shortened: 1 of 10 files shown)


mjcarroll (approved):

sloretz (approved):

PR #40593, opened 2024-04-14
Add do-mpc pip package to python.yaml for debian and ubuntu

KozaMateusz:

## Package name:

python3-do-mpc-pip

## Package Upstream Source:

[https://github.com/do-mpc/do-mpc](https://github.com/do-mpc/do-mpc)

## Purpose of using this:

This dependency is useful for quick development of the Model Predictive Control algorithm in the ros environment. 

- Debian: https://pypi.org/project/do-mpc/ 
- Ubuntu: https://pypi.org/project/do-mpc/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5575,6 +5575,13 @@ python3-dlib-pip:
     pip:
       depends: [cmake, build-essential, python3-dev]
       packages: [dlib]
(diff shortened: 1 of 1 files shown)


mjcarroll:

mjcarroll (approved):

sloretz (approved):

PR #40599, opened 2024-04-15
Add OpenTelemetry and Adafruit ServoKit Libraries to Rosdep

proffalken:

## Package name:

* adafruit-servokit-pip
* opentelemetry-api-pip
* opentelemetry-exporter-otlp-pip
* opentelemetry-sdk-pip

## Package Upstream Source:

* https://github.com/adafruit/Adafruit_CircuitPython_ServoKit
* https://github.com/open-telemetry/opentelemetry-python

## Purpose of using this:

These dependencies are proposed for the following reasons:
1. Adafruit ServoKit allows for control of standard hobby servos via the I2C PCA9685 driver boards, enabling small robot arms to be controlled via ROS easily on a Raspberry Pi or similar device
2. [Open Telemetry](https://www.opentelemetry.io) is rapidly becoming the defacto standard for monitoring software performance across the IT industry.  Adding these libraries enables us to visualise the interactions between ROS components in standard observability tooling such as Grafana, Honeycomb.io, or DataDog

Distro packaging links:

  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge
  - N/A - Python Pip Packge

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -29,6 +29,16 @@ adafruit-pca9685-pip:
   ubuntu:
     pip:
       packages: [adafruit-pca9685]
+adafruit-servokit-pip:
+  debian:
+    pip:
+      packages: [adafruit-circuitpython-servokit]
+  fedora:
+    pip:
+      packages: [adafruit-circuitpython-servokit]
+  ubuntu:
+    pip:
+      packages: [adafruit-circuitpython-servokit]
 autograd-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -20,13 +20,7 @@ adafruit-mcp3008-pip:
     pip:
       packages: [Adafruit-MCP3008]
 python3-adafruit-circuitpython-motor-pip:
-  debian:
-    pip:
-      packages: [adafruit-circuitpython-motor]
-  fedora:
-    pip:
-      packages: [adafruit-circuitpython-motor]
-  ubuntu:
+  '*':
     pip:
       packages: [adafruit-circuitpython-motor]
 adafruit-pca9685-pip:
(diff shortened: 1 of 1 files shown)


mjcarroll:

proffalken:
> Standard pip disclaimer: ROS packages that depend on pip keys cannot be released into a ROS distribution. They can only be depended on by from-source builds. Because of this, system packages are highly preferred to pip packages.

Thanks, unfortunately these packages are only available in pip format and I'm not keen to take on OS packaging for them.

Should I just continue to keep my own fork and use a custom rosdep format rather than asking for this to be merged?

mjcarroll:
> Should I just continue to keep my own fork and use a custom rosdep format rather than asking for this to be merged?

It's perfectly fine to merge them here, we just generally like to remind people that it's the case so they aren't surprised when they go to package something that depends on a pip package.

proffalken:
For completeness, this is the end result in Grafana when the otel libraries are included:

image
image
image
image

rosdep/python.yaml
```diff
@@ -19,6 +19,16 @@ adafruit-mcp3008-pip:
   ubuntu:
     pip:
       packages: [Adafruit-MCP3008]
+adafruit-circuitpython-motor-pip:
```
sloretz, line 22: Since these are pip only keys, please prefix them with `python3-`. That is `adafruit-circuitpython-motor-pip` should become `python3-adafruit-circuitpython-motor-pip`.

sloretz (dismissed):

rosdep/python.yaml
```diff
@@ -19,7 +19,17 @@ adafruit-mcp3008-pip:
   ubuntu:
     pip:
       packages: [Adafruit-MCP3008]
-adafruit-pca9685-pip:
```
clalancette, line 22: While the advice that @sloretz gave you here was correct for new keys, we shouldn't change the old ones (even if they don't meet our current standards).  That's because it might break people who are already using this.  Please put this one back to:

```
adafruit-pca9685-pip
```

clalancette (changes requested):

clalancette (approved):

nuclearsandwich (approved):

PR #40669, opened 2024-04-18
Change robot_localization source branch

ayrton04:
When I originally bloomed for `iron` a few days ago, I updated the upsteam devel branch to `iron-devel`, but it doesn't appear to have updated in the PR. I just ran it again, and it shows that it knows the upstream branch should be `iron-devel`, but the `source` stanza was still not getting updated in the PR. So I am just manually changing that line.

And yes, the branch [exists](https://github.com/cra-ros-pkg/robot_localization/tree/iron-devel).

```
$ bloom-release -t iron -r iron robot_localization --edit
ROS Distro index file associate with commit 'da9e9275d08c6d7db05995661026d93e4be4275b'
New ROS Distro index url: 'https://raw.githubusercontent.com/ros/rosdistro/da9e9275d08c6d7db05995661026d93e4be4275b/index-v4.yaml'
==> Fetching 'robot_localization' repository from 'https://github.com/ros2-gbp/robot_localization-release.git'
Cloning into '/tmp/tmptbk2vrz2'...
remote: Enumerating objects: 12893, done.
remote: Counting objects: 100% (12893/12893), done.
remote: Compressing objects: 100% (7347/7347), done.
remote: Total 12893 (delta 5334), reused 12686 (delta 5131), pack-reused 0
Receiving objects: 100% (12893/12893), 12.68 MiB | 1.15 MiB/s, done.
Resolving deltas: 100% (5334/5334), done.
Track 'iron' exists, editing...
Repository Name:
  <name>
    Name of the repository (used in the archive name)
  upstream
    Default value, leave this as upstream if you are unsure
  ['robot_localization']: 
Upstream Repository URI:
  <uri>
    Any valid URI. This variable can be templated, for example an svn url
    can be templated as such: "https://svn.foo.com/foo/tags/foo-:{version}"
    where the :{version} token will be replaced with the version for this release.
  ['https://github.com/cra-ros-pkg/robot_localization.git']: 
Upstream VCS Type:
  git
    Upstream URI is a git repository
  hg
    Upstream URI is a hg repository
  svn
    Upstream URI is a svn repository
  tar
    Upstream URI is a tarball
  ['git']: 
Version:
  :{auto}
    This means the version will be guessed from the devel branch.
    This means that the devel branch must be set, the devel branch must exist,
    and there must be a valid package.xml in the upstream devel branch.
  :{ask}
    This means that the user will be prompted for the version each release.
    This also means that the upstream devel will be ignored.
  <version>
    This will be the version used.
    It must be updated for each new upstream version.
  [':{auto}']: 
Release Tag:
  :{version}
    This means that the release tag will match the :{version} tag.
    This can be further templated, for example: "foo-:{version}" or "v:{version}"
    
    This can describe any vcs reference. For git that means {tag, branch, hash},
    for hg that means {tag, branch, hash}, for svn that means a revision number.
    For tar this value doubles as the sub directory (if the repository is
    in foo/ of the tar ball, putting foo here will cause the contents of
    foo/ to be imported to upstream instead of foo itself).
  :{ask}
    This means the user will be prompted for the release tag on each release.
  :{none}
    For svn and tar only you can set the release tag to :{none}, so that
    it is ignored.  For svn this means no revision number is used.
  [':{version}']: 
Upstream Devel Branch:
  <vcs reference>
    Branch in upstream repository on which to search for the version.
    This is used only when version is set to ':{auto}'.
  ['iron-devel']: 
ROS Distro:
  <ROS distro>
    This can be any valid ROS distro, e.g. humble, iron, noetic, rolling
  ['iron']: 
Patches Directory:
  <path in bloom branch>
    This can be any valid relative path in the bloom branch. The contents
    of this folder will be overlaid onto the upstream branch after each
    import-upstream.  Additionally, any package.xml files found in the
    overlay will have the :{version} string replaced with the current
    version being released.
  :{none}
    Use this if you want to disable overlaying of files.
  [None]: 
Release Repository Push URL:
  <url>
    (optional) Used when pushing to remote release repositories. This is only
    needed when the release uri which is in the rosdistro file is not writable.
    This is useful, for example, when a releaser would like to use a ssh url
    to push rather than a https:// url.
  :{none}
    This indicates that the default release url should be used.
  [None]: 
Saving 'iron' track.
```

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -5575,7 +5575,7 @@ repositories:
       test_pull_requests: true
       type: git
       url: https://github.com/cra-ros-pkg/robot_localization.git
-      version: ros2
+      version: iron-devel
     status: maintained
   robot_state_publisher:
     doc:


sloretz (approved):
[Holding for Iron sync](https://discourse.ros.org/t/preparing-for-iron-patch-and-sync-2024-04-22)

PR #40670, opened 2024-04-18
cob_common: 0.7.11-1 in 'noetic/distribution.yaml' [bloom]

fmessmer:
Increasing version of package(s) in repository `cob_common` to `0.7.11-1`:

- upstream repository: https://github.com/4am-robotics/cob_common.git
- release repository: https://github.com/4am-robotics/cob_common-release.git
- bloom version: `0.12.0`
- previous version for package: `0.7.10-1`

## cob_actions

- No changes

## cob_common

```
* Merge pull request #308 <https://github.com/4am-robotics/cob_common/issues/308> from fmessmer/noetic-devel
  [ROS1] cob4 eol cleanup
* remove cob_description
* remove raw_description
* Contributors: Felix Messmer, fmessmer
```

## cob_msgs

- No changes

## cob_srvs

- No changes

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1096,14 +1096,12 @@ repositories:
       packages:
       - cob_actions
       - cob_common
-      - cob_description
       - cob_msgs
       - cob_srvs
-      - raw_description
       tags:
         release: release/noetic/{package}/{version}
       url: https://github.com/4am-robotics/cob_common-release.git
-      version: 0.7.10-1
+      version: 0.7.11-1
     source:
       type: git
       url: https://github.com/4am-robotics/cob_common.git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1105,7 +1105,7 @@ repositories:
     source:
       type: git
       url: https://github.com/4am-robotics/cob_common.git
-      version: kinetic_dev
+      version: noetic-devel
     status: maintained
   cob_control:
     doc:
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1091,7 +1091,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/4am-robotics/cob_common.git
-      version: kinetic_release_candidate
+      version: noetic-release-candidate
     release:
       packages:
       - cob_actions
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -1107,6 +1107,24 @@ repositories:
       url: https://github.com/4am-robotics/cob_common.git
       version: noetic-devel
     status: maintained
+  cob_common_legacy:
+    doc:
+      type: git
+      url: https://github.com/4am-robotics/cob_common.git
+      version: kinetic_release_candidate
+    release:
+      packages:
+      - cob_description
+      - raw_description
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/4am-robotics/cob_common-release.git
+      version: 0.7.10-1
+    source:
+      type: git
+      url: https://github.com/4am-robotics/cob_common.git
+      version: kinetic_dev
+    status: end-of-life
   cob_control:
     doc:
       type: git


noetic/distribution.yaml
```diff
@@ -1091,14 +1091,30 @@ repositories:
...
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/4am-robotics/cob_common-release.git
+      version: 0.7.11-1
+    source:
+      type: git
+      url: https://github.com/4am-robotics/cob_common.git
+      version: noetic-devel
+    status: maintained
+  cob_common_legacy:
```
fmessmer, line 1110: I hope I'm doing the right thing here - could you please confirm that this PR will:
- release a new version `0.7.11` of `cob_actions`, `cob_common`, `cob_msgs` and `cob_srvs` from `noetic-release-candidate`
- keep the last release `0.7.10` for `cob_description` and `raw_description` available based on `kinetic_release_candidate`

sloretz (approved):
I think this does what you intend, but it's complicated enough that I'll ask for a second reviewer.

noetic/distribution.yaml
fmessmer: Thank you for the feedback - I guess the other linked PRs are then also what I intend.
Thus, go ahead with the review and merge process.

noetic/distribution.yaml
```diff
@@ -1091,14 +1091,30 @@ repositories:
     doc:
       type: git
       url: https://github.com/4am-robotics/cob_common.git
-      version: kinetic_release_candidate
+      version: noetic-release-candidate
```
clalancette, line 1094: It is uncommon for the `doc` version and the `source` version to be different.  It isn't strictly disallowed, but can you explain why these are different?

noetic/distribution.yaml
```diff
@@ -1091,14 +1091,30 @@ repositories:
...
     release:
       packages:
       - cob_actions
       - cob_common
-      - cob_description
       - cob_msgs
       - cob_srvs
+      tags:
+        release: release/noetic/{package}/{version}
+      url: https://github.com/4am-robotics/cob_common-release.git
```
clalancette, line 1103: Also, https://github.com/4am-robotics/cob_common-release/blob/14d26789362ce411824cdcfc0099328475d50e41/tracks.yaml#L139 uses `noetic-release-candidate` branch to release from.  Most likely, all of `doc`, `source`, and the `devel_branch` in tracks.yaml should be the same.  At least, that will make things the most understandable.

noetic/distribution.yaml
```diff
@@ -1091,14 +1091,30 @@ repositories:
...
+    source:
+      type: git
+      url: https://github.com/4am-robotics/cob_common.git
+      version: noetic-devel
+    status: maintained
+  cob_common_legacy:
+    doc:
+      type: git
+      url: https://github.com/4am-robotics/cob_common.git
+      version: kinetic_release_candidate
```
clalancette, line 1114: Same question here; this should probably be the same as the `source` version.

noetic/distribution.yaml
fmessmer: I actually updated the doc (noetic-release-candidate) and source (noetic-devel) branch to match the new devel (noetic-devel) branch for the packages released with version 0.7.11 under the cob_common entry

Or do you mean why a *_release_candidate branch is used for doc and a *-devel branch for source?
The latter, I actually dont know myself, I just kept to the scheme used before, i.e. kinetic_release_candidate for doc and kinetic_dev for source

If you ask me to change this, I'd rather do this in a follow-up PR that changes it consistebtly for all cob_* repos - as it might affect more repos beyond the 4 repos that I updated in the current cycle

What do you think?

fmessmer:
@quarkytale @clalancette 
the PR is marked "changes requested" - however, please see my https://github.com/ros/rosdistro/pull/40670#discussion_r1572774372 
How would you like to proceed here and what do you want me to change?

noetic/distribution.yaml
sloretz: @clalancette friendly ping

fmessmer: anything I can/should do here for now?

clalancette: > Or do you mean why a *_release_candidate branch is used for doc and a *-devel branch for source? The latter, I actually dont know myself, I just kept to the scheme used before, i.e. kinetic_release_candidate for doc and kinetic_dev for source It has probably been like this doc vs. source for years now...
> 
> If you ask me to change this, I'd rather do this in a follow-up PR that changes it consistebtly for all cob_* repos - as it might affect more repos beyond the 4 repos that I updated in the current cycle

All right, that is fine.  But I do encourage you to update it; like I said, it is somewhat unusual for the doc and source branches to be different.

clalancette (commented):
Sorry for the long delay in getting back to you here.  What this PR is doing is very complicated, and hard to follow.

In particular, while I understand that you are splitting this up, I'm also not sure how this is going to practically work for doing releases.  Since we are using the same source repository and the same release repository for `cob_common` and `cob_common_legacy`, how is a `bloom-release` on either of them supposed to work here?  If you do `bloom-release cob_common`, then it will release `cob_actions`, `cob_common`, `cob_msgs`, and `cob_srvs` to the new version (let's call it 0.7.12).  If you then do `bloom-release cob_common_legacy`, it will release `cob_actions`, `cob_common`, `cob_description`, `cob_msgs`, `cob_srvs`, and `raw_description`.  But that is duplicating what was done above, so this will always be out-of-sync.

Can you maybe explain how you intend to make this work?

fmessmer:
> Sorry for the long delay in getting back to you here. What this PR is doing is very complicated, and hard to follow.
> 
> In particular, while I understand that you are splitting this up, I'm also not sure how this is going to practically work for doing releases. Since we are using the same source repository and the same release repository for `cob_common` and `cob_common_legacy`, how is a `bloom-release` on either of them supposed to work here? If you do `bloom-release cob_common`, then it will release `cob_actions`, `cob_common`, `cob_msgs`, and `cob_srvs` to the new version (let's call it 0.7.12). If you then do `bloom-release cob_common_legacy`, it will release `cob_actions`, `cob_common`, `cob_description`, `cob_msgs`, `cob_srvs`, and `raw_description`. But that is duplicating what was done above, so this will always be out-of-sync.
> 
> Can you maybe explain how you intend to make this work?

I only intend to continue supporting `bloom-release cob_common` (i.e. for `cob_actions`, `cob_common`, `cob_msgs` and `cob_srvs`), i.e. updating the `rosdistro` entry for `cob_common`
I do not intend to edit the `rosdistro` entry for `cob_common_legacy` at all anymore (maybe except for the "doc vs. source" consistency change) - other than that, `cob_common_legacy` has not been released via `bloom-release` - I manually added this key...there will also not be any `bloom-release cob_common_legacy` in the future - I just want to keep the debian packages available in the pinned (legacy) version

clalancette:
> I only intend to continue supporting `bloom-release cob_common` (i.e. for `cob_actions`, `cob_common`, `cob_msgs` and `cob_srvs`), i.e. updating the `rosdistro` entry for `cob_common` I do not intend to edit the `rosdistro` entry for `cob_common_legacy` at all anymore (maybe except for the "doc vs. source" consistency change) - other than that, `cob_common_legacy` has not been released via `bloom-release` - I manually added this key...there will also not be any `bloom-release cob_common_legacy` in the future - I just want to keep the debian packages available in the pinned (legacy) version

I think the problem with this scheme is that it is tricky and different from all other releases in ROS, which is why we are balking at this.  If you are no longer able or willing to maintain this repository in the future, it will be very difficult for anyone else to pick this up and understand what is happening here.

One alternative that I might suggest is to actually create a `cob_common_legacy` source and release repository, and only put in the `raw_description` and `cob_description` packages in it.  That way it will look and behave the same way as all other repositories.  What do you think?

fmessmer:
I dont quite understand what is problematic about this. The two entries are completely separated by:
- disjunct package sets
- different release branch name
- different source branch name
- different tags

arent the branches in the release repo also package/branch/tag specific?

I would like to avoid moving the legacy packages to a separate repo!

fmessmer:
If you say this is too complicated, I will rather remove the legacy entries again

But then, I will also have to remove all downstream packages depending on those (then removed) legacy packages

fmessmer:
I mean, eventually, we at 4am-robotics, do not require these repos to be updated/released anymore...

We could also just not update/release *any* package of the cob-repos at all and leave everything in it current state, i.e. close the current PRs.

I just thought to keep those for the sake of nostalgy 🙂

clalancette:
> what would/could be potential (negative) consequences in the buildfarm/release workflow? is there a way to test/simulate the consequences in some way?

The basic issue is that it is not possible for anyone to do releases again for this repository, without it end up being a confusing mess.  And while I totally understand that you may not intend to do that, reality sometimes interferes and we need to do releases.

> I dont quite understand what is problematic about this. The two entries are completely separated by:
> 
>     * disjunct package sets
> 
>     * different release branch name
> 
>     * different source branch name

While they do have different source branch names, they have the same packages on multiple branches.  And that will cause problems in the release repository.  In particular, `bloom-release` will succeed with blooming either of them, but they will be "competing" to see which branch releases the shared packages.

I guess one other way to move forward here would be to remove all the duplicate packages on the `kinetic_release_candidate` branch.  That would fix most of my concerns here (though not all of them; this would still be a special snowflake repository).

fmessmer:
I thought about this some more...
I guess releasing the latest functional changes made to the (still maintained) packages is not worth the hustle of messing with the long-term stable releases of the cob_* series - at least there is no benefit for 4am-robotics but rather negative consequences for the buildfarm and other potential downstream dependencies

Thus, I guess it's ok then to just keep the last release (from Feb 2024) that still included all packages from the "old" source branch and just not trigger a release from the new dev branch - where now some packages have been dropped

fmessmer:
closing in favor of https://github.com/ros/rosdistro/pull/41016

PR #40578, opened 2024-04-14
Rename gc_spl to game_controller_spl

ijnek:
Repository was renamed recently from gc_spl to game_controller_spl. Instead of updating the name just for rolling's distribution.yaml, for consistency, I updated the names in humble and iron's distribution files too. I'm hoping there are no issues with this manual change.

* [Source repo](https://github.com/ros-sports/game_controller_spl)
* [Releas repo](https://github.com/ros2-gbp/game_controller_spl-release)

[Related issue in ros2-gbp](https://github.com/ros2-gbp/ros2-gbp-github-org/issues/430)

PR #40707, opened 2024-04-20
rename gc_spl to game_controller_spl in rolling

ijnek:
As requested in #40578, splitting up the changes into PR per distro.

> Repository was renamed recently from gc_spl to game_controller_spl. Instead of updating the name just for rolling's distribution.yaml, for consistency, I updated the names in humble and iron's distribution files too. I'm hoping there are no issues with this manual change.
>
> * [Source repo](https://github.com/ros-sports/game_controller_spl)
> * [Releas repo](https://github.com/ros2-gbp/game_controller_spl-release)
>
> [Related issue in ros2-gbp](https://github.com/ros2-gbp/ros2-gbp-github-org/issues/430)

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1711,7 +1711,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/game_controller_spl-release.git
-      version: 4.0.0-2
+      version: 4.0.1-1
     source:
       type: git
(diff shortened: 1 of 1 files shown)


clalancette:
This one will need to be rebased to fix the conflict.  After that, we can see what CI doesn't like about it.

ijnek:
Rebased this PR. The diff isn't great, but should be correct.

ijnek:
The CI failure was due to missing tags. I've updated the tags to match the latest ones on the new release repo.

PR #40709, opened 2024-04-20
rename gc_spl to game_controller_spl in humble

ijnek:
As requested in #40578, splitting up the changes into PR per distro.

> Repository was renamed recently from gc_spl to game_controller_spl. Instead of updating the name just for rolling's distribution.yaml, for consistency, I updated the names in humble and iron's distribution files too. I'm hoping there are no issues with this manual change.
>
> * [Source repo](https://github.com/ros-sports/game_controller_spl)
> * [Releas repo](https://github.com/ros2-gbp/game_controller_spl-release)
>
> [Related issue in ros2-gbp](https://github.com/ros2-gbp/ros2-gbp-github-org/issues/430)

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2176,6 +2176,31 @@ repositories:
       url: https://github.com/foxglove/schemas.git
       version: main
     status: developed
+  game_controller_spl:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Verified that the old and the new repository redirect to the same place, so this is the same codebase.

PR #40747, opened 2024-04-22
Add dm-env

peterdavidfagan:

## Package name:

dm-env

## Package Upstream Source:
[Source Code](https://github.com/google-deepmind/dm_env)
[PyPI](https://pypi.org/project/dm-env/)

## Purpose of using this:

For integrating ROS applications with dm-env API interface for modelling environments as a Markov Decision Process.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5575,6 +5575,10 @@ python3-dlib-pip:
     pip:
       depends: [cmake, build-essential, python3-dev]
       packages: [dlib]
+python3-dm-env-pip:
+  '*':
+    pip:
+      packages: [dm-env]
 python3-do-mpc-pip:
   debian:
     pip:


rosdep/python.yaml
```diff
@@ -173,6 +173,10 @@ cython3:
   gentoo: [dev-python/cython]
   nixos: [python3Packages.cython]
   ubuntu: [cython3]
+dm-env-pip:
```
sloretz, line 176: Please name the key `python3-dm-env-pip`: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1

sloretz (dismissed):
Please add an entry for `debian`



( stealing @mjcarroll's disclaimer :)

rosdep/python.yaml
peterdavidfagan: Resolved in [e14276c](https://github.com/ros/rosdistro/pull/40747/commits/e14276cb6e5120e62dc6f5935af7f2525d4cf78e).

peterdavidfagan:
It looks like there is no system package release for `dm-env` at the moment. I have updated with the python3 prefix as requested.

rosdep/python.yaml
```diff
@@ -173,6 +173,10 @@ cython3:
   gentoo: [dev-python/cython]
   nixos: [python3Packages.cython]
   ubuntu: [cython3]
+python3-dm-env-pip:
+  ubuntu:
+    pip:
+      packages: [dm-env]
```
sloretz, line 179:
```suggestion
python3-dm-env-pip:
  debian:
    pip:
      packages: [dm-env]
  ubuntu:
    pip:
      packages: [dm-env]
```

peterdavidfagan:
Apologies @sloretz I misunderstood adding entry for debian, added your commit suggestion for this.

clalancette:
Due to reasons not directly involved with this PR, CI is failing.  If you rebase this onto the latest, it should start succeeding.  Sorry for the extra work.

peterdavidfagan:
No worries, I just rebased and squashed some of the commits, hopefully CI passes following this change.

peterdavidfagan:
@clalancette @sloretz I am going the route of adding a virtualenv definition that ships with my workspace in order to handle Python dependencies. I think this might be cleaner for my use case as there are a number of dependencies I would need to add to rosdep. 

What are your thoughts on this? The Python community already has a nice set of infrastructure for managing dependencies that a lot of developers are already familiar with.

clalancette:
Arg.  Again due to changes outside of your PR, this needs to be rebased.  I'm really sorry about that.

> @clalancette @sloretz I am going the route of adding a virtual environment definition that ships with my workspace + dockerized applications in order to handle Python dependencies (using poetry). I think this might be cleaner for my use case as there are a number of dependencies I would need to add to rosdep. Of course this does mean that I cannot easily package my python ROS 2 packages for the ROS 2 ecosystem :( . I'll look to get time in future to add other packages to rosdep.

Right, that's the main downside to doing this all in a virtualenv; you won't be able to create ROS 2 packages for it easily.  Also, at present, ROS 2 doesn't integrate well with virtualenvs (though it is possible, and I have some PRs in the wings that will improve the situation).

Whether those tradeoffs are worth it for your deployment model is hard to say.

peterdavidfagan:
> Arg. Again due to changes outside of your PR, this needs to be rebased. I'm really sorry about that.

No worries I'll rebase again.

> Right, that's the main downside to doing this all in a virtualenv; you won't be able to create ROS 2 packages for it easily. Also, at present, ROS 2 doesn't integrate well with virtualenvs (though it is possible, and I have some PRs in the wings that will improve the situation). Whether those tradeoffs are worth it for your deployment model is hard to say.

Makes sense, one thing I would add is many research projects in python don't create packages but provided config files for setting up a local virtual environment. Right now its possible for me to release these types of files and provide users with instructions which may be enough for sharing research code. 

Anything that is written properly and more production grade would of course require proper packaging. I would also prefer to properly clean and package my software, in the short-term I will just release virtual env config files and later update any sets of software that are reused regularly as packages.

rosdep/python.yaml
```diff
@@ -173,6 +173,10 @@ cython3:
   gentoo: [dev-python/cython]
   nixos: [python3Packages.cython]
   ubuntu: [cython3]
+dm-env-pip:
+  ubuntu:
+    pip:
+      packages: [dm-env]
```
clalancette, line 179: Actually, sorry again.  A couple of things here:

1.  Since this is a pip-package, it should be `python3-dm-env-pip`.
2.  We haven't added this to the CONTRIBUTING.md yet, but we want to switch to wildcards for pip packages going forward.

So this whole thing would be:

```suggestion
python3-dm-env-pip:
  '*':
    pip:
      packages: [dm-env]
```

(don't forget to realphabetize once you change the key name)

clalancette (changes requested):

rosdep/python.yaml
peterdavidfagan: [ef51404](https://github.com/ros/rosdistro/pull/40747/commits/ef514049c134e621ebfad68b320926cfb1a45fa4)

peterdavidfagan:
Can I ping someone when I plan to do a batch of these?

clalancette:
> Can I ping someone when I plan to do a batch of these, or tag someone on GitHub?

We have a rotating schedule on who is assigned to this duty, so it isn't easy to point you to someone in particular.  I'd just open them up as you have them done, and we'll get to them.

peterdavidfagan:
> We have a rotating schedule on who is assigned to this duty, so it isn't easy to point you to someone in particular. I'd just open them up as you have them done, and we'll get to them.

Ok makes sense, I'll look to open prs once I get around to packaging everything. If I do a batch I'll include a message for this in the pr descriptions as it might be easier to batch them for reviews too.

peterdavidfagan:
image

clalancette (approved):

PR #40764, opened 2024-04-22
add openeb package to humble, iron, rolling

berndpfrommer:
# Please add openeb package to be indexed in the rosdistro.

rosdistros: humble, iron, rolling

# The source is here:

https://github.com/ros-event-camera/openeb

# Comment

The OpenEB library is the SDK for event based cameras based on Prophesee chips.

The metavision_driver package depends on this library. Currently I'm downloading, building and installing it in the metavision_driver package using FetchContent, but something is broken now on rolling/noble and I find it really hard to debug FetchContent. A separate package is also the cleaner solution.

Please add this package to rosdistro, thanks!

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -4937,6 +4937,16 @@ repositories:
         release: release/humble/{package}/{version}
       url: https://github.com/ros2-gbp/ompl-release.git
       version: 1.6.0-1
+  openeb:
+    doc:
+      type: git
+      url: https://github.com/ros-event-camera/openeb.git
+      version: ros2-latest
+    source:
+      type: git
+      url: https://github.com/ros-event-camera/openeb.git
+      version: ros2-latest
+    status: maintained
   openni2_camera:
     doc:
       type: git
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -4034,6 +4034,16 @@ repositories:
         release: release/iron/{package}/{version}
       url: https://github.com/ros2-gbp/ompl-release.git
       version: 1.5.2-4
+  openeb:
+    doc:
+      type: git
+      url: https://github.com/ros-event-camera/openeb.git
+      version: ros2-latest
+    source:
+      type: git
+      url: https://github.com/ros-event-camera/openeb.git
+      version: ros2-latest
+    status: maintained
   openni2_camera:
     doc:
       type: git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -3939,6 +3939,16 @@ repositories:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/ompl-release.git
       version: 1.5.2-4
+  openeb:
+    doc:
+      type: git
+      url: https://github.com/ros-event-camera/openeb.git
+      version: ros2-latest
+    source:
+      type: git
+      url: https://github.com/ros-event-camera/openeb.git
+      version: ros2-latest
+    status: maintained
   openni2_camera:
     doc:
       type: git


nuclearsandwich (changes requested):
> * [x]  All packages have a declared license in the package.xml

I don't actually see any package.xml files in the linked repository. Am I looking at the wrong branch?

***Edit*** *I was looking at the wrong branch.* The rest of the comments still stand however, I think it's better to package this as a vendor package rather than make a release of forked sources unless you intend the fork to be a permanent hard fork specific to the ROS ecosystem.

---

This appears to be an external package that is not specifically maintained for use in ROS.
My general recommendation is to propose such packages for upstreaming into Debian and Ubuntu so that they can be installed via rosdep rather than packaged on the ROS build farm.
My specific recommendation is to create a vendor package using `ament_cmake_vendor_package` https://github.com/ament/ament_cmake/tree/rolling/ament_cmake_vendor_package although this does not actually avoid fetchcontent as you're trying to, it does use the ROS 2 standard method for vendoring packages that are not themselves ROS based.
If you don't want to go that route in the long term, I'd suggest offering to contribute a package.xml file to the upstream project so it can be released into ROS as-is.

berndpfrommer:
Oh no ... I was so hoping this would be merged ... another week to hear back?
The default branch is set to "master", but the commit in this PR actually specifies the branch to be "ros2-latest". And "ros2-latest" does have a package.xml
I changed the default branch on the repo, and will update the link in the original comment above.

nuclearsandwich:
> Oh no ... I was so hoping this would be merged ... another week to hear back?
> ...
> The process is glacial. It's getting to the point that I have to use software to track open pull requests. I can't take it anymore.

I hear you here. The ROS core team is principally focused on the Jazzy release and the release process has gotten long in the tooth. Reducing the number of active steps for package contributors and maintainers alike is on my backlog.

> Please, please don't make me use ament_cmake_vendor_package. I spent so much time to get this working as is.

If you're trying to avoid "hiccups" that delay the review process, you would be better served in doing so by adhering to the standard conventions for packaging third-party software. My recommendation is to use ament_cmake_vendor_package, or at the very least, open a PR contributing your changes upstream so that your fork doesn't need to be maintained indefinitely. 

---

For fellow reviewers If another rosdistro reviewer approves this PR as-is, there is no need to block on getting my approval.

clalancette:
> If you're trying to avoid "hiccups" that delay the review process, you would be better served in doing so by adhering to the standard conventions for packaging third-party software. My recommendation is to use ament_cmake_vendor_package, or at the very least, open a PR contributing your changes upstream so that your fork doesn't need to be maintained indefinitely.

Yes, I this is the best path forward.  We don't currently have a formal written policy about this, but we are now working on one, and this will be part of it.

berndpfrommer:
I got the vendor package to work on humble and iron, but it fails to find the ``ament_cmake_vendor_package`` on rolling:
```
   Could not find a package configuration file provided by
    "ament_cmake_vendor_package" with any of the following names:
  
      ament_cmake_vendor_packageConfig.cmake
      ament_cmake_vendor_package-config.cmake
```

Although there is a dependency for it in ``package.xml``:
```
<buildtool_depend>ament_cmake</buildtool_depend>
<buildtool_depend>ament_cmake_vendor_package</buildtool_depend>
```

I'm using the github actions: setup-ros@0.7.4   and ros-ci@0.3.11
[Here](https://github.com/ros-event-camera/openeb_vendor/actions/runs/9001142037/job/24726739393) is a link to the failed build log. I don't ever see ``ros-rolling-ament-cmake-vendor-package`` being installed in the logs, but neither do I see that in the successful  builds on iron and humble.

clalancette:
> [Here](https://github.com/ros-event-camera/openeb_vendor/actions/runs/9001142037/job/24726739393) is a link to the failed build log. I don't ever see `ros-rolling-ament-cmake-vendor-package` being installed in the logs, but neither do I see that in the successful builds on iron and humble.

Yeah, it looks like it is never installed, which is why it can't be found.

The real problem here is that that action is targeting jammy, not noble.  You'll need to update that to noble for this to have any chance of working.

berndpfrommer:
Thanks for the pointer! I learned that the github runner needs a noble container image. Now everything builds successfully in CI.

I decided to not vendor the library for now, mostly because ament_cmake_vendor_package is not available on older ros distros and I don't want to fork my source tree.

I do have another package already in rosdistro (libcaer) that I will convert to use ament_cmake_vendor.

Closing this PR.

PR #40772, opened 2024-04-23
Add libgtkmm3, a C++ binder for GTK3

5730289021-NN:

## Package name:

libgtkmm3

## Package Upstream Source:

https://gitlab.gnome.org/GNOME/gtkmm.git

## Purpose of using this:
- gtkmm is a C++ binder, which ease C++ developer by not calling C API (from GTK) directly.
- gtkmm is already presented here but only version 2.X is available as `libgtkmm`.
- To prevent breaking change, I think the name `libgtkmm3` is a proper one.

Distro packaging links:

  - https://packages.debian.org/buster/libgtkmm-3.0-dev
  - https://packages.ubuntu.com/focal/libgtkmm-3.0-dev
  - https://packages.fedoraproject.org/pkgs/gtkmm3.0/gtkmm3.0/fedora-38.html
  - https://archlinux.org/packages/extra/x86_64/gtkmm3/
  - https://packages.gentoo.org/packages/dev-cpp/gtkmm
  - IF AVAILABLE
  - https://pkgs.alpinelinux.org/package/edge/community/x86_64/gtkmm3
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3889,6 +3889,12 @@ libgtkmm:
   fedora: [gtkmm24]
   gentoo: [dev-cpp/gtkmm]
   ubuntu: [libgtkmm-2.4-dev]
+libgtkmm3:
+  arch: [gtkmm3]
+  debian: [libgtkmm-3.0-dev]
+  fedora: [gtkmm30]
+  gentoo: [dev-cpp/gtkmm:3.0]
+  ubuntu: [libgtkmm-3.0-dev]
 libgts:
   arch: [gts]
   debian: [libgts-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -3890,6 +3890,7 @@ libgtkmm:
   gentoo: [dev-cpp/gtkmm]
   ubuntu: [libgtkmm-2.4-dev]
 libgtkmm3:
+  alpine: [gtkmm3]
   arch: [gtkmm3]
   debian: [libgtkmm-3.0-dev]
   fedora: [gtkmm30]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -3889,6 +3889,13 @@ libgtkmm:
   fedora: [gtkmm24]
   gentoo: [dev-cpp/gtkmm]
   ubuntu: [libgtkmm-2.4-dev]
+libgtkmm3:
+  alpine: [gtkmm3]
+  arch: [gtkmm3]
+  debian: [libgtkmm-3.0-dev]
+  fedora: [gtkmm30]
+  gentoo: [dev-cpp/gtkmm:3.0]
+  ubuntu: [libgtkmm-3.0-dev]
```
clalancette, line 3898: So this key seems to be mixing the devel and runtime libraries in some of the distributions.  Also, we generally prefer to have the key name match the Ubuntu package name.  So I think this should end up being:

```suggestion
libgtkmm3.0-dev:
  alpine: [gtkmm3-dev]
  arch: [gtkmm3]
  debian: [libgtkmm-3.0-dev]
  fedora: [gtkmm30-devel]
  gentoo: [dev-cpp/gtkmm:3.0]
  rhel: [gtkmm30-devel]
  ubuntu: [libgtkmm-3.0-dev]
```

clalancette (changes requested):

rosdep/base.yaml
5730289021-NN: I agree to your suggestions.

clalancette (approved):

nuclearsandwich (approved):

PR #40800, opened 2024-04-24
Add lidar_s_graphs to humble and iron distributions

hridaybavle:
## Package names:

  * [lidar_s_graphs](https://github.com/snt-arg/lidar_s_graphs.git)
  
## Package Upstream Source:

https://github.com/snt-arg/lidar_s_graphs.git

## Purpose of using this:

Repository to generate in real-time situational graphs for robot pose and high-Level map Optimization using 3D LiDAR data. Apart from the main package there also dependencies of `lidar_s_graphs` in the which will be uploaded to `distribution.yaml` in seperate PRs. The scientific paper can be found [here](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=10168233).

# Please Add This Package to be indexed in the rosdistro.

Iron and Humble

# The source is here:

https://github.com/snt-arg/lidar_s_graphs.git

# Checks
 - [ x] All packages have a declared license in the package.xml
 - [ x] This repository has a LICENSE file
 - [ x] This package is expected to build on the submitted rosdistro

PR #40804, opened 2024-04-24
Add graph_reasoning to rosdistro

hridaybavle:
## Package names:

  * [graph_reasoning](https://github.com/snt-arg/graph_reasoning.git)
  
## Package Upstream Source:

https://github.com/snt-arg/graph_reasoning.git

## Purpose of using this:

Reasoning on the low-level graph (i.e planes) generated by [lidar_s_graphs](https://github.com/snt-arg/s_graphs_msgs.git) for the automatic generation of high-level entities likes rooms, walls etc. 

PR related to `lidar_s_graphs` ---> https://github.com/ros/rosdistro/pull/40800

# Please Add This Package to be indexed in the rosdistro.

Iron and Humble

# The source is here:

https://github.com/snt-arg/graph_reasoning.git

# Checks
 - [ x] All packages have a declared license in the package.xml
 - [ x] This repository has a LICENSE file
 - [ x] This package is expected to build on the submitted rosdistro

PR #40809, opened 2024-04-24
Feature/add python distro

hridaybavle:
Please add the following python packages to `python.yaml`. 

## Package names:

  * [tensorboard](https://pypi.org/project/tensorboard/)
  * [pytorch-geometric](https://pypi.org/project/torch-geometric/)
  
## Purpose of using this:

pytorch-geometric required as dependencies for performing graph manipulations for [lidar_s_graphs](https://github.com/snt-arg/s_graphs_msgs.git). tensorboard required for visualization of the training process.

PR related to `lidar_s_graphs` ---> https://github.com/ros/rosdistro/pull/40800

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8337,7 +8337,6 @@ python3-pytest-xdist:
     "*": [python3-pytest-xdist]
     "7": ['python%{python3_pkgversion}-pytest-xdist']
   ubuntu: [python3-pytest-xdist]
-python3-pytorch-pip: *migrate_eol_2025_04_30_python3_pytorch_pip
 python3-pytorch-geometric-pip:
   debian:
     pip:
@@ -8345,6 +8344,7 @@ python3-pytorch-geometric-pip:
   ubuntu:
     pip:
       packages: [torch-geometric]
+python3-pytorch-pip: *migrate_eol_2025_04_30_python3_pytorch_pip
 python3-pytrinamic-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
I will note that `torch-geometric` has also been packaged in Debian: https://packages.debian.org/sid/python3-torch-geometric .  That would usually mean that we would prefer a Debian package key, rather than a `pip` key.  However, that package is only available in Debian sid, and not at all in Ubuntu, so in this case I'm OK with adding it in as a `pip` key.

clalancette:
Arg.  Due to reasons completely not having to do with this PR, it needs a rebase/merge with the latest on the `master` branch.  If you can do that, then CI should succeed.  Sorry for the inconvenience.

hridaybavle:
Hello @clalancette, I did the rebase with the master. Can you confirm if alls good.

clalancette (approved):

nuclearsandwich (approved):
We can use OS-level wildcards for these pip dependencies.

nuclearsandwich:
@hridaybavle so you are aware, packages that use pip-based dependencies are not eligible for release on the ROS build farm.

hridaybavle:
@nuclearsandwich thank you for the heads up. I am trying to do a ros2 release for our 
package (lidar_s_graphs)[https://github.com/ros/rosdistro/pull/40800], it depends on another ros2 python package (graph_reasoning)[https://github.com/ros/rosdistro/pull/40804] which needs the torch-geometric pip. What could be the solution you suggest in this case?

PR #40823, opened 2024-04-24
Fix graphicsmagick dep in gentoo

gentoo90:
If the virtual pulls `imagemagick` then build of [`nav2_map_server`](https://index.ros.org/p/nav2_map_server/#humble) fails

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1595,7 +1595,7 @@ graphicsmagick:
   arch: [graphicsmagick]
   debian: [libgraphicsmagick++1-dev, graphicsmagick-libmagick-dev-compat]
   fedora: [GraphicsMagick-c++-devel]
-  gentoo: [virtual/imagemagick-tools]
+  gentoo: [media-gfx/graphicsmagick]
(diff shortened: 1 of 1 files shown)


clalancette (approved):
I don't really know much about gentoo, but this seems reasonable to me based on: https://packages.gentoo.org/packages/media-gfx/graphicsmagick

gentoo90:
Also there is [a comment in `virtual/imagemagick-tools`](https://gitweb.gentoo.org/repo/gentoo.git/tree/virtual/imagemagick-tools/imagemagick-tools-0-r1.ebuild?id=031f524e4a368e419556867c49a2ee76af8fd249#n11) about that:
```bash
# This virtual is to be used **ONLY** for depending on the runtime
# tools of imagemagick/graphicsmagick. It should and cannot be used
# for linking against, as subslots are not transitively passed on.
# For linking, you will need to depend on the respective libraries
# in all consuming ebuilds and use appropriate sub-slot operators.
# See also: https://bugs.gentoo.org/314431
```

adityapande-1995 (approved):

PR #40824, opened 2024-04-24
add source entry for new package conditional_substitution

ottojo:

This package provides a single utility function for choosing substitutions in a launch file depending on a condition.

The source is here: https://github.com/ottojo/ros_conditional_substitution

# Checks

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -823,6 +823,17 @@ repositories:
       url: https://github.com/ros2/common_interfaces.git
       version: iron
     status: maintained
+  conditional_substitution:
+    doc:
+      type: git
+      url: https://github.com/ottojo/ros_conditional_substitution.git
+      version: main
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/ottojo/ros_conditional_substitution.git
+      version: main
+    status: maintained
   console_bridge_vendor:
     doc:
       type: git


clalancette:

You can copy-paste the below into your review comment when reviewing a new package addition into rosdistro.

  - [x] Top level license file: https://github.com/ottojo/ros_conditional_substitution/blob/43819308a242fd3cfce9b28acf04fb885723cf4f/LICENSE
- [x] Public source repo: https://github.com/ottojo/ros_conditional_substitution
- [x] Source repository contains ROS packages (though https://github.com/ottojo/ros_conditional_substitution/blob/43819308a242fd3cfce9b28acf04fb885723cf4f/setup.py#L18 could be updated)
- [ ] Each package meets [REP-144](https://www.ros.org/reps/rep-0144.html) naming conventions: I'm not sure about this one.  The name here is really generic, so I think violates [Global Guidelines bullet 1 and possibly bullet 4](https://www.ros.org/reps/rep-0144.html#global-guidelines).  Also, can you please explain how this differs from https://github.com/ros2/launch/pull/734 ?

ottojo:
It does not differ at all from https://github.com/ros2/launch/pull/734, which is why i opened https://github.com/ros2/launch/pull/776 and converted this back to draft. I only opened this because i didn't know about the implementation in launch, which so far is only available in rolling i think, and which i only discovered just after opening this :upside_down_face:

clalancette:
All right.  In that case, I'm going to go ahead and close this since we can get the same functionality from that PR.

PR #40861, opened 2024-04-26
Add missing nixos rules

bjsowa:
Added some nixos rules to fix generating nix derivations for rolling packages.

The `flake8-builtins`, `flake8-comprehensions` and `flake8-quotes` python packages are not available for nixos but they prevent generating derivation for `ament_flake8` package which is a dependency of many essential packages (e.g. rclpy). That's why I decided to ignore them for now. `ament_flake8` works without these packages as they only add more checks to flake8 tool.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4238,9 +4238,9 @@ liblmdb-dev:
 liblttng-ctl-dev:
   debian: [liblttng-ctl-dev]
   fedora: [lttng-tools-devel]
+  nixos: [lttng-tools]
   openembedded: [lttng-tools@openembedded-core]
   opensuse: [lttng-tools-devel]
-  nixos: [lttng-tools]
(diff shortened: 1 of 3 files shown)


wentasah:
LGTM. This increases the number of successfully generated packages in nix-ros-overlay for rolling from 1118 to 1138.

nuclearsandwich:
> This increases the number of successfully generated packages in nix-ros-overlay for rolling from 1118 to 1138

Good for an [oblique reference](https://en.wikipedia.org/wiki/THX_1138) pre-[Star Wars Day](https://en.wikipedia.org/wiki/Star_Wars_Day)

nuclearsandwich (approved):

nuclearsandwich:
Requires the second rosdep reviewer approval before merging.

clalancette (approved):

PR #41004, opened 2024-04-29
[rosdep] [catkin-lint] [debian/ubuntu] Use catkin-lint instead of python3-catkin-lint package

mikaelarguedas:
It looks like python3-catkin-tools was a name used in the past but not used anymore, and catkin-lint is used intead on all currently active distros.

## Ubuntu
$ rmadison python3-catkin-lint
 python3-catkin-lint | 1.6.6-1 | focal/universe | all

$ rmadison catkin-lint
 catkin-lint | 1.6.6-1  | focal/universe  | all
 catkin-lint | 1.6.16-1 | jammy/universe  | all
 catkin-lint | 1.6.22-1 | mantic/universe | all
 catkin-lint | 1.6.22-1 | noble/universe  | all

## Debian

$ rmadison python3-catkin-lint --url debian
python3-catkin-lint | 1.6.0-1       | oldoldstable | all
python3-catkin-lint | 1.6.12-1      | oldstable    | all

$ rmadison catkin-lint --url debian
catkin-lint | 1.6.12-1      | oldstable  | all
catkin-lint | 1.6.22-1      | stable     | all
catkin-lint | 1.6.22-1      | testing    | all
catkin-lint | 1.6.22-1      | unstable   | all

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5178,7 +5178,7 @@ python3-catkin-lint:
   openembedded: [python3-catkin-lint@meta-ros-common]
   rhel: [python3-catkin_lint]
   ubuntu:
-    '*': [python3-catkin-lint]
+    '*': [catkin-lint]
     bionic: null
 python3-catkin-pkg:
   alpine: [py3-catkin-pkg]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5172,7 +5172,7 @@ python3-cantools-pip:
       packages: [cantools]
 python3-catkin-lint:
   debian:
-    '*': [python3-catkin-lint]
+    '*': [catkin-lint]
     stretch: null
   fedora: [python3-catkin_lint]
   openembedded: [python3-catkin-lint@meta-ros-common]


clalancette (approved):

mikaelarguedas:
> Alternative question: Should we be distributing catkin-lint for more platforms from the bootstrap repository? We do [import it for bionic and focal](https://github.com/ros-infrastructure/reprepro-updater/blob/master/config/catkin_lint.upstream.yaml).

TL;DR my feeling is that it wouldn't hurt to import it, could be nice but not necessary anymore.


IIRC For a while it was necessary to import it as it was not in the official Ubuntu repos and then the package was moving faster than Ubuntu distros so having it in the ROS repos allowed to benefit from more recent versions.
Looks like now the package is stable (last commit from Jan 2023).

nuclearsandwich:
> > Alternative question: Should we be distributing catkin-lint for more platforms from the bootstrap repository? We do [import it for bionic and focal](https://github.com/ros-infrastructure/reprepro-updater/blob/master/config/catkin_lint.upstream.yaml).

> TL;DR my feeling is that it wouldn't hurt to import it, could be nice but not necessary anymore.

> For a while it was necessary to import it as it was not in the official Ubuntu repos and then the package was moving faster than Ubuntu distros so having it in the ROS repos allowed to benefit from more recent versions.

> Looks like now the package is stable (last commit from Jan 2023).

If the package was completely unavailable for our target platforms, as it was back in the day, then I think it would make sense to import and provide it via the ROS bootstrap repository. But now that it's available upstream and the upstream version is generally suitable. I'm very happy to have one less package, especially one that isn't currently maintained by members of the OSRF Infrastructure or ROS core teams, in the bootstrap repository's scope.

nuclearsandwich (approved):

PR #41032, opened 2024-05-01
Remove perception_open3d: no libopen3d-dev in Noble

SteveMacenski:
This removes `perception_open3d` from Jazzy and Rolling since `libopen3d` is not avilable on Noble (yet?). A ticket was filed with the maintainers https://github.com/isl-org/Open3D/issues/6730 and we're still waiting on a resolution about what we want to do about that. Until then, no reason to waste the build time or setup a vendor for a hopefully transient situation.

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -4136,6 +4136,20 @@ repositories:
       url: https://github.com/ros-perception/pcl_msgs.git
       version: ros2
     status: maintained
+  perception_open3d:
+    release:
+      packages:
+      - open3d_conversions
+      tags:
+        release: release/jazzy/{package}/{version}
+      url: https://github.com/ros2-gbp/perception_open3d-release.git
+    source:
+      test_commits: false
+      test_pull_requests: false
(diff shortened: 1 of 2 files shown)


nuclearsandwich (approved):

nuclearsandwich:
@SteveMacenski I'm not merging this immediately because I think it would also be acceptable make a smaller change that leaves this package indexed in the meantime:

1. Removing the `version` field from the `release` stanza would disable the binary packaging jobs
2. Setting `test_commits: false` and `test_pull_requests: false` in the `source` stanza would stop dev and PR jobs from running.

This also will avoid having to explain that this package is not "new" when it comes time to re-instate it.
If you'd prefer to pull the whole thing however that's okay by me but I wanted to recommend the less nuclear option.

--

For other reviewers. If the changes I propose are made and CI passes I'm an automatic ✅.

SteveMacenski:
Sounds good - changes reflected!

nuclearsandwich (approved):

PR #41118, opened 2024-05-09
Run `clean_rosdep_yaml.py` on rosdep YAML files

amalnanavati:
The rosdep YAML files have diverged from the formatting enforced by `scripts/clean_rosdep_yaml.py`. As a result, the script is not particularly useful since when someone runs it, it makes hundreds of lines of changes, as opposed to a targeted change on their contribution.

Thus, this PR brings the YAML formatting up-to-date by running:

- `python3 scripts/clean_rosdep_yaml.py rosdep/python.yaml rosdep/python.yaml`
- `python3 scripts/clean_rosdep_yaml.py rosdep/base.yaml rosdep/base.yaml`
- `python3 scripts/clean_rosdep_yaml.py rosdep/osx-homebrew.yaml rosdep/osx-homebrew.yaml`
- `python3 scripts/clean_rosdep_yaml.py rosdep/ruby.yaml rosdep/ruby.yaml`
- (Note: I skipped `gentoo.yaml` because formatting only removed the comment at the top that specifies that the file should not be removed and is kept for legacy purposes.)

For what it's worth, I ran that above script with `Python 3.11.3` on Mac OSx 13.2.1.


amalnanavati:
Given that the the validation fails, it seems like there is some mismatch between `clean_rosdep_yaml.py` and the validation scripts, surrounding whether to surround the gentoo packages in single quotes or not. Perhaps `clean_rosdep_yaml.py` has to be updated instead? I created an issue for it: #41119

PR #41254, opened 2024-05-19
Add PyXB-X

Ryanf55:

## Package name:

[PyXB-X](https://pypi.org/project/PyXB-X/)

## Package Upstream Source:

[PyXB-X
](https://github.com/renalreg/PyXB-X)

## Purpose of using this:

This is a pretty widely used python codegen tool for generating python classes based on XML schema. It's used in a wide variety of applications where you want to enforce the structure of XML but interract with the data objects in python in a safer way than a raw XML parser such as eTree.

Distro packaging links:

  - N/A - Pip
  - N/A - Pip

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8424,6 +8424,16 @@ python3-pyvista-pip:
   ubuntu:
     pip:
       packages: [pyvista]
+python3-pyxb-x-pip:
+  debian:
+    pip:
+      packages: [PyXB-X]
+  osx:
+    pip:
+      packages: [PyXB-X]
+  ubuntu:
+    pip:
+      packages: [PyXB-X]
 python3-pyxdg-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


mjcarroll:

rosdep/python.yaml
```diff
@@ -8424,6 +8424,16 @@ python3-pyvista-pip:
...
+python3-pyxb-x-pip:
+  debian:
+    pip:
+      packages: [PyXB-X]
+  osx:
+    pip:
+      packages: [PyXB-X]
+  ubuntu:
+    pip:
+      packages: [PyXB-X]
```
mjcarroll, line 8436: I believe this is a good place to use the wildcard.

```suggestion
python3-pyxb-x-pip:
  '*':
    pip:
      packages: [PyXB-X]
```

mjcarroll (approved):

clalancette (approved):

PR #41259, opened 2024-05-20
Mark jazzy as active.

marcoag:
Move Jazzy's distribution status from `pre-release` to `active`.


mjcarroll (approved):
:confetti_ball: :trumpet:

clalancette (approved):

PR #41305, opened 2024-05-22
add kornia pip dependency to python.yaml

Zarnack:

## Package name:

**python3-kornia-pip**

## Package Upstream Source:

https://github.com/kornia/kornia

## Purpose of using this:

This is a popular tool library to solve generic computer vision problems (e.g. image augmentation, image filtering or transformation) like OpenCV. Unlike the OpenCV functions, all of its components are differentiable, which means that they can be inserted directly into neural networks with PyTorch.

 pip: https://pypi.org/project/kornia/
 
 ## pip disclaimer
 
### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6748,6 +6748,19 @@ python3-kml2geojson-pip:
   ubuntu:
     pip:
       packages: [kml2geojson]
+python3-kornia-pip:
+  debian:
+    pip:
+      packages: [kornia]
+  fedora:
+    pip:
+      packages: [kornia]
+  osx:
+    pip:
+      packages: [kornia]
+  ubuntu:
+    pip:
+      packages: [kornia]
 python3-lark-parser:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -6748,6 +6748,19 @@ python3-kml2geojson-pip:
...
+      packages: [kornia]
+  fedora:
+    pip:
+      packages: [kornia]
+  osx:
+    pip:
+      packages: [kornia]
+  ubuntu:
+    pip:
+      packages: [kornia]
```
mjcarroll, line 6763:
```suggestion
python3-kornia-pip:
  '*':
    pip:
      packages: [kornia]
```

mjcarroll: We can use a platform wildcard here.

mjcarroll (approved):

adityapande-1995 (approved):

PR #41307, opened 2024-05-22
add supervision pip dependency to python.yaml

Zarnack:

## Package name:

**python3-supervision-pip**

## Package Upstream Source:

https://github.com/roboflow/supervision

## Purpose of using this:

A project that contains reusable model agnostic computer vision tools. It already contains several connectors for some popular image detection/classification/segmentation libraries (like Ultralytics, Transformers or MMDetection). Furthermore, it hugely simplifies annotating an image with detections made by a NN model. 

 pip: https://pypi.org/project/supervision/
 
 ## pip disclaimer
 
### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9297,16 +9297,7 @@ python3-suas-interop-clients-pip:
     pip:
       packages: [suas-interop-clients]
 python3-supervision-pip:
-  debian:
-    pip:
-      packages: [supervision]
-  fedora:
-    pip:
-      packages: [supervision]
-  osx:
-    pip:
-      packages: [supervision]
-  ubuntu:
+  '*':
     pip:
       packages: [supervision]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -9296,6 +9296,19 @@ python3-suas-interop-clients-pip:
...
+      packages: [supervision]
+  fedora:
+    pip:
+      packages: [supervision]
+  osx:
+    pip:
+      packages: [supervision]
+  ubuntu:
+    pip:
+      packages: [supervision]
```
mjcarroll, line 9311:
```suggestion
python3-supervision-pip:
  '*':
    pip:
      packages: [supervision]
```

Use platform wildcard.

nuclearsandwich (approved):

mjcarroll (approved):

PR #41381, opened 2024-05-29
Add rosdistro-pproto and rosdistro-shared repos

egorov-vn:

* rosdistro-shared
* rosdistro-pproto

# The source is here:

https://github.com/egorov-vn/rosdistro-shared.git
https://github.com/egorov-vn/rosdistro-pproto.git

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5861,6 +5861,21 @@ repositories:
       url: https://github.com/mrpt-ros-pkg/pose_cov_ops.git
       version: master
     status: maintained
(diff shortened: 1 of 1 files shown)


nuclearsandwich (changes requested):
These packages do not meet [REP-144](https://www.ros.org/reps/rep-0144.html) naming expectations.

`shared` is not an appropriate package name at all and there is not enough information in the readme to determine whether the package name `pproto` appropriately and concisely describes the package's utility.

PR #41413, opened 2024-05-30
Add cppzmq-dev to the libzmq3-dev key

azeey:
Starting in Debian Bookworm and Ubuntu Noble, the cppzmq headers have been extracted out of `libzmq3-dev` and placed in their own package, `cppzmq-dev`. This updates the `libzmq3-dev` key to include `cppzmq-dev` for Debian and Ubuntu to maintain parity with the other platforms.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6602,14 +6602,21 @@ libzip-dev:
   ubuntu: [libzip-dev]
 libzmq3-dev:
   arch: [zeromq]
-  debian: [libzmq3-dev]
+  debian:
+    '*': [libzmq3-dev, cppzmq-dev]
+    bullseye: [libzmq3-dev]
+    buster: [libzmq3-dev]
   fedora: [cppzmq-devel]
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):
We had an in-depth discussion about this and the "convenient moment" to fix this prompted by the new debian and ubuntu distributions doesn't apply to the other packages that we would retro-actively need to change and therefore would break in active distributions.

We could add dedicated keys for cppzmq-dev so that this can be separated in the future but it's not a requirement to merge this as is.

nuclearsandwich:
Although there is no direct evidence, @cottsay did assent to merging this as-is. He can disavow this if I am mistaken but I'm treating this as having received the necessary reviews.

ruffsl:
Oh boy, this bit me earlier this week while migrating to Jazzy using only rosdep to install vendored gazebo dependencies. Was really puzzled with `libzmq3-dev` already installed, but turns out I should've just update my rosdep list.

PR #41445, opened 2024-06-01
Added customtkinter pip dependency to python.yaml

alaa7alnajjar:

## Package name:

python3-customtkinter-pip

## Package Upstream Source:

https://github.com/TomSchimansky/CustomTkinter 

## Purpose of using this:

CustomTkinter is a python UI-library based on Tkinter, which provides new, modern and fully customizable widgets. They are created and used like normal Tkinter widgets and can also be used in combination with normal Tkinter elements.

- pip: https://pypi.org/project/customtkinter/

## pip disclaimer


clalancette:
@Mergifyio rebase

clalancette (approved):

audrow (approved):

PR #41512, opened 2024-06-06
Add nobleo_socketcan_bridge

Rayman:

iron

# The source is here:

~~https://github.com/nobleo/ros2_socketcan_bridge~~
https://github.com/nobleo/nobleo_socketcan_bridge

# Checks

~~@JWhitleyWork, since you are maintaining `ros2_socketcan`, are you OK with the name of this package? This is a from-scratch re-implementation of [socketcan_bridge](https://wiki.ros.org/socketcan_bridge) from ROS1. Thats why I want to call it `ros2_socketcan_bridge`~~

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -3986,6 +3986,12 @@ repositories:
       version: ros2
     status: developed
     status_description: ROS2 support is work-in-progress. Help wanted!
+  nobleo_socketcan_bridge:
+    source:
+      type: git
+      url: https://github.com/nobleo/nobleo_socketcan_bridge.git
+      version: main
+    status: maintained
   nodl:
     doc:
       type: git


sloretz (commented):
- [X] At least one of the following must be present
  - [ ] Top level license file: 
  - [X] Per package license files
- [X] License is [OSI-approved](https://opensource.org/licenses):
- [X] License correctly listed in package.xmls
- [X] Public source repo:
- [X] Source repository contains ROS packages

```console
  <name>ros2_socketcan_bridge</name>
```
</details>

```console
$ find . -name "package.xml" -exec grep --color=auto -e "<license[^>]*>" "{}" "+"
  <license>Apache-2.0</license>
```
</details>

---

The name prefix `ros2_` might be ok according to REP 144 because this is a binding of another non-ros library, but the requirement `The package name should describe what the package does, but also avoid naming collisions` is borderline because the name `ros2_socketcan_bridge` is so similar to these packages.

* [socketcan_bridge](https://index.ros.org/p/socketcan_bridge/#noetic)
* [ros2_socketcan](https://index.ros.org/p/ros2_socketcan/#rolling)
* [socketcan_interface](https://index.ros.org/p/socketcan_interface/github-ros-industrial-ros_canopen/#noetic)

I'll add this one to the next ROS 2 meeting.

clalancette:
> I'll add this one to the next ROS 2 meeting.

We ended up discussing this, and we think that there are too many packages with overlapping aims with similar names.

That said, we can take this package if it was renamed to include the organization name, as per REP-144.  In particular, something like `nobleo_socketcan` would work fine here, and make it clear that it is used specifically for Nobleo.

Rayman:
We are in the process of adding a package prefix to this package to get it accepted.

As a side note, I do think it's unfair that autoware is able to claim unprefixed package names, while I can not. The rules should be the same for every developer.

clalancette:
> As a side note, I do think it's unfair that autoware is able to claim unprefixed package names, while I can not. The rules should be the same for every developer.

We are doing our best here.  Unprefixed package names aren't a problem, per-se, but when there are several competing for basically the same name, we have to do something.  That something could be to force every single package to have a prefix, but that doesn't seem like a great solution either.  We are open to suggestions.

Rayman:
I've renamed the package to `nobleo_socketcan_bridge` and updated the PR

iron/distribution.yaml
```diff
@@ -3986,6 +3986,12 @@ repositories:
       version: ros2
     status: developed
     status_description: ROS2 support is work-in-progress. Help wanted!
+  nobleo_socketcan_bridge:
+    source:
```
Yadunund, line 3990: Do you want to add a doc entry?

Yadunund (approved):
Thanks for renaming the package!

@sloretz @clalancette do you approve of the latest name? its a LGTM from me but still holding this PR as Iron is in a sync hold

[Holding for Iron sync](https://discourse.ros.org/t/preparing-for-iron-patch-and-sync-2024-07-12/)

clalancette (approved):
This looks good to me, @Yadunund comments notwithstanding.  Thank you for renaming the package.

PR #41578, opened 2024-06-10
feat: add python3-pygnssutils-pip to python.yaml

chbroecker:

## Package name:

`python3-pygnssutils-pip`

No system level dependency is currently available.

## Package Upstream Source:

https://github.com/semuconsulting/pygnssutils

## Purpose of using this:

This dependency is being used for working with gnss in python, e.g. for communicating with an NTRIP server to receive GPS correction signals.

- PyPI: https://pypi.org/project/pygnssutils/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7986,6 +7986,10 @@ python3-pygments:
     '*': [python3-pygments]
     '7': null
   ubuntu: [python3-pygments]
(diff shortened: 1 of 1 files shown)


wjwwood (approved):

clalancette:

clalancette (approved):

PR #41583, opened 2024-06-10
Add netft_utils source for Iron

Ian118:

Iron

# The source is here:

~~https://github.com/tamu-edu/rad_lab_netft_utils/tree/iron~~
https://github.com/UTNuclearRoboticsPublic/netft_utils/tree/iron

# Checks

Adding @AdamPettinger

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -4038,7 +4038,7 @@ repositories:
   netft_utils:
     source:
       type: git
-      url: https://github.com/UTNuclearRoboticsPublic/netft_utils
+      url: https://github.com/UTNuclearRoboticsPublic/netft_utils.git
       version: iron
     status: maintained
   nlohmann_json_schema_validator_vendor:
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -4035,6 +4035,12 @@ repositories:
       url: https://github.com/nerian-vision/nerian_stereo_ros2.git
       version: default
     status: developed
+  netft_utils:
+    source:
+      type: git
+      url: https://github.com/tamu-edu/rad_lab_netft_utils.git
+      version: iron
+    status: maintained
   nlohmann_json_schema_validator_vendor:
     doc:
       type: git
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -4038,7 +4038,7 @@ repositories:
   netft_utils:
     source:
       type: git
-      url: https://github.com/tamu-edu/rad_lab_netft_utils.git
+      url: https://github.com/UTNuclearRoboticsPublic/netft_utils
       version: iron
     status: maintained
   nlohmann_json_schema_validator_vendor:


audrow:
Hi @Ian118, I see that the packages have license files, but would you add a top-level license file?

Ian118:
Yes, I have added it.

iron/distribution.yaml
```diff
@@ -3892,6 +3892,12 @@ repositories:
       url: https://github.com/nerian-vision/nerian_stereo_ros2.git
       version: default
     status: developed
+  netft_utils:
+    source:
+      type: git
+      url: https://github.com/tamu-edu/rad_lab_netft_utils.git
```
clalancette, line 3898: There are 2 main problems I can see with this repository:

1.  It is a fork of another repository.  While that doesn't automatically disqualify it from being listed here, we typically prefer to use the upstream where possible.  Can you describe the reason for the fork, and the relationship between this repository and the one it is forked from?
2.  The licensing of it is unclear.  In particular, https://github.com/tamu-edu/rad_lab_netft_utils/blob/main/netft_utils/License.txt uses similar text to a BSD 3-clause, but is not exactly the same.  Additionally, https://github.com/tamu-edu/rad_lab_netft_utils/blob/f5fd670aaf0a0ffa4054334cbc1f86ff16b99219/netft_utils/package.xml#L15 doesn't clearly state the license.

We can possibly work through that first issue, but because the license is not approved by the OSI, we actually aren't qualified to determine whether we can distribute this or not.  Is there anyway those licenses can be made standard 3-clause BSD?

clalancette (dismissed):

iron/distribution.yaml
Ian118: The reason for the fork is that my lab is maintaining the ROS2 port of the package, however we have pushed the changes to the upstream repository. I updated the pull request to reflect this change.
We have contacted the original authors of the package about the license, and I will let you know if anything changes.

RobertBlakeAnderson: Hi, I represent the authors of the upstream repo. I don't know that we can change the license since Los Alamos specified this one for work they have funded. It is, in all practical ways, identical to BSD-3 however. I could reach our to our Los Alamos contact and ask about this.

mitchpryor: I understand that this license is what Los Alamos National Labs will use for all software approved for open source release (https://www.lanl.gov/software/open-source-software.php). Would it be possible to get this license added to the approved list? They have several groups increasingly working in robotics so this may be the easier path.

clalancette: > I understand that this license is what Los Alamos National Labs will use for all software approved for open source release (https://www.lanl.gov/software/open-source-software.php). Would it be possible to get this license added to the approved list? They have several groups increasingly working in robotics so this may be the easier path.

Unfortunately, we really have no way of doing this.  We aren't lawyers, so we don't know if this license allows us to legally redistribute it.

mitchpryor: That makes sense. I have reached out to LANL to see if we can put the license through the OSI approval process (https://opensource.org/licenses/review-process). I have not done that before, but it looks like it takes 60 days.

Ian118:
We have submitted the license to the Open Source Initiative and are waiting for a decision.

iron/distribution.yaml
AdamPettinger: An update: I submitted the license to OSI and they said it should be easily approved. Now just waiting for that to be official

mjcarroll:
@Ian118 any update on the license process at OSI?

Ian118:
OSI will review the license at their following board meeting scheduled for September 20th.

AdamPettinger:
Hello all, the license has been approved, see [here](https://opensource.org/license/los-alamos-national-labs-bsd-3-variant)

mjcarroll:
Do you think that you can push another commit to this branch?  I think that the CI has gotten stale enough that I cannot re-run the actions here.

Ian118:
@mjcarroll @clalancette The repository has been pushed, could you review this change?

Yadunund (approved):

  - [x] Top level license file:
  - [x] Per package license files:
- [x] License is [OSI-approved](https://opensource.org/licenses): BSD and MIT
- [x] Public source repo: https://github.com/UTNuclearRoboticsPublic/netft_utils.git

PR #41591, opened 2024-06-11
Set clang-format key to clang-tools-extra on RHEL

christophebedard:

## Package name:

Set `clang-format` to `clang-tools-extra` on RHEL.

## Package Upstream Source:

N/A

## Purpose of using this:

`ros-jazzy-ament-clang-format` depends on `clang-format` on Ubuntu, but not on RHEL. The rosdep entry for `clang-format` on RHEL is `clang`, which does not provide `clang-format`. The RHEL package that provides `clang-format` is apparently `clang-tools-extra`: https://github.com/ros-tooling/setup-ros/pull/694#issuecomment-2159234767.

Since `clang-tidy` also maps to `clang-tools-extra` on RHEL and skips RHEL 7, I've done the same for `clang-format`.

  - https://packages.fedoraproject.org/pkgs/clang/clang-tools-extra/
  - https://pkgs.org/search/?q=clang-tools-extra

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -558,7 +558,7 @@ clang-format:
   osx:
     homebrew:
       packages: [clang-format]
-  rhel: [clang]
+  rhel: [clang-tools-extra]
   ubuntu: [clang-format]
 clang-tidy:
   alpine: [clang-extra-tools]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -552,7 +552,7 @@ clang-format:
   alpine: [clang]
   arch: [clang]
   debian: [clang-format]
-  fedora: [clang]
+  fedora: [clang-tools-extra]
   gentoo: [sys-devel/clang]
   nixos: [clang]
   osx:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -552,14 +552,14 @@ clang-format:
   alpine: [clang]
   arch: [clang]
   debian: [clang-format]
-  fedora: [clang-tools-extra]
+  fedora: [clang-tools-extra, git-clang-format]
   gentoo: [sys-devel/clang]
   nixos: [clang]
   osx:
     homebrew:
       packages: [clang-format]
   rhel:
-    '*': [clang-tools-extra]
+    '*': [clang-tools-extra, git-clang-format]
     '7': null
   ubuntu: [clang-format]
 clang-tidy:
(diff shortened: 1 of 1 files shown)


cottsay (approved):
Thanks, it looks like `fedora` is in the same boat. Mind fixing that as well?

https://packages.fedoraproject.org/pkgs/clang/clang-tools-extra/

christophebedard:
> Since clang-tidy also maps to clang-tools-extra on RHEL and skips RHEL 7, I've done the same for clang-format.

Oops, I didn't commit that part. Fixed.

> Thanks, it looks like `fedora` is in the same boat. Mind fixing that as well?
>
> https://packages.fedoraproject.org/pkgs/clang/clang-tools-extra/

Done!

It looks like it might be the same for alpine too, but I can only find packages with the `clang` version in the name, e.g., `[clang18-extra-tools](https://pkgs.alpinelinux.org/package/edge/main/x86_64/clang18-extra-tools)`: https://pkgs.alpinelinux.org/packages?name=clang*tools&branch=edge&repo=&arch=x86_64&maintainer=. Unless I'm missing something, the current value for `clang-tidy` for `alpine` is `clang-extra-tools`, which doesn't exist(?).

rosdep/base.yaml
```diff
@@ -552,13 +552,15 @@ clang-format:
   alpine: [clang]
   arch: [clang]
   debian: [clang-format]
-  fedora: [clang]
+  fedora: [clang-tools-extra]
```
clalancette, line 555: If we look at what binaries the `clang-format` tools provides on Ubuntu, it looks like it provides all of:

```
/usr/bin/clang-format
/usr/bin/clang-format-diff
/usr/bin/git-clang-format
```

On RHEL and Fedora, `clang-tools-extra` provides both `clang-format` and `clang-format-diff`, while `git-clang-format` comes in via the `git-clang-format` package.

Thus it seems to me that this should be:

```suggestion
  fedora: [clang-tools-extra, git-clang-format]
```

(the same goes for the `rhel` key below)

clalancette (changes requested):

rosdep/base.yaml
christophebedard: makes sense, done

clalancette (approved):
Looks good to me, but I'd like to get @cottsay 's approval again for the latest changes before merging.

cottsay (approved):

PR #41599, opened 2024-06-12
Add python3-dataclasses-json in rosdep/python.yaml

patrickKXMD:

## Package name:
`python3-dataclasses-json`

## Package Upstream Source:
https://github.com/lidatong/dataclasses-json

## Purpose of using this:
This dependency is used for handling json conversion of dataclasses in python nodes.

PyPI: https://pypi.org/project/dataclasses-json/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5463,6 +5463,16 @@ python3-cycler:
     '*': [python3-cycler]
     '7': null
   ubuntu: [python3-cycler]
+python3-dataclasses-json:
+  debian: [python3-dataclasses-json]
+  freebsd: [py39-dataclasses-json]
+  opensuse:
+    '*': [python310-dataclasses-json]
+    '15.2': null
+  ubuntu:
+    '*': [python3-dataclasses-json]
+    bionic: null
+    focal: null
 python3-datadog-pip:
   ubuntu:
     pip:


rosdep/python.yaml
```diff
@@ -5463,6 +5463,10 @@ python3-cycler:
     '*': [python3-cycler]
     '7': null
   ubuntu: [python3-cycler]
+python3-dataclasses-json:
```
clalancette, line 5466: So it looks like this is available on Ubuntu/Debian as a system package:

https://pkgs.org/search/?q=dataclasses-json

We prefer the system packages version (since things that depend on it can be released to the buildfarm), so please use the system packages for `arch`, `debian`, `ubuntu`, and `opensuse`.

clalancette (changes requested):

rosdep/python.yaml
patrickKXMD: @clalancette Hi, I've update this for `debian`, `freebsd`, `opensuse` and `ubuntu`, but the package in `arch` is not official, should I use it?

clalancette: > @clalancette Hi, I've update this for `debian`, `freebsd`, `opensuse` and `ubuntu`, but the package in `arch` is not official, should I use it?

No, we only list official packages in rosdep.  In that case, this looks good, thanks for updating!

clalancette (approved):

patrickKXMD:
@clalancette

clalancette:
> should I use pip in opensuse 15.2 and ubuntu focal ?

You can either do that, or set them to 'null' (as in https://github.com/ros/rosdistro/blob/e5aa984492c21934bce3c815a7faed33e16b7ed3/rosdep/python.yaml#L5900-L5910, for example).  Either will work.

clalancette (approved):

audrow (approved):

PR #41623, opened 2024-06-14
Add in liboctomap-dev key.

clalancette:
The situation with octomap and ROS is complicated.

As it stands, we effectively have a "vendor" package of octomap released into Humble, Iron, Jazzy, and Rolling. However, this is a problem on Jazzy and Rolling on Ubuntu 24.04 since the ABI between what we have released, and what is released in the operating system as "liboctomap-dev" is different.  That means that using it with downstream packages just doesn't work.

The intent here is to add this key in, update all downstream packages to use this key, and then remove octomap from all of Humble, Iron, Jazzy, and Rolling. This will avoid the ABI problem, and just use the system package everywhere.

See https://github.com/ros/rosdistro/issues/41622 for more information.

Note that I expect CI on this to fail until https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2024-71ae26cb75 lands in EPEL-9, about a week from now.

## Package name:

liboctomap-dev

## Package Upstream Source:

https://github.com/octomap/octomap

## Purpose of using this:

See the explanation above.

Several packages depend on having octomap available, including:

* hpp-fcl
* octomap_ros
* gz_dartsim_vendor
* mrpt2
* moveit_ros_occupancy_map_monitor
* moveit_core
* geometric_shapes
* octomap_rviz_plugins
* octomap_server

  - https://packages.debian.org/search?keywords=liboctomap-dev&searchon=names&suite=all&section=all
  - https://packages.ubuntu.com/search?keywords=liboctomap-dev&searchon=names&suite=all&section=all
  - https://pkgs.org/search/?q=octomap
  - https://pkgs.org/search/?q=octomap

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4549,6 +4549,11 @@ libnss3-dev:
   gentoo: [=dev-libs/nss-3*]
   nixos: [nss]
   ubuntu: [libnss3-dev]
+liboctomap-dev:
+  debian: [liboctomap-dev]
+  fedora: [octomap-devel]
+  rhel: [octomap-devel]
+  ubuntu: [liboctomap-dev]
 libogg:
   arch: [libogg]
   debian: [libogg-dev]


rosdep/base.yaml
```diff
@@ -4549,6 +4549,11 @@ libnss3-dev:
   gentoo: [=dev-libs/nss-3*]
   nixos: [nss]
   ubuntu: [libnss3-dev]
+liboctomap-dev:
+  debian: [liboctomap-dev]
+  fedora: [octomap-devel]
+  rhel: [octomap-devel]
```
audrow, line 4555: Should this be removed because it appears to only be on RHEL 8?

https://rhel.pkgs.org/8/epel-x86_64/octomap-devel-1.9.7-1.el8.x86_64.rpm.html

audrow (approved):
Looks good with happy PR jobs.

rosdep/base.yaml
clalancette: No, see my comment in the opening comment:

> Note that I expect CI on this to fail until https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2024-71ae26cb75 lands in EPEL-9, about a week from now.

clalancette: The package was now synced to RHEL-9, so CI finishes.  This is ready to be reviewed and merged.

marcoag (approved):

PR #41662, opened 2024-06-17
Add python3-transitions

lennewbrook:

## Package name:

python3-transitions

## Package Upstream Source:

https://github.com/pytransitions/transitions

## Purpose of using this:

python trasnsitrions is being used with ROS Humble and therefor needs to be available for Ubuntu :jammy and above

Distro packaging links:

  - REQUIRED
  - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4766,7 +4766,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/pybind11_json_vendor-release.git
-      version: 0.4.1-2
+      version: 0.5.0-1
     source:
       type: git
       url: https://github.com/open-rmf/pybind11_json_vendor.git


clalancette (changes requested):
This is not going to work for multiple reasons:

1.  This could break anyone using the `python-transitions` key.
2.  This isn't in alphabetical order.
3.  This is missing other needed operating systems, like Debian.

Please put the old key (`python-transitions`) back as it was, then add a completely new key in the appropriate place.  Also consider adding in other operating systems; you can get an idea of what is supported by looking at https://pkgs.org/search/?q=transitions

PR #41733, opened 2024-06-18
Update rmf_variants to 0.1.0-1 on jazzy

Yadunund:
Bloom succeeded but failed at the last step when opening the rosdistro PR. https://github.com/ros2-gbp/rmf_variants-release

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -5567,11 +5567,11 @@ repositories:
       tags:
         release: release/jazzy/{package}/{version}
       url: https://github.com/ros2-gbp/rmf_variants-release.git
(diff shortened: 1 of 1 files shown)


marcoag (approved):

marcoag:
Yup, that happens, you can always rerun `bloom` with:

```
  --pull-request-only, -p
                        skips the release actions and only tries to open a pull request
```

Yadunund:
> Yup, that happens, you can always rerun `bloom` with:
> 
> ```
>   --pull-request-only, -p
>                         skips the release actions and only tries to open a pull request
> ```

Thanks for the pro tip!

PR #41837, opened 2024-06-26
flex_sync: 1.0.0-1 in 'rolling/distribution.yaml' [bloom]

berndpfrommer:
Increasing version of package(s) in repository `flex_sync` to `1.0.0-1`:

- upstream repository: https://github.com/ros-misc-utilities/flex_sync.git
- release repository: https://github.com/ros2-gbp/flex_sync-release.git
- bloom version: `0.12.0`

## flex_sync

```
* first release as a ROS2 package
* Contributors: Bernd Pfrommer
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -1650,6 +1650,11 @@ repositories:
       type: git
       url: https://github.com/ros-misc-utilities/flex_sync.git
       version: rolling
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/flex_sync-release.git
+      version: 1.0.0-1
     source:
       type: git
       url: https://github.com/ros-misc-utilities/flex_sync.git


christophebedard:
This looks fine, but just one observation: I see you have version 1.0.0 for Rolling here and version 1.3.0 for Jazzy in #41840. If the same code is shared for all distros, you _could_ just release the same version in all distros. Otherwise, the version number for a newer distro is usually greater than the version number for an older distro.

berndpfrommer:
TLDR:

I use the following minor version numbering:
M.0.n = rolling
M.1.n = humble
M.2.n = iron
M.3.n = jazzy
M.4.n = ... you get the idea ..

This is ordered ascending with release number if you exclude rolling. Setting rolling to zero avoids collision in the future.

Longer version:

Having that said, I would love to be able to release the same version across all distros, since I keep the code base identical, and anyways work around distro API differences with #defines to avoid managing different source trees.
The reason I keep different version numbers is  because I don't understand the release process, so I slavishly follow the cookbook recipe and I used a separate track for each distro. At some point I got some error messages about version number conflicts  so google said should use different minor numbers to avoid that. Presto, that's why I idiotically have a ballooning number of package versions. I'm not sure exactly what I should put in the source and doc entry in the ros distros. Should I have just one "ros2" branch for my repos, instead of a branch for each distro, and put that "ros2" branch in the distribution.yaml? (It looks like that would probably be the right thing to do).  And do I need a separate track when running bloom-release? I'm very nervous about getting the packaging state screwed up to the point of no return.

Glad you bring this up BTW. I don't really know where to ask questions related to best practices. The discourse is not meant for tech questions, and robotics stack exchange doesn't feel like the right place either.

christophebedard:
I don't have a _ton_ of experience with this (so maybe @clalancette can confirm), but I believe this is what you'd need to do:

1. Keep the separate tracks in your release repo's `tracks.yaml`
2. Have all relevant tracks (i.e., all active distros) point `devel_branch:` to the same branch, e.g., `ros2`
    - I believe you can just change the `tracks.yaml` file manually for this and push to the release repo
4. Have all tracks point to the same version
    - This will happen when you run `bloom-release` separately for each distro. Since the tracks all point to the same branch, they'll all find and use the same version
    - You might need to choose a new common version number that is greater than all current version numbers; I'm not sure how bloom and/or apt act if some version numbers suddenly decrease
5. Update the `version:` for `doc:` and `source:` in the `${ROS_DISRO}/distribution.yaml` file here in ros/rosdistro to point to the same branch (`ros2`)

Then bloom & the build farm will create "separate" binaries for each distro that will point to the same tag on the same branch.

For example, the tracks for Iron, Jazzy, and Rolling of the `angles` packages use the same branch (`ros2`) and the same version (`1.16.0`):

1. `tracks.yaml`: https://github.com/ros2-gbp/angles-release/blob/8158d4bb244c8300d6473c64844a7dec28b0f881/tracks.yaml#L176-L253
2. rosdistro Rolling: https://github.com/ros/rosdistro/blob/7bc7b72d338ca9a3694e10fb148b89d0f8bdb599/rolling/distribution.yaml#L311-L326
3. rosdistro Jazzy: https://github.com/ros/rosdistro/blob/7bc7b72d338ca9a3694e10fb148b89d0f8bdb599/jazzy/distribution.yaml#L311-L326
4. rosdistro Iron: https://github.com/ros/rosdistro/blob/7bc7b72d338ca9a3694e10fb148b89d0f8bdb599/iron/distribution.yaml#L310-L325

> I'm very nervous about getting the packaging state screwed up to the point of no return.

Yeah I feel you :laughing: 

> Glad you bring this up BTW. I don't really know where to ask questions related to best practices. The discourse is not meant for tech questions, and robotics stack exchange doesn't feel like the right place either.

I think Robotics StackExchange would be appropriate. Since this is common enough, I think it would be useful to document this use-case in the ROS 2 docs somewhere under https://docs.ros.org/en/rolling/How-To-Guides/Releasing/Releasing-a-Package.html.

> At some point I got some error messages about version number conflicts so google said should use different minor numbers to avoid that.

Oh, did you get this error with bloom or apt?

berndpfrommer:
Thanks for the valuable suggestions. You will probably see a related PR soon where I try this on one of my repos. I guess I can just bump the major number to avoid "apt" calamity.

Sorry, I no longer recall what was giving the errors, this was good while ago. It was when I ran bloom-release, but I no longer remember which part complained.

christophebedard:
Can you just confirm that you want to go ahead with this series of PRs for flex_sync?

berndpfrommer:
Closing this PR.  @christophebedard you convinced me to nix this PR and try the single-version approach.

berndpfrommer:
https://robotics.stackexchange.com/questions/112024/how-to-maintain-a-single-package-version-numbers-across-multiple-ros2-distros/112025#112025

PR #41870, opened 2024-06-28
Add python3-opencv-pip to obtain the newest version possible

KozaMateusz:

## Package name:

opencv-python

## Package Upstream Source:

[https://pypi.org/project/opencv-python/](https://github.com/opencv/opencv)

## Purpose of using this:

NMSBoxesBatched was added in a later version of OpenCV which is available only on PyPi.

- pip: [https://pypi.org/project/opencv-python/](https://pypi.org/project/opencv-python/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7454,6 +7454,10 @@ python3-opencv:
     '7': null
     '8': null
   ubuntu: [python3-opencv]
+python3-opencv-pip:
+  '*':
+    pip:
+      packages: [opencv-python]
 python3-opengl:
   debian: [python3-opengl]
   fedora: [python3-pyopengl]


clalancette (commented):
I'm honestly not sure that this is a good idea, particularly for libraries like this that we know are compiled.  We are just dealing with ABI problems from another library where we accidentally mixed up a system-installed package and a ROS-specific package (see https://github.com/ros/rosdistro/issues/41622).

Thus, my thinking here is that we should be heavily encouraging people to use the system-provided versions, and we shouldn't provide pip keys for things that the system already has.  But I'll ask the rest of the team what they think about it.

KozAAAAA:
So, what's the solution if this pull request doesn't get approved?

clalancette:
> So, what's the solution if this pull request doesn't get approved?

You can use a local rosdep database in that case.  See http://docs.ros.org/en/independent/api/rosdep/html/contributing_rules.html on how to do that.

clalancette:
So we did discuss this, and the general thought was that we generally should not do this.  It leads to too much confusion, and particularly in this kind of case, can lead to ABI problems.

So I'm going to close this out for now.

PR #41871, opened 2024-06-28
Use the right branch on geometry_tutorials

ahcorde:
Related with this PR https://github.com/ros/ros_tutorials/pull/169

I created turtlesim_msgs which is not backportable to other distros. This PR use the right branch

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -2653,7 +2653,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros/geometry_tutorials.git
-      version: ros2
+      version: humble
     release:
       packages:
       - geometry_tutorials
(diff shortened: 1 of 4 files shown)


clalancette (commented):
I'm going to suggest that since we are renaming branches on `geometry_tutorials`, we also switch the Rolling branch to `rolling`.  It is just more consistent with what we do in the core.

ahcorde:
I created the branch `rolling` and update rosdistro. 

What we should do with the `ros2` branch in the `geometry_tutorials` repository?

clalancette:
> What we should do with the `ros2` branch in the `geometry_tutorials` repository?

I would just leave it as-is.  It is historical, and used in older distributions.

clalancette (approved):
This looks good to me, but we need to hold off on merging it so we don't break the Jazzy sync freeze.

PR #41946, opened 2024-07-03
Add PyQt6 to older Ubuntu versions with Pip

InvincibleRMC:
## Purpose of using this:

To make Guis with PyQt6 on older Ubuntu Versions

Distro packaging links:

https://pypi.org/project/PyQt6/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -470,8 +470,10 @@ pyqt6-dev-tools:
   fedora: [python3-pyqt6-base]
   ubuntu:
     '*': [pyqt6-dev-tools]
-    focal: null
-    jammy: null
+    focal:
+      pip: PyQt6
+    jammy:
+      pip: PyQt6
 pyrebase-pip:
   debian:
     pip:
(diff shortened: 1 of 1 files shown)


marcoag:
I would say we usually don't accept pip packages that have a system version. In this case for `focal` and `jammy` it seems that `pyqt5` is available, even though an older version there's still a system version of it. With that said let me still run it through the ROS Core meeting and come back with the outcome of that.

marcoag:
Adding a precedent: https://github.com/ros/rosdistro/pull/41870#pullrequestreview-2148978970

InvincibleRMC:
Following up on this I would like to note that pyqt5 is available but eol https://endoflife.date/qt

InvincibleRMC:
If this solution does not get accepted would it be possible to add a python3-pyqt6-pip instead only for releases that don't have a pre-built package?

marcoag:
The problem is the releases you mentioned have a pre-built package, just that is from a previous version.

PR #41960, opened 2024-07-03
Add Mavproxy pip

InvincibleRMC:

## Package Upstream Source:

https://github.com/ArduPilot/MAVProxy

## Purpose of using this:

Allows communicating to UAVs over tcp. I used this greatly for configuration of my UAV before using ROS to control it.

Distro packaging links:

  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6950,6 +6950,16 @@ python3-matplotlib:
     '7': null
   slackware: [python3-matplotlib]
   ubuntu: [python3-matplotlib]
+python3-mavproxy-pip:
+ debian:
+    pip:
+      packages: [MAVProxy]
+  fedora:
+    pip:
+      packages: [MAVProxy]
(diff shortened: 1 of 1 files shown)


marcoag (approved):

rosdep/python.yaml
```diff
@@ -6950,6 +6950,16 @@ python3-matplotlib:
     '7': null
   slackware: [python3-matplotlib]
   ubuntu: [python3-matplotlib]
+python3-mavproxy-pip:
```
Yadunund, line 6953: > Standard pip disclaimer: ROS packages that depend on pip keys cannot be released into a ROS distribution. They can only be depended on by from-source builds. Because of this, system packages are highly preferred to pip packages.

Yadunund (approved):

PR #41961, opened 2024-07-03
Add pyqtdarktheme pip

InvincibleRMC:

## Package Upstream Source:

https://github.com/5yutan5/PyQtDarkTheme

## Purpose of using this:

Has nice CSS for creating dark themes for PyQt. 

  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8247,7 +8247,7 @@ python3-pyproj:
     '7': null
   ubuntu: [python3-pyproj]
 python3-pyqtdarktheme-pip:
-   debian:
+  debian:
     pip:
       packages: [pyqtdarktheme]
   fedora:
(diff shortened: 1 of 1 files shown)


marcoag (approved):

rosdep/python.yaml
```diff
@@ -8275,6 +8275,19 @@ python3-pyqt5.qtwebengine:
   openembedded: ['${PYTHON_PN}-pyqt5@meta-qt5']
   opensuse: [python3-qtwebengine-qt5]
   ubuntu: [python3-pyqt5.qtwebengine]
+python3-pyqtdarktheme-pip:
```
Yadunund, line 8278: > Standard pip disclaimer: ROS packages that depend on pip keys cannot be released into a ROS distribution. They can only be depended on by from-source builds. Because of this, system packages are highly preferred to pip packages.

Yadunund (approved):

PR #41977, opened 2024-07-05
Add python3-pyqt6.qtmultimedia

NoahMollerstuen:

## Package name:

python3-pyqt6.qtmultimedia

## Package Upstream Source:

https://pypi.org/project/PyQt6/

## Purpose of using this:

Used for working with audio and video in GUI applications

Distro packaging links:

  - https://packages.debian.org/bookworm/python3-pyqt6.qtmultimedia
  - https://packages.ubuntu.com/noble/python3-pyqt6.qtmultimedia
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8317,11 +8317,12 @@ python3-pyqt5.qtwebengine:
   ubuntu: [python3-pyqt5.qtwebengine]
 python3-pyqt6.qtmultimedia:
   debian:
-    bullseye: null
     '*': [python3-pyqt6.qtmultimedia]
+    bullseye: null
   ubuntu:
-    '*': null
-    'noble': [python3-pyqt6.qtmultimedia]
+    '*': [python3-pyqt6.qtmultimedia]
+    focal: null
+    jammy: null
 python3-pyqtdarktheme-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8304,7 +8304,9 @@ python3-pyqt5.qtwebengine:
   opensuse: [python3-qtwebengine-qt5]
   ubuntu: [python3-pyqt5.qtwebengine]
 python3-pyqt6.qtmultimedia:
-  debian: [python3-pyqt6.qtmultimedia]
+  debian:
+    bullseye: null
+    '*': [python3-pyqt6.qtmultimedia]
   ubuntu:
     '*': null
     'noble': [python3-pyqt6.qtmultimedia]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -8275,6 +8275,11 @@ python3-pyqt5.qtwebengine:
   openembedded: ['${PYTHON_PN}-pyqt5@meta-qt5']
   opensuse: [python3-qtwebengine-qt5]
   ubuntu: [python3-pyqt5.qtwebengine]
+python3-pyqt6.qtmultimedia:
+  debian: [python3-pyqt6.qtmultimedia]
```
marcoag, line 8279: [`Humble` and `Iron` run on `bullseye`](https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027). The package only seems to have started to be available since `bookworm`: https://packages.debian.org/search?keywords=python3-pyqt6.qtmultimedia&searchon=names&suite=all&section=all

Maybe it's better to make sure it doesn't resolve to a nonexistent package in bullseye:

```suggestion
  debian:
    bullseye: null
    '*': [python3-pyqt6.qtmultimedia]
```

marcoag (dismissed):

rosdep/python.yaml
NoahMollerstuen: That makes sense; thanks for reviewing

Yadunund:
@NoahMollerstuen could you address the git conflict?

NoahMollerstuen:
> @NoahMollerstuen could you address the git conflict?

Done!

rosdep/python.yaml
```diff
@@ -8303,6 +8303,13 @@ python3-pyqt5.qtwebengine:
   openembedded: ['${PYTHON_PN}-pyqt5@meta-qt5']
   opensuse: [python3-qtwebengine-qt5]
   ubuntu: [python3-pyqt5.qtwebengine]
+python3-pyqt6.qtmultimedia:
+  debian:
+    bullseye: null
+    '*': [python3-pyqt6.qtmultimedia]
+  ubuntu:
+    '*': null
+    'noble': [python3-pyqt6.qtmultimedia]
```
clalancette, line 8312: We typically assume that if a distribution has it, then newer ones will as well.  So we should reverse the order here:

```suggestion
  debian:
    '*': [python3-pyqt6.qtmultimedia]
    bullseye: null
  ubuntu:
    '*': [python3-pyqt6.qtmultimedia]
    focal: null
    jammy: null
```

clalancette (changes requested):
Besides what I put inline, this needs to be rebased for unrelated reasons.  All of that should fix CI.

clalancette (approved):

sloretz (approved):

PR #41979, opened 2024-07-08
Add tinkerforge pip

BenKlee:

# The source is here:

https://pypi.org/project/tinkerforge/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9518,6 +9518,16 @@ python3-timm-pip:
   ubuntu:
     pip:
       packages: [timm]
+python3-tinkerforge-pip:
+  debian:
+    pip:
+      packages: [tinkerforge]
+  fedora:
+    pip:
+      packages: [tinkerforge]
+  ubuntu:
+    pip:
+      packages: [tinkerforge]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -10142,6 +10142,16 @@ tilestache:
   debian: [tilestache]
   fedora: [python-tilestache]
   ubuntu: [tilestache]
+tinkerforge:
```
marcoag, line 10145:
```suggestion
tinkerforge-pip:
```

Can we add the `-pip`, this helps easily identify the dependency as a pip package and reduces the chances of a future clash of the label with a system package.

marcoag (dismissed):

rosdep/python.yaml
clalancette: Actually, to pass our naming this needs to be `python3-tinkerforge-pip` (and needs to be moved into the correct place in the file to be alphabetical).

mjcarroll:
@BenKlee friendly ping, can you address the feedback here?

mjcarroll (approved):

clalancette (approved):

PR #41120, opened 2024-05-09
Index Beluga documentation and sources

hidmic:
# Please Add This Package to be indexed in the rosdistro.

`beluga`

# The source is here:

https://github.com/Ekumen-OS/beluga

# Checks
 - [x] All packages have a declared license in the package.xml
 - [x] This repository has a LICENSE file
 - [x] This package is expected to build on the submitted rosdistro

PR #42119, opened 2024-07-17
Disable all `beluga` devel jobs

hidmic:
Follow-up to #41120. Devel jobs on the farm have been trying and failing to build some https://github.com/Ekumen-OS/beluga packages, the ones including `pip` dependencies that we don't release. This patch (hopefully) disables these jobs. We have our own CI pipelines and policy to cover commits, PRs, and releases.

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -866,6 +866,8 @@ repositories:
       type: git
       url: https://github.com/Ekumen-OS/beluga.git
       version: main
+      test_commits: false
+      test_pull_requests: false
     status: developed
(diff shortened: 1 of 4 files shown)


hidmic:
FYI @nahueespinosa @serraramiro1 @glpuga

sloretz (approved):

sloretz:
Noetic is in a sync, but this won't affect released packages so I'm ok with merging it https://discourse.ros.org/t/preparing-for-noetic-sync-2024-07-18/38648

PR #42144, opened 2024-07-22
Add `tkintermapview` pip dependency to python.yaml

jack-obrien:

## Package name:

`tkintermapview`

## Package Upstream Source:

https://github.com/TomSchimansky/TkinterMapView

## Purpose of using this:

This dependency is being used to build GUIs with maps in my ROS project. It is a dependency of my project, so it would be valuable to include it as a rosdep key so that others can easily include it too.

Distro packaging links:

Pip only, since the project is relatively young.

https://pypi.org/project/tkintermapview/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9594,6 +9594,10 @@ python3-tk:
   opensuse: [python3-tk]
   rhel: ['python%{python3_pkgversion}-tkinter']
   ubuntu: [python3-tk]
+python3-tkintermapview-git:
+  '*':
+    pip:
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

clalancette (approved):

clalancette (changes requested):
This is completely not your fault, but this PR needs to be rebased onto the latest.  Once that is done, this should pass CI.  Apologies.

jack-obrien:
No worries, should be good now

clalancette (approved):

PR #42157, opened 2024-07-22
add ros_battery_monitoring repository

ottojo:
Please add the https://github.com/ipa320/ros_battery_monitoring repository to rosdistro for indexing and later binary release of the contained packages:
* battery_state_broadcaster
* battery_state_rviz_overlay

The first package implements a ros2_control broadcaster for BatteryState messages, while the node in the second package converts these packages into OverlayText messages for display as 2d overlays in rviz.

Feedback on package naming is appreciated, in particular whether `battery_state_broadcaster` should be named `ros2_control_battery_state_broadcaster` or if that would wrongly imply any "official affiliation" with the ros2_control project.

# Checks

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -6109,6 +6109,13 @@ repositories:
       url: https://github.com/osrf/ros2launch_security.git
       version: main
     status: maintained
+  ros_battery_monitoring:
+    source:
+      test_pull_requests: true
+      type: git
+      url: https://github.com/ipa320/ros_battery_monitoring.git
+      version: main
+    status: developed
   ros_canopen:
     release:
       packages:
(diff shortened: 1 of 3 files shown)


mjcarroll:
I'm not sure, but I think there is a potential issue with REP 144 package naming here: 

> A package name should not contain ros as it is redundant. Exceptions include core packages and ROS bindings of an upstream library (e.g. moveit_ros)

But also, in this case, it's just in the repository name, rather than in the package names, so maybe we are okay.  @clalancette what do you think?

clalancette:
> But also, in this case, it's just in the repository name, rather than in the package names, so maybe we are okay. @clalancette what do you think?

Yeah, I'm fine with it because it is just the repository name.

clalancette (approved):

PR #42168, opened 2024-07-22
Add python-transforms3d-pip For All Ubuntu Versions

danthony06:

## Package name:

python-transforms3d-pip

## Package Upstream Source:

[transforms3d](https://matthew-brett.github.io/transforms3d/)

## Purpose of using this:

I need this package for `rolling`. I think every supported distribution is currently explicitly called out, except for `noble`, so I simplified the rule to install it for all Ubuntu versions.

Distro packaging links:

This is a pip package.

rolling

# The source is here:

https://matthew-brett.github.io/transforms3d/

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4287,16 +4287,8 @@ python-transforms3d-pip:
     pip:
       packages: [transforms3d]
   ubuntu:
-    '*': null
-    bionic:
-      pip:
-        packages: [transforms3d]
-    focal:
-      pip:
-        packages: [transforms3d]
-    jammy:
-      pip:
-        packages: [transforms3d]
+    pip:
+      packages: [transforms3d]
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

clalancette (approved):

danthony06:
Thanks

danthony06:
@clalancette Sorry, it looks like this may be causing problems on `rolling`. My ROS-I CI is failing [here](https://github.com/swri-robotics/marti_common/actions/runs/10049091224/job/27883690955?pr=741#step:3:1098). It looks like Ubuntu 24.04 is no longer allowing us to install pip package system-wide in the past. Is this something we should update the ROS-I CI pipeline to work with, or is this a larger problem with pip packages in `rolling`?

PR #42174, opened 2024-07-23
fix(noetic): update imaga transport plugins to 1.15.0-1 again

daenny:
Re-apply: https://github.com/ros/rosdistro/pull/35994


mjcarroll:
Thanks for flagging this @daenny.  @sloretz did a few touchups and I think we are good to go here.

PR #42205, opened 2024-07-25
Add PyRoboSim to python.yaml

sea-bass:

## Package name:

Python package : `pyrobosim`
rosdep entry : `python3-pyrobosim-pip`

## Package Upstream Source:

https://github.com/sea-bass/pyrobosim
https://github.com/sea-bass/pyrobosim_ros

## Purpose of using this:

PyRoboSim has existed as a Python package with a ROS 2 wrapper for about 2 years. However, the ROS 2 wrapper lives within the source code and thus is not well-integrated into the ROS 2 ecosystem for distribution.

I am now working on breaking up the `pyrobosim` repo into a ROS-free package (the one on PyPi below) and its ROS 2 wrappers will be officially bloomed soon.

So, this PR handles only the Python package, and I will next break out a new `pyrobosim_ros` repo and bloom that.

This is available on PyPi here: https://pypi.org/project/pyrobosim/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8343,6 +8343,10 @@ python3-pyro-ppl-pip:
   ubuntu:
     pip:
       packages: [pyro-ppl]
+python3-pyrobosim-pip:
+  '*':
+    pip:
+      packages: [pyrobosim]
 python3-pyrr-pip:
   debian:
     pip:


clalancette:
> I am now working on breaking up the `pyrobosim` repo into a ROS-free package (the one on PyPi below) and its ROS 2 wrappers will be officially bloomed soon.

Just so you are aware, this will not be possible.  It is not possible to bloom a package that has `pip`-only dependencies, as our buildfarm does not allow installing from `pip`.

We can still merge this key in, but it will only be possible to use it in from-source builds.  Let us know how you want to proceed.

sea-bass:
Ah, thanks for the clarification, @clalancette!

I guess I'd still like to proceed here so `rosdep` tells users about the missing pip install when source building the ROS package.

clalancette (approved):

mjcarroll (approved):

PR #42315, opened 2024-08-03
Add `python3-smbus2`

InvincibleRMC:

## Package name:

python3-smbus2

## Package Upstream Source:

https://github.com/kplindegaard/smbus2

## Purpose of using this:

Used for low level communication same as python3-smbus2-pip but for built packages.

Distro packaging links:

https://packages.debian.org/sid/python3-smbus2
https://packages.ubuntu.com/noble/python3-smbus2

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9252,7 +9252,6 @@ python3-smbus:
     '8': null
   ubuntu: [python3-smbus]
 python3-smbus2:
-  arch: [python-smbus2]
   debian: [python3-smbus2]
   ubuntu:
     '*': [python3-smbus2]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -9251,6 +9251,12 @@ python3-smbus:
     '*': [python3-i2c-tools]
     '8': null
   ubuntu: [python3-smbus]
+python3-smbus2:
+  debian: [python3-smbus2]
+  ubuntu:
+    '*': [python3-smbus2]
+    focal: null
+    jammy: null
```
nuclearsandwich, line 9259: There actually is a Jammy package so unless there's something known to be wrong with it, we don't need to null this key: https://packages.ubuntu.com/jammy/python3-smbus2
```suggestion
```

sloretz (approved):

marcoag (approved):

PR #42333, opened 2024-08-06
adding python3-nep-pip

simeoncapy:

## Package name:

NEP

## Package Upstream Source:

* Source: [https://github.com/enriquecoronadozu/NEP](https://github.com/enriquecoronadozu/NEP)
* PIP: [https://pypi.org/project/nep/](https://pypi.org/project/nep/)
* Documentation: [https://enriquecoronadozu.github.io/NEP/](https://enriquecoronadozu.github.io/NEP/) and [https://enrique-coronado.gitbook.io/nep-docs](https://enrique-coronado.gitbook.io/nep-docs)

## Purpose of using this:

This package is used to communicate between two devices regardless of the OS. It uses the same publisher/subscriber system as ROS and can be used in several languages (Python, C#, Javascript, C++...). It can be an alternative to ROS when just the communication is required or with a language not available with ROS.

Distro packaging links:

- PyPi: https://pypi.org/project/nep/

NB: I ran the `pytest` and it returned `9 passed, 1 warning`. I also tried to do `rosdep install --from-paths src -y --ignore-src` and it worked (but I only tried on Ubuntu 22.04). The package also works on Windows but I didn't add it since it is not supported on rosdep.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7232,6 +7232,13 @@ python3-nclib-pip:
   ubuntu:
     pip:
       packages: [nclib]
+python3-nep-pip:
+  osx:
+    pip:
+      packages: [nep]
+  ubuntu:
+    pip:
+      packages: [nep]
 python3-nest-asyncio:
   alpine: [py3-nest_asyncio]
   arch: [python-nest-asyncio]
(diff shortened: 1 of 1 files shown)


sloretz (approved):

sloretz (approved):

PR #42353, opened 2024-08-08
Add PySide6 to rosdep keys

samyarsadat:

## Package name:

PySide6

## Package Upstream Source:

https://github.com/qtproject/pyside-pyside-setup

## Purpose of using this:

This package is a Python wrapper for Qt 6. It allows one to create Qt applications using Python.

- PyPi: https://pypi.org/project/PySide6/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8544,8 +8544,8 @@ python3-pyside6:
     bullseye:
       pip:
         packages: [PySide6]
-    trixie: [libpyside6-dev]
     sid: [libpyside6-dev]
+    trixie: [libpyside6-dev]
   osx:
     pip:
       packages: [PySide6]
@@ -8569,56 +8569,56 @@ python3-pyside6:
     oracular: [libpyside6-dev]
 python3-pyside6.qtasyncio:
   debian:
-    trixie: [python3-pyside6.qtasyncio]
     sid: [python3-pyside6.qtasyncio]
+    trixie: [python3-pyside6.qtasyncio]
   ubuntu:
     oracular: [python3-pyside6.qtasyncio]
 python3-pyside6.qtcharts:
   debian:
-    trixie: [python3-pyside6.qtcharts]
     sid: [python3-pyside6.qtcharts]
+    trixie: [python3-pyside6.qtcharts]
   ubuntu:
     oracular: [python3-pyside6.qtcharts]
 python3-pyside6.qtconcurrent:
   debian:
-    trixie: [python3-pyside6.qtconcurrent]
     sid: [python3-pyside6.qtconcurrent]
+    trixie: [python3-pyside6.qtconcurrent]
   ubuntu:
     oracular: [python3-pyside6.qtconcurrent]
 python3-pyside6.qtcore:
   debian:
-    trixie: [python3-pyside6.qtcore]
     sid: [python3-pyside6.qtcore]
+    trixie: [python3-pyside6.qtcore]
   ubuntu:
     oracular: [python3-pyside6.qtcore]
 python3-pyside6.qtdatavisualization:
   debian:
-    trixie: [python3-pyside6.qtdatavisualization]
     sid: [python3-pyside6.qtdatavisualization]
+    trixie: [python3-pyside6.qtdatavisualization]
   ubuntu:
     oracular: [python3-pyside6.qtdatavisualization]
 python3-pyside6.qtgui:
   debian:
-    trixie: [python3-pyside6.qtgui]
     sid: [python3-pyside6.qtgui]
+    trixie: [python3-pyside6.qtgui]
   ubuntu:
     oracular: [python3-pyside6.qtgui]
 python3-pyside6.qtqml:
   debian:
-    trixie: [python3-pyside6.qtqml]
     sid: [python3-pyside6.qtqml]
+    trixie: [python3-pyside6.qtqml]
   ubuntu:
     oracular: [python3-pyside6.qtqml]
 python3-pyside6.qttest:
   debian:
-    trixie: [python3-pyside6.qttest]
     sid: [python3-pyside6.qttest]
+    trixie: [python3-pyside6.qttest]
   ubuntu:
     oracular: [python3-pyside6.qttest]
 python3-pyside6.qtwidgets:
   debian:
-    trixie: [python3-pyside6.qtwidgets]
     sid: [python3-pyside6.qtwidgets]
+    trixie: [python3-pyside6.qtwidgets]
   ubuntu:
     oracular: [python3-pyside6.qtwidgets]
 python3-pysnmp:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8536,7 +8536,7 @@ python3-pyside2.qtwidgets:
     '*': [python3-pyside2]
     '8': null
   ubuntu: [python3-pyside2.qtwidgets]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -8442,6 +8442,10 @@ python3-pyside2.qtwidgets:
     '*': [python3-pyside2]
     '8': null
   ubuntu: [python3-pyside2.qtwidgets]
+python3-pyside6-pip:
+  '*':
+    pip:
+      packages: [PySide6]
```
marcoag, line 8448: It seems like PySide6 is coming for next [Ubuntu](https://packages.ubuntu.com/search?keywords=pyside6&searchon=names&suite=all&section=all) and [Debian](https://packages.debian.org/search?keywords=pyside6&searchon=names&suite=all&section=all) distros. Since we rather not have pip packages when the system package are available, I would suggest to scope this to specific distros (feel free to add more):

```suggestion
  debian:
    bullseye:
      pip:
        packages: [PySide6]
    bookworm:
      pip:
        packages: [PySide6]
  ubuntu:
    focal:
      pip:
        packages: [PySide6]
    jammy:
      pip:
        packages: [PySide6]
    noble:
      pip:
        packages: [PySide6]
```

Feel free to add other distros.

marcoag (dismissed):

marcoag:

rosdep/python.yaml
samyarsadat: I have incorporated your suggestion, and also added keys for RHEL 8, RHEL 9, and OSX.

audrow (dismissed):

clalancette (commented):
This one ends up being tricky.

We know that system packages for pyside6 are coming; they are already available for [Ubuntu Oracular](https://packages.ubuntu.com/source/oracular/pyside6) and [Debian Sid](https://packages.debian.org/source/sid/pyside6).

However, as a matter of course we don't usually add in keys for pre-release distributions.

There are 2 ways we could go here:
1.  Take in this `python3-pyside6-pip` key for now, and later add in a `python3-pyside6` key.
2.  Right now go for a `python3-pyside6` key, and fill it out with `pip` for everything except for very new Ubuntu and Debian distributions.

Personally, I think we should go for 2, but I can be convinced otherwise.

@mjcarroll @alsora Any thoughts here?

samyarsadat:
I also think number 2 is a better option, but there is one small issue with it.
As with PySide2, the PySide6 binary packages are split up into different modules (ex. `python3-pyside6.qtqml`, `python3-pyside6.qtgui`, etc.), so keys for all of those would have to be added as well (at least for the latest Debian and Ubuntu distros that are going to use system packages).

clalancette:
> As with PySide2, the PySide6 binary packages are split up into different modules (ex. `python3-pyside6.qtqml`, `python3-pyside6.qtgui`, etc.), so keys for all of those would have to be added as well (at least for the latest Debian and Ubuntu distros that are going to use system packages).

Yeah, agreed.  However, we can do that piecemeal.  What particular parts of PySide6 are you interested in?

alsora:
I agree with option 2.
It also seems slightly more consistent with the rest of the file, where the `-pip` suffix is used only for those dependencies that always come from `pip`, while no suffix is used for dependencies where we use a mixture of `pip` and system packages.

samyarsadat:
> > As with PySide2, the PySide6 binary packages are split up into different modules (ex. `python3-pyside6.qtqml`, `python3-pyside6.qtgui`, etc.), so keys for all of those would have to be added as well (at least for the latest Debian and Ubuntu distros that are going to use system packages).
> 
> Yeah, agreed. However, we can do that piecemeal. What particular parts of PySide6 are you interested in?

That works. I currrently need `qtcore`, `qtgui`, `qtconcurrent`, `qtqml`, and `qtwidgets`.

Should there be one key named `python3-pyside6` that installs PySide6 thourgh pip on distros with no system packages and `libpyside6-dev`, `python3-pyside6.qtcore`, `python3-pyside6.qtgui`, `python3-pyside6.qtconcurrent`, `python3-pyside6.qtqml`, and `python3-pyside6.qtwidgets` on distros with system packages, or should there be individual keys for all of the different modules?

clalancette:
> or should there be individual keys for all of the different modules?

I would start with individual keys.  That gives packagers maximum control.  If it turns out that there is really a need for a "consolidated" key, we can always add that in later.

samyarsadat:
I've updated the `python3-pyside6` key and added all the new keys for the different modules.
I previously said that I needed `qtcore`, `qtgui`, `qtconcurrent`, `qtqml`, and `qtwidgets`, but I've added a few more:
`qtasyncio`, `qtcharts`, `qtdatavisualization`, `qttest`.

sloretz (approved):

rosdep/python.yaml
```diff
@@ -8442,6 +8442,81 @@ python3-pyside2.qtwidgets:
...
   ubuntu: [python3-pyside2.qtwidgets]
+python3-pyside6:
+  debian:
+    bookworm:
+      pip:
+        packages: [PySide6]
+    bullseye:
+      pip:
+        packages: [PySide6]
+    sid: [libpyside6-dev]
```
nuclearsandwich, line 8453:
```suggestion
    trixie: [libpyside6-dev]
    sid: [libpyside6-dev]
```

Add the current Debian Testing distribution as well. I would even consider dropping the `sid` definition altogether in favor of just trixie but don't mind it being there.

It's worth making the change for each of the definitions below as well.

nuclearsandwich (approved):

nuclearsandwich:
@sloretz I modified the PR to include trixie definitions, would you please give it an updated review with my changes.

rosdep/python.yaml
nuclearsandwich: Done in https://packages.debian.org/testing/python3-pyside6.qtasyncio

sloretz: Looks like this made CI unhappy

```
    ERR: list out of alphabetical order line 8526.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8551.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8557.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8563.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8569.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8575.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8581.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8587.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8593.  'sid' should come before 'trixie'
    ERR: list out of alphabetical order line 8599.  'sid' should come before 'trixie'
```

clalancette:
I updated this to be more forwards-compatible.  With this in place, I'm happy, so I'm going to approve.  I would still like another approval before merging.

clalancette (approved):

cottsay (approved):
Looks good. Could use an OS-level wildcard for the pip rule on `python3-pyside6` and drop the explicit osx and rhel rules if you like.

clalancette:
> Looks good. Could use an OS-level wildcard for the pip rule on `python3-pyside6` and drop the explicit osx and rhel rules if you like.

Thanks for the review.  I'm going to leave things as-is for now, we can always follow-up later on.

PR #42387, opened 2024-08-09
Removing Mapviz

danthony06:
I'm removing Mapviz from Jazzy until we release some of its dependencies so that I don't get spammed about failing builds from the buildfarm.

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -3359,27 +3359,6 @@ repositories:
       url: https://github.com/Neargye/magic_enum.git
       version: master
     status: maintained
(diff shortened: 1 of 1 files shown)


marcoag (approved):

marcoag:
I see that only `mapviz_plugins` is failing on `jazzy`: https://repo.ros2.org/status_page/ros_jazzy_default.html?q=mapviz
Maybe wanna consider scoping the removal just to just that package and keep the rest?

danthony06:
Thanks, but I'd like to get everything released at once.

marcoag (approved):

PR #42459, opened 2024-08-15
Remove performance_test from rosdistro

### LINKED ISSUE #42458: Performance_test buildfarm issues
Hey there!

apparently our package is not building properly in the buildfarm, we created an issue 1year ago, but, did not get any feedback: https://gitlab.com/ApexAI/performance_test/-/issues/166

Is there anything should we do? Our nighly pipelines are working fine with the osrf docker images.

Thanks in advance!

LanderU:
Closes #42458

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -3884,21 +3884,6 @@ repositories:
       url: https://github.com/ros-perception/perception_pcl.git
       version: foxy-devel
     status: maintained
-  performance_test:
-    doc:
-      type: git
-      url: https://gitlab.com/ApexAI/performance_test.git
-      version: 1.2.1
-    release:
-      tags:
-        release: release/foxy/{package}/{version}
-      url: https://github.com/ros2-gbp/performance_test-release.git
-      version: 1.2.1-1
-    source:
-      type: git
-      url: https://gitlab.com/ApexAI/performance_test.git
-      version: master
-    status: maintained
   performance_test_fixture:
     release:
       tags:
--- a/galactic/distribution.yaml
+++ b/galactic/distribution.yaml
@@ -3072,21 +3072,6 @@ repositories:
       url: https://github.com/ros-perception/perception_pcl.git
       version: foxy-devel
     status: maintained
-  performance_test:
-    doc:
-      type: git
-      url: https://gitlab.com/ApexAI/performance_test.git
-      version: 1.2.1
-    release:
-      tags:
-        release: release/galactic/{package}/{version}
-      url: https://github.com/ros2-gbp/performance_test-release.git
-      version: 1.2.1-1
-    source:
-      type: git
-      url: https://gitlab.com/ApexAI/performance_test.git
-      version: master
-    status: maintained
   performance_test_fixture:
     release:
       tags:
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5941,21 +5941,6 @@ repositories:
       url: https://github.com/ros-perception/perception_pcl.git
       version: ros2
     status: maintained
-  performance_test:
-    doc:
-      type: git
-      url: https://gitlab.com/ApexAI/performance_test.git
-      version: 1.2.1
-    release:
-      tags:
-        release: release/humble/{package}/{version}
-      url: https://github.com/ros2-gbp/performance_test-release.git
-      version: 1.2.1-1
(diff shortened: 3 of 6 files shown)


audrow:
We normally don't modify EOL distros. I think removing a package may be an exception. I've asked in the OSRF slack.

tfoote (changes requested):
For preventing the build failures this is overkill. To do that specifically you'll want to only remove the release stanzas. If you remove the entire entry you will also remove the indexing and documentation which means that the community won't be able to discover your packages anywhere near as easily. 

> did not get any feedback: https://gitlab.com/ApexAI/performance_test/-/issues/166

Note that you @ mentioned one developer on your own repository, which I don't think that they're watching. If you want help please use one of the support channels (such as robotics.stackexchange.com ) and not rely on a single ping of a single developer   on a non-standard channel. 

However I think that there's likely just some things that can be fixed to resolve your issues. 

For example I took a quick look at the the Hdev test looks like it's failing for some linting errrors: https://build.ros2.org/job/Hdev__performance_test__ubuntu_jammy_amd64/79/  You should be able to fix those. 

It is also flagged that the released version was out of date. Has it been attempted to review the errors and update the release?

ErikAtApex:
Thanks @tfoote . I fixed the linting errors, and perhaps the JUnit error also. I should be able to slowly push this to a better place during the next several days.

At the end of the jobs, several emails are sent. For example:

https://build.ros2.org/job/Jdev__performance_test__ubuntu_noble_amd64/12/console

Sending e-mails to: audrow+build.ros2.org@openrobotics.org ros2-buildfarm-humble@googlegroups.com opensource@apex.ai alban.tamisier@apex.ai MiguelCompany@eprosima.com SniderThanYou@gmail.com erik.snider@apex.ai 3488932-albtam@users.noreply.gitlab.com lander.usategui@gmail.com

How can we unsubscribe our personal email addresses from this list?

tfoote:
> How can we unsubscribe our personal email addresses from this list?

The emails are an aggregate of the maintainers listed in the package.xml as well as any committers (from the git commits) who have committed to the branch under tests since the last successful build. Now that the build has gone green it should only be the package maintainers (with one email now saying it was successful to all the people who were notified in the past that it had failed,)  And if there's a regression in the future it will be to the committers with changes since the last successful build. 

So you can control it by either adjusting the package.xml or changing what you use as your commit email.

ErikAtApex:
@tfoote thanks for the explanation.

Now that it has gone green, what do I need to do here to update? I never really learned what this repo is for. Should I submit an MR to update version numbers to more recent tags?

tfoote:
If there's new content making new releases would be great. 

To update a version here you should be running bloom to make a new release and it will automatically open the PR here when it's done for you with the new version. There's an overview here: https://docs.ros.org/en/jazzy/How-To-Guides/Releasing/Releasing-a-Package.html You'll be executing the subsequent releases section but I'd suggest reading through the other two for more context first. Bloom will make all the target platform specific tags in the release repos for you based on the tags that you point it at. 

The only real thing to be managed directly here might be the `source` `version` field which should be the relevant branch for development in that rosdistro, and this is what the dev jobs will pick up on and test against each associated rosdistro.  And we usually recommend the `doc` `version` be the same as the dev branch such that you can fix documentation and it will get rebuilt without needing to make a new release. 

So I'm going to close this. And bloom should open new PRs for new versions. And if you have different branches for development on the different rosdistros you could update those for each distro as is appropriate.

ErikAtApex:
@tfoote thank you!

ErikAtApex:
@tfoote I updated all of the versions, and the build farm has a new error:

https://build.ros2.org/job/Ibin_uJ64__performance_test__ubuntu_jammy_amd64__binary/542/

```
15:29:37 c++: fatal error: Killed signal terminated program cc1plus
```

It looks like the job is being killed? But i don't see a reason why. Any ideas?

tfoote:
Is this a compilation unit that uses a lot of memory? It might have been reaped by the oom killer. 

The farm is busy at the moment and we don't hard enforce memory limits. So it out another parallel build might have run over allocation. It will retry automatically, but I've bumped it to accelerate it too.

ErikAtApex:
@tfoote there's a good amount of template instantiation due to the wide variety of message types, so I would not be surprised if this is the case. Thanks for looking into it.

PR #42464, opened 2024-08-16
Add python3-minimalmodbus-pip

kousheekc:

## Package name:

minimalmodbus

## Package Upstream Source:

https://github.com/pyhys/minimalmodbus

## Purpose of using this:

This dependency is being used for modbus RTU and modbus ASCII communication for Python. This is why I think it's valuable to be added to the rosdep database.

Distro packaging links:

PyPi: [https://pypi.org/project/minimalmodbus/](https://pypi.org/project/minimalmodbus/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7066,6 +7066,19 @@ python3-meson-pip:
   ubuntu:
     pip:
       packages: [meson]
+python3-minimalmodbus-pip:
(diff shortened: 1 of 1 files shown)


audrow (approved):

mjcarroll:

mjcarroll (approved):

PR #41664, opened 2024-06-17
Add python3-openpyxl

SimoneNardi:
Please add the following dependency to the rosdep database.

## Package name:

openpyxl

## Package Upstream Source:

Source code is available at the [link](https://foss.heptapod.net/openpyxl/openpyxl)

## Purpose of using this:

openpyxl is a Python library to read/write Excel 2010 xlsx/xlsm/xltx/xltm files.
Documentation is available at the [link](https://openpyxl.readthedocs.io/en/stable/)

## Links to Distribution Packages

- Debian: 
  - https://packages.debian.org/openpyxl
- Ubuntu: 
  - https://packages.ubuntu.com/openpyxl

PR #42483, opened 2024-08-19
rosdep: openpyxl

130s:

Taking over a closed PR https://github.com/ros/rosdistro/pull/41664

## Package name:

`openpyxl`

## Package Upstream Source:

https://foss.heptapod.net/openpyxl/openpyxl

## Purpose of using this:

The pkg provides an access the data inside the file that is formated in `.xls*`, which can be a source of (large) dataset, and a data output file that can be easily imported on a viewer-friendly (cloud) reader tools.

- Debian: https://pkgs.org/search/?q=openpyxl
  - REQUIRED
- Ubuntu: https://pkgs.org/search/?q=openpyxl
  - REQUIRED
- Fedora: https://pkgs.org/search/?q=openpyxl
  - IF AVAILABLE
- Arch: https://pkgs.org/search/?q=openpyxl
  - IF AVAILABLE
- Gentoo: https://packages.gentoo.org/packages/dev-python/openpyxl
  - IF AVAILABLE
- Alpine: https://pkgs.org/search/?q=openpyxl
  - IF AVAILABLE
- openSUSE: https://pkgs.org/search/?q=openpyxl
- RHEL https://pkgs.org/search/?q=openpyxl

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7552,13 +7552,14 @@ python3-openhsi-pip:
   ubuntu:
     pip:
       packages: [openhsi]
-python3-openpyxl-pip:
-  debian:
-    pip:
-      packages: [openpyxl]
-  ubuntu:
-    pip:
-      packages: [openpyxl]
+python3-openpyxl:
+  alpine: [py3-openpyxl]
+  arch: [python-openpyxl]
+  debian: [python3-openpyxl]
+  fedora: [python3-openpyxl]
+  gentoo: [dev-python/openpyxl]
+  opensuse: [python3-openpyxl]
+  ubuntu: [python3-openpyxl]
 python3-opentelemetry-api-pip:
   '*':
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7552,6 +7552,13 @@ python3-openhsi-pip:
   ubuntu:
     pip:
       packages: [openhsi]
+python3-openpyxl-pip:
+  debian:
+    pip:
+      packages: [openpyxl]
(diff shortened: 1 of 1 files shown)


audrow:
It seems like there's a RHEL package that could be added.

rosdep/python.yaml
```diff
@@ -7552,6 +7552,15 @@ python3-openhsi-pip:
...
     pip:
       packages: [openhsi]
+python3-openpyxl:
+  alpine: [py3-openpyxl]
+  arch: [python-openpyxl]
+  debian: [python3-openpyxl]
+  fedora: [python3-openpyxl]
+  gentoo: [dev-python/openpyxl]
+  opensuse: [python3-openpyxl]
+  rhel: ['python%{python3_pkgversion}-openpyxl']
```
130s, line 7562: > It seems like there's a RHEL package that could be added.

Thanks for catching. Added RHEL.

Since I see some entries for RHEL uses this `%{python3_pkgversion}` notation, I followed that without understanding what for it is needed (and also not understanding why non-RHEL entries don't use that).

rosdep/python.yaml
```diff
@@ -7552,6 +7552,15 @@ python3-openhsi-pip:
...
```
clalancette, line 7562: Yeah, this is a vestige which we don't need anymore:

```suggestion
  rhel: [python3-openpyxl]
```

clalancette (changes requested):

clalancette (approved):

alsora (approved):

PR #42504, opened 2024-08-20
Add python3-easyocr

Plexiglas:

python3-easyocr:
Package Upstream Source:
https://github.com/JaidedAI/EasyOCR

Purpose of using this:
Used for optical character detection and recognition.

Distro packaging links:

pip https://pypi.org/project/easyocr/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5688,6 +5688,16 @@ python3-easydict:
     focal:
       pip:
         packages: [easydict]
+python3-easyocr:
+  debian:
+    pip:
+      packages: [easyocr]
+  fedora:
+    pip:
+      packages: [easyocr]
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


mjcarroll:

mjcarroll (approved):

rosdep/python.yaml
```diff
@@ -5688,6 +5688,16 @@ python3-easydict:
     focal:
       pip:
         packages: [easydict]
+python3-easyocr:
```
clalancette, line 5691:
```suggestion
python3-easyocr-pip:
```

clalancette (changes requested):

clalancette:
There's been no response to this in a month, so closing it out.  If you'd still like to get this in, please feel free to respond to the comments and reopen.

PR #42560, opened 2024-08-23
Add Turtle Nest for indexing

Jannkar:
# Please Add Turtle Nest to be indexed in the rosdistro.

humble, iron, jazzy, rolling

# The source is here:

https://github.com/Jannkar/turtle_nest

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -10018,6 +10018,12 @@ repositories:
       url: https://github.com/wep21/turbojpeg_compressed_image_transport.git
       version: humble
     status: maintained
+  turtle_nest:
+    source:
+      type: git
+      url: https://github.com/Jannkar/turtle_nest.git
+      version: main
+    status: developed
   turtlebot3:
     doc:
       type: git
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -8084,6 +8084,12 @@ repositories:
       url: https://github.com/wep21/turbojpeg_compressed_image_transport.git
       version: rolling
     status: maintained
+  turtle_nest:
+    source:
+      type: git
+      url: https://github.com/Jannkar/turtle_nest.git
+      version: main
+    status: developed
   turtlebot3_msgs:
     doc:
       type: git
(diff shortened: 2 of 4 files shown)


clalancette:

  - [x] Top level license file: https://github.com/Jannkar/turtle_nest/blob/main/LICENSE
- [x] License is [OSI-approved](https://opensource.org/licenses): https://opensource.org/license/apache-2-0
- [x] Public source repo: https://github.com/Jannkar/turtle_nest
- [x] Source repository contains ROS packages - yes, though the organization is a bit different than we are used to.  There are 4 top-level directories, but only one of them contains a ROS package (turtle_nest)
- [x] Each package meets [REP-144](https://www.ros.org/reps/rep-0144.html) naming conventions - The package name doesn't exactly explain what it does, but I also don't think it violates REP-144.  @alsora what do you think?

clalancette (approved):

clalancette:
@Jannkar Due to changes unrelated to your PR, this needs a rebase to pass CI.  Thanks in advance!

alsora:
> Each package meets [REP-144](https://www.ros.org/reps/rep-0144.html) naming conventions - The package name doesn't exactly explain what it does, but I also don't think it violates REP-144. @alsora what do you think?

I agree with your assesment.
I think we have a lot of packages that use a "codename" as package name, which is not descriptive at all of what the package does.
Prohibiting that, would cause a proliferation of packages with either very similar names or very long names.

alsora (approved):

Jannkar:
Thank you! I've now rebased.

I was also thinking about the package structure, as I noticed that usually the repo seems to be either a single package or set of packages. I chose this structure, since I wanted to avoid including for example docker stuff and images into the package itself. Is it fine like this?

mikeferguson:
> I think we have a lot of packages that use a "codename" as package name, which is not descriptive at all of what the package does (unless you know the project).

To be fair, ROS is not really an "operating system". Further, most of the core packages aren't really all that "robot" specific either...

PR #42617, opened 2024-08-29
fix Fedora/RHEL rosdep for `libudev-devel`

Grvzard:
https://github.com/mcatalancid/libudev
The udev development has moved to systemd

  - https://packages.fedoraproject.org/pkgs/systemd/systemd-devel/
  - https://pkgs.org/search/?q=systemd-devel


clalancette (approved):
Makes sense to me; I believe this has been the case going all the way back to Fedora 18, and at least RHEL-8, so this should be safe to do.

mjcarroll (approved):

PR #42722, opened 2024-09-04
robot_upstart: 1.0.3-1 in 'jazzy/distribution.yaml' [bloom]

civerachb-cpr:
Increasing version of package(s) in repository `robot_upstart` to `1.0.3-1`:

- upstream repository: https://github.com/clearpathrobotics/robot_upstart.git
- release repository: https://github.com/clearpath-gbp/robot_upstart-release.git
- bloom version: `0.12.0`

## robot_upstart

```
* Source workspace after exporting domain ID and RMW
* Contributors: Roni Kreinin
```

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -6196,6 +6196,21 @@ repositories:
       url: https://github.com/ros/robot_state_publisher.git
       version: jazzy
     status: maintained
+  robot_upstart:
+    doc:
+      type: git
+      url: https://github.com/clearpathrobotics/robot_upstart.git
+      version: foxy-devel
+    release:
+      tags:
+        release: release/jazzy/{package}/{version}
+      url: https://github.com/clearpath-gbp/robot_upstart-release.git
+      version: 1.0.3-1
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


jazzy/distribution.yaml
```diff
@@ -6196,6 +6196,21 @@ repositories:
...
+      version: foxy-devel
+    release:
+      tags:
+        release: release/jazzy/{package}/{version}
+      url: https://github.com/clearpath-gbp/robot_upstart-release.git
+      version: 1.0.3-1
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/robot_upstart.git
+      version: foxy-devel
```
christophebedard, line 6212: I'm assuming that `foxy-devel` here for Jazzy is intentional, since it's also used for Galactic and Humble.

christophebedard:
This is a new package in Jazzy but it's already in some other distros, so skipping the checks.

jazzy/distribution.yaml
civerachb-cpr: That's correct; there were no changes needed since the Foxy release, so the same branch is still being used for all ROS 2 releases.

PR #42823, opened 2024-09-12
added python3 noisereduce

rafa-martin:

## Package name:

- `noisereduce`

## Package Upstream Source:

- https://github.com/timsainb/noisereduce

## Purpose of using this:

Noise reduction in python using spectral gating (speech, bioacoustics, audio, time-domain signals).

Distro packaging links:

- Python: https://pypi.org/project/noisereduce/


sloretz (dismissed):
LGTM with one change applied

clalancette (approved):

audrow (approved):

PR #42827, opened 2024-09-12
Remove mrpt2 package

jlblancoc:
Deprecated by packages mrpt_ros and python_mrpt_ros.  There are now no packages depending on mrpt2 (checked in Jenkins, "downstream projects"), all are ported to mrpt_ros.

### DIFF
--- a/iron/distribution.yaml
+++ b/iron/distribution.yaml
@@ -3616,14 +3616,6 @@ repositories:
       url: https://github.com/ika-rwth-aachen/mqtt_client.git
       version: main
     status: maintained
-  mrpt2:
-    release:
-      tags:
-        release: release/iron/{package}/{version}
-      url: https://github.com/ros2-gbp/mrpt2-release.git
-      version: 2.13.5-1
-    status: end-of-life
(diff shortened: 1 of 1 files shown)


sloretz:
ROS Iron is almost end-of-life (November 2024). How about we leave it in Iron just in case removing it causes issues without much time to fix them?

nuclearsandwich:
> ROS Iron is almost end-of-life (November 2024). How about we leave it in Iron just in case removing it causes issues without much time to fix them?

+1, let's let Iron reach end-of-support without removing this package.

> There are now no packages depending on mrpt2 (checked in Jenkins, "downstream projects"), all are ported to mrpt_ros.

This accounts for all of the publicly released packages, but not packages which are privately used.

jlblancoc:
Ok, close this PR then if you see it appropriate.

My main actual motivation to completely delete it is that this old,
monolithic version is so big that it quite often timeouts in the farms, and
I guess that's painful for all of you too 😀
But I have nothing against going on and leaving it with the EOL notice as
is.

PR #42870, opened 2024-09-16
rosdep: add python3-ikpy-pip

severin-lemaignan:

## Package name:

`ikpy`

## Package Upstream Source:

https://github.com/Phylliade/ikpy

## Purpose of using this:

ikpy is a lightweight Python-based inverse kinematics library. It is used by eg [`hri_body_detect`](https://github.com/ros4hri/hri_body_detect) to create kinematically-consistent human models on the fly.

https://pypi.org/project/ikpy/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6576,6 +6576,10 @@ python3-ignition-math6:
   ubuntu:
     focal: [python3-ignition-math6]
     jammy: [python3-ignition-math6]
+python3-ikpy-pip:
+  '*':
+    pip:
+      packages: [ikpy]
 python3-imageio:
   debian: [python3-imageio]
   fedora: [python3-imageio]


humble/distribution.yaml
```diff
@@ -6412,6 +6412,20 @@ repositories:
       url: https://github.com/mrpt-ros-pkg/pose_cov_ops.git
       version: master
     status: maintained
+  privacy_msgs:
```
sloretz, line 6415: These look like the changes from #42828. Please rebase this PR so it includes only the new rosdep key.

rosdep/python.yaml
```diff
@@ -6576,6 +6576,10 @@ python3-ignition-math6:
   ubuntu:
     focal: [python3-ignition-math6]
     jammy: [python3-ignition-math6]
+python3-ikpy-pip:
+  '*':
+    pip:
```
sloretz, line 6581: Standard pip disclaimer: ROS packages that depend on pip keys cannot be released into a ROS distribution. They can only be depended on by from-source builds. Because of this, system packages are highly preferred to pip packages.

sloretz (dismissed):

humble/distribution.yaml
severin-lemaignan: oops. Fixed.

rosdep/python.yaml
severin-lemaignan: yes, fully aware... 🤷‍♂️

severin-lemaignan: `ikpy` is unfortunately not packaged upstream.

sloretz (approved):

clalancette (approved):

PR #42927, opened 2024-09-24
performance_test: 2.3.0-1 in 'humble/distribution.yaml' [bloom]

ErikAtApex:
Increasing version of package(s) in repository `performance_test` to `2.3.0-1`:

- upstream repository: https://gitlab.com/ApexAI/performance_test.git
- release repository: https://github.com/ros2-gbp/performance_test-release.git
- bloom version: `0.12.0`
- previous version for package: `1.2.1-1`

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -6051,7 +6051,7 @@ repositories:
     doc:
       type: git
       url: https://gitlab.com/ApexAI/performance_test.git
-      version: 1.2.1
+      version: master
     release:
       tags:
         release: release/humble/{package}/{version}
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -6056,7 +6056,7 @@ repositories:
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/ros2-gbp/performance_test-release.git
-      version: 1.2.1-1
+      version: 2.3.0-1
     source:
       type: git
       url: https://gitlab.com/ApexAI/performance_test.git


ErikAtApex:
I am not very experienced with bloom. Should the `performance_test/doc/version` entry also be updated to 2.3.0-1? I don't really understand why it is still 1.2.1-1.

clalancette:
> I am not very experienced with bloom. Should the `performance_test/doc/version` entry also be updated to 2.3.0-1? I don't really understand why it is still 1.2.1-1.

In almost all circumstances, the `doc` branch should match the `source` branch (there are cases where it makes sense for them to be different, but they are fairly uncommon).  So I'd update it to be `master` in this case.

ErikAtApex:
Thanks @clalancette . I changed it to `master`, and I'll do the same on the other two MRs.

audrow:
You might not have access, since you're not a maintainer. I reran jobs.

audrow:
It looks like we're still getting CI errors:

```
```

ErikAtApex:
Thanks @audrow . Why would we get that CI error on this MR only, and not the other two? They're all pointing to the same tag, and package.xml does exist:

https://gitlab.com/ApexAI/performance_test/-/blob/master/performance_test/package.xml?ref_type=heads

audrow:
I am not sure 🤔

ErikAtApex:
@audrow Let's try to isolate it, I suppose. I'll revert the rolling commit, and open a new MR for it. Maybe the humble will pass and rolling will fail, or maybe they will pass independently :shrug:

ErikAtApex:
@audrow the new MR has the same problem, but at least this MR is now green.

PR #42984, opened 2024-09-29
CI: Add workflow that builds and releases cache on new tags

oysstu:
This PR contains a simple workflow to generate and release a cache from each pushed tag. This would make targeting specific snapshots using rosdep and rosinstall_generator more convenient. The compressed archive is fairly small and does not take long to generate since it bases it off the current index cache.

Let me know if this makes sense to have running and if any changes are needed.


oysstu:
> Overall doing this to have the cache items with the tags sounds interesting. The tools were designed to work without a cache being necessary but they will be much slower. What do you use for configuring the tools to find the release artifacts?

Not sure I understand the question. Are you referring to the release artifacts output by this workflow?  They should be available at the following url (if using bash).
```
<repo>/releases/download/${TAG}/${TAG%%/*}-${TAG#*/}-cache.yaml.gz
```

I've been using this with rosinstall_generator by modifying the cache url in index-v4.yaml and setting ROSDISTRO_INDEX_URL. If I check out this repository at a specified tag and leave the cache url unmodified, it will output the wrong versions (e.g. jazzy/2024-05-23 should generate rosbag2 with version 0.26.3, but it outputs 0.26.5). I just realized that this behavior is probably a bug in rosintall_generator.

Even if this is fixed, I need the cache, because I'm interested in the output using the --upstream flag.

mjcarroll:
@tfoote @cottsay @oysstu What's the status here?

ros-discourse:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-march-17-2026/53310/1

tfoote:
This is a good direction. There's superseding work in #50112 so I'm going to close this one.

PR #43003, opened 2024-10-01
Added adi_3dtof_image_stitching package

raebelchristo-adi:

## Package name:

adi_3dtof_image_stitching

## Package Upstream Source:

https://github.com/analogdevicesinc/adi_3dtof_image_stitching/tree/v1.1.0

## Purpose of using this:

`adi_3dtof_image_stitching` dependency allows a user to stitch together multiple depth and IR images published by the ADI's ADTF3175D ToF sensor. This has a significance in wide angle applications to increase the field of view.

# Checks

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -77,6 +77,11 @@ repositories:
       url: https://github.com/ros-acceleration/adaptive_component.git
       version: rolling
     status: developed
+  adi_3dtof_image_stitching:
+    source:
+      type: git
+      url: https://github.com/analogdevicesinc/adi_3dtof_image_stitching.git
+      version: humble-devel
   adi_tmcl:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -79,7 +79,8 @@ repositories:
     source:
       type: git
       url: https://github.com/analogdevicesinc/adi_3dtof_image_stitching.git
-      version: v1.1.0
+      version: noetic-devel
+    status: maintained
   adi_tmc_coe:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -75,6 +75,11 @@ repositories:
       url: https://github.com/ros/actionlib.git
       version: noetic-devel
     status: maintained
+  adi_3dtof_image_stitching:
+    source:
+      type: git
+      url: https://github.com/analogdevicesinc/adi_3dtof_image_stitching.git
+      version: v1.1.0
   adi_tmc_coe:
     doc:
       type: git


sloretz:
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2024-10-03/)

MichaelOrlov (changes requested):
The license should be one of the open-source ones listed in https://opensource.org/licenses (preferably one of the "Popular licenses").
However, the license file https://github.com/analogdevicesinc/adi_3dtof_image_stitching/blob/v1.1.0/LICENSE
has an Analog Devices license which is not listed in the https://opensource.org/licenses

raebelchristo-adi:
We have used the **BSD-3-Clause** license which comes under the "Popular/Strong Community" Category as listed in [https://opensource.org/licenses](https://opensource.org/licenses). We only replaced the year and copyright holder placeholders with the corresponding year and the name of our organization.

MichaelOrlov:
@raebelchristo-adi I don't know how it happened, but https://github.com/analogdevicesinc/adi_3dtof_image_stitching/blob/v1.1.0/LICENSE have ADI license
image
And this is what gitlab officially shows under the what license this project is.

However, at the same time in the https://github.com/analogdevicesinc/adi_3dtof_image_stitching/blob/v1.1.0/package.xml
mentioned BSD license.
image

The licenses must be the same. Please adjust.

As regards your latest commit with changing the source version to the `noetic-devel` branch instead of the pinned version with a  tag. We usually do not allow this "workaround".

raebelchristo-adi:
> The licenses must be the same. Please adjust.

Could it be that the extra lines appended at the bottom of the LICENSE file that makes github think it is an ADI license rather than BSD-3? In that case I might need to check with an internal resource regarding how this can be altered.

> As regards your latest commit with changing the source version to the `noetic-devel` branch instead of the pinned version with a tag. We usually do not allow this "workaround".

We had an edit in the package.xml and CMakelists.txt that was merged into the noetic-devel branch. Hence we wanted this to be the main source branch as all maintenance happens here. Do you have any other suggestions if this workaround is not allowed?

raebelchristo-adi:
Regarding license changes. we are reviewing the alterations for the LICENSE file internally. I will update about it here.

MichaelOrlov (commented):
@sloretz @clalancette Please correct me if I am wrong, if I understand correctly we shouldn't allow sources of the packages to track development branches like https://github.com/analogdevicesinc/adi_3dtof_image_stitching/tree/humble-devel unless some extraordinary cases.
The `source: version:` should refer to the tag in the repo associated with the release for the package. 
Otherwise, if source version will be set to track the development branch and, If it will be a regression, it will be no way to quickly revert changes by the ROS Boss when he preparing  be-weekly synch.

sloretz:
@MichaelOrlov The `source: version:` should be [the development branch from which release are made](https://docs.ros.org/en/rolling/How-To-Guides/Releasing/Index-Your-Packages.html#make-changes-to-your-fork) rather than a tag. The `source` block is used by tools like `rosinstall_generator --upstream-development` to get the latest unreleased sources for a package.

> Otherwise, if source version will be set to track the development branch and, If it will be a regression, it will be no way to quickly revert changes by the ROS Boss when he preparing be-weekly synch.

Syncs are only affected by content in the `release` block.

MichaelOrlov:
@sloretz Thanks for the clarification.
@raebelchristo-adi As soon as licensing issue will be resolved this PR will be good to go.

raebelchristo-adi:
The license of the repository has been updated. It now shows as BSD-3-Clause.

cottsay (approved):

MichaelOrlov (approved):

clalancette (approved):

PR #43171, opened 2024-10-14
Add python3-bloom to rosdep/python.yaml

frankkusters:

## Package name:

python3-bloom

## Package Upstream Source:

https://github.com/ros-infrastructure/bloom

## Purpose of using this:

Bloom is a release automation tool, designed to make generating platform specific release artifacts from source projects easier.

  - https://packages.debian.org/bookworm/python3-bloom
  - https://packages.ubuntu.com/jammy/python3-bloom
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - https://rhel.pkgs.org/8/epel-x86_64/python3-bloom-0.11.2-2.el8.noarch.rpm.html

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5029,6 +5029,12 @@ python3-bitstring:
     '*': [python3-bitstring]
     '7': [python36-bitstring]
   ubuntu: [python3-bitstring]
+python3-bloom:
+  debian: [python3-bloom]
+  rhel:
+    '*': null
+    '8': [python3-bloom]
+  ubuntu: [python3-bloom]
 python3-bluerobotics-ping-pip:
   debian:
     pip:


clalancette (changes requested):
I think we need to be really careful with this one, since it is provided both by upstream Ubuntu, and also by us.

Pinging @cottsay for a look.

cottsay (approved):
I don't see a very strong use case to have this here, but I also don't have any reason to object to adding it.

clalancette (approved):

frankkusters:
> I don't see a very strong use case to have this here, but I also don't have any reason to object to adding it.

The reason for adding this is that we package pretty much _everything_ as ROS packages in our project.* We also have a package for our build scripts, which depend on bloom.

\* It makes setting up a new developer machine incredibly easy: `rosdep install`, done.

PR #43173, opened 2024-10-14
opencv-contrib-python added to rosdep/python.yaml

pep248:

## Package name:

opencv-contrib-python

## Package Upstream Source:
[opencv-contrib-python](https://pypi.org/project/opencv-contrib-python/)

## Purpose of using this:

Some libraries for marker detection in OpenCV are only available in said package.
This package is automatically installed when using `pip install opencv-python` but not when using `apt-get install python-opencv` or  `apt-get install python3-opencv`.
Since the request to make OpenCV installable through pip was discarded [here](https://github.com/ros/rosdistro/pull/41870), I would like to, at least, have the opencv-contrib-python package available.

Distro packaging links:

- PyPi: https://pypi.org/project/opencv-contrib-python/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -2602,6 +2602,16 @@ python-opencv:
     '7': [opencv-python]
   slackware: [opencv]
   ubuntu: [python-opencv]
+python-opencv-contrib-pip:
+  debian:
+    pip:
+      packages: [opencv-contrib-python]
+  fedora:
+    pip:
+      packages: [opencv-contrib-python]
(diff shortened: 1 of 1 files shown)


methylDragon (approved):
Okay because I do not see an Python opencv-contrib package on the debian package or ubuntu lists

pep248:
Why stale?
What else do I need for the merge to be successfull?

agonzat:
Hi, what is blocking this pr?

sloretz (approved):

PR #43178, opened 2024-10-14
Add dh_python to rosdep/base.yaml

frankkusters:

## Package name:

dh_python

## Package Upstream Source:

https://salsa.debian.org/python-team/tools/dh-python

## Purpose of using this:

dh-python is used when a Debian package is created from a ROS package using the scripts generated by [bloom](https://github.com/ros/rosdistro/pull/43171).

  - https://packages.debian.org/bookworm/dh-python
  - https://packages.ubuntu.com/noble/dh-python

This package is only applicable to distributions which are Debian-based.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -812,6 +812,9 @@ dfu-util:
   gentoo: [app-mobilephone/dfu-util]
   nixos: [dfu-util]
   ubuntu: [dfu-util]
+dh-python:
+  debian: [dh-python]
+  ubuntu: [dh-python]
 disper:
   debian: [disper]
   nixos: [disper]


clalancette (commented):
I'm not totally convinced that this is a good candidate for a rosdep key.  As you say in your description, it is Debian-only, so it will never have equivalents on other platforms.

Could you explain a bit why you want to add it?

frankkusters:
> I'm not totally convinced that this is a good candidate for a rosdep key. As you say in your description, it is Debian-only, so it will never have equivalents on other platforms.
> 
> Could you explain a bit why you want to add it?

We package pretty much _everything_ as ROS packages in our project. It makes setting up a new developer machine incredibly easy with `apt install` (binary) or `rosdep install` (from source).

We also have packages for our build scripts, which depend on bloom and dh-python. Hence the need for us to have them as rosdep keys.

In a sense we're solving the same problem as the `ros-dev-tools` package [1]:

> I’ve always been annoyed with the long list of dependencies that had to be installed in order to develop ROS software. Spurred by @mikeferguson 's tweet I finally tried to do something about it.
> [..]
> `ros-dev-tools` has the following dependencies:
> [..]
> The new variants are now part of the standard installation path in the documentation.

With the added benefit that if `dh-python` is added as a rosdep key, we can fully use ROS infrastructure for our dependencies, keeping the number of tools to a minimum.

[1]: https://discourse.ros.org/t/ros-developer-tools-now-in-binary-form/29802 ""

clalancette:
> With the added benefit that if `dh-python` is added as a rosdep key, we can fully use ROS infrastructure for our dependencies, keeping the number of tools to a minimum.

Got it.  Thanks for the explanation.  I don't really have a strong objection, so I'll go ahead and approve this.

clalancette (approved):

marcoag (approved):

PR #43190, opened 2024-10-15
Add pytest-order

ruffsl:
- https://pypi.org/project/pytest-order/

## Package name:

python3-pytest-order

## Package Upstream Source:

- https://github.com/pytest-dev/pytest-order
- https://pypi.org/project/pytest-order/

## Purpose of using this:

This is pytest plugin to run your tests in a specific order, such combining with launch-pytest to run smoke tests before longer or more thorough tests, to enable more efficient fail fast testing strategies, saving time on starting test the you know would fail, without having to consolidating all your test files and test cases into a monolithic fixture.

> Note: pytest-order is the successor to `pytest-ordering`

Distro packaging links:

  - https://packages.debian.org/bookworm/python3-pytest-order
  - https://packages.ubuntu.com/noble/python3-pytest-order
  - IF AVAILABLE
  - https://archlinux.org/packages/extra/any/python-pytest-order/
  - https://packages.gentoo.org/packages/dev-python/pytest-order
  - IF AVAILABLE
  - IF AVAILABLE
  - https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=pytest-order
  - https://software.opensuse.org/package/python3-pytest-order
  - IF AVAILABLE

ROSDISTRO NAME

# The source is here:

https://github.com/pytest-dev/pytest-order

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8775,7 +8775,11 @@ python3-pytest-order:
   nixos: [python3Packages.pytest-order]
   openembedded: [python3-pytest-order@meta-ros-common]
   opensuse: [python3-pytest-order]
-  ubuntu: [python3-pytest-order]
+  ubuntu:
+    '*': [python3-pytest-order]
+    focal:
+      pip:
+        packages: [pytest-order]
 python3-pytest-pylint:
   arch:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8773,7 +8773,6 @@ python3-pytest-order:
   debian: [python3-pytest-order]
   gentoo: [dev-python/pytest-order]
   nixos: [python3Packages.pytest-order]
-  openembedded: [python3-pytest-order@meta-ros-common]
   opensuse: [python3-pytest-order]
   ubuntu:
     '*': [python3-pytest-order]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8769,18 +8769,12 @@ python3-pytest-mock:
   rhel: ['python%{python3_pkgversion}-pytest-mock']
   ubuntu: [python3-pytest-mock]
 python3-pytest-order:
-  alpine: [py3-pytest-order]
   arch: [python-pytest-order]
   debian: [python3-pytest-order]
-  fedora: [python3-pytest-order]
   gentoo: [dev-python/pytest-order]
   nixos: [python3Packages.pytest-order]
   openembedded: [python3-pytest-order@meta-ros-common]
   opensuse: [python3-pytest-order]
-  osx:
-    pip:
-      packages: [pytest-order]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -8768,6 +8768,14 @@ python3-pytest-mock:
       packages: [pytest-mock]
   rhel: ['python%{python3_pkgversion}-pytest-mock']
   ubuntu: [python3-pytest-mock]
+python3-pytest-order:
+  arch: [python-pytest-order]
+  debian: [python3-pytest-order]
+  gentoo: [dev-python/pytest-order]
+  nixos: [python3Packages.pytest-order]
```
ruffsl, line 8775: Not sure what to put here, as nix namespaces the package with `311` or `312`, not just `3`.

rosdep/python.yaml
```diff
@@ -8768,6 +8768,17 @@ python3-pytest-mock:
...
+  opensuse: [python3-pytest-order]
```
ruffsl, line 8776: Note sure why the linter is failing for opensuse, as the URL lists support for `SUSE SLE-15-SP2`.

rosdep/python.yaml
```diff
@@ -8768,6 +8768,17 @@ python3-pytest-mock:
...
+  ubuntu:
+    '*': [python3-pytest-order]
+    focal:
+      pip:
+        packages: [pytest-order]
```
ruffsl, line 8781: Is this how we should handle focal as the oldest active LTS?

rosdep/python.yaml
clalancette: I'd just remove it in this case.

rosdep/python.yaml
clalancette: I think they have "aliases", or something like that, that make this work.  CI seems happy enough with it.

That said, nixos isn't a required platform, so if you are uncomfortable with it, just remove it.

rosdep/python.yaml
clalancette: Yes.  If you don't care about focal working at all, you can also do:

```
focal: null
```

clalancette (changes requested):

rosdep/python.yaml
ruffsl: If CI is happy, I'm fine.

rosdep/python.yaml
ruffsl: Removed.

rosdep/python.yaml
ruffsl: Simplified.

clalancette (approved):

ruffsl:
Shoo stale bot, shoo. Folks are just getting back from ROSCon.

cottsay (approved):

PR #43211, opened 2024-10-16
libfranka: 0.14.2-1 in 'humble/distribution.yaml' [bloom]

AndreasKuhner:
Increasing version of package(s) in repository `libfranka` to `0.14.2-1`:

- upstream repository: https://github.com/frankaemika/libfranka.git
- release repository: https://github.com/frankaemika/libfranka-release.git
- bloom version: `0.12.0`
- previous version for package: `0.13.6-1`

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -4036,11 +4036,15 @@ repositories:
       version: master
     status: maintained
   libfranka:
+    doc:
+      type: git
+      url: https://github.com/frankaemika/libfranka-release.git
+      version: release/foxy/libfranka
     release:
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/frankaemika/libfranka-release.git
-      version: 0.13.6-1
+      version: 0.14.2-1
     source:
       type: git
       url: https://github.com/frankaemika/libfranka.git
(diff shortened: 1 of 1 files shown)


humble/distribution.yaml
```diff
@@ -4036,11 +4036,15 @@ repositories:
       version: master
     status: maintained
   libfranka:
+    doc:
+      type: git
+      url: https://github.com/frankaemika/libfranka-release.git
```
sloretz, line 4041: Often the doc entry [is the source repo](https://docs.ros.org/en/rolling/How-To-Guides/Documenting-a-ROS-2-Package.html#ci-docs-ros-org) rather than the release repo so that documentation can be updated without making a new release. Are you sure you want the documentation to be built from the release repo?

sloretz (commented):
[Holding for Humble sync](https://discourse.ros.org/t/preparing-for-humble-sync-2024-10-17/40097)

humble/distribution.yaml
AndreasKuhner: Yes true -> redirected.

Other question, that now the version changed is not a problem for the existing releases right? Both the 0.13.6 and 0.14.2 will be available...?

clalancette: No, we only provide the latest version of the package.  Old versions are removed during the sync process.

PR #43327, opened 2024-11-01
Add nanobind, an alternative to pybind11

BrunoB81HK:

## Package name:

nanobind-dev

## Package Upstream Source:

[nanobind](https://github.com/wjakob/nanobind)

## Purpose of using this:

Nanobind is a lightweight alternative to pybind11. It can be useful to bind c++ code to python and vice-versa. 

- [Arch](https://archlinuxarm.org/packages/any/nanobind)
- [Debian Sid](https://packages.debian.org/sid/nanobind-dev)
- [Ubuntu Oracular](https://packages.ubuntu.com/oracular/nanobind-dev)
- [Ubuntu Noble](https://packages.ubuntu.com/noble/nanobind-dev)
- [FreeBSD 13-14-15](https://www.freshports.org/devel/nanobind)
- [PIPY](https://pypi.org/project/nanobind/)

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7228,9 +7228,6 @@ nanobind-dev:
     "*":
       pip:
         packages: [nanobind]
-  "*":
-    pip:
-      packages: [nanobind]
 nanoflann:
   debian:
     '*': [libnanoflann-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7218,7 +7218,11 @@ muparser:
   ubuntu: [libmuparser-dev]
 nanobind-dev:
   arch: [nanobind]
-  debian: [nanobind-dev]
+  debian:
+    '*': [nanobind-dev]
+    bookworm:
+      pip:
+        packages: [nanobind]
   freebsd: [nanobind]
   ubuntu:
     noble: [nanobind-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7225,9 +7225,11 @@ nanobind-dev:
         packages: [nanobind]
   freebsd: [nanobind]
   ubuntu:
-    noble: [nanobind-dev]
-    oracular: [nanobind-dev]
-    "*":
+    '*': [nanobind-dev]
+    focal:
+      pip:
+        packages: [nanobind]
+    jammy:
       pip:
         packages: [nanobind]
 nanoflann:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7217,9 +7217,7 @@ muparser:
   opensuse: [muparser-devel]
   ubuntu: [libmuparser-dev]
 nanobind-dev:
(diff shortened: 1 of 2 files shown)


rosdep/python.yaml
```diff
@@ -7271,6 +7271,10 @@ python3-myst-parser-pip:
   ubuntu:
     pip:
       packages: [myst-parser]
+python3-nanobind-pip:
```
clalancette, line 7274: So it looks like this package is available on at least some distributions: https://pkgs.org/search/?q=nanobind .

So I'm going to suggest renaming this key to `python3-nanobind`, fill it in for the new distributions that have it, and fall back to the `pip` key for those that do not.  You won't be able to use this key to release things on the buildfarm, but at least that will be possible in the future.

clalancette (changes requested):

BrunoB81HK:
After looking at `pybind11-dev`, think i will move the key to [`base.yaml`](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml). The package is not really a python package but more of a development package.

Also, I'm not sure the changes I made regarding the distros are good.

rosdep/base.yaml
```diff
@@ -7216,6 +7216,18 @@ muparser:
...
+    pacman:
+      packages: [nanobind]
+  debian: [nanobind-dev]
+  freebsd: [nanobind]
+  ubuntu:
+    noble: [nanobind-dev]
+    oracular: [nanobind-dev]
+    "*":
+      pip:
+        packages: [nanobind]
```
clalancette, line 7230: We should do this the other way around so that this will work into the future:

```suggestion
  ubuntu:
    '*': [nanobind-dev]
    focal:
      pip:
        packages: [nanobind]
    jammy:
      pip:
        packages: [nanobind]
```

rosdep/base.yaml
```diff
@@ -7216,6 +7216,18 @@ muparser:
   nixos: [muparser]
   opensuse: [muparser-devel]
   ubuntu: [libmuparser-dev]
+nanobind-dev:
+  arch:
+    pacman:
+      packages: [nanobind]
+  debian: [nanobind-dev]
```
clalancette, line 7223:
```suggestion
  debian:
    '*': [nanobind-dev]
    bookworm:
      pip:
        packages: [nanobind]
```

rosdep/base.yaml
```diff
@@ -7216,6 +7216,18 @@ muparser:
   nixos: [muparser]
   opensuse: [muparser-devel]
   ubuntu: [libmuparser-dev]
+nanobind-dev:
+  arch:
+    pacman:
+      packages: [nanobind]
```
clalancette, line 7222:
```suggestion
  arch: [nanobind]
```

clalancette (dismissed):

sloretz (approved):

clalancette (approved):

PR #43421, opened 2024-11-08
migrate grasping_msgs to ros2-gbp (jazzy)

mikeferguson:


christophebedard:
The new release repo seems to have the same content, so it looks good to me!

I don't think it matters here/right now, but you should also update the `release_repo_url` values in your `tracks.yaml` file: https://github.com/ros2-gbp/grasping_msgs-release/blob/8dfe269f6edb3137c6d60267ab0cf26cf0ab4c6c/tracks.yaml

mikeferguson:
@christophebedard thanks for the heads up about the tracks

PR #43482, opened 2024-11-13
Add python3-ollama-pip to rosdistro

Barry-Xu-2018:

## Package name:

python3-ollama-pip

## Package Upstream Source:

https://github.com/ollama/ollama-python

## Purpose of using this:

This dependency allows access to ollama's servers for robot AI processing

Distro packaging links:

    - pip
        - https://pypi.org/project/ollama/
   - pip
        - https://pypi.org/project/ollama/
    - pip
        - https://pypi.org/project/ollama/
    - pip
        - https://pypi.org/project/ollama/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7485,6 +7485,19 @@ python3-odrive-pip:
   ubuntu:
     pip:
       packages: [odrive]
+python3-ollama-pip:
+  debian:
+    pip:
+      packages: [ollama]
+  fedora:
+    pip:
+      packages: [ollama]
+  rhel:
+    pip:
+      packages: [ollama]
+  ubuntu:
+    pip:
+      packages: [ollama]
 python3-onnxruntime-gpu-pip:
   arch:
     pip:


rosdep/python.yaml
```diff
@@ -7485,6 +7485,19 @@ python3-odrive-pip:
...
+python3-ollama-pip:
+  debian:
+    pip:
+      packages: [ollama]
+  fedora:
+    pip:
+      packages: [ollama]
+  rhel:
+    pip:
+      packages: [ollama]
```
fujitatomoya, line 7497: @Barry-Xu-2018 `ros2ai` only supports ubuntu platform only. technically i do not have any platform specific code in the implementation, but that is so far what i verified. so probably, we would want to remove untested platform here?

Barry-Xu-2018: @fujitatomoya  

Yes. 
I considered that other packages might also use it, so I didn't limit it to the Ubuntu platform. Despite the different platforms, since it's installed via pip, it should use the same Python package.

sloretz (approved):

fujitatomoya (approved):
lgtm!

@sloretz thank you very much for the review.

PR #43485, opened 2024-11-13
Add python3-validators to rosdistro

Barry-Xu-2018:

## Package name:

python3-validators-pip

## Package Upstream Source:

https://github.com/python-validators/validators/

## Purpose of using this:

For robot AI processing, while using AI python packages like openai or ollama, they will rely on this python package.

Distro packaging links:

    - pip
        - https://pypi.org/project/validators/
   - pip
        - https://pypi.org/project/validators/
    - pip
        - https://pypi.org/project/validators/
    - pip
        - https://pypi.org/project/validators/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -10257,7 +10257,7 @@ python3-validators:
   debian: [python3-validators]
   fedora: [python3-validators]
   opensuse: [python3-validators]
-  ubuntu: 
+  ubuntu:
     '*': [python3-validators]
     focal: null
 python3-vcstool:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -10257,14 +10257,9 @@ python3-validators:
   debian: [python3-validators]
   fedora: [python3-validators]
   opensuse: [python3-validators]
-  ubuntu:
+  ubuntu: 
     '*': [python3-validators]
-    bionic:
-      pip:
-        packages: [validators]
-    focal:
-      pip:
-        packages: [validators]
+    focal: null
 python3-vcstool:
   alpine: [vcstool]
   debian: [python3-vcstool]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -10256,6 +10256,7 @@ python3-uvloop:
 python3-validators:
   debian: [python3-validators]
   fedora: [python3-validators]
+  opensuse: [python3-validators]
   ubuntu:
     '*': [python3-validators]
     bionic:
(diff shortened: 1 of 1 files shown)


sloretz (dismissed):
It looks like this is available from system packages. Please create a [system package key instead of a pip key](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1)

* fedora: [python3-validators](https://packages.fedoraproject.org/pkgs/python-validators/python3-validators/)
* debian: [python3-validators](https://packages.debian.org/bookworm/python3-validators)
* ubuntu: [python3-validators](https://packages.ubuntu.com/jammy/python3-validators)

fujitatomoya (commented):
@Barry-Xu-2018 thanks for opening the PR. can you address @sloretz comments?

Barry-Xu-2018:
> @Barry-Xu-2018 thanks for opening the PR. can you address @sloretz comments?

Thanks for the reminder. It is done.

Barry-Xu-2018:
`python3-validators`  wasn't provided in `Bionic` and `Focal` version for Ubuntu.  
So for the Bionic and Focal versions of Ubuntu, I'll continue using pip to install validators.

clalancette (changes requested):
The bot thinks we may be able to satisfy this on `opensuse` as well.  Once that is added, this should be good to go.

Barry-Xu-2018:
> The bot thinks we may be able to satisfy this on `opensuse` as well. Once that is added, this should be good to go.

Okay. I added `opensuse` platform.

fujitatomoya (approved):

rosdep/python.yaml
```diff
@@ -10253,6 +10253,18 @@ python3-uvloop:
...
+  fedora: [python3-validators]
+  opensuse: [python3-validators]
+  ubuntu:
+    '*': [python3-validators]
+    bionic:
+      pip:
+        packages: [validators]
+    focal:
+      pip:
+        packages: [validators]
```
nuclearsandwich, line 10267: Bionic is un-supported so we shouldn't add handling for it. Focal is nearly EOL so I think it's not worth adding unless you're actually using it.

```suggestion
```

nuclearsandwich (dismissed):
Thanks for continuing to iterate @Barry-Xu-2018. I have a few recommended changes based on the current state.

Barry-Xu-2018:
> Bionic is no longer supported so we don't need to add it. Unless you're actually using the pip package on Focal I don't think it's worth adding there either since its nearly end-of-support itself.

Get it. Thank you for your information.   
I will remove Bionic and Focal.

clalancette (approved):
I think we need to have "something" for `focal`, otherwise the linters will fail.  So I think this is good as it is.

PR #43502, opened 2024-11-14
Add python3-streamz to rosdep/python.yaml

galou:

## Package name:

`python3-streamz`: Pipelines to manage continuous streams of data

## Package Upstream Source:

https://github.com/python-streamz/streamz/

## Purpose of using this:

We use `streamz` to apply a sliding window average on raw input data for logging and user interface.

Distro packaging links:

  - https://packages.debian.org/search?keywords=streamz&searchon=names&suite=stable&section=all
  - https://packages.ubuntu.com/oracular/python3-streamz
  - not available
  - not available
  - not available
  - not available
  - not available
  - https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/development/python-modules/streamz/default.nix#L83
  - https://software.opensuse.org/package/python3-streamz
  - https://debian.pkgs.org/12/debian-main-amd64/python3-streamz_0.6.4-1_all.deb.html

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9886,6 +9886,18 @@ python3-streamlit-pip:
   '*':
     pip:
       packages: [streamlit]
+python3-streamz:
+  debian:
+    bookworm: [python3-streamz]
+    bullseye: [python3-streamz]
+  ubuntu:
+    '*': [python3-streamz]
+    bionic:
+      pip:
+        packages: [streamz]
+    focal:
+      pip:
+        packages: [streamz]
 python3-suas-interop-clients-pip:
   debian:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9892,9 +9892,6 @@ python3-streamz:
     bullseye: [python3-streamz]
   ubuntu:
     '*': [python3-streamz]
-    bionic:
-      pip:
-        packages: [streamz]
     focal:
       pip:
         packages: [streamz]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9890,6 +9890,7 @@ python3-streamz:
   debian:
     bookworm: [python3-streamz]
     bullseye: [python3-streamz]
+  opensuse: [python3-streamz]
   ubuntu:
     '*': [python3-streamz]
     focal:


sloretz (dismissed):
It looks like this package is available as system packages. Please use those instead of `pip` entries.

```
Key 'python3-streamz' might be satisfied by debian package named 'python3-streamz': https://packages.debian.org/bookworm/python3-streamz
```

```
Key 'python3-streamz' might be satisfied by opensuse package named 'python3-streamz': https://software.opensuse.org/package/python-streamz
```

galou:
There's also the package for Tumbleweed but I don't know how to add it (`opensuse: [python-streamz]`?).

galou:
What is preferred, should I rebase and squash or simply push a new commit?

clalancette:
> What is preferred, should I rebase and squash or simply push a new commit?

Either one is fine.  We'll end up squashing it in the end.

rosdep/python.yaml
```diff
@@ -9886,6 +9886,18 @@ python3-streamlit-pip:
...
       packages: [streamlit]
+python3-streamz:
+  debian:
+    bookworm: [python3-streamz]
+    bullseye: [python3-streamz]
+  ubuntu:
+    '*': [python3-streamz]
+    bionic:
+      pip:
+        packages: [streamz]
```
clalancette, line 9897: Bionic is EOL, so we don't need to add it:

```suggestion
```

rosdep/python.yaml
```diff
@@ -9886,6 +9886,18 @@ python3-streamlit-pip:
   '*':
     pip:
       packages: [streamlit]
+python3-streamz:
+  debian:
+    bookworm: [python3-streamz]
+    bullseye: [python3-streamz]
+  ubuntu:
```
clalancette, line 9893:
```suggestion
  opensuse: [python3-streamz]
  ubuntu:
```

clalancette (changes requested):

rosdep/python.yaml
galou: `python3-streamz` is only available as experimental package if I understand correctly. Moreover, on Tumbleweed it is available as `python-streamz`.

galou:
I committed the suggested changes because I got no answer about my doubts with OpenSuse.

clalancette (approved):

quarkytale (approved):

PR #43895, opened 2024-12-13
Fix Nix packages that incorrectly used Python 2

lopsided98:
`pythonPackages` normally refers to Python 2 in nixpkgs, but I have had an override in nix-ros-overlay that remaps it to Python 3, dating back to the days when I had to support both Python 2 and Python 3 for different distros. This has led to several packages in rosdep mistakenly using `pythonPackages` rather than `python3Packages`. Fix these to allow me to remove the override for ROS 2.


clalancette (approved):

lopsided98:
Not stale, just needs review

audrow (approved):

PR #43915, opened 2024-12-14
Add coal source for Humble

nim65s:
Hi,

The hpp-fcl package has been renamed to coal for its v3.0.0.
It was requested to add a new source: https://github.com/ros2-gbp/ros2-gbp-github-org/pull/671

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -932,6 +932,12 @@ repositories:
       url: https://github.com/carologistics/clips_vendor.git
       version: main
     status: maintained
+  coal:
+    source:
+      type: git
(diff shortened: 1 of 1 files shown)


clalancette (approved):

  - [x] Top level license file: https://github.com/coal-library/coal/blob/devel/LICENSE
- [x] License is [OSI-approved](https://opensource.org/licenses): https://opensource.org/license/bsd-3-clause
- [x] Public source repo: https://github.com/coal-library/coal
- [x] Each package meets [REP-144](https://www.ros.org/reps/rep-0144.html) naming conventions: This is borderline, since it is so short.  However, it is unique enough, and doesn't cause any obvious collisions, so approved.

PR #43975, opened 2024-12-18
Revert tuw msgs to 0.2.1 1

audrow:
This reverts https://github.com/ros/rosdistro/pull/43875 and https://github.com/ros/rosdistro/pull/43802 to get back to a building version of `tuw_msgs`.

It looks like the problem is that the December 5 release is the one that added in the new package, and made `tuw_msgs` depend on `tuw_std_msgs`.

FYI, @maxbader.


clalancette (commented):
Looks good to me!

Issue #438, opened 2013-02-06
Add support for virtual packages


tfoote:
Regarding Virtual Packages, this seems like a valid new feature, but it is orthogonal to the other issues currently mentioned around metapackages. Also the need for this feature seems not that urgent, so unless some volunteer wants to write up a REP, prototype and such, it is not going to happen anytime soon. Anyone might open a ticket so that others may upvote it.

tfoote:
Fixed in ros-infrastructure/rosdep#468

PR #43981, opened 2024-12-19
Revert "pinocchio: 3.3.0-1 in 'jazzy/distribution.yaml' [bloom] (#438…

marcoag:
This reverts commit 1786a2e3afdddbc2e5e15789823c319c54052415.

The new version brings a regression due to a linking error in `jazzy`: https://build.ros2.org/view/Jbin_uN64/job/Jbin_uN64__pinocchio__ubuntu_noble_amd64__binary/34/console

I'm reverting this update so the package don't get wiped out in the upcoming `jazzy` patch sync.

FYI: @nim65s please take a look.

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -5172,7 +5172,7 @@ repositories:
       tags:
         release: release/jazzy/{package}/{version}
       url: https://github.com/ros2-gbp/pinocchio-release.git
-      version: 3.3.0-1
+      version: 2.6.21-3
     source:
       type: git
       url: https://github.com/stack-of-tasks/pinocchio.git


nim65s:
Thanks @marcoag for the heads-up !

```
06:03:07 clang++: error: unable to execute command: Killed
06:07:42 clang++: error: linker command failed due to signal (use -v to see invocation)
```
Can this be a OoM issue on the builder ?
The exact same thing did work on humble: https://build.ros2.org/view/Hbin_uJ64/job/Hbin_uJ64__pinocchio__ubuntu_jammy_amd64__binary/ ; so I'm not sure what to do here.

marcoag:
Could be OoM as eigenpy was failing for OoM issues and it's a dependency of this package. However `pinocchio` is not succeeding on retries so I don't think I can do much on my side. Memory is limited on the jobs, can you see if you can reduce the amount that the package uses during build? Maybe a start point is to look on the differences with humble that might make build more lean?

nim65s:
https://github.com/ros2-gbp/pinocchio-release/compare/release/humble/pinocchio/3.3.0-1..release/jazzy/pinocchio/3.3.0-1 :
> There isn’t anything to compare.
> 
> release/humble/pinocchio/3.3.0-1 and release/jazzy/pinocchio/3.3.0-1 are identical. 

So that might just be luck. Or maybe clang take less RAM between `15:03:38 -- The CXX compiler identification is Clang 14.0.0` and `05:44:58 -- The CXX compiler identification is Clang 18.1.3`.

I'll try to benchmark RAM use of ld / gold /  mold / lld 14 / lld 18.

Diffing logs also shows that Jazzy is on `AMD EPYC 7R32` while Humble is on `Intel(R) Xeon(R) Platinum 8124M`  maybe there is a difference in the available RAM in the runners too ?

PR #40677, opened 2024-04-19
Add doc and source entries for rmw_zenoh into rolling/distribution.yaml

Yadunund:
# Please Add This Package to be indexed in the rosdistro.

`rmw_zenoh`

# The source is here:

https://github.com/ros2/rmw_zenoh

# Checks
 - [x] All packages have a declared license in the package.xml
 - [x] This repository has a LICENSE file
 - [x] This package is expected to build on the submitted rosdistro

PR #44093, opened 2025-01-02
Add doc and source entry for rmw_zenoh (humble)

Yadunund:
Similar to https://github.com/ros/rosdistro/pull/40677 but for Humble.


clalancette (changes requested):
This one is targeting the wrong file, it should be `humble/distribution.yaml`.

clalancette (approved):

PR #44119, opened 2025-01-08
Add platformio

Tacha-S:

## Package name:

platformio

## Package Upstream Source:

https://github.com/platformio/platformio-core

## Purpose of using this:

I use this to write the latest program to the microcontroller board during the build process.

Distro packaging links:

  - https://packages.debian.org/search?keywords=platformio&searchon=names&suite=all&section=all
  - https://packages.ubuntu.com/search?keywords=platformio&searchon=names&suite=all&section=all
  - https://packages.fedoraproject.org/search?query=platformio
  - not available
  - not available
  - not available
  - not available
  - OPTIONAL
  - not available
  - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -368,6 +368,7 @@ platformio:
   debian:
     pip:
       packages: [platformio]
+  fedora: [python-platformio]
 pydocstyle:
   alpine: [py3-pydocstyle]
   arch: [python-pydocstyle]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -365,7 +365,7 @@ platformio:
   debian:
     pip:
       packages: [platformio]
-  fedora: [python-platformio]
+  fedora: [platformio]
   ubuntu:
     '*': [platformio]
     focal: null
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -361,6 +361,13 @@ pika:
     pip:
       packages: [pika]
   ubuntu: [python-pika]
+platformio:
+  ubuntu:
+    '*': [platformio]
+    focal: null
+  debian:
+    pip:
+      packages: [platformio]
 pydocstyle:
   alpine: [py3-pydocstyle]
   arch: [python-pydocstyle]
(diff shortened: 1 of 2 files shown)


christophebedard:
Is the package available on Debian? We require at least Ubuntu and Debian: https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md#types-of-pull-requests

Tacha-S:
As indicated by the [packages.debian.org/search?keywords=platformio&searchon=names&suite=all&section=all](https://packages.debian.org/search?keywords=platformio&searchon=names&suite=all&section=all) in the Debian section, this package does not exist in Debian.
However, it can be installed on Debian using `pip`.
In such cases, should it be written in `python.yaml` instead?

christophebedard:
Yes, you can add it as `python3-platformio-pip` in `rosdep/python.yaml`, but note that ROS packages that depend on `pip` keys cannot be released into a ROS distribution. It will only be useful for packages that get built from source.

Tacha-S:
I moved it to python.yml.

rosdep/python.yaml
```diff
@@ -361,6 +361,14 @@ pika:
     pip:
       packages: [pika]
   ubuntu: [python-pika]
+platformio:
+  debian:
+    pip:
+      packages: [platformio]
+  fedora: [python-platformio]
```
clalancette, line 368:
```suggestion
  fedora: [platformio]
```

(on Fedora, it looks like the `platformio` package provides the command-line utils, equivalent to what `platformio` does on Ubuntu)

clalancette (changes requested):

rosdep/python.yaml
Tacha-S: It seems so.
I have made the corrections.

clalancette (approved):

Tacha-S:
@christophebedard @clalancette Do I still need to take any action?

christophebedard:
We usually require 2 approvals for new rosdep keys, and sometimes it results in PRs getting lost a bit. Sorry for the delay.

christophebedard (approved):

PR #44122, opened 2025-01-08
Add torchrl

baconbrot:

## Package name:

torchrl

## Package Upstream Source:

https://github.com/pytorch/rl

## Purpose of using this:

Reinforcement Learning with pytorch.

https://pypi.org/project/torchrl/

# Checks


christophebedard (approved):
There doesn't seem to be any system package for this, so a pip key is fine with me.

clalancette (approved):

PR #44124, opened 2025-01-09
README: expand description of packaging metadata

scpeters:
This is not a typical pull request; it's just updating the README with an expanded description of the two types of data contained in this repository in a way that I think is more readable. Thanks in advance for helping to correct any inaccuracies in my phrasing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,12 @@
-This repo maintains the lists of repositories defining ROS distributions.
+The rosdistro repository stores two independent datasets used in ROS:
 
-It is the implementation of [REP 143](http://ros.org/reps/rep-0143.html).
+1. The lists of repositories that curate ROS packages for each ROS distributions,
+   implementing the data structure defined in [REP 143](http://ros.org/reps/rep-0143.html).
+   Any ROS package release will generate pull requests to the distribution files
+   in this repository.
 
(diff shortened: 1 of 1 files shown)


jlblancoc:
:+1: 
I would prefer another wording instead of `datasets`, which for most of us in robotics has a different meaning than implied here... perhaps: 

```
-The rosdistro repository stores two independent datasets used in ROS:
+This repository maintains:
```

scpeters:
> 👍 I would prefer another wording instead of `datasets`, which for most of us in robotics has a different meaning than implied here

how about "This repository maintains two independent sets of packaging metadata"?

christophebedard:
I also think that would be better.

scpeters:
> > 👍 I would prefer another wording instead of `datasets`, which for most of us in robotics has a different meaning than implied here
> 
> how about "This repository maintains two independent sets of packaging metadata"?

applied in https://github.com/ros/rosdistro/pull/44124/commits/5170aead2ff689e96ee79b15702911a0e1ba2c45

christophebedard (approved):

nuclearsandwich-ai (approved):
Thanks!

nuclearsandwich (approved):

PR #44182, opened 2025-01-14
Add `rmw_desert` to rolling

dcostan:

Rolling

# The source is here:

[https://github.com/signetlabdei/rmw_desert](https://github.com/signetlabdei/rmw_desert)

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -6008,6 +6008,16 @@ repositories:
       url: https://github.com/ros2/rmw_dds_common.git
       version: rolling
     status: maintained
+  rmw_desert:
+    doc:
+      type: git
+      url: https://github.com/signetlabdei/rmw_desert.wiki.git
+      version: 1.0
+    source:
+      type: git
+      url: https://github.com/signetlabdei/rmw_desert.git
+      version: 1.0
+    status: maintained
   rmw_fastrtps:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


mjcarroll:
@dcostan I just took a look, and it seems that while you have the GPL in the `LICENSE` file, the `package.xml` specifies LGPL.  Mind making sure these are consistent?

dcostan:
@mjcarroll Fixed the license

rolling/distribution.yaml
```diff
@@ -6037,6 +6037,16 @@ repositories:
       url: https://github.com/ros2/rmw_dds_common.git
       version: rolling
     status: maintained
+  rmw_desert:
+    doc:
+      type: git
+      url: https://github.com/signetlabdei/rmw_desert.wiki.git
```
mjcarroll, line 6043: Maybe @cottsay can chime in here, I recall that the doc stanza is trying to run the doc job on the buildfarm.  It generally expects that you are pointing to the source code with a `dox` file in the root.  I'm not sure this will work.

clalancette: That's correct.  This stanza should point to the source repository.

clalancette (changes requested):

mjcarroll (approved):

clalancette (approved):

PR #40564, opened 2024-04-12
Add python3-pymongo-pip and python3-motor-pip keys for debian and ubuntu

wilml:
Please add the following dependency to the rosdep database.

## Package name:

* python3-pymongo-pip
* python3-motor-pip

## Package Upstream Source:
* https://github.com/mongodb/mongo-python-driver/
* https://github.com/mongodb/motor

## Purpose of using this:

The pymongo package distributed by apt is a major release behind (v3.11 compared to v4.6) missing a set of critical new features and functionality.

Additionally Motor is an asynchronous driver for MongoDB that allows for data to be accessed asynchronously. 

## Links to Distribution Packages

* pip: https://pypi.org/project/pymongo/
* pip: https://pypi.org/project/motor/

PR #44281, opened 2025-01-24
Add black-pip key for debian and ubuntu

jackiescanlon:

## Package name:

black-pip

## Package Upstream Source:

https://github.com/psf/black

## Purpose of using this:

In apt the current version supported is 21.12b0 and the latest stable version available from black is 24.10.0 which includes bug fixes and feature updates that are desired. Also, 21.12b0 is not supported by the VS Code black extension.

* pip: https://pypi.org/project/black/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -113,6 +113,13 @@ black:
   ubuntu:
     '*': [black]
     bionic: null
+black-pip:
+  debian:
+    pip:
+      packages: [black]
+  ubuntu:
+    pip:
+      packages: [black]
 canopen-pip:
   debian:
     pip:


tfoote (changes requested):
Black is packaged in the native packaging: https://packages.ubuntu.com/oracular/black

This rule will cause conflicts with users of the native packaging and so we can't add a pip rule for this. There are definitely new features in lots of packages but we have to keep coordinated with the broader distribution.

jackiescanlon:
Thanks @tfoote, I meant for this to only be an alternative to the native, not a replacement. Also, I see a precedent set elsewhere to allow the pip package to be added as an option if the native package is a major release behind: https://github.com/ros/rosdistro/pull/40564/files

Is this precedent no longer being followed?

tfoote:
We have a very clear policy in the review guidelines: https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md
`Native packages are strongly preferred. If a native package exists for the key in question, then that should always be used over e.g. a pip key.`

I will be submitting a revert for the linked PR at  https://github.com/ros/rosdistro/pull/44287

jackiescanlon:
Okay, thanks for your consideration.

PR #44305, opened 2025-01-29
Add isaacsim pip with cache pkgs

baconbrot:

## Package name:
Edit: In the newly released Isaac Sim 4.5.0, the package structure has changed. Packages are now installed via pip extras, and I propose adding the following extras to rosdep to replace the old ones:

python3-isaacsim-all-pip
python3-isaacsim-rl-pip
python3-isaacsim-ros-pip
python3-isaacsim-ros2-pip
python3-isaacsim-extscache-pip

~~python3-isaacsim-pip~~
~~python3-isaacsim-extscache-physics-pip~~
~~python3-isaacsim-extscache-kit-pip~~
~~python3-isaacsim-extscache-kit-sdk-pip~~

## Package Upstream Source:

NA

## Purpose of using this:

Simulating ROS integrated robots using NVIDIA Omniverse IsaacSim.
~~The main package `isaacsim` can be supplemented by any subset of the other three cache packages to accelerate first time startup.~~

There are no special distribution packages, only pip.
pip:
 - https://pypi.org/project/isaacsim/
 - ~~https://pypi.org/project/isaacsim-extscache-physics/~~
 - ~~https://pypi.org/project/isaacsim-extscache-kit/~~
 - ~~https://pypi.org/project/isaacsim-extscache-kit-sdk/~~

Edit: Available extras can be inspected in the `isaacsim-4.5.0.0.tar.gz` file available on PyPI.

## Additional notes
Installation with rosdep might require the user to set the --as-root pip:false flag, e.g.
`rosdep install -i -y --from-paths src --as-root pip:false`

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6796,6 +6796,22 @@ python3-interpreter:
   rhel: [python3]
   slackware: [python3]
   ubuntu: [python3-minimal]
+python3-isaacsim-extscache-kit-pip:
+  '*':
+    pip:
+      packages: [isaacsim-extscache-kit]
+python3-isaacsim-extscache-kit-sdk-pip:
+  '*':
+    pip:
+      packages: [isaacsim-extscache-kit-sdk]
+python3-isaacsim-extscache-physics-pip:
+  '*':
+    pip:
+      packages: [isaacsim-extscache-physics]
+python3-isaacsim-pip:
+  '*':
+    pip:
+      packages: [isaacsim]
 python3-j1939-pip: *migrate_eol_2025_04_30_python3_j1939_pip
 python3-jinja2:
   alpine: [py3-jinja2]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6796,22 +6796,26 @@ python3-interpreter:
   rhel: [python3]
   slackware: [python3]
   ubuntu: [python3-minimal]
-python3-isaacsim-extscache-kit-pip:
+python3-isaacsim-all-pip:
   '*':
     pip:
-      packages: [isaacsim-extscache-kit]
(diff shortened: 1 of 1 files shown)


baconbrot:
This package is currently only available for python 3.10. Should the pip install be restricted to e.g. Ubuntu jammy (which has default python version 3.10.X)?

baconbrot:
IsaacSim 4.5.0 was released during this PR.

The new version replaces `isaacsim` with `isaacsim[all]`.  
Since pip extras aren't supported yet (see `ros-infrastructure/rosdep#900`), I'll manually add each package from `[all]` to `python3-isaacsim-pip`, similar to how `python-pytorch-pip` installs both `pytorch` and `torchvision`.  

Additionally, I'll add `python3-isaacsim-extscache-pip` for the new `[extscache]` extra.  
Next to `python3-isaacsim-extscache-pip`, I'll also add packages for the following newly introduced extras:  
- `rl`  
- `ros`  
- `ros2`

baconbrot:
Still required

cottsay (approved):
I don't have any particular feedback here. The rules look really big but I think it makes sense to break out the full list of packages included in the extra like that.

> This package is currently only available for python 3.10. Should the pip install be restricted to e.g. Ubuntu jammy (which has default python version 3.10.X)?

I don't think so. It would be different if we were talking about the system package manager, but you could absolutely use pip on one of those platforms with a different version of the interpreter. In any case, pip will give a graceful error if it finds that you're trying to install using an unsupported interpreter since `isaacsim` properly declares the constraint.

Standard disclaimer that rosdep keys backed by pip rules can't be used by packages which are built on the official buildfarm.

marcoag (approved):

PR #44520, opened 2025-02-19
Revert "ros_battery_monitoring: 1.0.1-1 in 'jazzy/distribution.yaml' [bloom]"

ottojo:
Reverts ros/rosdistro#44393

It seems this update (which updated realtime_tools header names) is not actually compatible with jazzy, which we only noticed now. Does rosdistro accept downgrades or should i create a new release on the (now created) jazzy branch?

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -7331,7 +7331,7 @@ repositories:
       tags:
         release: release/jazzy/{package}/{version}
       url: https://github.com/ros2-gbp/ros_battery_monitoring-release.git
-      version: 1.0.1-1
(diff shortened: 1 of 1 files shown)


marcoag (approved):

marcoag:
> Reverts #44393
> 
> It seems this update (which updated realtime_tools header names) is not actually compatible with jazzy, which we only noticed now. Does rosdistro accept downgrades or should i create a new release on the (now created) jazzy branch?

The downgrade should work, yes! There's no need for further work.

marcoag (approved):

PR #44552, opened 2025-02-20
Add pfs

Timple:

jazzy pfs

# The source is here:

See modified files

# Checks

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -5395,6 +5395,16 @@ repositories:
       url: https://github.com/ros2/performance_test_fixture.git
       version: jazzy
     status: maintained
+  pfs:
+    doc:
+      type: git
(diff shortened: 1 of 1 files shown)


marcoag (approved):

sloretz:
This one is a little unusual in that it doesn't use ROS, but has a package.xml. 

It looks like this one got talked about on the Feb 25th ROS PMC meeting. The notes suggest asking to change the name of the package in the `package.xml` to something other than `pfs` because there's a possible collision with `pfstools`/`libpfs2` in Ubuntu and Debian https://launchpad.net/ubuntu/+source/pfstools . 

I don't have a good suggestion on how to move forward. I might suggest looking for a Debian maintainer that is willing to package it for Debian, and use `ament_vendor` to release it as `pfs_vendor` in the meantime.

PR #44571, opened 2025-02-21
boost_sml_vendor: 1.1.11-1 in 'rolling/distribution.yaml' [bloom]

Timple:
Increasing version of package(s) in repository `boost_sml_vendor` to `1.1.11-1`:

- upstream repository: https://github.com/nobleo/boost_sml_vendor.git
- release repository: https://github.com/ros2-gbp/boost_sml_vendor-release.git
- bloom version: `0.12.0`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -835,6 +835,21 @@ repositories:
       type: git
       url: https://github.com/tesseract-robotics/boost_plugin_loader.git
       version: main
+  boost_sml_vendor:
+    doc:
+      type: git
+      url: https://github.com/nobleo/boost_sml_vendor.git
+      version: main
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/boost_sml_vendor-release.git
+      version: 1.1.11-1
+    source:
+      type: git
+      url: https://github.com/nobleo/boost_sml_vendor.git
+      version: main
+    status: maintained
   camera_ros:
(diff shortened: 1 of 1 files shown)


marcoag:
> Is this bot comment really required for PRs generated with the bloom tool?

I would say it's useful yes, there are considerations that are not guaranteed just because bloom was used: i.e.:
- Your package name should comply to [REP-144 ROS Package Naming](https://www.ros.org/reps/rep-0144.html)

marcoag (approved):

Timple:
> * Your package name should comply to [REP-144 ROS Package Naming](https://www.ros.org/reps/rep-0144.html)

But that part is tackled earlier on right? Because before one can create a bloom release, a release repository needs to be created. And for that a rosdistro entry is required. 
And that last step is not automated, so the bot can stay there and the package name compliance is checked.

marcoag:
> > * Your package name should comply to [REP-144 ROS Package Naming](https://www.ros.org/reps/rep-0144.html)
> 
> But that part is tackled earlier on right? Because before one can create a bloom release, a release repository needs to be created. And for that a rosdistro entry is required. And that last step is not automated, so the bot can stay there and the package name compliance is checked.

I'm not sure if I follow here. I don't think a rosdistro entry is needed for a release repository. Also release repositories can be created anywhere, even though we recomend `ros2-gbp`. In any case this line was just an example. I also think the comment helps provide context to the requirements when releasing a package.

PR #44608, opened 2025-02-25
Add python3-utm-pip

Timple:

## Package name:

python3-utm-pip

## Package Upstream Source:

[TODO link to source repository](https://pypi.org/project/utm/)

## Purpose of using this:

UTM conversions with python

Distro packaging links:

  - REQUIRED
  - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -10548,11 +10548,11 @@ python3-usb:
   gentoo: [dev-python/pyusb]
   nixos: [python3Packages.pyusb]
   openembedded: [python3-pyusb@meta-python]
+  ubuntu: [python3-usb]
 python3-utm-pip:
   '*':
     pip:
       packages: [utm]
-  ubuntu: [python3-usb]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
This is packaged on Noble and newer: https://packages.ubuntu.com/noble/python3-utm

As per our policy we can't have a pip rule that overrides it. We can use pip to backfill older platforms, but since it's packaged now if have a rule it should be using the native packages where available.

Timple:
Ow wauw, I did miss that! Thought I tried an apt install :thinking: .

My bad! I'll come back with a new proposal!

Timple:
https://github.com/ros/rosdistro/pull/44654

PR #44609, opened 2025-02-26
Rename release repository in active distributions.

nuclearsandwich:
The old repository url will still work via GitHub redirect unless a new fastrtps-release repository is created.


nuclearsandwich:
Renamed via ros2-gbp here: https://github.com/ros2-gbp/ros2-gbp-github-org/pull/704

Yadunund (approved):

PR #44623, opened 2025-02-26
Fix geographiclib for generic Ubuntu version

chris-hagen:
## Package name:

geographiclib

## Package Upstream Source:

rosdep

## Purpose of using this:

geographiclib is broken for generic Ubuntu version. The default value `libgeographiclib-dev` is used. The apt package list defines the package as `libgeographic-dev`.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1418,7 +1418,7 @@ geographiclib:
   openembedded: [geographiclib@meta-ros-common]
   rhel: [GeographicLib-devel]
   ubuntu:
-    '*': [libgeographiclib-dev]
+    '*': [libgeographic-dev]
     focal: [libgeographic-dev]
     jammy: [libgeographic-dev]
 geographiclib-tools:


mjcarroll (approved):

rosdep/base.yaml
```diff
@@ -1418,7 +1418,7 @@ geographiclib:
   openembedded: [geographiclib@meta-ros-common]
   rhel: [GeographicLib-devel]
   ubuntu:
-    '*': [libgeographiclib-dev]
+    '*': [libgeographic-dev]
```
christophebedard, line 1421: Looks like it's `libgeographiclib-dev` for Ubuntu Noble: https://packages.ubuntu.com/noble/libgeographiclib-dev. So I'm not sure this change is needed.

christophebedard (changes requested):
As-is, this change is not valid. See the CI failure and my comment.

rosdep/base.yaml
sloretz: I agree that the rosdep key is correct as is.

focal, jammy [libgeographic-dev](https://packages.ubuntu.com/jammy/libgeographic-dev)
Noble and beyond [https://packages.ubuntu.com/noble/libgeographiclib-dev](https://packages.ubuntu.com/noble/libgeographiclib-dev)

sloretz (commented):
> geographiclib is broken for generic Ubuntu version. The default value libgeographiclib-dev is used. The apt package list defines the package as libgeographic-dev.

Please open an issue with steps to reproduce. It looks like the rosdep key is correct as is.

sloretz:
I'll close this for now since the rosdep key appears correct.

PR #44640, opened 2025-02-28
Add rustup dependency to rosdep base.yaml

Mallets:

## Package name:

rustup

## Package Upstream Source:

https://github.com/rust-lang/rustup

## Purpose of using this:

[rmw_zenoh](https://github.com/ros2/rmw_zenoh) internal core is written in Rust and it uses the `cargo` as dependency for the build. While this is currently working, installing directly `cargo` is pinning the Rust toolchain to a quite old version in many distributions. For example, on Ubuntu `24.04` the shipped Rust toolchain is `1.75`.

Since Jazzy the Zenoh project has globally pinned the Rust version to [1.75](https://github.com/eclipse-zenoh/zenoh/blob/2311334ebb82378e341d350a08d0667169c7f211/rust-toolchain.toml).
However, this is now a limitation for the project since newer Rust toolchains have been release and downstream dependencies already started requiring more recent Rust versions (e.g. https://github.com/eclipse-zenoh/zenoh/pull/1800).

[rustup](https://rustup.rs/) is a tool that installs [Rust](https://www.rust-lang.org/) on the system from the official release channels, enabling to easily switch between toolchain versions and keep them updated.
By depending on `rustup` instead of `cargo` [rmw_zenoh](https://github.com/ros2/rmw_zenoh) (and any other Rust project) will be able to use the required Rust toolchain and not forced to use the toolchain shipped with any distro.

Distro packaging links:

  - https://packages.debian.org/sid/rustup
  - https://packages.ubuntu.com/search?keywords=rustup&searchon=names&suite=noble&section=all
  - https://packages.fedoraproject.org/pkgs/rustup/rustup/
  - https://archlinux.org/packages/?sort=&q=rustup&maintainer=&flagged=
  - https://packages.gentoo.org/packages/dev-util/rustup
  - https://formulae.brew.sh/formula/rustup
  - https://pkgs.alpinelinux.org/package/edge/community/x86_64/rustup
  - https://nixos.wiki/wiki/Rust
  - https://software.opensuse.org/package/rustup
  - Not available

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8487,6 +8487,15 @@ rustc:
   opensuse: [rust]
   rhel: [rust]
   ubuntu: [rustc]
+rustup:
+  arch: [rustup]
+  debian: [rustup]
+  fedora: [rustup]
+  gentoo: [dev-util/rustup]
+  nixos: [rustup]
+  opensuse: [rustup]
+  rhel: [rustup]
+  ubuntu: [rustup]
 sbcl:
   alpine: [sbcl]
   arch: [sbcl]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8495,7 +8495,6 @@ rustup:
   gentoo: [dev-util/rustup]
   nixos: [rustup]
   opensuse: [rustup]
-  rhel: [rustup]
   ubuntu: [rustup]
 sbcl:
   alpine: [sbcl]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8488,6 +8488,7 @@ rustc:
   rhel: [rust]
   ubuntu: [rustc]
 rustup:
+  alpine: [rustup]
   arch: [rustup]
   debian: [rustup]
   fedora: [rustup]


tfoote (changes requested):
The rule itself here is ok, but we may need to have some extra rules about compatibility with debian packaging, like we do with pip such that people know it's incompatible with being able to make a release. 

The proposed changes  moving the zenoh compiler forward connected to this are a major issue that needs to be discussed.

tfoote:
Now seeing the build test results it looks like the proposed rules don't exist on several platforms either.

rosdep/base.yaml
```diff
@@ -8487,6 +8487,15 @@ rustc:
...
   ubuntu: [rustc]
+rustup:
+  alpine: [rustup]
+  arch: [rustup]
+  debian: [rustup]
+  fedora: [rustup]
+  gentoo: [dev-util/rustup]
+  nixos: [rustup]
+  opensuse: [rustup]
+  ubuntu: [rustup]
```
christophebedard, line 8498: You can make it version-specific, e.g.:

```suggestion
  ubuntu: [rustup]
    '*': [rustup]
    focal: null
    jammy: null
```

christophebedard:
Since there has been no activity for ~1 month, I'll close this. If you still want to proceed, please re-open and address the CI failures.

PR #44646, opened 2025-03-01
foxglove_msgs: update repo url

amacneil:

**foxglove_msgs**

- updated the repo URL
- set `status: developed` consistently across all distros
- added `packages` key to explicitly list the packages (we plan to add more in the future)

I'm not sure if the URL should be corrected for EOL distros, please let me know if you would prefer I revert those.

# The source is here:

https://github.com/foxglove/foxglove-sdk/tree/main/ros/foxglove_msgs

# Checks

### DIFF
--- a/foxy/distribution.yaml
+++ b/foxy/distribution.yaml
@@ -1563,20 +1563,18 @@ repositories:
   foxglove_msgs:
     doc:
       type: git
-      url: https://github.com/foxglove/foxglove-sdk.git
+      url: https://github.com/foxglove/schemas.git
       version: main
     release:
-      packages:
-      - foxglove_msgs
       tags:
         release: release/foxy/{package}/{version}
       url: https://github.com/ros2-gbp/ros_foxglove_msgs-release.git
       version: 2.1.1-1
     source:
       type: git
-      url: https://github.com/foxglove/foxglove-sdk.git
+      url: https://github.com/foxglove/schemas.git
       version: main
-    status: developed
+    status: maintained
   gazebo_ros2_control:
     doc:
       type: git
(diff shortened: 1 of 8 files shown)


sloretz:
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2025-03-06/)

christophebedard:
The sync hold is now over, but I don't think we accept changes to `distribution.yaml` files of EOL distros (Foxy, Galactic, Iron, Melodic).

christophebedard (dismissed):
To be clearer: we indeed do not take changes to `distribution.yaml` files of EOL distro.

mjcarroll:
@amacneil friendly ping. We don't take changes to EOL distros, instead preferring those to remain frozen snapshots in time.  If you can get those removed, we can get this in.

sloretz:
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2025-04-24/)

emersonknapp (approved):

emersonknapp:
Okay, to move this forward, @amacneil I hope you don't mind I went ahead and pushed two commits fixing this up into mergeable state:
1. revert the changes to EOL distros
2. Remove `packages:` per formatter preferences. Though https://www.ros.org/reps/rep-0141.html#distribution-file doesn't actually say you can't have a single-package listing, it does say "If no package is specified, one package with the name of the repository is assumed.", and in this case that name is `foxglove_msgs`. So, the formatter didn't like having a single listing of `foxglove_msgs` since the distribution file assumes that when not specified.

Unfortunately we've just entered another Noetic freeze
[Holding for Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2025-05-08/)

But as soon as that is lifted we can merge this

amacneil:
@emersonknapp @mjcarroll bit of a slow response here, but thanks for fixing up and landing the PR!

PR #44705, opened 2025-03-09
Added python3-mss-pip

peci1:

## Package name:

python3-mss-pip

## Package Upstream Source:

https://github.com/BoboTiG/python-mss

## Purpose of using this:

Fast multiplatform Python screenshotter.

Distro packaging links:

Only found on PyPI: https://pypi.org/project/mss/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7394,6 +7394,10 @@ python3-msm-pip:
   ubuntu:
     pip:
       packages: [msm]
+python3-mss-pip:
+  '*':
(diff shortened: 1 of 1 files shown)


peci1:
I'd still like this added. There's no problem having this package, or is there?

christophebedard:

christophebedard (approved):
Approved.

Note that we require 2 approvals for PRs adding rosdep keys.

quarkytale (approved):

PR #44741, opened 2025-03-13
Add silero-vad pip package rule to rosdep

vsevolod-misiul:

## Package name:

silero-vad

## Package Upstream Source:

https://github.com/snakers4/silero-vad

## Purpose of using this:

The `silero-vad` package is used for voice activity detection (VAD) in speech processing. It is a lightweight, high-performance VAD model for real-time applications.

https://pypi.org/project/silero-vad/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9731,6 +9731,10 @@ python3-sila2lib-pip:
   ubuntu:
     pip:
       packages: [sila2lib]
+python3-silero-vad-pip:
+  '*':
+    pip:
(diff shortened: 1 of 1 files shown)


christophebedard:

christophebedard:
Please rebase your branch to fix unrelated issues.

christophebedard (approved):
Thanks, approved.

Note that we require 2 approvals for new rosdep keys.

quarkytale (approved):

PR #44829, opened 2025-03-19
Add galois to python.yaml

Silviis:

## Package name:

python3-galois-pip

## Package Upstream Source:
https://github.com/mhostetter/galois

## Purpose of using this:

A performant NumPy extension for Galois fields and their applications

Distro packaging links:

Package not found in most distros package manager. It can be found in PyPI:  [https://pypi.org/project/galois/](https://pypi.org/project/galois/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6336,16 +6336,7 @@ python3-future:
   rhel: ['python%{python3_pkgversion}-future']
   ubuntu: [python3-future]
 python3-galois-pip:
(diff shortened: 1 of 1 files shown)


christophebedard:

I indeed couldn't find this for Ubuntu or Debian (note that this is not the same package: https://packages.ubuntu.com/noble/galois).

christophebedard (approved):

quarkytale (approved):

PR #44839, opened 2025-03-20
add python3-freezegun rule

mqcmd196:

## Package name:

python3-freezegun

## Package Upstream Source:

https://github.com/spulec/freezegun

## Purpose of using this:

This dependency is being used to release our package. 

Distro packaging links:

https://github.com/jsk-ros-pkg/jsk_common/tree/master/jsk_data

  - https://packages.debian.org/source/bookworm/freezegun
  - https://launchpad.net/ubuntu/jammy/+package/python3-freezegun
  - https://packages.fedoraproject.org/pkgs/python-freezegun/python3-freezegun/
  - https://packages.gentoo.org/packages/dev-python/freezegun

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6315,6 +6315,13 @@ python3-fpdf2-pip:
   ubuntu:
     pip:
       packages: [fpdf2]
+python3-freezegun:
+  debian: [python3-freezegun]
+  fedora: [python3-freezegun]
+  gentoo: [dev-python/freezegun]
+  opensuse: [python3-freezegun]
+  rhel: [python3-freezegun]
+  ubuntu: [python3-freezegun]
 python3-ftdi1:
   arch: [libftdi]
   debian: [python3-ftdi1]


rosdep/python.yaml
```diff
@@ -6315,6 +6315,13 @@ python3-fpdf2-pip:
   ubuntu:
     pip:
       packages: [fpdf2]
+python3-freezegun:
+  debian: [python3-freezegun]
+  fedora: [python3-freezegun]
+  gentoo: [dev-python/freezegun]
+  opensuse: [python3-freezegun]
+  rhel: [python3-freezegun]
```
christophebedard, line 6323: It is only available on RHEL 9? https://pkgs.org/search/?q=python3-freezegun

If so, since we also currently support RHEL 8, this should be:

```suggestion
  rhel:
    '*': [python3-freezegun]
    '8': null
```

christophebedard: My bad, the search only shows RHEL 9 results by default :facepalm:

christophebedard (approved):
Approved.

We require two approvals for these rosdep key PRs.

quarkytale (approved):

PR #44883, opened 2025-03-26
Adding libaditof Package to Ros Index

raebelchristo-adi:

## Package name:

libaditof

## Package Upstream Source:

[https://github.com/analogdevicesinc/libaditof/tree/main](https://github.com/analogdevicesinc/libaditof/tree/main)

## Purpose of using this:

This dependency serves as an SDK that provides API calls for the ADTF3175 time of flight sensors developed at ADI.

# Checks

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5297,6 +5297,12 @@ repositories:
       url: https://github.com/lgsvl/lgsvl_msgs.git
       version: noetic-devel
     status: developed
+  libaditof:
+    source:
+      type: git
+      url: https://github.com/analogdevicesinc/libaditof.git
+      version: main
+    status: maintained
   libcreate:
     doc:
       type: git
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -5298,6 +5298,10 @@ repositories:
       version: noetic-devel
     status: developed
   libaditof:
+    doc:
+      type: git
+      url: https://github.com/analogdevicesinc/libaditof.git
+      version: main
     source:
       type: git
       url: https://github.com/analogdevicesinc/libaditof.git


christophebedard:
New package checklist:

  - [x] Top level license file: https://github.com/analogdevicesinc/libaditof/blob/eb7d2b9000a8392683a3a810a9edd8d46857ca39/LICENSE
- [ ] License is [OSI-approved](https://opensource.org/licenses): MIT (dependencies?)
- [ ] License correctly listed in package.xmls: https://github.com/analogdevicesinc/libaditof/blob/eb7d2b9000a8392683a3a810a9edd8d46857ca39/package.xml#L10 (what about dependencies?)
- [x] Public source repo: https://github.com/analogdevicesinc/libaditof

----

I noticed that you have multiple submodules in your repository. This complicates licensing and packaging.

Generally, licenses of anything distributed with your package should be listed in the `package.xml`. Looking at the licenses of the repos included using submodules, they cover a bit more than MIT. Also, when relying on external dependencies, you should try to use system packages (e.g., Ubuntu system packages). If they are not available through system packages, you can use vendor packages, which you may have to create yourself if they are not available. Looks like all submodules are available as Ubuntu Focal system packages:

- libzmq: https://packages.ubuntu.com/focal/libzmq3-dev (rosdep key: `libzmq3-dev`)
- cppzmq: I believe this is included in the `libzmq3-dev` Ubuntu system package in Ubuntu <24.04
- glog: https://packages.ubuntu.com/focal/libgoogle-glog-dev (rosdep key: `libgoogle-glog-dev`)
- libwebsockets: https://packages.ubuntu.com/focal/libwebsockets-dev (rosdep key: `libwebsockets-dev`)
- protobuf: https://packages.ubuntu.com/focal/libprotobuf-dev (rosdep key: `protobuf-dev`)

So you could list these rosdep keys as dependencies (`<depend>`) in your `package.xml` and remove the submodules. You'd need to update your `CMakeLists.txt` to find the system package. Then users building your package from source would have to install the system packages (e.g., using rosdep).

I found this relevant discussion on ROS Discourse (seems like it was also about an Analog Devices package!): https://discourse.ros.org/t/buildfarm-make-install-error/39231.

Also, note that ROS Noetic will be EOL in ~2 months.

christophebedard (changes requested):
Just so that this doesn't accidentally get merged.

raebelchristo-adi:
Alright, so here's the issue. This package wasn't primarily designed for ROS. It's a generic package which was given a package.xml in order to be built as a ros package so that it worked seamlessly with our other ros packages. The development team had stated that they needed a very specific version of the given submodules, which can't be attained through rosdep.

Since we need very specific versions, we wish to keep them as submodules. However, the problem right now is licensing right? If we addressed the licenses of each submodule in our package.xml, will our package become valid? I am wondering if adding additional license tags using [this (page 10)](https://roscon.ros.org/2023/talks/Licensing_of_ROS_Packages_Changes_to_package.xml_and_new_tool_support.pdf) as reference, would help solving this issue. Any advice regarding licensing is welcome since I am not quite well versed in it.

christophebedard:
> I am wondering if adding additional license tags using [this (page 10)](https://roscon.ros.org/2023/talks/Licensing_of_ROS_Packages_Changes_to_package.xml_and_new_tool_support.pdf) as reference, would help solving this issue. Any advice regarding licensing is welcome since I am not quite well versed in it.

That REP has not been accepted/merged yet: https://github.com/ros-infrastructure/rep/pull/347. The current best practice is to list all licenses and just put an XML comment next to each one saying what part the license is for, e.g., https://github.com/ros2/rosbag2/blob/9351e6a9a46b54b37b2edd31f73e90d923e754ee/liblz4_vendor/package.xml#L10-L12. You can use the `file` attribute with a relative path to the corresponding license file, though: https://ros.org/reps/rep-0149.html#id17.

I didn't see them before, but you should also list the licenses of the third-party dependencies in the `dependencies/` directory: https://github.com/analogdevicesinc/libaditof/tree/main/dependencies

> Alright, so here's the issue. This package wasn't primarily designed for ROS. It's a generic package which was given a package.xml in order to be built as a ros package so that it worked seamlessly with our other ros packages. The development team had stated that they needed a very specific version of the given submodules, which can't be attained through rosdep.
> 
> Since we need very specific versions, we wish to keep them as submodules. However, the problem right now is licensing right? If we addressed the licenses of each submodule in our package.xml, will our package become valid?

I understand your constraints. Unfortunately, it's also about potential clashes if other packages include a different version of the same dependency, as mentioned here: https://discourse.ros.org/t/buildfarm-make-install-error/39231/2. However, your package is using `add_subdirectory()` and not `ExternalProject_Add()`, so I'm not sure if it's the same situation.

Let me get other people's input on this. I've added this to the agenda for Tuesday's ROS PMC meeting; I'll get back to you.

raebelchristo-adi:
I guess that works for us. We are fine with having just `source` entry for now. We also just realized that we also want to add a `doc` entry as well. I hope that's okay with you if I update the PR right? As it is only going to be indexed for source and doc, do we still need to update the `package.xml` to include the licenses of submodules and dependencies folder?

We also intend to update this repository in some time to use system dependencies rather than version specific submodules, however this will cross the EOL for Noetic. We will release that version on ROS2.

christophebedard:
> We also just realized that we also want to add a doc entry as well. I hope that's okay with you if I update the PR right?

Of course, adding a `doc: entry is fine.

> As it is only going to be indexed for source and doc, do we still need to update the package.xml to include the licenses of submodules and dependencies folder?

It would be good to do it (or at least mention that there are third-party dependencies that have their own licenses too), but it's not required to add `source:` and `doc:` entries here.

> We also intend to update this repository in some time to use system dependencies rather than version specific submodules, however this will cross the EOL for Noetic. We will release that version on ROS2.

Sounds good.

raebelchristo-adi:
Updated the PR to contain the `doc:` entry. If everything looks good, please merge this PR. Thank you.

christophebedard (approved):

christophebedard:
Now we just need to [hold for the Noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2025-04-03/)

PR #44907, opened 2025-03-29
Fix branches for control_toolbox on humble

christophfroehlich:
I missed to update the branches here.

According to
https://github.com/ros2-gbp/control_toolbox-release/blob/master/tracks.yaml#L69


christophebedard:
[Holding for Humble sync](https://discourse.ros.org/t/preparing-for-humble-sync-and-patch-release-2025-03-31/42859)

christophebedard (approved):

PR #44918, opened 2025-03-31
#44821: include rti-connext-7.3.0 in base.yaml

lobolanja:
If you're making a new release with bloom please use bloom to create the pull request automatically (except for the naming review request which must be made manually).
If you've already run the release bloom has a `--pull-request-only` option you can use.-->

**As I understand it, previous updates to the rti-connext-dds packages were handled by OSRF—we provided the binaries, and they added them to the ROS bootstrap repositories. @mjcarroll can you confirm that? **

## Package name:

rti-connext-dds-7.3.0

## Package Upstream Source:

TODO (waiting clarification from OSRF on how to proceed)

## Purpose of using this:

This new package will bring new features, improved stability, and bug fixes compared to rti-connext-dds-6.0.1. We’ve also already had ROS 2 users requesting this update.

  - REQUIRED [TODO] 
  - REQUIRED [TODO] we are waiting for getting assistance on this
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available

ROSDISTRO NAME

# The source is here:

http://sourcecode_url

# Checks

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8478,8 +8478,8 @@ rti-connext-dds-7.3.0:
   nixos: []
   ubuntu:
     '*': [rti-connext-dds-7.3.0-ros]
-    jammy: null
     focal: null
+    jammy: null
 rtmidi:
   arch: [rtmidi]
   debian: [librtmidi-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8475,10 +8475,10 @@ rti-connext-dds-6.0.1:  # Supported until ubuntu 24.04
     focal: [rti-connext-dds-6.0.1]
     jammy: [rti-connext-dds-6.0.1]
     noble: [rti-connext-dds-6.0.1]
-rti-connext-dds-7.3.0-ros:  # supported from ubuntu 20.04 focal
+rti-connext-dds-7.3.0-ros:  # supported from ubuntu 24.04 noble
   nixos: []
   ubuntu:
-    '*': [rti-connext-dds-7.3.0-ros]
+    noble: [rti-connext-dds-7.3.0-ros]
 rtmidi:
   arch: [rtmidi]
   debian: [librtmidi-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8472,15 +8472,13 @@ rti-connext-dds-5.3.1:
 rti-connext-dds-6.0.1:  # Supported until ubuntu 24.04
   nixos: []
   ubuntu:
-    bionic: null
     focal: [rti-connext-dds-6.0.1]
     jammy: [rti-connext-dds-6.0.1]
     noble: [rti-connext-dds-6.0.1]
-rti-connext-dds-7.3.0:
+rti-connext-dds-7.3.0:  # supported from ubuntu 20.04 focal
   nixos: []
   ubuntu:
     '*': [rti-connext-dds-7.3.0]
-    bionic: null
 rtmidi:
   arch: [rtmidi]
   debian: [librtmidi-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8469,10 +8469,17 @@ rti-connext-dds-5.3.1:
   ubuntu:
     bionic: [rti-connext-dds-5.3.1]
     focal: [rti-connext-dds-5.3.1]
-rti-connext-dds-6.0.1:
+rti-connext-dds-6.0.1: # Supported until ubuntu 24.04
   nixos: []
   ubuntu:
-    '*': [rti-connext-dds-6.0.1]
+    bionic: null
+    focal: [rti-connext-dds-6.0.1]
+    jammy: [rti-connext-dds-6.0.1]
+    noble: [rti-connext-dds-6.0.1]
+rti-connext-dds-7.3.0:
+  nixos: []
+  ubuntu:
+    '*': [rti-connext-dds-7.3.0]
     bionic: null
 rtmidi:
   arch: [rtmidi]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8475,7 +8475,7 @@ rti-connext-dds-6.0.1:  # Supported until ubuntu 24.04
     focal: [rti-connext-dds-6.0.1]
     jammy: [rti-connext-dds-6.0.1]
     noble: [rti-connext-dds-6.0.1]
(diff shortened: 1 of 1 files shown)


lobolanja:
@nuclearsandwich do we need to address this PR before RMW freeze?

mjcarroll (approved):

mjcarroll:
One approval from me, I'll let @nuclearsandwich get the second.

rosdep/base.yaml
```diff
@@ -8469,11 +8469,16 @@ rti-connext-dds-5.3.1:
...
-rti-connext-dds-6.0.1:
+rti-connext-dds-6.0.1:  # Supported until ubuntu 24.04
   nixos: []
   ubuntu:
-    '*': [rti-connext-dds-6.0.1]
-    bionic: null
+    focal: [rti-connext-dds-6.0.1]
+    jammy: [rti-connext-dds-6.0.1]
+    noble: [rti-connext-dds-6.0.1]
+rti-connext-dds-7.3.0-ros:  # supported from ubuntu 24.04 noble
```
nuclearsandwich, line 8478: rosdep convention is to match the Debian package name but I actually considered matching the previous RTI package name in #44972 and considering the `-ros` an implementation detail.

I think the main thing it affects is needing to know which connext versions use a `ros`-suffixed package when specifying connext version in package.xml dependencies. Whether that's enough of a gripe to be worth defying convention is not up to me.

rosdep/base.yaml
```diff
@@ -8469,11 +8469,16 @@ rti-connext-dds-5.3.1:
...
+  nixos: []
+  ubuntu:
+    noble: [rti-connext-dds-7.3.0-ros]
```
nuclearsandwich, line 8481: In order to maintain forward compatibility we usually invert the definition here:
```suggestion
  ubuntu:
    '*': [rti-connext-dds-7.3.0-ros]
    jammy: null
    focal: null
```

This will allow us to use this package in 26.04 without requiring re-definition. The method by which we currently populate bootstrap repositories will insure this package is propagated to 26.04 when we first create it. However, if the Connext maintainers prefer explicitly defining which distros are supported for each rosdep definition I'm alright with that.

rosdep/base.yaml
```diff
@@ -8469,11 +8469,16 @@ rti-connext-dds-5.3.1:
...
     focal: [rti-connext-dds-5.3.1]
-rti-connext-dds-6.0.1:
+rti-connext-dds-6.0.1:  # Supported until ubuntu 24.04
   nixos: []
   ubuntu:
-    '*': [rti-connext-dds-6.0.1]
-    bionic: null
+    focal: [rti-connext-dds-6.0.1]
+    jammy: [rti-connext-dds-6.0.1]
+    noble: [rti-connext-dds-6.0.1]
```
nuclearsandwich, line 8477: I would actually suggest leaving the rti-connext-dds-6.0.1 definition as-is. The suite could actually be completely elided since bionic is no longer a checked / supported distro but it doesn't do much harm to leave it as it was.

rosdep/base.yaml
lobolanja: @nuclearsandwich 

I made that change mostly by intuition, since I wasn’t entirely sure how to represent it.
However, I’d prefer to keep the `-ros` suffix as an implementation detail.

To do that, would it be enough to change both parts to the following?

```
rti-connext-dds-7.3.0:
  nixos: []
  ubuntu:
    '*': [rti-connext-dds-7.3.0]
    jammy: null
    focal: null
```

rosdep/base.yaml
lobolanja: @nuclearsandwich Just a quick question to avoid any issues—can you confirm whether I should use `rti-connext-dds-7.3.0` instead of `rti-connext-dds-7.3.0-ros` in [this PR](https://github.com/ros2/rmw_connextdds/pull/181) in `rti_connext_dds_cmake_module/package.xml`?

rosdep/base.yaml
lobolanja: check changes

rosdep/base.yaml
lobolanja: done!

rosdep/base.yaml
```diff
@@ -8474,6 +8474,12 @@ rti-connext-dds-6.0.1:
   ubuntu:
     '*': [rti-connext-dds-6.0.1]
     bionic: null
+rti-connext-dds-7.3.0:
```
mjcarroll, line 8477: This line should match what is in the downstream package.xml (and it does).

rosdep/base.yaml
```diff
@@ -8474,6 +8474,12 @@ rti-connext-dds-6.0.1:
...
+  nixos: []
+  ubuntu:
+    '*': [rti-connext-dds-7.3.0]
```
mjcarroll, line 8480: This line should match the name of the package that is to be installed from debs (so should have the `-ros` suffix).

nuclearsandwich-ai, line 8480:
```suggestion
    '*': [rti-connext-dds-7.3.0-ros]
```

Apologies for creating ambiguity here @lobolanja. The package is keeping the `-ros` suffix. My comments about dropping it only applied to the rosdep key name.

It's probably best just to keep it everywhere and follow convention. Sorry for the confusion.

lobolanja: @nuclearsandwich @mjcarroll   I think we can go with that approach—keeping the `-ros` suffix only for the package that’s installed via the .deb, and using `rti-connext-dds-7.3.0` as the key and in the package.xml.

At the end of the day, the real dependency is Connext 7.3.0, so the `-ros` suffix doesn’t really matter from a functionality standpoint, it is just to differentiate the package from RTI regular Debian packages.

lobolanja: @mjcarroll thanks for clarifying.

rosdep/base.yaml
lobolanja: Thanks! @mjcarroll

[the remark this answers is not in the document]
lobolanja: @nuclearsandwich @mjcarroll it is fine tho have this "warning" ?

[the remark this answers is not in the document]
nuclearsandwich: Intentional. Discussed here: https://github.com/ros/rosdistro/pull/44918#discussion_r2032162264

nuclearsandwich (approved):

PR #44920, opened 2025-03-31
Add python3-einops-pip

stergiosba:

## Package name:

einops

## Package Upstream Source:

[PyPI - Einops](https://pypi.org/project/einops/)

## Purpose of using this:

Flexible and powerful tensor operations for readable and reliable code.
Supports numpy, pytorch, tensorflow, jax, and [others](https://pypi.org/project/einops/#supported-frameworks).

Distro packaging links:

- Debian: https://pypi.org/project/einops/
  - Its from PyPi
- Ubuntu: https://pypi.org/project/einops/
  - Its from PyPi

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5749,6 +5749,13 @@ python3-easydict:
     focal:
       pip:
         packages: [easydict]
+python3-einops-pip:
+  debian:
+    pip:
+      packages: [einops]
+  ubuntu:
+    pip:
+      packages: [einops]
 python3-elasticsearch:
   debian: [python3-elasticsearch]
   fedora: [python3-elasticsearch]


christophebedard:

christophebedard (approved):

christophebedard:
Note that we require 2 approvals for new rosdep keys.

stergiosba:
Not sure I understood what you meant. Can you spell it out for me? If you can't that is ok

christophebedard:
If you're looking to release a package (meaning that your package gets built into a binary by the buildfarm), it cannot have a dependency (declared in its `package.xml`) that is actually installed from `pip`. It could only depend on something from the Ubuntu system packages (or the equivalent for other platforms, e.g., Debian). For example, see the `python3-elasticsearch` key below the key you added in `rosdep/python.yaml`.

This PR is OK, because you can still use this rosdep key in your package so that the dependency gets installed when users clone your package into their workspace and run `rosdep install` before building your package from source. You just can't release your package on the buildfarm; people have to build it from source.

Does that help?

sloretz (approved):
Note to future viewers: [`python3-einops` is in Debian Sid](https://packages.debian.org/sid/all/science/python3-einops) and [in Ubuntu Plucky](https://packages.ubuntu.com/plucky/python3-einops). In a year or so this should become a normal rosdep key instead of a pip key.

stergiosba:
This was very clear thank you.

PR #44939, opened 2025-04-01
Document 'rosdep-needs-second-approval' label in REVIEW_GUIDELINES.md

christophebedard:
This was briefly discussed in the 2025-04-01 ROS PMC meeting.

I've noticed that the 2-approval requirement for rosdep key PRs often leads to PRs falling through the cracks. I think this is because it’s hard to know if a PR is just waiting for a 2nd approval by looking at the list of PRs, especially if the review bot approved it. PR submitters don’t really know about this requirement and just see their PR being approved and waiting there.

To help make PRs that are only waiting on a second approval (and merge) more visible, the `rosdep-needs-second-approval` label can be used. Document this.

For example, I added the label to this PR: https://github.com/ros/rosdistro/pull/44920.


REVIEW_GUIDELINES.md
```diff
@@ -69,6 +69,7 @@ There are a few different types of pull requests that are opened against this re
     * New rosdep keys are typically only accepted for supported Ubuntu distros which have ROS releases; the ROS support timeline generally follows the Ubuntu LTS support timeline.
     * If there is no good alternative for a pip package to add to rosdep keys, use `python3-PACKAGENAME-pip` to name the key.
     * If a key conforms to all standards above, then it can be approved. Once the pull request has two approvals, it will be merged.
+        * To help make pull requests that are only waiting for a second approval & merge more visible, the `rosdep-needs-second-approval` label can be used.
```
tfoote, line 72:
```suggestion
        * To help make pull requests that are only waiting for a second approval & merge more visible, the `rosdep-needs-second-approval` label can be used.
        * Also please request review explicitly from the other person on the rosdistro review rotation.
```

tfoote (approved):
This looks good. i'd also like to suggest the best practice is to explicitly request the review for the other rosdistro maintainer for the week.

christophebedard:
> This looks good. i'd also like to suggest the best practice is to explicitly request the review for the other rosdistro maintainer for the week.

Good call, that's probably going to help even more.

christophebedard:
Thanks for the review!

PR #34558, opened 2022-09-14
Add jax and jaxlib mapping

halejo-as:
Please add the following dependency to the rosdep database.

## Jax and JaxLib:

`jax`
`jaxlib`

## Package Upstream Source:

[jax](https://github.com/google/jax)

## Purpose of using this:

What’s new is that JAX uses [XLA](https://www.tensorflow.org/xla) to compile and run your NumPy programs on GPUs and TPUs.

## Links to Distribution Packages

- Debian:
  https://pypi.org/project/jax/
  https://pypi.org/project/jaxlib/
- Ubuntu: 
  https://pypi.org/project/jax/
  https://pypi.org/project/jaxlib/

PR #44942, opened 2025-04-02
Add python3-jraph-pip

stergiosba:

## Package name:

jraph

## Package Upstream Source:

[PyPI - Jraph](https://pypi.org/project/jraph/)

## Purpose of using this:

Jraph - A library for graph neural networks in jax. 

Distro packaging links:

- Debian: https://pypi.org/project/jraph/
  - Its from PyPi
- Ubuntu: https://pypi.org/project/jraph/
  - Its from PyPi

## Possible issues:
1) The proposed addition, `jraph`, depends on [`jax`](https://github.com/jax-ml/jax) and installing it via `rosdep` on my local machine installs `jax` but only its CPU version. Reminder:`jax` code can be executed by CPU and GPU but using `rosdep`. Is there a way to differentiate the two? Is it necessary?

2) I also see that `jax` is **not** added as a `rosdep` key. Should we add `jax` first and then any dependent package? I would love to complete the #34558 PR first if that is necessary.


sloretz:
> Reminder:jax code can be executed by CPU and GPU but using rosdep. Is there a way to differentiate the two? Is it necessary?

[I see what you mean](https://docs.jax.dev/en/latest/installation.html#installation). It looks like jax recommends installing different dependency groups ([PEP 735](https://peps.python.org/pep-0735/)?) depending on what hardware the machine has. I don't know of a good way to map those to rosdep keys.

I would recommend that `rosdep` installs the CPU version only by default, with users needing to `pip install jax[group-matching-their-hardware]` manually to use GPUs or TPUs. I'll bring this up in the next ROS PMC meeting to see if anyone else has more ideas.

> I also see that jax is not added as a rosdep key. Should we add jax first and then any dependent package? 

If your package is only using APIs from `jraph`, then I don't think you need to add a rosdep key for `jraph`s dependencies, including `jax`. If your package is also using `jax` APIs, then I do think you should add a rosdep key for `jax`. That way if your package were to stop depending on `jraph` in the future, it would still depend on `jax`.

stergiosba:
Typically we install `jax` firstly and then install everything depending on it like `jraph`. Similarly to `numpy` for example. On a second thought I do thing that `jax` should be added explicitly as a key. In this case do we add two keys one for CPU and one for specific architectures? Starting with cuda and seeing from there?

stergiosba:
@sloretz So what is an update on this?

ros-discourse:

https://discourse.ros.org/t/ros-pmc-minutes-for-april-8-2025/43138/1

stergiosba:
I just paste the relevant part from the ros-discourse meeting:

**Conclusion**: 
`rosdep` doesn’t support this, and probably won’t with some additional feature development. For now, we can use dependency-specific keys like `jax-nvidia` to differentiate from `jax`

So if my understanding is correct. 
The user will have to install the `jax` version they want, either `jax` for CPU or `jax[cuda]` for NVIDIA GPUs and then install any `jax` dependent packages likes `jraph` and that should "just work" right?

stergiosba:
Its unclear to me what information you need. Can you be more specific. Thanks!

sloretz:
Hi @stergiosba!

In short, I think this key is fine as is.

> The user will have to install the jax version they want, either jax for CPU or jax[cuda] for NVIDIA GPUs and then install any jax dependent packages likes jraph and that should "just work" right?

Yes, I think that's correct. We talked about adding a rosdep key like `jax-cuda` that would map to `jax[cuda]`, but there's a couple issues:

* If your package doesn't care what hardware is used, then adding `jax-cuda` would unnecessarily limit your users.
* rosdep doesn't have a way to check if `jax[cuda]` vs `jax` have been installed, so if a user asks for `jax[cuda]`, but `jax` is already installed then the cuda support might not be added.

Having a `jraph` key that installs the CPU version of `jax` seems reasonable. Users with special hardware will have to manually install `jax[...]` to get hardware specific versions.

sloretz (approved):

stergiosba:
Two points that are also relevant but I didn't mention

**Point 1**
---
Even if someone, let's say by mistake, installs `jax[cuda]` but they don't have a GPU at all, then `jax` reverts to CPU automatically and the code runs, albeit with the expected discrepancies in performance that the two different executors bring to the table. 

So adding the `jax-cuda` rosdep key doesn't limit the users from this perspective according to my understanding.

**Point 2**
---
This is based on my knowledge of by using both CPU and GPU versions of `jax` using pip:
If `jax` (the CPU version) is already installed (either via pip or rosdep) and then the user wants to switch to GPU, when they do:
 ```
 pip install -U "jax[cuda12]"
```
This will only bring in the added `cuda` related dependencies with suggestive names. For example:  `jax-cuda12-plugin` and `cuda12-pjrt`.

I will open another PR for `jax` and `jax-cuda` together if the tests pass locally.

Thanks again for the discussion.

stergiosba:
@sloretz I don't know who to ask for a 2nd review on this. gentle ping :)

mjcarroll (approved):

PR #44972, opened 2025-04-02
Add keys for rti-connext-dds-7.3.0.

nuclearsandwich:
Available for both Ubuntu Noble amd64 and arm64. Unlike previous releases where we used an empty package for arm64.

## Package name:

rti-connext-dds-7.3.0

## Package Upstream Source:
Sourced from the [RTI APT repository](https://community.rti.com/static/documentation/connext-dds/current/doc/manuals/debian_packages/repo.html) imported into ros_bootstrap.

## Purpose of using this:

Updated package for a ROS 2 middleware implementation.

Distro packaging links:

https://community.rti.com/static/documentation/connext-dds/current/doc/manuals/debian_packages/repo.html

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -8474,6 +8474,11 @@ rti-connext-dds-6.0.1:
   ubuntu:
     '*': [rti-connext-dds-6.0.1]
     bionic: null
+rti-connext-dds-7.3.0:
+  nixos: []
+  ubuntu:
+    '*': [rti-connext-dds-7.3.0]
+    jammy: null
 rtmidi:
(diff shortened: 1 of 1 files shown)


christophebedard:
Should we close #44918?

nuclearsandwich:
> Should we close #44918?

Actually, that PR is probably closer to what needs to happen as it uses the custom `-ros` package rather than the official RTI debs. (The `-ros` deb includes the non-commercial license).

emersonknapp:
Closing in favor of https://github.com/ros/rosdistro/pull/44918 which has been merged. If I've misread the situation, let me know!

PR #45033, opened 2025-04-08
Try making Build & Test action use ubuntu latest

### LINKED ISSUE #45032: Upgrade github actions to use newer ubuntu version than 20.04
```
[rosdistro / rosdep checks (3.8)](https://github.com/ros/rosdistro/actions/runs/14337415587/job/40187953787)
This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101
```

sloretz:
Hopefully Fixes #45032

### DIFF
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -12,7 +12,7 @@ permissions:
 jobs:
   checks:
     name: rosdistro / rosdep checks
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         python-version: [3.8]
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -35,7 +35,7 @@ jobs:
       run: pytest -s test
   yamllint:
     name: Yaml Linting
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         python-version: [3.8]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Hopefully this will not be fragile and need to be pinned.

.github/workflows/build_test.yaml
```diff
@@ -12,7 +12,7 @@ permissions:
 jobs:
   checks:
     name: rosdistro / rosdep checks
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
nuclearsandwich, line 15:
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/build_test.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
       run: pytest -s test
   yamllint:
     name: Yaml Linting
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
nuclearsandwich, line 38:
```suggestion
    runs-on: ubuntu-24.04
```

nuclearsandwich (commented):
I have a preference for using an explicit Ubuntu version. GitHub, rather than we, are in charge of when ubuntu-latest rolls forward and it usually seems to happen right when we don't want it to. By setting the date we get into situations like this where we get hit by a brownout and need to update but if we use ubuntu-latest we have surprise rolls forward. So I see it as a kind of pick-your-poison.

If the `ubuntu-latest` flavor is preferred by rosdistro maintainers I won't put my foot down.

ros-discourse:

https://discourse.ros.org/t/ros-pmc-minutes-for-april-8-2025/43138/1

PR #45267, opened 2025-04-21
Use ros2-gbp release repository for Rolling.

nuclearsandwich:
As we prepare for the kilted branching, things go much smoother if all repositories are already in the ros2-gbp GitHub org.

There is no ros2-gbp PR since this repository was previously mirrored into ros2-gbp. I've updated the release repo on ros2-gbp with all of the new changes in the other release repository.

FYI @floweisshardt, @zvezdan94

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -8224,7 +8224,7 @@ repositories:
     release:
       tags:
         release: release/rolling/{package}/{version}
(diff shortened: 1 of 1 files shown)


Yadunund (approved):

floweisshardt:
Thanks for taking action on this!

PR #45516, opened 2025-04-28
remove skyway

nakakura:
I removed it because my package is giving build errors.
I am sorry.

### DIFF
--- a/melodic/distribution.yaml
+++ b/melodic/distribution.yaml
@@ -13320,21 +13320,6 @@ repositories:
       url: https://gitlab.com/InstitutMaupertuis/simple_rviz_plugin.git
       version: melodic
     status: maintained
-  skyway:
-    doc:
-      type: git
-      url: https://github.com/ntt-t3/skyway_for_ros.git
-      version: main
-    release:
-      tags:
-        release: release/melodic/{package}/{version}
-      url: https://github.com/ntt-t3/skyway_for_ros-release.git
-      version: 0.0.2-3
-    source:
-      type: git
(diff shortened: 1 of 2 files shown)


nakakura:
Here is the relevant error
https://build.ros.org/job/Nbin_ufhf_uFhf__skyway__ubuntu_focal_armhf__binary/777/display/redirect

noetic/distribution.yaml
```diff
@@ -11664,21 +11664,6 @@ repositories:
       url: https://github.com/mikeferguson/simple_grasping.git
       version: ros1
     status: maintained
-  skyway:
-    doc:
-      type: git
-      url: https://github.com/ntt-t3/skyway_for_ros.git
-      version: main
-    release:
```
sloretz, line 11672: Please remove only the release stanza, and keep the `doc` and `source` stanza. That way your package will be listed on index.ros.org, but won't be built by the buildfarm.

melodic/distribution.yaml
```diff
@@ -13320,21 +13320,6 @@ repositories:
       url: https://gitlab.com/InstitutMaupertuis/simple_rviz_plugin.git
       version: melodic
     status: maintained
-  skyway:
```
sloretz, line 13323: Please don't change `melodic/distribution.yaml`. ROS Melodic is end-of-life, so this file is essentially frozen.

sloretz (changes requested):

nakakura:
Thanks for the comment.
I removed only the release stanza in noetic/distribution.yaml.

sloretz (approved):

PR #45572, opened 2025-04-30
Remove rai_interfaces from rolling

knicked:
Please remove rai_interfaces from the rolling distribution index. When creating the initial PR I meant to add it to Humble and Jazzy distributions only, but I also included rolling by mistake.
The package is giving build errors on rolling and we don't actually target this distribution with this package, so it can be removed.

Sorry for the inconvenience.


mjcarroll (approved):

mjcarroll:
You may also want to add it to Kilted, as this happened after the branch.

PR #45576, opened 2025-04-30
Remove specific Python versions from NixOS rosdep entries

wentasah:
NixOS provides Python packages under names that include specific Python version (e.g. `python312Packages` for Python 3.12) as well as under unversioned names (`python3Packages`), which provide packages for the currently default Python version. rosdep entries should use the unversioned names to be future-proof.

This PR replaces all versioned names with unversioned. The majority of existing entries of already unversioned.


mjcarroll (approved):

sloretz (approved):
Thank you for the PR!

PR #45040, opened 2025-04-08
ardrone_sdk and ardrone_sumo from ardrone-ros2 in humble, jazzy and rolling

vtalpaert:
Please add the following dependency to the rosdep database.

## Package name:

- ardrone_sdk
- ardrone_sumo

## Package Upstream Source:

https://github.com/vtalpaert/ardrone-ros2

## Purpose of using this:

Parrot released the ARDRONE SDK (ARSDK for short) version 3 almost 9 years ago. Some projects made use for ROS1 notably
- https://github.com/AutonomyLab/parrot_arsdk
- https://github.com/arnaud-ramey/rossumo

Their build process is not suited for release due to the `repo init` which git clones during the installation process.
Instead, we use github actions to build the library into a tarball available for download at
- https://github.com/vtalpaert/ardrone-sdk-native

The proposed packages to index make ARSDK3 available to developers, users should include a `ardrone_sdk` dependency in package.xml/CmakeLists.txt, as showed in the `ardrone_sumo` package.

Note that the build requires an intermediate `arsdk3` package before building `ardrone_sdk` for release

# Please Add This Package to be indexed in the rosdistro.

- humble
- jazzy
- rolling

# The source is here:

https://github.com/vtalpaert/ardrone-ros2

The [license permission from Parrot](https://github.com/Parrot-Developers/ARSDKBuildUtils/blob/master/LICENSE.md) for the SDK is BSD 3. Therefore the `ardrone_sdk` package includes the license file.

# Checks
 - [x] All packages have a declared license in the package.xml
 - [x] This repository has a LICENSE file
 - [x] This package is expected to build on the submitted rosdistro

PR #45602, opened 2025-05-02
ardrone_ros: 1.0.0-1 in 'rolling/distribution.yaml' [bloom]

vtalpaert:
Increasing version of package(s) in repository `ardrone_ros` to `1.0.0-1`:

- upstream repository: https://github.com/vtalpaert/ardrone-ros2.git
- release repository: https://github.com/ros2-gbp/ardrone_ros-release.git
- bloom version: `0.12.0`

## ardrone_sdk

```
* Export Parrot ARSDK3 as a library with all cmake confguration to use find_package
* Includes the headers, shared libraries (.so), native unix samples and Parrot licence as required
* Contributors: Victor Talpaert
```

## ardrone_sumo

```
* ROS2 node to control the Jumping Sumo Parrot drone
* Connect over private WiFi to retrieve the video feed, battery level and send motor commands
* No IMU data, no standard units for motor commands (currently in percentage of maximum value)
* Contributors: Victor Talpaert
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -439,6 +439,18 @@ repositories:
       version: master
     status: developed
   ardrone_ros:
+    doc:
+      type: git
+      url: https://github.com/vtalpaert/ardrone-ros2.git
+      version: main
+    release:
+      packages:
+      - ardrone_sdk
+      - ardrone_sumo
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/ardrone_ros-release.git
+      version: 1.0.0-1
     source:
       type: git
       url: https://github.com/vtalpaert/ardrone-ros2.git


vtalpaert:
The reviewer should pay attention to two things

# Licence

The ARSDK library is distributed in open source by Parrot. The licence can be found [here](https://github.com/Parrot-Developers/ARSDKBuildUtils/blob/master/LICENSE.md), it allows to distribute the binaries as long as the release includes the licence. The package `ardrone_sdk` includes the licence as required.

See comment by @wjwwood in [indexing PR](https://github.com/ros/rosdistro/pull/45040#issuecomment-2791618905)

# arsdk3 package as build dependency but not released

To obtain a node for the Jumping Sumo robot, here are all the build steps:

- The Parrot ARSDK library is built in [a separate github CI](https://github.com/vtalpaert/ardrone-sdk-native) as a tarball. This is a separate step since the original build process uses a git clone step. The versioned tarball allows consecutive deterministic builds 
- We use CMake ExternalProject to download the tarball. To correctly export the library .so files, we need to list them in the cmake configure step. Only ExternalProject only downloads the tar at build time so we cannot glob in the directory. To this end, we first create a `arsdk3` ament package which installs inside the `ardrone_sdk` directory
- Then, we declare the `ardrone_sdk` package. It can effectively list all the libraries it will export using cmake glob. This library includes the headers, shared libraries, native unix samples and Parrot licence as required
- Finally, we release the `ardrone_sumo` package that requires the previous package to build and run. Other developers will only require the `ardrone_sdk` dependency to add their own nodes for other Parrot drones
 
Since the arsdk3 package is only for building, I added it to the [ignored packages](https://github.com/ros2-gbp/ardrone_ros-release/blob/master/rolling.ignored). But it results in rosdep not finding the package

```bash
==> git-bloom-generate -y rosdebian --prefix release/rolling rolling -i 1 --os-name ubuntu
ROS Distro index file associate with commit '9e1ed5591ec7a084027f3d0f3f05dd922184a44b'
New ROS Distro index url: 'https://raw.githubusercontent.com/ros/rosdistro/9e1ed5591ec7a084027f3d0f3f05dd922184a44b/index-v4.yaml'
Explicitly ignoring package 'arsdk3' because it is in the `rolling.ignored` file.
Generating source debs for the packages: ['ardrone_sumo', 'ardrone_sdk']
Debian Incremental Version: 1
Debian Distributions: ['noble']
Releasing for rosdistro: rolling

Pre-verifying Debian dependency keys...
Running 'rosdep update'...
Could not resolve rosdep key 'arsdk3'
Failed to resolve arsdk3 on ubuntu:noble with: Error running generator: Failed to resolve rosdep key 'arsdk3', aborting.
arsdk3 is depended on by these packages: ['ardrone_sdk']
<== Failed
❌  The following generator action reported that it is missing one or more
    rosdep keys, but that the key exists in other platforms:
'['/usr/bin/git-bloom-generate', '-y', 'rosdebian', '--prefix', 'release/rolling', 'rolling', '-i', '1', '--os-name', 'ubuntu']'

If you are absolutely sure that this key is unavailable for the platform in
question, the generator can be skipped and you can proceed with the release.
Skip generator action and continue with release [y/N]? y

Action skipped, continuing with release.
```

**Is this the correct way to exclude the package ?** I do not want arsdk3 to be release since it is only a CMakeLists.txt package

mjcarroll (approved):

PR #45645, opened 2025-05-06
Change development branch of leo packages to ros2 in kilted

bjsowa:


christophebedard (approved):
Same thing I mentioned in https://github.com/ros/rosdistro/pull/45646#issuecomment-2856054842. It's not required for this PR, but you should also update the `devel_branch` for Kilted in the corresponding release repos.

PR #45653, opened 2025-05-07
Added touchlab-comm-py rosdep

VladimirIvan:

## Package name:

touchlab-comm-py

## Package Upstream Source:

[https://pypi.org/project/touchlab-comm-py/](https://pypi.org/project/touchlab-comm-py/)

## Purpose of using this:

This is a python wrapper for Touchlab tactile sensors.


mjcarroll (approved):

christophebedard (approved):

PR #45674, opened 2025-05-09
Add live555_vendor to source index

roehling:

jazzy, rolling

# The source is here:

https://github.com/fkie/live555_vendor
http://live555.com/liveMedia/

This is a vendoring-friendly copy of the live555 multimedia library which has been augmented with a CMake configuration for convenient use in ROS packages. The library itself has been taken unmodified from upstream.

# Checks

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -4134,6 +4134,12 @@ repositories:
       url: https://github.com/adityapande-1995/linux_isolate_process.git
       version: main
     status: maintained
+  live555_vendor:
+    source:
+      type: git
+      url: https://github.com/fkie/live555_vendor.git
+      version: main
+    status: maintained
   log_view:
     doc:
       type: git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -3582,6 +3582,12 @@ repositories:
       url: https://github.com/adityapande-1995/linux_isolate_process.git
       version: main
     status: maintained
+  live555_vendor:
+    source:
+      type: git
+      url: https://github.com/fkie/live555_vendor.git
+      version: main
+    status: maintained
   log_view:
     doc:
       type: git


emersonknapp:
New package review checklist

  - [ ] Top level license file:
- [ ] License is [OSI-approved](https://opensource.org/licenses):
- [ ] Public source repo:
- [ ] Source repository contains ROS packages

emersonknapp (approved):
Approved, but I have one small question I want to double check at tomorrow's ROS 2 PMC meeting before merging.

roehling:
I didn't realize that FSFAP is not OSI approved. I was mainly looking for a permissive and short license, but I can change it if need be.

emersonknapp:
Yes, please! If you can change out that license for one in https://opensource.org/licenses - then we can merge with no questions.

ros-discourse:

https://discourse.ros.org/t/ros-pmc-minutes-for-may-13-2025/43771/1

peci1:
What's a little weird is that this package [is available in Debian](https://pkgs.org/search/?q=liblivemedia-dev), but [not in Ubuntu (since after Focal)](https://launchpad.net/ubuntu/+source/liblivemedia). I tried googling why would Ubuntu drop it, but I haven't succeeded.

However, now there is the situation that there is a package that could be a ROS package on some systems and a system package on other systems. As [already mentioned](https://github.com/ros/rosdistro/pull/37717#issuecomment-2032220323), current rosdep cannot resolve a dependency that is partly system and partly ros index.

But I see this PR adds only source info, which would probably be okay.

peci1:
Oh, deep research has actually found something: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981451 . This bugreport tells the library was removed from Debian.

And here's what confused me: pkgs.org apparently browses also non-official repos. liblivemedia-dev is only available from https://www.deb-multimedia.org/ and not from the official repos. So this library is no longer available in any ROS2-relevant distros.

roehling:
It is correct that live555 was removed due to non-free code, but that code has actually been replaced with LGPL licensed code as of version 2025.04.25 ([changelog](http://www.live555.com/liveMedia/public/changelog.txt)). So it is possible that live555 will be reintroduced to Debian (and thereby Ubuntu) at some point, but for now, having a vendor package is the only way to use the library with the current ROS 2 releases.

roehling:
Yes, that is the goal indeed.

mjcarroll (approved):

PR #45685, opened 2025-05-10
jsk_3rdparty: 2.1.30-1 in 'noetic/distribution.yaml' [bloom]

k-okada:
Increasing version of package(s) in repository `jsk_3rdparty` to `2.1.30-1`:

- upstream repository: https://github.com/jsk-ros-pkg/jsk_3rdparty.git
- release repository: https://github.com/tork-a/jsk_3rdparty-release.git
- distro file: `noetic/distribution.yaml`
- bloom version: `0.11.2`
- previous version for package: `2.1.28-1`

## aques_talk

```
* CI:aques_talk : run aques_talk only when download was succeeded (#534 <https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/534>)
* Add ROS-O 24.04 test (#521 <https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/521>)
* Contributors: Kei Okada, Yoshiki Obinata
```

## assimp_devel

- No changes

## bayesian_belief_networks

```
* bayesian_belief_networks: fix deb build (#525 <https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/525>) --single-version-externally-managed only available on ubuntu<20
  checked with 'git clean -xfd .; bloom-generate rosdebian --os-name=ubuntu --os-version=noble --ros-distro one  --skip-pip; dpkg-buildpackage -b -us -u'
* Add ROS-O 24.04 test (#521 <https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/521>)
* Contributors: Kei Okada
```

## chaplus_ros

- No changes

## collada_urdf_jsk_patch

```
* [ros-o] collada_urdf_jsk_patch: skip cmake/pkg config generation (#517 <https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/517>)
* Contributors: Kei Okada
```

## dialogflow_task_executive

```
* Add ROS-O 24.04 test (#521 <https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/521>)
* Contributors: Kei Okada
```

## downward

- No changes

## emotion_analyzer

```
* Add Emotion Analyzer (#527 <https://github.com/jsk-ros-pkg/jsk_3rdparty/issues/527>)
* Contributors: Ayaha Nagata, Yoshiki Obinata
```

## ff

- No changes

## ffha

- No changes

## gdrive_ros

- No changes

## google_chat_ros

```
[...]

PR #45713, opened 2025-05-12
Remove slic package

sloretz:
I noticed this package had no license when reviewing https://github.com/ros/rosdistro/pull/45685#pullrequestreview-2833872745

It seems like it comes from here https://github.com/PSMM/SLIC-Superpixels . Without a license it's not clear if the buildfarm can redistribute it in binary form. This PR removes the package from the buildfarm.

@k-okada FYI

https://github.com/jsk-ros-pkg/jsk_3rdparty/blob/a3b5eb682efc0f0e0ad6ca43faae4ad5a96867f3/3rdparty/slic/package.xml#L15

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4759,7 +4759,6 @@ repositories:
       - rosping
       - rostwitter
       - sesame_ros
-      - slic
       - switchbot_ros
       - voice_text
       - webrtcvad_ros


k-okada:
thanks, I also removed `<run_depend>slic</run_depned>` from jsk_3rdparty/package.xml.
To note, there are unanswered question about SLIC license at https://github.com/PSMM/SLIC-Superpixels/issues/14
SLIC algorithm is alredy implimented at opencv_contrib, so we could  use this version https://github.com/PSMM/SLIC-Superpixels/pull/5#issuecomment-243398615

PR #45782, opened 2025-05-14
Switch source and doc branches to kilted for open-rmf repos

Yadunund:
Sources branches for `kilted` now exist in open-rmf repos.

### DIFF
--- a/kilted/distribution.yaml
+++ b/kilted/distribution.yaml
@@ -5966,7 +5966,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/open-rmf/rmf_cmake_uncrustify.git
-      version: kilted
+      version: rolling
     release:
       tags:
         release: release/kilted/{package}/{version}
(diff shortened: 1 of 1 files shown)


christophebedard (approved):
Will you also update the `devel_branch`es in the corresponding release repos? E.g. https://github.com/ros2-gbp/ament_cmake_catch2-release/blob/0ede69c6ee0186bde98113881ae7757723ad9046/tracks.yaml#L147

kilted/distribution.yaml
```diff
@@ -5960,13 +5960,13 @@ repositories:
...
       url: https://github.com/open-rmf/rmf_building_map_msgs.git
-      version: main
+      version: kilted
     status: developed
   rmf_cmake_uncrustify:
     doc:
       type: git
       url: https://github.com/open-rmf/rmf_cmake_uncrustify.git
-      version: rolling
+      version: kilted
```
christophebedard, line 5969: CI is complaining that this repo doesn't have a `kilted` branch

mxgrey: It looks like this repo hasn't seen any changes in 4 years. We can probably just keep it on `rolling`.

Yadunund: We should probably just remove that repo from Rolling since it was a temporary package and changes from it landed in `ament_uncrustify`.

christophebedard (approved):

PR #45793, opened 2025-05-15
phidgets_drivers: Switch to dedicated branch "kilted"

mintar:


christophebedard (approved):
It's not required for this PR, but you should also update the `devel_branch` in the release repo: https://github.com/ros2-gbp/phidgets_drivers-release/blob/ea7ff33c6f0377ccbdae964427b63cc52b978401/tracks.yaml#L195

PR #45803, opened 2025-05-16
Adding hokuyo-for-ROS2 to documentation index for distro

marcosvcr:

## Purpose of using this:
We tested the hokuyo_node and urg_node2 libraries, but neither succeeded in reading data from the sensor. So, I created a new library that links the Hokuyo drivers with Ament.

Distro packaging links:

- Debian: https://packages.debian.org/buster/ruby
- Ubuntu: sudo apt install ruby-full

ROS Humble

# The source is here:

https://github.com/marcosvcr/hokuyo-for-ROS2.git

# Checks
 - [ x] All packages have a declared license in the package.xml
 - [ x] This repository has a LICENSE file
 - [ x] This package is expected to build on the submitted rosdistro

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3525,14 +3525,14 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/hls_lfcd_lds_driver.git
       version: rolling-devel
     status: developed
-  hokuyo-for-ROS2:
+  hokuyo_node2:
     doc:
       type: git
-      url: https://github.com/marcosvcr/hokuyo-for-ROS2.git
+      url: https://github.com/marcosvcr/hokuyo_node2.git
       version: main
     source:
       type: git
-      url: https://github.com/marcosvcr/hokuyo-for-ROS2.git
+      url: https://github.com/marcosvcr/hokuyo_node2.git
       version: main
     status: developed
   homing_local_planner:
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -3525,6 +3525,16 @@ repositories:
       url: https://github.com/ROBOTIS-GIT/hls_lfcd_lds_driver.git
       version: rolling-devel
     status: developed
+  hokuyo-for-ROS2:
+    doc:
+      type: git
+      url: https://github.com/marcosvcr/hokuyo-for-ROS2.git
+      version: main
+    source:
+      type: git
+      url: https://github.com/marcosvcr/hokuyo-for-ROS2.git
+      version: main
+    status: developed
   homing_local_planner:
     doc:
       type: git


christophebedard:
[Holding for Humble sync](https://discourse.ros.org/t/preparing-for-humble-sync-2025-05-16/43770)

marcosvcr:
What does happen now? What are the next steps to add my repo to the indexer?

marcosvcr:
Why is taking so long to merge my repo? @christophebedard

christophebedard (changes requested):
This process takes a bit of time, especially for new packages, which need to be carefully reviewed.

----

New package checklist:

  - [x] Top level license file: https://github.com/marcosvcr/hokuyo_node2/blob/main/LICENSE
    - https://github.com/marcosvcr/hokuyo_node2/blob/5b7fa54b4ba603008000de6d90456f840d2e2d90/hokuyo_node2/package.xml#L8
- [x] Public source repo: https://github.com/marcosvcr/hokuyo_node2
    - I'd recommend reading REP-144 (linked just above).
    - Your package is named `hokuyo_node2`, which I believe is meant to be a successor to https://github.com/ros-drivers/hokuyo_node. I think this is OK, but see my comments below.

Other comments:

* Your package will _not_ build on the build farm if it requires the code in this directory here: https://github.com/marcosvcr/hokuyo-for-ROS2/tree/main/urg_library. That library needs to be a ROS package, and then your package needs to depend that package in its `package.xml`. It looks like the same library exists here under the name `urg_c`: https://github.com/ros-drivers/urg_c/tree/ros2-devel. You should try using that instead. You can depend on it like `urg_node` depends on it here: https://github.com/ros-drivers/urg_node/blob/ros2-devel/package.xml#L31. Then you can try having your package and the `urg_c` package in your workspace and build them. If that works, then it should build on the build farm.
* You said you created this package because the hokuyo_node (which I assume is https://github.com/ros-drivers/hokuyo_node) and urg_node2 (which I assume is https://github.com/Hokuyo-aut/urg_node2) libraries didn't work. Have you tried `urg_node` https://github.com/ros-drivers/urg_node? Often, spending a bit more time trying to get something to work is better than creating a whole new package :smile: 

Overall, we cannot accept your package as-is because it will not build on the build farm. I also recommend considering trying out/using `urg_node`.

marcosvcr:
Hi I have made the changes as you requested. Unfortunately the urg_node2 didn't work in my situation when, connect the hokuyo via USB. I hope this time everything is fine. Fingers crossed @christophebedard

christophebedard:
I don't think your package will build on the build farm, because you're including `urg_c` headers (https://github.com/marcosvcr/hokuyo_node2/blob/8fc76331f0c90446ade805f1cbc0fdaf5686ca27/hokuyo_node2/src/urg_node.cpp#L3-L4) but you aren't declaring a dependency on it in your `package.xml` (https://github.com/marcosvcr/hokuyo_node2/blob/8fc76331f0c90446ade805f1cbc0fdaf5686ca27/hokuyo_node2/package.xml). Adding `<depend>urg_c</depend>` to your `package.xml` should do the trick.

marcosvcr:
I added the dependency and worked fine on my computer. It should build on the build farm now. @christophebedard

christophebedard (approved):
Thanks for confirming. Looks good now!

marcosvcr:
I have added the rclcpp and sensor_msgs on the pachage.xml

PR #45808, opened 2025-05-16
Revert "jsk_3rdparty: 2.1.31-3 in 'noetic/distribution.yaml' [bloom] …

sloretz:
…(#45761)"

This reverts commit 74f44d9a1dda86c52411092f01e15b80f340f8cc.

This appears to have caused a regression in jsk_3rdparty on amd64 https://repositories.ros.org/status_page/ros_noetic_default.html?q=REGRESSION

The root cause is voicevox fails to build https://build.ros.org/job/Nbin_uF64__voicevox__ubuntu_focal_amd64__binary/

This would remove `emotion_analyzer`, which would be unfortunate. @k-okada Think you can fix the voicevox regression?

### DIFF
--- a/noetic/distribution.yaml
+++ b/noetic/distribution.yaml
@@ -4738,7 +4738,6 @@ repositories:
       - collada_urdf_jsk_patch
       - dialogflow_task_executive
       - downward
-      - emotion_analyzer
       - ff
       - ffha
       - gdrive_ros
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

k-okada:
@sloretz @mjcarroll thanks. 
Should be fixe by following patch, releasing new version  https://github.com/ros/rosdistro/pull/45825
https://github.com/tork-a/jsk_3rdparty-release/commit/af66c2897e1b5a3f16755ee1426ba8683694974a
https://github.com/tork-a/jsk_3rdparty-release/commit/3a41b6ac888e23bf4ed26f3c63342daff6310f18

PR #45834, opened 2025-05-19
imu_tools: Switch to dedicated branch "kilted"

mintar:


christophebedard:
Thanks for also updating the `devel_branch` in the release repo!

PR #45855, opened 2025-05-20
trimble_driver: index into jazzy

andre-nguyen:

## Package name:

trimble_driver

## Package Upstream Source:

https://github.com/trimble-oss/trimble_driver_ros.git

## Purpose of using this:

Trimble customers can use this package to read data from various GNSS Receivers and INS products.

Distro packaging links:
 N/A

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -9754,6 +9754,12 @@ repositories:
       url: https://github.com/ros-drivers/transport_drivers.git
       version: main
     status: developed
+  trimble_driver:
+    source:
+      type: git
+      url: https://github.com/trimble-oss/trimble_driver_ros.git
+      version: jazzy
+    status: developed
   tsid:
     doc:
       type: git


christophebedard:
New package checklist:

  - [x] Top level license file: https://github.com/trimble-oss/trimble_driver_ros/blob/jazzy/LICENSE
- [x] License is [OSI-approved](https://opensource.org/licenses): BSD-2
    - `gsof_msgs`: https://github.com/trimble-oss/trimble_driver_ros/blob/3db9a88276bfe32fbfab55345597906c520d07ef/gsof_msgs/package.xml#L10
    - `trimble_driver`: https://github.com/trimble-oss/trimble_driver_ros/blob/3db9a88276bfe32fbfab55345597906c520d07ef/trimble_driver/package.xml#L13
    - `trimble_interfaces`: https://github.com/trimble-oss/trimble_driver_ros/blob/3db9a88276bfe32fbfab55345597906c520d07ef/trimble_interfaces/package.xml#L9
- [x] Public source repo: https://github.com/trimble-oss/trimble_driver_ros
    - I think `gsof_msgs` is unique enough, but I would recommend prefixing it with `trimble_` so that its contents are a bit clearer, especially since Googling "gsof" alone doesn't return much. What do you think?

christophebedard:
[Holding for Jazzy sync](https://discourse.ros.org/t/preparing-for-jazzy-sync-2025-05-23/)

andre-nguyen:
>   * I think `gsof_msgs` is unique enough, but I would recommend prefixing it with `trimble_` so that its contents are a bit clearer, especially since Googling "gsof" alone doesn't return much. What do you think?

Good point, changed it to `trimble_gsof_msgs`. This [documentation page](https://receiverhelp.trimble.com/alloy-gnss/en-us/GSOFmessages_Overview.html) should come up when searching for trimble gsof.

>   * I think maybe `trimble_driver` should be `trimble_gsof_driver` if it really only deals with Trimble's GSOF, but you know better, so I leave that part up to you.

This one is harder to answer, there is a different protocol for some Trimble Applanix products that used to be part of the package but was removed because we couldn't get clearance (yet). There is still some will to get that out public but it has not happened yet.

christophebedard:
Alright, sounds good, thanks for renaming `gsof_msgs` to `trimble_gsof_msgs`.

Now this only needs to wait for the Jazzy freeze to be lifted.

christophebedard (approved):

PR #45864, opened 2025-05-20
rviz_satellite: 4.2.0-1 in 'rolling/distribution.yaml' [bloom]

Timple:
Increasing version of package(s) in repository `rviz_satellite` to `4.2.0-1`:

- upstream repository: https://github.com/nobleo/rviz_satellite.git
- release repository: https://github.com/nobleo/rviz_satellite-release.git
- bloom version: `0.12.0`

## rviz_satellite

```
* Fixed demo (install rviz file)
* Add libproj-dev as dependency at package.xml
* Add support for sub-region tile servers (local maps)
* fix: time conversion constant in publish_demo_data
* Contributors: Ahmad Kamal Nasir, Christian Geller, David Alejo Teissière, Tim Clephas, nmitropoulos
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -7847,6 +7847,21 @@ repositories:
       url: https://github.com/teamspatzenhirn/rviz_2d_overlay_plugins.git
       version: main
     status: maintained
+  rviz_satellite:
+    doc:
+      type: git
+      url: https://github.com/nobleo/rviz_satellite.git
+      version: main
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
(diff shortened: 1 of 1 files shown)


christophebedard (approved):
This package already exists in other distros, so all good.

I'll note that we strongly prefer having the release repo under the ros2-gbp org, especially for Rolling, since it allows us to automatically release your package into new distros. See [this guide](https://docs.ros.org/en/rolling/How-To-Guides/Releasing/Release-Team-Repository.html) and the note at the very bottom. Since that note says this is not currently a hard requirement, and since you're already using the same release repo in your org for other distros, I think we can proceed with this PR, but this is something to consider in the future.

PR #45913, opened 2025-05-21
Add python3-posix-ipc-pip rule

baconbrot:
## Package name:

python3-posix-ipc-pip

## Package Upstream Source:

[Github](https://github.com/osvenskan/posix_ipc)

## Purpose of using this:
From the repo: "creation and manipulation of POSIX inter-process semaphores, shared memory and message queues"

pip: [pypi.org](https://pypi.org/project/posix-ipc/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -8228,6 +8228,10 @@ python3-polling2-pip:
   ubuntu:
     pip:
       packages: [polling2]
+python3-posix-ipc-pip:
+  '*':
+    pip:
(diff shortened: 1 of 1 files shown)


marcoag:
This package seems to be available for [Jammy, Focal](https://launchpad.net/ubuntu/+source/python-posix-ipc), Fedora and RHEL 9. Can we add system packages and only use pip for those OS that don't have it at the sytem level?

mjcarroll (approved):

PR #45954, opened 2025-05-22
Add 'taskflow' rosdep rule

miguelprada:

## Package name:

`taskflow`

## Package Upstream Source:

[Source repository](https://github.com/taskflow/taskflow).

## Purpose of using this:

This contains a header-only C++ library for parallel programming. It can be used to compose computational graphs declaring precedence dependencies between nodes and then executing them in parallel. More info in the [taskflow project's main page](https://taskflow.github.io/).

:warning:  Note that this was briefly [packaged and released by `ros-industrial` into ROS Noetic](https://github.com/ros-industrial-release/taskflow-release). I'm unsure of the implications of this (e.g. whether a different rosdep key should be used to avoid confusion or conflicts).

  - https://packages.debian.org/trixie/libtaskflow-cpp-dev
  - https://packages.ubuntu.com/noble/libtaskflow-cpp-dev
  - https://packages.fedoraproject.org/pkgs/taskflow/taskflow-devel/
  - Not available ([only in AUR](https://aur.archlinux.org/packages/cpp-taskflow)).
  - https://packages.gentoo.org/packages/dev-cpp/taskflow
  - https://formulae.brew.sh/formula/taskflow#default
  - Not available
  - https://search.nixos.org/packages?channel=24.11&show=taskflow&from=0&size=50&sort=relevance&type=packages&query=taskflow
  - Not available
  - Not available

Also:
- FreeBSD
  - https://ports.freebsd.org/cgi/ports.cgi?query=taskflow&stype=all&sektion=all
- OpenEmbedded
  - https://layers.openembedded.org/layerindex/recipe/419692/

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6523,6 +6523,23 @@ libsystemd-dev:
   openembedded: [systemd@openembedded-core]
   ubuntu:
     '*': [libsystemd-dev]
+libtaskflow-cpp-dev:
+  debian:
+    '*': [libtaskflow-cpp-dev]
+    bookworm: null
+    bullseye: null
+  fedora: [taskflow-devel]
+  freebsd: [taskflow]
+  gentoo: [dev-cpp/taskflow]
+  nixos: [taskflow]
+  openembedded: [taskflow@meta-ros1-noetic]
+  osx:
+    homebrew:
+      packages: [taskflow]
+  ubuntu:
+    '*': [libtaskflow-cpp-dev]
+    focal: null
+    jammy: null
 libtclap-dev:
   arch: [tclap]
   debian: [libtclap-dev]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Please change the Debian and Ubuntu rules to be forward compatible. 

Like this

```
debian:
  '*': [libtaskflow-cpp-dev]
  bookworm: null
  bullseye: null
```

So that it won't have to be updated each new release.

miguelprada:
> Please change the Debian and Ubuntu rules to be forward compatible.

I hope 57479bb covers all Debian and Ubuntu versions which should be excluded from the general wildcard evaluation.

tfoote (approved):
Thanks for the update, covering the currently supported lts releases should be enough.

mjcarroll (approved):

miguelprada:
Thanks @tfoote and @mjcarroll!

PR #46007, opened 2025-05-23
Set plotjuggler packages to last building version.

marcoag:
Rolling back `plotjuggler` and `plotjuggler_ros` to last building versions.

It seems that [latest changes related to 3rd party dependencies](https://github.com/facontidavide/PlotJuggler/commit/63ced5aa2b20eacd5f90f93ea6f093cbe1dc9e31) don't go well with the buildfarm debian packaging. Rolling back to avoid regression in the [upcoming sync](https://discourse.ros.org/t/preparing-for-jazzy-sync-2025-05-23).

FYI: @facontidavide

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -5854,7 +5854,7 @@ repositories:
       tags:
         release: release/jazzy/{package}/{version}
       url: https://github.com/ros2-gbp/plotjuggler-release.git
-      version: 3.10.4-1
+      version: 3.9.2-1
     source:
       type: git
       url: https://github.com/facontidavide/PlotJuggler.git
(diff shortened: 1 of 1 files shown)


marcoag:
Thanks for understanding @facontidavide.

It seems like the `CPMAddPackage` way to pull dependencies won't be compatible with the buildfarm build package process. I believe that dependencies need to be declared at the `package.xml` level to be found when building the debian package. 

I would say the ideal way would be to use as many system packages as possible, I.e. I saw there a rosdep [backward-ccp key](https://github.com/ros/rosdistro/blob/40e9a19365064a7791d77cd5f1aaf6aba6104684/rosdep/base.yaml#L2748C1-L2748C20) you could add to your package-xml if you can use the system's version and then vendor whatever is not provided by the OS. Alternatively, maybe less clean but easier, you could move the ones not in the system back to the 3rd party directory on the repo, as it was previously done.

The [ros_buildfarm]() package provides a way to run the jobs locally: https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/jobs/release_jobs.rst#run-the-release-job-locally. It takes some setup but I can confirm it can be done :grin:.

PR #46048, opened 2025-05-24
Update ROS keys to new ones used in ros-apt-source

claraberendsen:
### Description
As the title says this PR updates the key to the new ones (with new expiration date). 
Key values are pulled from: 
- ros.key: https://github.com/ros-infrastructure/ros-apt-source/blob/main/ros-apt-source/keys/ros2-archive-keyring.gpg
- ros.asc: https://github.com/ros-infrastructure/ros-apt-source/blob/main/ros2-release/RPM-GPG-KEY-ROS2


tfoote (approved):
Validated as matching

claraberendsen:
@tfoote it seems I don't have merge powers here. If yo could merge this one would be much appreciated

PR #46052, opened 2025-05-24
Update mp2p_icp dev branch name

jlblancoc:
This is an update of the default branch name changed from the old "master" to the new "develop", for consistency with the rest of the org projects.


jlblancoc:
Should I undo the change for noetic since it's EOL??

marcoag (approved):

PR #46139, opened 2025-05-27
Add MuJoCo pip rosdistro entry

saikishor:
Add `mujoco` and `mujoco-mjx` to the rosdistro to make it installable with rosdep

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7473,22 +7473,6 @@ python3-mss-pip:
   '*':
     pip:
       packages: [mss]
-python3-mujoco-pip:
-  arch:
-    pip:
-      packages: [mujoco]
-  debian:
-    pip:
-      packages: [mujoco]
-  fedora:
-    pip:
-      packages: [mujoco]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
There are mujoco packages already natively on some of our support platforms, including our primary one Ubuntu

https://packages.ubuntu.com/source/noble/mujoco

Also Debian as of bookworm: https://packages.debian.org/source/bookworm/mujoco

As such our policy is to use the native package

saikishor:
> There are mujoco packages already natively on some of our support platforms, including our primary one Ubuntu
> 
> https://packages.ubuntu.com/source/noble/mujoco
> 
> Also Debian as of bookworm: https://packages.debian.org/source/bookworm/mujoco
> 
> As such our policy is to use the native package

Thank you for the review. The problem here is 2.2.2 is the relatively old version. Right now, we have 3.3.2 upstream, that's the reason for this PR. If the policy is to use a native package, I understand.

Thank you

PR #46201, opened 2025-05-30
adding gersemi

DangitBen:

## Package name:

`python3-gersemi-pip`

## Package Upstream Source:

https://github.com/BlankSpruce/gersemi
https://pypi.org/project/gersemi/

## Purpose of using this:

`gersemi` is a CMake linter and formatter, as opposed to `cmake-lint` which only reports problems but cannot automatically fix them.

 Adding `gersemi` to rosdep for use in ament formatting tools.

All following are the same, since it is only available via pip.

- Debian: https://pypi.org/project/gersemi/
- Ubuntu: https://pypi.org/project/gersemi/
- Fedora: https://pypi.org/project/gersemi/
- Arch: https://pypi.org/project/gersemi/
- Gentoo: https://pypi.org/project/gersemi/
- macOS: https://pypi.org/project/gersemi/
- Alpine: https://pypi.org/project/gersemi/
- NixOS/nixpkgs: https://pypi.org/project/gersemi/
- openSUSE: https://pypi.org/project/gersemi/
- rhel: https://pypi.org/project/gersemi/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6434,6 +6434,10 @@ python3-geopy:
   fedora: [python3-geopy]
   nixos: [python3Packages.geopy]
   ubuntu: [python3-geopy]
+python3-gersemi:
+  '*':
+    pip:
+      packages: [gersemi]
 python3-gi:
   arch: [python-gobject]
   debian: [python3-gi]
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

sloretz (approved):

PR #46387, opened 2025-06-06
lms1xx: 1.0.1-1 in 'jazzy/distribution.yaml' [bloom]

tonybaltovski:
Increasing version of package(s) in repository `lms1xx` to `1.0.1-1`:

- upstream repository: https://github.com/clearpathrobotics/LMS1xx.git
- release repository: https://github.com/clearpath-gbp/LMS1xx-release.git
- bloom version: `0.12.0`

## lms1xx

```
* Added statemachine to prevent driver from blocking executor.
* Contributors: Roni Kreinin
```

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -4174,6 +4174,21 @@ repositories:
       url: https://github.com/fkie/live555_vendor.git
       version: main
     status: maintained
+  lms1xx:
+    doc:
+      type: git
+      url: https://github.com/clearpathrobotics/LMS1xx.git
+      version: humble-devel
+    release:
+      tags:
+        release: release/jazzy/{package}/{version}
+      url: https://github.com/clearpath-gbp/LMS1xx-release.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/LMS1xx.git
+      version: humble-devel
+    status: maintained
   log_view:
     doc:
       type: git


jazzy/distribution.yaml
```diff
@@ -4174,6 +4174,21 @@ repositories:
...
+      version: humble-devel
+    release:
+      tags:
+        release: release/jazzy/{package}/{version}
+      url: https://github.com/clearpath-gbp/LMS1xx-release.git
+      version: 1.0.1-1
+    source:
+      type: git
+      url: https://github.com/clearpathrobotics/LMS1xx.git
+      version: humble-devel
```
Yadunund, line 4190: Just to confirm, the source branch for `jazzy` is `humble-devel`?

Yadunund (dismissed):
  - [ ] Top level license file:
- [ ] License is [OSI-approved](https://opensource.org/licenses):
- [x] Public source repo:

@tonybaltovski thanks for the PR. There are a few changes needed before we can accept this package.

1. A LICENSE file is missing in the source repo
2.  The `package.xml` lists as the license as `LGPL` but this does not follow SPDX convention (eg. `LGPL-3.0-only `).
3. The package does not follow REP-144 conventions. 
Specifically
```
if a package is a driver, append driver
```
I would recommend the name `sick_lms1xx_driver` or `clearpath_lms1xx_driver` to clearly distinguish the fact that this driver is not maintained by SICKAG like [these ones](https://github.com/ros/rosdistro/blob/b8ccabe4562edf3b3f3d6be9d3b968a01b5ccaef/jazzy/distribution.yaml#L8997-L9091)

jazzy/distribution.yaml
tonybaltovski: That is correct.

ros-discourse:

https://discourse.ros.org/t/ros-pmc-minutes-for-june-10-2025/44219/1

mjcarroll:
Discussed in PMC, due to history of this package, we can consider it grandfathered in.

PR #46392, opened 2025-06-07
This PR adds a static snapshot of `rosdep` entries for the now end-of-life (EOL) Ubuntu Xenial distribution. 

nilesh646:

## Package name:

TODO

## Package Upstream Source:

## Purpose of using this:

Distro packaging links:

  - REQUIRED
  - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

ROSDISTRO NAME

# The source is here:

http://sourcecode_url

# Checks

### DIFF
--- a/rosdep/rosdep-xenial.yaml
+++ b/rosdep/rosdep-xenial.yaml
@@ -1,0 +1,5411 @@
+ace:
+  ubuntu:
+  - libace-dev
+ack:
+  ubuntu:
+    '*':
+    - ack
+ack-grep:
+  ubuntu:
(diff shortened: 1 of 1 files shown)


tfoote:
Note that if you want an older version of the dependencies you can change your rosdep sources to historical tags in the git repository. We don't need to maintain it in the current head for it to be accessible.

PR #46440, opened 2025-06-11
Update distribution.yaml

ksatyaki:
Update version of package MoD to 1.1.0-4 from 1.1.0-1

Minor update fixed OMPL Includes not being found. Fixes [this issue](https://github.com/OrebroUniversity/mod/issues/1)

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5192,7 +5192,7 @@ repositories:
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/OrebroUniversity/mod-release.git
-      version: 1.1.0-4
+      version: 1.1.1-1
     source:
       test_pull_requests: true
       type: git
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -5192,7 +5192,7 @@ repositories:
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/OrebroUniversity/mod-release.git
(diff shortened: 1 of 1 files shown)


ksatyaki:
Ah. I thought the 4 from 1 in the 4th part of version should be enough. Let me re-release.

christophebedard:
`catkin_prepare_release` should bump the patch number (`major.minor.patch`) by default. Are you running these commands on the latest commit on your `master` branch? https://bitbucket.org/mapsofdynamics/mod/commits/branch/master. I don't see a commit with the message "1.1.1" on `master` and I don't see a `1.1.1` tag either. Bloom looks at the latest tag, which is currently `1.1.0`, so it tries to (re)release that.

ksatyaki:
I think I have now pushed those. I would also like to change the upstream and release repos to the github.com/OrebroUniversity ones. It had not seen the 1.1.1 tag since it was looking at the wrong upstream repo, I think. Should be fixed now. :-)

christophebedard (approved):
Yeah, this looks good now!

Your release repo correctly points to your Bitbucket repo: https://github.com/OrebroUniversity/mod-release/blob/d410c0f884eb72a907c70f30ba7d23f61406b435/tracks.yaml#L27. However, yes, it might be simpler to point everything (release repo and actual source code repo) to GitHub.

ksatyaki:
Thanks !

PR #46443, opened 2025-06-12
rsl: 1.2.0-3 in 'humble/distribution.yaml' [bloom]

nbbrooks:
Increasing version of package(s) in repository `rsl` to `1.2.0-3`:

- upstream repository: https://github.com/PickNikRobotics/RSL.git
- release repository: https://github.com/ros2-gbp/RSL-release.git
- bloom version: `0.12.0`

## rsl

```
* Remove devcontainer
* Fix new clang-tidy warnings
* Update Catch2
* Ensure FindEigen3.cmake is not used (#126 <https://github.com/PickNikRobotics/RSL/issues/126>)
* Eigen: Add NO_MODULE to find_package (#125 <https://github.com/PickNikRobotics/RSL/issues/125>)
  Work around RHEL 8 failure due to a faulty FindEigen3.cmake module being found
* Hide symbols from shared libraries by default and support DLL creation
* Fix static_string.cpp compilation with MSVC standard library
  std::array's iterator is a pointer in libc++ and libstdc++ so
  const auto* works fine. However the array iterator is not a
  pointer on MSVC so compilation fails. Removing the asterisk still
  results in a pointer type being used but causes a clang-tidy check
  to fail so we have to disable that.
* Disable TRY macro tests on MSVC
* Disable compiler warnings by default
  Unconditionally enabling -Werror is a heavy-handed approach and not
  ideal when shipping code to be used by many third parties. In fact
  it's even better to disable warnings by default since compiler
  warnings are not hard requirements. They're merely a development
  tool that developers should opt into.
* Contributors: Chris Thrasher, Christoph Fröhlich, Griswald Brooks, mosfet80
```

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -9743,7 +9743,7 @@ repositories:
       tags:
         release: release/humble/{package}/{version}
       url: https://github.com/ros2-gbp/RSL-release.git
-      version: 1.1.0-1
+      version: 1.2.0-3
     source:
       type: git
       url: https://github.com/PickNikRobotics/RSL.git


nbbrooks:
This release should fix a CMake issue involving RHEL. I am guessing the build farm RHEL jobs were manually disabled in the past - could we get them turned on again?

christophebedard:
[Holding for Humble sync](https://discourse.ros.org/t/preparing-for-humble-sync-2025-06-13/44216)

christophebedard:
> This release should fix a CMake issue involving RHEL. I am guessing the build farm RHEL jobs were manually disabled in the past - could we get them turned on again?

I do see that it's disabled here: https://repo.ros2.org/status_page/ros_humble_rhel.html?q=rsl. Looks like it can be re-enabled here: https://build.ros2.org/job/Hbin_rhel_el864__rsl__rhel_8_x86_64__binary/.

@cottsay do some packages' Hbin_rhel jobs get disabled (manually or automatically) if they don't successfully build for a period of time? Can I re-enable it from the Jenkins UI if the maintainer says that the package should now build on RHEL?

christophebedard:
I manually enabled the src & bin jobs on June 16th, and both jobs passed:

1. https://build.ros2.org/job/Hsrc_el8__rsl__rhel_8__source/
2. https://build.ros2.org/job/Hbin_rhel_el864__rsl__rhel_8_x86_64__binary/

However, they then got automatically disabled by the nightly reconfigure job: https://build.ros2.org/job/Hrel_rhel_reconfigure-jobs/2447/consoleFull#console-section-6. It seems like they're blacklisted somewhere:

> The following packages are being disabled because of white-/blacklisting:
> ...
>   - rsl

I was trying to find a manually-curated list, such as [this one](https://github.com/ros-infrastructure/ros_buildfarm_config/blob/c1bef62b897e100f8d2d152ad8a3bad2cc73bd08/galactic/release-rhel-build.yaml#L17), but I haven't found anything yet. And I don't think it's disabled because an upstream project is disabled, because none of its upstream projects (except src) are disabled.

PR #46633, opened 2025-06-25
rosdep: nixos: Add python3-filelock package

hacker1024:
This add the NixOS entry for `python3-filelock`.

https://nixpkgs.dev/python313Packages.filelock

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5965,6 +5965,7 @@ python3-filelock:
   arch: [python-filelock]
   debian: [python3-filelock]
   fedora: [python3-filelock]
+  nixos: [python3Packages.filelock]
   opensuse: [python3-filelock]
   rhel: [python3-filelock]
   ubuntu: [python3-filelock]


rosdep/python.yaml
```diff
@@ -5965,6 +5965,7 @@ python3-filelock:
   arch: [python-filelock]
   debian: [python3-filelock]
   fedora: [python3-filelock]
+  nixos: [python3Packages.filelock]
```
tfoote, line 5968: The link you provide is for python313Packages.filelock
```suggestion
  nixos: [python313Packages.filelock]
```

Is this a valid alias or does it need to be more specific like this?

hacker1024: It is a valid alias, and refers to the main Python version being used with the ROS distro as a whole. search.nixos.org does not index these for some reason.

(Also note that all other Python packages so far use the same prefix)

wentasah (approved):
LGTM.

sloretz (approved):

bjsowa (approved):

nuclearsandwich (approved):

PR #46764, opened 2025-07-03
Add ifm3dpy to rosdep database

Finn2708:

## Package name:

ifm3dpy

## Package Upstream Source:

https://github.com/ifm/ifm3d

## Purpose of using this:

This dependency is being used for interfacing with ifm pmd-based 3D ToF Perception devices. As these devices are used in many mobile robots, it is valuable in the rosdep database.

There are no distro-specific packages I am aware of. But it should be available with pip everywhere.

  - REQUIRED
  - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6721,6 +6721,10 @@ python3-ifcfg:
   ubuntu:
     bionic: [python3-ifcfg]
     focal: [python3-ifcfg]
+python3-ifm3dpy-pip:
+  '*':
+    pip:
+      packages: [ifm3dpy]
 python3-ignition-gazebo6:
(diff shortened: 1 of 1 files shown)


Finn2708:
Hi,

do I need to request a review somewhere? 
What is needed to move this forward?

christophebedard (approved):

Note that we require 2 approvals for rosdep keys.

ahcorde (approved):

PR #46833, opened 2025-07-09
Add `reduct-py` pip index

AnthonyCvn:

## Package name:

- python3-reduct-py-pip
- python3-mcap-pip

## Package Upstream Source:

- https://github.com/reductstore/reduct-py
- https://github.com/foxglove/mcap

## Purpose of using this:

These dependencies are used in the `reductstore_agent` ROS 2 package for writing and managing MCAP files (`mcap`) and uploading them to ReductStore (`reduct-py`). These libraries are not available via system package managers and must be installed via pip.

This is required to support ROS 2 logging/recording workflows with ReductStore as a backend.

These packages are only available on PyPI and must be installed via pip. Not available in system package managers.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7260,6 +7260,25 @@ python3-mavproxy-pip:
   ubuntu:
     pip:
       packages: [MAVProxy]
+python3-mcap-pip:
+  debian:
+    pip:
+      packages: [mcap]
+  fedora:
+    pip:
+      packages: [mcap]
+  gentoo:
+    pip:
+      packages: [mcap]
+  nixos:
+    pip:
+      packages: [mcap]
+  osx:
+    pip:
+      packages: [mcap]
+  ubuntu:
+    pip:
+      packages: [mcap]
 python3-mccabe:
   arch: [python-mccabe]
   debian: [python3-mccabe]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -7260,6 +7260,19 @@ python3-mavproxy-pip:
   ubuntu:
     pip:
       packages: [MAVProxy]
+python3-mcap-pip:
```
tfoote: I suspect that this is going to collide with our mcap vendor packages. 

https://index.ros.org/p/mcap_vendor/

Until this is available on all platforms natively we can't switch to a rosdep key.

tfoote (changes requested):

tfoote (approved):
Thanks for using mcap which is already available. This lgtm now

rosdep/python.yaml
```diff
@@ -9411,6 +9411,19 @@ python3-rdflib:
   rhel:
     '8': ['python%{python3_pkgversion}-rdflib']
   ubuntu: [python3-rdflib]
+python3-reduct-py-pip:
+  debian:
```
asymingt: Just a drive-by comment here, mainly for @nuclearsandwich . Is there a reason we didn't do the following wildcard expansion?

```yaml
python3-reduct-py-pip:
  '*':
    pip:
      packages: [reduct-py]
```

nuclearsandwich: OS level wildcards were only added recently (well, 2022 apparently but in ROS years that's still puppy age) https://github.com/ros-infrastructure/rosdep/blob/master/CHANGELOG.rst#0220-2022-06-24 and it doesn't always occur to suggest them as a result. It's a good change to suggest here.

rosdep/python.yaml
```diff
@@ -9411,6 +9411,19 @@ python3-rdflib:
...
+      packages: [reduct-py]
+  fedora:
+    pip:
+      packages: [reduct-py]
+  osx:
+    pip:
+      packages: [reduct-py]
+  ubuntu:
+    pip:
+      packages: [reduct-py]
```
nuclearsandwich: Thanks to the OS wildcards in rosdep 0.22.0 this can be simplified.

```suggestion
python3-reduct-py-pip:
  '*':
    pip:
      packages: [reduct-py]
```

nuclearsandwich (dismissed):
One suggested cleanup and otherwise this looks ready!

christophebedard (approved):
Standard pip disclaimer: ROS packages that depend on `pip` keys cannot be released into a ROS distribution.
They can only be depended on by from-source builds.
Because of this, system packages are highly preferred to pip packages.

PR #46843, opened 2025-07-09
add jrl_cmakemodules on rolling

nim65s:

jrl_cmakemodules

This package is already used as a submodule in multiple others (eigenpy/coal/pinocchio/tsid), but it can now be installed as standalone and be depended on

# The source is here:

https://github.com/jrl-umi3218/jrl-cmakemodules

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -3115,6 +3115,12 @@ repositories:
       url: https://github.com/ros-drivers/joystick_drivers.git
       version: ros2
     status: maintained
+  jrl_cmakemodules:
+    source:
+      type: git
+      url: https://github.com/jrl-umi3218/jrl-cmakemodules.git
+      version: master
+    status: maintained
   kdl_parser:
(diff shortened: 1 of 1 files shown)


asymingt:
Hi @nim65s , this PR meets the criteria for a `A new binary, source, or documentation package in a ROS distribution.` per the [review guidelines](https://github.com/ros/rosdistro/blob/master/REVIEW_GUIDELINES.md). This means that it has to meet a few criteria, one of which is that it needs a LICENSE file in the repo. I see your does contain the file, but it only has a short statement about what license applies, but not the terms of the license. It looks like you are releasing under GPL-3. Here is a link to a LICENSE that you can use: https://www.gnu.org/licenses/gpl-3.0.txt

cc. @nuclearsandwich to fact-check me on this one.

quarkytale:
Friendly ping @nim65s to update the license file

emersonknapp (approved):

PR #46855, opened 2025-07-11
Add rowan pip

knmcguire:

## Package name:

* Rowan (pip package)
* 
## Package Upstream Source:

https://github.com/glotzerlab/rowan

## Purpose of using this:

This is used for essential packages in the [crazyswarm2 project](https://github.com/IMRCLab/crazyswarm2) (just recently released as the [crazyflie* packages](https://index.ros.org/p/crazyflie/))

Distro packaging links:

This package only exists in PyPi: https://pypi.org/project/rowan/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9623,11 +9623,8 @@ python3-rospkg-modules:
     pip:
       packages: [rospkg]
   ubuntu: [python3-rospkg-modules]
-python3-rowan:
-  debian:
-    pip:
-      packages: [rowan]
-  ubuntu:
+python3-rowan-pip:
+  '*':
     pip:
       packages: [rowan]
 python3-rsa:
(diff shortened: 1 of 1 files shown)


asymingt (approved):
Changes LGTM, as there is no pre-existing debian / ubuntu package for `python3-rowan`. However, we need a second review on this from somebody with a deeper understanding of dependencies and pip interaction to merge.

nuclearsandwich:
> This is used for essential packages in the [crazyswarm2 project](https://github.com/IMRCLab/crazyswarm2) (just recently released as the [crazyflie* packages](https://index.ros.org/p/crazyflie/))

@knmcguire pip packages can't be used as dependencies of build farm packages since bloom only supports resolving system packages during Debian builds.

If this is still useful for unreleased packages we can add it, but otherwise it may be spurious.

asymingt:
@nuclearsandwich  can you approve if you are happy with the PR as-is, and one of us can merge.

nuclearsandwich (approved):

PR #46859, opened 2025-07-11
Add mujoco

clarissagsai:

## Package name:

mujoco

## Package Upstream Source:

https://github.com/google-deepmind/mujoco

## Purpose of using this:

MuJoCo is a fast, accurate physics engine designed for research and development in robotics and biomechanics, offering fine-grained control over simulation and advanced features like soft contacts and inverse dynamics. Including it in the ROS distribution would streamline integration for roboticists, enabling high-fidelity simulation within the ROS ecosystem without complex external dependencies.

- Python:
  - https://pypi.org/project/mujoco/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7482,6 +7482,13 @@ python3-mss-pip:
   '*':
     pip:
       packages: [mss]
+python3-mujoco-pip:
+  debian:
+    pip:
+      packages: [mujoco]
+  ubuntu:
+    pip:
+      packages: [mujoco]
 python3-multimethod-pip:
   debian:
     pip:


rosdep/python.yaml
```diff
@@ -7473,6 +7473,13 @@ python3-mss-pip:
   '*':
     pip:
       packages: [mss]
+python3-mujoco-pip:
```
asymingt, line 7485: This looks like a valid pip package for now. However, we might want to notify the package manager because it seems like `libmojoco-dev` is available as a debian package, but `python3-mojoco` is not.  It might be the case that Python bindings will be released as a follow-on through the `python3-mujoco` package.

nuclearsandwich (approved):
The ideal would be to file a request to package (RTP) in the [Debian bug tracker](https://www.debian.org/devel/wnpp/) to add a binary package for the python bindings. But in the meantime since this dependency is unavailable and not actively packaged for future releases of Debian/Ubuntu. Adding a pip key is reasonable today.

rosdep/python.yaml
nuclearsandwich: The ideal would be to file a request to package (RTP) in the [Debian bug tracker](https://www.debian.org/devel/wnpp/) to add a binary package for the python bindings. But in the meantime since this dependency is unavailable and not actively packaged for future releases of Debian/Ubuntu. Adding a pip key is reasonable today.

clarissagsai: Thank you :pray:

PR #46901, opened 2025-07-14
moveit_msgs: 2.7.0-1 in 'rolling/distribution.yaml'

nbbrooks:
This was manually generated due to GH classic token limitations.

Increasing version of package(s) in repository moveit_msgs to 2.7.0-1:

*  upstream repository: https://github.com/moveit/moveit_msgs.git
*  release repository: https://github.com/ros2-gbp/moveit_msgs-release.git
*  distro file: `rolling/distribution.yaml`
*  bloom version: `0.13.0`
*  previous version for package: `2.6.0-1`

## moveit_msgs

```
* Add GetMultiStateValidity service (#186 <https://github.com/ros-planning/moveit_msgs/issues/186>)
* CI: Update actions
* Contributors: Lucian James, Robert Haschke
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4168,7 +4168,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/moveit_msgs-release.git
-      version: 2.6.0-1
+      version: 2.7.0-1
     source:
       type: git
       url: https://github.com/ros-planning/moveit_msgs.git


Yadunund (approved):

Yadunund:
Hmm interesting, the `Validate rosdistro` check is failing due to `RuntimeError: rolling: missing package.xml file for package "moveit_msgs"`.

nbbrooks:
I will give that a shot. Perhaps what was involved was I attempted to squash the commits generated by `catkin_generate_changelog` and `catkin_prepare_release` so I didn't end up with this awkward git history like with the last moveit release.

image

I did not realize there were tags involved so perhaps squashing them caused this.

I will give this a shot tomorrow with a 2.7.1 release with no squashing.

Yadunund:
@nbbrooks i am going to go ahead with the rolling sync. We can resolve the issue with blooming moveit_msgs in the meanwhile. I wonder if we should try setting up the package for first time release.

urfeex:
@nbbrooks the `release/rolling/moveit_msgs` seems not to be the problem, that does [exist](https://github.com/ros2-gbp/moveit_msgs-release/tree/release/rolling/moveit_msgs). The problematic ref is `337b97c` which is taken from https://github.com/ros2-gbp/moveit_msgs-release/blob/6134a9a343f9c7690441abf8828c0e2faad66a8a/patches.conf.

The [`upstream`](https://github.com/ros2-gbp/moveit_msgs-release/tree/upstream) branch is still pointing to 2.6.0, so that's probably where your bloom-release got wrong somewhere. The patch is referring to a commit that is missing, hence it is failing from my understanding.

Are there any updates from your side on this -- also concerning #47018 ?

urfeex:
To come back to this I experimented a bit today. From my understanding, the token is only needed for the final PR creating part.

Generating and pushing to the release repo should not require any token at all from my understanding. Having https->ssh redirection and an ssh key available as [explained in the ROS wiki](https://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization#Re-route_https_through_ssh) should be enough to run `bloom-release <release_entry> --rosdistro rolling --no-pull-request`.

The token is then required to clone the rosdistro repository, push to the repo's fork and opening the PR. The first two items can be achieved without the token by changing the lines

https://github.com/ros-infrastructure/bloom/blob/6199b2a76f50f120a398c48c44b1e1ffd6c5f38f/bloom/commands/release.py#L741-L742

to

```python
rosdistro_url = 'https://github.com/{base_repo_id}.git'.format(**locals())
fork_template = 'https://github.com/{head_org}/{head_repo}.git'
```

then, running `bloom-release <release_entry> --rosdistro rolling --pull-request-only` will generate the PR content and attempt to open the PR, which isn't possible without a proper token. But the PR can conveniently be created by copying the PR body from the output and opening the PR by clicking on the PR link in the output. This only works if the rosdistro fork already exists for your account as creating the fork would require a token again.

Note: All of the above was done having a fine-grained token with public repo access configured.

nbbrooks:
> @nbbrooks the `release/rolling/moveit_msgs` seems not to be the problem, that does [exist](https://github.com/ros2-gbp/moveit_msgs-release/tree/release/rolling/moveit_msgs). The problematic ref is `337b97c` which is taken from https://github.com/ros2-gbp/moveit_msgs-release/blob/6134a9a343f9c7690441abf8828c0e2faad66a8a/patches.conf.

@urfeex thank you for looking into this. That makes sense to me. Not sure where the 337b97c ref came from but probably user error. I am still learning more about this and if it would be simplest to just manually fix all the generated patch.conf files this one time.

Regarding https://github.com/ros/rosdistro/pull/47018 there is no update yet.

@v4hn will be helping out with this sometime this week on resolving these issues..

emersonknapp (approved):

PR #46945, opened 2025-07-17
Add pip version of typing-extensions package

bastianschildknecht:

## Package name:

typing-extensions (pip package)

## Package Upstream Source:

[https://github.com/python/typing_extensions](https://github.com/python/typing_extensions)

## Purpose of using this:

typing-extensions is a Python module that serves two related purposes:
- Enable use of new type system features on older Python versions. For example, typing.TypeGuard is new in Python 3.10, but typing_extensions allows users on previous Python versions to use it too.
- Enable experimentation with new type system PEPs before they are accepted and added to the typing module.

Note: A rosdep package already exists for the system package `python3-typing-extensions`; however, this PR adds the pip version, as the system package often contains outdated versions of the package, which prevents modern type annotations from being available.

Distro packaging links:

- Python:
  - https://pypi.org/project/typing-extensions/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -10676,6 +10676,10 @@ python3-typing-extensions:
     bionic:
       pip:
         packages: [typing-extensions]
+python3-typing-extensions-pip:
+  '*':
+    pip:
+      packages: [typing-extensions]
 python3-tz:
   alpine: [py3-tz]
   arch: [python-pytz]


nuclearsandwich:
> Note: A rosdep package already exists for the system package python3-typing-extensions; however, this PR adds the pip version, as the system package often contains outdated versions of the package, which prevents modern type annotations from being available.

pip keys in the official rosdep database are only added when no system dependency is available ([CONTRIBUTING.md](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip-1)).
Using system packages on supported distributions is the trade-off we make to have long term stability in ROS releases.

If this is something that you want to use for local development of packages off the official ROS build farm you can use add custom rosdep sources and define these keys there: https://github.com/ros-infrastructure/rosdep/blob/master/doc/sources_list.rst

Since this PR cannot be merged, I will go ahead and close it.

PR #46946, opened 2025-07-17
Add pip version of docker package

bastianschildknecht:

## Package name:

docker (pip package)

## Package Upstream Source:

[https://github.com/docker/docker-py](https://github.com/docker/docker-py)

## Purpose of using this:

docker-py is a Python library for the Docker Engine API. It lets you do anything the docker command does, but from within Python apps – run containers, manage containers, manage Swarms, etc.

Note: A rosdep package already exists for the system package python3-docker; however, this PR adds the pip version, as the system package often contains outdated versions of the package, which are missing important bug fixes and new features. 

Distro packaging links:

- Python:
  - [https://pypi.org/project/docker/](https://pypi.org/project/docker/)

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5719,6 +5719,10 @@ python3-docker:
   rhel:
     '7': ['python%{python3_pkgversion}-docker']
   ubuntu: [python3-docker]
+python3-docker-pip:
+  '*':
+    pip:
+      packages: [docker]
 python3-docopt:
   arch: [python-docopt]
   debian: [python3-docopt]


tfoote (changes requested):
> Note: A rosdep package already exists for the system package python3-docker; however, this PR adds the pip version, as the system package often contains outdated versions of the package, which are missing important bug fixes and new features.

As part of being in a distro is that we have to work with the QA'd versions. If we potentially just push this ahead to the pip version we are potentially breaking everything else that might depend on the system installed version. 

THis is in the review policy as: 
> Native packages are strongly preferred. If a native package exists for the key in question, then that should always be used over e.g. a pip key.

bastianschildknecht:
> > Note: A rosdep package already exists for the system package python3-docker; however, this PR adds the pip version, as the system package often contains outdated versions of the package, which are missing important bug fixes and new features.
> 
> As part of being in a distro is that we have to work with the QA'd versions. If we potentially just push this ahead to the pip version we are potentially breaking everything else that might depend on the system installed version.
> 
> THis is in the review policy as:
> 
> > Native packages are strongly preferred. If a native package exists for the key in question, then that should always be used over e.g. a pip key.

I understand, but to me, it is not at all clear what the ROS intended way is to pull upstream fixes and features for packages in older ROS distros without up-to-date system packages. In these cases, rosdep becomes unusable, and one would need to engineer some parallel dependency resolving system just to get access to bug fixes.

emersonknapp:
I've added this to the agenda for the August 12 ROS PMC meeting to discuss what the recommended path forward is. 

The use case is well understood, but the recommendation toward system packages in rosdep is for strong reasons, given that rosdistro produces system packages, and there is not a way to manage mixed package managers with guarantees of good results. You're right that in practice, many projects do have a side channel for language-specific dependencies that are in packages not targeted for rosdistro release.

Adding standard pip disclaimer for reference
> Standard pip disclaimer: ROS packages that depend on `pip` keys cannot be released into a ROS distribution.
> They can only be depended on by from-source builds.
> Because of this, system packages are highly preferred to pip packages.

ros-discourse:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-august-12-2025/49548/3

PR #47024, opened 2025-07-23
Remove pybind11-json-vendor package from rolling

luca-della-vedova:
This PR removes the `pybind11-json-vendor` package from Rolling and marks it as unmaintained.

The package was used in the Open-RMF project but since it was vendored a system dependency was added to Ubuntu, the package migrated to it and it is now archived https://github.com/open-rmf/pybind11_json_vendor/

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -5488,16 +5488,11 @@ repositories:
       type: git
       url: https://github.com/open-rmf/pybind11_json_vendor.git
       version: main
-    release:
-      tags:
-        release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/pybind11_json_vendor-release.git
-      version: 0.5.0-1
     source:
       type: git
       url: https://github.com/open-rmf/pybind11_json_vendor.git
       version: main
-    status: developed
+    status: unmaintained
   pybind11_vendor:
     doc:
       type: git


rolling/distribution.yaml
```diff
@@ -5488,16 +5488,11 @@ repositories:
       type: git
       url: https://github.com/open-rmf/pybind11_json_vendor.git
       version: main
-    release:
```
asymingt, line 5491: @nuclearsandwich  is there a way of measuring the potential impact removing this will have on third parties that might depend on the vendored version? Would it not be safer to keep it in, but just change the status to unmaintained?

Yadunund: One reason for removing the `release` tag is to prevent the tagged version of the package from being released to future ROS distros (eg., L-turtle).
And it's usually up to the maintainers to factor in the impact of removing a package and we don't want to gatekeep that decision especially on Rolling.

asymingt: In that case, LGTM and I will prod @nuclearsandwich  for a second review.

asymingt (approved):
LGTM

mjcarroll (approved):

mxgrey:
Removing this vendor package from the rosdistro release was a totally reasonable thing to do, but unfortunately an [unreasonable dependency problem](https://openrobotics.zulipchat.com/#narrow/channel/526042-Infrastructure-General/topic/Conflicting.20version.20of.20nlohmann-json.20in.20Resolute/with/593414087) upstream in Ubuntu Resolute has made it necessary to bring this vendor package back.

I've taken the step of [un-archiving the repo](https://github.com/open-rmf/pybind11_json_vendor/pull/18) and want to check what the right way to bring this back into rosdistro is.

Should I simply bloom-release `pybind11_json_vendor` as normal, or are there additional considerations that I need to account for?

mxgrey:
https://github.com/ros/rosdistro/pull/51575 is the outcome of doing a `bloom-release` on the resuscitated `pybind11_json_vendor` repo. It looks like a reasonable revert of this PR, except perhaps we should change the status from `unmaintained` to `maintained`?

PR #47089, opened 2025-07-28
Add MuJoCo MJX entry

saikishor:
Package name:

mujoco-mjx

Package Upstream Source:

https://mujoco.readthedocs.io/en/stable/mjx.html

Purpose of using this:

This package is a re-implementation of the [MuJoCo physics engine](https://github.com/google-deepmind/mujoco) in [JAX](https://github.com/jax-ml/jax). This library is developed and maintained by Google DeepMind and is kept up-to-date with the latest developments in MuJoCo itself.

    Python:
       https://pypi.org/project/mujoco-mjx/

Complements https://github.com/ros/rosdistro/pull/46859

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7486,6 +7486,22 @@ python3-mss-pip:
   '*':
     pip:
       packages: [mss]
+python3-mujoco-mjx-pip:
+  arch:
+    pip:
+      packages: [mujoco-mjx]
+  debian:
+    pip:
+      packages: [mujoco-mjx]
+  fedora:
+    pip:
+      packages: [mujoco-mjx]
+  osx:
+    pip:
+      packages: [mujoco-mjx]
+  ubuntu:
+    pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -7486,6 +7486,22 @@ python3-mss-pip:
   '*':
     pip:
       packages: [mss]
+python3-mujoco-mjx-pip:
```
asymingt, line 7495: Hello @saikishor, thanks for adding this dependency to ROS. Have you considered using a wildcard to specify a pip key across all distros:

```yaml
python3-mujoco-mjx-pip:
  '*':
    pip:
      packages: [mujoco-mjx]
```

saikishor: You are right!

saikishor: Done

mjcarroll (approved):

PR #47090, opened 2025-07-28
adding python bleak

Geibinger:

## Package name:

bleak

## Package Upstream Source:

https://github.com/hbldh/bleak

## Purpose of using this:

This library provides a cross‑platform Python BLE client API. It can be used for BLE/ROS2 integrations.

- **Debian** (bookworm):  
  https://packages.debian.org/bookworm/python3-bleak  
- **Ubuntu** (noble):  
  https://packages.ubuntu.com/noble/python3-bleak

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5088,10 +5088,7 @@ python3-bitstring:
     '7': [python36-bitstring]
   ubuntu: [python3-bitstring]
 python3-bleak:
-  debian:
-    bookworm: [python3-bleak]
-    trixie: [python3-bleak]
-    sid: [python3-bleak]
+  debian: [python3-bleak]
   ubuntu:
     '*': [python3-bleak]
     focal: null
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5088,8 +5088,14 @@ python3-bitstring:
     '7': [python36-bitstring]
   ubuntu: [python3-bitstring]
 python3-bleak:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5087,6 +5087,9 @@ python3-bitstring:
     '*': [python3-bitstring]
     '7': [python36-bitstring]
   ubuntu: [python3-bitstring]
+python3-bleak:
```
asymingt, line 5090: Hi @Geibinger, you can see below that our automated checks picked up that `python3-bleak` is not available on `jammy` or `focal`. You might have to do something like this to signal that it is not available:

```yaml
python3-bleak:
  debian: [python3-bleak]
  ubuntu:
    '*': [python3-bleak]
    focal: null
    jammy: null
```

rosdep/python.yaml
```diff
@@ -5087,6 +5087,15 @@ python3-bitstring:
...
+  debian:
+    bookworm: [python3-bleak]
+    trixie: [python3-bleak]
+    sid: [python3-bleak]
```
nuclearsandwich: Since all supported debian releases contain the package the specific release can be elided.

```suggestion
  debian: [python3-bleak]
```

nuclearsandwich (approved):
Looks good after one further suggestion.

asymingt (approved):
Change LGTM

christophebedard (approved):

PR #47205, opened 2025-07-31
Update libqt5-* rosdep keys for Ubuntu 24.04 (Noble) compatibility

g-arjones:
These packages were renamed on Ubuntu 24.04. This patch renames them on 'noble', while keeping the names on 'focal' and 'jammy'.


christophebedard (approved):
LGTM. Note that we require 2 approvals for rosdep keys/changes.

mjcarroll (approved):

PR #47264, opened 2025-08-08
Add python3-mongomock-pip for debian and ubuntu

jackiescanlon:

## Package name:

`python3-mongomock-pip`

## Package Upstream Source:

https://github.com/mongomock/mongomock

## Purpose of using this:

Library to use for testing code that uses a mongo database. Not available via apt (yet)

- Debian: https://pypi.org/project/mongomock/
- Ubuntu: https://pypi.org/project/mongomock/

# Note

I experienced the same issue reported [here](https://github.com/ros/rosdistro/pull/46675#issue-3184234496) (`ModuleNotFoundError: No module named 'lark'`) when trying to build the tests, and adding `lark` to my `requirements.txt` fixed the issue. However I saw that that was later removed from that PR so it seems like maybe I shouldn't commit it here. Let me know if I should.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7428,13 +7428,20 @@ python3-mock:
       packages: [mock]
   rhel: ['python%{python3_pkgversion}-mock']
   ubuntu: [python3-mock]
-python3-mongomock-pip:
+python3-mongomock:
   debian:
-    pip:
-      packages: [mongomock]
+    '*': [python3-mongomock]
+    bookworm:
+      pip:
+        packages: [mongomock]
   ubuntu:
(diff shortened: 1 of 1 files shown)


jackiescanlon:
@mjcarroll would you be able to review this for me? Thanks!

mjcarroll (approved):

nuclearsandwich:
This package actually is available as of [Trixie](https://packages.debian.org/trixie/python3-mongomock) and [Noble](https://packages.ubuntu.com/plucky/python3-mongomock) so instead of a full pip key we could use a system package key with pip fallback for older distributions.

PR #47270, opened 2025-08-11
add hydro-core from pypi

peter-mitrano-ar:

## Package name:

hydra-core

## Package Upstream Source:

https://github.com/facebookresearch/hydra

## Purpose of using this:

From their website
| A framework for elegantly configuring complex applications.

Used mostly in ML context. Can load yaml files and configure python classes based on them.

Distro packaging links:

- Debian:
  - not available
- Ubuntu:
  - not available
- Python:
    - https://pypi.org/project/hydra-core/

As far as I am aware this code is _only_ packages via pypi. I have checked debian and ubuntu. The `hydra` packages that exist there are logon crackers and are unrelated.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6727,6 +6727,9 @@ python3-httpx:
     focal:
       pip:
         packages: [httpx]
+python3-hydra-core-pip:
+  '*':
+    pip: [hydra-core]
 python3-hypothesis:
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

mjcarroll:

(Though looks like you are already aware!)

nuclearsandwich (approved):

PR #47284, opened 2025-08-12
add 'open3d' to rolling

christian-rauch:

## Package name:

~~`open3d`~~
`open3d_vendor`

## Package Upstream Source:

https://github.com/isl-org/Open3D.git

## Purpose of using this:

Open3D is an open-source library that supports rapid development of software that deals with 3D data.

`jazzy`, `rolling`

# The source is here:

https://github.com/christian-rauch/open3d_colcon

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4852,7 +4852,7 @@ repositories:
       url: https://github.com/ompl/ompl.git
       version: main
     status: developed
-  open3d:
+  open3d_vendor:
     source:
       type: git
       url: https://github.com/christian-rauch/open3d_colcon.git
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -4852,6 +4852,11 @@ repositories:
       url: https://github.com/ompl/ompl.git
       version: main
     status: developed
+  open3d:
+    source:
+      type: git
+      url: https://github.com/christian-rauch/open3d_colcon.git
+      version: main
   open_manipulator:
     doc:
       type: git


rolling/distribution.yaml
```diff
@@ -4852,6 +4852,11 @@ repositories:
       url: https://github.com/ompl/ompl.git
       version: main
     status: developed
+  open3d:
```
mjcarroll: While we don't have explicit guidance, on packages that are purely vendoring upstream dependencies, the preference is to put `_vendor` as a suffix on the package name.  So in this case `open3d_vendor`

christian-rauch: My understanding is that "vendor" packages are those that automatically fall back to system dependencies, when they are available. In this case, I want to provide the binary packages that contain GPU/CUDA backends, which are not available in the fallback system packages.

In this sense, `open3d` would be similar to the [`ompl` package](https://github.com/ompl/ompl/blob/main/package.xml) which is available without the `_vendor` suffix in rolling:
https://github.com/ros/rosdistro/blob/233e48cb8fa7f428a663553174b6bd8f26c7596d/rolling/distribution.yaml#L4840-L4854

`ros-jazzy-ompl` and `ros-rolling-ompl` are provided at version 1.7 although the system version (`libompl-dev`) is also available at version 1.5.

ros-discourse:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-august-12-2025/49548/1

rolling/distribution.yaml
mjcarroll: We discussed this in the PMC meeting today.  The general thought was that since OMPL is created by basically making the upstream package ROS-aware (adding a package.xml with a build type CMake), and blooms happen directly from that repo, it makes sense to not have `_vendor`.

On the other hand, the open3d package is fetching external sources and building them in a ROS-compatible way, which looks much more like a vendor package, so the suffix would be suitable here.

The other note was that open3d will be part of the next Ubuntu release, so maybe a vendor package makes sense to eventually point off to the OS variant of the package.  If the version in the system isn't sufficient, then unconditionally building the vendor package may still make sense (we do this in at least one right now that I'm struggling to remember).

christian-rauch: > The general thought was that since OMPL is created by basically making the upstream package ROS-aware (adding a package.xml with a build type CMake), and blooms happen directly from that repo, it makes sense to not have `_vendor`.

Ok. That argument makes sense to me.

> The other note was that open3d will be part of the next Ubuntu release, so maybe a vendor package makes sense to eventually point off to the OS variant of the package.

I explicitly wanted this package here to persist in parallel to the system package, because this is the "official" binary distribution by Intel that also contains the GPU/CUDA backend. So I explicitly do not want the `open3d` package to fall back to the system package (e.g. apt packages `libopen3d-dev` and `libopen3d0.19`), as this would remove the GPU/CUDA backend and would exclude some applications, such as running SLAM in real-time.

> If the version in the system isn't sufficient, then unconditionally building the vendor package may still make sense (we do this in at least one right now that I'm struggling to remember).

Do you mean with "unconditionally building", in this concrete case, that the GPU-enabled Open3D package (`open3d` or `open3d_vendor`) would be available in any case, and that users can choose between this package and the system version (via the rosdep key `libopen3d`)? In principle, this is what I aim for.

But the `_vendor` is confusing in this case. If I look at other "vendor" packages, such as [`orocos_kdl_vendor`](https://github.com/ros2/orocos_kdl_vendor/blob/rolling/orocos_kdl_vendor/package.xml) then I see that this package depends on the system package (`<depend>liborocos-kdl-dev</depend>`). And this is explicitly not what I intended. I want the `open3d` package with the "official" binary distribution from upstream (i.e. the gold standard) with the GPU support to be separate from the distribution package, and I want users to be able to chose between those packages by specifying either the rosdep key for the "official" binary package, or using the distribution package.

If this package would only be a temporary workaround for the missing apt package `libopen3d-dev` in Ubuntu 24.04, I would call this a "vendor" package. But in this case, my aim is to provide the "official" Open3D binaries with full feature support (CUDA, etc.) in parallel to the system package.

mjcarroll: > then I see that this package depends on the system package

This doesn't _always_ have to be the case.  Some people have chosen to either not depend on the system package, or have an override via cmake like `ALWAYS_VENDOR`.  One example I can think of is  https://github.com/eProsima/foonathan_memory_vendor.git

christian-rauch: I still find the terminology confusing. I am more or less just a package maintainer in the same sense that most Linux distributions just maintain upstream packages, but they wouldn't suffix all Debian packages just because they distribute them.

With the argument that this package is not to replace the Open3D system package, but coexist in parallel so that users can choose between the two variants (system version without GPU support and official version with GPU support), should this be seen as a vendor package?

mjcarroll: > With the argument that this package is not to replace the Open3D system package, but coexist in parallel so that users can choose between the two variants (system version without GPU support and official version with GPU support), should this be seen as a vendor package?

Yes, I think that's the strongest argument to call it a vendor package.

christian-rauch: Alright. I added the `_vendor` suffix to the package name.

PR #46902, opened 2025-07-14
moveit_msgs: 2.7.0-2 in 'kilted/distribution.yaml'

nbbrooks:
This was manually generated due to GH classic token limitations.

Increasing version of package(s) in repository moveit_msgs to 2.7.0-2:

*  upstream repository: https://github.com/moveit/moveit_msgs.git
*  release repository: https://github.com/ros2-gbp/moveit_msgs-release.git
*  distro file: `kilted/distribution.yaml`
*  bloom version: `0.13.0`
*  previous version for package: `2.6.0-1`

## moveit_msgs

```
* Add GetMultiStateValidity service (#186 <https://github.com/ros-planning/moveit_msgs/issues/186>)
* CI: Update actions
* Contributors: Lucian James, Robert Haschke
```

PR #47286, opened 2025-08-13
moveit_msgs: 2.7.1-1 in 'kilted/distribution.yaml' [bloom]

v4hn:
Increasing version of package(s) in repository `moveit_msgs` to `2.7.1-1`:

- upstream repository: https://github.com/moveit/moveit_msgs.git
- release repository: https://github.com/ros2-gbp/moveit_msgs-release.git
- distro file: `kilted/distribution.yaml`
- bloom version: `0.13.0`
- previous version for package: `2.6.0-2`

## moveit_msgs

```
* Address 2.7.0 release issues via re-release as 2.7.1
* Contributors: Nathan Brooks
```

### DIFF
--- a/kilted/distribution.yaml
+++ b/kilted/distribution.yaml
@@ -4220,7 +4220,7 @@ repositories:
       tags:
         release: release/kilted/{package}/{version}
       url: https://github.com/ros2-gbp/moveit_msgs-release.git
-      version: 2.6.0-2
+      version: 2.7.1-1
     source:
       type: git
       url: https://github.com/ros-planning/moveit_msgs.git


v4hn:
This replaces the broken https://github.com/ros/rosdistro/pull/46902 with @nbbrooks's new tagged version 2.7.1 .
I reverted/removed various patches/ branches in the gbp repository to clean up the invalid commit references that were introduced during the previous broken release process and went with the last tagged upstream version.

Obligatory statement: I think this is the appropriate place to point out that the whole gbp-repository-creates-a-myriad-of-branches machinery should not exist outside the buildfarm-build process in the first place (and several companies do not have it in their internal buildfarms). The use of all target-platform specific branches is to add custom patches on the **package maintainer** side for platforms where they are not included in the upstream repository. I'm only aware of a handful of maintainers around ROS who would not be desperately confused by this mechanism where it is actually needed and I believe they agree that it does not make sense in ROS. ROS package maintainers are pretty much always responsible for their packages on *all* target platforms instead of being responsible for a number of packages for only a single target platform. But the overhead generates weird erroneous situations like this where something goes wrong.

mjcarroll (approved):

ros-discourse:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-august-19-2025/49758/1

PR #47288, opened 2025-08-13
rosdep: adding dependencies for openembedded

vermaete:
Python:

  * pyhton3-wheel: https://layers.openembedded.org/layerindex/recipe/299625/
  * python3-attrs: https://layers.openembedded.org/layerindex/recipe/162120/
  * python3-filelock: https://layers.openembedded.org/layerindex/recipe/340331/
  * python3-httpx: https://layers.openembedded.org/layerindex/recipe/398317/
  * python3-platformdirs: https://layers.openembedded.org/layerindex/recipe/339299/
  * python3-setproctitle: https://layers.openembedded.org/layerindex/recipe/419953/
  * python3-toml: https://layers.openembedded.org/layerindex/recipe/414804/
  * python3-tqdm: https://layers.openembedded.org/layerindex/recipe/97211/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4300,6 +4300,7 @@ python-tqdm:
     buster: [python-tqdm]
     stretch: [python-tqdm]
   fedora: [python-tqdm]
+  openembedded: [python3-tqdm@meta-python]
   ubuntu:
     bionic: [python-tqdm]
 python-transforms3d-pip:
@@ -4980,6 +4981,7 @@ python3-attrs:
   debian: [python3-attr]
   fedora: [python3-attrs]
   nixos: [python3Packages.attrs]
+  openembedded: [python3-attrs@openembedded-core]
   opensuse: [python3-attrs]
   rhel: [python3-attrs]
   ubuntu: [python3-attr]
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

rosdep/python.yaml
robwoolley, line 1: Using ${PYTHON_PN} was correctly historically when OpenEmbedded had both python and python3.  I think it would be better to simplify things and just use python3 going forward as the python-simplejson recipe no longer exists.  (The meta-python2 layer went EOL in January 2020)

Replacing ${PYTHON_PN} with python3 is a general change that would be worth doing if you're up for it.

robwoolley: I manually checked all the entries.  They all appear correct.

robwoolley (approved):
I checked the file for occurrences of ${PYTHON_PN}.  There appear to be 62 references.  I suggest approving these changes as they are technically correct.  I suggest creating a new PR to replace ${PYTHON_PN} with python3

rosdep/python.yaml
mjcarroll: Thanks @robwoolley!

nuclearsandwich (approved):

PR #47292, opened 2025-08-14
Revert ros2_control due to jazzy breaking deprecation.

marcoag:
Ros2_control added a deprecation in [4.34.0](https://github.com/ros-controls/ros2_control/blob/4.34.0):

https://github.com/ros-controls/ros2_control/blob/d026efc28a26f17d67b9221261a409b79fe3577a/hardware_interface/CHANGELOG.rst?plain=1#L15

This breaks downstream packages [feetech_ros2_driver](https://github.com/JafarAbdi/feetech_ros2_driver) and [topic_based_ros2_control](https://github.com/PickNikRobotics/topic_based_ros2_control). I'm reverting to the latest version without the deprecation to avoid the wipe out of those packages in the [upcoming jazzy patch sync](https://discourse.openrobotics.org/t/preparing-for-jazzy-sync-and-patch-release-2025-08-14/).

I'll revert this change as soon as the sync is done.

FYI: @bmagyar

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -7725,7 +7725,7 @@ repositories:
       tags:
         release: release/jazzy/{package}/{version}
       url: https://github.com/ros2-gbp/ros2_control-release.git
-      version: 4.35.0-1
+      version: 4.33.0-1
     source:
       type: git
       url: https://github.com/ros-controls/ros2_control.git


bmagyar:
@marcoag I stand somewhat corrected. I've just found your PRs addressing the deprecated API. While I appreciate the effort, I think for packages where this issue turns up, removing the `-Werror` is a more lasting fix for the buildfarm.

marcoag:
> @marcoag , I'd like to challenge this policy. Deprecation notices are important to warn the userbase on older distros and to facilitate migrating to new API gradually. Some packages using unreasonable flags in their compilation setting (typically the `-Wall -Werror` combo) shouldn't be the reason to roll our releases back.

First thanks for taking time to address the issue. I would say this is not a policy per se but more of a ROS boss decision. I would add this shouldn't be considered a a roll back since the package was not out, but more of a delay to give downstream packages time to fix the issues. I will try to have a look at issues like this on a more weekly basis and not only when releases are coming so maintainers get more time to react to possible suggestions.

We recommend no breaking changes on packages released on stable ROS releases however it's always up to the discretion of the maintainers. Leaving aside the fact that a deprecation might or might not be acceptable on a stable release, the fact is that this one is breaking downstream packages. Given the upcoming sync the a decision had to be made to either wipe out those packages or delay the release of a newer version of `ros2_control`. Any decision has its downsides but when it comes to ROS Bossing a release I believe the best is to choose the one that brings less hassle to the  ROS release user base. I think in this case delaying the release of `ROS2_control` and allowing time to fix the downstream packages so they don't get wiped out suddenly was the best that could be done.
> 
> I think a workflow where people are notified _not as part of a revert a day before a sync_ would help ensure a healthy community environment. I'd expect that first notifications would go out to the maintainers of the failing packages and if they fail to address the issue, they get cut off the build. If they have active maintainers around, it should not be a problem to get back in. If we aren't allowed to add deprecation notices to widely used packages, we'll end up stalling the ecosystem like in ROS1.

Notifications are sent to failing packages maintainers as they fail to build on the buildfarm. Do you also mean notifying upstream packages maintainers causing the downstream breakage? not sure if this can create a burst of unwanted notifications though... 

I do see an opportunity for a policy in this direction where a specific period of time is given to maintainers for package fixes after which they can face wipe out. This is something I could bring up to the ROS PMC for discussion.

> @marcoag I stand somewhat corrected. I've just found your PRs addressing the deprecated API. Thank you for doing that but while I appreciate the effort, I think for packages where this issue turns up, removing the `-Werror` is a more lasting fix for the buildfarm.

I do agree with this and add it to the PRs. Thanks!

PR #47303, opened 2025-08-15
Firefox geckodriver

BlueNoise:

## Package name:
geckodriver

## Package Upstream Source:
https://github.com/mozilla/geckodriver

## Purpose of using this:
This dependency provides the WebDriver server for automating Mozilla Firefox, used by Selenium and other browser automation frameworks.  
We require this for automated testing and interaction with web interfaces in ROS-based systems.

- Debian: 
  - **Not available**
- Ubuntu focal: https://launchpad.net/ubuntu/focal/+package/firefox-geckodriver

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1428,6 +1428,9 @@ gdal-bin:
   gentoo: [sci-libs/gdal]
   nixos: [gdal]
   ubuntu: [gdal-bin]
+geckodriver:
+  ubuntu:
+    focal: [firefox-geckodriver]
 geographiclib:
   debian:
     '*': [libgeographiclib-dev]


sloretz (commented):
Thank you for the PR!

I don't think we should take this rosdep key. It looks like the Debian package was removed in Ubuntu Jammy https://bugs.launchpad.net/ubuntu/+source/firefox/+bug/1968266 , and when I look at the documentation [it seems to recommend installing geckodriver via snap](https://firefox-source-docs.mozilla.org/testing/geckodriver/Usage.html#Running-Firefox-in-an-container-based-package).

We could add this key for Ubuntu Focal, but [Ubuntu Focal reached end of life earlier this year](https://ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare). I think we'd need to make rosdep aware of `snap` on ubuntu to support installing this dependency going forward.

---
I'll bring this up at the next ROS 2 PMC meeting.

tfoote (changes requested):
I agree, that I think merging this for an EOL distro doesn't make sense. We don't often prune the keys but since it's EOL this would already be elligable for pruning. I would suggest that you create your own rosdep sources list that you can add to your system's sources since this is outside of the scope of what we want to maintain publicly, but clearly is valuable to you

PR #47308, opened 2025-08-18
Add orbbec_camera and orbbec_camera_msgs to ROS Humble index

obyalian:
# Add OrbbecSDK_ROS2 packages to ROS humble and jazzy Index

This PR indexes the following packages from the OrbbecSDK_ROS2 repository into the ROS humble and jazzy distribution:

- orbbec_camera
- orbbec_camera_msgs

Repository URL: https://github.com/orbbec/OrbbecSDK_ROS2.git  
Branch: v2-main  
Status: maintained

# Checks
 - [x] These packages are expected to build on the submitted ROS Humble distribution

# Notes
Both packages are part of the same repository but are separate ROS packages with independent package.xml files.


jazzy/distribution.yaml
```diff
@@ -5832,6 +5832,18 @@ repositories:
       url: https://github.com/hatchbed/opensw_ros.git
       version: ros2
     status: developed
+  orbbec_camera:
+    source:
+      type: git
+      url: https://github.com/orbbec/OrbbecSDK_ROS2.git
+      version: v2-main
```
JafarAbdi, line 5919: Should we have a different name for v2 vs the main branch? Like something similar to behaviortree? https://github.com/ros/rosdistro/blob/master/jazzy/distribution.yaml#L906 both seem to be supported

PR #47569, opened 2025-09-09
Update the development branch of rosbridge_suite in Jazzy

bjsowa:


cottsay (commented):
It should be updated to match here as well: https://github.com/ros2-gbp/rosbridge_suite-release/blob/860af510566e744467aae52b8b892a237e5cca53/tracks.yaml#L145

cottsay (approved):
Thanks, looks good

PR #47688, opened 2025-09-17
Add pip fallbacks for python3-docstring-parser

ndahn:
Please update the following dependency.

## Package name:

python3-docstring-parser (*rule update*)

## Package Upstream Source:

https://github.com/rr-/docstring_parser

## Purpose of using this:

This package parses python docstrings and makes them available for programmatic digestion. It's very useful when creating new (CLI) tools, as it allows to generate e.g. help strings for functions and parameters without duplication.

Distro packaging links:

I previously assumed that the os_name fallback would also apply to missing os_version tags (related issue: https://github.com/ros-infrastructure/rosdep/pull/1015). This fixes this oversight.

- pypi
  - https://pypi.org/project/docstring-parser/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5747,6 +5747,9 @@ python3-docstring-parser:
   nixos: [python3Packages.docstring-parser]
   ubuntu:
     '*': [python3-docstring-parser]
+    focal:
+      pip:
+        packages: [docstring-parser]
     jammy:
       pip:
         packages: [docstring-parser]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5736,10 +5736,16 @@ python3-docstring-parser:
     pip:
       packages: [docstring-parser]
   debian:
+    '*':
+      pip:
+        packages: [docstring-parser]
     trixie: [python3-docstring-parser]
   fedora: [python3-docstring-parser]
   nixos: [python3Packages.docstring-parser]
   ubuntu:
+    '*':
+      pip:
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -5736,10 +5736,16 @@ python3-docstring-parser:
     pip:
       packages: [docstring-parser]
   debian:
+    '*':
+      pip:
+        packages: [docstring-parser]
```
mjcarroll, line 5742: I believe in this case, that you would want to create a `python3-docstring-parser-pip` package and depend on that.

We keep the dependencies explicit as you can't build a binary debian against pip dependencies.

mjcarroll: Actually, this is the correct way to do this.

mjcarroll (approved):

clalancette (changes requested):
Honestly, I think we should redo this entire key to be more like our other ones and future-proof.  The GitHub UI won't quite let me do this, but I think this entire key should look like:

```
python3-docstring-parser:
  debian:
    '*': [python3-docstring-parser]
    bookworm:
      pip:
        packages: [docstring-parser]
    bullseye:
      pip:
        packages: [docstring-parser]
  fedora: [python3-docstring-parser]
  nixos: [python3Packages.docstring-parser]
  ubuntu:
    '*': [python3-docstring-parser]
    jammy:
      pip:
        packages: [docstring-parser]
```

clalancette (approved):

PR #47772, opened 2025-09-24
openembedded: python(3)pyudev

vermaete:
Removed the not longer existing Python2 version of pyudev.

The Python3 version of pyudev is in the meta-python layer See: https://layers.openembedded.org/layerindex/recipe/66922/

@robwoolley Could you review please?


robwoolley:
Thanks, Jan!  Confirmed: looks good.

clalancette (approved):

christophebedard (approved):

PR #47773, opened 2025-09-24
openembedded: python(3)-six

vermaete:
Removed the not longer existing Python2 recipe of six.

The Python3 version of six is at openembedded-core. See: https://layers.openembedded.org/layerindex/recipe/51784/

@robwoolley Could you review please?

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3883,7 +3883,6 @@ python-six:
   fedora: [python-six]
   gentoo: [dev-python/six]
   nixos: [pythonPackages.six]
-  openembedded: ['${PYTHON_PN}-six@openembedded-core']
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
This one is failing CI because 'openembedded' comes before 'rhel' alphabetically.

rosdep/python.yaml
```diff
@@ -8321,6 +8320,7 @@ python3-posix-ipc:
     pip:
       packages: [posix-ipc]
   fedora: [python3-posix-ipc]
+  openembedded: [python3-six@openembedded-core]
```
clalancette, line 8323: This is now modifying the incorrect key.

clalancette (changes requested):

PR #47818, opened 2025-09-26
Update foxglove_bridge to 3.2.1, update repo and remove old repo

ericmlujan:
Update foxglove_bridge and foxglove_msgs to 3.2.1.

This PR is manually generated because we are migrating foxglove_bridge (previously in the foxglove/ros-foxglove-bridge) repo to Foxglove's monorepo (foxglove/foxglove-sdk). This repo provides both the foxglove_bridge and foxglove_msgs packages, and we needed to remove the old foxglove_msgs repo to avoid conflicts/bloom throwing an assertion.

See #44646 for the Rolling version of this PR.

### DIFF
--- a/kilted/distribution.yaml
+++ b/kilted/distribution.yaml
@@ -2189,19 +2189,22 @@ repositories:
       url: https://github.com/ros-drivers/four_wheel_steering_msgs.git
       version: ros2
     status: maintained
-  foxglove_bridge:
+  foxglove-sdk:
     doc:
       type: git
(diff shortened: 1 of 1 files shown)


christophebedard (approved):
I understand that this doesn't remove any package; it just combines the existing `foxglove_bridge` and `foxglove_msgs` packages under the new/renamed foxglove/foxglove-sdk repository (+ bumps their versions): https://github.com/foxglove/foxglove-sdk/tree/main/ros/src

so LGTM

PR #47916, opened 2025-10-03
openembedded: adding python3-tabulate

vermaete:
@robwoolley Could you review please


asymingt (approved):
I have verified the exists here: http://layers.openembedded.org/layerindex/branch/master/layer/meta-python/

LGTM

christophebedard (approved):

christophebedard:
I think we can proceed with this. If Rob does have feedback, we can address it in a later PR.

PR #48280, opened 2025-10-21
add easysnmp to python.yaml

NouiraSaeve:

## Package name: 
easysnmp

## Package Upstream Source: 
https://github.com/easysnmp/easysnmp

## Purpose of using this:
easysnmp is a Python library that provides a simple interface for working with SNMP

Distro packaging links:

https://pypi.org/project/easysnmp/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -11366,7 +11366,7 @@ zulip-pip:
   ubuntu:
     pip:
       packages: [zulip]
-python3-easysnmp-pip:
+python3-easysnmp:
   ubuntu:
     noble: [python3-easysnmp]
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -5795,6 +5795,11 @@ python3-easydict:
     focal:
       pip:
         packages: [easydict]
+python3-easysnmp:
+  ubuntu:
+    noble: [python3-easysnmp]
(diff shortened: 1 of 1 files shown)


mjcarroll:
Just curious if the version already in Debian/Ubuntu would be sufficient here? https://launchpad.net/ubuntu/noble/amd64/python3-easysnmp/0.2.6-1build1

rosdep/python.yaml
```diff
@@ -5836,6 +5836,11 @@ python3-easydict:
     focal:
       pip:
         packages: [easydict]
+python3-easysnmp:
+  ubuntu:
+    noble: [python3-easysnmp]
+    pip:
+      packages: [easysnmp]
```
cottsay: The way we typically structure a rule set like this is to make the wildcard rule "forward-looking" so that it supports future releases without modification. Then we back-fill the "supported" releases that don't have the package with either an explicit `null` or another rule, like pip.

```suggestion
  ubuntu:
    '*': [python3-easysnmp]
    focal:
      pip:
        packages: [easysnmp]
    jammy:
      pip:
        packages: [easysnmp]
```

cottsay (dismissed):

emersonknapp (approved):

skyegalaxy (approved):

PR #48406, opened 2025-10-30
Changed owner of rplidar_ros repository for jazzy

blockdoth:
## Purpose
The purpose of this PR is to update the ownership of the rplidar_ros repository from an out of date version originally maintained by @allenh1 to the more up to date repository maintained by a @deyouslamtec, a commercial representative from [Slamtec](https://www.slamtec.com/). The maintainer has been made aware #https://github.com/Slamtec/rplidar_ros/issues/164 but has not initiated any change yet.

## My Issue
My personal issue is that the current repo does not have support for the C1 Lidar for ros2, while the official one does.

## Relevant distro's
For me this issue primarily concerns jazzy, because that is the distro I am actively using for my project. However since I am not fully aware of how  this chance fits into the distrobution release cycles I am open to guidance on this.

#### The original repository
https://github.com/allenh1/rplidar_ros
#### The updated repository
https://github.com/Slamtec/rplidar_ros

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -8965,6 +8965,10 @@ repositories:
       version: master
     status: maintained
   rplidar_ros:
+    doc:
+      type: git
+      url: https://github.com/Slamtec/rplidar_ros.git
+      version: ros2
     release:
       tags:
         release: release/jazzy/{package}/{version}
@@ -8973,7 +8977,7 @@ repositories:
     source:
       test_pull_requests: true
       type: git
-      url: https://github.com/allenh1/rplidar_ros.git
+      url: https://github.com/Slamtec/rplidar_ros.git
       version: ros2
     status: developed
   rpyutils:


basvd:
I'm also interested in this update for the same reason: C1 support.

However, I'm confused by the proposed change. Don't you mean to set the source URL to `https://github.com/Slamtec/rplidar_ros`? The one by deyouslamtec does not contain the 2.1.5 tag which includes the C1 support.

In fact, I think the jazzy distribution YAML should resemble/be identical to the humble one:
```
  rplidar_ros:
    doc:
      type: git
      url: https://github.com/Slamtec/rplidar_ros.git
      version: ros2
    release:
      tags:
      url: https://github.com/ros2-gbp/rplidar_ros-release.git
      version: 2.1.4-1
    source:
      test_pull_requests: true
      type: git
      url: https://github.com/Slamtec/rplidar_ros.git
      version: ros2
    status: developed
```

I suppose the version should also be bumped to 2.1.5, but I am not familiar with the rosdistro process.

blockdoth:
You are right, my mistake. I did remember noticing but didnt act on it. Pushed changes with an added docs section aswell

As for the versioning, not sure how to proceed, waiting on input from somebody with more knowledge about the ROS distro lifecycle conventions

mjcarroll:
> Let's wait a bit for the original maintainer to reply to see if they're fine with it.

I would also like to wait for original maintainer.  This is also the appropriate thing to do by policy.

Just checking the history here, did @allenh1 create the ROS 2 fork because upstream hadn't done it yet, and now upstream correctly supports ROS 2?

mjcarroll:
> Hunter happens to be a coworker, so I ~harassed~ pinged him internally.

https://osrf.github.io/code_of_conduct/

joshnewans:
As the creator (6 months ago) of the issue mentioned in the original post I would be keen for this to be resolved so that I can update my tutorials which use this package.

However it would be best if we can get `kilted` and `rolling` updated as well to avoid this issue in the future.

mjcarroll:
Thanks @allenh1, this all sounds about how I recall it.

As a further datapoint, this appears to have already been done for humble: https://github.com/ros/rosdistro/blob/master/humble/distribution.yaml#L9935-L9950, so it may have just fallen through the cracks here.

joshnewans:
Just checking @mjcarroll this was only for jazzy, yeah? Do we need to submit separate PRs for kilted and rolling so that it's not like humble where we have the same problem again?

mjcarroll:
> Just checking @mjcarroll this was only for jazzy, yeah? Do we need to submit separate PRs for kilted and rolling so that it's not like humble where we have the same problem again?

This is only for jazzy.  AS far as I can tell, we never had releases for kilted or rolling, so you can set those up fresh.

joshnewans:
> This is only for jazzy. AS far as I can tell, we never had releases for kilted or rolling, so you can set those up fresh.

Hmm it looks like there's no `doc` entry but it does have `release` and `source` (both still referencing `allenh1`).

https://github.com/ros/rosdistro/blob/9bb250386c83a3d6a21008153581ebdb659854b7/kilted/distribution.yaml#L7949-L7960

https://github.com/ros/rosdistro/blob/9bb250386c83a3d6a21008153581ebdb659854b7/rolling/distribution.yaml#L7886-L7897

mjcarroll:
> Hmm it looks like there's no doc entry but it does have release and source (both still referencing allenh1).

You are absolutely right! (I swear I'm not an AI).  I actually went and did a text search, but it was probably before the file finished loading, so I didn't see it.  In that case, you need to make these adjustments to those distros as well.

joshnewans:
Thanks, I'll submit an equivalent PR for the other repos in the coming days.

Issue #48237, opened 2025-10-19
Update entries to use system packages over PyPI ones to prevent shadowing


BlueberryCatt:
A bunch of Python packages that were previously available only on PyPI have been added as native system libraries in major Linux distributions (e.g., Ubuntu and Debian). According to the [contributing guidelines](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-rosdep-rules), system packages should be preferred over PyPI ones. This practice helps prevent conflicts that arise when PyPI-installed packages shadow system libraries. In addition, packages built by the ROS buildfarm are prohibited from depending on pip-only keys.

A similar case was discussed in [#19058](https://github.com/ros/rosdistro/pull/19058#issuecomment-788115489), where PyPI packages were replaced with system libraries once the latter became available.

The following table summarizes several entries that still install packages from PyPI even though equivalent system libraries now exist in Ubuntu and Debian (more exists for other distributions). For each entry, one example package is shown, preferring Ubuntu when available.

<details>
  <summary>Click to show/hide table of entries that now have corresponding system packages</summary>

| Entry Key                           | Example Platform | New System Package              |
| ----------------------------------- | ---------------- | ------------------------------- |
| `python3-flake8-black`              | Ubuntu Noble     | `python3-flake8-black`          |
| `python3-mongomock`                 | Debian Bookworm  | `python3-mongomock`             |
| `python3-xmlschema`                 | Ubuntu Noble     | `python3-xmlschema`             |
| `python3-fpdf2-pip`                 | Ubuntu Noble     | `python3-fpdf`                  |
| `python3-anytree-pip`               | Debian Trixie    | `python3-anytree`               |
| `python3-lingua-franca-pip`         | Ubuntu Noble     | `python3-lingua-franca`         |
| `python3-marshmallow-dataclass-pip` | Ubuntu Noble     | `python3-marshmallow-dataclass` |
| `python3-openai-pip`                | Ubuntu Noble     | `python3-openai`                |
| `python3-sense-hat-pip`             | Ubuntu Noble     | `python3-sense-hat`             |
| `python3-aio-pika-pip`              | Ubuntu Noble     | `python3-aio-pika`              |
| `python3-ansible-runner-pip`        | Ubuntu Noble     | `python3-ansible-runner`        |
| `python3-backoff-pip`               | Ubuntu Noble     | `python3-backoff`               |
[...]

PR #48659, opened 2025-11-17
Replace pip installations with native packages for Python deps on newer Debian/Ubuntu distros

BlueberryCatt:
As mentioned in #48237 , a bunch of packages from PyPI have been added as native system libraries in Ubuntu and Debian, this pr updates the corresponding entries to make them available on the build farm and avoid shadowing problems.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -7543,8 +7543,7 @@ python3-mock:
   rhel: ['python%{python3_pkgversion}-mock']
   ubuntu: [python3-mock]
 python3-mongomock:
-  debian:
-    '*': [python3-mongomock]
+  debian: [python3-mongomock]
   ubuntu:
     '*': [python3-mongomock]
     focal:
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

rosdep/python.yaml
```diff
@@ -7507,9 +7545,6 @@ python3-mock:
 python3-mongomock:
   debian:
     '*': [python3-mongomock]
-    bookworm:
-      pip:
-        packages: [mongomock]
```
cottsay, line 7512: For clarity, this package is available for Bookworm so removing the rule falls back to the apt rule.

https://packages.debian.org/bookworm/python3-mongomock

This could be simplified to just `debian: [python3-mongomock]` though (sorry I can't use the GitHub suggestion UI for it)

cottsay (commented):
LGTM, just one comment

rosdep/python.yaml
BlueberryCatt: Thanks for the reminder! I simplified it in the latest commit

ahcorde (approved):

PR #48735, opened 2025-11-22
apriltag_hitch_estimation: 0.0.1-1 in 'humble/distribution.yaml' [bloom]

li9i:
Increasing version of package(s) in repository `apriltag_hitch_estimation` to `0.0.1-1`:

- upstream repository: https://github.com/li9i/apriltag-hitch-estimation.git
- release repository: https://github.com/li9i/apriltag-hitch-estimation-release.git
- bloom version: `0.13.0`

### DIFF
--- a/humble/distribution.yaml
+++ b/humble/distribution.yaml
@@ -584,6 +584,21 @@ repositories:
       url: https://github.com/ros-misc-utilities/apriltag_detector.git
       version: release
     status: developed
+  apriltag_hitch_estimation:
+    doc:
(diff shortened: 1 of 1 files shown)


wjwwood (approved):
I think this is ok, but I also think it might be better namespaced, as starting with `apriltag_` makes it seem like it's part of that project, which from what I can see may not be the case? Anyways, I think this could be fine as-is, I just wanted to raise the issue and get a second opinion.

PR #48818, opened 2025-11-30
Add qml6_ros2_plugin to distribution.yaml

StefanFabian:
Added qml6_ros2_plugin with its source and documentation details. This is the Qt6 version of the QML ROS 2 module.

## Package name:

qml6_ros2_plugin

## Package Upstream Source:

https://github.com/StefanFabian/qml6_ros2_plugin

# Checks

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -6900,6 +6900,16 @@ repositories:
       url: https://github.com/ros-visualization/python_qt_binding.git
       version: jazzy
     status: maintained
+  qml6_ros2_plugin:
+    doc:
+      type: git
+      url: https://github.com/StefanFabian/qml6_ros2_plugin.git
+      version: jazzy
+    source:
+      type: git
+      url: https://github.com/StefanFabian/qml6_ros2_plugin.git
+      version: jazzy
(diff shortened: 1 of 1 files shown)


StefanFabian:
This is just the Qt 6 version of `qml_ros2_plugin` btw:
https://github.com/ros/rosdistro/blob/19230b546726bdbb3b249db5613db6b8e81da90f/jazzy/distribution.yaml#L6903-L6917

It requires a separate package name because you can install both Qt5 and Qt6 in parallel, but they are not source-compatible.

mjcarroll:
New-ish package, so worth double checking.

  - [x] Top level license file: https://github.com/StefanFabian/qml6_ros2_plugin/blob/jazzy/LICENSE
- [x] License is [OSI-approved](https://opensource.org/licenses): https://opensource.org/license/mit
- [x] Public source repo: https://github.com/StefanFabian/qml6_ros2_plugin

Using "ros2" in the library is acceptable here, as it falls in the category of "ROS bindings of an upstream library" detailed here: https://ros.org/reps/rep-0144.html#id12

mjcarroll (approved):

PR #48882, opened 2025-12-05
ardrone_ros: 2.0.2-1 in 'rolling/distribution.yaml' [bloom]

vtalpaert:
Increasing version of package(s) in repository `ardrone_ros` to `2.0.2-1`:

- upstream repository: https://github.com/vtalpaert/ardrone-ros2.git
- release repository: https://github.com/ros2-gbp/ardrone_ros-release.git
- bloom version: `0.13.0`

## ardrone_sdk

```
* change libncurses to libncurses-dev since we not build full sdk
* Contributors: Victor Talpaert
```

## ardrone_sumo

- No changes

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -505,7 +505,7 @@ repositories:
       tags:
         release: release/rolling/{package}/{version}
       url: https://github.com/ros2-gbp/ardrone_ros-release.git
-      version: 1.1.0-1
+      version: 2.0.2-1
     source:
       type: git
       url: https://github.com/vtalpaert/ardrone-ros2.git


vtalpaert:
@tfoote I added my sources like you suggested in https://robotics.stackexchange.com/a/116912

- License in source files are preserved
- Package license is now the one from Parrot
- License is included in the binary distribution (in cmake using INSTALL FILES)

 I chose to commit a copy of some sdk files in my repo because I didn't want the build process to use repo sync to do a git clone when building. Previously, I built the library in a tarball since I believed it was somewhat better (with md5 verification). The current solution is simply a re-commit, which is fine since the sdk is discontinued at version 3.

Other ROS1 repos use SDK2, so this should be an improvement for all Parrot drone users.

vtalpaert:
For reference https://github.com/ros/rosdistro/pull/45602#issuecomment-3003419538

mjcarroll (approved):
This LGTM, @tfoote can you give one more sanity check on the license issues that you were concerned about?

ahcorde (approved):

PR #49085, opened 2025-12-17
Clean up python-configparser entries in rosdep

Alex2099v1:
Remove deprecated python-configparser entries for all distributions.

## Package name:

TODO

## Package Upstream Source:

## Purpose of using this:

Distro packaging links:

  - REQUIRED
  - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

ROSDISTRO NAME

# The source is here:

http://sourcecode_url

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1103,11 +1103,12 @@ python-concurrent.futures:
   debian: [python-concurrent.futures]
   gentoo: [virtual/python-futures]
   ubuntu: [python-concurrent.futures]
+# in the standard library in python3 
 python-configparser:
-  arch: [python2-configparser]
-  debian: [python-configparser]
-  gentoo: [dev-python/configparser]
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Can you clarify what issue/problem you're trying to resolve. The comment says that this is built into python 3 however this is the python 2 key. If it's no longer available the correct thing to do would be to remote it not set the list to a null set. 

This is likely to break anyone on older systems. We were just discussing retirement policies in another thread. There's a level that all the python2 keys can be retired enmasse I suspect. But we should make that decision systematically versus doing it one key at a time.

mjcarroll (changes requested):
Please address lint failure (trailing spaces)

PR #49202, opened 2025-12-25
Add usd-core/pxr

baconbrot:

## Package name:

pip package name: usd-core

## Package Upstream Source:

https://github.com/PixarAnimationStudios/OpenUSD
https://pypi.org/project/usd-core/

## Purpose of using this:

OpenUSD python bindings. Use-cases for this include creation of simulation environments for Isaac Sim/Lab.
From [pypi](https://pypi.org/project/usd-core/):
> Universal Scene Description (USD) is a system that scalably encodes and interchanges static and time-sampled 3D geometry, shading and lighting data between Digital Content Creation applications. Domain-specific schema modules define the geometry, shading, and lighting encoding atop USD's domain-agnostic core. It provides a low-level data model that supports plugin implementations, and a scenegraph built on top of a multi-threaded "composition engine" for creating complex scenes out of reusable, modular assets.

Primarily via pip https://pypi.org/project/usd-core/
A fedora package is available. Should the rosdep key be named `python3-usd` according to the fedora package name?
I did not find any ubuntu/debian apt packages for this.

- Pip: https://pypi.org/
  - https://pypi.org/project/usd-core/
  - https://packages.fedoraproject.org/pkgs/usd/python3-usd/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -11065,7 +11065,7 @@ python3-usb:
   nixos: [python3Packages.pyusb]
   openembedded: [python3-pyusb@meta-python]
   ubuntu: [python3-usb]
-python3-usd-core:
+python3-usd:
   '*':
     pip:
       packages: [usd-core]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -11065,6 +11065,11 @@ python3-usb:
   nixos: [python3Packages.pyusb]
   openembedded: [python3-pyusb@meta-python]
   ubuntu: [python3-usb]
+python3-usd-core:
+  '*':
+    pip:
+      packages: [usd-core]
+  fedora: [python3-usd]
 python3-utm:
   debian:
     '*': [python3-utm]


mjcarroll (approved):

rosdep/python.yaml
```diff
@@ -11065,6 +11065,11 @@ python3-usb:
   nixos: [python3Packages.pyusb]
   openembedded: [python3-pyusb@meta-python]
   ubuntu: [python3-usb]
+python3-usd-core:
```
tfoote:
```suggestion
python3-usd:
```

tfoote (approved):
Approving with one change.

tfoote:
I don't see it yet in the debian packaging requests: https://www.debian.org/devel/wnpp/requested nor in the existing packages. But if it's in Fedora already that would probably be a good thing to put a request in for so that it can be something potentially used in released packages. 

We'll switch to the main rule when it becomes available which I'm pretty confident it will in the future.

PR #49203, opened 2025-12-26
Support libgsl on debian bookworm

Timple:
I thought it was easiest to deprecate wheezy since it's EOL.

Required to fix buildfarm errors with the `mp_units_vendor` package.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -4316,10 +4316,7 @@ libgrpc:
   ubuntu: [libgrpc++-dev]
 libgsl:
   arch: [gsl]
-  debian:
-    buster: [libgsl-dev]
-    stretch: [libgsl-dev]
-    wheezy: [libgsl0-dev]
+  debian: [libgsl-dev]
   fedora: [gsl-devel]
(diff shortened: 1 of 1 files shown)


rosdep/base.yaml
```diff
@@ -4316,10 +4316,7 @@ libgrpc:
   ubuntu: [libgrpc++-dev]
 libgsl:
```
Timple, line 4317: Should this not be called `libgsl-dev` actually? As per ubuntu entry.

mjcarroll: Probably, but we can do that in a follow-up.

mjcarroll (approved):

rosdep/base.yaml
tfoote: Yeah, probably. But as this is an old one. We usually bias towards grandfathering it and maintaining it as is instead of creating churn and breaking people for the sake of "cleanlyness".

tfoote (approved):
Good to have the new stable name everywhere now.

PR #49279, opened 2026-01-05
Update lz4 keys of base.yaml

kshinjo-pfr:
## Package name:

lz4

## Purpose of using this:

I want to add `lz4` command dependency to my package. But current `lz4` key in ubuntu is resolved as `liblz4-dev`, which is actually a library package.
https://github.com/ros/rosdistro/pull/3938

It seems that this keys are already used as dependency to dev package. ( e.g. https://github.com/ros/rosdistro/pull/16915 looks like fixing dependency of lz4 in ros_comm in melodic )

So I will just add command package to the key in addition to existing keys.

  - https://packages.debian.org/trixie/lz4
  - https://launchpad.net/ubuntu/+source/lz4
  - https://packages.fedoraproject.org/pkgs/lz4/lz4/
  - https://software.opensuse.org/package/lz4
  - https://pkgs.org/download/lz4

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7749,16 +7749,16 @@ lua5.2-dev:
 lz4:
   alpine: [lz4-dev]
   arch: [lz4]
-  debian: [liblz4-dev]
-  fedora: [lz4-devel]
+  debian: [liblz4-dev, lz4]
+  fedora: [lz4-devel, lz4]
   freebsd: [liblz4]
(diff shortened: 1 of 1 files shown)


Timple (approved):

Timple:
Makes sense as `liblz4-dev` is already a key as well. So packages can still link to only `liblz4-dev` if they want.

mjcarroll (approved):

tfoote (approved):

PR #49289, opened 2026-01-06
tf_tree_terminal: 2.0.0-2 in 'humble/distribution.yaml', 2.0.0-1 in  'jazzy/distribution.yaml' and 'rolling/distribution.yaml'

Tanneguydv:

HUMBLE, JAZZY, ROLLING

# The source is here:

https://github.com/Tanneguydv/tf_tree_terminal-release

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -9365,7 +9365,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/Tanneguydv/tf_tree_terminal.git
-      version: master
+      version: 2.0.0
     release:
       tags:
         release: release/rolling/{package}/{version}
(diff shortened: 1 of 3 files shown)


rolling/distribution.yaml
```diff
@@ -9361,6 +9361,21 @@ repositories:
...
     status: maintained
+  tf_tree_terminal:
+    doc:
+      type: git
+      url: https://github.com/Tanneguydv/tf_tree_terminal.git
+      version: master
+    release:
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/Tanneguydv/tf_tree_terminal-release.git
```
mjcarroll, line 9372: This needs to be in the ros2-gbp organization so that maintainers can cut releases to new distros (like Lyrical in a few months)

mjcarroll (changes requested):
Please submit these with source/doc tags first and then generate the releases via bloom.

rolling/distribution.yaml
Tanneguydv: Ok, how do I do ?

PR #49332, opened 2026-01-08
Update Python rosdep: Install gRPC from PIP instead of APT

tropappar:
Please update the following dependencies in the rosdep database.

## Package name:

- python3-grpcio / grpcio
- python3-grpc-tools / grpcio-tools

## Package Upstream Source

- https://github.com/grpc/grpc
- https://github.com/grpc/grpc/tree/master/tools/distrib/python/grpcio_tools

## Purpose of using this

The update is needed because the APT version is outdated. See [this protobuf issue](https://github.com/protocolbuffers/protobuf/issues/14372).

- Python
  - https://pypi.org/project/grpcio/
  - https://pypi.org/project/grpcio-tools/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6725,13 +6725,13 @@ python3-grip:
   ubuntu: [grip]
 python3-grpc-tools:
   debian:
-    '*': [python3-grpc-tools]
-    stretch: null
+    pip:
+      packages: [grpcio-tools]
   nixos: [python3Packages.grpcio-tools]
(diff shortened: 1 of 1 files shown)


tropappar:
The gRPC packages depend on python3-typing-extensions / typing_extensions which are already installed as a dependency by several ROS packages (e.g. launch-ros).

So, one need to run rosdep with the PIP ignore installed option:

```bash
 export PIP_IGNORE_INSTALLED=1
```

PR #49333, opened 2026-01-08
Index RQml

StefanFabian:

rqml

# The source is here:

https://github.com/StefanFabian/rqml.git

# OSRF Discourse Release Post

https://discourse.openrobotics.org/t/goodbye-rqt-hello-rqml-new-release/51697

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -8084,6 +8084,12 @@ repositories:
       url: https://github.com/ros2/rpyutils.git
       version: rolling
     status: developed
+  rqml:
+    source:
+      type: git
+      url: https://github.com/StefanFabian/rqml.git
+      version: rolling
+    status: maintained
   rqt:
     doc:
       type: git


emersonknapp (dismissed):
Notes:
* `package.xml`s lists `GPLv3` - which doesn't match the full name or the SPDX identifier. See https://opensource.org/license/gpl-3-0
  * We'll still accept the deprecated `GPL-3.0` but would prefer you commit to the modern language `GPL-3.0-only` 
* https://github.com/StefanFabian/rqml/blob/rolling/rqml_plugin_example/package.xml lists `MIT` license, but this is not an included license file in the repo

StefanFabian:
@emersonknapp Thanks for the review :+1: 
One question regarding the example package. I wanted to include it in the repository to ensure consistency with the library. However, for the example, as it may be copied as a starting point, a more permissive license is preferable.
Is it an issue that the license of that package is more permissive?
How should I properly declare that?
Add a license file to the package subfolder?

clalancette (approved):

  - [x] Top level license file: https://github.com/StefanFabian/rqml/blob/rolling/LICENSE
  - [x] Per package license files: https://github.com/StefanFabian/rqml/blob/rolling/rqml_plugin_example/LICENSE
- [x] License is [OSI-approved](https://opensource.org/licenses): GPLv3/MIT
- [x] Public source repo: https://github.com/StefanFabian/rqml
- [x] Each package meets [REP-144](https://reps.openrobotics.org/rep-0144/) naming conventions

PR #49404, opened 2026-01-13
Add tf_tree_terminal source to humble, jazzy, kilted, rolling

Tanneguydv:

- humble 
- jazzy
- kilted
- rolling

# The source is here:
https://github.com/Tanneguydv/tf_tree_terminal

# Checks

### DIFF
--- a/kilted/distribution.yaml
+++ b/kilted/distribution.yaml
@@ -9480,7 +9480,7 @@ repositories:
     source:
       type: git
       url: https://github.com/Tanneguydv/tf_tree_terminal.git
-      version: 2.0.0
+      version: master
     status: developed
   tinyspline_vendor:
     doc:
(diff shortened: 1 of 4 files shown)


kilted/distribution.yaml
```diff
@@ -9476,6 +9476,12 @@ repositories:
       url: https://github.com/DLu/tf_transformations.git
       version: main
     status: maintained
+  tf_tree_terminal:
+    source:
+      type: git
+      url: https://github.com/Tanneguydv/tf_tree_terminal.git
+      version: 2.0.0
```
Yadunund: The version in the `source` block corresponds the development branch and not tag. Looking at the repo I believe this should be `master`. Please update for other distributions. 

```suggestion
      version: master
```

Yadunund (dismissed):

kilted/distribution.yaml
Tanneguydv: Thank you, I will do so.
The error message for the rosdistro checks was not clear to me, thank you for the explanation

mjcarroll (approved):

PR #49418, opened 2026-01-13
Add language-pack- rules for Debian and RHEL

peci1:
Please update the following dependencies in the rosdep database.

## Package name:

language-pack-de
language-pack-en

  - the locales do not have any package, they need to be manually configured; however, Bookworm has by default a C.UTF-8 locale which is quite close to en_US, so I decided to mark -en locale package as satisfied by default. However, a precise solution would be to put `null` even to the english version.
- Fedora:
  - https://fedora.pkgs.org/42/fedora-x86_64/glibc-langpack-de-2.41-1.fc42.x86_64.rpm.html
  - https://fedora.pkgs.org/42/fedora-x86_64/glibc-langpack-en-2.41-1.fc42.x86_64.rpm.html
- rhel:
  - https://rhel.pkgs.org/9/red-hat-ubi-baseos-aarch64/glibc-langpack-en-2.34-231.el9_7.2.aarch64.rpm.html
  - de locale is not supported in RHEL according to pkgs.org

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2607,12 +2607,12 @@ konsole:
   ubuntu: [konsole]
 language-pack-de:
   debian: null
-  fedora: [filesystem]
-  rhel: [glibc-langpack-de]
+  fedora: [glibc-langpack-de]
+  rhel: null
   ubuntu: [language-pack-de]
 language-pack-en:
   debian: []
-  fedora: [filesystem]
+  fedora: [filesystem, glibc-langpack-en]
   rhel: [glibc-langpack-en]
   ubuntu: [language-pack-en]
 lcov:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2608,7 +2608,7 @@ konsole:
 language-pack-de:
   debian: null
   fedora: [glibc-langpack-de]
-  rhel: null
+  rhel: [glibc-langpack-de]
   ubuntu: [language-pack-de]
 language-pack-en:
   debian: []
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2612,7 +2612,7 @@ language-pack-de:
   ubuntu: [language-pack-de]
 language-pack-en:
   debian: []
-  fedora: [filesystem, glibc-langpack-en]
+  fedora: [glibc-langpack-en]
   rhel: [glibc-langpack-en]
   ubuntu: [language-pack-en]
 lcov:
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -2606,10 +2606,14 @@ konsole:
   nixos: [plasma5Packages.konsole]
   ubuntu: [konsole]
 language-pack-de:
+  debian: null
   fedora: [filesystem]
+  rhel: [glibc-langpack-de]
   ubuntu: [language-pack-de]
 language-pack-en:
+  debian: []
   fedora: [filesystem]
+  rhel: [glibc-langpack-en]
   ubuntu: [language-pack-en]
 lcov:
   arch: [lcov]


ahcorde (approved):

rosdep/base.yaml
```diff
@@ -2606,10 +2606,14 @@ konsole:
   nixos: [plasma5Packages.konsole]
   ubuntu: [konsole]
 language-pack-de:
-  fedora: [filesystem]
+  debian: null
+  fedora: [glibc-langpack-de]
+  rhel: null
```
cottsay: > de locale is not supported in RHEL according to pkgs.org

It's part of the `baseos` repo. I don't know where pkgs.org gets its data, but it's there.

https://repo.almalinux.org/almalinux/9/BaseOS/x86_64/os/Packages/glibc-langpack-de-2.34-231.el9_7.2.x86_64.rpm


```suggestion
  rhel: [glibc-langpack-de]
```

rosdep/base.yaml
```diff
@@ -2606,10 +2606,14 @@ konsole:
   nixos: [plasma5Packages.konsole]
   ubuntu: [konsole]
 language-pack-de:
-  fedora: [filesystem]
+  debian: null
```
cottsay, line 2609: From what I can tell, all of the locales are provided together in Debian under the `locales-all` package: https://packages.debian.org/trixie/locales-all


```suggestion
  debian: [locales-all]
```

cottsay (changes requested):

rosdep/base.yaml
peci1: Well, I'm always super confused regarding RHEL and its clones... so is AlmaLinux the same as RHEL?

rosdep/base.yaml
peci1: locales-all is a 250 MB package... I think very few people want to have that installed...

rosdep/base.yaml
cottsay: In this context, yes. ([there are differences, of course](https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux_derivatives)).

rosdep/base.yaml
cottsay: :shrug: I'd guess I don't know what the use case is. If it's needed for your project to work, it kinda seems like you'd want it installed...

If this is indeed the correct package and you're proposing that we elect not to use it here, we should probably note it somewhere in case someone else wants to use the rule and expects it to work like all the other rosdep rules.

rosdep/base.yaml
```diff
@@ -2606,10 +2606,14 @@ konsole:
...
+  fedora: [glibc-langpack-de]
+  rhel: [glibc-langpack-de]
   ubuntu: [language-pack-de]
 language-pack-en:
-  fedora: [filesystem]
+  debian: []
+  fedora: [filesystem, glibc-langpack-en]
```
cottsay: Seems appropriate to drop `filesystem` from the rule like you did for `language-pack-de`.

```suggestion
  fedora: [glibc-langpack-en]
```

rosdep/base.yaml
peci1: I've created an issue for clarifying the RHEL packages search: https://github.com/ros/rosdistro/issues/49424.

rosdep/base.yaml
peci1: Yes, I also thought about that. Committing.

rosdep/base.yaml
peci1: I don't actually need the -de locale at all :)

It got added in commit https://github.com/ros/rosdistro/commit/6e8c638260ac12da0a7a9eadb6cf6c5c5f5faad7#diff-0765244367c2a04f58d49d5cade813aef00080c9671077c2f5795c95b606ecdeR932

And you added the fedora rule in https://github.com/peci1/rosdistro/commit/d52042d41131a49b1c6ab656ab0576ceeab74ea0 .

Nobody touched it since.

Now I needed to make the -en rule working, so I also thought I could fix the -de locale. But if there is a problem with it, I can drop -de from this PR and keep only the -en part.

The use-case I have for -en is to have a system with en_US.UTF-8 locale installed, or a non-minimalistic C.UTF-8 . That's in order to use iconv transliteration, which doesn't work with the default minimal C locale. And the images used in CI only contain this minimal C locale by default.

I don't know what's better:

- I want -de locale and rosdep says it doesn't have it
- I want -de locale and I get them all (please note that it not only takes up lots of space, but it also makes any APT operation take ages because of locale regeneration (just a theory, I haven't tried it)).

cottsay: I see, thanks for the context.

I'd guess adding the explicit `null` here doesn't actually change the behavior of any of our tools, and would serve as a clue to investigate previous changes for anyone stumbling across it in the future.

Hi future developers! 👋 Good luck!

cottsay (approved):

PR #49555, opened 2026-01-22
synapticon_ros2_control: 0.3.0-1 in 'rolling/distribution.yaml' [bloom]

floweisshardt:
Increasing version of package(s) in repository `synapticon_ros2_control` to `0.3.0-1`:

- upstream repository: https://github.com/synapticon/synapticon_ros2_control.git
- release repository: https://github.com/synapticon/synapticon_ros2_control-release.git
- bloom version: `0.13.0`
- previous version for package: `0.1.2-1`

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -9202,7 +9202,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/synapticon/synapticon_ros2_control.git
-      version: main
+      version: rolling
     release:
       tags:
         release: release/rolling/{package}/{version}
@@ -9211,7 +9211,7 @@ repositories:
     source:
       type: git
       url: https://github.com/synapticon/synapticon_ros2_control.git
-      version: main
+      version: rolling
     status: maintained
   sync_tooling_msgs:
     doc:
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -9206,8 +9206,8 @@ repositories:
     release:
       tags:
         release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/synapticon_ros2_control-release.git
-      version: 0.1.2-1
+      url: https://github.com/synapticon/synapticon_ros2_control-release.git
+      version: 0.3.0-1
     source:
       type: git
       url: https://github.com/synapticon/synapticon_ros2_control.git


rolling/distribution.yaml
```diff
@@ -9202,16 +9202,16 @@ repositories:
...
       type: git
       url: https://github.com/synapticon/synapticon_ros2_control.git
-      version: main
+      version: rolling
     release:
       tags:
         release: release/rolling/{package}/{version}
-      url: https://github.com/ros2-gbp/synapticon_ros2_control-release.git
-      version: 0.1.2-1
+      url: https://github.com/synapticon/synapticon_ros2_control-release.git
```
christophebedard, line 9209: Can you explain why you made this change? We prefer using a release repo under ros2-gbp.

https://docs.ros.org/en/rolling/How-To-Guides/Releasing/Release-Team-Repository.html#what-if-my-existing-release-repo-isn-t-on-ros2-gbp

christophebedard (changes requested):

rolling/distribution.yaml
mjcarroll: For rolling, it's not just preferred, it's required.

floweisshardt: it's the same repo as we're hosting the other release distro versions. We'll investigate switching to ros2-gbp

floweisshardt: reference to https://github.com/ros/rosdistro/pull/49553 (humble) and https://github.com/ros/rosdistro/pull/49554 (jazzy)

ahcorde: any updates?

emersonknapp: It seems that you already have the `ros2-gbp` repository for these packages, it doesn't seem to be a matter of "switching to ros2-gbp", but actually a question of why you're trying to switch off of it in this version bump.

As @christophebedard and @mjcarroll noted, the Rolling distribution release is only supported for `ros2-gbp`-backed packages, but if you do have a reason to maintain your own separate release repositories, you are not prevented in any way from releasing to the stable annual distributions. The only thing you lose out on is automatic inclusion when the new distro snapshot is cut from Rolling.

If you do want to use exclusively the external repository and unlist from Rolling, you can remove the entry here. Or, I would encourage you to re-bloom to the existing `ros2-gbp` repository to get a new version into Rolling.

Unfortunately a deal of nuance there, and ultimately the choice is yours as the package maintainer.

mjcarroll (approved):

PR #49716, opened 2026-02-06
Fixed upstream branches for twist_mux (Jazzy/Kilted/Rolling)

peci1:
Fixing upstream branches of repo twist_mux for Jazzy/Kilted/Rolling.

The branch used by rosdistro before this change was foxy-devel, which [however only has versions up to 4.0.1](https://github.com/ros-teleop/twist_mux/blob/foxy-devel/package.xml). But if you look in binary versions for J/K/R, they are 4.4.0, which is only provided by the rolling branch.

@bmagyar please check this change makes sense.

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -11478,7 +11478,7 @@ repositories:
     doc:
       type: git
       url: https://github.com/ros-teleop/twist_mux.git
-      version: foxy-devel
+      version: rolling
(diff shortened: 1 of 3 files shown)


bmagyar (approved):
Thanks for fixing this! I've also just checked `humble` and that one looks correct.

ahcorde (approved):

PR #49855, opened 2026-02-14
Add `libopencv-highgui` and `libopencv-videoio` dependencies

InvincibleRMC:

## Package name:
`libopencv-highgui` and `libopencv-videoio`

## Package Upstream Source:

https://github.com/opencv/opencv

## Purpose of using this:

Want to use these packages

Distro packaging links:

  - REQUIRED
  - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5306,6 +5306,32 @@ libopencv-imgproc-dev:
   openembedded: [opencv@meta-oe]
   rhel: [opencv-devel]
   ubuntu: [libopencv-imgproc-dev]
+libopencv-highgui:
+  debian:
+    bookworm: [libopencv-highgui406]
+    bullseye: [libopencv-highgui45]
+    trixie: [libopencv-highgui410]
+  fedora: [opencv-highgui]
+  rhel:
+    '*': [opencv-highgui]
+    '8': [opencv-core]
+    '9': [opencv-core]
+  ubuntu:
+    jammy: [libopencv-highgui4.5d]
+    noble: [libopencv-highgui406t64]
+libopencv-videoio:
+  debian:
+    bookworm: [libopencv-videoio406]
+    bullseye: [libopencv-videoio45]
+    trixie: [libopencv-videoio410]
+  fedora: [opencv-videoio]
+  rhel:
+    '*': [opencv-videoio]
+    '8': [opencv-core]
+    '9': [opencv-core]
+  ubuntu:
+    jammy: [libopencv-videoio4.5d]
+    noble: [libopencv-videoio406t64]
 libopenexr-dev:
   arch: [openexr]
   debian: [libopenexr-dev]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -5278,6 +5278,7 @@ libopencv-highgui:
     '9': [opencv-core]
   ubuntu:
     '*': [libopencv-highgui410]
+    focal: null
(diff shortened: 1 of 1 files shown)


MichaelOrlov (approved):

rosdep/base.yaml
```diff
@@ -5266,6 +5266,19 @@ libopencv-dev:
...
+    bookworm: [libopencv-highgui406]
+    bullseye: [libopencv-highgui45]
+    trixie: [libopencv-highgui410]
+  fedora: [opencv-highgui]
+  rhel:
+    '*': [opencv-highgui]
+    '8': [opencv-core]
+    '9': [opencv-core]
+  ubuntu:
+    jammy: [libopencv-highgui4.5d]
```
clalancette, line 5282: While we are here, we should probably add a "catch-all" for the newer distributions:


```suggestion
    '*': [libopencv-highgui410]
    jammy: [libopencv-highgui4.5d]
```

rosdep/base.yaml
```diff
@@ -5266,6 +5266,19 @@ libopencv-dev:
   opensuse: [opencv-devel]
   rhel: [opencv-devel]
   ubuntu: [libopencv-dev]
+libopencv-highgui:
+  debian:
+    bookworm: [libopencv-highgui406]
+    bullseye: [libopencv-highgui45]
+    trixie: [libopencv-highgui410]
```
clalancette: There's a couple different problems that I see here.

The first is that I think bullseye should be `highgui4.5`.  The second is that we should use the `highgui410` for future distributions (I know they keep changing it, but at least for both trixie and forky this name is the same.  So:


```suggestion
  debian:
    '*': [libopencv-highgui410]
    bookworm: [libopencv-highgui406]
    bullseye: [libopencv-highgui4.5]
```

rosdep/base.yaml
```diff
@@ -5306,6 +5319,19 @@ libopencv-imgproc-dev:
   openembedded: [opencv@meta-oe]
   rhel: [opencv-devel]
   ubuntu: [libopencv-imgproc-dev]
+libopencv-videoio:
+  debian:
+    bookworm: [libopencv-videoio406]
+    bullseye: [libopencv-videoio45]
+    trixie: [libopencv-videoio410]
```
clalancette: Same here; I think the bullseye entry is wrong, and we may as well get ready for forky as well:


```suggestion
    '*': [libopencv-videoio410]
    bookworm: [libopencv-videoio406]
    bullseye: [libopencv-videoio4.5]
```

rosdep/base.yaml
```diff
@@ -5306,6 +5319,19 @@ libopencv-imgproc-dev:
...
+  fedora: [opencv-videoio]
+  rhel:
+    '*': [opencv-videoio]
+    '8': [opencv-core]
+    '9': [opencv-core]
+  ubuntu:
+    jammy: [libopencv-videoio4.5d]
```
clalancette, line 5337:
```suggestion
    '*': [libopencv-videoio410]
    jammy: [libopencv-videoio4.5d]
```

clalancette (changes requested):

clalancette (approved):

PR #49894, opened 2026-02-17
Add a pip install option for nanobind

eholum:
The `nanobind-dev` debian is outdated and will not be updated, so this is still installing v1.9.1, whereas nanobind is at version 2.11. The recommended install method from the [maintainers is pip](https://nanobind.readthedocs.io/en/latest/installing.html#install-via-pip-recommended), which is actively released to. In the [base deps](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L8208), `nanobind-dev` is installable from apt everywhere that is available and defaults to pip otherwise.

This change adds a pip only option (`python3-nanobind-pip`) dependency everywhere. This is to support packages that depend on pip installed version, rather than the apt.


MichaelOrlov (approved):

eholum:
> Can you please give some links as to where this is stated?

Sure thing, and sorry had to do a little more digging into why the deb exists when it's not mentioned in nanobind's repo or docs. The package is maintained by the [debian python team](https://salsa.debian.org/python-team/packages/nanobind/-/tree/debian/latest/debian?ref_type=heads) as a downstream consumer of nanobind. So they are pushing new [releases for ubuntu](https://packages.ubuntu.com/search?keywords=nanobind-dev) from time to time,

* Jammy: [doesn't exist](https://packages.ubuntu.com/search?keywords=nanobind-dev)
* Noble (24.04): frozen at [1.9.2](https://packages.ubuntu.com/noble/nanobind-dev)
* plucky/questing: frozen at [2.5.0](https://packages.ubuntu.com/plucky/nanobind-dev)
* resolute: at [2.9.2](https://packages.ubuntu.com/resolute/nanobind-dev)

So why isn't noble in particular being updated despite being ~2 years old? Presumably because nanobind requires a [full ABI transition](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1085019) on every minor bump, so back porting newer versions isn't really going to happen. Seems like there are consistently issues moving it forward on unstable as well,

* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1085019
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1123044
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1127353

> the fact that it is still available in Debian and Ubuntu means we should prefer the system packages, even if they are outdated.

I definitely agree that system packages should be preferred, but in this case the system package either doesn't exist or is ABI-incompatible with what any nanobind 2.x project actually needs. And developers trying to build their packages across distros end up with 2.11 on humble/by default and 1.9 on jazzy/kilted, which are so fundamentally different that we ended up just bypassing rosdep and installing from pip anyway. 

So IMO in this case it would be preferable to deliver the stable releases from the nanobind maintainers, either that or remove the default pip install from rosdep to avoid confusion.

eholum:
Thanks for the info, @clalancette!

> We could remove the pip keys, but I suspect that someone will come back to add them later on. Maybe (ideally?) we could specify a particular version for nanobind for Humble, i.e. the 1.x series.

If possibly that might be the safest option if the intent is to provide something "stable"? The mixing of install mechanisms as it exists currently is what brought me to this problem, and that would have prevented it. Alternatively, though I'm not sure if this would cause more or less confusion... but would adding the pip entry as a python package from non-supported distros make sense? Namely, leave `nanobind-dev` in place in `base.yaml` for the distros that support it, but make an entry in `python.yaml` for `python3-nanobind-pip` for the pip dependency?

emersonknapp (changes requested):
> > Namely, leave `nanobind-dev` in place in `base.yaml` for the distros that support it, but make an entry in `python.yaml` for `python3-nanobind-pip` for the pip dependency?
> 
> We don't typically do that, but I'd be OK with it. I'll put it on the agenda for the ROS PMC meeting.

Not typical, but I'm not sure we have a hard rule on this. A careful review of https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-rules plus "Native packages are strongly preferred. (They are required for packaging and have upgrade and conflict tracking.)" _could_ lead to the conclusion "not both" but it's not the only interpretation. At least `python3-typeguard`/`python3-typeguard-pip` and `python3-smbus2`/`python3-smbus2-pip` provide precedent of having both definitions.

I see that this did make it to the PMC agenda but there wasn't a note made about a resolution so I'm not sure if it was discussed. 

I will conclude tentatively as reviewer: we definitely _do not_ remove the system package definition, but we _may_ add additional `-pip` definition for users who accept the limitations of using it (primarily, not possible to release to rosdistro).

emersonknapp (approved):

rosdep/python.yaml
```diff
@@ -7738,6 +7738,19 @@ python3-myst-parser-pip:
...
+      packages: [nanobind]
+  fedora:
+    pip:
+      packages: [nanobind]
+  osx:
+    pip:
+      packages: [nanobind]
+  ubuntu:
+    pip:
+      packages: [nanobind]
```
cottsay: Let's just use an os-level wildcard here.

```suggestion
  '*':
    pip:
      packages: [nanobind]
```

eholum: Thanks, @cottsay!

sea-bass:
Hello all! I was following the status of this PR since a set of packages we want to get on the buildfarm is dependent on a recent enough version of `nanobind`.

> Missing required distribution links in PR body: Debian, Ubuntu

I'm seeing other entries in this file that do just have a single top-level wildcard (`'*'`) entry. Are these grandfathered in before the automated review rule, or is it appropriate to keep as is?
 
> Found native package(s), prefer native over pip:
>   Found native Debian package: python3-nanobind (https://packages.debian.org/search?keywords=python3-nanobind&searchon=names&suite=all&section=all)

Based on the PMC discussion notes [here](https://discourse.openrobotics.org/t/ros-pmc-minutes-for-march-10-2026/53282), it seems this one was going to move forward as a pip entry rather than a debian entry because the debians have some incorrect versions in some distros?

tfoote:
Closing the loop from the PMC meeting yes this generally does not meet our policy of strongly preferring the system packaged versions. But because this system packaged version unmaintained and wrong we should remove it from the rosdep database and allow this pip package to be used instead. So to move this along please remove the system version and then I think this will be good to go.

clalancette:
> Closing the loop from the PMC meeting yes this generally does not meet our policy of strongly preferring the system packaged versions. But because this system packaged version unmaintained and wrong we should remove it from the rosdep database and allow this pip package to be used instead. So to move this along please remove the system version and then I think this will be good to go.

Hm, is that what we really want to do?  There is a package released in Jazzy, Kilted, and Rolling that depends on the system version (`nanoeigenpy`).  And in Ubuntu 26.04, the system package for `nanobind-dev` is 2.11.0, which seems about right.  So I still stand by my original thought that we should add *both* nanobind (system package) and `python3-nanobind-pip` (pip package), even though that is not our usual thing to do.

eholum:
> So I still stand by my original thought that we should add both nanobind (system package) and python3-nanobind-pip (pip package), even though that is not our usual thing to do.

TBH this makes sense to me, the original issue (and change) was confusion from the fact that the existing `nanobind-dev` key installs from [_both_ apt AND pip](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L8337):

```
nanobind-dev:
  arch: [nanobind]
  debian:
    '*': [nanobind-dev]
    bookworm:
      pip:
        packages: [nanobind]
  freebsd: [nanobind]
  nixos: [python3Packages.nanobind]
  ubuntu:
    '*': [nanobind-dev]
    focal:
      pip:
        packages: [nanobind]
    jammy:
      pip:
        packages: [nanobind]
```

Our application just needs consistency. So my vote would be to remove the pip entries from `nanobind-dev` and add the `python3-nanobind-pip`. But I get wanting to leave things unbroken and using the apt package where possible so happy to follow whatever you want to do!

(please let me know either way and I can make changes)

sea-bass:
Any updates here on preferences so we can proceed? We have another upcoming release of our project and it would be nice if we could get the rosdep key fix as part of it so it can build correctly on the buildfarm for all distros.

clalancette:
I think that this is good as-is, and I'll add my review as such.  Can you please rebase this onto the latest?  I think that will resolve the CI problem.

clalancette (approved):

PR #50075, opened 2026-03-02
[openembedded] pyflakes: remove the deprecating PYTHON_PN

vermaete:
PYTHON_PN is deprecated in OpenEmbedded/Yocto.
There is no python2 version of the pyflakes package anymore.

@robwoolley: could you ack/nack
Although there is a more recent version of the recipe in meta-python, I would stick for now to the recipe of meta-ros.


emersonknapp (approved):
I don't have a lot of context to evaluate this, but also am not hearing alarm bells. Based on above conversation, @vermaete can you just confirm that you still want to move forward with this PR? I couldn't tell if the actions you identified were separate or a replacement for this.

cottsay (approved):

PR #50112, opened 2026-03-05
feat: add github action to cut release and preserve distribution cache

asymingt:
Duplicates: https://github.com/ros/rosdistro/pull/42984

This adds a new release-cutting GitHub action, which automatically tags a release for us. Specifically, it regenerates the distribution cache tarball for the release, and then pushes an orphaned commit that updates the keys in `index-v4.yaml`and `index.yaml` to point to the release artifact in GitHub. Finally it tags this commit, an bundles the updated yamls and distribution cache alongside it as artifacts. Example:

Screenshot 2026-04-08 at 10 52 08 AM

The action is triggered manually through the github Action tab, by clicking on `Cut a new ROS release` and then using the drop-down `manual` menu item to select the distribution and date stamp.

Screenshot 2026-04-08 at 9 57 49 AM

Once Lyrical merges I will submit a follow-on PR to automatically manage sync holds for us. Including this as part of this PR is too complicated and hard to review.


.github/workflows/apply_sync_hold.yaml
```diff
@@ -0,0 +1,45 @@
...
+        required: true
+        type: choice
+        options:
+        - 'jazzy'
+        - 'kilted'
+        - 'lyrical'
+        - 'rolling'
+        default: 'rolling'
+jobs:
+  apply-sync-hold:
```
mjcarroll: So in addition to the label, we need to make it so that future PRs that come in during the sync hold window will be marked with the label.  @cottsay and I were discussing keeping the sync status somewhere in the repo that would be set by this action and unset by the following one.  What do you think about that?

asymingt: I've been exploring doing it with repo variables. I have the ability to toggle sync variable states, and I'm working on blocking PRs from merging based on the state value.

asymingt: Decided to avoid scope creep and stick to the original plan. We'll do the sync hold management as a follow-on.

.github/workflows/cut_new_release.yaml
```diff
@@ -0,0 +1,110 @@
...
+    - name: Install extra tools via pip
+      run: |
+        python -m pip install --upgrade pip
+        pip install rosdistro
+
+    # Rebuild the cache from scratch.
+    - name: Rebuild cache
+      run: |
+        cd ${{ env.RELEASE_DISTRIBUTION }}
+        sed -i "s|https://github.com/ros-gbp|https://github.com/ros2-gbp|g" distribution.yaml
```
cottsay, line 86: This workaround can't make it into production. If there's a bug in the index, it should be fixed.

.github/workflows/cut_new_release.yaml
```diff
@@ -0,0 +1,110 @@
...
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+        sed -i "s|${{ env.RELEASE_CACHE_OLD_URL }}|${{ env.RELEASE_CACHE_NEW_URL }}|g" index.yaml
+        sed -i "s|${{ env.RELEASE_CACHE_OLD_URL }}|${{ env.RELEASE_CACHE_NEW_URL }}|g" index-v4.yaml
+        git add index.yaml index-v4.yaml
+        git commit -m "chore: update cache entry for ${{ env.RELEASE_DISTRIBUTION }}"
+        git tag -f ${{ env.RELEASE_TAG }}
+        git push -f origin tags/${{ env.RELEASE_TAG }}
+
+    # Force-push the release.
```
cottsay, line 102: Why are we force-pushing to rosdistro?

.github/workflows/cut_new_release.yaml
asymingt: Great point. I forgot about this step in the workflow. It won't be needed going forward as the GBP repos are correct on master. I needed this workaround when trying to rebuild a distribution cache for an older release in 2025. Fixing the error in the way you suggest would be rewriting git history, which I assume can't happen. Is it ok if I just remove this step completely?

.github/workflows/cut_new_release.yaml
asymingt: To overwrite an existing tag from the past if we choose to update it to include the distribution cache artifacts. If we plan to use this action strictly for future releases then it should be fine to remove the force in this command. Would that work for you?

traversaro:
Thanks a lot @asymingt for working on this, this would be really useful also in the robostack context, xref: 
* https://github.com/RoboStack/vinca/pull/146
* https://github.com/RoboStack/vinca/issues/93

PR #50219, opened 2026-03-12
Add package python3-questionary to python packages

L-Koch:
# Please add the following dependency to the rosdep python database.

## Package name:

questionary

## Package Upstream Source:

https://github.com/tmbo/questionary

## Purpose of using this:

The questionary pip package adds a user friendly and adaptive way to select configuration options. As such, the package will be used to select configurations.

## Link to PyPi

https://pypi.org/project/questionary/

## License

This package is under the MIT License.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9629,6 +9629,10 @@ python3-qtpy:
   rhel:
     '8': [python3-QtPy]
   ubuntu: [python3-qtpy]
+python3-questionary-pip:
+  '*':
+    pip:
+      packages: [questionary]
 python3-quick2wire-api-pip:
   '*':
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -9628,6 +9628,12 @@ python3-qtpy:
   rhel:
     '8': [python3-QtPy]
   ubuntu: [python3-qtpy]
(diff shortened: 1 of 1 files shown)


christophebedard (approved):
Standard pip disclaimer: ROS packages that depend on `pip` keys cannot be released into a ROS distribution.
They can only be depended on by from-source builds.
Because of this, system packages are highly preferred to pip packages.

InvincibleRMC (changes requested):
@christophebedard This should be `python3-questionary`

https://packages.ubuntu.com/noble/python3-questionary

https://fedora.pkgs.org/43/fedora-updates-aarch64/python3-questionary-2.1.0-2.fc43.noarch.rpm.html

It's fine to use pip if you want it on some other distribution.

christophebedard (dismissed):
ah you're right, I didn't check if Ubuntu already provided it. Then this should be a normal/non-pip key pointing to the corresponding system packages.

https://github.com/ros/rosdistro/blob/1301920e4ce4176f8081569045a7373b9783f9ce/REVIEW_GUIDELINES.md#L60

InvincibleRMC (approved):

mjcarroll (approved):

mjcarroll (approved):

PR #50302, opened 2026-03-18
python: add unified-planning pip dependency

TarikViehmann:

## Package name:

unified-planning

## Package Upstream Source:

https://github.com/aiplan4eu/unified-planning

## Purpose of using this:

The unified-planning framework was introduced within the AIPlan4EU project as a state-of-the-art planning framework.
I am preparing a new release of the clips_executive reasoning framework that makes use of unified-planning to handle PDDL planning problems.

https://pypi.org/project/unified-planning/

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -11059,6 +11059,10 @@ python3-unidiff:
   opensuse: [python-unidiff]
   rhel: [python3-unidiff]
   ubuntu: [python3-unidiff]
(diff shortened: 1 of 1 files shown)


InvincibleRMC (approved):
Approved with the standard caveat that pip dependencies cannot be used for releasing packages on the ros build farm.

mjcarroll (approved):

PR #50612, opened 2026-04-08
Change ros_babel_fish source version from 'rolling' to 'kilted'

StefanFabian:
Currently, the automated builds fail because they track rolling rather than the kilted branch, and rolling just had interface changes that are incompatible with kilted.
https://build.ros2.org/job/Kdev__ros_babel_fish__ubuntu_noble_amd64/15/console


StefanFabian:
@ahcorde Could you merge this please? :)
This should only change what the build farm runs on push.
Currently, this sends me failed emails every time I push to rolling.

ahcorde (approved):

PR #50631, opened 2026-04-08
Added new package video_to_image_msg_publisher  on humble

gstavrinos:
## Package name:

video_to_image_msg_publisher

## Package Upstream Source:
https://github.com/gstavrinos/video_to_image_msg_publisher


skyegalaxy (dismissed):
looks like rosdep validation is complaining about your package missing a package.xml. I think your source code, CMakeLists and package.xml need to be pushed down into another folder with the package name for it to pass

StefanFabian:
@skyegalaxy, sorry for intruding. I was curious how an added package could fail validation.
A subfolder shouldn't be necessary, or if it is, that's new. I have multiple packages with a structure like this (e.g., https://github.com/StefanFabian/qml6_ros2_plugin), and it has never caused an issue.
Don't have any idea what the issue is, either, though. Perhaps the second line with `<?xml-model`, but that's grasping at straws.

gstavrinos:
Awaiting confirmation that everything is alright.

PR #50632, opened 2026-04-08
Added new package video_to_image_msg_publisher on jazzy

gstavrinos:
## Package name:

video_to_image_msg_publisher

## Package Upstream Source:
https://github.com/gstavrinos/video_to_image_msg_publisher


skyegalaxy (dismissed):
looks like rosdep validation is complaining about your package missing a package.xml. I think your source code, CMakeLists and package.xml need to be pushed down into another folder with the package name for it to pass

gstavrinos:
Awaiting confirmation that everything is alright.

PR #50770, opened 2026-04-16
Add NixOS key for python3-qt-bindings

wentasah:
This is meant to be used in relation with https://github.com/lopsided98/nix-ros-overlay/pull/833.

Unlike in other distros, on NixOS, we depend on just one "meta package", which will then inject the required dependencies (Qt5 or Qt6) depending on the compilation context. This is explained more in the referenced PR, and the short summary is that on NixOS, we don't use different versions of the underlying distribution so we cannot determine the Qt version based on that.


wjwwood (approved):

wentasah:
Rebased to master to resolve the conflict and remove stale label.

nim65s (approved):

ahcorde (approved):

PR #50797, opened 2026-04-18
rqml: 3.26.41-1 in 'rolling/distribution.yaml' [bloom]

StefanFabian:
Increasing version of package(s) in repository `rqml` to `3.26.41-1`:

- upstream repository: https://github.com/StefanFabian/rqml
- release repository: https://github.com/ros2-gbp/rqml-release.git
- bloom version: `0.13.0`

## rqml

```
* Initial release.
* Contributors: Aljoscha Schmidt, Stefan Fabian
```

## rqml_core

```
* destroy plugin on close
  Matches behavior observed when plugins are restored from a config and
  initialization happens in PluginManager::factoryFn instead of
  createPlugin
* Initial release.
* Contributors: Stefan Fabian, dzajac
```

## rqml_default_plugins

```
* Initial release.
* Contributors: Aljoscha Schmidt, Stefan Fabian
```

## rqml_plugin_example

```
* Initial release.
* Contributors: Stefan Fabian
```

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -8494,6 +8494,20 @@ repositories:
       version: rolling
     status: developed
   rqml:
+    doc:
+      type: git
+      url: https://github.com/StefanFabian/rqml.git
+      version: rolling
+    release:
+      packages:
+      - rqml
+      - rqml_core
+      - rqml_default_plugins
+      - rqml_plugin_example
+      tags:
+        release: release/rolling/{package}/{version}
+      url: https://github.com/ros2-gbp/rqml-release.git
+      version: 3.26.41-1
     source:
       type: git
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

StefanFabian:
@ahcorde The build fails on the build farm because FetchContent does not work there (according to AI, the Debian build process isolates from the internet, so it fails).
Do you have a suggestion on how to solve this consistently with ROS conventions?
Should I create a separate vendor package?
`spdlog_vendor` uses `ament_vendor(...)`, is this the correct approach to create a vendor package?
`tinyxml_vendor`, on the other hand, uses `ExternalProject_Add` though I don't understand why that works and FetchContent doesn't.

PR #50908, opened 2026-04-25
add ros2-perf to rosdistro.

evshary:

## Package name:

perf

## Package Upstream Source:

https://github.com/evshary/ros2-perf

## Purpose of using this:

The package is for an easy benchmark on the current ROS environment, including throughput and latency.
If this can be installed from the official apt repository, it will be easier for people to use.

Distro packaging links:

  - REQUIRED
  - REQUIRED
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - IF AVAILABLE
  - OPTIONAL
  - IF AVAILABLE
  - IF AVAILABLE

humble
jazzy
kilted
lyrical
rolling

# The source is here:

https://github.com/evshary/ros2-perf

# Checks


wjwwood:
Since this is a high profile package name (and ros2cli verb name), I put this on the agenda for the ROS PMC meeting (it's public, at 9am pst on Tuesdays) next week to discuss.

evshary:
Hi @wjwwood, thanks for your comment. I had no chance to attend the meeting. Is there any update from the ROS PMC meeting? Maybe I can rename the package to `bench` if `perf` is controversial to the community. Please let me know your thoughts. Thank you!
BTW, I suppose the CI failure doesn't come from my changes, right?

mjcarroll:
From the meeting, we discussed that since it is a CLI-type package, it should probably follow the convention of ros2foo as the package name.

As far as `perf`, it was a little undecided.  Many people associate `perf` with the linux `perf` tool (https://www.brendangregg.com/perf.html) which allows for deep profiling/introspection of runtime resource usage.  This feels more like a throughput/bandwidth related tool, so the name `perf` could be a bit confusing.

sloretz:
[Holding for Lyrical sync](https://discourse.openrobotics.org/t/preparing-for-lyrical-sync-2026-06-22/55537)

evshary:
@mjcarroll Thank you for the comments!

I understand the concern that `perf` may be confused with the Linux `perf` profiling tool.

To avoid that ambiguity, I plan to rename the package from `perf` to `bench`, since the tool is more focused on ROS 2 latency/throughput benchmarking rather than runtime profiling or resource introspection.

Regarding the CLI naming convention, I would like to clarify one point. Do you mean that this package should be implemented as a `ros2cli` extension, such as `ros2 bench ...`, instead of being invoked through `ros2 run` or `ros2 launch`?

My current intention is for users to run launch files like:

```bash
ros2 launch bench latency.launch.py
```

I prefer `bench` over `ros2bench` because it keeps the user-facing command shorter and cleaner, but I can reconsider if `ros2bench` is strongly preferred by the ROS release team.

evshary:
Hi @wjwwood @mjcarroll please let me know if `bench` works. I can rename it accordingly. Thank you.

Yadunund:
👋🏼 `rosdsitro` reviewer for the week here

I have a slightly different take here. 

We already have https://gitlab.com/ApexAI/performance_test indexed and released via the build farm (https://github.com/ros/rosdistro/blob/bdfe0dc83e404d6890d05ecdfeb7bbf8c1b24e80/rolling/distribution.yaml#L5987)  and I see that you've contributed to the upstream package. 

Do you see a path where the tooling in your `ros2-perf` package can go into `performance_test` instead?

evshary:
Hi @Yadunund 

Thanks for your suggestion! The tool is used for a quick benchmark on ROS 2 and has some slight differences from Apex AI tools.
I thought it would be easier for people to download the tool rather than build it themselves. That's the reason I tried to upload the tool.
Given my recent career changes, I won't be able to maintain it as often, so closing it is the better choice.
Thank you all for your attention to the issue!

PR #51195, opened 2026-05-06
Use a custom User-Agent for rosdep_repo_check

cottsay:
It seems that archive.ubuntu.com is now blocking requests which use the default urllib User-Agent.

Should resolve issues like this: https://github.com/ros/rosdistro/actions/runs/25393469511/job/74637085237


claraberendsen (approved):
LGTM!

PR #51286, opened 2026-05-08
[robotraconteur,robotraconteur-companion] Use system packages on resolute and forky

### LINKED ISSUE #50609: Use Ubuntu Resolute packages for robotraconteur and robotraconteur_companion
The `robotraconteur` and `robotraconteur_companion` packages are now included in Ubuntu Resolute as packages `python3-robotraconteur, librobotraconteur-dev` and `librobotraconteur-companion-dev`. The packages are also included in Debian testing release Forky. I need to point `rosdep` to use the OS packages on these distros but still use the ROS packages on other distros. I do not see documentation explaining how to configure this.

How do I configure this?

johnwason:
`robotraconteur` and `robotraconteur_companion` have recently been added to the Ubuntu and Debian system repositories for Resolute and Forky. This PR updates `base.yml` to use the system packages when available.

I am not clear if rosdep will fall back to the ROS packages on Ubuntu Noble and the other Debian/Redhat operating systems that do not have system packages. 

Resolves #50609

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -9753,6 +9753,16 @@ robin-map-dev:
   ubuntu: [robin-map-dev]
 robotino-api2:
   ubuntu: [robotino-api2]
+robotraconteur:
+  ubuntu:
+    resolute: [python3-robotraconteur, librobotraconteur-dev]
+  debian:
+    forky: [python3-robotraconteur, librobotraconteur-dev]
+robotraconteur_companion:
+  ubuntu:
+    resolute: [librobotraconteur-companion-dev]
+  debian:
+    forky: [librobotraconteur-companion-dev]
 rsync:
   arch: [rsync]
   debian: [rsync]
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -9754,15 +9754,15 @@ robin-map-dev:
 robotino-api2:
   ubuntu: [robotino-api2]
 robotraconteur:
-  ubuntu:
-    resolute: [python3-robotraconteur, librobotraconteur-dev]
   debian:
     forky: [python3-robotraconteur, librobotraconteur-dev]
-robotraconteur_companion:
   ubuntu:
-    resolute: [librobotraconteur-companion-dev]
+    resolute: [python3-robotraconteur, librobotraconteur-dev]
+robotraconteur_companion:
   debian:
     forky: [librobotraconteur-companion-dev]
+  ubuntu:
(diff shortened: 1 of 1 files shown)


cottsay (changes requested):
Thanks for the PR, I'm going to block review here until we understand how the obsolescence process will work when replacing the existing ROS packages like this.

@mjcarroll, we may want to add this to the PMC agenda. I only recall this happening once previously.

johnwason:
Keep open I am still waiting for a response

rosdep/base.yaml
```diff
@@ -9753,6 +9753,16 @@ robin-map-dev:
   ubuntu: [robin-map-dev]
 robotino-api2:
   ubuntu: [robotino-api2]
+robotraconteur:
```
tfoote, line 9756: This key should follow the rosdep pattern and match the Ubuntu package name. 

For new things typically you want to label older distros as null and use a wildcard rule to indicate current and latest (aka the ones that don't match the null rules)

rosdep/base.yaml
```diff
@@ -9753,6 +9753,16 @@ robin-map-dev:
...
+  debian:
+    forky: [python3-robotraconteur, librobotraconteur-dev]
+  ubuntu:
+    resolute: [python3-robotraconteur, librobotraconteur-dev]
+robotraconteur_companion:
```
tfoote, line 9761: Same comment down here. Fix the key and use the nulls for all older supported platforms and weildcard for current and expected newer ones so it will work for newer releases without updating here.

tfoote (changes requested):
Closing the loop on how to do the transition from ros package to rosdep key. If you want to be using the exact same codebase I believe that you can do conditional logic in your package.xml dependencies. 

The key to this is that the rosdep key and the package key are different. 

```
<!-- Use the ROS package for before Lyrical -->
  <depend condition="$ROS_DISTRO < 'lyrical'">robotraconteur</depend>

  <!-- Use the rosdep key for Lyrical and anything that comes after it -->
  <depend condition="$ROS_DISTRO >= 'lyrical'">robotraconteur-dev</depend>
```

I don't see examples of the greater than but there's a more fragile logic here using ifs:
https://github.com/OctoMap/octomap_rviz_plugins/blob/00db1ac4d4bd7a6d0efb4b2ef31077344ebd2c8e/package.xml#L19

clalancette:
No response in 2 months, so closing this for now.  Please feel free to reopen and respond to @tfoote 's comments.  Thanks!

PR #51354, opened 2026-05-11
add libsocketcan

ottojo:

## Package Upstream Source:

https://github.com/linux-can/libsocketcan

## Purpose of using this:

Building ROS nodes and device drivers for devices using SocketCAN

  - https://packages.debian.org/de/sid/libsocketcan-dev
  - https://packages.ubuntu.com/noble/libsocketcan-dev
  - https://packages.fedoraproject.org/pkgs/libsocketcan/libsocketcan-devel/
  - not available
  - not available
  - not available
  - not available
  - not available
  - not available
  - https://rhel.pkgs.org/9/epel-x86_64/libsocketcan-devel-0.0.12-3.el9.x86_64.rpm.html

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -7010,6 +7010,10 @@ libsndfile1-dev:
   openembedded: [libsndfile1@openembedded-core]
   rhel: [libsndfile-devel]
   ubuntu: [libsndfile1-dev]
+libsocketcan-dev:
+  debian: [libsocketcan-dev]
+  fedora: [libsocketcan-devel]
+  ubuntu: [libsocketcan-dev]
 libsodium-dev:
   alpine: [libsodium-dev]
   arch: [libsodium]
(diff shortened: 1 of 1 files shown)


mjcarroll (commented):
While you are here, can you add the identified suse key?

ottojo:
> While you are here, can you add the identified suse key?

I added the suggested package name. The website does just say `libsocketcan`, not `libsocketcan-dev`, but i assume that exists if the CI suggests it? Maybe the "-dev" package always exists on suse so it's not listed? I don't have a suse install ready to test this.

cottsay:
Hello, the automation will only suggest packages it actually finds in the repositories, so the `-devel` (not `-dev`) package is probably the right one even if it isn't listed on the packaging dashboard.

cottsay (approved):

mjcarroll (approved):

PR #51532, opened 2026-05-15
Fix depthai_ros_v3 Jazzy source branch

moratom:
This fixes the `depthai_ros_v3` Jazzy rosdistro metadata to point `doc.version` and `source.version` at the correct source branch, `v3_jazzy`.

The entry previously used `v3-jazzy`, but that branch name was a mistake. The v3 source branches use underscores, matching the existing `v3_humble` branch naming, and the Jazzy release work was done from `v3_jazzy`.

The release entry is unchanged and remains at `3.2.1-1`.


moratom:
Bump

PR #51568, opened 2026-05-19
Add datamodel-code-generator dependencies for Ubuntu

arjo129:
## Package name:

datamodel-code-generator 

## Package Upstream Source:

https://github.com/koxudaxi/datamodel-code-generator

## Purpose of using this:

[`rmf_api_msgs` ](https://github.com/open-rmf/rmf_api_msgs/tree/main) uses this for generating python models. So far we have been advising users to manually pip install this package, but since it ships with resolute, it'd be good to ship it directly as part of the package.xml.

Distro packaging links: 

- Debian: https://packages.debian.org/trixie/python3-datamodel-code-generator
- Ubuntu: https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=datamodel-code-generator
- Fedora: Not Available
- Arch: Not Available
- Gentoo: Not Available
- macOS: Not Available
- Alpine: Not Available
- NixOS/nixpkgs: https://search.nixos.org/packages?channel=25.11&query=datamodel-code-generator#show=python312Packages.datamodel-code-generator
- openSUSE: https://software.opensuse.org/package/python-datamodel-code-generator
- rhel: Not Available

# The source is here:

https://github.com/koxudaxi/datamodel-code-generator

# Checks

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -193,8 +193,9 @@ datamodel-code-generator:
         packages: [datamodel-code-generator]
     trixie: [python3-datamodel-code-generator]
   fedora:
-    pip:
-      packages: [datamodel-code-generator]
+    '*':
+      pip:
+        packages: [datamodel-code-generator]
   nixos: [python312Packages.datamodel-code-generator]
   opensuse:
     '*':
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -191,15 +191,12 @@ datamodel-code-generator:
   fedora:
     pip:
       packages: [datamodel-code-generator]
-  gentoo:
-    pip:
-      packages: [datamodel-code-generator]
   nixos: [python312Packages.datamodel-code-generator]
   opensuse: [python-datamodel-code-generator]
   ubuntu:
-    resolute: [python3-datamodel-code-generator]
     pip:
       packages: [datamodel-code-generator]
+    resolute: [python3-datamodel-code-generator]
 dpath-pip:
   ubuntu:
     pip:
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -188,20 +188,23 @@ datamodel-code-generator:
     pip:
       packages: [datamodel-code-generator]
   debian:
-    pip:
-      packages: [datamodel-code-generator]
+    '*':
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

rosdep/python.yaml
```diff
@@ -183,6 +183,26 @@ cython3:
...
   nixos: [python3Packages.cython]
   ubuntu: [cython3]
+datamodel-code-generator:
+  arch:
+    pip:
+      packages: [datamodel-code-generator]
+  debian:
+    pip:
+      packages: [datamodel-code-generator]
+    trixie: [python3-datamodel-code-generator]
```
clalancette, line 195: I'm not convinced that this is the right syntax here.  In all other cases in this file, we have something like:

```
  debian:
    '*':
      pip:
        packages: [datamodel-code-generator]
    trixie: [python3-datamodel-code-generator]
```

I'm going to suggest we switch to that.

clalancette (changes requested):

rosdep/python.yaml
arjo129: Ive updated it. Is this the same for other distros or is it debian only?

clalancette: It should be for all of them.  The rest look right except for `fedora`, so we should probably fix that one as well.

arjo129: [bc2ef25](https://github.com/ros/rosdistro/pull/51568/commits/bc2ef25c59f74d217ad9a3687abc1ca9efd849b6)

rosdep/python.yaml
```diff
@@ -183,6 +183,30 @@ cython3:
   gentoo: [dev-python/cython]
   nixos: [python3Packages.cython]
   ubuntu: [cython3]
+datamodel-code-generator:
+  arch:
+    pip:
+      packages: [datamodel-code-generator]
```
clalancette: Sorry, I just noticed this one as well:


```suggestion
  arch:
    '*':
      pip:
        packages: [datamodel-code-generator]
```

clalancette (changes requested):

clalancette (approved):

PR #51650, opened 2026-05-23
Make python-argparse entry for NixOS empty

wentasah:
This is the same what some other distros like Ubuntu do. This dependency corresponds to the argparse module for Python 2.7, which is no longer present in current Nixpkgs/NixOS.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -641,7 +641,7 @@ python-argparse:
   freebsd: [py27-argparse]
   gentoo: [dev-lang/python]
   macports: [py27-argparse]
-  nixos: [python]
+  nixos: []
   openembedded: []
   opensuse: [python]
   osx:


nim65s (approved):

cottsay:
This entire key should really just be removed. Python 2 hasn't been supported for a long time, and this package was merged into the Python standard library 16 years ago. I'll add a ROS PMC agenda item to discuss dropping all Python 2 rules.

Because the standard library provides it, I think that referencing the new package (python itself) is the _correct_ way to do this. The empty list probably started showing up when even that package was no longer provided by the distribution, though the rule should have been removed instead.

The idea is that someone should expect that rosdep will satisfy a resource (identified by a key) with packages backing the rules for a platform. If there are no packages available to satisfy the resource, there should be no rules for that platform. An empty package list indicates that the resource is implicitly available on all installations of that platform.

tl;dr - the other rules listed here with empty package lists are wrong, and we shouldn't propagate that antipattern.

mjcarroll:
We did discuss this week.  Consensus was to drop python2 keys, but probably do it incrementally (10-20% at a time) just to make sure that we don't break anything.

Issue #51715, opened 2026-05-29
Remove EOL Python 2 rosdep keys


mjcarroll:
Based on this: https://github.com/ros/rosdistro/pull/51650

@cottsay proposed that we remove all existing python2 keys from rosdistro, as Python 2.7 is long EOL.
We should do this in batches of approximate 20% of the keys from lowest risk to highest risk of anybody depending on them (though they really, really shouldn't at this point).

## Batches Checklist:
- [x] Stage 1 (20%): #51716
- [x] Stage 2 (40%): #51717
- [x] Stage 3 (60%): #51718
- [ ] Stage 4 (80%): #51719
- [ ] Stage 5 (100%): #51720

PR #51716, opened 2026-05-29
rosdep: drop legacy Python 2 keys - Stage 1 (20%)

mjcarroll:


tfoote (approved):
With us being 5+ years beyond EOL date that's a full Ubuntu LTS later. 

Spot checking these also appear to be approximately unused

PR #51717, opened 2026-05-29
rosdep: drop legacy Python 2 keys - Stage 2 (40%)

mjcarroll:

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -1566,15 +1566,6 @@ python-monotonic:
   nixos: [pythonPackages.monotonic]
   opensuse: [python2-monotonic]
   ubuntu: [python-monotonic]
-python-msgpack:
-  arch: [python2-msgpack]
-  debian: [python-msgpack]
-  fedora: [python-msgpack]
-  gentoo: [dev-python/msgpack]
-  nixos: [pythonPackages.msgpack]
-  openembedded: ['${PYTHON_PN}-msgpack@meta-python2']
-  opensuse: [python2-msgpack]
-  ubuntu: [python-msgpack]
 python-multicast:
   fedora:
     pip: [py-multicast]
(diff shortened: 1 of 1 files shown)


rosdep/python.yaml
```diff
@@ -2623,148 +2142,32 @@ python-qt4-gl:
...
-python-redis:
-  debian: [python-redis]
-  fedora: [python-redis]
-  gentoo: [dev-python/redis-py]
-  ubuntu: [python-redis]
 python-redis-pip:
   ubuntu:
     pip:
       packages: [redis]
-python-requests:
```
tfoote, line 2698: This is the first one I've seen that actually has users, but it's only ROS 1 distros though. We've retired keys in the past so this isn't bround breaking. And confirms that this is all older content.

rosdep/python.yaml
```diff
@@ -3266,14 +2403,6 @@ python-tablib-pip:
...
       packages: [tablib]
-python-tabulate:
-  debian:
-    buster: [python-tabulate]
-    stretch: [python-tabulate]
-  gentoo: [dev-python/tabulate]
-  nixos: [pythonPackages.tabulate]
-  ubuntu:
-    bionic: [python-tabulate]
 python-tabulate-pip:
```
tfoote, line 2406: This one should be stripped too. I'll check that it's in one of the other requests

tfoote: Confirmed in https://github.com/ros/rosdistro/pull/51720/changes#diff-d8721f41777cc41754c7a34c95d1ef9c8f6fbce7575d129bcbcd32383cde667bL997

tfoote (approved):
Looks good to remove these. They're all python2 rules that should get cleared out as non of our platforms support it. I did a spot check and there's a few with legacy dependencies.

PR #51719, opened 2026-05-29
rosdep: drop legacy Python 2 keys - Stage 4 (80%)

mjcarroll:


wjwwood:
Waiting on conflict resolution, I glanced at it, but didn't seem obvious to me, so I'll come back around to it.

sloretz (approved):

ahcorde (approved):

christophebedard:
what's the logic used here? Is this just (blindly?) removing `python-*` keys? Because I've recently seen new `python-*` keys being added

mjcarroll:
We have been adding python keys as `python3` for some time.  I think once this is complete (the 5th PR) we should add a rule to prevent the reintroduction of any new `python-` keys, right?

christophebedard:
That makes sense. Then what's going to happen to these new `python-pathvalidate-pip` & `python-piper-tts-pip` keys? https://github.com/ros/rosdistro/pull/51971. We just keep those but stop accepting new `python-*` keys?

mjcarroll:
Yeah, I meant to get these in much quicker so we wouldn't have the transition period. 

We should not introduce any new python- keys from now on.

mjcarroll:
https://github.com/ros-infrastructure/rosdistro-reviewer/issues/97

christophebedard (approved):

0xjairo:
This appears to affect noetic **Python 3** dependencies declared as `python-`, not just Python 2. Was that the intent?

mjcarroll:
> Was that the intent?

No.  Which keys specifically?

ovaag:
> > Was that the intent?
> 
> No. Which keys specifically?

Examples of some python- keys that resolve to Python 3 packages and broke for us on noetic now after this:
- python-pymavlink → pip: pymavlink. There's no python3-pymavlink successor, It now fails to resolve on noetic with no alternative.
- python-pyzbar-pip → pip: pyzbar. python3-pyzbar was added in #48968, but its ubuntu: focal entry is null, so there's no working pyzbar key on noetic.

0xjairo:
we were affected by these 2:

- `python-pytz-pip`
- `python-progressbar2-pip`

PR #51732, opened 2026-05-31
[openembedded] python3-smbus2: added recipe

vermaete:
@see: https://layers.openembedded.org/layerindex/recipe/60953/

@robwoolley This one was undefined in the generated files


tfoote (approved):
Looks reasonable.

Will need  a second review. PReferably from @robwoolley but this is simple enough anyone will probably be fine.

robwoolley:
> Looks reasonable.
> 
> Will need a second review. PReferably from @robwoolley but this is simple enough anyone will probably be fine.

Looks good to me!

PR #51733, opened 2026-05-31
[openembedded] python3-graphviz: added recipe

vermaete:
See: https://layers.openembedded.org/layerindex/recipe/131361/

@robwoolley: could you review please?  This recipe was undefined in the generated files, and temp. fixed in `ros-distro.inc`.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -6742,7 +6742,7 @@ python3-graphviz:
   fedora: [python3-graphviz]
   gentoo: [dev-python/graphviz]
   nixos: [python3Packages.graphviz]
-  openembedded: [python3graphviz@meta-python]
+  openembedded: [python3-graphviz@meta-python]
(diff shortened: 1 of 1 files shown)


mjcarroll:
This one seems to be failing checks, can you take a look?

rosdep/python.yaml
```diff
@@ -6742,6 +6742,7 @@ python3-graphviz:
   fedora: [python3-graphviz]
   gentoo: [dev-python/graphviz]
   nixos: [python3Packages.graphviz]
+  openembedded: [python3graphviz@meta-python]
```
cottsay: Probably just missing the `-`:
```suggestion
  openembedded: [python3-graphviz@meta-python]
```

tfoote (approved):
lgtm, needs a second preferably @robwoolley

robwoolley:
This looks good to me.

PR #51784, opened 2026-06-02
Add pip entries for dash and dash-cytoscape

Wesdebaas:

## Package name:

dash
dash-cytoscape

## Package Upstream Source:

https://github.com/plotly/dash
https://github.com/plotly/dash-cytoscape

## Purpose of using this:

These are Python packages for building interactive web-based dashboards. They are used by ROS2 packages that provide browser-based visualization dashboards for monitoring robot state. dash-cytoscape specifically was flagged as a missing dependency for a ROS2 package.

    - not available
    - not available
    - not available
    - not available
    - not available
    - not available
    - not available
    - not available
    - not available
    - not available

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -3602,6 +3602,14 @@ python3-cycler:
     '*': [python3-cycler]
     '7': null
   ubuntu: [python3-cycler]
+python3-dash-cytoscape-pip:
+  '*':
+    pip:
+      packages: [dash-cytoscape]
+python3-dash-pip:
+  '*':
+    pip:
+      packages: [dash]
 python3-dataclasses-json:
   debian: [python3-dataclasses-json]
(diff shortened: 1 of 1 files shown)


tfoote (dismissed):
This isn't passing the linter on ordering. 

Can you also confirm that you understand that pip rules can be used in a source workspace, but can never be used in a full release? 

This doesn't look like something expected to be packaged natively.

Wesdebaas:
Confirmed. I understand these pip rules only work for building from source in a workspace and can never be used in a full ROS buildfarm binary release. The packages aren't available natively in any distro, and our use case is source-workspace builds only. The ordering linter is now passing on the latest commits.

mjcarroll (approved):

clalancette (approved):

PR #51785, opened 2026-06-02
Add python3-azure-storage-blob-pip to rosdep/python.yaml

evTessellate:

## Package name:

python3-azure-storage-blob-pip

## Package Upstream Source:

https://pypi.org/project/azure-storage-blob/

## Purpose of using this:

This dependency is being used for uploading data on a remote Azure blob server which is a type of server optimize to store massive amount of unstructured data. This is very relevant in robotics to store data such as bags, videos and other high volume data. This is why I think it's valuable to be added to the rosdep database.

Distro packaging links: https://pypi.org/project/azure-storage-blob/

- pypi.python.org
https://pypi.org/project/azure-storage-blob/

# Documentation
https://learn.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob?view=azure-python

# The source is here:

https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/storage/azure-storage-blob/azure/storage/blob

# Checks
 - [x] All packages have a declared license in the PyPi.org
 - [x] This repository has a LICENSE file in the PyPi.org in its source code
 - [x] This pip package was tested via rosdep resolve, rosdep install and via the pytest of the rosdistro.

### DIFF
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -4235,7 +4235,7 @@ python3-awsiotsdk-pip:
   ubuntu:
     pip:
       packages: [awsiotsdk]
-python3-azure-storage-blob:
+python3-azure-storage:
   debian: [python3-azure-storage]
   fedora: [python3-azure-storage-blob]
   ubuntu: [python3-azure-storage]
--- a/rosdep/python.yaml
+++ b/rosdep/python.yaml
@@ -101,16 +101,6 @@ azure-mgmt-storage-pip:
   ubuntu:
     pip:
       packages: [azure-mgmt-storage]
-azure-storage-blob-pip:
-  debian:
-    pip:
-      packages: [azure-storage-blob]
-  fedora:
-    pip:
-      packages: [azure-storage-blob]
-  ubuntu:
-    pip:
-      packages: [azure-storage-blob]
 azure-storage-file-share-pip:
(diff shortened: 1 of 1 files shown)


evTessellate:
Hi @mjcarroll, it's my first time contributing to rosdistro. I was wandering what the pending CI checks are waiting for? Is it because it requires an additional reviewer? Thank you and have a nice day!

tfoote (dismissed):
A quick check shows azure-storage-blob being available in native packages. 

https://packages.ubuntu.com/resolute/python3-azure-storage

As such we should use those instead of the pip package. Pip packages cannot be used as dependencies of packages that get released and if installed on the system will cause conflicts with the native ones that are better managed. 

Please update the rules to use native packages on the platforms not the pip entires. The default template had a list of sites to check.

tfoote:
By default CI doesn't run for new contributors as a security precaution on GitHub to make sure not to ddos or potentially enable drive by malicious code in actions from unknown entities.

evTessellate:
@tfoote Thanks for pointing out the native packages. I must confess that I didn't look very far in the native packages, since I saw `pip` instructions. I will be more alert next time.

mjcarroll (approved):

ahcorde (approved):

PR #51790, opened 2026-06-02
Add jpype1 package for python3-jpype-pip

Rezenders:

## Package name:

JPype1

## Package Upstream Source:

[link to source repository](https://github.com/jpype-project/jpype)
[link to pip repository](https://pypi.org/project/jpype1/)

## Purpose of using this:

This package is used to import Java libraries into Python.

Distro packaging links:

  - https://pypi.org/project/jpype1/


tfoote (changes requested):
jpype is packaged as a native package: https://packages.ubuntu.com/resolute/python3-jpype

That should be used instead of pip where available. And only fallback to pip on older unsupported platforms.

PR #51951, opened 2026-06-09
Add chafa rosdep key

codevilot:
Adds a rosdep key for the chafa terminal graphics tool.\n\nThis is needed by echo_graphic so the runtime dependency can be declared in package.xml and installed automatically by standard ROS dependency tooling.\n\nMappings added:\n- arch: chafa\n- debian: chafa\n- fedora: chafa\n- rhel: chafa\n- ubuntu: chafa

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -551,6 +551,12 @@ cgal-qt5-dev:
   fedora: [CGAL-devel]
   gentoo: ['sci-mathematics/cgal[qt5]']
   ubuntu: [libcgal-qt5-dev]
+chafa:
+  arch: [chafa]
+  debian: [chafa]
+  fedora: [chafa]
+  rhel: [chafa]
+  ubuntu: [chafa]
 checkinstall:
   debian: [checkinstall]
   nixos: [checkinstall]


mjcarroll (approved):

codevilot:
The current required rosdep check appears to be failing on the existing `libgstreamer-plugins-good1.0-0` rule for Ubuntu Resolute, rather than on the newly added `chafa` rule.

The `chafa` rule itself passes the automated rosdep review and YAML lint checks, so I’ll leave the current change as-is unless additional platform mappings or other updates are requested.

cottsay:
> The current required rosdep check appears to be failing on the existing `libgstreamer-plugins-good1.0-0` rule for Ubuntu Resolute, rather than on the newly added `chafa` rule.

<details>

There's a known bug in some of the legacy check automation which considers the full "diff" from your branch HEAD to the target branch HEAD, meaning that any lines which were removed from the target since your branch was last synchronized with the target are seen as "added" in your branch. Again, it's a bug, and it's not present in the latest generation of the tooling (rosdistro-reviewer).

Brand new authors require that maintainers "authorize" the PR build CI, and this latency makes it more likely that new changes land in the target branch before the automation can run.

</details>

I just merged `master` back into your branch and immediately triggered CI, and it looks like things are passing now.

skyegalaxy (approved):

PR #52171, opened 2026-06-18
[openembedded] added python3-prettytable

vermaete:
see: https://layers.openembedded.org/layerindex/recipe/105997/

@robwoolley: could you review?

There is in meta-ros-common an older recipe for python3-prettytable.  But this one could be removed (@master) because bitbake isn't using it.


robwoolley:
This commit looks good to me.  I checked if there was a recommended format for the git commit title, but there doesn't seem to be a standard as far as I can see.  This format also matches previous commits for OpenEmbedded that were merged.

Approved.

cottsay (approved):

PR #52331, opened 2026-06-26
Please add the package ias_rviz_plugins to Jazzy Jalisco.

weraleiso:
This package provides a collection of RViz plugins for the InterActiveSpace (IAS). The package contains classical and new, ROS1 packages that were recently ported to ROS2 Jazzy Jalisco. As suggested by [#1783 (comment)](https://github.com/ros2/rviz/issues/1783#issuecomment-4669018555), the authors would like to make this collection available to the community as release package. 

Jazzy Jalisco

# The source is here:
https://github.com/weraleiso/ias_rviz_plugins

# Checks

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -4399,6 +4399,29 @@ repositories:
       url: https://github.com/p2p-industries/hyveos_ros_msgs.git
       version: main
     status: developed
+  ias_rviz_plugins:
+    doc:
+      type: git
+      url: https://github.com/weraleiso/ias_rviz_plugins.git
+      version: jazzy
+    release:
+      packages:
+      - ias_rviz_plugins
+      - rviz_plugin_display_lighting
+      - rviz_plugin_display_particles
+      - rviz_plugin_display_projector
+      - rviz_plugin_display_stereoscopic
+      - rviz_plugin_view_animated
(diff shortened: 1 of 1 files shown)


jazzy/distribution.yaml
```diff
@@ -4433,6 +4433,29 @@ repositories:
...
+      version: jazzy
+    release:
+      packages:
+      - ias_rviz_plugins
+      - rviz_plugin_display_lighting
+      - rviz_plugin_display_particles
+      - rviz_plugin_display_projector
+      - rviz_plugin_display_stereoscopic
+      - rviz_plugin_view_animated
+      - rviz_plugin_view_animated_msgs
```
clalancette, line 4449: Almost all of the rest of our third-party rviz plugins are prefixed with a name; that distinguishes them from "official" rviz plugins.  In this case, your plugins should be named: `ias_rviz_plugin_display_lighting`, etc.

clalancette (changes requested):
Besides the naming problem, this is also failing CI.  Please fix both of the issues.

christophebedard:
Since there's been no activity, I'll close this PR. You can keep commenting and we can always re-open the PR if needed.

In this case, adding an `ias_*` prefix to your packages is an easy way forward.

PR #52341, opened 2026-06-27
[openembedded] python2-backports-ssl: remove recipe

vermaete:
The Python 2 recipe was removed from meta-python:
https://github.com/openembedded/meta-openembedded/commit/05d0c5eee319055816bbea8b9dd972f723f68019

And there is no dependency in meta-ros on this packages (anymore).

@robwoolley: could you verify?


robwoolley:
Confirmed, that recipe is no longer present in supported releases: https://layers.openembedded.org/layerindex/recipe/114330/

Looks good to me.

cottsay (approved):

PR #52396, opened 2026-07-01
Add libqt6-gui dependencies

danthony06:
Adding `libqt6-gui` equivalent dependencies of the existing `libqt5-gui` key.


danthony06:
I'm working on fixing the wrong packages.

danthony06:
I think this is good to go now. I just have the complaint about the key name not matching the package, but that's expected.

mjcarroll (approved):

christophebedard (approved):

PR #52400, opened 2026-07-02
Add Resolute support for gurumdds-3.2

mmminjae:
Add Ubuntu Resolute support for the gurumdds-3.2 rosdep key.

The package is available from our public APT repository and is required by rmw_gurumdds for ROS 2 Lyrical.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -1994,6 +1994,7 @@ gurumdds-3.2:
   ubuntu:
     jammy: [gurumdds-3.2]
     noble: [gurumdds-3.2]
+    resolute: [gurumdds-3.2]
 gv:
   debian: [gv]
   fedora: [gv]


mjcarroll:
I believe that we will have to bring your current package into our package mirror.

Rosdep doesn't allow for keys that are outside of either the Ubuntu repositories or the packages.ros repositories.

CC: @cottsay because I think that you had to do this last time?

clalancette:
Ping @cottsay on this one, because I think we need to pull gurumdds-3.2 into our repositories.

mmminjae:
gurumdds-3.2 packages for Resolute are now available in the ROS public repositories. I believe this should satisfy the prerequisite for adding the corresponding rosdep entry. Thank you for your review.

cottsay (approved):

PR #52640, opened 2026-07-14
Adding libqtgui key for Jammy

danthony06:
Adds a Qt5 package for libqtgui for `jammy` to go with the other distros.


danthony06:
This will make it easier to release code for both Lyrical (Qt6) and Humble, Jazzy, and Kilted which depend on Qt5 because libqtgui can be added as a dependency in a package.xml.

mjcarroll (approved):

skyegalaxy (approved):

PR #52673, opened 2026-07-16
Adding qt-base-dev key for jammy

danthony06:
Adds missing key for `qt-base-dev` for `jammy` so that packages can correctly pick up the Qt5 or Qt6 version of this package, depending on which distro they are running.


danthony06:
I think this is the last of these changes I'm requesting. Sorry to flood you with so many requests.

mjcarroll (approved):

ahcorde (approved):

PR #52678, opened 2026-07-16
Adding RAFCON to index

JohannesErnst:

HUMBLE

# The source is here:

https://github.com/DLR-RM/RAFCON

# Checks
   NOTE: RAFCON comes with a license file for Eclipse Public License - v 1.0


JohannesErnst:
Update: While it was originally tested with `humble` we are now using RAFCON with `jazzy` which also works fine. So maybe the PR should be changed to `jazzy` instaed

PR #52699, opened 2026-07-17
added RAFCON in distribution.yaml

JohannesErnst:

JAZZY and HUMBLE

# The source is here:

https://github.com/DLR-RM/RAFCON.git

# Checks
   NOTE: RAFCON in itself is independent from ROS2 but can (and should) be used complementary. Therefore, it does not have a `package.xml`.

### DIFF
--- a/jazzy/distribution.yaml
+++ b/jazzy/distribution.yaml
@@ -8438,6 +8438,11 @@ repositories:
       url: https://github.com/ros2-gbp/radar_msgs-release.git
       version: 0.2.2-4
     status: maintained
+  rafcon:
+    doc:
+      type: git
+      url: https://github.com/DLR-RM/RAFCON.git
+      version: 2.5.0
   rai_interfaces:
     release:
       tags:


mjcarroll:
  - [x] Top level license file: https://github.com/DLR-RM/RAFCON/blob/master/LICENSE.md
- [x] License is [OSI-approved](https://opensource.org/licenses): https://opensource.org/license/EPL-1.0
- [x] Public source repo: https://github.com/DLR-RM/RAFCON
- [ ] Source repository contains ROS packages

```console
<OUTPUT HERE>
```
</details>

```console
$ find . -name "package.xml" -exec grep --color=auto -e "<license>" "{}" "+"
<OUTPUT HERE>
```
</details>

mjcarroll:
As far as I can tell, this doesn't contain any ROS packages or a `package.xml` file?

JohannesErnst:
@mjcarroll yes that's correct: RAFCON in itself is independent from ROS2 but they can (and should) be used complementary. Therefore, it does not have a `package.xml`. Is this needed to be part in the ros index? Our main motivation is to get more visibility from the ros community.

JohannesErnst:
pinging this again as maybe it fell through the cracks. Is there any more information needed?

mjcarroll:
We actually discussed this in relation to another package in [the weekly PMC meeting](https://discourse.openrobotics.org/t/ros-pmc-minutes-for-july-28-2026/57022/1).

Generally, when a package is indexed, it is expected that it is going to be used by downstream ROS packages.  In order to do that, either:

* the upstream repository would have a `package.xml` file
* It is contributed as a `_vendor` package which brings it into ROS packaging
* A `package.xml` file is injected as part of the bloom process.

I guess this comes down to how you expect it to be used? Does it need to be binary packaged for downstream users?  Do you have examples of it being integrated into ROS nodes?

JohannesErnst:
Thanks for the clarification, I think I understand the expectation better now. Our use case is a little bit different and might therefore not fit so well into the ROS packaging ecosystem: I initially thought the `rosdistro` gathers packages that use ROS, rather than packages that are used by ROS and downstream packages.

RAFCON is a visual flow control tool to program robotic skills and orchestrate interfaces which then talk to modules via middleware. While the middleware of choice is usually ROS (and we provide examples how to start nodes and perform service calls from inside RAFCON), our package is middleware agnostic. Therefore it is also not packaged as a ROS package for now and may indeed not be the best fit for the ROS index in its current form.

Thanks for taking the time to explain the rationale behind the indexing requirements.

PR #53118, opened 2026-08-07
NixOS: Make several Qt entries Qt5/6-independent

wentasah:
On Ubuntu and RHEL, these entries refer to Qt5 or Qt6 depending on the underlying distribution. In Nix, we build all packages against the same distribution (nixos-unstable), so we have to make this distinction differently. We define the rosdep keys to resolve to package set `qt5or6` and this package set is then defined in nix-ros-overlay as an alias of `qt5` or `qt6` depending on ROS distribution. The `qt5or6` package set is introduced in https://github.com/lopsided98/nix-ros-overlay/pull/905 and will be merged after this PR.

This is similar approach to what we used previously for `python3-qt-bindings`.

### DIFF
--- a/rosdep/base.yaml
+++ b/rosdep/base.yaml
@@ -6866,7 +6866,7 @@ libqtcore:
     bookworm: [libqt6core6]
   fedora: [qt6-qtbase]
   gentoo: ['dev-qt/qtcore:6']
-  nixos: [qt6.qtbase]
+  nixos: [qt5or6.qtbase]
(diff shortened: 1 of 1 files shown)


clalancette (commented):
I'm not familiar with nixos, but can you explain why you want to do this?  I would have guessed that was is in there is already basically right, but there may be some nuance of nixos I don't understand.

clalancette (approved):

PR #53152, opened 2026-08-10
Update RHEL 8 dependencies from null to qt5 packages

danthony06:
This updates the Qt packages that map to either Qt5 or Qt6 for RHEL 8 so that generic references can be made to these dependencies for repos that are building for both Qt5 and Qt6.


danthony06:
Looks like https://github.com/ros/rosdistro/pull/53020 already covered this.

Issue #50999, opened 2026-04-30
Old REP-143 link in README.md


mwopfner:
The Link to REP 134 in the README.md is outdated. It points to http://ros.org/reps/rep-0143.html

I think it should now point to: https://reps.openrobotics.org/rep-0143/
This contains the newest version and was updated at Oct 17, 2025

PR #53210, opened 2026-08-13
Update remaining REP links to reps.openrobotics.org

mjcarroll:
Follow-up to #50999 / #53209, which fixed only the README.

The header comments in every `*/distribution.yaml`, `index.yaml`, `index-v4.yaml`, the groovy/hydro/indigo/jade build files, and `releases/targets.yaml` still pointed at the stale `http://ros.org/reps/rep-XXXX.html` URLs. This repoints them at `https://reps.openrobotics.org/rep-XXXX/`, which serves the current revision of each.

Covers REP 3, 141, 143 and 153; I confirmed all four resolve at the new location. Comment-only change across 36 files — no distribution data touched, one line per file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 This repository maintains two independent sets of packaging metadata used in ROS:
 
 1. The lists of repositories that curate ROS packages for each ROS distributions,
-   implementing the data structure defined in [REP 143](http://ros.org/reps/rep-0143.html).
+   implementing the data structure defined in [REP 143](https://reps.openrobotics.org/rep-0143/).
    Any ROS package release will generate pull requests to the distribution files
    in this repository.
 
--- a/ardent/distribution.yaml
+++ b/ardent/distribution.yaml
@@ -1,6 +1,6 @@
 %YAML 1.1
 # ROS distribution file
-# see REP 143: http://ros.org/reps/rep-0143.html
+# see REP 143: https://reps.openrobotics.org/rep-0143/
 ---
 release_platforms:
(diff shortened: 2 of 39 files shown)


tfoote (approved):
Thanks for the cleanup lgtm

tfoote:
This is failing validation because the strings are embedded in the validation/testing in the rosdistro library and that needs to be updated before this will merge cleanly: https://github.com/search?q=repo%3Aros-infrastructure%2Frosdistro+ros.org%2Freps&type=code I ticketed it at: https://github.com/ros-infrastructure/rosdistro/issues/203

mjcarroll:
Holding this until ros-infrastructure/rosdistro#204 is merged and released to PyPI.

The header comment these files carry is both written and verified by the `rosdistro` package — `verify._yaml_header_lines()` hardcodes the `ros.org` URL — so `test_verify_files_identical` fails here against the released package regardless of rebasing.

It also would not have stuck on its own: bloom writes distribution files through `writer.yaml_from_distribution_file`, which prepends that same header, so the next release PR for each distro would have quietly reverted these lines.

\#204 makes the package write the new URL while accepting either, so the two repos do not have to merge in lockstep.

Yadunund:
@mjcarroll is this ready for review now?

mjcarroll:
This is blocked on needing [rosdistro release](https://github.com/ros-infrastructure/rosdistro).  I have flagged it to Scott, he's going to get that out at some point in a coordinated way.

PR #53263, opened 2026-08-14
add example_robot_data on rolling

nim65s:

example_robot_data, a set of robot URDFs for benchmarking and developed examples

# The source is here:

https://github.com/Gepetto/example-robot-data

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -2106,6 +2106,13 @@ repositories:
       url: https://github.com/ros2/example_interfaces.git
       version: rolling
     status: maintained
+  example_robot_data:
+    source:
+      test_commits: false
+      type: git
+      url: https://github.com/gepetto/example-robot-data.git
+      version: devel
+    status: maintained
   examples:
     doc:
       type: git


christophebedard:
I'm slightly concerned about the name because "example robot data" is pretty generic.

Since this seems to come from a research lab, one potential solution would be to prepend the name of the lab to the package name. If you still prefer keeping the name, I can bring it up with the ROS PMC and see what other people think.

---

Another concern is licensing, since this redistributes a bunch of data from other sources. More concretely, I believe we can't accept the CC-BY-NC-ND-4.0 license: we were just recently talking about non-commercial CC licenses and were told we cannot accept it (going forward).

christophebedard (changes requested):
Preventing this from being merged before we discuss the naming and licensing concerns

nim65s:
Thanks for your review.

Yes, this name is generic on purpose, because this package provide a generic collection of robot descriptions (ie. "without xacro") to ease maintenance of tests, benchmarks and showcases in other packages.

I personally don't believe it would really make sense to prepend any entity name here, because this package is not about "the description of the robots from some particular lab", but rather "some common files anybody may need to develop their software", and contributions are welcome, even for assets not used by the main authors.

Still, I understand how this kind of names can be sensitive in rosdistro, and I don't think I have all the elements to take this decision, so I should let you choose. Prepending the github org name would be fine I guess. Awkward and boring on my side, but I can deal with this if necessary.

---

About CC-BY-NC-ND, I have some good news: this information was outdated, and fixed here: https://github.com/Gepetto/example-robot-data/pull/385

christophebedard:
After talking with the rest of the ROS PMC, it sounds like a non-prefixed "example" name is OK. However, "robot data" is still generic. The main title in the README is specifically "Example robot _URDFs_" and all packages are `*_description` packages (which you also said in your comment above), so `example_robot_descriptions` would be more specific.

---

The PMC did wonder about the intention here, though. This is fine licensing/legal-wise, but the README says:

> These source files do not intend to substitute their original repositories."

Then what's the goal of releasing this package on the buildfarm? Personally, I'm worried having copies of files would fracture the community and lead to issues later if an upstream description file gets updated and your copy doesn't. If your goal is aggregation, then a ROS metapackage would be a lot better.

We don't currently have a rule about the content of packages, but we'd like to hear your intentions & thoughts on this.

nim65s:
Thanks again for working on this :)

I believe `example_robot_descriptions` is indeed a better name for this project, thanks for the suggestion. We can easily change the ROS name in upstream package.xml and the key in this PR I think.

Other renamings (at least github name/url, cmake project, c++ header dir, python module, and packages in AUR, conda PyPI & nixpkgs) would be breaking way too much stuff, so this would be a longer term task for us.

---

In order to test an algorithm on some robots, one may want to test it on a bunch of different robots, eg. humanoids, quadrupeds, arms, wheeled, etc. 

Therefore we get software dependencies. This is what this PR is trying to address.

### The role of this project

Example use cases:
- https://github.com/loco-3d/crocoddyl/blob/devel/unittest/bindings/test_impulses.py use hyq (quadruped) & icub (humanoid)
- https://github.com/machines-in-motion/mim_solvers/blob/devel/tests/python/ use solo (quadruped) and ur5 (arm)
- https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/python/bindings_urdf.py use romeo (humanoid) and ur5 (arm)
- https://github.com/humanoid-path-planner/hpp-core/blob/devel/tests/spline-path.cc use romeo (legged humanoid) and baxter (wheeled humanoid)

Btw, baxter_description was available only in hydro, indigo and groovy, and romeo_description  in electric, fuerte, hydro, indigo, jade & kinetic.

Those tests don't really need the upstream updates of the robot descriptions (nor the xacro or gazebo features) that may change between ros distros, and may break tests for bad reasons. Those tests don't even really need the exact eg. baxter or romeo descriptions, just plausible ones. So adding dependencies to baxter_description and romeo_description for one test seems like a huge burden.

Of course, when someone want to make an experiment on a real robot, they must use the correct *_description for their ros distro with the correct xacro / ros2_control / gazebo / etc., that is why this project does not intend to substitute the original repositories.

### The goal of adding this project to the buildfarm

The goal in adding this project on the buildfarm is to improve the packaging, the CI, and the distribution with ROS tools of the packages that are using it for their tests, benchmarks or showcases.

Currently, we can not add eg. `<test_depend>example_robot_data</test_depend>` in the package.xml of a package that is build by the buildfarm, otherwise the buildfarm would fail. Then, if someone want to run `colcon test` with one of those packages in their workspace, they will need to manually install this first, because colcon won't know about this dependency.

We could workaround those issues by adding that  `<test_depend>example_robot_data</test_depend>` upstream, and then patch it out in ros2-gbp, but then we would still have to manually clone it in our dev or CI workspaces.

If we can add this package to the buildfarm, all those use cases will just work out of the box :)

christophebedard:
Thanks for the explanation. I think we're OK with the content.

Now the only blocker is the rename. Once it's renamed to `example_robot_descriptions`, we can merge it. I _think_ you can keep the GitHub repo name as-is but use `example_robot_descriptions` in the `package.xml` and as the key/entry name here.

nim65s:
wow, thanks for the reminder, good bot

PR #53570, opened 2026-08-30
[openembedded] python3-black: move from meta-ros to meta-python

vermaete:
The recipe in meta-python of meta-openembedded is more recent.

@robwoolley I can remove the recipe from meta-ros2, if you agree and once merged into rosdistro and released on Rolling.


christophebedard (approved):
Makes sense to me to just re-use a recipe instead of maintaining one in meta-ros2

mjcarroll (approved):

PR #53847, opened 2026-09-08
rewire_ros: 0.1.1-1 in rolling

alvgaona:

rolling

# The source is here:

Source repository: https://github.com/rewire-run/rewire-ros
Release repository: https://github.com/rewire-run/rewire-ros-release, generated with bloom. 

# Checks

### DIFF
--- a/rolling/distribution.yaml
+++ b/rolling/distribution.yaml
@@ -7210,6 +7210,16 @@ repositories:
       url: https://github.com/ros2/resource_retriever_service.git
       version: rolling
     status: maintained
+  rewire_ros:
+    doc:
+      type: git
+      url: https://github.com/rewire-run/rewire-ros.git
+      version: main
+    source:
+      type: git
+      url: https://github.com/rewire-run/rewire-ros.git
+      version: main
+    status: developed
   rig_reconfigure:
     release:
       tags:


christophebedard (changes requested):
I'm not sure we can take this, since it distributes proprietary software: https://github.com/rewire-run/rewire-ros#license

For one, the `package.xml` should mention your proprietary license as an additional `<license>`. What we do for multi-license packages is to add a `<!-- comment -->` next to each `<license>`, e.g., next to `<license file="LICENSE">Apache-2.0</license>` indicating that it covers this package and next to the other/proprietary `<license>` indicating that it covers rewire itself.

Then, since the license is not OSI-compatible (https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#guidelines-for-package-contents), we can't take it.

AFAIK, the process in this case is to create a formal agreement/license with the OSRF to properly authorize it to distribute your software. If that's something you're interested in, I can point you to the right people.

christophebedard:
I think we can accept your package if you completely remove the option to bundle your proprietary software with the ROS package build (https://github.com/rewire-run/rewire-ros/blob/a2d396d0bab37131a05ccfe2e73330065a7decc4/CMakeLists.txt), which would mean your package stays strictly Apache-2.0, but users have to install the rewire binaries separately. Let me know what you think.

alvgaona:
> I think we can accept your package if you completely remove the option to bundle your proprietary software with the ROS package build (https://github.com/rewire-run/rewire-ros/blob/a2d396d0bab37131a05ccfe2e73330065a7decc4/CMakeLists.txt), which would mean your package stays strictly Apache-2.0, but users have to install the rewire binaries separately. Let me know what you think.

Thanks @christophebedard. 

This was certainly an issue I was expecting, thanks for the confirmation. In terms of the binaries, I could definitely not bundle them and keep the package slim. In the future when I open the codebase I can bundle it. Let me get back to you in a couple days with the package updates.

PR #54044, opened 2026-09-18
Add opengl-es rosdep key

jlblancoc:
## Summary
- Adds a new `opengl-es` rosdep key mapping to Debian/Ubuntu's `libgles-dev`, which ships the `GLES`/`GLES2`/`GLES3` headers and the `glesv2` pkgconfig file.
- Motivated by MRPT's `mrpt_opengl` package: it probes for these headers at its own build time (`CHECK_INCLUDE_FILE`) and bakes the result into an installed public header. Downstream packages building against that header need the same GLES dev headers present in their own build environment, which the existing `opengl` key (mesa GL/GLU only) does not provide. This surfaced as a real build failure once a ROS buildfarm image happened not to have those headers pre-installed by coincidence.
- Only `debian` and `ubuntu` are covered; I verified `libgles-dev` provides the needed headers/pkgconfig on both (including current Ubuntu 26.04). I did not add entries for other platforms since I could not confirm exact package names with confidence (e.g. Arch's `libglvnd` covers GLES2/3 but not the legacy `GLES/gl.h`). Happy to extend if a maintainer can confirm names for other platforms.

## Test plan
- [x] `python3 -m pytest test/rosdep_formatting_test.py` passes locally
- [ ] CI on this PR


jlblancoc:
Good catch — renamed the key to `libgles-dev` to match the Ubuntu package name.

mjcarroll (approved):

PR #54066, opened 2026-09-19
Remove osrf_pycommon and launch_system_modes from rolling

wentasah:
`osrf_pycommon` is being dropped in favor of installing it as a standalone Python package via the system package manager (`python3-osrf-pycommon` rosdep key):. It was already removed from ros2.repos so now, we remove it also from rosdistro. See https://github.com/ros2/launch/pull/817 for details.

`launch` and `launch_ros` packages have already switched away from depending on the ROS-released package, so there are no remaining maintained consumers in rolling.

Also remove `launch_system_modes` and `system_modes_examples`. The former depends on `osrf_pycommon` and the latter on the former. These packages are unmaintained (see https://github.com/micro-ROS/system_modes).


rolling/distribution.yaml
clalancette: Given that the entire repository as marked as unmaintained, and system_modes itself is also failing to build, I'm going to suggest that we just remove this entire block.

clalancette (changes requested):

rolling/distribution.yaml
wentasah: Done.

clalancette (approved):

